PR #24430, opened 2020-05-08
Variables: enables cancel for slow query variables queries

hugohaggmark:
**What this PR does / why we need it**:
This PR tries to handle cancellation of slow running QueryVariable queries when navigating away from a dashboard that contains slow running QueryVariables. I also added a Cancel button during load after X seconds just in case. 

image

That button could need some better wording and design, leave any feedback improvement you have on that. 

If you agree that this is a valid solution I'll add tests for backendSrv, initDashboard and Variables.

**Which issue(s) this PR fixes**:
Relates #24026

**Special notes for your reviewer**:
to test this I've used:
`make devenv sources=prometheus2,slow_proxy_mac`

this makes all request for the `gdev-slow-prometheus` data source take 60s. I'm using the following [dashboard json](https://gist.github.com/hugohaggmark/9158996a57d9d96932ab2ec9695794da) to test this scenario out.

PR #25447, opened 2020-06-08
Dashboard: Fix loading custom home dashboard

torkelo:
In https://github.com/grafana/grafana/pull/24430  I added a check so the loading dashboard did not crash when loading home dashboard call was cancelled. But the check was wrong as it did not account for the scenario when you have a custom home dashboard


hugohaggmark (approved):
Awesome catch!  Are there any tests that need updating?

torkelo:
Updated a test that was wrong and added a new test

PR #25454, opened 2020-06-08
Make tsdb dataframes response able to carry encoded/decoded frames

marefr:
Make tsdb dataframes response able to carry encoded and/or decoded frames

### DIFF
--- a/pkg/services/alerting/conditions/query_test.go
+++ b/pkg/services/alerting/conditions/query_test.go
@@ -55,7 +55,7 @@ func TestQueryCondition(t *testing.T) {
 
 			Convey("should fire when avg is above 100 on dataframe", func() {
 				ctx.frame = data.NewFrame("",
-					data.NewField("time", nil, []time.Time{time.Now()}),
+					data.NewField("time", nil, []time.Time{time.Now(), time.Now()}),
 					data.NewField("val", nil, []int64{120, 150}),
 				)
 				cr, err := ctx.exec()
@@ -75,7 +75,7 @@ func TestQueryCondition(t *testing.T) {
 
 			Convey("Should not fire when avg is below 100 on dataframe", func() {
 				ctx.frame = data.NewFrame("",
-					data.NewField("time", nil, []time.Time{time.Now()}),
+					data.NewField("time", nil, []time.Time{time.Now(), time.Now()}),
 					data.NewField("val", nil, []int64{12, 47}),
 				)
 				cr, err := ctx.exec()
@@ -198,12 +198,8 @@ func (ctx *queryConditionTestContext) exec() (*alerting.ConditionResult, error) 
 	}
 
 	if ctx.frame != nil {
-		bFrame, err := ctx.frame.MarshalArrow()
-		if err != nil {
-			return nil, err
-		}
 		qr = &tsdb.QueryResult{
-			Dataframes: [][]byte{bFrame},
+			Dataframes: tsdb.NewDecodedDataFrames(data.Frames{ctx.frame}),
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/tsdb/models.go
```diff
@@ -85,3 +88,88 @@ func NewTimeSeries(name string, points TimeSeriesPoints) *TimeSeries {
 		Points: points,
 	}
 }
+
+type DataFrames interface {
+	Encoded() ([][]byte, error)
+	Decoded() (data.Frames, error)
+}
+
+type decodedDataFrames struct {
```
kylebrandt, line 97: I think this can be the same struct declaration as encoded frame, but with a different name if you want a different type.

pkg/tsdb/models.go
```diff
@@ -85,3 +88,88 @@ func NewTimeSeries(name string, points TimeSeriesPoints) *TimeSeries {
...
+	decoded data.Frames
+	encoded [][]byte
+}
+
+func NewDecodedDataFrames(decodedFrames data.Frames) DataFrames {
```
kylebrandt, line 102: Perhaps NewDataFrames, with a type switch might be more straight forward when consuming the package?

pkg/tsdb/models.go
```diff
@@ -85,3 +88,88 @@ func NewTimeSeries(name string, points TimeSeriesPoints) *TimeSeries {
...
+	return &decodedDataFrames{
+		decoded: decodedFrames,
+	}
+}
+
+func (ddf *decodedDataFrames) Encoded() ([][]byte, error) {
+	if len(ddf.encoded) == 0 {
```
kylebrandt, line 109: perhaps these should check for nil  and not length of zero? I'm not entirely sure, but perhaps alerting dataframes being nil and length of zero could mean different things.

pkg/tsdb/models.go
```diff
@@ -85,3 +88,88 @@ func NewTimeSeries(name string, points TimeSeriesPoints) *TimeSeries {
 		Points: points,
 	}
 }
+
+type DataFrames interface {
```
marefr, line 92: Would have been nice to have this in SDK/data package and use it in `backend.QueryDataResponse` instead of only encoded arrow bytes. The encoded arrow bytes makes sense sending over the gRPC wire so we want to keep that. But in Grafana we sometime want to just send the encoded arrow bytes straight to the frontend, but in other cases, e.g. alerting, we need the the decoded frames. 

Maybe too late to change that in SDK since it's a breaking change, why this PR suggests adding this in Grafana. Thoughts?

pkg/services/alerting/conditions/query_test.go
```diff
@@ -55,7 +55,7 @@ func TestQueryCondition(t *testing.T) {
 
 			Convey("should fire when avg is above 100 on dataframe", func() {
 				ctx.frame = data.NewFrame("",
-					data.NewField("time", nil, []time.Time{time.Now()}),
+					data.NewField("time", nil, []time.Time{time.Now(), time.Now()}),
```
marefr, line 58: This was an "invalid data frame" before since the length of time and value fields didn't add up. I'm not really sure why this wasn't caught earlier since in my opinion we are doing the same thing as before.

pkg/tsdb/models.go
marefr: How do you mean, taking an interface as argument?

pkg/tsdb/models.go
marefr: Good point. Had that first, but changed when investigating alerting test failure.

pkg/tsdb/models.go
kylebrandt: Yes, where the interface{} will be valid in the type switch at runtime if it is [][]byte or data.Frames

marefr: I rather have typed contracts that shows the intent from usage.

kylebrandt:
I think that is okay then, but should add comments then explaining when to
use which.

kylebrandt (approved):

PR #25466, opened 2020-06-08
Docs: Images in alerts require an external image storage provider

DavidWittman:
Be more explicit about the fact that providing images in alert notifications requires an external image storage provider. Previously this read as if it was only necessary for publicly accessible images, but even if they are served from Grafana, the administrator must set up the "local" image provider in the configuration.

By default, Grafana doesn't have an external image storage provider configured, so images are never uploaded after they are rendered. As a result, the images are not available in alert notifications.

I'm not sure if this is a given, but I spent a good amount of time troubleshooting this in order to get images working in my alert notifications so I figured I'd add it to the docs.


oddlittlebird (approved):
Great improvement, thank you!

PR #21010, opened 2019-12-10
WIP: Panel Communication & New Event Bus Design

torkelo:
We need a way for panels to listen to events and emit events. 

# A
We tried first an approach to use a context to expose an event system using rxjs: #19660 
It felt a bit too complicated to me, and also did not allow for subscribing to specific events without having to filter on type name. 

# B
I then tried a very simple approach wrapping the AppEvent emitter and exposing via these functions added directly top PanelProps. This was nice as this allowed PanelChrome to handle unsubscribing to event on unmount (so panels do not have to remember this). 

```ts
export interface PanelProps<T = any> {
  ....
  subscribeToEvent<P = any>(event: AppEvent<P>, handler: PanelEventHandler<P>): () => void;
  emitEvent<P = any>(event: AppEvent<P>, payload?: P): void;
}
```

# C
Want to try a different approach where instead of passing a event handler the subscribeToEvent function would return an Observable.

https://github.com/grafana/grafana/tree/communication-between-panels-v2

```ts
export interface PanelProps<T = any> {
   ...
  getEventStream<T extends PanelEvent<any>>(eventType: PanelEventType<T>): Observable<T>;
  emitEvent<T>(event: PanelEvent<T>): void;
}
```

This did not add much capability and did not allow PanelChrome anymore to help with unsubscribing, Panel had store subscription & unsub in componentWillUnmount. 

# D

The new typed emitter and the `AppEvent<T>` is great, but the need to use an eventFactory and event variables instead of types feels a bit off. Wanted to try a different event bus typing that allowed a single event type to be used for both subscription and emit. 

The new EventBus, allows you to write code like this:

```ts
 bus.on(LoginEvent, event => {
    console.log(event.payload.logins);
 });

bus.emit(new LoginEvent({ logins: 10 }));
```

See how the LoginEvent type can be used both as the argument in a subscription call, and as a way to create new events. This is made possible thanks to making Events classes. 

Updated Panel Props in this PR solution passes an eventBus (using a simple interface), this is a special sub bus group that records subscriptions and makes sure all subscriptions are unsubscripted when panel unmounts. 

EventBus (interface)
EventBusSrv  (main bus service implementation, uses RxJS Subject)
EventBusGroup (records subscriptions and allows you to unsubscribe to all)

Note to reviews:
All lines that contain CoreEvents.panelChangeView are unrelated to the main change.

PR #25467, opened 2020-06-09
Table: Adds adhoc filtering

### LINKED ISSUE #25072: Table: Add ad-hoc filters support to new table panel
**What happened**:
Can't use the filter feature in table panel

image

Create a table with elasticsearch datasource.

Feature work if I import old dashboards. I can see panel json config : "type": "table-old"

**Environment**:
- Grafana version: 7.0.0
- Data source type & version: elasticsearch
- OS Grafana is installed on: docker ubuntu official image

hugohaggmark:
This PR adds ad hoc filtering to the new React Table component.

Fixes #25072


public/app/plugins/panel/table/TablePanel.tsx
```diff
@@ -62,6 +65,20 @@ export class TablePanel extends Component<Props> {
...
+    const panelModel = getDashboardSrv()
+      .getCurrent()
+      .getPanelById(this.props.id);
+    const datasource = panelModel?.datasource;
+
+    if (!datasource) {
+      return;
+    }
+
+    dispatch(applyFilterFromTable({ datasource, key, operator, value }));
```
hugohaggmark, line 79: Not super happy with this, maybe we should introduce a prop on PanelProps that allows plugins to dispatch actions/events?

packages/grafana-ui/src/components/Table/TableCell.tsx
```diff
@@ -1,38 +1,59 @@
...
+  if (filterable && onFilterAdded) {
+    return (
+      <FilterableTableCell
+        cell={cell}
+        field={field}
+        tableStyles={tableStyles}
+        onFilterAdded={onFilterAdded}
+        cellProps={cellProps}
+      />
+    );
```
hugohaggmark, line 48: Would love to get feedback on this refactor 🙏

dprokop: This sort of relates to my previous comment on how we enable the filtering callback. Since we want to inject domain-specific props to the table cell I think we would have to introduce a render prop for the cell. Then not sure we have/will have a case for filterable prop in near time feature, hence introducing this additional prop feels like introducing unnecessary complexity. WDYT?

packages/grafana-ui/src/components/Table/Table.tsx
```diff
@@ -36,9 +37,10 @@ export interface Props {
   noHeader?: boolean;
   resizable?: boolean;
   initialSortBy?: TableSortByFieldState[];
-  onCellClick?: TableFilterActionCallback;
+  onCellClick?: TableCellClickActionCallback;
   onColumnResize?: TableColumnResizeActionCallback;
   onSortByChange?: TableSortByActionCallback;
+  onFilterAdded?: TableFilterActionCallback;
```
dprokop, line 43: This feels a bit too domain-specific as for a generic table component

hugohaggmark: Not sure what the solution would be otherwise? Maybe we could sync on this or maybe you could elaborate further?

packages/grafana-ui/src/components/Table/TableCell.tsx
```diff
@@ -1,38 +1,59 @@
...
-    onClick = () => onCellClick(cell.column.Header as string, cell.value);
+    return (
+      <ClickableTableCell
+        cell={cell}
+        field={field}
+        tableStyles={tableStyles}
+        onCellClick={onCellClick}
+        cellProps={cellProps}
+      />
+    );
```
torkelo, line 36: I do not think we need this clickable cell and callback anymore, this was just done as simple version of ad-hoc filters, Explore table should use the onFilterAdded callback instead

packages/grafana-ui/src/components/Table/TableCell.tsx
torkelo: not sure what the best solution is, we need this in Explore as well. filterable is a field.config property. think a prop for this on Table is ok. wonder if we should name the callback a bit differently, like onCellFilterAdded , as I think we will add filters on columns as well in the future (think filters in excel on the headers), to filter values in a table in the table component it'self.

packages/grafana-ui/src/components/Table/TableCell.tsx
hugohaggmark: Gotcha, but this means that Explore also would need to implement both `filter for` and `filter out` functionality.

public/app/plugins/panel/table/TablePanel.tsx
torkelo: question is what kind of actions/events? My thinking was to have an event bus on panel props, so panels can subscribe to and listen to events, ala https://github.com/grafana/grafana/pull/21010 
need to pick that PR up again. Got stuck on the event bus design and how to migrate or make it compatible with the panel events & app events (so we do not have two event systems).

hugohaggmark: Ah yes, I remembered that one. Then I will just leave this as is for now.

packages/grafana-ui/src/components/Table/TableCell.tsx
hugohaggmark: I like this suggestion; it makes the api clearer.

hugohaggmark:
@dprokop @torkelo renamed prop to `onCellFilterAdded` and removed the clickable cell. Also, I added the changes to Explore.

packages/grafana-ui/src/components/Table/TableCell.tsx
hugohaggmark: Made changes to Explore to accommodate this new api. Please have a look at changes @kaydelaney @aocenas 🙏

packages/grafana-ui/src/components/Table/TableCell.tsx
dprokop: Yeah, I like the suggestion as well. Think as long as the props are all optional this makes sense.

aocenas (commented):
Code looks good but did not test. Let me know if you need some testing for this.

torkelo:
took this for a spin, works great! awesome!  I changed tooltip position to top so that the first icon (filter for value) tooltip did no cover the right (filter out) icon

torkelo (approved):

hugohaggmark:
> took this for a spin, works great! awesome! I changed tooltip position to top so that the first icon (filter for value) tooltip did no cover the right (filter out) icon

Great catch!

PR #25486, opened 2020-06-09
CloudWatch Logs: "limit" query command now respected

### LINKED ISSUE #25182: CloudWatch Logs: "limit" parameter in Insights query ignored
**What happened**:
When using a simple CloudWatch Insights query with a limit parameter, the limit parameter is ignored, and more results are shown.

Only show as many results as the limit requests.

 Query a log group over a time period that contains more than 20 entries, and use a limit parameter in the query, e.g.
```
fields﻿ ﻿@timestamp﻿,﻿ ﻿@message
|﻿ ﻿sort﻿ ﻿@timestamp﻿ ﻿desc
|﻿ ﻿limit﻿ ﻿20
```

**Environment**:
- Grafana version: Grafana v7.0.1 (ef5b586d7d)
- User OS & Browser: Mac OS 10.15.4, Chrome 83.0.4103.61
- Grafana plugins: NA
- Others: NA

kaydelaney:
Previously the `StartQuery` action was dispatched with a default `limit` parameter of 1000, which seems to override any use of the `limit` command in a CloudWatch Logs query.
The `limit` parameter is now only used in the action payload if explicitly provided.

Closes #25182

### DIFF
--- a/pkg/tsdb/cloudwatch/log_actions.go
+++ b/pkg/tsdb/cloudwatch/log_actions.go
@@ -200,13 +200,19 @@ func (e *CloudWatchExecutor) executeStartQuery(ctx context.Context, logsClient c
 	// so that a row's context can be retrieved later if necessary.
 	// The usage of ltrim around the @log/@logStream fields is a necessary workaround, as without it,
 	// CloudWatch wouldn't consider a query using a non-alised @log/@logStream valid.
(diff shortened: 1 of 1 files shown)


aocenas:
> The limit parameter is now only used in the action payload if explicitly provided.

In what cases do we provide that?

aknuds1 (approved):
Looks good AFAICT.

aknuds1:
@bergquist You know if this is a sensible change?

aknuds1:
@kaydelaney Which milestone should this have? I removed the cherry-pick needed label, since we don't know if it's actually headed for cherry picking.

aocenas (approved):

PR #25490, opened 2020-06-09
Transform: added merge transform that will merge multiple series/tables into one table

### LINKED ISSUE #24891: v7: Tables no longer merge on matching columns
With Grafana v7, I set up two Prometheus queries, like this:

```
sum by (cluster, region, namespace) (kube_pod_container_resource_requests_cpu_cores)
```

```
sum by (cluster, region, namespace) (kube_pod_container_resource_limits_cpu_cores)
```

In previous versions, results were merged on matching tags, which is what I wanted. Now, we have to be explicit about how data is being merged. I don't mind this, but it's extremely limited. The "outer join" transformation only allows us to join on a single label when I want to join on the three I call out in my query.

I can use `label_join` in my PromQL query to create a new label to join on. This works, but then I get joined data where I have two cluster, region, and namespace. I don't have any way to hide one of the duplicate columns.

In my opinion, this is a significant regression and makes tables unusable. Is there any plan to allow for joining time series on multiple columns?

### LINKED ISSUE #24945: Transforms: Time series to rows
This was a transform that existed in the old table panel (think it was the default). 

We need to have something similar implemented as a 7.0 Transformation. 

Takes many time series and concat's them, and adds a Metric field for the time series name. 

image

mckn:
In versions prior to Grafana 7 the table panel had a set of transformers that took multiple queries that returned table data or series data and transformed them into one single table result.

This PR adds that support back to Grafana 7 in the form of a transformer.

Fixes #24945 
Fixes #24891

Will we always have a `field.name`? Please have a look at the https://github.com/grafana/grafana/blob/94c42455474b4bb855fe7ca9a25a2195e8a794af/packages/grafana-data/src/transformations/transformers/merge/dataFrameBuilder.ts and see if I use the field names properly.


docs/sources/panels/transformations.md
```diff
@@ -93,6 +94,28 @@ After I apply the transformation, there is no time value and each column has bee
...
+
+In the example below, we are visualizing multiple queries returing table data before applying the transformation.
+
+{{< docs-imagebox img="/img/docs/transformations/table-data-before-merge-7-1.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+Below you can see the same example but after applying the merge transformer.
+
+{{< docs-imagebox img="/img/docs/transformations/table-data-after-merge-7-1.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+If any the queries returns time series data a `Metric` column containing the name of the query will be added. This value can be customized e.g. by defining `Label` on the source query.
```
dprokop, line 109:
```suggestion
If any of the queries return time series data a `Metric` column containing the name of the query will be added. This value can be customized e.g. by defining `Label` on the source query.
```

packages/grafana-data/src/transformations/transformers/merge/dataFrameBuilder.ts
```diff
@@ -0,0 +1,131 @@
...
+
+type ValueMapper = (frame: DataFrame, valueIndex: number, timeIndex: number) => Record<string, any>;
+
+const fieldNames = {
+  metric: 'Metric',
+  time: 'Time',
+  value: 'Value',
+};
+
+export class DataFrameBuilder {
```
dprokop, line 20: Nit: name of this file should be `DataFrameBuilder` according to https://github.com/grafana/grafana/blob/master/contribute/style-guides/frontend.md#file-and-directory-naming-conventions

mckn:
> I edited the docs in single comments, didn't want to inadvertently approve code too early.

Perfect! As always, great feedback! Thanks 🙏

packages/grafana-data/src/transformations/transformers/merge/DataFrameBuilder.ts
```diff
@@ -0,0 +1,131 @@
...
+  private createValueMapper(): ValueMapper {
+    return (frame: DataFrame, valueIndex: number, timeIndex: number) => {
+      return frame.fields.reduce((values: Record<string, any>, field, index) => {
+        const value = field.values.get(valueIndex);
+
+        if (index === timeIndex) {
+          values[fieldNames.time] = value;
+
+          if (this.displayMetricField) {
+            values[fieldNames.metric] = getFrameDisplayName(frame);
```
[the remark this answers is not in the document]
mckn: Also, you only have one query defined so the transform won't be applied. Do you have time to talk some more about this @torkelo?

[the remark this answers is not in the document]
mckn: ah, now when I look into it a bit more you can ignore my comment above 🙈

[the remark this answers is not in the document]
torkelo: Wondering if in this case where if it's just two fields (time & value), and value field = TIME_SERIES_VALUE_FIELD , then we should keep that name (TIME_SERIES_VALUE_FIELD)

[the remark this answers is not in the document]
mckn: yes, totally agree! Adding tests to reproduce it aws.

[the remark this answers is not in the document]
mckn: So the reason is that the displayName is cached to the previous one. So I will override that one when creating the new dataFrame.

mckn:
> Took this for a spin and works really well!
> 
> Found one thing that looks a bit off, for a time series query & result from Prometheus the value volume looks off
> 
> Screenshot from 2020-06-10 17-02-34

@torkelo I think this is ready for another round of reviewing 🙏

docs/sources/panels/transformations.md
```diff
@@ -93,6 +94,28 @@ After I apply the transformation, there is no time value and each column has bee
 
 {{< docs-imagebox img="/img/docs/transformations/reduce-after-7-0.png" class="docs-image--no-shadow" max-width= "1100px" >}}
 
+### Merge
+
+Use this transformation to _merge_ the result from multiple queries into one single result based on the time field. This is really helpful when using the table panel visualization.
```
oddlittlebird, line 99:
```suggestion
Use this transformation to combine the result from multiple queries into one single result based on the time field. This is helpful when using the table panel visualization. 
```

docs/sources/panels/transformations.md
```diff
@@ -93,6 +94,28 @@ After I apply the transformation, there is no time value and each column has bee
...
 
+### Merge
+
+Use this transformation to _merge_ the result from multiple queries into one single result based on the time field. This is really helpful when using the table panel visualization. 
+
+In the example below, we are visualizing multiple queries returning table data before applying the transformation.
+
+{{< docs-imagebox img="/img/docs/transformations/table-data-before-merge-7-1.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+Below you can see the same example after applying the merge transformation.
```
oddlittlebird, line 105:
```suggestion
Here is the same example after applying the merge transformation.
```

docs/sources/panels/transformations.md
```diff
@@ -93,6 +94,28 @@ After I apply the transformation, there is no time value and each column has bee
...
+
+{{< docs-imagebox img="/img/docs/transformations/table-data-after-merge-7-1.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+If any of the queries return time series data a `Metric` column containing the name of the query will be added. This value can be customized by defining `Label` on the source query.
```
oddlittlebird, line 109:
```suggestion
If any of the queries return time series data, then a `Metric` column containing the name of the query is added. You can customize this value by defining `Label` on the source query.
```

docs/sources/panels/transformations.md
```diff
@@ -93,6 +94,28 @@ After I apply the transformation, there is no time value and each column has bee
...
+
+In the example below, we are visualizing multiple queries returning time series data before applying the transformation.
+
+{{< docs-imagebox img="/img/docs/transformations/time-series-before-merge-7-1.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+Below you can see the same example after applying the merge transformation.
```
oddlittlebird, line 115:
```suggestion
Here is the same example after applying the merge transformation.
```

oddlittlebird (approved):
Docs look good, cannot speak to the code.

torkelo (approved):
Looks great!

PR #25498, opened 2020-06-10
Transformations: Group by and aggregate on multiple fields

### LINKED ISSUE #24716: Display the number of occurence of each value of a field
## What would you like to be added ?

A way to display the number of occurrence of each value of a specified field in my data.

Example of data received:
| timestamp | level | message |
| --- | --- | ---
| 100001 | "warning" | "Warning 401"
| 100002 | "error" | "Error 102"
| 100003 | "info" | "Info message 1"
| 100004 | "error" | "Error 100"
| 100005 | "info" | "Info message 2"
| 100006 | "error" | "Error 100"

From that data, I would like to display the following:

| level | count |
| --- | ---
| "error" | 3 |
| "info"  | 2 |
| "warning" | 1 |

## Why is this needed ?

To answer the question **_How many occurrence of each value of this field is there in the data?_**
I haven't find an easy way to display this info (lead me to it if there is one).

## How should we implement that (suggestion)

I see this very suited to be implemented as a simple [Transformation](https://grafana.com/docs/grafana/latest/panels/transformations/), specifying which field name we want to apply it on.

Totalus:

Adding a new transformation that calculates the number of occurrences of each value of the specified field and displays the count of each value as result. See #24716 for details.

Fixes #24716 

- I based it off of other transformations source code.
- The UI (`OccurrencesTransformerEditor.tsx`) is the same as the _Outer Joint_ transformation as the options for this transformation are the same.
- I modified the doc, but I don't know where to upload the actual images. They don't seem to be stored on the git repo.


torkelo:
Thanks for contributing! 

Wonder if we should turn this into a more general group by transform?

mckn:
Great contribution! I agree with @torkelo regarding the `group by`.

mckn (changes requested):
Awesome initiative and thanks for adding this! I think we should do a couple of changes in this PR to make this feature even more awesome.

Instead of only supporting occurrences I think that changing this one into a proper `groupBy` would be really great. One part of that is already in place and that is the grouping of the field that now is called occurrences.

What we also need to add is the possibility to add any of the other fields and in what way you want to aggregate them. If we take `GROUP BY` in SQL as example you can do something like:

```sql
SELECT COUNT(CustomerID), Country
FROM Customers
GROUP BY Country;
```
That basically mean that you will group your data by country and then also for each group add a column where we count all the Customer ID's for each group. We would like to have something similar for this transformation.

What do you think about changing the editor so you first select what field you want to group by. When that one is selected we could add the possibility to add other fields and how those should be aggregated. We already have the most common aggregation functions in place if you look at the reduce transformer. So those could basically be reused in this scenario.

So to make this feature ready to be merged I would suggest the following things:

1. Rename this transform to `groupBy`.
2. Add the possibility to add any of the other fields and an aggregation column for that field in the editor.
3. Make sure we add a column for the other selected fields and properly aggregate them according to the selected function.

@Totalus what do you think about this? I am happy to help out if you need any help or have any questions.

mckn:
> Hello,
> I understand what you mean. This would allow to do more than just count the number of occurrences, but also calculate the `min`, `max`, `average`, etc. for each different value of the specified field to group by. Smart ! I'll look into it.
> 
> I'm wondering if we could find a better name than _Group by_. For me _Group by_ by itself doesn't mean much, especially since I am not a experienced SQL query writer. Also the `group by` operation seems to always be followed by a calculation function. I find most of the transformations name confusing currently, so I'm trying to find the best self explanatory name for this one. Would it make it clearer for users to name it **Group by and Reduce** ?

Super! This is great 🙌 

Regarding the naming I see your point and I agree and disagree at the same time 🙈 For people used to the concept of `group by` from other products/languages this behavior is expected. So in that sense it is good to follow the classic way of doing a group by. So for them naming it something else might be confusing. 

But for people not used to the details of that concept it might be confusing. That is why we have a description text below the name of the transform. So we can give the user a short description of what the transform does.

So I think naming it group by to start, get some feedback from the community and later change it if needed is the way to go. If that makes sense?

torkelo:
created an issue for the group by: https://github.com/grafana/grafana/pull/25498

Totalus:
Work in progress. I still have a few additional commits to make before pushing for a review.

mckn:
> Work in progress. I still have a few additional commits to make before pushing for a review.

Awesome! No stress, please let me know if I can help out with something :)

Totalus:
Hello all,
I've implemented the beast ! It's flexible and seems to work well.

In short, you specify the **Group By** field, and for any arbitrary number of other fields, you can specify one or multiple calculations to be done on that field.

Its implemented in two steps:

1) Restructure the data by grouping together entries with the same **GroupBy** value
2) Performing all the specified calculations on each group and adding a new column for each calculation

I took advantage of the `reduceField` function. Intuitively I would have done it a bit differently, but it might be a bit more optimize this way as some calculations are performed at the same time in this function. It also reuses more code.

A few question/comments:

- I never really programmed in TypeScript, so I'm not familiar with the good typing practices. There are a two or three typing errors that would need to be arranged. I'm not sure what is the right way to arrange them. Maybe someone could go through and look into this, or let me know how I should arrange them.
- The transformation editor UI could use a bit of tweaking to make it cleaner and nicer. I didn't really put time on the visuals.
- I made some basic documentation. The best in my opinion would be to take captures, but I don't know how to add the images to the repo and I don't have off the shelf a good source of test data to illustrate the various possibilities properly. I could look into this if needed.
- I haven't updated the test yet. I want to get your comments first.

If some of you want to help with this, you're welcome.

Let me know what should be the next steps.

docs/sources/panels/transformations.md
```diff
@@ -222,6 +223,65 @@ After I apply the transformation, my labels appear in the table as fields.
...
+| 2020-07-07 09:30:05 | server 2    | 88              | OK
+| 2020-07-07 09:25:05 | server 2    | 88              | OK
+|
+| 2020-07-07 11:34:20 | ***server 3***    | 62              | OK
+| 2020-07-07 10:31:22 | ***server 3***    | 55              | OK
+| 2020-07-07 09:30:57 | ***server 3***    | 62              | Rebooting
+
+Then you can apply various calculations on each group. The result is reduced to one line per group.
+
+For instance, **On Field** *CPU Temperature*, we can **Calculate** the average. We will obtain the average of each server's CPU Temperature as a result :
```
oddlittlebird, line 263: Why is "On Field" bold? Is this a field name? Per the [Documentation style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-style-guide.md) and [Microsoft style guide](https://docs.microsoft.com/en-us/style-guide/welcome/), field names are bold, not italicized. Do not use bold for emphasis.

Totalus: It's the terminology used in the Transformer Editor UI if you take a look at it. Same with **Calculate** and **Group By**. I wanted to highlight them somehow. I haven't read the Documentation style guide. I'll take a look. *CPU Temperature* is a field name.

I might eventually take more time to do this properly with real screen captures, it would make it much clearer to understand.

mckn:
> Hello all,
> I've implemented the beast ! It's flexible and seems to work well.
> 
> In short, you specify the **Group By** field, and for any arbitrary number of other fields, you can specify one or multiple calculations to be done on that field.
> 
> Its implemented in two steps:
> 
> 1. Restructure the data by grouping together entries with the same **GroupBy** value
> 2. Performing all the specified calculations on each group and adding a new column for each calculation
> 
> I took advantage of the `reduceField` function. Intuitively I would have done it a bit differently, but it might be a bit more optimize this way as some calculations are performed at the same time in this function. It also reuses more code.
> 
> A few question/comments:
> 
> * I never really programmed in TypeScript, so I'm not familiar with the good typing practices. There are a two or three typing errors that would need to be arranged. I'm not sure what is the right way to arrange them. Maybe someone could go through and look into this, or let me know how I should arrange them.
> * The transformation editor UI could use a bit of tweaking to make it cleaner and nicer. I didn't really put time on the visuals.
> * I made some basic documentation. The best in my opinion would be to take captures, but I don't know how to add the images to the repo and I don't have off the shelf a good source of test data to illustrate the various possibilities properly. I could look into this if needed.
> * I haven't updated the test yet. I want to get your comments first.
> 
> If some of you want to help with this, you're welcome.
> 
> Let me know what should be the next steps.

This looks great!! I will do some more testing and have a look at the code a bit more later this week 🙌

packages/grafana-data/src/transformations/transformers/groupBy.ts
```diff
@@ -0,0 +1,175 @@
...
+        for (let [fieldName, calculations] of calculationsByField) {
+          if (fieldName === null || fieldName === groupByFieldName) {
+            continue;
+          }
+
+          // This won't be so intuitive as way to build the result, but if we want to take advantage
+          // of the reduceField function, we'll have to loop on the data this way. We will build a few
+          // fields (columns) at the time, corresponding the each field we want to make some calculations on.
+
+          let calculationResults = {};
```
mckn, line 125:
```suggestion
          let calculationResults: Record<string, any[]> = {};
```

packages/grafana-data/src/transformations/transformers/groupBy.ts
```diff
@@ -0,0 +1,175 @@
...
+          // Now we add the fields to the new fields
+          for (let calc of calculations) {
+            let f = {
+              name: fieldName + ' (' + calc + ')',
+              values: new ArrayVector(calculationResults[calc]),
+              type: FieldType.other,
+              config: {},
+            };
+
+            f.type = guessFieldTypeForField(f);
```
mckn, line 161:
```suggestion
            f.type = guessFieldTypeForField(f) ?? FieldType.string;
```

mckn:
@Totalus awesome work with this PR! I have done some more testing and I think it works as expected and looks good.

## A couple of things before we can get this one merged:

1. Fix the type issues that I have added as comments in this PR.
1. Cleanup code comments and console.logs.
1. Do some small adjustments to the UI (I will attach some suggestions below).

When these things are fixed I think we can merge this PR to master.

-----------------------

## Suggestions on UI updates:

Make sure that we have the proper spacing between the different components (4px). The remove button should also take full height of the row.

Skärmavbild 2020-08-07 kl  11 09 15

Move the "Add Field Calculations" button so it is placed after the "Group By" from group. It should also be changed to a "secondary" variant so it becomes gray-ish.

Skärmavbild 2020-08-07 kl  11 09 15

Simplify the UI by preventing the user from adding the same field multiple times. If you want to add more calculations you should just add them on the existing line.

Skärmavbild 2020-08-07 kl  11 22 26

Totalus:
Great. I'll check this out.

I also wanted to improve the documentation, but I have hard time finding a proper testing data set to showoff the possibilities and take captures... Also, how can I add images to the doc ? Where are they stored ?

oddlittlebird:
Images are stored in a non-public repo. If you want to add images to the doc, please attach them to the PR. We will take them, put them in the appropriate place, and update or create links as needed.

Totalus:
> ## A couple of things before we can get this one merged:
> 
>     1. Fix the type issues that I have added as comments in this PR.
> 
>     2. Cleanup code comments and console.logs.
> 
>     3. Do some small adjustments to the UI (I will attach some suggestions below).
> 
>     4. Add tests that verifies the behavior of the transform.

So, I made quite a few commits. First I worked out the four points mentioned above. Here's a snapshot of the updated UI by the way:

UI snapshot

I also fixed some other issues I found that were causing the transformation to crash in certain conditions.

Side question, is the _TestData DB_ code part of Grafana ? I would make some modification to allow me to build a better data set for testing, and taking captures. Otherwise I would stick with the current doc I guess.

public/app/core/components/TransformersUI/GroupByTransformerEditor.tsx
```diff
@@ -0,0 +1,157 @@
...
+          allowMultiple
+          stats={config[1]}
+          onChange={stats => {
+            onConfigChange([config[0], stats as ReducerID[]]);
+          }}
+          menuPlacement="bottom"
+        />
+      </div>
+      <div className="gf-form">
+        <Button icon="trash-alt" onClick={onDelete} style={{ height: '100%' }} size="sm" variant="secondary" />
```
Totalus, line 57: Are inline stile ok ? I haven't find any `gf-form-*` classes that fit the need.

torkelo: Not sure what your trying to do here, we have not had this need to for height 100% on buttons before

torkelo: will try to test

Totalus: @mckn suggested that the delete (trash) button should take full height of the row:

> ## Suggestions on UI updates:
> 
> Make sure that we have the proper spacing between the different components (4px). The remove button should also take full height of the row.
> 
> Skärmavbild 2020-08-07 kl 11 09 15
>

packages/grafana-data/src/transformations/transformers/groupBy.ts
```diff
@@ -0,0 +1,179 @@
...
+import { DataFrame, FieldType, Field } from '../../types/dataFrame';
+import { DataTransformerInfo } from '../../types/transformations';
+import { getFieldDisplayName } from '../../field/fieldState';
+import { ArrayVector } from '../../vector/ArrayVector';
+import { guessFieldTypeForField } from '../../dataframe/processDataFrame';
+import { reduceField, ReducerID } from '../fieldReducer';
+
+export interface GroupByTransformerOptions {
+  byField: string | null;
+  calculationsByField: Array<[string | null, ReducerID[]]>;
```
torkelo, line 11: this needs to change so we can support multiple group by fields. 

The tuple array type is a bit hard work with. think something like this could be easier and more readable
```
export interface GroupByTransformerOptions {
 fields: Record<string, GroupByFieldOptions>
}

export interface GroupByFieldOptions {
 aggregations: ReducerId[];
 groupBy: boolean; 
}
```

So just single map for field name -> options, and each option object can have a groupBy Boolean to true/false (if true aggregations is ignored/empty)

torkelo (commented):
One big thing missing is the ability to group by multiple fields. 

The UX is also a bit confusing. Think just a list of all fields and toggle button group with "Group by | Aggregate | Ignore"  could make it easier, and if you select Aggregate we show the reducer multi dropdown.

Totalus:
> One big thing missing is the ability to group by multiple fields.

Mmm... I didn't know it was also a possibility to group by multiple fields (never used SQL GroupBy option), but after searching a bit, I understand how it would work. Let me see if I can pull it off.

> The UX is also a bit confusing. Think just a list of all fields and toggle button group with "Group by | Aggregate | Ignore" could make it easier, and if you select Aggregate we show the reducer multi dropdown.

I would not display all the fields by default. I ran into multiple (Github) issues where people are having a lots of fields in their queries, which would quickly fill the transformation editor, making it harder to navigate.

I would consider two options:
- The lazy one (less changes): make the `GroupBy` input selector a multi-select so we can select more than one field.
- The less lazy one: similar to what you suggest: one row per field where you can select `GroupBy` or `Aggregate` and the corresponding reducers, but instead of having all the rows displayed by default, they could be dynamically added (start with one row. When the field name is selected, a new row appears. That or have an Add button to add a new row, either).

To me both options seem functional and decent. At this point its mostly a matter of preferences I believe.

Totalus:
- [x] GroupBy multiple fields

Snapshot

So now we just need to decide what we want to do with the UI. I think the one liner layout would be better if we want to prevent a GroupBy field of also being aggregated. It does not seem to make much sense to apply calculations on one of the GroupBy fields since we know the value and its all the same (for data in the same group).

I'll need some help again for the TS errors.

mckn:
> * [x]  GroupBy multiple fields
> 
> Snapshot
> 
> So now we just need to decide what we want to do with the UI. I think the one liner layout would be better if we want to prevent a GroupBy field of also being aggregated. It does not seem to make much sense to apply calculations on one of the GroupBy fields since we know the value and its all the same (for data in the same group).
> 
> I'll need some help again for the TS errors.

Sweet! Looks really great! I will have a look at the code today and try to help out with the TS errors. hmm, great example above with the multiple group by. Let's see if we can tweak the UI a little bit more.

mckn:
> Great. I'll check this out.
> 
> I also wanted to improve the documentation, but I have hard time finding a proper testing data set to showoff the possibilities and take captures... Also, how can I add images to the doc ? Where are they stored ?

As @oddlittlebird said they are stored in a separate repository. But I can recommend using tables as I did in the following example: https://grafana.com/docs/grafana/latest/panels/transformations/#merge

If you still want to use images you can add them in a comment in this PR and I will add them in the other private repository.

packages/grafana-data/src/transformations/transformers/groupBy.ts
```diff
@@ -0,0 +1,199 @@
...
+        
+          {
+            ["goupByField1Value1","groupByField2Value1", ...] : {fieldName1:Field, fieldname2:Field, ...},
+            ["goupByField1Value2","groupByField2Value2", ...] : {fieldName1:Field, fieldname2:Field, ...},
+            ["goupByField1Value3","groupByField2Value3", ...] : {fieldName1:Field, fieldname2:Field, ...},
+            ...
+          }
+        */
+
+        const groupedData = {};
```
mckn, line 77:
```suggestion
        const groupedData: Record<string, Record<string, MutableField>> = {};
```

The following suggestion should fix the errors with the typings. *Note:* you also need to import the `MutableField` by adding `import { MutableField } from '../../dataframe';`.

public/app/core/components/TransformersUI/GroupByTransformerEditor.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const unusedFieldNameOptions = fieldNames
+    .filter(name => !usedFieldNames.includes(name))
+    .map((item: string) => ({ label: item, value: item }));
+
+  const onSelectField = (value: SelectableValue<string>) => {
+    onChange({
+      ...options,
+      byFields: value.map(i => i.value),
+    });
+  };
```
mckn, line 82:
```suggestion
  const onSelectField = (values: Array<SelectableValue<string>>) => {
    const fields = values.reduce((all: string[], current) => {
      if (!isString(current?.value)) {
        return all;
      }

      all.push(current.value);
      return all;
    }, []);

    onChange({
      ...options,
      byFields: fields,
    });
  };
```

To fix the type error being prompted in this file.

packages/grafana-data/src/transformations/transformers/groupBy.ts
mckn: I agree with @torkelo regarding the structure of the `calculationsByField`. Might be worth having a look at this before we merge this one since it will be stored in the json model of the dashboard.

Totalus: Intuitively I would set `fields` to an Array instead of a Record considering how the UI and the options would be added/removed (on a per line basis) and also how the transform needs to interpret the configuration (builds arrays out of the config options, which is slightly easier starting from an Array, though we agree it can be done fairly easily with both types).

```
export interface GroupByTransformerOptions {
 fields:  GroupByFieldOptions[];
}

export interface GroupByFieldOptions {
 fieldName: string;
 aggregations: ReducerId[];
 groupBy: boolean; 
}
```

Do you see any disadvantage to use the above structure instead ? @mckn @torkelo

mckn: So one advantage of using a `Record<string, GroupByFieldOptions>` is that you don't need to loop through the `options.fields` list to find the field. You can easily just get it by using `options.fields[fieldName]`.

But at the same time this will be pretty small array so I don't think it will have any noticeable benefits. I think you can use your suggested structure above. It makes it easier to argue and read the code then the existing tuple array.

torkelo: The benefits are more in diffs and merges and how it's stored, we sort all props on alphabetical order

mckn: @torkelo great input! I wasn't aware of that 🙌

mckn: So to explain it a bit more in detail. The transformation configuration will be stored in the dashboard `json` data structure that represents the dashboard when it is persisted (saved). When doing changes/comparing this structure it is better to work with a record because of how we store the data.

Totalus:
Here's the new _per-line_ layout (one line per field):

Snapshot

- Lines are dynamically added
- The add button disables when there is no more available fields (could make it disappear maybe, would be cleaner)
- Fields that are `Group By` are automatically positioned at the top
- Fields are NOT sorted in alphabetical orders (I figured the user could want to keep the added order in the result maybe)

@torkelo We can change the labels if needed, that's a minor issue. I used Calculate instead of Aggregate, just because it feels more natural to read **For field [fieldName] calculate [some stats]**, which makes it self explanatory. We can also change the sorting behavior.

packages/grafana-data/src/transformations/transformers/groupBy.ts
Totalus: So currently I'm using the following structure:

```
export enum GroupByOperationID {
  aggregate = 'aggregate',
  groupBy = 'groupby',
}

export interface GroupByFieldOptions {
  fieldName: string | null;
  aggregations: ReducerID[];
  operation: GroupByOperationID;
}

export interface GroupByTransformerOptions {
  fields: Record<string, GroupByFieldOptions>;
  fieldsArray: GroupByFieldOptions[];
}
```

It is much easier to work with an array considering how the UI is made, but if we really prefer a Record, I can convert it to a Record to store it in the transform options and convert it back to an Array when needed. This way it would be stored as a Record in the options.

The main thing to consider here is the ordering. Currently I am assuming the user would set a certain order that he could want to keep in the fields he wants to aggregate, which would not really be guaranteed if we store the info in a Record, but the ordering could be dealed with the Organise Transform instead, so we could just sort it however we want and the user can reorganize it afterwards.

What do you think ? How should we sort in the UI ? @torkelo @mckn

torkelo:
Ah, yes. Order is important. Maybe an array is best

Totalus:
If everybody is happy, I think we're good to go.

mckn:
> If everybody is happy, I think we're good to go.

Awesome work @Totalus! I will do some testing before I give a green light but really love the work you have put into this one.

wohoo

torkelo (commented):
a bit conflicted on the UI and having to add every field manually, not sure if that is good or bad. I think with this model the array storage model is ok, but if we always show a config line for every field then an object (keyed by field name) would make sense again.

Totalus:
> 
> 
> a bit conflicted on the UI and having to add every field manually, not sure if that is good or bad. I think with this model the array storage model is ok, but if we always show a config line for every field then an object (keyed by field name) would make sense again.

From reading a few other issues, I understood that some users have a lot of fields in their data set. Allowing them to select the ones they want to apply the transform on only makes it much cleaner. I also expect that eventually some might ask to support Regex for field matching, as they are for other transformations. In this perspective, dynamically added rows are better.

However this does not mean we need to store the configuration as an array. We could still convert to an Record where the keys are the field names (or eventually the Regex).

As I mentioned previously:
>The main thing to consider here is the ordering. Currently I am assuming the user would set a certain order that he could want to keep in the fields he wants to aggregate, which would not really be guaranteed if we store the info in a Record, but the ordering could be dealed with the Organise Transform instead, so we could just sort it however we want and the user can reorganize it afterwards.

The field order only has an impact on the results column ordering. So at the end, the question of using a Record or an Array to store the configuration comes down to asking: **Do we want to keep the user's row ordering in this transformation ?**

torkelo:
> From reading a few other issues, I understood that some users have a lot of fields in their data set.

Is that not an argument to show all instead of having to add every field one by one? really tricky, depends on how many columns you want do show in the final result

Totalus:
> 
> 
> > From reading a few other issues, I understood that some users have a lot of fields in their data set.
> 
> Is that not an argument to show all instead of having to add every field one by one? really tricky, depends on how many columns you want do show in the final result

I don't know. I'm thinking if I had 50 fields in my data set, I would probably prefer being able to select which one I want to apply the transformation on, otherwise I would have 50 rows filling up my UI, but only actually need to configuring a small number of them. It would force me to scroll down and most of the rows are just useless because I'm not configuring them anyway. I can also easily find the required fields with the auto-completion of the `Select` component. Having fewer rows allows me to get a quick understanding of my transformation configuration at a glance, rather than having to scroll all the rows to see which are configured. That's my personal preference though.

After if you consider that you may have dynamically created fields in your data set and you want to process them with a Regex, then it makes sense to be able to manually enter the field name and dynamically add rows from the UI, rather than having a fix number of rows corresponding to the number of fields.

One other possibility would be to add all the fields (rows) in alphabetical order as default in this same UI. Then you don't need to add them manually, but you can also remove them if you want. Best of both worlds I guess.

torkelo:
Can't you limit the number of fields in your query? or in the filter transform?

Totalus:
> 
> 
> Can't you limit the number of fields in your query? or in the filter transform?

I'm not experiencing this use case personally, and yes you could filter out the fields in another prior transform or in the query itself.

Look, if you need it to be that way, I'll make it that way. At this point its just a mater of personal preferences and I don't want to debate over that. Plus I'm getting a tired of working on this PR.

torkelo:
> Look, if you need it to be that way, I'll make it that way. At this point its just a mater of personal preferences and I don't want to debate over that. Plus I'm getting a tired of working on this PR.

Totally understand that! Sorry, jus trying to refine the UX for this. We have can take over all the remaining tweaks! super grateful for all the work you have put in!

Totalus:
There you go.

snapshot

A remove button also appears at the end of a row if the field is not present anymore in the input data, giving the user the option to remove it.

mckn:
> There you go.
> 
> snapshot
> 
> A remove button also appears at the end of a row if the field is not present anymore in the input data, giving the user the option to remove it.

@Totalus 

Really awesome work! I like the latest version of the UI and think we can merge this one. Get some feedback and change it afterward if needed. I will do a small adjustment to make the UI more intuitive but lets merge it after that one.

Great work! Really appreciate it and sorry for not being so responsive last couple of weeks. Have had focus on another project that have taken a bit to much time.

mckn (approved):
thank you

torkelo:
Awesome! 

The UI editor needs more work, think we need to handle longer field names better (maybe use table layout to get field names to align). And the FieldCalculationsSelector is untyped. And the it should show fields in the same order as the data (currently it sorts field names). 

Would like to test using a radio button group (https://developers.grafana.com/ui/latest/index.html?path=/story/forms-radiobuttongroup--full-width) for the group by | calculate | ignore selector. 

Then there is the question of showing fields that exist in options but not in data, what do you think should the user have to manually remove those? Or automatically when they any option change?

Totalus:
> The UI editor needs more work, think we need to handle longer field names better (maybe use table layout to get field names to align). And the FieldCalculationsSelector is untyped. And think it should show fields in the same order as the data (currently it sorts field names).

@torkelo At this point, I'll let you tweak it up to your taste.

> Would like to test using a radio button group (https://developers.grafana.com/ui/latest/index.html?path=/story/forms-radiobuttongroup--full-width) for the group by | calculate | ignore selector.

That is going to take up more horizontal space though, up to you.

> Then there is the question of showing fields that exist in options but not in data, what do you think should the user have to manually remove those? Or automatically when they any option change?

At first I was going to remove them automatically, but then I figured that if the user is configuring his query and transforms, and for some reason he decides to change a bit his query and it ends up returning no data, then he would loose all his transofmation configuration automatically, which I wanted to avoid. A possible way would be to remove the absent fields if the input data is not empty, or something like that. That could work too.

We could also improve the current behavior if we keep it, and make sure only the absent fields that were not ignored are kept in the list, with an option to remove them.

Anyway, at this point, I'm letting you finish it up.

mckn:
> > The UI editor needs more work, think we need to handle longer field names better (maybe use table layout to get field names to align). And the FieldCalculationsSelector is untyped. And think it should show fields in the same order as the data (currently it sorts field names).
> 
> @torkelo At this point, I'll let you tweak it up to your taste.
> 
> > Would like to test using a radio button group (https://developers.grafana.com/ui/latest/index.html?path=/story/forms-radiobuttongroup--full-width) for the group by | calculate | ignore selector.
> 
> That is going to take up more horizontal space though, up to you.
> 
> > Then there is the question of showing fields that exist in options but not in data, what do you think should the user have to manually remove those? Or automatically when they any option change?
> 
> At first I was going to remove them automatically, but then I figured that if the user is configuring his query and transforms, and for some reason he decides to change a bit his query and it ends up returning no data, then he would loose all his transofmation configuration automatically, which I wanted to avoid. A possible way would be to remove the absent fields if the input data is not empty, or something like that. That could work too.
> 
> We could also improve the current behavior if we keep it, and make sure only the absent fields that were not ignored are kept in the list, with an option to remove them.
> 
> Anyway, at this point, I'm letting you finish it up.

I will help out making these changes :)

mckn:
Example with radio button group:

Skärmavbild 2020-08-27 kl  07 29 43

torkelo:
The 100% width of the calculation select results in a bit broken layout.
Screenshot from 2020-08-28 15-42-31

Sorry for so many meandering UX feedback thoughts, this has been harder to find a UX for than I thought. Not 100% happy still. 

Have potentially one more UX idea to try.

Mockup 
```
Group by: Region, Data center   | + | 
Select: Max(Sensor1), Last(Sensor1), Max(Sensor2)   | + |
```

So similar to how it was before :) but I feel like we have explored and gone back and forth with this too much already to lets merge this and explore UI change in the future.

mckn:
> The 100% width of the calculation select results in a bit broken layout.
> Screenshot from 2020-08-28 15-42-31
> 
> Sorry for so many meandering UX feedback thoughts, this has been harder to find a UX for than I thought. Not 100% happy still.
> 
> Have potentially one more UX idea to try.
> 
> Mockup
> 
> ```
> Group by: Region, Data center   | + | 
> Select: Max(Sensor1), Last(Sensor1), Max(Sensor2)   | + |
> ```
> 
> So similar to how it was before :) but I feel like we have explored and gone back and forth with this too much already to lets merge this and explore UI change in the future.

Hmm, that sounds like a good plan! Probably something we can try to get some UX feedback on. I will have a look at the long field name issue that you posted above.

PR #25503, opened 2020-06-10
Variables: Fixes maximum call stack bug for empty value

### LINKED ISSUE #25485: Variables: Text input variable does not work because "Maximum call stack size exceeded"
Text variables in dashboards does not work on master.
Screenshot from 2020-06-09 14-54-34
Screenshot from 2020-06-09 14-54-12
Screenshot from 2020-06-09 14-54-09

Seems like here https://github.com/grafana/grafana/blob/00a9af00fc8a4c6d75dedd164da545e80c79a791/public/app/features/variables/textbox/actions.ts#L28 the value from url is always empty string what ever I type in and then it fires this event in cycle.

hugohaggmark:
When using an empty value for a Text variable the logic that checked for variable changes in URL got into an endless loop. This PR tries to remedy this.

Fixes #25485

### DIFF
--- a/public/app/core/services/bridge_srv.test.ts
+++ b/public/app/core/services/bridge_srv.test.ts
@@ -12,4 +12,32 @@ describe('when checking template variables', () => {
     expect(findTemplateVarChanges(b, a)).toEqual({ 'var-xyz': 'hello' });
     expect(findTemplateVarChanges(a, b)).toEqual({ 'var-xyz': '' });
   });
+
+  it('then should ignore equal values', () => {
+    const a: UrlQueryMap = {
+      'var-xyz': 'hello',
+      bbb: 'ignore me',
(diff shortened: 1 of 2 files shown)


hugohaggmark:
Is there some additional work that is needed before approval here @aocenas @tskarhed

tskarhed (approved):
Looks good! Error is no longer there!

aocenas:
@hugohaggmark 
So there was still some issues based on the conversations in slack mainly:
> But you should not come to the state when you see `query: some text` while the actual value is something else. 

This still seems like a bug to me.

aocenas:
But I am ok if that is solved separately.

hugohaggmark:
Would be great if you could file that as a separate issue @aocenas 🙏

PR #25075, opened 2020-05-25
Settings: Expand variables in configuration

sakjur:
**What this PR does / why we need it**:
This PR improves support for environment variables in configuration by breaking the logic out and adding tests.

This PR also adds support for using files in place of environment variables by introducing a new `$__<provider>{<argument>}` format in addition to the old `${<envvar>}`. We have a workaround for this in the Docker launcher already, but this adds the functionality to core Grafana.

This opens up for the option of not having to have `tls_cert_file` and `tls_cert` as two separate fields, it also allows you to mount for example Kubernetes or Vault secrets as a filesystem and access them that way (which is slightly more secure if set up properly than using environment variables as open files are not dumped into audit logs or core dumps unlike environment variables).

In addition to that, my changes are made to be expandable (we're going to use this in Grafana Enterprise) to make sure it's easy to add new providers for variables in configuration.

PR #25508, opened 2020-06-10
Setting: Add docs for option variable expansion #25075

sakjur:
Breaking out the docs from #25075 so we can merge them with the 7.1 beta rather than having to merge them directly and placing a confusing block in the docs.


oddlittlebird (approved):
Looks great!

torkelo:
nice!

PR #25521, opened 2020-06-10
Docs: Updated URL for developing plugins

scoren-gl:
The existing link for developing plugins was redirecting to the legacy plugin docs page for developers. I updated it to the new link for latest post-7.0 release. 👍


marcusolsson (approved):
Nice! Thank you for this!

Issue #18841, opened 2019-09-02
Replace racially charged language "whitelist/blacklist"


naomijub:
**What would you like to be added**:
I'd like to deprecate the "whitelist/blacklist" language in Grafana, replacing them with "safelist/blocklist" or "allowlist/denylist" (This has been done in many places like [Graphite](https://github.com/graphite-project/graphite-web/issues/1569) )

**Why is this needed**:
This is important to help Grafana become more inclusive.

torkelo:
Ah, good feedback! never thought about the origin of these words!

daniellee:
We don't have any blacklists except for tslints rules - as that is an external lib that would be hard to change. Unless I am missing something?

We use whitelist in three places: 
- whitelisted cookies for datasource config 
- `data_source_proxy_whitelist` [configuration setting](https://grafana.com/docs/installation/configuration/#data-source-proxy-whitelist).
- In the [docs for Azure Monitor](https://grafana.com/docs/features/datasources/azuremonitor/#azure-monitor-metrics-whitelist)

naomijub:
@daniellee  The idea is to replace any references to `whitelist` to something like `allowlist` and `blacklist` to something like `denylist`. So `data_source_proxy_whitelist` would become `data_source_proxy_allowlist` and in the Azure Monito:

> Azure Monitor Metrics **Allowlist**
> Not all metrics returned by the Azure Monitor API have values. The Grafana data source has a **allowlist** to only return metric names if it is possible they might have values. This **allowlist** is updated regularly as new services and metrics are added to the Azure cloud. You can find the current **allowlist** here.

I can open an issue in Tslint as well for their `blacklist`

kylebrandt:
Also the auth proxy  https://grafana.com/docs/grafana/latest/auth/auth-proxy/

PR #25533, opened 2020-06-11
Docs: Add guidelines for inclusive language

marcusolsson:

Adds guidelines for how to avoid using charged language.

#18841

### DIFF
--- a/contribute/style-guides/documentation-style-guide.md
+++ b/contribute/style-guides/documentation-style-guide.md
@@ -173,6 +173,12 @@ Two words if used as a verb, one word if used as a noun.
 
 This section provides guidelines on how to avoid using charged language in documentation.
 
+**Exceptions:**
+
+* When referring to a configuration or settings used by technologies outside the Grafana project, prefer the original name to avoid confusion.
+
+  For example, use "master" when referring to the default Git branch.
+
 #### Allowing and blocking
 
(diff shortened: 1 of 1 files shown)


daniellee (approved):
LGTM

The tricky part is 3rd party libraries or technologies.

marcusolsson:
Agreed. I tried explaining it under Exceptions, using the Git master branch as example, but now realize that this should apply to all guidelines under Inclusive language.

contribute/style-guides/documentation-style-guide.md
```diff
@@ -168,3 +168,29 @@ Two words if used as a verb, one word if used as a noun.
 
 * Set up the workspace.
 * Initial setup might take five minutes.
+
+### Inclusive language
```
oddlittlebird, line 172: I think this section should be an h2 and above word usage. 

Also, periods should be inside the quotation marks in this section, not outside.

oddlittlebird (changes requested):
Whoops, forgot to make the earlier comment a review.

marefr:
@oddlittlebird can you fix the changes you've suggested since Marcus is on vacation?

oddlittlebird:
@marefr , will do

oddlittlebird (approved):

PR #25540, opened 2020-06-11
POC: Only run certain build steps for PR's if certain files have changed

marefr:
Proof of concepts for skipping certain build steps for PR's if certain files haven't been changed. Attempt of quick win to reduce build times.

**Notes:**
- Not sure comparing against origin/master will work in all cases (forks etc).
- CircleCI provides parameters for revision (current commit) and base revision (commit before) so using that will only allow to compare changed files since the last commit - maybe that's what we want?  

**Examples:**
- Build of this PR before adding condition to build everything if circleci config changed: https://app.circleci.com/pipelines/github/grafana/grafana/10538/workflows/b8ed1aa8-b0a9-48ba-adae-c25879acd244 (Duration 7m 1s)
- Build with changes in `docs/` folder: https://app.circleci.com/pipelines/github/grafana/grafana/10561/workflows/0dac3312-3ef2-44bf-bd3e-7d2122da8e5b
- Build with changes in backend files: https://app.circleci.com/pipelines/github/grafana/grafana/10563/workflows/3d2b71c7-32e2-4f9b-811f-e6a338fd0557
- Build with changes in frontend files: https://app.circleci.com/pipelines/github/grafana/grafana/10564/workflows/a896f62b-6343-4ece-ba51-b4704b533baf
- Build with changes in backend and frontend files:  https://app.circleci.com/pipelines/github/grafana/grafana/10565/workflows/9ea0dd07-3f6f-44dc-aee4-f623cb90b264

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ jobs:
             fi
       - checkout
       - halt-if-not-matching-files:
-          regex: ^(go.mod|go.sum|.*\.go)
+          regex: ^(go\.mod|go\.sum|.*\.go)+
       - run:
           name: CI job started
           command: ./scripts/ci-job-started.sh
@@ -163,7 +163,7 @@ jobs:
             fi
       - checkout
       - halt-if-not-matching-files:
-          regex: ^(e2e/|emails/|packages/|public/|package.json|yarn.lock)+
+          regex: ^(package\.json|yarn\.lock|e2e/|emails/|packages/|public/)+
       - restore_cache:
           keys:
             # Use own cache for front-end builds, since for some reason it can't be unpacked for parts of the
@@ -330,9 +330,9 @@ jobs:
           at: /tmp/workspace
       - checkout
       - halt-if-not-matching-files:
-          regex: ^(e2e/|emails/|packages/|public/|package.json|yarn.lock)+
+          regex: ^(package\.json|yarn\.lock|e2e/|emails/|packages/|public/)+
       - halt-if-not-matching-files:
-          regex: ^(go.mod|go.sum|.*\.go)+
+          regex: ^(go\.mod|go\.sum|.*\.go)+
       - run:
           name: CI job started
           command: ./scripts/ci-job-started.sh
@@ -373,9 +373,9 @@ jobs:
           at: /tmp/workspace
       - checkout
       - halt-if-not-matching-files:
-          regex: ^(e2e/|emails/|packages/|public/|package.json|yarn.lock)+
+          regex: ^(package\.json|yarn\.lock|e2e/|emails/|packages/|public/)+
       - halt-if-not-matching-files:
-          regex: ^(go.mod|go.sum|.*\.go)+
+          regex: ^(go\.mod|go\.sum|.*\.go)+
       - run:
           name: CI job started
           command: "./scripts/ci-job-started.sh"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -61,6 +61,35 @@ commands:
             chmod +x grabpl
             mv grabpl /tmp
 
+  halt-if-not-matching-files:
+    description: "Requires files of last commit to match provided glob pattern. If no matches are found, the step will halt and be marked as successful."
```
bergquist, line 65: <3

Perhaps we can check all commits since the branch was created?

`git log --oneline master...myBranch``

bergquist (commented):
I think its worth a try. 
If something breaks we should notice when master doesn't build correctly.

tskarhed:
We might want to only run publish-storybook for changes in `packages/grafana-ui`.

.circleci/config.yml
```diff
@@ -557,6 +610,8 @@ jobs:
     executor: grafana-publish
     steps:
       - checkout
+      - halt-if-not-matching-files:
+          regex: ^(e2e/|emails/|packages/|public/|package.json|yarn.lock)+
```
tskarhed, line 614:
```suggestion
          regex: ^(packages/grafana-ui)+
```

hugohaggmark:
I think e2e test should be run even if only backend files have changed, WDYT?

.circleci/config.yml
marefr: The current approach should already check all commits in branch compared to origin/master `git log --oneline --pretty="format:" --name-only origin/master..`

marefr:
Question, if packaging/docker haven't been changed can we skip that step?

tskarhed (approved):
Storybook/Frontend looks good! It will be interesting to see how this impacts our productivity.

hugohaggmark:
Rerunning backend, frontend and backend and frontend builds.

marefr:
@hugohaggmark end-to-end tests can only be run if both backend and frontend have been built. So either we have to always build backend and frontend if there are backend or frontend changes or only run end-to-end test if there are backend and frontend changes. 

Third alternative would be to always build backend if backend or frontend changes and always run end-to-end if frontend changes.

Thought?

xlson (approved):
Seems like a good start at speeding up the build. A bit sad that it doesn't show in the flow visualizer on circleci as that means its pretty hard to understand that certain steps have been skipped.

Great job!

marefr:
@xlson what's your input on this question https://github.com/grafana/grafana/pull/25540#issuecomment-642617580 ?

hugohaggmark:
> @hugohaggmark end-to-end tests can only be run if both backend and frontend have been built. So either we have to always build backend and frontend if there are backend or frontend changes or only run end-to-end test if there are backend and frontend changes.
> 
> Third alternative would be to always build backend if backend or frontend changes and always run end-to-end if frontend changes.
> 
> Thought?

Gotcha, as far as I can remember when e2e has failed it's because of changes in FrontEnd (that might change) so maybe we should run e2e every time there is a FrontEnd change? I guess that means we always need to build back end when front end changes, not sure what the time cost would be for that?

marefr:
@hugohaggmark should be fine. The frontend is much slower building than the backend

hugohaggmark:
@marefr outstanding work on this! 🌟  It's not my intention to hold up this improvement so go ahead with the changes you think make the most sense.

torkelo:
> Gotcha, as far as I can remember when e2e has failed it's because of changes in FrontEnd (that might change) so maybe we should run e2e every time there is a FrontEnd change? I guess that means we always need to build back end when front end changes, not sure what the time cost would be for that? 

Yes, it's very important that e2e tests run on PRs and for that, we need to build release package (tarball) as e2e tests use that

.circleci/config.yml
torkelo: But the description says only last commit, so the description wrong? `Requires files of last commit to match provided glob pattern`

PR #25542, opened 2020-06-11
UI/Switch: Makes Switch ID generation more stable

kaydelaney:
Previously a new ID was being generated for a switch every time it was rendered. This PR ensures each switch's ID is only generated once.


mckn (approved):
Great work! LGTM!

PR #25544, opened 2020-06-11
Explore/Logs: Fix tooltip display for log graph

kaydelaney:
Previously hovering over the Explore graph in Logs mode would display multiple values, but all with the label 'Value' and omitting the corresponding graph color.
image

This PR ensures the series' name and color is displayed along with the value(s).
image


peterholmberg (approved):
Nice!

peterholmberg (approved):

PR #25546, opened 2020-06-11
Tab: Make active tab clickable and add hyperlink functionality

### LINKED ISSUE #25536: Tabs: The main navigation tabs should support being anchor hrefs and support clicking on active tab
1) The change to tabs component during 7.0 made them not handle/support link hrefs (so you cannot ctrl click to open tabs). 
2) The active tab can no longer be clicked (this is a big problem in a few pages where you need to be able to click the active tab to go back) 

the second point might sound strange but think GitHub, when view an issue the issues tab is active, but you can still click it to go back to issue list (same when you create new issue and want to go back to issue list, or have a PR open). 

Problem page in grafana, server admin > user page. No way to go back to user list now. 

Screenshot from 2020-06-11 11-38-03

tskarhed:

Fixes #25536
Adds `href` attribute to Tab component

- Apply onChangeTab for active tab
- Mouse: pointer
- When href attribute exists it renders an anchor element


torkelo (approved):
Awesome!

dprokop (changes requested):
Think we need to adjust the Tabs a bit. In this implementation, the clickable (ctrl-clickable) area is limited to the link only. This means that you need to be "quite" precise to i.e. open the link in new window:

image

IMO the entire tab li should be a link, so the padding belongs to the anchor.

tskarhed:
I agree. Probably did this fix a bit too quickly to remember that. I need to figure out why the e2e tests are failing here too.

packages/grafana-ui/src/components/Tabs/Tab.tsx
```diff
@@ -11,31 +11,44 @@ import { Counter } from './Counter';
 export interface TabProps extends HTMLProps<HTMLLIElement> {
   label: string;
   active?: boolean;
+  /** With this it is possible to use the tab as a hyperlink. Use in cases where the tabs update location. */
```
dprokop, line 14:
```suggestion
  /** With provided, it is possible to use the tab as a hyperlink. Use in cases where the tabs update location. */
```

packages/grafana-ui/src/components/Tabs/Tab.tsx
```diff
@@ -11,31 +11,44 @@ import { Counter } from './Counter';
...
+  href?: string;
   icon?: IconName;
+  /** May cause unintended behaviour when used with href */
```
dprokop, line 17: uhm, not sure this fits into props table comment. Think this should be a part of the component documentation.

packages/grafana-ui/src/components/Tabs/Tab.tsx
```diff
@@ -11,31 +11,44 @@ import { Counter } from './Counter';
...
   onChangeTab: () => void;
+  /** A number next to the text. Usually used to display number of items in a tab's view. */
```
dprokop, line 19:
```suggestion
  /** A number rendered next to the text. Usually used to display the number of items in a tab's view. */
```

packages/grafana-ui/src/components/Tabs/Tab.tsx
```diff
@@ -11,31 +11,44 @@ import { Counter } from './Counter';
...
       <li
         {...otherProps}
-        className={cx(tabsStyles.tabItem, active && tabsStyles.activeStyle)}
-        onClick={() => {
-          if (!active) {
-            onChangeTab();
-          }
+        className={cx(!href && tabsStyles.padding, tabsStyles.tabItem, active && tabsStyles.activeStyle)}
+        onClick={(event: React.MouseEvent<HTMLLIElement>) => {
+          onChangeTab();
```
dprokop, line 40: Think this may be problematic. Imagine user providing a `href` and at the same time `onChangeTab`. He/she does not have a chance to cancel the event in that situation. We have two paths here:
1. If `href` is provided stop the propagation in the `onClick` handler.
2. Change signature of the onChangeTab to be `onChangeTab(event: ...)` so that the responsibility is delegated to the consumer.

Think the second approach is the one we want. For example ant.d follows this pattern: https://ant.design/components/tabs/

dprokop (approved):
Awesome, left some nits to consider:)

PR #23439, opened 2020-04-08
DataSource/Cloudwatch: Add Auth Option and Cache Config Per DataSource

patstrom:
**What this PR does / why we need it**:

This PR does two main things. 

## First thing: Authentication
It refactors the entire tsdb/cloudwatch authentication flow and adds a `AWS SDK Default` authentication option that does what one would expect. All other options are refactored to only perform *exactly* what is specified and thus any implicit behaviour is removed.

The details of this are isolated to the first commit of this PR (8db20e4f31) and it contains some more information.

## Second thing:  No more global caches
The current cloudwatch implementation makes frequent use of global caches, both for credentials (or sessions after 8db20e4f31) and for custom metrics and dimensions. This is obviously a problem when someone for example adds two cloudwatch datasources or when custom metrics are requested for two different regions at the same time. To be honest I haven't really bothered to find out all the ways this can cause a problem but caching everything on a per-datasource basis would solve them, as well as make more sense conceptually.

The changes are in commit c37277eac6 and you can find more information there

**Which issue(s) this PR fixes**:

This PR is the "proper" solution mentioned [here](https://github.com/grafana/grafana/issues/20473#issuecomment-559181587) for #20473.

While https://github.com/grafana/grafana/pull/21594 does solve this specifically for EKS IAM Roles for Service Accounts the new `AWS SDK Default` (which just uses the default chain as specified by the SDK) is more future-proof in the sense that any new feature can be taken advantage of by simply updating the dependency rather than replicating it.

**Special notes for your reviewer**:

*I need help testing this*! 

This MR could have wide implications for people relying on these potential bugs and implicit behavior. I do doubt anyone is actually relying on the bugs but I think *a lot* of user will have to update their Cloudwatch authentication used. This can of course be frustrating but personally I find it even more frustrating that sometimes things are implicit and sometimes they are not. In order to set up Grafana in an EKS cluster myself the first time I didn't want any of the available plugins (I wanted SDK default) so I just chose one and provided bad options until it eventually used some default credential as fallback.

I've tested this locally with all the different options. Credentials file and static credentials will *only* choose their respective providers so no implicit behavior at all there, since it works locally it should work everywhere. I still need to make sure the AWS SDK Default and Assume Role (which uses the sdk default for AssumeRole credentials) works in the following cases :

- [x] EC2 Instance with assigned role
- [x] Docker inside EC2 Instance with assigned role
- [x] ECS Task using the `aws_container_credentials_relative_uri` environment variable
[...]

PR #9718, opened 2017-10-30
fix cloudwatch ec2_instance_attribute

mtanda:
Fixes, https://github.com/grafana/grafana/issues/9667
Sorry, I may confirm that this feature is working, but not...

PR #25555, opened 2020-06-11
cloudwatch: Consolidate client logic

patstrom:
Consolidate logic for getting clients in CloudWatch data source. Also clean up the CloudWatch test suite.

This is part 1 of PRs to replace https://github.com/grafana/grafana/pull/23439


patstrom:
@aknuds1

aknuds1:
I've synced up your branch against latest master in [this branch](https://github.com/grafana/grafana/tree/patstrom-cloudwatch-internal-interface). Do you want to merge it into your own? I can also push to your branch if you want.

patstrom:
Very busy at the moment unfortunately. I'll try to make some room for it this week.

aknuds1:
@patstrom If you want, I can just push my [changes](https://github.com/grafana/grafana/tree/patstrom-cloudwatch-internal-interface) to your branch.

patstrom:
There doesn't seem to be any `diff` between `patstrom:f/cloudwatch-internal-interface` and `grafana:patstrom-cloudwatch-internal-interface` that I can find (or at least that Github shows)? What am I missing?

aknuds1:
Let me double check after lunch.

aknuds1:
@patstrom Checking locally, there are 1459 files that differ between the two.

patstrom:
I wrote to you in slack but you have notifications turned off.

Looking at the changes in `pkg/tsdb/cloudwatch` it seems like the interface to create a new cloudwatchExecutor has changed and that there are now factory functions for certain aws service clients. Is that the gist of your changes? Everything else seems to be in service of that or unrelated.

aknuds1:
@patstrom I'm not sure which branch you're referring to. [patstrom-cloudwatch-internal-interface](https://github.com/grafana/grafana/tree/patstrom-cloudwatch-internal-interface) is basically the same branch as this, minus merge conflicts.

patstrom:
Yeah go ahead and push your fixes.

aknuds1:
Thanks, I've pushed my conflict resolutions to your branch.

aknuds1:
There's a CI error, looking into it.

aknuds1:
We've finally been able to discuss this in full internally, and have come to a conclusion. Would you mind merging the changes in [this branch](https://github.com/grafana/grafana/tree/chore/cloudwatch-interface) into your PR? We think it's a better solution, because in our opinion letting `awsCredsCache` implement the `clientCache` interface violates the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). Basically, the credentials cache gets a second responsibility, which is to fetch clients. Furthermore, our solution simplifies testing and depends less on fakes/mocks.

patstrom:
Sounds reasonable. Sure.

Personally I am not a huge fan of global variables like that though. I would much rather have an interface and a "default" implementation. That was the idea with my original implementation. Perhaps the interface should have been called `awsClients` and a implementation `awsClientsCache`.

The `chore/cloudwatch-interface` branch does actually implement the very same interface, it's just not explicit. I guess that technically also violates the single responsibility principle (as in a `cloudWatchExecutor` is not really concerned with fetching clients _per se_, but rather executing queries).

How would you feel about something like

```golang
type awsClients interface {
  getCWClient(region string) (cloudwatchiface.CloudWatchAPI, error)
  // etc
}

// or whatever you want to call it. The "default" one that just fetches clients
type naiveClientsFetcher struct {}

func (n *naiveClientsFetcher) getCWClient(...) (...) {
  // create and return an actual client
}

type awsClientsCache struct {
  clients awsClients
  cache map[key]value
}

func (c *awsClientsCache) getCWClient(...) (...) {
  if v, ok := c.cache[key]; ok {
    return v
  }
  v := c.clients.getCWClient(...)
  c.cache[key] = v
  return v
}

type cloudwatchExecutor struct {
   client awsClients // either use a "naiveClientsFetcher" or wrap a "naiveClientsFetcher" in a "awsClientsCache"
}
```

While the _interface_ might not adhere to the single responsibility principle the actual _implementations_ will (not that an interface necessarily must adhere to that principle). It would also actually explicitly say what the interface is and delegate the responsibility of fetching & caching clients from the `cloudWatchExecutor` to something else.

In this PR only the non-caching one would be implemented, where the caching one would be implemented in the next one (when fixing the cache).

aknuds1:
Which global variables are you referring to, the factory functions? I don't think you can argue that CloudWatchExecutor violates SRP though, since it doesn't provide that service via its interface. Its interface is in fact only the `Query` method. Fetching clients is something it does for itself, in order to _fulfill_ its responsibility.

Regarding changing the implementation in our branch, I don't wish to iterate any further on this PR. We've already spent enough time on it, and we have plenty other work.

After merging, we can move on to the next PR and evaluate that on its own merits.

patstrom:
Ah yeah thinking about it some more I don't think it it will matter much in the end. I'll try to get the merge up today.

aknuds1:
Can you remind me of why you remove caching of logs, EC2 and RGTA clients in this PR? Is there a certain reasoning behind this change?

aknuds1:
@mtanda Do you know why  CloudWatch logs, EC2 and RGTA clients are cached (while the CloudWatch client is not)? FYI, this PR proposes to drop the caching of clients.

patstrom:
I had to remove the "caching" of EC2 and RGTA clients in order to unify the interface for fetching in the first place. The other alternative would've been to add a `cwSvc cloudwatchiface.CloudWatchAPI` field to the `cloudwatchExecutor` and a corresponding `ensureCloudwatchClient` function that does a nil-check and I felt that was the wrong way forward.

The Cloudwatch logs client cache I removed because it did a cloudwatchlogs-specific caching on a region basis. While that is what we want (eventually) I thought the proper way was to remove it for now in order to be able to unify the interface and then later on start caching the `sessions` on a region-basis instead. That way all the different clients can use the same cache-implementation.

So the idea here moving forward is that once there is a unified interface for fetching AWS clients we can move on to fix the now buggy cache by using the data source id as key and start caching the actual `Session` (on a region-basis) instead of `Credentials`, which will be a performance improvement.

aknuds1:
@patstrom So if we ignore the reasons why you had to remove the caching in the first place, what would break if we re-introduce client caching (according to the original naive scheme) in this PR? I'm leaning towards re-introducing client caching in this PR, so we don't break anything and instead propose an improved(?) caching system in another PR.

patstrom:
The reasons for actually removing the caching in this PR is purely in order to de-clutter the code base. Re-introducing the very same `ec2Svc`, `rgtaSvc` and `logsClientsByRegion` and instead manipulating them in the corresponding "get" function shouldn't break anything when compared to the previous behavior (as far as I can tell).

Do note that all caching in this package is buggy though. Nothing is properly invalidated. E.g if you provide a working configuration, do some queries (and stuff is cached) and then update the configuration to something that shouldn't work, it still will (in some cases).

I guess it is a matter of taste if you want to remove it now or later (either when replacing them with something proper or just removing them to fix the bugs). Personally I felt that given that they are buggy and that there are two different client cache implementations I felt it better to just remove them since they made the code much more difficult to reason about.

aknuds1:
Thanks for clarifying that @patstrom. It sounds as if I might just code up a reintroduction of caching for this PR, to keep the current scheme, unless my colleagues think differently.

mtanda:
> @mtanda Do you know why CloudWatch logs, EC2 and RGTA clients are cached (while the CloudWatch client is not)? FYI, this PR proposes to drop the caching of clients.

I didn't intend to cache client, I need to use mock client for test code.
https://github.com/grafana/grafana/pull/9718/commits/5255c9c70cbe845a6d500737610e8933078c18d4
(Sorry, I didn't read discussion)

aknuds1:
@mtanda Aha, thanks for clarifying that.

aknuds1:
@patstrom I've re-introduced client caching in [this branch](https://github.com/grafana/grafana/tree/patstrom-cloudwatch-internal-interface). I also removed initialization of cloudWatchExecutor.DataSource when it gets instantiated, since I realized it always gets initialized in the `Query` method anyway. Also fixed a merge conflict. Please look my changes over and merge them into your branch, or let me know if you see any problems.

aknuds1 (approved):
LGTM now, but there's a remaining minor merge conflict. I've fixed it in [this branch](https://github.com/grafana/grafana/tree/patstrom-cloudwatch-internal-interface).

aknuds1:

patstrom:
Cheers! :tada:

Do you want me to update the other MRs to be based on this instead?

aknuds1:
@patstrom Just base them (rebase, merge, whatever) on master, since this PR has been merged.

patstrom:
Ah yeah that's what I meant :sweat_smile:

PR #25548, opened 2020-06-11
CloudWatch: Re-implement authentication

aknuds1:
**What this PR does / why we need it**:
PR to extract new CloudWatch authentication scheme from @patstrom's PRs #25558 and #23439.

The original PR touches on too many aspects to be practically reviewable - this PR is an attempt to extract the authentication scheme changes.

# Remote Credentials Method
According to [AWS documentation](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials), the default chain uses the EC2 IAM role automatically.

According to [AWS Documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html), the default chain should pick up ECS credentials automatically.

# Web Identity Method

[These docs](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html) indicate that the SDK supports this provider, but doesn't say if it's part of the default chain. I've read the SDK source code, and looks as if it is picked up by default, but we need to test this (under EKS?).

# Credentials Caching

From reading the [session documentation](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/), I can see that sessions should be cached. So it sounds as if we're doing the wrong thing by caching credentials instead of sessions. Also, the [credentials documentation](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#Credentials) indicates that credentials are cached in the SDK. Last, but not least, in this PR we won't create credentials directly any longer, so only sessions can be cached.

# Backwards Compatibility

I think it should be backwards compatible since all of the current authentication types are supported. The "arn" authentication type from the UI should just be mapped to the default authentication method, and the `assumeRoleArn` property should be applied if present.

# Testing

### Configured Profile
Arve has tested this.

### Default Profile
Arve has tested this.

## Explicit Keys Method
Arve has tested this.

## Default Method
Arve has tested this.

## Assume Role Method (Backwards Compatibility)
Arve has tested this (falls back to default method).

## Role Assumption
Arve has tested this.

# TODO:

- [x] Figure out if we should keep web identity auth method -> Should be handled by SDK default chain, from reading the source, but need to test
- [x] Figure out if we should keep remote credentials auth method -> Should be handled by SDK default chain, see above notes
- [x] Figure out if we should keep credentials caching -> No, see notes above (sessions should be cached instead, but best done in another PR?)
- [x] Figure out backwards compatibility story
[...]

PR #25556, opened 2020-06-11
cloudwatch: Cache executor-related things by id

patstrom:
This commit introduces changes that simplifies the AWS caching by
keeping one executor per data source id that keeps all caches as struct
fields. This is in contrast to the pre-commit behavior where different
things were kept in different global caches with keys that tried to
limit them to a data source.
    
The pre-commit behavior might have worked out but there are definitely
places where the global cache could have returned a false-positive cache
hit (for example one data source using another data sources
credentials).
    
The Custom Metrics caches can probably be simplified since they would
only need to use region and namespace as key but this commit is simply a
refactor that tries to keep all current functionality.

Fixes #

This is part 2 of the PRs to replace https://github.com/grafana/grafana/pull/23439 and is based on https://github.com/grafana/grafana/pull/25555


patstrom:
@aknuds1

patstrom:
Ah sorry the status is that I have been on vacation on currently have a full plate at work. I hope to get to this (and the other PRs) in a week or two.

aknuds1:
Sorry, thought I had removed the stale label on this one. Now that @papagian is back from holiday, I think she and I will try to review your PRs together, since at least the auth scheme PR needs several pairs of eyes.

patstrom:
Great! I'll try to update all the relevant PRs as soon as possible!

aknuds1:
@patstrom There's no need to rush, we need to work through the PRs and develop an idea of what changes should be made and the effect they will have.

patstrom:
Let me know if you have any questions or just want some clarifications about what I had in mind!

aknuds1:
No worries, we will ask when we review and are unsure about anything.

aknuds1:
Can you sync this PR with master? We're interested in implementing session caching, in order to enable [the PR](https://github.com/grafana/grafana/pull/25548) to re-design the CloudWatch auth scheme.

patstrom:
So you've already implemented the sessions cache in that PR I'll wait with this until the new authentication is merged. Otherwise there will just be even more conflicts.

aknuds1:
Yeah, probably best to wait as I produced what I hope is an adequate quick fix in the meantime.

PR #25557, opened 2020-06-11
cloudwatch: Cache sessions instead of credentials

patstrom:
Caching the AWS SDK Credentials struct is not the recommended way to use
the AWS SDK for Go. The pre-commit behavor is that an instance of type
`Credentials` is generated when needed that is then used to create a
`session.Session` that can be used to query the AWS API.

Creating a session causes environment variables to be read and
potentially files to be opened and scanned. This is somewhat expensive
and unnecessary. A `session.Session` will, once created, handle things
like caching this information on its own. In other words, it makes far
more sense to cache a `session.Session` rather than a `Credentials`.

This commit moves from a `Credentials` cache to a `session.Session`
cache. The key thing to note here is that the session contains
information about which region it should query, which is why we must now
cache sessions on a region basis as well as on a datasource basis.

A potential for simplification is to have a more generic
"region-specific-information" cache that could contain e.g both an
appropriate session and the region specific custom metrics. The custom
metric caches are left as-is in this commit as to avoid scope-creep.

Fixes #

This is part 3 of PRs to replace https://github.com/grafana/grafana/pull/23439 and is based on https://github.com/grafana/grafana/pull/25556


patstrom:
@aknuds1

PR #25558, opened 2020-06-11
cloudwatch: Rework CloudWatch authentication

patstrom:
The pre-commit behavor for CloudWatch authentication is a little
confusing and implicit. There is always some implicit fallback to
default-ish type of credentials and it is a little unclear as to how one
should configure their data source.

The main point of this commit is to expose the AWS SDK Default
authentication provider. This does exactly what it says, it uses the
default credentials chain. This will allow Grafana to take advantage of
any future providers by just updating the SDK (rather than
re-implementing them as in the case of the remotecred and the
webidentityprovider).

It will also actually expose a proper, explicit
authentication provider for many different environments. In the
pre-commit behavior the way to use e.g EKS IAM Roles for Service
Accounts is to provide faulty credentials in another provider so it can
fall back to the web identity provider, which is just ridiculous. The
situation is similar for ECS tasks and EC2 instance roles.

In order to resolve the situation of removing this confusing, implicit
behavior this commit, as mentioned, implements the AWS SDK Default
authentication provider. It also removes any implicit behavior from the
Credentials file and Access & Secret key providers.

The ARN provider has been reworked entirely since it never was a
provider in the first place, or at least not conceptually. Specifying a
assumeRoleArn is more akin to saying what Grafana should do with the
credentials, not how to obtain them. The pre-commit behavior here is
that when the ARN authentication type is specified it will use the
default credentials chain to assume the role.

This commit changes this and exposes the assumeRoleArn option for all
authentication providers and uses the given credentials to assume the
specified role, which it will then use to query for metrics

Fixes #

This is part 4 (the final part) of PRs to replace https://github.com/grafana/grafana/pull/23439 and is based on https://github.com/grafana/grafana/pull/25557


marefr:
I'll let Arve review this. My only 2 cents is if this is breaking changes I would like to see what the migration path would be:
- Possible to handle old auth type/config settings in backward/forward compatible way instead of throwing an error?
- Documentation of breaking change the only possible way?

patstrom:
I'll try to think of how to handle the breaking changes when we get to this PR!

aknuds1:
Closing this one in favour of #25548, since it has the same purpose and I already started on it. Please continue the discussion there.

Please bear in mind that #25548 starts out very simple on purpose, so our team can figure out how authentication should be done. The idea is to port relevant changes from this PR (and the main one?) gradually.

patstrom:
I just want to summarize my though from our slack conversation here as well.

I see how your WIP and this PR overlap in purpose, but they differ greatly between upon what foundation they are built. I honestly think the first 3 PRs are all worthwhile changes (that doesn't introduce any breaking changes) by themselves and if you agree I think it would be more appropriate to close your WIP in favor of this PR. I just don't see why it would be worthwhile to re-write the breaking changes on master rather than reviewing the code already written.

Also please bear in mind that the commit in this PR is actually really small as well, mostly changes to the documentation. It's just that currently it is based on the other 3 PRs.

public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
```diff
@@ -121,16 +121,18 @@ export class ConfigEditor extends PureComponent<Props, State> {
...
+              </InlineFormLabel>
               <Select
                 className="width-30"
                 value={authProviderOptions.find(authProvider => authProvider.value === options.jsonData.authType)}
                 options={authProviderOptions}
                 defaultValue={options.jsonData.authType}
                 onChange={option => {
-                  if (options.jsonData.authType === 'arn' && option.value !== 'arn') {
-                    delete this.props.options.jsonData.assumeRoleArn;
-                  }
```
hugohaggmark, line 133: Not sure what the impact is when we remove this if statement. Is this because ARN is no longer part of `authProviderOptions`?. Does the test cases reflect this change?

hugohaggmark (commented):
FrontEnd code looks great, I left a question for you to consider.

aknuds1:
@patstrom As stated in the description of #25548, the intention is to extract work (selectively) from this PR.

patstrom:
I know that @aknuds1, but realize that your *intention* is to do it, this *has already done it*. So why spent days writing that code instead of using the code from the original PR which you *intend to duplicate*?

public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
patstrom: Yes it does! I have removed the conditional part of rendering the `arn` div and updated the snapshots.

The idea is that `arn` isn't actually an authentication provider but rather something you do with your selected authentication provider.

123BLiN:
I'm sorry for asking here, but there is too much issues around about `assumeRoleArn` behaviour and I'm not sure if I should create a new one.
I have a problem trying to assume role cross-account in kops and kiam (latest grafana 7.1.1), based on other issues it seems it is mostly working in EKS, but it is using `assume-role-with-web-identity` instead of `assume-role`, so am I understand it correct that second one is still broken and this PR is intended to fix it and use the default behaviour from go-aws-sdk that should work as expected for both?

PR #25566, opened 2020-06-12
Dashboard: fix promql query editor resolution field's position to prevent overflow

MisterSquishy:
currently the resolution select overflows its container and makes certain options inaccessible without the keyboard:
2020-06-12 at 09 28

this moves the menu below the select so that the panel can be scrolled to accommodate:
2020-06-16 at 08 33

suggestion for the [developer guide](https://github.com/grafana/grafana/blob/master/contribute/developer-guide.md) -- in addition to the `ulimit` thing i also had to set a `GOPATH` env var to run the backend. this manifested as a bunch of `go: writing go.mod cache: mkdir /pkg: read-only file system` errors when running `make run`, which was relatively easy to troubleshoot but would potentially still be nice to call out in that doc

### DIFF
--- a/public/app/plugins/datasource/prometheus/components/PromQueryEditor.tsx
+++ b/public/app/plugins/datasource/prometheus/components/PromQueryEditor.tsx
@@ -155,7 +155,7 @@ export class PromQueryEditor extends PureComponent<Props, State> {
             <div className="gf-form-label">Resolution</div>
             <Select
               isSearchable={false}
-              maxMenuHeight={150}
+              menuPlacement="bottom"
               options={INTERVAL_FACTOR_OPTIONS}
               onChange={this.onIntervalFactorChange}
               value={intervalFactorOption}
--- a/public/app/plugins/datasource/prometheus/components/PromQueryEditor.tsx
+++ b/public/app/plugins/datasource/prometheus/components/PromQueryEditor.tsx
@@ -155,6 +155,7 @@ export class PromQueryEditor extends PureComponent<Props, State> {
             <div className="gf-form-label">Resolution</div>
             <Select
               isSearchable={false}
+              maxMenuHeight={150}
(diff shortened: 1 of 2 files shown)


aocenas:
@dprokop how is this solved in general with the grafana ui? Personally I would think any dropdowns and selects should be able to overflow the container or be positioned so they do not need to be resized otherwise there will still be edge cases. For example in this case any change to the query editor or the query tab will require change in the max height here which is not ideal.

torkelo:
you can force the menu to always go down

aocenas:
@torkelo I assume that can create the same issue just on the bottom edge? I would assume the dropdown should have different z-index or created by portal but maybe there are different edge cases to that. So just wondering if we have some existing solutions or if we just do a short term fix here.

torkelo:
if it goes down it will just make the outer container scrollable

dprokop:
> @dprokop how is this solved in general with the grafana ui? Personally I would think any dropdowns and selects should be able to overflow the container or be positioned so they do not need to be resized otherwise there will still be edge cases. For example in this case any change to the query editor or the query tab will require change in the max height here which is not ideal.

@aocenas unfortunately we don't have a general solution for this. Currently, in such cases, we suggest setting the dropdown to open downwards. The Select component needs a lot of love, as it's extremely rigid and hard to maintain current. We have tried implementing portaling for react-select but whatever we did, following the documentation or not, did not result invconsistent behaviour.

aocenas:
@MisterSquishy can you check changing the menuPlacement option, whether it would make sense to to do that here?

MisterSquishy:
sounds good! `menuPlacement` is probably a more resilient fix here; updated. I've also had a lot of trouble making `react-select` behave in a component library 😅

zoltanbedi (approved):
lgtm thanks for contributing @MisterSquishy

PR #25581, opened 2020-06-14
GraphPanel: Make legend values clickable series toggles

### LINKED ISSUE #25402: Clicking on legend for "min/avg/max/total" does not select the metric
**What happened**:

I have a dashboard which has a legend and showing avg values next to the series name. So my label looks like this: `web_server_01  Current: 27.75`. When I click on `web_server_01` I can see only the the specific series. However, if I click on the "Current" or "27.75" nothing happens.

Since those numbers are part of the series I would expect to be able to select it wherever I click

1. Go to a dashboard (.e.g https://play.grafana.org/d/000000012/grafana-play-home?orgId=1&editPanel=2) 
2. Enable "Current" in legend
3. Click on "Current"

image

**Environment**:

hshoff:

- fixes clicking on legend for "min/avg/max/total" does not select the metric (#25402)
- makes legend values clickable series toggles by reusing `onLabelClick()` handler by passing it to the LegendValue. applies to both table and non-table legends.

before
before_legend-label-value

after
after_legend-label-value

Fixes #25402 

This patch does have an effect on text selection for legend values (double-click, drag-to-select).  These compound events start as clicks so the series toggles during the operation. This matches the behavior of the legend labels on click, so no change to this behavior was made. 
after_text-selection

### DIFF
--- a/public/app/plugins/panel/graph/Legend/LegendSeriesItem.tsx
+++ b/public/app/plugins/panel/graph/Legend/LegendSeriesItem.tsx
@@ -71,7 +71,13 @@ export class LegendItem extends PureComponent<LegendItemProps, LegendItemState> 
       if (this.props[valueName]) {
         const valueFormatted = series.formatValue(series.stats[valueName]);
         legendValueItems.push(
-          <LegendValue key={valueName} valueName={valueName} value={valueFormatted} asTable={asTable} />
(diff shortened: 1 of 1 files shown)


torkelo (approved):
Looks good! thanks for contributing!

PR #25587, opened 2020-06-15
Chore: Include golangci-lint with grafana/build-container

aknuds1:
Include golangci-lint with grafana/build-container so it doesn't have to be installed by CI. I noticed this matters since builds would break due to failing to download golangci-lint. It should also make builds faster.


bergquist (approved):
Looks good :)

PR #25590, opened 2020-06-15
Chore: Upgrade CI build Dockerfile to latest Debian Stretch

aknuds1:
Upgrade the Dockerfile for grafana/build-container to latest Debian Stretch and also upgrade its Node and Go versions.

I would experience some node/yarn segfaults, so I make sure to use official Debian packages now instead of downloading generic binaries. Don't know if it actually helps, but I haven't seen the segfaults again at least.

### DIFF
--- a/scripts/build/ci-build/Dockerfile
+++ b/scripts/build/ci-build/Dockerfile
@@ -1,5 +1,5 @@
 # Use old Debian (this has support into 2022) in order to ensure binary compatibility with older glibc's.
-FROM debian:stretch-20200514 AS toolchain
+FROM debian:stretch-20200607 AS toolchain
 
 ENV OSX_SDK_URL=https://s3.dockerproject.org/darwin/v2 \
     OSX_SDK=MacOSX10.10.sdk \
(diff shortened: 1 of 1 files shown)


jessabe (approved):

scripts/build/ci-build/Dockerfile
```diff
@@ -107,18 +108,16 @@ RUN apt-get update && \
...
-    curl -fsS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -   && \
-    echo "deb [arch=amd64] https://dl.yarnpkg.com/debian/ stable main"     \
-      | tee /etc/apt/sources.list.d/yarn.list                           && \
-    apt-get update && apt-get install -yq --no-install-recommends yarn      && \
+    ln -s /usr/bin/llvm-dsymutil-6.0 /usr/bin/dsymutil && \
+    curl -fsL https://deb.nodesource.com/setup_12.x | bash - && \
+    apt-get update && apt-get install -yq nodejs=${NODEVERSION} && \
+    curl -fsS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    apt-get update && apt-get install -yq yarn=${YARNVERSION} && \
```
papagian, line 116: I guess `--no-install-recommends` is removed intentionally?

aknuds1: @papagian I opted to copy the [official](https://classic.yarnpkg.com/en/docs/install/#debian-stable) installation instruction: `sudo apt update && sudo apt install yarn`. I don't know how they design the package with regards to what "recommendations" get installed.

They do mention in the same document though that you can use `--no-install-recommends` if you don't want to install Node (and use e.g. nvm instead). Based on that, it sounds as if we should use their recommendations, in order to ensure that system dependencies are in place.

papagian (approved):
LGTM

bergquist (approved):
LGTM!

Builds as expected.

PR #25602, opened 2020-06-15
GraphPanel: Add support for local time formats in graph panel

### LINKED ISSUE #1459: Localize time_format in graphs
Hello dear developers team.
Thank you for a awesome product, but I have one problem.

Now there is a hadcoded US date format in [time_format function](https://github.com/grafana/grafana/blob/master/src/app/panels/graph/graph.js#L373).
The most annoying case is display month and day. When I see something like "2/3" I'm a bit confused. Is it "the second of Mart" or "the third of February"?
The most sad thing, that I can't to configure this behaviour. 

Unfortunately the simplest way (and may be the most proper) doesn't help here. I mean [toLocaleString](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) with additional options. You can return different array of options instead of hardcoded format pattern and this method convert date in accordance with a right locale.
But in our case there is a jquery plot and it requires date format for converting timestamp by itself.

So, the second way is make some kind of mapping locale -> format array. [Example](http://stackoverflow.com/questions/2388115/get-locale-short-date-format-using-javascript/9893752#9893752). It seems a bit ugly. But it could be a single working solution.

May be I missed some obvious and better solutions. That's why I didn't create a pull request. =)

alex-shpak:

Hi! This PR supposed to fix https://github.com/grafana/grafana/issues/1459
Using browser internalisation support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

If `DateTimeFormat` is available `localTimeFormat` function will try to figure correct `momentjs` format.

Examples:
`en-GB`
image

`en-US`
image

`ru-RU`
image

`ko-KR`
image

Fixes #1459

This change is made to be as non-intrusive as possible. Alternatively formatting with `momentjs` might be replaced by browser API via `graphTickFormatter`.

Browser support seems to be okay, https://caniuse.com/#search=Intl since grafana 7.0 dropped support for IE11 https://grafana.com/docs/grafana/latest/installation/requirements/#supported-web-browsers


magjac:
This is excellent. Do you have an example using [ISO format](https://en.wikipedia.org/wiki/ISO_8601)? Or `sv_SE` if there is no such format.

alex-shpak:
@magjac Hi!
For `se-SV` it produces
image

It is formatting according browser locale, so it is not possible to set ISO8601 specifically.

hugohaggmark:
@alex-shpak thank you so much for this contribution, but there are a few things that I'd like to mention
- There should be some way to opt in or opt out, i.e. I think this should be an overall setting for the org/user profile and not a setting for a specific plugin.
- We're working/discussing on how the next generation of the graph panel should look like and work when we migrate it to React, so adding feature to the existing Graph is something we currently avoid
- You should provide some tests

torkelo:
Looks is looking promising! 

What is adding the comma in `10/6, 10.00? Any change to remove that?

packages/grafana-ui/src/components/Graph/utils.ts
```diff
@@ -119,19 +119,45 @@ export const graphTimeFormat = (ticks: number | null, min: number | null, max: n
...
+    month: 'MM',
+    day: 'DD',
+    hour: 'HH',
+    minute: 'mm',
+    second: 'ss',
+    weekday: 'ddd',
+    era: 'N',
+    dayPeriod: 'A',
+    timeZoneName: 'Z',
+  };
```
torkelo, line 160: This is  called a lot of times, wonder if there is a possibility to cache/memoize this

packages/grafana-ui/src/components/Graph/utils.ts
```diff
@@ -119,19 +119,45 @@ export const graphTimeFormat = (ticks: number | null, min: number | null, max: n
...
     }
-    return 'YYYY-MM';
+    return localTimeFormat({ year: 'numeric', month: '2-digit', hour12: false }, 'YYYY-MM');
   }
 
-  return 'HH:mm';
+  return localTimeFormat({ hour: '2-digit', minute: '2-digit', hour12: false }, 'HH:mm');
+};
+
+const localTimeFormat = (options: Intl.DateTimeFormatOptions, fallback: string): string => {
```
dprokop, line 142: I think this should belong to https://github.com/grafana/grafana/blob/1a0c1a39e4974220890f6a8d003e545847c15863/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts

Also, as @hugohaggmark mentioned, some tests for this are more than welcome :)

dprokop (changes requested):

packages/grafana-ui/src/components/Graph/utils.ts
alex-shpak: Yes, In fact it can be pre-generated once. I will make that change.

alex-shpak:
@torkelo 
> What is adding the comma in `10/6, 10.00? Any change to remove that?

Hi!
So `const parts = new Intl.DateTimeFormat('default', options).formatToParts(new Date());`
returns parts of formatted date in browser's locale, for example:

```
const options = { month: '2-digit', day: '2-digit', hour: '2-digit', minute: '2-digit', hour12: false }
const parts = new Intl.DateTimeFormat('se-SV', options).formatToParts(new Date());
```

will produce
```
0: {type: "day", value: "16"}
1: {type: "literal", value: "/"}
2: {type: "month", value: "06"}
3: {type: "literal", value: ", "}
4: {type: "hour", value: "11"}
5: {type: "literal", value: ":"}
6: {type: "minute", value: "34"}
```

non-literal types are mapped then to `momentjs` formatting options like `HH`, `DD`, etc, and `literals` are just left as it is. So here is where comma comes from: `3: {type: "literal", value: ", "}`

For dropping comma and using specific format some other solution would be better.

alex-shpak:
I pushed some changes:

1. Pre-generate formats to run only once,
2. Added tests, including tests for `graphTimeFormat`, Note, it runs in node environment, therefore behaviour is not exactly the same as browser.

packages/grafana-ui/src/components/Graph/utils.ts
alex-shpak: Hi!
I can move if you think that is better.
From my point of view `localTimeFormat` is different from other functions in `dateTimeFormatters.ts`

`localTimeFormat` returns date format, rather than formatted date. But I can move if you say so.

alex-shpak:
@hugohaggmark 
Hi! Tests has been added.
I think in case of adding option in settings, it is better to have dropdown with some common date formats to select from, instead of relying on browser internalisation support 🤔

hugohaggmark:
@alex-shpak that's outstanding! ⭐   I think you should consider the comments from both Torkel and Dominik too :)

alex-shpak:
@hugohaggmark Thanks, sure! I think I answered all comments.

torkelo:
This is looking really good now but think we are missing 2 parts

* Making all date formatting locale-aware (https://github.com/grafana/grafana/blob/80e9546cec826eb93bdb0e9c2d3cc1fbd3483b5c/packages/grafana-data/src/datetime/formatter.ts#L94)

alex-shpak:
More changes has been pushed:

1. @dprokop `localTimeFormat` with tests is moved to [formats.ts](https://github.com/grafana/grafana/pull/25602/files#diff-7e9d58bdc2d5d8db3f6ad630e20173d8). Putting it to `dateTimeFormatters.ts` creates circular dependency.
2. @torkelo I prepared change of default format in here https://github.com/grafana/grafana/pull/25602/files#diff-7e9d58bdc2d5d8db3f6ad630e20173d8R35-R36, commented ATM, I don't feel confident to change it because I'm not familiar enough with the codebase. Your call.
3. Added `Browser Local` format to Units:
   image

alex-shpak:
Hi there! Any update? :)

alex-shpak:
Well, bot is threatening me :D
Ping for review @torkelo @hugohaggmark @dprokop

tiagomnz:
+1

torkelo:
Ignore comment (just so I can find this later) : https://github.com/grafana/grafana/pull/13429#issuecomment-430272485

torkelo (commented):
Been unable to get this locale thing to work, always get US formatted dates no matter what locale I switch to (installed Locale switcher extension). 

```
new Date().toLocaleDateString();
"8/25/2020"

new Date().toLocaleDateString("sv-SE");
"2020-08-25"

navigator.languages
(3) ["sv", "en-US", "en"]
`

Not sure why the DateTimeFormat & toLocaleDateString do not use the current locale

**Comment by alex-shpak** [comment]
Hi!
> new Date().toLocaleDateString();

I think it is a not straigh-forward how browser picks `default` locale.
Solution can be to change to `new Intl.DateTimeFormat([...navigator.languages], options)`
That works for me in chrome 84 and Firefox 80 by changing language in settings. Also OS settings might affect it:
image

I can push that change to the branch.

File: packages/grafana-data/src/datetime/formats.test.ts [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+        month: '2-digit',
+        day: '2-digit',
+
+        hour: '2-digit',
+        minute: '2-digit',
+        second: '2-digit',
+      },
+      ''
+    );
+    expect(format).toBe('MM/DD/YYYY, HH:mm:ss A');
```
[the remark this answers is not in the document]
alex-shpak: Hi!

`YYYY` comes from this mapping: https://github.com/alex-shpak/grafana/blob/feature/graph-local-dates/packages/grafana-data/src/datetime/formats.ts#L20
It transform any year field to momentjs format `YYYY`

Something to keep in mind: Tests are running in Node environment, not a real browser. So it is using node's Intl api for this. So that comma comes from `Node` formatting.

In node 12
```
new Intl.DateTimeFormat('default', options).formatToParts(new Date());
[
  { type: 'month', value: '08' },
  { type: 'literal', value: '/' },
  { type: 'day', value: '25' },
  { type: 'literal', value: '/' },
  { type: 'year', value: '20' },
  { type: 'literal', value: ', ' },
  { type: 'hour', value: '05' },
  { type: 'literal', value: ':' },
  { type: 'minute', value: '00' },
  { type: 'literal', value: ':' },
  { type: 'second', value: '16' },
  { type: 'literal', value: ' ' },
  { type: 'dayPeriod', value: 'PM' }
]
```

see `{ type: 'literal', value: ', ' },` after `{ type: 'year', value: '20' },`

alex-shpak:
Related: https://stackoverflow.com/questions/57666095/browser-default-locale-intl-datetimeformat-vs-navigator-language

torkelo:
That is great, started work based on this PR to add support for making all this configurable (as well as supporting local browser format)

torkelo:
https://github.com/grafana/grafana/pull/27216

torkelo:
Closing this in favor of https://github.com/grafana/grafana/pull/27216 

It adds formating options as well as experimental browser locale format (based on Intl.DateTimeFormat) I do not think it's 100% usable yet for example it formats US time formats using 24 hour clock but also adding PM/AM , need to figure out a solution for that

alex-shpak:
Cool, 
@torkelo Do you need any of my assist or you will take it from here?

Regarding 24/12h style, likely issue is [here](https://github.com/grafana/grafana/pull/27216/files#diff-7e9d58bdc2d5d8db3f6ad630e20173d8R105), hours always translated to `HH`, but for 12h time it should be `hh` according to `momentjs` docs https://flaviocopes.com/momentjs/

So adding little logic to translating should help. Or always enforce 24h format with `hour12: false` options.

torkelo:
> Regarding 24/12h style, likely issue is here, hours always translated to HH, but for 12h time 

Yes but how to know which to map to based on current locale?

alex-shpak:
🤔  Probably add a bit of logic to mapping.
https://github.com/grafana/grafana/pull/27216/files#diff-7e9d58bdc2d5d8db3f6ad630e20173d8R86-R115

Pseudo code: if `parts` contains `{ type: 'dayPeriod' }` then `hh` else `HH`

I can't push to that branch, but I can prepare some snippet.

PR #25604, opened 2020-06-15
CloudWatch Logs: Adds template variable support to log groups

### LINKED ISSUE #25099: CloudWatch Logs: Variables in log group name don't get interpolated
**What happened**:

Using variables in log group names seems to be not possible like:

`/aws/ecs/containerinsights/$cluster/performance`

Variables values are  rendered.

Cloudwatch Log Insights Query Syntax with Log Group Name like: `/aws/ecs/containerinsights/$cluster/performance`

**Environment**:
- Grafana version: 7.0
- Data source type & version: Cloudwatch
- User OS & Browser: Mac / Safari
- Grafana plugins:
- Others:

### LINKED ISSUE #25185: CloudWatch Logs: Log Group not saved in panel settings when saving dashboard
**What happened**:
- Editing a log type panel with a CloudWatch Logs data source
- Set a log group
- Save the dashboard
- Close the dashboard
- Come back to the dashboard, edit the panel
- The "log group" field in the panel settings is empty and has to be filled out again
- This does not seem to affect fetching results, only the edit mode, so the log group is saved somewhere, just not displayed

Display the log group that's been saved with the panel when editing the panel

See above

-
**Environment**:
Grafana version: Grafana v7.0.1 (ef5b586)
Data source type & version: CloudWatch
OS Grafana is installed on: CentOS 7
User OS & Browser: Mac OS 10.15.4, Chrome 83.0.4103.61
Grafana plugins: NA
Others: NA

kaydelaney:
Lets a user enter custom values into the log groups selector, which are then interpolated with any template variable values.

Closes #25099 
Closes https://github.com/grafana/grafana/issues/25185

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/components/LogsQueryEditor.tsx
+++ b/public/app/plugins/datasource/cloudwatch/components/LogsQueryEditor.tsx
@@ -12,7 +12,7 @@ import { CloudWatchLanguageProvider } from '../language_provider';
 import CloudWatchLink from './CloudWatchLink';
 import { css } from 'emotion';
 
-type Props = QueryEditorProps<CloudWatchDatasource, CloudWatchQuery>;
+type Props = QueryEditorProps<CloudWatchDatasource, CloudWatchQuery> & { allowCustomValue?: boolean };
 
 const labelClass = css`
   margin-left: 3px;
@@ -20,7 +20,7 @@ const labelClass = css`
 `;
 
 export const CloudWatchLogsQueryEditor = memo(function CloudWatchLogsQueryEditor(props: Props) {
-  const { query, data, datasource, onRunQuery, onChange, exploreId, exploreMode } = props;
+  const { query, data, datasource, onRunQuery, onChange, exploreId, exploreMode, allowCustomValue } = props;
 
   let absolute: AbsoluteTimeRange;
   if (data?.request?.range?.from) {
@@ -55,6 +55,7 @@ export const CloudWatchLogsQueryEditor = memo(function CloudWatchLogsQueryEditor
       absoluteRange={absolute}
       syntaxLoaded={isSyntaxReady}
       syntax={syntax}
+      allowCustomValue={allowCustomValue}
       ExtraFieldElement={
         <FormLabel className={`gf-form-label--btn ${labelClass}`} width="auto" tooltip="Link to Graph in AWS">
           <CloudWatchLink query={query as CloudWatchLogsQuery} panelData={data} datasource={datasource} />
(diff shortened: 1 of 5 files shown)


public/app/plugins/datasource/cloudwatch/components/LogsQueryEditor.tsx
```diff
@@ -12,15 +12,15 @@ import { CloudWatchLanguageProvider } from '../language_provider';
 import CloudWatchLink from './CloudWatchLink';
 import { css } from 'emotion';
 
-type Props = QueryEditorProps<CloudWatchDatasource, CloudWatchQuery>;
+type Props = QueryEditorProps<CloudWatchDatasource, CloudWatchQuery> & { allowCustomValue?: boolean };
```
aocenas, line 15: Nitpick: we probably do not need this as a prop if we never set it to false.

public/app/plugins/datasource/cloudwatch/components/LogsQueryField.tsx
```diff
@@ -130,6 +131,12 @@ export class CloudWatchLogsQueryField extends React.PureComponent<CloudWatchLogs
       return Promise.resolve();
     }
 
+    // No need to fetch matching log groups if the search term isn't valid
+    // This is also useful for preventing searches when a user is typing out a log group with template vars
+    if (!/^[\.\-_/#A-Za-z0-9]+$/.test(searchTerm)) {
```
aocenas, line 136: Can we add some link to cloudwatch docs where the allowed chars are mentioned for future reference?

aocenas (commented):
Seems like there is some weird behavior.
1. insert custom group
2. it does not run a query and just return LogGroup is required error. 
3. Add another group from the dropdown
4. Now it sends the query and returns error that the custom log group does not exists which is correct.
5. Remove existing log group, keeping only the custom one
6. This now works ok, runs the query and returns log group does not exist

Probably connected if I save panel with just custom log group it is not preserved when I reload. If I have custom + existing log group the values are persisted.

Also can we add some tests for the new functionality?

marefr (commented):
Don't really have anything to say about the code. Let me know if you want me to test it eventually. 

Reading [here](https://grafana.com/docs/grafana/latest/features/datasources/cloudwatch/#query-variable), but nothing special about log groups. Wondering if it's already possible to define a variable query that populates with log groups or not? If not, maybe something to consider?

public/app/plugins/datasource/cloudwatch/components/LogsQueryEditor.tsx
kaydelaney: Should probably make it more explicit but it's false by omission in Explore, since template variables aren't supported there

public/app/plugins/datasource/cloudwatch/components/LogsQueryField.tsx
kaydelaney: Sure

kaydelaney:
@aocenas Re: weird behavior, can you give slightly more explicit instructions to reproduce? Having difficulty getting the behavior you're describing.

aocenas:
Cool was trying to repro and seem to be working fine now.

Can we also add tests for this if possible?

kaydelaney:
@aocenas Yes, writing some tests now.

aocenas (approved):
Nice work

PR #25606, opened 2020-06-15
GraphPanel: Fix annotations overflowing panels

### LINKED ISSUE #24898: annotation region overlay and hover card overflow graph bounds
annotation-bug

**What happened**:

1. annotation region overlay overflows graph + panel
2. annotation hover card will also overflow

1. annotation region overflow is clipped to graph bounds
2. annotation hover card is fit to graph bounds

1. add graph panel
1. add annotation
2. brush to zoom on leading edge of annotation event overlay

- only able to reproduce with leading edge of annotation region overlay

**Environment**:
- Grafana version: 7.0
- Data source type & version: random walk
- OS Grafana is installed on: macOS catalina
- User OS & Browser: macOS catalina, Chrome 83 & Firefox 76
- Grafana plugins: -
- Others: -

hshoff:

- fixes annotation region overlay and hover card overflow graph bounds (#24898)
- adds region marker whenever annotation overlaps visible time range

Before [flot-events](https://github.com/mojoaxel/flot-events) was always rendering region markers and lines without respecting the visible range. It also would only render region markers if the annotation start was within the visible range, this would hide the marker/edit popover whenever the leading edge of the region was hidden. This PR fixes this by:
- clamping annotation overlays to graph bounds
- rendering region markers for any part of the the annotation overlay in the visible range (within, overlap, leading, trailing)
- skip rendering region lines if they are not within visible range

before
before

after
after

Fixes #24898

This further forks the `flot-events` vendor flot plugin. Since it appears it was already forked for Grafana (TypeScript, createAnnotationTooltip, createEditPopover), I assumed this was ok.

### DIFF
--- a/public/app/plugins/panel/graph/jquery.flot.events.ts
+++ b/public/app/plugins/panel/graph/jquery.flot.events.ts
@@ -287,8 +287,11 @@ export class EventMarkers {
     // var o = this._plot.getPlotOffset();
 
     $.each(this._events, (index, event) => {
+      const options = event.getOptions();
+      const insidePlot = this._insidePlot(options.min) || this._insidePlot(options.timeEnd);
+      const overlapPlot = this._overlapPlot(options.min, options.timeEnd);
       // check event is inside the graph range
-      if (this._insidePlot(event.getOptions().min) && !event.isHidden()) {
+      if ((insidePlot || overlapPlot) && !event.isHidden()) {
         event.visual().draw();
       } else {
         event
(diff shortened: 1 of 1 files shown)


tskarhed:
I'm having some issue creating a new annotation here. Can you confirm this is also happening for you?
Peek 2020-06-16 08-47

hshoff:
Good catch. Fixed. 

```diff
- const [[xmin, xmax]] = [o.left, o.left + this._plot.width()];
+ const [xmin, xmax] = [o.left, o.left + this._plot.width()];
```

tskarhed (approved):
It looks good to me! Thanks for the contribution! :smiley: :partying_face:

PR #25609, opened 2020-06-15
Build: fixed strict null errors to get master green.

mckn:
Reduce number of strict null errors to make master green again.


aknuds1 (approved):
Probably good, but CI fails.

aknuds1:
Also, which milestone should it be? 7.1 or should we get the fixes into 7.0.4?

torkelo (approved):

aknuds1:
Giving this milestone 7.1, please change it in case it's wrong.

PR #25610, opened 2020-06-15
@grafana/e2e: improvements

stevenvachon:


stevenvachon:
There's a problem with master branch, so tests are failing here too. Will have to wait to merge.

packages/grafana-e2e/src/flows/addPanel.ts
```diff
@@ -89,6 +81,14 @@ export const addPanel = (config?: Partial<AddPanelConfig>): any =>
     return e2e().wrap({ config: fullConfig });
   });
 
+// @todo this actually returns type `Cypress.Chainable`
+const closeOptions = (): any =>
```
scottlepp, line 85: can these be async functions?

scottlepp (approved):
lgtm - curious if we can use async/await instead of .then

packages/grafana-e2e/src/flows/addPanel.ts
stevenvachon: Nope, and it's something that I dislike about Cypress. Its API is Promise-like, but not true promises. The asynchronous behaviour is abstracted away from us into a queue. It produces some ugly code that's difficult to debug.

hugohaggmark: I agree with @stevenvachon this is a source of headaches, confusion and has made me rethink our choice of Cypress several times. We used puppeteer before Cypress and the code you wrote for puppeteer was more similar to the code you write on a daily base. Cypress has some pleasant features that puppeteer hasn't like the `dev` mode with time travel.

If this continues to confuse us and slow us down, maybe we should make this super clear in our e2e docs?

stevenvachon: I'll be putting together some documentation and will surely link to the Cypress page explaining their confusing API.

hugohaggmark: Thank you @stevenvachon

PR #25613, opened 2020-06-15
DataLinks: Add internal links in table and allow custom query

aocenas:
- Adds support for internal links in table.
- Allow custom serialized query for internal link.

There are few issues with this but not sure how to fix short term.
### 1. Internal links are a bit simplistic
For Loki the config has single string input for the query but this is not enough to support any datasource. The main goal was to support links to Jaeger and Zipkin which it does but was a bit hardcoded.
In general the query is target data source specific and so the definition of the query should be in target specific way, something like providing a full query editor in the derived field config or something like that

### 2. The current API on of `getLinks` on dataFrame is not ideal imho.
There are 2 concepts
1. `DataLink` which defines the link from data point of view.
2. `LinkModel` which is the actual link data with interpolated values.

With internal links the distinction is even bigger. `DataLink` defines just the query and target data source. `LinkModel` says what the href is going to be and what happens onClick (which isn't the same thing, in Explore we open split view on click even though we have regular link with href).

So the `DataLinks` are driven by the data but not `LinkModel`. For correct `LinkModel` we need to know where we are as the actions could be application context specific (ie open split in Explore, maybe something else in dashboards).

### Next steps
In 7.2 we will probably need to refactor the way we handle links so it is more straight forward and the responsibilities are more clear in the code. In rough terms the process should be:
1. datasource supply dataframe with dataLinks
2. Explore/dashboard applies overrides to dataframe
3. Explore/dashboard creates and passes on getLinks context
4. visualisation passes dataFrame to getLinks to get final linkModel

As there isn't time for that in this release this will be merged as is and then worked on in 7.2


public/app/features/explore/TableContainer.tsx
```diff
@@ -38,12 +41,23 @@ export class TableContainer extends PureComponent<TableContainerProps> {
...
   render() {
-    const { loading, onCellFilterAdded, showingTable, tableResult, width } = this.props;
+    const { loading, onCellFilterAdded, showingTable, tableResult, width, splitOpen, range } = this.props;
 
     const height = this.getTableHeight();
     const tableWidth = width - config.theme.panelPadding * 2 - PANEL_BORDER;
     const hasTableResult = tableResult?.length;
 
+    if (hasTableResult) {
+      // Bit of code smell here. We need to add links here to the frame modifying the frame on every render.
```
aocenas, line 51: Result of the problem 2. Need access to redux and navigation for the links.

public/app/features/explore/state/actions.ts
```diff
@@ -706,24 +706,31 @@ export function splitOpen(options?: { datasourceUid: string; query: string }): T
...
-      //  query shape.
+      rightState.queries = [];
+      rightState.graphResult = undefined;
+      rightState.logsResult = undefined;
+      rightState.tableResult = undefined;
+      rightState.queryKeys = [];
+      urlState.queries = [];
+      rightState.urlState = urlState;
+
+      dispatch(splitOpenAction({ itemState: rightState }));
```
aocenas, line 718: This worked before but triggered few cycles of setting and resetting queries resulting in blinking of different query and sometimes race conditions.

public/app/features/explore/utils/links.ts
```diff
@@ -20,6 +20,10 @@ export function getFieldLinksForExplore(
   const data = getLinksFromLogsField(field, rowIndex);
   return data.map(d => {
     if (d.link.meta?.datasourceUid) {
+      // TODO: fix the ambiguity here
+      // This looks weird but in case meta.datasourceUid is set we save the query in url which will get
+      // interpolated into href
+      const parsedQuery = JSON.parse(d.linkModel.href);
```
aocenas, line 26: Problem 1. Probably could do this in a type safe way (at least a bit more typesafe) though not sure where such changes would ripple.

public/app/features/explore/utils/ResultProcessor.ts
```diff
@@ -69,6 +69,8 @@ export class ResultProcessor {
       return null;
     }
 
+    // We are going from dataframe to table model and back because of the merging which is done on tableModel at the
+    // moment.
```
torkelo, line 73: We actually merged a DataFrame merge transform yesterday! so now we can look at getting rid of this costly DataFrame -> Table -> DataFrame mess :)

torkelo (commented):
Sorry for late feedback, it's a lot to take in and this some complex stuff :) 

the internal/external types feel unnecessary somehow (maybe), can these internal links not be implemented using the onClick callback on DataLink? then there would be no need for this filtering

aocenas:
@torkelo 
The problem is inside the datasource where the DataLinks are created there is no information about what onClick should actually do or even what the url should be. To be able to do that it would have to know how explore urls look like or know how to open a split pane in explore.

So internal link is basically just the information of what datasource and what query should be opened but id does not know how to do that. Also "how" is context specific, like in explore we open new split onClick but in dashboard it will be just link to Explore without any special onClick behavior.

So the processing is different between those two types:
#### External link
1. Template is specified by the user or datasource
2. Template is interpolated, which needs the data values
3. LinkModel is created and can be used for an html link in the visualisation.
#### Interal link
1. Target datasource and query template is specified by user (DerivedFields in loki) or by datasource (x-ray link between trace list and single trace view)
2. Query template is interpolated which needs the data values (like the traceID from the list)
3. Based on application context, onClick functionality is added (in explore we open new split on click but still have <a> with href so it can be opened in separate tab. In dashboard there will be no onClick I assume just simple html <a>) and href is created (right now just explore url but to create it with proper params is something that is part of explore codebase and that can change)
4. LinkModel is created and can be used for an html link in the visualisation.

We may not need the onClick for the split function. At least internal feedback was that people rather open it in new tab anyway. But that would not take away the need to create the final href outside of the datasource and so we need to have different data model to pass the values from datasource to grafana. For example how to serialize explore href with query and all the state is not something that 3rd party datasource should try to duplicate as that would be pretty fragile.

Another thing I talked about with @dprokop is that we do not need to change the DataLink to add the type attribute. That would mean we do not need migration and the main purpose of that was to make type assertions easier but that is not strictly necessary.

So personally I do not see easier or nicer solution right now. In abstract terms the problem is that the information needed to create the final internal link is split between the datasource where the data link is defined and application itself where it will be presented. I do not see a way around that.

torkelo:
I see, thanks for the detail, sorry I could not join yesterday. 

> The problem is inside the datasource where the DataLinks are created there is no information about what onClick 

Understand that, the reason I proposed it was that you could add the onClick to the data links in the TableContainer in Explore, as with this solution you needed to that anyway. But that might make it more complicated.

> Another thing I talked about with @dprokop is that we do not need to change the DataLink to add the type attribute. That would mean we do not need migration and the main purpose of that was to make type assertions easier but that is not strictly necessary.

Good, I think the main issue I had with this was the added complexity of internal vs external checks and filtering if we can limit that a bit would be good.

dprokop:
@aocenas I've tried to come up with some other solution that would use onClick higher up in the link definition hierarchy. I was thinking that the `onClick` could be defined closer to the data source, with UI context provided. Unfortunately, we don't have a way to identify UI context in a generic way, and it feels like the complexity would be too high, especially with defining the basic behavior of simple hrefs that just do the redirect and allow opening in new tab. 

So, I think solution proposed in this PR should be fine. As discussed, I would just like to see if we can get around the DataLinkInternal/External interfaces in some way.

packages/grafana-data/src/field/fieldOverrides.ts
```diff
@@ -419,10 +412,22 @@ export const getLinksSupplier = (
       },
     };
 
-    info.href = replaceVariables(info.href, variables);
-    info.title = replaceVariables(info.title, variables);
-    info.href = locationUtil.processUrl(info.href);
+    if (link.internal) {
+      // For internal links at the moment only destination is Explore.
+      return mapInternalLinkToExplore(link, variables, {} as any, field);
```
aocenas, line 417: To support the internal links in dahboard table panel I needed to add it here. Find the general data flow of links confusing so personally don't think it should be done here as it isn't tied to overrides but not sure what would be a better place right now.

packages/grafana-data/src/types/dataLink.ts
```diff
@@ -28,16 +32,19 @@ export interface DataLink {
   // Not saved in JSON/DTO
   onClick?: (event: DataLinkClickEvent) => void;
 
-  // At the moment this is used for derived fields for metadata about internal linking.
-  meta?: {
-    datasourceUid?: string;
+  // If dataLink represents internal link this has to be filled. Internal link is defined as a query in a particular
+  // datas ource that we want to show to the user. Usually this results in a link to explore but can also lead to
+  // more custom onClick behaviour if needed.
+  internal?: {
```
aocenas, line 38: Change to the types since last commit. Removed 2 different types and just made the internal link attributes more type safe. At the same time though it's not ideal as you have to check for this and for functions taking DataLink you do not know if they will handle internal links properly.

packages/grafana-data/src/utils/dataLinks.ts
```diff
@@ -12,3 +17,80 @@ export const DataLinkBuiltInVars = {
   // name of the calculation represented by the value
   valueCalc: '__value.calc',
 };
+
+export function mapInternalLinkToExplore(
```
aocenas, line 21: I had to move this inside the grafana/data because otherwise there would be cyclic imports from the field overrides code.

packages/grafana-data/src/utils/url.ts
```diff
@@ -129,3 +131,24 @@ export const urlUtil = {
   appendQueryToUrl,
   getUrlSearchParams,
 };
+
+export function serializeStateToUrlParam(urlState: ExploreUrlState, compact?: boolean): string {
```
aocenas, line 135: Also this needed to be moved here because of cyclic imports.

public/app/features/explore/TableContainer.tsx
```diff
@@ -38,12 +41,23 @@ export class TableContainer extends PureComponent<TableContainerProps> {
...
+      // Should work fine in essence but still not the ideal way to pass props. In logs container we do this
+      // differently and sidestep this getLinks API on a dataframe
+      for (const field of tableResult.fields) {
+        field.getLinks = (config: ValueLinkConfig) => {
+          return getFieldLinksForExplore(field, config.valueRowIndex, splitOpen, range);
+        };
+      }
```
torkelo, line 58: Can't this be done in the explore ResultProcessor instead of in the render path?

packages/grafana-data/src/field/fieldOverrides.ts
torkelo: I think this could be added to ValueLinkConfig , so callers of getLinks can provide a custom handler for how internal links should be processed / evaluated

aocenas: The problem is how to get it there. This is called by the visualisation but I do not want the visualisation to know if it is inside the explore or dashboard or how to deal with the link. I think that should responsibility of the dashboard code or explore code. So I would have to pass it from there somehow anyway.

public/app/features/explore/TableContainer.tsx
aocenas: I need the splitOpen action and tbh I did not know how to get that in the ResultProcessor. I would like to move it a bit up to Explore component for example so it is done just once for both logs and table and this was done just because it would need more refactor. Also I actually still don't like that `field.getLinks` is used just to transfer this function around tbh I think this is a presentation concern not data concern so this feels iffy to me. Especially as I am mutating the dataFrame but otherwise I would need a structural copy, again to add a presentation info to it.

torkelo: > I need the splitOpen action and tbh I did not know how to get that in the ResultProcessor. 

I thin you can, ResultProcessor is called from a thunk action, all you need is to pass the dispatch function to the ResultProcessor and any dependant state .

aocenas: Well passing the dispatch function into a reducer seems kinda weird tbh. Not sure how much of an improvement that would be. Also I thought action payloads should be json serializable.

torkelo: ResultProcessor is not part of any reducer but an action. 

The code here modifying the links is a data mutation step so to me makes perfect sense to have that be part of the state & data processing and not the render path. 

We have no requirement that action or state payloads needs to be serializable

torkelo: but then again having onClick handlers assigned in the state processing is maybe equally bad in the opposite way

aocenas: Not sure if we are talking about the same thing, the processing is done here https://github.com/grafana/grafana/blob/7b7000c5bcb445dcf039d50ab4feed68842ef056/public/app/features/explore/state/reducers.ts#L550 so it is part of the reducer and it needs to get the dispatch from the action payload.

But yes in general I feel that onClick handler is directly associated with the presentation part ie Explore knowing what to do onClick (open a split window) so moving that to data processing still does not feel better.

I think ideally the getLinks should be decoupled from the dataFrame. As mentioned DataLinks should be part of the DataFrame but not LinkModel (which is what dataFrame.getLinks returns). So to me having a `getLinks(dataFrame, field, rowIndex)` function makes more sense than `dataFrame.getLinks`. Then we could just have custom linkContext to propagate that function around and so Explore will be wrapping the components with Explore specific implementation while Dashboard with something else. That would prevent having to drill down the props too deep while keeping those 2 concerns separate and there would be no need to modify the dataframe in the presentation layer.

packages/grafana-data/src/field/fieldOverrides.ts
```diff
@@ -419,10 +422,25 @@ export const getLinksSupplier = (
       },
     };
 
-    info.href = replaceVariables(info.href, variables);
-    info.title = replaceVariables(info.title, variables);
-    info.href = locationUtil.processUrl(info.href);
+    if (link.internal) {
```
aocenas, line 425: Main part that makes internal links work in dashboards

packages/grafana-data/src/types/explore.ts
```diff
@@ -0,0 +1,20 @@
+import { ExploreMode } from './datasource';
+import { RawTimeRange } from './time';
+import { LogsDedupStrategy } from './logs';
+
+export interface ExploreUrlState {
```
aocenas, line 5: Moved here from explore code to prevent import cycles

packages/grafana-data/src/utils/dataLinks.ts
```diff
@@ -12,3 +26,96 @@ export const DataLinkBuiltInVars = {
...
+}
+
+function interpolateQuery<T extends DataQuery = any>(
+  link: DataLink,
+  scopedVars: ScopedVars,
+  replaceVariables: InterpolateFunction
+): T {
+  let stringifiedQuery = '';
+  try {
+    stringifiedQuery = JSON.stringify(link.internal?.query || '');
```
aocenas, line 102: Not sure if this is the best way to interpolate variables inside random query object. One caveat is that this will replace it also in keys not only values.

packages/grafana-ui/src/components/Table/DefaultCell.tsx
```diff
@@ -20,7 +20,23 @@ export const DefaultCell: FC<TableCellProps> = props => {
     <div className={tableStyles.tableCell}>
       {link ? (
         <Tooltip content={link.title}>
-          <a href={link.href} target={link.target} title={link.title} className={tableStyles.tableCellLink}>
+          <a
```
aocenas, line 23: Needed so that in explore we can inject custom onClick behaviour to table.

packages/grafana-data/src/utils/dataLinks.ts
```diff
@@ -12,3 +26,96 @@ export const DataLinkBuiltInVars = {
   // name of the calculation represented by the value
   valueCalc: '__value.calc',
 };
+
+type Options = {
+  onClickFn?: (options: { datasourceUid: string; query: any }) => void;
+  replaceVariables: InterpolateFunction;
+  getDataSourceSettingsByUid: (uid: string) => DataSourceInstanceSettings | undefined;
```
aocenas, line 33: We need these 2 functions injected as we cannot import stuff from runtime here.

public/app/features/explore/TableContainer.tsx
torkelo: Ah, was sure ResultProcessor was part of the action before. 

> So to me having a getLinks(dataFrame, field, rowIndex) 

But how would we pass this function around? this is called by individual Cells in the table component. 

Your thinking Table component should have a getLinks prop so that caller can handle the DataLink -> LinkModel differently depending on where the Table is used and we pass this function all the way down to Cells? That could work. 

thoughts @dprokop ?

aocenas: > But how would we pass this function around? this is called by individual Cells in the table component.

> Your thinking Table component should have a getLinks prop so that caller can handle the DataLink -> LinkModel differently depending on where the Table is used and we pass this function all the way down to Cells? That could work.

Well I think something like a custom react context would be better so it does not need to be passed across too many levels of components but in essence yes it will be a separate dependency from the dataFrame.

So something like:
```jsx
// Explore.tsx
<GetLinksContextProvider value={(dataFrame, field, index) => { /* process links */}}>
  <Table data={dataFrame}/>
</GetLinksContextProvider>
```

```jsx
// Cell.tsx
// This can have some default fallback if nothing is supplied from parent so it does not crash in places where linking does not make sense
const getLinks = useGetLinksContext()
const links = getLinks(dataFrame, field, index) 
```

torkelo: nice, that looks very interesting / better

packages/grafana-data/src/types/data.ts
```diff
@@ -102,6 +103,8 @@ export interface Column {
   filterable?: boolean;
   unit?: string;
   custom?: Record<string, any>;
+  // To transfer data from dataFrame column
+  links?: DataLink[];
```
torkelo, line 107: this should no longer be needed

public/app/features/explore/TableContainer.tsx
dprokop: > So to me having a getLinks(dataFrame, field, rowIndex) function makes more sense than dataFrame.getLinks

I agree with this. Just thinking about the field variables would need to be calculated twice in this case for the field state and links separately. not sure tbh how big of a problem that is, maybe something we don't necessarily need to think too much about. I guess there won't be hundreds of thousands of fields in the data frames. I like the context idea too!

packages/grafana-data/src/types/data.ts
torkelo: as the table translation is no longer used for table in explore

aocenas: removed

aocenas:
@torkelo or @dprokop need an approve to merge this.

packages/grafana-data/src/field/fieldOverrides.test.ts
```diff
@@ -118,6 +118,7 @@ describe('applyFieldOverrides', () => {
           overrides: [],
         },
         replaceVariables: (value: any) => value,
+        getDataSourceSettingsByUid: undefined as any,
```
dprokop, line 121: I assume we are going to remove this in the refactor of `getLinks`?

aocenas: Yes, right now I do not see any easy way around this, but as we talked, this should be refactored into a (hopefully) more straightforward processing without the need to pass all these functions around.

packages/grafana-data/src/field/fieldOverrides.ts
dprokop: This part is to be refactored out from here too, right?

packages/grafana-data/src/types/explore.ts
dprokop: That's because of `mapInternalLinkToExplore` becoming part of grafana-data as for now?

dprokop, line 5:
```suggestion
/** @internal */
export interface ExploreUrlState {
```

packages/grafana-data/src/types/explore.ts
```diff
@@ -0,0 +1,20 @@
...
+  datasource: string;
+  queries: any[]; // Should be a DataQuery, but we're going to strip refIds, so typing makes less sense
+  mode: ExploreMode;
+  range: RawTimeRange;
+  ui: ExploreUIState;
+  originPanelId?: number;
+  context?: string;
+}
+
+export interface ExploreUIState {
```
dprokop, line 15:
```suggestion
/** @internal */
export interface ExploreUIState {
```

packages/grafana-data/src/utils/url.ts
dprokop: Can we be more explore-especific here in naming: `serializeStateToUrlParam` -> `serializeExploreStateToUrlParam`? I guess this will be brought back to core when we refactor.

dprokop (approved):
I think this looks good for now, loads of work @aocenas! Happy to see this happening to reveal the issues with the current links processing logic.

packages/grafana-data/src/types/explore.ts
aocenas: what does the internal mean?

torkelo (approved):

PR #25626, opened 2020-06-16
cloudwatch: avoid sorting the result

### LINKED ISSUE #25494: Cloudwatch SORT metrics math function is not respected for more than 10 results
**What happened**:
When using [`SORT(SEARCH(...), SUM, DESC)`](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/using-metric-math.html#metric-math-syntax) to sort the results of a SEARCH query with 10+ results, the time series are sorted based on their labels rather than the order returned by Cloudwatch.

Because the rank of the result is affixed to the label, this results in the ordering of the rank 1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9 (assuming the result contains 12 time series). I suspect [this](https://github.com/grafana/grafana/blob/5f923be2bf67deadab90dcd328e84bee8a9b41cf/pkg/tsdb/cloudwatch/response_parser.go#L69) is the line that causes this.

I expected Grafana to respect the order returned by Cloudwatch (1, 2, 3, ...), which helps in visualizing multiple time series when stacked.

SEARCH and SORT for metrics that return more than 10 time series and inspect the response array in Grafana.

N/A

**Environment**:
- Grafana version: 7.0.3

bergquist:
Grafana should respect the returned sort order
by default. If the users wants another order
we should use transformations to change it.

fixes #25494

### DIFF
--- a/pkg/tsdb/cloudwatch/response_parser.go
+++ b/pkg/tsdb/cloudwatch/response_parser.go
@@ -66,7 +66,6 @@ func parseGetMetricDataTimeSeries(metricDataResults map[string]*cloudwatch.Metri
 	for k := range metricDataResults {
 		metricDataResultLabels = append(metricDataResultLabels, k)
 	}
-	sort.Strings(metricDataResultLabels)
 
 	partialData := false
 	result := tsdb.TimeSeriesSlice{}


aknuds1 (approved):

marefr (commented):
Not sure this will guarantee the order of the source result? Go's map iteration order is random if not mistaken. Reason sorting have been added is to at least be able to write deterministic test.

aknuds1:
@marefr I actually agree, I didn't notice that the array came from a map.

aknuds1 (changes requested):
This is probably too simplistic. As @marefr points out, the order is non-deterministic due to the labels coming from a map.

bergquist:
I was kind of expecting this :)

> Go's map iteration order is random if not mistaken

Since Cloudwatch returns the result as an array I guess we need to save the order during parsing and sort them again before returning. 

Tests should definitely be deterministic.

Closing this and writing a comment on the issue.

PR #25628, opened 2020-06-16
Show the add new channel button when a user has no notifications

gehtman:
Show the add notification channel button also when there are no notification channels

Users without notifications don't see the add notification button


peterholmberg:
@gehtman we have a `<EmptyListCTA />` component if the list is empty. Not sure why we want two buttons doing the same thing?

torkelo:
Agree, we already have a big  button for this

PR #25629, opened 2020-06-16
API: Recognize MSSQL data source URLs

### LINKED ISSUE #25067: Cannot connect to SQL Named Instance
**What happened**:
When trying to connect to a named instance when using version 7.0 we cannot specify a named instance. When using a backslash (`\`) to seperate the server name from the instance name, it gives us an error. See attached.

**What you expected to happen**: We should we able to connect as in the past to the server. This worked in version 6.7.2

Try to create a connection to a new instance using server\InstanceName. Is seems that you interpret each single slash (back slash) as two slashes.

**Anything else we need to know?**: No

**Environment**:
- Grafana version: 7.0
- Data source type & version: SQL Server
- OS Grafana is installed on: CentOS
- User OS & Browser: Chrome

- Grafana plugins:
- Others:

ServerName
Connection

aknuds1:
Implement MSSQL specific data source URL validation logic.

Fixes #25067.

I read the MSSQL [driver](https://github.com/denisenkom/go-mssqldb) documentation in order to figure out valid connection strings. I figure I should probably add support for the two other MSSQL URL formats too, but starting out with this.

### DIFF
--- a/pkg/api/pluginproxy/ds_proxy_test.go
+++ b/pkg/api/pluginproxy/ds_proxy_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/grafana/pkg/api/datasource"
 	"github.com/grafana/grafana/pkg/components/securejsondata"
 	"github.com/grafana/grafana/pkg/models"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


aknuds1:
@reuvygroovy Can you please test this PR and see if it works for you?

reuvygroovy:
How can I test it? What should I install/replace?

aknuds1:
@reuvygroovy Are you able to build Grafana from source?

reuvygroovy:
No, haven't done that yet. If you explain how though I am willing to try.

aknuds1:
Read the [developer guide](https://github.com/grafana/grafana/blob/master/contribute/developer-guide.md).

papagian (approved):
LGTM
It would be nice to have some tests for the different cases that `parseURL` supports.
Also, off the scope of this one, ideally each datasource should have each one validation method (by implementing the same interface) but this requires a lot of refactoring.

aknuds1:
@papagian Those tests are implemented via TestNewDataSourceProxy_MSSQL. If mssql.ParseURL has a bug it should be caught there. It's preferable to test as few implementation details as possible.

papagian:
> @papagian Those tests are implemented in TestNewDataSourceProxy_MSSQL. If mssql.ParseURL has a bug it should be caught there. It's preferable to test as few implementation details as possible, to reduce maintenance/cognitive load.

correct! forget about my comment!

pkg/api/pluginproxy/ds_proxy_test.go
```diff
@@ -583,11 +584,62 @@ func TestNewDataSourceProxy_ProtocolLessURL(t *testing.T) {
...
 	cfg := setting.Cfg{}
 	plugin := plugins.DataSourcePlugin{}
+
 	_, err := NewDataSourceProxy(&ds, &plugin, &ctx, "api/method", &cfg)
 
 	require.NoError(t, err)
 }
 
+// Test wth MSSQL type data sources.
+func TestNewDataSourceProxy_MSSQL(t *testing.T) {
```
marefr, line 594: Testing the proxy with MSSQL doesn't make sense.

aknuds1: @marefr Why is that a problem? The proxy supports data sources of MSSQL type, same as the handlers for adding and updating data sources. The benefit of testing from this level is that you're testing fewer implementation details, and you can see that the chain of code is working.

aknuds1: @marefr OK after our conversation I understand what you mean. I'll try to look into it.

aknuds1:
@reuvygroovy Can you try the latest [nightly](https://grafana.com/grafana/download/7.1.0-96b2dad4pre)? It should have the fix.

reuvygroovy:
Works! Awesome! Thanks :-)

aknuds1:
@reuvygroovy Great!

gitszakal:
Hello Team,
Unfortunately I have the same issue.

I have an SQLExpess database and a Windows installed Grafana 7.5.4. on the same PC.
Host: localhost\SQLEXPRESS:1433 brings this failure:
_Validation error, invalid URL: "localhost\\SQLEXPRESS:1433"_
With double \\ the host I get this:
_Validation error, invalid URL: "localhost\\\\SQLEXPRESS:1433"_
I have tried to add data source PostgreSQL, but I get the same issue.
Just with name SQLEXPRESS:1433 brings this failure:
_dial tcp: lookup SQLEXPRESS: no such host_

I am not an experienced programmer, could you please help me on this, explaining on an easy level? :)
Thank you in advance

reuvygroovy:
Did you try computer name instead of localhost?

Also try

.\SQLExpress

Try also without port

gitszakal:
Thank you for your willing to help :)
Yes, I tried with computer name, with, without port, but always the same.
.\ I haven't tried before, but now this brings the same message as well:
Validation error, invalid URL: ".\\\\SQLEXPRESS:1433"

reuvygroovy:
can you connect with Management Studio on the same PC? on a different computer?

gitszakal:
Connection to the database on the same PC works with Management Studio.
I haven't tried from different computer, but maybe I can try to find a way to try... Does it makes any difference on the host name connection?

PR #25631, opened 2020-06-16
Chore: use fork-ts-checker plugin for eslint check

zoltanbedi:

This pr removes eslint-loader from webpack.configs so the build time can be faster. Also upgrades fork-ts-checker-plugin to version 5 so it will be compatible with webpack 5.

Some measurements on my pc:

yarn start (without eslint-loader): 73253ms
yarn start (with eslint-loader): 113371ms

yarn build (without eslint-loader): 248.47s
yarn build (with eslint-loader): 289.10s

in both scripts we can compile almost **~1 minute** faster this way.


aknuds1 (approved):
Not really familiar with it, but LGTM

torkelo (approved):
Looks good!

stevenvachon (approved):
Nice! I wasn't aware of this and it looks as though we may have been doubling up checks. Perhaps we should also change our "lint" npm scripts to use webpack, so that our paths are only in one place.

zoltanbedi:
> Perhaps we should also change our "lint" npm scripts to use webpack, so that our paths are only in one place.

I'd rather not do that. This change only going to lint packages that is compiled through webpack. So no test files for example.

I had to do some changes because for my first try it only checked files under public/app. So I've added package paths that are compiled. Compilation was still fast but eslint check become slower. So I've enabled cache for eslint which made it **BLAZING** fast again.

scripts/webpack/webpack.dev.js
```diff
@@ -89,7 +89,17 @@ module.exports = (env = {}) =>
...
             eslint: {
               enabled: true,
-              files: ['public/app/**/*.{ts,tsx}', 'e2e/suite1/**/*.{ts,tsx}', 'public/test/**/*.{ts,tsx}'],
+              files: [
+                'public/app/**/*.{ts,tsx}',
+                'packages/grafana-ui/src/**/*.{ts,tsx}',
+                'packages/grafana-data/src/**/*.{ts,tsx}',
+                'packages/grafana-runtime/src/**/*.{ts,tsx}',
+                'packages/grafana-e2e-selectors/src/**/*.{ts,tsx}',
+                'packages/jaeger-ui-components/src/**/*.{ts,tsx}',
```
stevenvachon, line 98: wouldn't `packages/**/*.{ts,tsx}` suffice?

zoltanbedi: good catch but sadly I couldn’t make that work. It failed with an error I couldn’t manage to solve.

stevenvachon: Okay, maybe add a comment explaining why. Aside from that, the changes LGTM.

stevenvachon (approved):

PR #25635, opened 2020-06-16
plugins: Validate that all files correspond to manifest

aknuds1:
Modify plugin signature validation to check that all files included with plugin correspond to manifest.


pkg/plugins/manifest.go
```diff
@@ -82,49 +84,130 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
 	return manifest, nil
 }
 
+func getChksums(dpath, manifestPath string, log log.Logger) (map[string]string, error) {
```
kylebrandt, line 87: while not exported, I think the complexity of this warrants a func comment explaining what it does.

pkg/plugins/manifest.go
```diff
@@ -82,49 +84,130 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+		defer f.Close()
+		if _, err := io.Copy(h, f); err != nil {
+			return err
+		}
+
+		relPath, err := filepath.Rel(dpath, path)
+		if err != nil {
+			return err
+		}
+		chksums[relPath] = fmt.Sprintf("%x", h.Sum(nil))
```
kylebrandt, line 155: Curious why appending nil? Maybe a comment?

pkg/plugins/manifest.go
```diff
@@ -82,49 +84,130 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
-		if _, err := io.Copy(h, f); err != nil {
-			log.Warn("Couldn't read plugin file", "plugin", plugin.Id, "filename", fp)
-			return PluginSignatureModified
+		gotHash, ok := chksums[p]
+		if !ok {
+			log.Warn("Plugin file from manifest is missing", "plugin", plugin.Id, "filename", fp)
 		}
-		sum := hex.EncodeToString(h.Sum(nil))
-		if sum != hash {
+		if gotHash != hash {
```
kylebrandt, line 203: Your call as I'm not so sure on this. But since the zero value of both `gotHash` and `hash` is an empty string and we have continued from above, this could maybe too weak of a check if somehow one ends up with empty string equals empty string. But possibly in the context this is just be overly defensive programming.

Alternatively one could check that it isn't empty (assuming it never should be?), or maybe make it nullable (a pointer).  If empty string check would work that is probably simpler.

pkg/plugins/plugins.go
```diff
@@ -140,7 +140,11 @@ func (pm *PluginManager) Init() error {
 		if p.IsCorePlugin {
 			p.Signature = PluginSignatureInternal
 		} else {
-			p.Signature = getPluginSignatureState(pm.log, p)
+			var err error
```
kylebrandt, line 143: technically I think you could remove this line, but just style so up to you.

kylebrandt (approved):
Looks good to me. I don't know to much about the manifests, but it makes sense.

One request for a code function comment, and then other items at your discretion.

pkg/plugins/manifest.go
aknuds1: Thanks @kylebrandt, will fix it.

pkg/plugins/manifest.go
aknuds1: @kylebrandt `h.Sum(nil)` is the standard way of getting the checksum.

pkg/plugins/manifest.go
aknuds1: @kylebrandt I guess an empty 256-bit checksum would be a bug? It should just be defensive programming to guard against this, but if we want to, we would also have to add these guards in grafana/build-pipeline, in the code to generate the manifest.

pkg/plugins/plugins.go
aknuds1: @kylebrandt I would have to try, I'm not sure if `err` is defined otherwise.

aknuds1: @kylebrandt Oh yeah, it does work since `err` is defined already in the function.

pkg/plugins/manifest.go
aknuds1: A bit weird, but that's Go :)

pkg/plugins/manifest.go
```diff
@@ -82,49 +84,133 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+			info, err := os.Stat(finalPath)
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				return nil
+			}
+
+			if _, err := filepath.Rel(dpath, finalPath); err != nil {
+				return fmt.Errorf("symbolic link %q targets a file outside of the plugin directory: %q", path, finalPath)
```
ryantxu, line 123: why do we care?

ryantxu:
IIUC, this means that a plugin could not write to the install folder?  settings or license etc would need to go elsewhere?

What is this trying to protect against?  From my perspective, we really just need to make sure the executable and any `module.ts` is in the manifest -- I worry that forcing everything in the extracted folder to exist in the manifest will cause problems, and I don't understand what it protects

pkg/plugins/manifest.go
aknuds1: @ryantxu Why should a plugin contain a symlink that points to an external file? At the very least, the current checksumming of plugins won't really work, since we'll be checksumming an external file at the time of packaging, that will very likely point to a file with different content (or not existing at all) when ran by Grafana. I'm also worried it will be an attack vector, so best to avoid the scenario don't you think?

aknuds1:
> IIUC, this means that a plugin could not write to the install folder? settings or license etc would need to go elsewhere?

That's a good point @ryantxu, I guess we'll have to think about that. Is it possible that the manifest should only be evaluated when the plugin is installed?

> What is this trying to protect against? From my perspective, we really just need to make sure the executable and any `module.ts` is in the manifest -- I worry that forcing everything in the extracted folder to exist in the manifest will cause problems, and I don't understand what it protects

The idea is to make the manifest account for all files installed by the plugin. Isn't the manifest otherwise relatively worthless, if files not listed by it are ignored?

ryantxu:
The purpose of the manifest is to:
* verify that the author created the file
* make sure the executables/modules we run are the ones that were published

For security, I think only the executable and `**/module.js` are required.

I am nervous about changing the manifest contract to mean *everything* in folder came from the bundled zip file.  Things like the oracle plugin include a few setup files that can/should be deleted after install.

I assume the symlink code is solving something?  problem is it protecting against?  I also don't understand how this is worthless if it does not list everything?

ryantxu:
For security, i suppose we could add a check that makes sure:
1. `**/module.js` is listed in the manifest
1. `**/plugin.json` is listed in the manifest
1. for each plugin.json, look for the [executable
](https://github.com/grafana/google-sheets-datasource/blob/master/src/plugin.json#L7) property and make sure `${value}*` is in the manifest

When I say "for each" the main issue is with app plugins that define a tree of plugins -- the manifest will only exist at the root element.

aknuds1:
@ryantxu OK, so you're saying that instead of verifying that everything in the plugin is according to the manifest, we should only check **/module.js, **/plugin.json and each plugin.json's executable (plus whatever else is listed in the manifest)?

marefr:
Everything started here: https://github.com/grafana/grafana-image-renderer/pull/152
Continued here: https://github.com/grafana/build-pipeline/pull/85

With the image renderer plugin we ship chromium as part of the plugin and in this case the plugin runs another executable (chromium). To me it makes sense to include all of its files in manifest. My first idea was to skip following symbolic links. After some discussions with Arve we agreed upon following the symlinks and validate them before signing the plugin.

> For security, I think only the executable and **/module.js are required.

Feels a bit insecure since if I'm not mistaken a plugin file (static resources at least) can be requested from the UI by URL and that could be a potential security issue. Currently we don't include images in manifest which I really don't see why either. 

> Things like the oracle plugin include a few setup files that can/should be deleted after install.

From a security perspective I wouldn't want to run those scripts unless I can verify they haven't been tampered with. Can't the files be left in the plugin folder?

> Should we also have a signature to verify the manifest isn't tampered with?

Don't we already have that? The content of plugin manifest file is signed with our private key and Grafana uses the public key to validate it have been signed using the private key..

aknuds1:
> > Should we also have a signature to verify the manifest isn't tampered with?
> 
> Don't we already have that? The content of plugin manifest file is signed with our private key and Grafana uses the public key to validate it have been signed using the private key..

@marefr I honestly don't remember if we have that right now. We need to discuss this further together I guess, in order to figure out the points in this thread.

marefr:
@aknuds1 after discussion with @ryantxu yesterday we came to the conclusion that these changes are not needed in a first step. For now we're fairly confident that skipping including symbolic links in manifest, but still validating in build pipeline tool is the most straight forward thing to start with.

aknuds1:
Closing since we've reached the conclusion that we should be verifying integrity at installation time instead.

Issue #11409, opened 2018-03-28
Installing & enabling plugins through provisioning


xlson:
It's currently possible to [provision](http://docs.grafana.org/administration/provisioning/) datasources and dashboards but not plugins. When using the Docker container this is solved to some extent by being able to provision/install plugins using an env variable. 

This leaves app plugins without provisioning as they need to be enabled after they have been installed. It also leaves installs without docker without provisioning support for plugins.

### Proposal:
- extend provisioning to support plugins
- extend provisioning to support app plugins

Related to grafana/grafana-docker#133

iMartyn:
I think the `grafana-cli` needs an 'plugins enable' command equivalent to `plugins install` that it now has, then the docker `run.sh` can have this functionality.

PaulGallon:
This is the last thing we need to keep our Grafana environment entirely in Git! Love the new provisioning features.

kethahel99:
Any traction on this?

xlson:
@kethahel99 not yet, but we're definitely keeping this in mind. The more attention this gets, the likelier it is that we implement it.

onyxet:
@xlson Hi. Have you tried `GF_INSTALL_PLUGINS: grafana-clock-panel,grafana-simple-json-datasource` this for plugin provisioning? I think it's solving all possible issues with plugins

xlson:
@onyxet Unfortunately that only solves the problem for users running Grafana with Docker. This issue is aimed at developing a solution for non-docker users.

onyxet:
@PaulGallon Sorry but can't agree with you. I just checked `GF_INSTALL_PLUGINS: camptocamp-prometheus-alertmanager-datasource` this env with `grafana/grafana:5.3.4` image and datasource is ready out the box. So I believe that for docker users it's working as mentioned

onyxet:
@xlson I see. I guess that systemd configuration with env configuration like:
`Environment=GF_INSTALL_PLUGINS: camptocamp-prometheus-alertmanager-datasource` might be working solution. Here is some doc about that:
http://docs.grafana.org/installation/configuration/#using-environment-variables

Or I didn't understand the issue?

xlson:
@onyxet Some plugins are ready right out of the box, some aren't and require extra configuration (which requires provisioning). I understand that you are trying to be helpful but I'm afraid the issue documented here is not solved by linking to Grafana docs as the issue is that Grafana itself does not do provisioning of plugins.

[...]

Issue #25526, opened 2020-06-11
Provisioning plugins using config files 


bergquist:
It should be possible to provision Grafana with plugins using config files. On startup, Grafana should make sure that the specified plugins are installed and configured correctly. If the plugins are missing Grafana should fetch them using the same code as the CLI. If the plugins are not set up as configured, Grafana should update the settings for that plugin (Only app plugins can be configured currently)

The config files should allow the operator to specify which plugin that should be installed and at what version. The initial version should support specific versions or `latest`. Specifying what versions that accepted new patches (ex ^1.2.0) etc is out of scope for the initial implementation. 

Plugins can only be installed once per instance but can have app settings per org. So I suggest that each plugin can have a settings array of how that plugin should be configured in Grafana database. Ex

```yaml
apiVersion: 1

plugins:
  - name: grafana-piechart-panel
    version: 1.2.3
  - name: grafana-alerting-ui-app
    version: latest
    settings: 
    - orgId: 1	
      enabled: true
      pinned: true
      jsonData: 
        url: https://some.cloudy.service
      secureJsonData:
        token: <foo-bar-baz>
    - orgId: 2	
      enabled: true
      jsonData: 
        url: https://some.cloudy.service
      secureJsonData:
        token: <foo-bar-baz>
```

marefr:
Great. Not sure that installing the plugins needs to be part of the solution (in a first step). Because installing requires the instance to be restarted.

marefr:
Duplicate of #11409

PR #25649, opened 2020-06-17
Provisioning: Adds support for enabling app plugins

marefr:
Adds support for enabling app plugins using provisioning. Installation of plugins must be done in advance for now. 

Ref #11409

Not sure if the configuration makes sense, Decided to put focus on apps instead of plugins in general since we currently only allow app plugins to be enabled. Data sources are already handled separated. Maybe would make more sense to have plugin provisioning with apps sections, very similar but allows it to be extended more easily in the future if needed.

Would it make it more explicit to have separate sections for enabled/disabled apps similar to what data sources and notifiers have (datasources, deletedDatasources, e.g.
```yaml
disabledApps:
  - type: some-id

enabledApps:
  - type: some-other-id
```

Feedback very welcome.


pkg/api/admin_provisioning.go
```diff
@@ -21,6 +22,14 @@ func (server *HTTPServer) AdminProvisioningReloadDatasources(c *models.ReqContex
 	return Success("Datasources config reloaded")
 }
 
+func (server *HTTPServer) AdminProvisioningReloadApps(c *models.ReqContext) Response {
+	err := server.ProvisioningService.ProvisionApps()
+	if err != nil {
+		return Error(500, "", err)
```
aknuds1, line 28: I'm not sure it's safe to return the error message to the client? For the sake of security, I think we should log the error here:

```suggestion
		log.Error("Provisioning apps failed", "error", err)
		return Error(500, "", "Internal server error")
```

pkg/plugins/queries.go
```diff
@@ -85,3 +85,9 @@ func GetEnabledPlugins(orgId int64) (*EnabledPlugins, error) {
 
 	return &enabledPlugins, nil
 }
+
+// IsAppInstalled check if an app plugin with provided plugin ID is installed.
```
aknuds1, line 89:
```suggestion
// IsAppInstalled checks if an app plugin with provided plugin ID is installed.
```

pkg/services/provisioning/apps/app_provisioner.go
```diff
@@ -0,0 +1,84 @@
...
+		query := &models.GetPluginSettingByIdQuery{OrgId: app.OrgID, PluginId: app.PluginID}
+		err := bus.Dispatch(query)
+		if err != nil && err != models.ErrPluginSettingNotFound {
+			return err
+		}
+
+		if err != models.ErrPluginSettingNotFound {
+			app.PluginVersion = query.Result.PluginVersion
+			app.Pinned = query.Result.Pinned
+		}
```
aknuds1, line 51: I think the following makes the error handling much easier to verify:
```suggestion
		if err != nil {
			if err != models.ErrPluginSettingNotFound {
				return err
			}
		} else {
			app.PluginVersion = query.Result.PluginVersion
			app.Pinned = query.Result.Pinned
		}
```

pkg/services/provisioning/apps/app_provisioner.go
```diff
@@ -0,0 +1,84 @@
...
+
+		ap.log.Info("updating app from configuration ", "type", app.PluginID, "enabled", app.Enabled)
```
aknuds1, line 53: I think the standard is using sentence case when logging:
```suggestion
		ap.log.Info("Updating app from configuration ", "type", app.PluginID, "enabled", app.Enabled)
```

pkg/services/provisioning/apps/app_provisioner_test.go
```diff
@@ -0,0 +1,95 @@
...
+	t.Run("Should return error when config reader returns error", func(t *testing.T) {
+		expectedErr := errors.New("test")
+		reader := &testConfigReader{err: expectedErr}
+		ap := AppProvisioner{log: log.New("test"), cfgProvider: reader}
+		err := ap.applyChanges("")
+		require.Equal(t, expectedErr, err)
+	})
+
+	t.Run("Should apply configurations", func(t *testing.T) {
+		bus.AddHandler("test", func(query *models.GetOrgByNameQuery) error {
```
aknuds1, line 23: I really think it's best to remove these handlers after the test, via `t.Cleanup`, so they won't be accidentally used by other tests.

pkg/services/provisioning/apps/config_reader.go
```diff
@@ -0,0 +1,130 @@
...
+	return &configReaderImpl{log: logger}
+}
+
+func (cr *configReaderImpl) readConfig(path string) ([]*appsAsConfig, error) {
+	var apps []*appsAsConfig
+	cr.log.Debug("Looking for app provisioning files", "path", path)
+
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		cr.log.Error("Failed to read app provisioning files from directory", "path", path, "error", err)
```
aknuds1, line 33: Should this be a warning instead, since we don't treat it as an error?

pkg/services/provisioning/apps/config_reader.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			if app != nil {
+				apps = append(apps, app)
+			}
+		}
+	}
+
+	cr.log.Debug("Validating apps")
+	if err = validateRequiredField(apps); err != nil {
```
aknuds1, line 52:
```suggestion
	if err := validateRequiredField(apps); err != nil {
```

pkg/services/provisioning/apps/config_reader.go
```diff
@@ -0,0 +1,130 @@
...
+	err = validateApps(apps)
+	if err != nil {
+		return nil, err
+	}
+
+	return apps, nil
+}
+
+func (cr *configReaderImpl) parseAppConfig(path string, file os.FileInfo) (*appsAsConfig, error) {
+	filename, _ := filepath.Abs(filepath.Join(path, file.Name()))
```
aknuds1, line 67: I think the error should be checked, to be on the safe/consistent side.

pkg/services/provisioning/apps/config_reader.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func validateRequiredField(apps []*appsAsConfig) error {
+	for i := range apps {
+		var errStrings []string
+		for index, app := range apps[i].Apps {
+			if app.PluginID == "" {
+				errStrings = append(
+					errStrings,
+					fmt.Sprintf("App item %d in configuration doesn't contain required field type", index+1),
```
aknuds1, line 89: Errors shouldn't be [capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).
```suggestion
					fmt.Sprintf("app item %d in configuration doesn't contain required field type", index+1),
```

pkg/services/provisioning/apps/config_reader.go
```diff
@@ -0,0 +1,130 @@
...
+
+func validateApps(apps []*appsAsConfig) error {
+	for i := range apps {
+		if apps[i].Apps == nil {
+			continue
+		}
+
+		for _, app := range apps[i].Apps {
+			if !plugins.IsAppInstalled(app.PluginID) {
+				return fmt.Errorf("plugin not installed: %s", app.PluginID)
```
aknuds1, line 110: Easier debugging in case ID contains spaces, or should happen to be empty:
```suggestion
				return fmt.Errorf("plugin not installed: %q", app.PluginID)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+	emptyFolder       = "./testdata/test-configs/empty_folder"
+	unknownApp        = "./testdata/test-configs/unknown-app"
+	correctProperties = "./testdata/test-configs/correct-properties"
+)
+
+func TestConfigReader(t *testing.T) {
+	t.Run("Broken yaml should return error", func(t *testing.T) {
+		reader := newConfigReader(log.New("test logger"))
+		_, err := reader.readConfig(brokenYaml)
+		require.NotNil(t, err)
```
aknuds1, line 24:
```suggestion
		require.Error(t, err)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+	})
+
+	t.Run("Skip invalid directory", func(t *testing.T) {
+		cfgProvider := newConfigReader(log.New("test logger"))
+		cfg, err := cfgProvider.readConfig(emptyFolder)
+		if err != nil {
+			t.Fatalf("readConfig return an error %v", err)
+		}
```
aknuds1, line 32:
```suggestion
		require.NoError(t, err)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+		require.Len(t, cfg, 0)
+	})
+
+	t.Run("Unknown app should return error", func(t *testing.T) {
+		cfgProvider := newConfigReader(log.New("test logger"))
+		_, err := cfgProvider.readConfig(unknownApp)
+		require.NotNil(t, err)
```
aknuds1, line 39:
```suggestion
		require.Error(t, err)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+		require.Equal(t, "plugin not installed: nonexisting", err.Error())
+	})
+
+	t.Run("Read incorrect properties", func(t *testing.T) {
+		cfgProvider := newConfigReader(log.New("test logger"))
+		_, err := cfgProvider.readConfig(incorrectSettings)
+		require.NotNil(t, err)
```
aknuds1, line 46:
```suggestion
		require.Error(t, err)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+		require.Equal(t, "App item 1 in configuration doesn't contain required field type", err.Error())
```
aknuds1, line 47:
```suggestion
		require.Equal(t, "app item 1 in configuration doesn't contain required field type", err.Error())
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+	})
+
+	t.Run("Can read correct properties", func(t *testing.T) {
+		plugins.Apps = map[string]*plugins.AppPlugin{
+			"test-plugin":   {},
+			"test-plugin-2": {},
+		}
+
+		_ = os.Setenv("ENABLE_PLUGIN_VAR", "test-plugin")
```
aknuds1, line 56:
```suggestion
		err := os.Setenv("ENABLE_PLUGIN_VAR", "test-plugin")
		require.NoError(t, err)
```

pkg/services/provisioning/apps/config_reader_test.go
```diff
@@ -0,0 +1,87 @@
...
+		defer func() {
+			_ = os.Unsetenv("ENABLE_PLUGIN_VAR")
+		}()
```
aknuds1, line 59:
```suggestion
		t.Cleanup(t, func() {
			_ = os.Unsetenv("ENABLE_PLUGIN_VAR")
		})
```

pkg/services/provisioning/apps/types.go
```diff
@@ -0,0 +1,57 @@
+package apps
+
+import "github.com/grafana/grafana/pkg/services/provisioning/values"
+
+// appsAsConfig is normalized data object for apps config data. Any config version should be mappable
```
aknuds1, line 5:
```suggestion
// appsAsConfig is a normalized data object for apps config data. Any config version should be mappable.
```

pkg/services/provisioning/apps/types.go
```diff
@@ -0,0 +1,57 @@
...
+type appFromConfigV0 struct {
+	OrgID          values.Int64Value     `json:"org_id" yaml:"org_id"`
+	OrgName        values.StringValue    `json:"org_name" yaml:"org_name"`
+	Type           values.StringValue    `json:"type" yaml:"type"`
+	Disabled       values.BoolValue      `json:"disabled" yaml:"disabled"`
+	JSONData       values.JSONValue      `json:"jsonData" yaml:"jsonData"`
+	SecureJSONData values.StringMapValue `json:"secureJsonData" yaml:"secureJsonData"`
+}
+
+// appsAsConfigV0 is mapping for zero version configs. This is mapped to its normalised version.
```
aknuds1, line 31:
```suggestion
// appsAsConfigV0 is a mapping for zero version configs. This is mapped to its normalised version.
```

pkg/services/provisioning/apps/types.go
```diff
@@ -0,0 +1,57 @@
...
+type appsAsConfigV0 struct {
+	Apps []*appFromConfigV0 `json:"apps" yaml:"apps"`
+}
+
+// mapToNotificationFromConfig maps config syntax to normalized notificationsAsConfig object. Every version
```
aknuds1, line 36:
```suggestion
// mapToAppsFromConfig maps config syntax to a normalized notificationsAsConfig object. Every version
```

pkg/services/provisioning/apps/types.go
```diff
@@ -0,0 +1,57 @@
...
+// of the config syntax should have this function.
+func (cfg *appsAsConfigV0) mapToAppFromConfig() *appsAsConfig {
```
aknuds1, line 38: I think mapToAppsFromConfig is more correct, since it returns apps in plural.
```suggestion
func (cfg *appsAsConfigV0) mapToAppsFromConfig() *appsAsConfig {
```

pkg/services/provisioning/provisioning.go
```diff
@@ -107,6 +118,12 @@ func (ps *provisioningServiceImpl) ProvisionDatasources() error {
 	return errutil.Wrap("Datasource provisioning error", err)
 }
 
+func (ps *provisioningServiceImpl) ProvisionApps() error {
+	appPath := path.Join(ps.Cfg.ProvisioningPath, "apps")
+	err := ps.provisionApps(appPath)
+	return errutil.Wrap("App provisioning error", err)
```
aknuds1, line 124:
```suggestion
	return errutil.Wrap("app provisioning error", err)
```

aknuds1 (changes requested):
Please see my suggestions and questions.

conf/provisioning/apps/sample.yaml
```diff
@@ -0,0 +1,11 @@
+# # config file version
+apiVersion: 1
+
+# apps:
```
bergquist, line 4: Thinking a few steps ahead I think I would like this to be part of provisioning plugins. Having both plugins and app provisioning seems slightly confusing to me.

torkelo:
Nice! this was quickly done!

conf/provisioning/apps/sample.yaml
aknuds1: @bergquist Good point - I was also wondering if it shouldn't be for plugins, but I don't know the context.

bergquist: As described in https://github.com/grafana/grafana/issues/25526 
But if totally fine, not downloading plugins to start with.

docs/sources/administration/provisioning.md
```diff
@@ -208,6 +208,36 @@ datasources:
       httpHeaderValue2: 'Bearer XXXXXXXXX'
 ```
 
+## Apps
+
+> This feature is available from v7.1
+
+It's possible to manage (enable/disable) apps in Grafana by adding one or more yaml config files in the [`provisioning/apps`]({{< relref "../installation/configuration/#provisioning" >}}) directory. Each config file can contain a list of `apps` that will be updated during start up. Grafana will update each app to match the configuration file.
```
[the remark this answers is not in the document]
oddlittlebird: Please change "It's possible" to "You can" in all instances.

docs/sources/administration/provisioning.md
```diff
@@ -65,7 +65,7 @@ Currently we do not provide any scripts/manifests for configuring Grafana. Rathe
...
+It's possible to manage datasources in Grafana by adding one or more yaml config files in the [`provisioning/datasources`]({{< relref "../installation/configuration/#provisioning" >}}) directory. Each config file can contain a list of `datasources` that will be added or updated during start up. If the datasource already exists, Grafana will update it to match the configuration file. The config file can also contain a list of datasources that should be deleted. That list is called `deleteDatasources`. Grafana will delete datasources listed in `deleteDatasources` before inserting/updating those in the `datasource` list.
```
oddlittlebird, line 68:
```suggestion
You can manage data sources in Grafana by adding one or more YAML config files in the [`provisioning/datasources`]({{< relref "../installation/configuration/#provisioning" >}}) directory. Each config file can contain a list of `datasources` that will be added or updated during start up. If the data source already exists, then Grafana updates it to match the configuration file. The config file can also contain a list of data sources that should be deleted. That list is called `deleteDatasources`. Grafana deletes data sources listed in `deleteDatasources` before inserting or updating those in the `datasource` list.
```

docs/sources/administration/provisioning.md
```diff
@@ -208,6 +208,36 @@ datasources:
...
+      key: value
+    # <map> fields that will be converted to json, encrypted and stored in secureJsonData. Custom per app.
+    secureJsonData:
+      # kay/value pairs of string to string
+      key: value
+```
+
 ## Dashboards
 
 It's possible to manage dashboards in Grafana by adding one or more yaml config files in the [`provisioning/dashboards`]({{< relref "../installation/configuration.md" >}}) directory. Each config file can contain a list of `dashboards providers` that will load dashboards into Grafana from the local filesystem.
```
oddlittlebird, line 243:
```suggestion
You can manage dashboards in Grafana by adding one or more YAML config files in the [`provisioning/dashboards`]({{< relref "../installation/configuration.md" >}}) directory. Each config file can contain a list of `dashboards providers` that load dashboards into Grafana from the local filesystem.
```

pkg/api/admin_provisioning.go
marefr: No, it's not needed. Happening automatically, see https://github.com/grafana/grafana/blob/7b7000c5bcb445dcf039d50ab4feed68842ef056/pkg/api/common.go#L91 which uses message "Internal Server Error" and https://github.com/grafana/grafana/blob/7b7000c5bcb445dcf039d50ab4feed68842ef056/pkg/api/common.go#L49-L52 which log the error.

pkg/services/provisioning/apps/config_reader.go
marefr: We use error level for data source and notifiers so something to consider for later.

pkg/services/provisioning/apps/config_reader.go
marefr: Good point

conf/provisioning/apps/sample.yaml
marefr: After discussion with Carl we think it make sense to rename the package apps to plugins, but still keep the object "apps" in that yaml file. Point was that data sources are provisioned separately currently and is also subject for installation in the future. It wouldn't make sense to have separate provisioning files for installation and configuration and we maybe want to change this in the future.

pkg/api/admin_provisioning.go
aknuds1: Ah OK, a bit unexpected from the API.

aknuds1 (approved):
LGTM!

docs/sources/administration/provisioning.md
```diff
@@ -208,9 +208,39 @@ datasources:
       httpHeaderValue2: 'Bearer XXXXXXXXX'
 ```
 
+## Plugins
+
+> This feature is available from v7.1
+
+It's possible to manage plugins in Grafana by adding one or more YAML config files in the [`provisioning/plugins`]({{< relref "../installation/configuration/#provisioning" >}}) directory. Each config file can contain a list of `apps` that will be updated during start up. Grafana will update each app to match the configuration file.
```
oddlittlebird, line 215:
```suggestion
You can manage plugins in Grafana by adding one or more YAML config files in the [`provisioning/plugins`]({{< relref "../installation/configuration/#provisioning" >}}) directory. Each config file can contain a list of `apps` that will be updated during start up. Grafana updates each app to match the configuration file.
```

oddlittlebird (approved):
Docs look good, one minor suggestion for improvement.

lesovsky:
What about provisioning panel plugins? As I can see it doesn't supported
```
grafana_1          | t=2020-07-18T12:54:30+0000 lvl=eror msg="Server shutdown" logger=server reason="Service init failed: app provisioning error: app plugin not installed: grafana-polystat-panel"
```

FireEmerald:
@marefr The documentation of [plugins provisioning](https://grafana.com/docs/grafana/latest/administration/provisioning/#plugins) should be expanded. I'm confused about the naming. Is a "app plugin" the same as a "plugin" from https://grafana.com/grafana/plugins or do you mean a "plugin" of type "app" - this would mean the provisioning does not work for panels?

marefr:
@FireEmerald yes, "plugin" of type "app". Here's a list of official and community app plugins: https://grafana.com/grafana/plugins?type=app. 

Plugin provisioning does not install any plugins. You have to make sure to install any plugins before Grafana starts. Only provisioning of app plugins makes sense to support for now since they requires to be configured and enabled per organization in Grafana - panel plugins have no configuration and is always enabled and should be installed before Grafana is started.

forumsheth:
hey @marefr  I am trying to provision a app plugin but still getting the below error
eror msg="Server shutdown" logger=server reason="Service init failed: app provisioning error: app plugin not installed: devopsprodigy-kubegraf-app"

Can you suggest what needs to be done? I wan this app to be installed and enabled automatically  with grafana deployment

marefr:
Please read my earlier comment, e.g. 

> Provisioning doesn't install any plugins. Plugins should be installed before Grafana starts.

PR #25653, opened 2020-06-17
Dashboard: Redirects for old (pre 7.0) edit & view panel urls

### LINKED ISSUE #25459: Support old panelId urls for view & edit panel
We need to add redirects for old links that use panelId , to redirect viewPanel & editPanel 

In 7.0 we changed the view panel & edit panel urls from 

From
 https://play.grafana.org/d/000000012/grafana-play-home?panelId=2&orgId=1&edit&fullscreen

To
 https://play.grafana.org/d/000000012/grafana-play-home?editPanel=2&orgId=1 

For view panel (only difference is missing edit url param)

From
 https://play.grafana.org/d/000000012/grafana-play-home?panelId=2&orgId=1&fullscreen

To
 https://play.grafana.org/d/000000012/grafana-play-home?viewPanel=2&orgId=1

torkelo:
Closes #25459

### DIFF
--- a/pkg/middleware/dashboard_redirect_test.go
+++ b/pkg/middleware/dashboard_redirect_test.go
@@ -55,4 +55,21 @@ func TestMiddlewareDashboardRedirect(t *testing.T) {
 			})
 		})
 	})
+
+	Convey("Given the dashboard legacy edit panel middleware", t, func() {
+		bus.ClearBusHandlers()
+
+		middlewareScenario(t, "GET dashboard by legacy edit url", func(sc *scenarioContext) {
+			sc.m.Get("/d/:uid/:slug", RedirectFromLegacyPanelEditURL(), sc.defaultHandler)
+
+			sc.fakeReqWithParams("GET", "/d/asd/dash?orgId=1&panelId=12&fullscreen&edit", map[string]string{}).exec()
+
+			Convey("Should redirect to new dashboard edit url with a 301 Moved Permanently", func() {
+				So(sc.resp.Code, ShouldEqual, 301)
+				redirectURL, _ := sc.resp.Result().Location()
+				So(redirectURL.String(), ShouldEqual, "/d/asd/d/asd/dash?editPanel=12&orgId=1")
+			})
+		})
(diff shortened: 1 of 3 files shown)


pkg/middleware/dashboard_redirect_test.go
```diff
@@ -55,4 +55,21 @@ func TestMiddlewareDashboardRedirect(t *testing.T) {
 			})
 		})
 	})
+
+	Convey("Given the dashboard legacy edit panel middleware", t, func() {
```
aknuds1, line 59: We might want to rewrite this using the standard testing library.

marefr (approved):
LGTM

Any scenario where fullscreen param is not provided since this will only apply when fullscreen param is provided. Someone having old links maybe with only panelId param set?

torkelo:
no the old urls always had panelId & fullscreen, and then edit no edit depending on view/edit mode. which is why I simplified the state a big for 7.0 to just have editPanel  and viewPanel to cover both states with just two states instead of 3 properties

pkg/middleware/dashboard_redirect_test.go
torkelo: Then we have to convert this whole file, what test runners are you using?

aknuds1: @torkelo We just use `go test` for running tests. In this case, I think it would be easy to write this test without Convey. That's what I usually do when adding new tests, just write them with the standard library, in addition to the existing Convey ones (which I guess we'll clean up at some point).

torkelo: go test is terrible, have scroll a bunch to see what failed due to stack traces that are more than a screen long, and it does not rerun tests on file change (no watch), so the whole experience I find very poor. Like having a test runner in a tmux split that runs tests on file change and print failures in readable form. 

I tried to rewrite this to go testing library but sadly it depends on middlewareScenario which uses GoConvey so have to rewrite hundreds of tests to go testing library to convert this test :(

Issue #25547, opened 2020-06-11
Influxdb: Remaining Flux items for 7.1


kylebrandt:
List of Influx Changes:
 - Adds support for Flux
 - Data source configuration editor changes
   - Adds "Enable Flux" switch. When enabled, adds "Organization" and "Default Bucket" fields for use.
 - Query editor changes
   - Now has dropdown "Query" for the type of query. Values are "Classic", "InfluxQL", "Flux". Before there was no dropdown and there was only the equivalent of Classic and InfluxQL. What is now "Classic" was the default editor, and what is now "InfluxQL" used to be accessed by clicked the pencil edit icon. "Flux" is new.
  - When the Flux query editor is used:
    - Supported Macros: v.TimeRangeStart, v.TimeRangeStop, v.windowPeriod, v.bucket, v.defaultBucket, v.organization
    - Interpolated query (macros expanded) can be viewed in the Query Inspector

Outstanding TODO/Issues/Questions:

TODO:

 - [ ] [docs] Document above and add some query examples to documentation (WIP: #25551)
 - [x] [frontend] "Token" should only be displayed with "Enable Flux" and moved to the "Flux" section of the datasource configuration editor (PR #25662)
 - [ ] [frontend] (cleanup) Query properties unrelated to flux are still sent (and saved) from the frontend. I believe they are properties used by other types of flux queries
 - [ ] [frontend+backend] Test datasource should also test flux (when `enable flux` in datasource configuration) 
 - [ ] Need to make sure existing influx queries still work when upgrading from older than 7.1 
 - [ ] Healthcheck -- verify everything using the healthcheck endpoint

Questions:

 - [x] Should this be marked beta?
    * (rm) I think so, we can remove it if we think stable
 - [ ] Private flux plugin is getting more commits since merging flux to master. Not sure the intent around this and diverging?
    * (rm) The changes are from a branch I hoped would be merged before you migrated :( I have added the same changes in #25725, #25726.  Note that the flux plugin is not "private" it is just a normal plugin, that has been poorly supported for 2 years
 - [ ] Confused about the v.bucket macro, because I don't see how you would set it to something other than the defaultBucket?
    * (rm) correct -- eventually we want to actually send variables to flux rather than expand them.  the `v.variable` syntax is a flux construct that lets you send the values in a different part of the request.  However that is not yet implemented (on their side) so we will just expand them for now.

Bugs:

[...]

PR #25662, opened 2020-06-17
Influx: move token to 'enable flux' section within config editor

kylebrandt:
The token is only used for Flux as I understand it, so this move that config item.

Part of collection of misc flux items in #25547


ryantxu:
I think we should take a more dramatic approach like: #25723

This will make sure the same configs work for IFQL and Flux -- it will also force everything to use the backend support when configured

dprokop:
@kylebrandt @ryantxu I assume this is no longer relevant since #25723 is merged?

PR #25663, opened 2020-06-17
Datasources: add new tab and feature toggle

AgnesToulet:
Add a new tab and a feature toggle to enable a new Enterprise feature (Datasource Insights)

### DIFF
--- a/packages/grafana-data/src/types/config.ts
+++ b/packages/grafana-data/src/types/config.ts
@@ -38,6 +38,7 @@ export interface FeatureToggles {
    * Available only in Grafana Enterprise
    */
   meta: boolean;
+  datasourceInsights: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


alexanderzobnin (commented):
Is it OK that we have 404 on the insights page? Looks like there's no route in grafana to that pages. Or it will be implemented in future PR?

AgnesToulet:
Yes, I just created the other PR in the Enterprise repo.

alexanderzobnin:
Maybe it makes sense to show this only in Enterprise version to prevent confused 404 if feature toggled in open source?

AgnesToulet:
You shouldn't be able to see the Insights tab without an Enterprise license even if the feature is enabled.

alexanderzobnin (approved):

PR #25472, opened 2020-06-09
Backend plugins: Refactor to allow shared contract between core and external backend plugins

marefr:
**What this PR does / why we need it**:
Refactor to allow shared contract between core and external backend plugins allowing core backend data sources in Grafana to be implemented in same way as an external backend plugin.

This is currently based on #25454, but when that is merged I will rebase this one.
Also dependent on https://github.com/grafana/grafana-plugin-sdk-go/pull/187

**Special notes for your reviewer**:
- Verified image renderer v2.0.0 (v2 protocol) works
- Verified image renderer v1.0.12 (v1 protocol) works
- Verified test data source works (v2 protocol)
- Verified adding a temporary new core data source for manual test/verification when calling resource (see `testdatasourcev2` package)

PR #25665, opened 2020-06-17
Influx/flux: include interpolated query with error for query inspector

kylebrandt:
that includes a Frame with meta and the ExecutedQueryString.
part of #25547

This is a bit icky. I think we may say somewhere in error handling either return an error or a Frame (need to look into this), which doesn't work out well here. This is general area of some inconsistency we ended up with between the frontend and backend for data queries.

That being said it works. With this change you can now see the interpolated query even when there is an error executing that interpolated string.


kylebrandt:
@marefr in

https://github.com/grafana/grafana/pull/25472#discussion_r438356582

> "After second thought I don't think we need to care about this. Looking at wrapper v2 doesn't do that"

Good call based on this - wouldn't have been able to pass the ExecutedQueryString.

ivanahuckova (approved):
LGTM

PR #25676, opened 2020-06-17
StatPanel: Option showing name instead of value and more

### LINKED ISSUE #24076: StatPanel: Migration from singlestat does not support the metric name option
In the old singlestat we had a "name" option in the stat option dropdown. 

And was handled using a custom case: 
https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/singlestat/module.ts#L212

There is no equivalent option in the new stat panel.

### LINKED ISSUE #24486: Stat Panel: Tooltip for title and value (if value is hidden)
Tooltip showing measurement name for stat panel

**Why is this needed**:
I use the stat panel with background colors to show status of sensors. I set the size of the panel to be quite small, so I can see the color/status but the text is too small to be legible. It would be great if a tooltip could be set so I could simply hover over the box to see the measurement name if necessary.

I attempted doing something similar using Data Links, but it didn't seem capable of pulling data from the query.

torkelo:
Just a first stab / idea for how to solve the feature to show the name instead of the value in the singlestat panel 

Closes #24076
Closes #24486

This is something the old singlestat could do. 

I propose adding a "textMode" option (struggled a lot naming this) that controls what the stat panel will show. 

* Auto (like it is today, only show name and value if there are more than one series/fields) 
* Value (only show value, show name in the tooltip if there is any, good for big grids with lots of values)
* Name (show only name instead of value, show value in tooltip)
* Value and name force showing both value and name (for single series / fields) scenarios. Makes it easier to handle this scenario. Right now users need to specify a custom display name template to get this if there is only one series/field.
* None (show nothing, name and value in the tooltip), good for lots of small cells & overview dashboards

So why is Name option is useful, well think of it like the value is not interesting (maybe it's 0 and 1) and that is visualized via the color already, so the only interesting things are the color and the name. 

So this would involve adding a tooltip as well. 

* [x] The logic of clearing title if there is only one DisplayValue have to be removed from getFieldDisplayValues so that we do not loose this info (so we can show it in tooltip or force showing it if option is set to always show name and value). 
* [x] Need to move the logic of hiding title for single series results to BarGauge and Gauge (or apply this new textMode option to those visualizations as well)
* [x] Add tooltip logic
* [x] Write unit tests
* [x] Update devenv dashboards 
* [x] Update docs
* [x] Update singlestat migration button 

Screenshot from 2020-06-17 18-38-39


packages/grafana-ui/src/components/BigValue/BigValue.tsx
```diff
@@ -32,6 +32,17 @@ export enum BigValueJustifyMode {
   Center = 'center',
 }
 
+/**
+ * Options for how the value & title are to be displayed
+ */
+export enum BigValueNameAndValueOption {
```
ryantxu, line 38:
```suggestion
export enum BigValueTextOption {
```

packages/grafana-ui/src/components/BigValue/BigValue.tsx
```diff
@@ -32,6 +32,17 @@ export enum BigValueJustifyMode {
...
+  Auto = 'auto',
+  Value = 'value',
+  ValueAndName = 'value_and_value',
```
ryantxu, line 41:
```suggestion
  ValueAndName = 'value_and_name',
```

packages/grafana-ui/src/components/BigValue/BigValue.tsx
```diff
@@ -43,6 +54,7 @@ export interface Props extends Themeable {
   graphMode: BigValueGraphMode;
   justifyMode?: BigValueJustifyMode;
   alignmentFactors?: DisplayValueAlignmentFactors;
+  nameAndValue?: BigValueNameAndValueOption;
```
ryantxu, line 57:
```suggestion
  showText?: BigValueTextOption;
```

public/app/plugins/panel/stat/module.tsx
```diff
@@ -9,6 +9,21 @@ export const plugin = new PanelPlugin<StatPanelOptions>(StatPanel)
   .setPanelOptions(builder => {
     addStandardDataReduceOptions(builder);
 
+    builder.addSelect({
+      path: 'nameAndValue',
```
ryantxu, line 13: I don't really like `nameAndValue` as a property -- maybe `showText` or `textMode`?

ryantxu (commented):
I like the behavior -- the names need work ;)  happy to discuss more

public/app/plugins/panel/stat/module.tsx
torkelo: changed to textMode

packages/grafana-ui/src/components/BigValue/BigValue.tsx
torkelo: changed to textMode and BigValueTextMode

docs/sources/panels/visualizations/stat-panel.md
```diff
@@ -48,3 +48,19 @@ Use the following options to refine your visualization:
...
+  - **Value -** Show only value never name. Name will be shown in hover tooltip instead.
+  - **Value and name -** Always show value and name.
+  - **Name -** Show name in place of value. Will show value in a hover tooltip.
+  - **None -** Show nothing (empty). Will show both name and value in hover tooltip.
+
+## Text mode
+
+> Only available in Grafana 7.1+
+
+By default the stat will just value for a single series or field and both value and name for multiple series or fields. You can use the Text mode option to control what text the panel is to render. If the value is not that important, only name & color is, change the `Text mode` to **Name**. The value will still be used to determine color and is available via a tooltip.
```
oddlittlebird, line 62:
```suggestion
By default, the Stat panel displays: 
- Just the value for a single series or field.
- Both the value and name for multiple series or fields.

You can use the Text mode option to control what text the panel renders. If the value is not important, only name and color is, then change the `Text mode` to **Name**. The value will still be used to determine color and is displayed om a tooltip.
```

oddlittlebird (commented):
Minor wording adjustments.

public/app/plugins/panel/bargauge/BarGaugePanel.tsx
```diff
@@ -18,10 +18,15 @@ export class BarGaugePanel extends PureComponent<PanelProps<BarGaugeOptions>> {
     menuProps: DataLinksContextMenuApi
   ): JSX.Element => {
     const { options } = this.props;
-    const { value, alignmentFactors, orientation, width, height } = valueProps;
+    const { value, alignmentFactors, orientation, width, height, count } = valueProps;
     const { field, display, view, colIndex } = value;
     const { openMenu, targetClassName } = menuProps;
 
+    // hide title if only 1 gauge is being displayed and no custom displayName set
+    if (count === 1 && !field.displayName) {
```
mckn, line 26: nit: would it be possible to move this code somewhere so it can be shared between the `BarGaugePanel` and the `GaugePanel`?

mckn (approved):
Looks good to me! Work really well when testing it 🙌 

Added one super small nit that might be worth looking at.

mckn:
One question regarding the tooltip. 

I see that we have a logic that displays different values depending on the text mode. I understand the reason for this but I am thinking that it might be better to always display both name and value in the tooltip and with some formatting e.g. 

**name:** frontend-server
**value:** 30%

public/app/plugins/panel/stat/types.ts
```diff
@@ -65,7 +56,7 @@ export function addStandardDataReduceOptions(
   builder.addCustomEditor({
     id: 'reduceOptions.calcs',
     path: 'reduceOptions.calcs',
-    name: 'Value',
+    name: 'Calculation',
```
dprokop, line 59: nice!

public/app/plugins/panel/gauge/GaugePanel.tsx
```diff
@@ -12,10 +12,15 @@ export class GaugePanel extends PureComponent<PanelProps<GaugeOptions>> {
...
     const { options } = this.props;
-    const { value, width, height } = valueProps;
+    const { width, height, count, value } = valueProps;
     const { field, display } = value;
     const { openMenu, targetClassName } = menuProps;
 
+    // hide title if only 1 gauge is being displayed and no custom displayName set
+    if (count === 1 && !field.displayName) {
+      display.title = undefined;
+    }
```
dprokop, line 22: Nit: not sure why you want to hide title in such scenario? Maybe instead we could have an option to hide the title explicitly?

This also causes some vertical misalignment in the panel:
image

dprokop (approved):
Works great, really love this addition! Left some small nits that can be addressed separately :)

dprokop:
> I see that we have a logic that displays different values depending on the text mode. I understand the reason for this but I am thinking that it might be better to always display both name and value in the tooltip and with some formatting e.g.
name: frontend-server
value: 30%

Yes, I resonate with this

public/app/plugins/panel/gauge/GaugePanel.tsx
torkelo: this was just replicating logic that was before in getFieldDisplayValues

torkelo: strange, not seeing that missalignment, the logic in Gauge is to check for null or undefined, so it does not use any space if title is undefined

torkelo:
pushed changes, moved the single series no name logic to an imutable function that is shared between Gauge and BarGauge. 

Also updated tooltip, but I tested your suggestion always show Name and Value and that made it much harder to parse the text for the cases where it only contains name or value. So that did not work. But when both value and name is shown the prefixes and line breaked helped.

public/app/plugins/panel/bargauge/BarGaugePanel.tsx
torkelo: moved to shared function

PR #25677, opened 2020-06-17
@grafana/e2e: improvements

stevenvachon:
See commit message(s).

### DIFF
--- a/packages/grafana-e2e/src/flows/addPanel.ts
+++ b/packages/grafana-e2e/src/flows/addPanel.ts
@@ -51,6 +51,9 @@ export const addPanel = (config?: Partial<AddPanelConfig>): any =>
         .scrollIntoView();
     });
 
+    // @todo instead wait for '@pluginModule'
+    e2e().wait(2000);
+
     openOptions();
 
     openOptionsGroup('settings');
--- a/packages/grafana-e2e/src/flows/addPanel.ts
+++ b/packages/grafana-e2e/src/flows/addPanel.ts
@@ -87,6 +87,9 @@ export const addPanel = (config?: Partial<AddPanelConfig>): any =>
 
     e2e().wait('@chartData');
 
+    // Wait for RxJS
+    e2e().wait(500);
+
     // @todo remove `wrap` when possible
     return e2e().wrap({ config: fullConfig });
   });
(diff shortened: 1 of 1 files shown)


briangann (approved):
LGTM!

packages/grafana-e2e/src/flows/addPanel.ts
```diff
@@ -77,6 +85,11 @@ export const addPanel = (config?: Partial<AddPanelConfig>): any =>
       .get('button[title="Apply changes and go back to dashboard"]')
       .click();
 
+    e2e().wait('@chartData');
+
+    // Wait for RxJS
+    e2e().wait(500);
```
scottlepp, line 91: seems brittle

scottlepp (changes requested):
the wait for rxjs bit seems a bit brittle.  can we log a message or something so when that times out we know why?  otherwise the test will fail and we won't know the reason

packages/grafana-e2e/src/flows/addPanel.ts
stevenvachon: I don't understand. It doesn't kill it after 500ms, it simply waits. I don't know why this happens, but Cypress `screenshot()` causes RxJS to fail. What would logging a message accomplish?

scottlepp: what happens if rxjs is not loaded?

scottlepp: is there a way to determine what you are waiting for has loaded?

stevenvachon: Not that I know of. RxJS isn't a global.

scottlepp: i'm ignorant here with cypress.  can you execute something that would test the existence of rxjs?

scottlepp (approved):

packages/grafana-e2e/src/flows/addPanel.ts
hugohaggmark: This sounds very disturbing, I'd like to look into this closer but will not have time today. RxJs is our library for streaming data from the backendSrv to visualisations so we really don't know when it's completed.

stevenvachon: "Streaming" doesn't sound like a normal request, so perhaps it can be covered in the near future with https://github.com/cypress-io/cypress/issues/687

stevenvachon: > i'm ignorant here with cypress. can you execute something that would test the existence of rxjs?

Not that I can think of. We can `wait()` for requested URLs (currently only `XMLHttpRequest` and not `fetch()`).

PR #24648, opened 2020-05-13
CloudWatch: Unify look of query mode select between dashboard and explore

aocenas:
Use the same component and make the options more descriptive:

Screenshot from 2020-05-13 20-04-19

PR #25691, opened 2020-06-18
Datasource/CloudWatch: More robust handling of different query modes

kaydelaney:
A small refactor which changes how the CloudWatch datasource handles the case of multiple queries with different query modes. Rather than simply checking the query mode of the first target and assuming the remaining queries are of the same mode, this PR introduces code that separates the supplied queries based on mode.
This PR is mostly useful as groundwork for future Logs/Metrics unification work.

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/components/PanelQueryEditor.tsx
+++ b/public/app/plugins/datasource/cloudwatch/components/PanelQueryEditor.tsx
@@ -1,7 +1,8 @@
 import React, { PureComponent } from 'react';
+import pick from 'lodash/pick';
 import { ExploreQueryFieldProps, ExploreMode } from '@grafana/data';
 import { Segment } from '@grafana/ui';
-import { CloudWatchQuery, CloudWatchMetricsQuery, CloudWatchLogsQuery } from '../types';
+import { CloudWatchQuery } from '../types';
 import { CloudWatchDatasource } from '../datasource';
 import { QueryInlineField } from './';
 import { MetricsQueryEditor } from './MetricsQueryEditor';
@@ -17,7 +18,7 @@ const apiModes = {
 export class PanelQueryEditor extends PureComponent<Props> {
   render() {
     const { query } = this.props;
-    const apiMode = query.apiMode ?? query.queryMode ?? 'Metrics';
+    const apiMode = query.queryMode ?? 'Metrics';
 
     return (
       <>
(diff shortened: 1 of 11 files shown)


aocenas:
Does not seem to be working for me, it sends logs query in metrics mode:
Screenshot from 2020-06-29 13-01-21

aocenas:
seems like there is some issue with the mode switching, even though I switch to metrics, the queryMode state stays in logs.

kaydelaney:
Hm, strange, I'll check it out

aocenas:
Seems like we have both apiMode and queryMode in the query but we are only using apiMode now: https://github.com/grafana/grafana/blob/3b383149db7408b1669159ad2edad9c141498ece/public/app/plugins/datasource/cloudwatch/components/PanelQueryEditor.tsx#L29

Probably something for backward compatibility or something.

aocenas:
This is the change https://github.com/grafana/grafana/pull/24648#discussion_r424631956

public/app/plugins/datasource/cloudwatch/components/PanelQueryEditor.tsx
```diff
@@ -25,9 +25,22 @@ export class PanelQueryEditor extends PureComponent<Props> {
...
-            onChange={({ value }) =>
-              this.props.onChange({ ...query, apiMode: (value as 'Metrics' | 'Logs') ?? 'Metrics' })
-            }
+            onChange={({ value }) => {
+              const newMode = (value as 'Metrics' | 'Logs') ?? 'Metrics';
+              if (newMode === 'Metrics') {
+                this.props.onChange({
+                  ...(query as CloudWatchMetricsQuery),
+                  apiMode: newMode,
+                  queryMode: newMode,
```
aocenas, line 34: I think this may not be correct. In explore we have more combinations. Ie we can have apiMode = logs but queryMode = metrics. So this should probably not be in sync but as 2 separate things. One pushed from query editor second from explore.

kaydelaney: `queryMode` is just supposed to exist in service of [discriminated unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions), though in retrospect it may not be the best name.
In general I think I need to take a look at `CloudWatchMetricsQuery` and `CloudWatchLogsQuery` as I think the way they're handled right now isn't great and a bit confusing

aocenas: Ah ok right we actually do not care about the explore mode in the datasource. In that case I assume we may actually remove the queryMode? I think it was present only in Explore (and I introduced apiMode by copying code from dashboards editor) so I think it should be safe to remove.

public/app/plugins/datasource/cloudwatch/components/PanelQueryEditor.tsx
```diff
@@ -17,17 +18,35 @@ const apiModes = {
...
+                const commonProps = pick(
+                  query,
+                  'id',
+                  'region',
+                  'namespace',
+                  'refId',
+                  'hide',
+                  'key',
+                  'queryType',
+                  'datasource'
```
aocenas, line 41: Is datasource part of the query?

aocenas:
There seems to be some issues still.

When trying the cheat sheet clicking on the example query I get to state where I have logs query in metrics mode:
Screenshot from 2020-07-08 23-46-18

which does not happen in master.

Second issue seems to be that there is some leaking of values into the query when query type is metrics:
Screenshot from 2020-07-08 23-48-36

Things like `liveStreaming` or `showingTable` are internal explore variables so no idea how they got into the query.

kaydelaney:
Cheat sheet issue is fixed now, but re: that second issue:
> Second issue seems to be that there is some leaking of values into the query when query type is metrics:
Agree that this is strange and not ideal, but it also seems to be the behavior on master. Might be best to make a separate issue?

aocenas:
> Agree that this is strange and not ideal, but it also seems to be the behavior on master. Might be best to make a separate issue?

Ah ok, I thought I did not see it in master but did not dig much deeper there. In that case can you create a separate issue for this so we can take a look later?

aocenas (approved):

PR #16315, opened 2019-03-31
Add support for reading data source provisioner secrets from files

bobmshannon:
Closes #12368, which also includes background details on the use cases surrounding this feature.

PR #25692, opened 2020-06-18
Provisioning: Apply settings expanders

### LINKED ISSUE #12368: Support reading config value from file when provisioning datasources
By allowing the provisioning system to read config values from disk as an alternative to regular strings, we enable orchastrators such as k8s, mesos and open shift to provide the token value as a file. This way the token management/lifecycle will be completly managed to the orchastrator without changing the datasource.yaml file. 

Ex in kubernetes case
```
<brancz> concretely the service account token is located in /var/run/kubernetes.io/serviceaccount/token
```

cc @brancz

sakjur:
Adds support for the variable expanders introduced to the configuration file in #25075 to provisioning.

Fixes #


pkg/services/provisioning/values/values.go
```diff
@@ -171,7 +177,9 @@ func transformInterface(i interface{}) (interface{}, interface{}) {
 	case reflect.Map:
 		return transformMap(i.(map[interface{}]interface{}))
 	case reflect.String:
-		return interpolateValue(i.(string))
+		// TODO: Handle error
+		val, raw, _ := interpolateValue(i.(string))
+		return val, raw
```
sakjur, line 182: I'm not sure what we'd want to do here

pkg/services/provisioning/values/values_test.go
```diff
@@ -241,6 +246,30 @@ func TestValues(t *testing.T) {
...
+	type Data struct {
+		Val StringValue `yaml:"val"`
+	}
+
+	f, err := ioutil.TempFile(os.TempDir(), "file expansion *")
+	require.NoError(t, err)
+	file := f.Name()
+
+	defer func() {
+		require.NoError(t, os.Remove(file))
```
aknuds1, line 259: I think it's safer to ensure the file is closed before you try to remove it. Depending on the platform, removing an open file could fail:
```suggestion
		require.NoError(t, f.Close())
		require.NoError(t, os.Remove(file))
```

pkg/services/provisioning/values/values_test.go
```diff
@@ -241,6 +246,30 @@ func TestValues(t *testing.T) {
...
+	}()
+
+	const expected = "hello, world"
+	_, err = f.WriteString(expected)
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	data := &Data{}
+	err = yaml.Unmarshal([]byte(fmt.Sprintf("val: $__file{%s}", file)), data)
+	assert.NoError(t, err)
```
aknuds1, line 269: I think the test should halt immediately and not continue with assertions after an error.
```suggestion
	require.NoError(t, err)
```

marefr (changes requested):
Great. Any issue to reference maybe? Documentation needs update.

Hadn't seen the referenced PR. Guess it's fine, but seems a bit non-optimal that you for example have to spread out usage of vault in several places but I guess you have reasons for it. As an end user if I have configured vault I would expect Grafana to always try and use that automatically where needed, but maybe not that easy.

aknuds1 (changes requested):
Generally looks good, but left some suggestions (open to discussion).

sakjur:
> Great. Any issue to reference maybe? Documentation needs update.

Only a request in an internal roadmap doc for supporting provisioning in addition to config files.

> Hadn't seen the referenced PR. Guess it's fine, but seems a bit non-optimal that you for example have to spread out usage of vault in several places but I guess you have reasons for it. As an end user if I have configured vault I would expect Grafana to always try and use that automatically where needed, but maybe not that easy.

It could be, but that would require us to define the structure we expect of the users' Vault setups. If an organization has a rigid structure from Vault, making sure Grafana is flexible makes it easier for them. In addition to that supporting reading options from files is also almost as important as Vault, as that's what the Kubernetes community is doing for secrets (although we're already supporting that but with a hack in the Docker image).

I'd like to add opinionated parsing of Vault configuration eventually. We have the foundations for that now that we have started with an unopinionated model first.

pkg/services/provisioning/values/values_test.go
sakjur: Reverted this, I already close the file immediately after writing to it so I got an error for trying to close a file twice

marefr (approved):
LGTM when fixed lint error. Think you can reference #12368, but it doesn't close that because some users
 requested dynamic read from files whenever a data source is used where's this will only be read once/on provision update

aknuds1 (approved):
LGTM!

sakjur:
> Think you can reference #12368, but it doesn't close that because some users
requested dynamic read from files whenever a data source is used where's this will only be read once/on provision update

@marefr ✅ I don't see the ask for dynamic read in that thread, am I missing something?

marefr:
@sakjur sorry, was a https://github.com/grafana/grafana/pull/16315#issuecomment-482157743 in one of the PR's that tried to solve the feature request

bergquist:
Great!

This and https://github.com/grafana/grafana/pull/25075/files#diff-9ba8047fed34dc348a4be93767dae68aR22 seems to lack documentation. @sakjur I guess you will do that later? :)

PR #25693, opened 2020-06-18
Grafana UI: Apply disabled styles to LinkButton

Clarity-89:

Currently there is no way to disable `LinkButton` since it's using `a` tag, for which `disabled` attribute doesn't work. In order to actually disable `LinkButton` we need to manually apply disabled styles and remove pointer events + tabindex to make it un-clickable.

### DIFF
--- a/packages/grafana-ui/src/components/Button/Button.tsx
+++ b/packages/grafana-ui/src/components/Button/Button.tsx
@@ -182,7 +182,7 @@ export const LinkButton = React.forwardRef<HTMLAnchorElement, ButtonLinkProps>(
 
     return (
       <a
-        className={cx(styles.button, className, linkButtonStyles)}
+        className={cx(styles.button, linkButtonStyles, className)}
         {...otherProps}
         ref={ref}
         tabIndex={disabled ? -1 : 0}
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/Button/Button.tsx
```diff
@@ -167,8 +171,22 @@ export const LinkButton = React.forwardRef<HTMLAnchorElement, ButtonLinkProps>(
...
+        disabledStyles,
+        css`
+          pointer-events: none;
+        `
+      );
+
     return (
-      <a className={cx(styles.button, className)} {...otherProps} ref={ref}>
+      <a
+        className={cx(styles.button, className, linkButtonStyles)}
```
tskarhed, line 185: Correct me if I'm wrong, but if you put the className last it can overwrite previous styles? Maybe that's a desirable effect.
```suggestion
        className={cx(styles.button, linkButtonStyles, className)}
```

tskarhed (changes requested):
Add this as a knob in the Button story, then we're all good to go!

packages/grafana-ui/src/components/Button/Button.tsx
Clarity-89: I was thinking the same, but then decided we'd like the disabled styles to be enforced so they're not accidentally overridden, but now not sure about that 🤔

tskarhed (approved):

PR #25694, opened 2020-06-18
Explore: Use merge transform instead of DataFrame -> toLegacyTable -> Merge -> toDataFrame

torkelo:
In 7.1 we recently added a merge transform that behaves like the old table merge function so now we do not need this costly data juggling :)

### DIFF
--- a/public/app/features/explore/utils/ResultProcessor.ts
+++ b/public/app/features/explore/utils/ResultProcessor.ts
@@ -4,13 +4,12 @@ import {
   DataFrame,
   FieldType,
   TimeZone,
-  toDataFrame,
   getDisplayProcessor,
   ExploreMode,
   PreferredVisualisationType,
+  standardTransformers,
 } from '@grafana/data';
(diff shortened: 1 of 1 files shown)


aocenas:
@torkelo not sure I know how to properly test this but it seems there are some differences now:

In master
Screenshot from 2020-06-22 15-02-03

This PR
Screenshot from 2020-06-22 15-02-39

Not sure this is a meaningful test case, basically this is Loki doing an instant query and then joining both (timeseries data + instant query) x 2 queries which is probably a bug but seems like there are slight differences in the merging algo which aren't tested.

torkelo:
The first table (master) also looks bugged with dates in the value column. 

The way to test is with multiple Prometheus queries

aocenas:
> The first table (master) also looks bugged with dates in the value column.

> The way to test is with multiple Prometheus queries

Yup could be some other issues there. With prometheus the result looks good.

aocenas (approved):

PR #25698, opened 2020-06-18
Templating: fixes variables not being interpolated after dashboard refresh

### LINKED ISSUE #23841: Global variable loses value upon dashboard refresh
**What happened**:
Global variable loses its value after refreshing the dashboard.

After refreshing the dashboard, the values should also get passed.

**Before:**
Screen Shot 2020-04-23 at 5 11 24 PM

**After refreshing the dashboard:**
Screen Shot 2020-04-23 at 5 11 31 PM

1. Disable HTML sanitization
2. Create a Dashboard > Choose text as visualization
3. Select HTML and enter "<span style>${__dashboard.name}</span>"
4. Give the dashboard a name so the value gets pulled into the ${__dashboard.name}
5. Refresh the dashboard (upper right of the screen) see that it doesn't pull the value itself and reverts back to "${__dashboard.name}"

Issue is specific to global variables. I tested this out on the dashboard variables and they are working as intended.

### LINKED ISSUE #25103: DataLinks: ${__dashboard} variable does not interpolate correctly after query refresh
**What happened**:
Created a data link as:
```d/${__dashboard}/${__dashboard.name}?var-host=${__value.text}```
The link renders as:
```https://hostname/grafana/d/$%7B__dashboard%7D/$%7B__dashboard.name%7D?var-host=host2```

The dashboard variables should be expanded in the rendered link to produce a URL similar to:
```https://hostname/grafana/d/AHwBKXnZk/interface-errors?var-host=host2```

Display some data in a table panel.
Add an override with a data-link containing `${__dashboard}`

Tangentially related to: #22380

**Environment**:
- Grafana version: v7.0.0
- Panel: table

### LINKED ISSUE #25359: global variables e.g. user.login stop expanding when change time range
**What happened**:
The global variable stopped expanding when adjusting the time range.

The variable continued to expand to its value in normal usage.

This occured in both Safari and Chrome on Mac OS and Chrome on iOS
It happened when the time range was changed either with the drop down or with the drag zoom control

On Mac OS, it almost always corrects when you log out and then log back in again
On iOS Chrome, it fixed also when selecting another dashboard and then reselecting the dashboard that includes the global var.

**Environment**:
- Grafana version: Grafana v7.1.0-db3b38efpre (467348bded)
- Data source type & version: InfluxDB v1.7.0
- OS Grafana is installed on: linux in container :master tag
- User OS & Browser: Mac Safari/Chrome iOS Chrome
- Grafana plugins: GF_INSTALL_PLUGINS=marcuscalidus-svg-panel,blackmirror1-singlestat-math-panel,briangann-gauge-panel,jdbranham-diagram-panel,natel-plotly-panel"
- Others:

### LINKED ISSUE #25757: ${__user.login} template not loaded properly on variable Init
**What happened**:
I upgraded from Grafa 6.7 to 7.1 And created a  variable to retrieve some value based on logged user. But the value only getting an update when saving the query, when the dashboard loading after login the value not found either its don't have previous value.

The Initiated variable should set on the Dashboard on Load

Query : 
`select security_group_institutions.institution_id from security_users inner join security_group_users on security_group_users.security_user_id = security_users.id inner join security_group_institutions on security_group_institutions.security_group_id = security_group_users.security_group_id where  security_users.username = '${__user.login}'  group by security_group_institutions.institution_id limit 1`
Screenshot from 2020-06-23 09-25-02

**Environment**:
- Grafana version: 7.1
- Data source type & version: Mysql 5.7
- User OS & Browser: Ubuntu & Chrome
- Grafana plugins: Default Plugins 
- Others:

hugohaggmark:
This PR deprecates `setGlobalVariable` in TemplateSrv and moves this to Redux with a the MetaVariableModel. This was a bit rushed but the issues it fixes have been open some time so it would be great to close them. The question is: is this a good first step to make TemplateSrv more aligned with the new variable system.

Fixes #23841
Fixes #25359
Fixes #25103
Fixes #25757

I will not be online when you decide if this is a suitable solution or not. Trash this if it seems like a dumb way forward :)


torkelo (commented):
I think this looks great and definitely the right direction! My only feeling is that "meta" is not the right type name, maybe "system" or "internal" type name is better? Global would also work but would be confusing when we introduce global (shared variables across dashboards)

hugohaggmark:
> I think this looks great and definitely the right direction! My only feeling is that "meta" is not the right type name, maybe "system" or "internal" type name is better? Global would also work but would be confusing when we introduce global (shared variables across dashboards)

I agree, my first suggestion was `dynamic` but I like `system`.

torkelo:
@mckn can you take over this PR (when you're done with the timezone). 

Just a minor rename needed of the variable type from meta to `system` and any other review tweaks you find.

mckn:
> @mckn can you take over this PR (when you're done with the timezone).
> 
> Just a minor rename needed of the variable type from meta to `system` and any other review tweaks you find.

yes, I will fix it :)

mohamednizar:
@hugohaggmark will this fix this issue? https://github.com/grafana/grafana/issues/25757

torkelo (approved):
Looks great!

allomorphy:
Excited to beta test this as soon as it is in master. 🤩

mckn (approved):
LGTM

mohamednizar:
@dprokop in which pre-release this fix available?

allomorphy:
Hi @dprokop,
Amazing 🙏 

Surely @mohamednizar and I could run by tomorrow the code committed per https://github.com/grafana/grafana/commit/6e3e9b7c395d1dd0772bbf0455bc427c7453019b

With something like this:
`docker run -d -p 3000:3000 --name grafana grafana/grafana:master` 
🤔

salanki:
Please note that this does not work as intended in 7.1.0-beta1, see: https://github.com/grafana/grafana/issues/25757

PR #25702, opened 2020-06-18
Units: Match 'length' case with other categories

der-eismann:
**What this PR does / why we need it**: This PR is changing the `length` category to `Length`, because it was the only unit category starting with a lowercase letter.

Just a tiny PR, but this really bugged me.


peterholmberg (approved):
Haven't noticed until now but :exploding_head:

PR #25251, opened 2020-06-01
Panel Inspect: use Monaco editor for json display

ryantxu:
Alternative to #25199

Rather than use the helper library: https://github.com/suren-atoyan/monaco-react, this tries to manage monaco more directly.

TODO: 
- [x] Verify CodeEditor works when imported from @grafana/ui in plugins. Not sure rollup will bundle everything as expected. (Fixed by dcce054eb36f750db8e53f86e90c3e929c0d5a79)
- [ ] Add tests for CodeEditor API (onChange, onSave etc)

PR #25723, opened 2020-06-19
InfluxDB: different config UI for 1x vs 2x

ryantxu:
This changes the "enableFlux" flag to a dropdown version for 1x and 2x

This also changes:
* different UI for 1x vs 2x
* only bakend support to 2x (removes all the HTTP configs since they are not used in the client)
* Moves the flux query editor to react.  When #25251 is merged, we can use monaco editor
* adds some sample queries for flux
* adds a beta notice (we can remove when we have more confidence later)

Alternative to #25662 (see also #25547)

image


pkg/tsdb/influxdb/influxdb.go
```diff
@@ -92,7 +92,13 @@ func (e *InfluxDBExecutor) Query(ctx context.Context, dsInfo *models.DataSource,
 		return nil, err
 	}
 
-	httpClient, err := dsInfo.GetHttpClient()
+	var httpClient *http.Client
+	version := dsInfo.JsonData.Get("version").MustString("1x")
+	if version == "2x" {
+		glog.Info("TODO use org & token to create a new client")
```
ryantxu, line 98: @kylebrandt -- I have asked them to confirm what the proper request structure is.  My understanding was basic auth with user=org and token=password, but that does not seem to work

kylebrandt:
I'm a bit confused. Is the idea that the datasource can be configured for 1.x _or_ 2.x, or is just that the drop down filters areas of configuration?

If it is the latter, than a tab or some other UI element might make sense?

ryantxu:
The server you are talking to is *either* 1.x or 2.x -- and when 2.x we only use token/org -- when 1.x it has all the HTTP settings and configures a database/password that only applies in 1x

kylebrandt (approved):

PR #25726, opened 2020-06-19
InfluxDB: support aggregate results

ryantxu:
Depends on https://github.com/grafana/grafana/pull/25725

The timeseries detection is off and should not depend on a `_measurment` field existing.

Lets wait till #25725 is merged to look a this


pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -152,6 +154,43 @@ func TestExecuteGrouping(t *testing.T) {
...
+			testDataPath: "aggregate.csv",
+		}
+
+		dr := ExecuteQuery(ctx, QueryModel{MaxDataPoints: 100}, runner, 50)
+
+		if dr.Error != nil {
+			t.Fatal(dr.Error)
+		}
+
+		str, _ := dr.Frames[0].StringTable(-1, -1)
```
ryantxu, line 171: @kylebrandt I know this breaks your brain -- but look at how pretty it is ;)

kylebrandt: Not when the test changes/fails. Then the diff will be a mess - so this is bad for maintainable code.

pkg/tsdb/influxdb/flux/builder.go
```diff
@@ -129,20 +128,29 @@ func (fb *FrameBuilder) Append(record *query.FluxRecord) error {
 		}
 
 		if fb.isTimeSeries {
+			frameName := ""
+			name := record.ValueByKey("_measurement")
+			if name != nil {
+				frameName = name.(string)
```
kylebrandt, line 134: Since this is a core datasource, I think it is better to be a defensive on any type assertion.s therefore this should use the same `ok` pattern as earlier.

pkg/tsdb/influxdb/flux/builder.go
```diff
@@ -129,20 +128,29 @@ func (fb *FrameBuilder) Append(record *query.FluxRecord) error {
...
+			}
+			name = record.ValueByKey("_field")
+
 			// Series Data
 			labels := make(map[string]string)
 			for _, name := range fb.labels {
 				labels[name] = record.ValueByKey(name).(string)
```
kylebrandt, line 141: Another assertion that should likely be checked.

kylebrandt (changes requested):
Some assertions that should be checked to be safe, and as per documentation (https://pkg.go.dev/github.com/grafana/grafana-plugin-sdk-go/data?tab=doc#Frame.StringTable) do not use the output of StringTable for comparing in tests.

kylebrandt (approved):

PR #25728, opened 2020-06-20
POC: support hidden fields

### LINKED ISSUE #24092: Table: Ability to hide fields but still reference them in data links

I would like the ability to hide fields from the table but still reference them in data links.

**Why is this needed**:

First of all, I really like the new Table component. Thanks for the great work!

While migrating some dashboards from the old table to the new one I noticed that the new table component lacks one feature: When adding a data link via a field override it does not allow referencing filtered fields, for example `http://example.com?foo=${__data.fields.foo}` where `foo` was hidden by the "Filter by name" or "Organize fields" transformation.

I would like the ability to hide fields from the table but still reference them in data links, for example by adding a field override property "Hide" or by having the ability to reference filtered fields.

**Environment:**

  * Grafana version: 7.0.0-beta1

ryantxu:
Super early POC to support hidden fields.

Currently when we filter fields, they are totally removed from the results -- this means you can not use them for template variables or further calculations.

this is not ready for review... but posting so I don't lose it :)

Fixes #24092


torkelo (commented):
For things like this a class wrapper for DataFrame would have been killer to have a special iterator function for including or excluding hidden fields :)

Because now we need to update all loops for fields to check hidden flag. So in the absence of a class, should we add a utility function ether like dataFrameForEachField(frame, true/false, callback) to a getFields(frame, true/false) .

torkelo:
an idea I have is to move hidden fields to a sperate array on the DataFrame, I think the need to access hidden fields should be quite minimal (data links, and coming dynamic expressions) and should be easy to add checking the hidden fields there. 

That feels like a better solution as if we just have a hidden flag we need to check that flag in so many locations, especially as we do not have a DataFrame class with a custom field iterator

ryantxu:
I'll close this for now -- likely a different approach will make more sense

DanielGnzlzVll:
👎

simPod:
@ryantxu can you crosslink the different approach PR when available? Thanks!

simPod:
https://github.com/grafana/grafana/pull/41443

PR #25733, opened 2020-06-22
Users: Changed hardcoded limit from 10 to 50 when getting users to add them into a team

### LINKED ISSUE #25552: The UserPicker component has a hardcoded limit of 10.
I'm trying to add Users to a Team.  I followed all the instructions, but when I see the dropdown menu with the list of users, it only ever includes the same 10 users.  

I tracked this down to line 47 of UserPicker.tsx:
https://github.com/grafana/grafana/blob/1e88e508220159cddb846e5934af3fe3ae39bef9/public/app/core/components/Select/UserPicker.tsx#L47

Is there a reason not to list more than 10?  If so, can we be allowed to page through 10 at a time instead of being limited to the first 10?

rvillablanca:

Currently there is a bug which limits the returned user list to 10 when calling endpoint `/api/org/users/lookup`, this value is set in the frontend side and it is harcoded.

This limit is considered very low so this PR removes that value from the frontend and fixed it with a max value of 100 in the backend. Also as the limit parameter was removed from frontend site, the parsed limit value will be 0 which in this case the query would be without limits, so I decided to fixed it to 100 when there is no limit parameter or when it is higher that 100.

I know 100 is a subjetive value but it seems more appropiated for this case. Also this is kind of a workaround because a better solution would be to have pagination or selectable limits values, though I don't have much knowledge about the average value of users so perhaps this fine for this case.

Fixes https://github.com/grafana/grafana/issues/25552

I'm new here and I was looking at the code looking for tests with sample data, users for this case for example, but I couldn't find them so any suggestiong about what I can take a look into would be appreciated.

### DIFF
--- a/pkg/api/org_users.go
+++ b/pkg/api/org_users.go
@@ -64,7 +64,12 @@ func GetOrgUsersForCurrentOrgLookup(c *models.ReqContext) Response {
 		return Error(403, "Permission denied", nil)
 	}
 
-	orgUsers, err := getOrgUsersHelper(c.OrgId, c.Query("query"), c.QueryInt("limit"))
+	limit := c.QueryInt("limit")
+	if limit > 50 {
+		limit = 50
+	}
+
+	orgUsers, err := getOrgUsersHelper(c.OrgId, c.Query("query"), limit)
 	if err != nil {
 		return Error(500, "Failed to get users for current organization", err)
 	}
--- a/pkg/api/org_users.go
+++ b/pkg/api/org_users.go
@@ -65,8 +65,8 @@ func GetOrgUsersForCurrentOrgLookup(c *models.ReqContext) Response {
 	}
 
 	limit := c.QueryInt("limit")
-	if limit > 50 {
-		limit = 50
+	if limit == 0 || limit > 100 {
+		limit = 100
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/api/org_users.go
```diff
@@ -64,7 +64,12 @@ func GetOrgUsersForCurrentOrgLookup(c *models.ReqContext) Response {
 		return Error(403, "Permission denied", nil)
 	}
 
-	orgUsers, err := getOrgUsersHelper(c.OrgId, c.Query("query"), c.QueryInt("limit"))
+	limit := c.QueryInt("limit")
+	if limit == 0 || limit > 100 {
+		limit = 100
+	}
+
+	orgUsers, err := getOrgUsersHelper(c.OrgId, c.Query("query"), limit)
```
torkelo, line 72: This API is used from the users tab under configuration which has no pagination currently, so this low limit here will be a problem

rvillablanca: I'm not quite sure about that, when I see the api calls from users tab under configuration it is `/api/org/users` which is mapped to method `GetOrgUsersForCurrentOrg`, the other api call when adding users to a team is `/api/org/users/lookup` which is mapped to `GetOrgUsersForCurrentOrgLookup`, and this last endpoint is being called only from `UserPicker` component :thinking:

PR #25734, opened 2020-06-22
Annotations: Add Data field to UpdateAnnotation and PatchAnnotation

### LINKED ISSUE #21696: UpdateAnnotation should update field Data too
file: grafana/pkg/api/annotations.go
Function PostAnnotation creates new annotation and allows you to set field Data from cmd.Data.
Function UpdateAnnotation do not read this field from cmd. 
Add please the string "Data: cmd.Data" to function UpdateAnnotation (line:164) to have possibilty to update data field in the database

**Why is this needed**:
We store some additional information in the data field while working with annotations. And we want to have possibility to update this information. At present we have to delete and recreate annotation to update data field.

jeet-parekh:

Adds Data field to UpdateAnnotation and PatchAnnotation.

Fixes #21696.

Question: I noticed that the Data field for PostGraphiteAnnotation is of type string. And for PostAnnotation, it is of type `*simplejson.Json`. For UpdateAnnotation and PatchAnnotation, I have used type `*simplejson.Json`. I guess that would cause an issue when trying to update or patch Graphite annotations...?


pkg/api/annotations.go
```diff
@@ -212,6 +214,8 @@ func PatchAnnotation(c *models.ReqContext, cmd dtos.PatchAnnotationsCmd) Respons
 		existing.EpochEnd = cmd.TimeEnd
 	}
 
+	existing.Data = cmd.Data
```
marefr, line 217: Need an if statement to check if `cmd.Data` is provided.

pkg/api/annotations.go
```diff
@@ -162,6 +162,7 @@ func UpdateAnnotation(c *models.ReqContext, cmd dtos.UpdateAnnotationsCmd) Respo
 		EpochEnd: cmd.TimeEnd,
 		Text:     cmd.Text,
 		Tags:     cmd.Tags,
+		Data:     cmd.Data,
```
marefr, line 165: A bit unsure whether this could be a breaking change.

marefr (dismissed):
See comments

pkg/api/annotations.go
jeet-parekh: I was confused about this initially. I thought that if the user wants to clear the data from the annotation, then they would pass null for `cmd.Data`. But yeah, it makes sense to me now. Fixed this in the next commit.

pkg/api/annotations.go
jeet-parekh: It could be. The docs say that it "Updates all properties". But it could still break things for some users. What do you suggest?

jeet-parekh:
@marefr, could you please review this again?

jeet-parekh:
I was waiting for your comment on the `UpdateAnnotation` function. Will make the necessary changes once we decide what to do about the potential breaking change.

aknuds1:
I think @marefr is going on holiday, so this might have to wait a bit. In the meantime, would you like for me to sync your branch with latest master (in order to fix the Drone CI check)?

aknuds1:
I'm also worried about potential breaking changes, so I don't think we can't conclude on this before having analyzed the PRs effects internally.

jeet-parekh:
@aknuds1, I'll do a rebase and force push in a while.

aknuds1:
@jeet-parekh I think it's better if you merge, and don't force push, since I think that might mess with previous review comments.

jeet-parekh:
Ah okay, I didn't know that you squash merge PRs. Will do a merge then.

pkg/api/annotations.go
aknuds1: @marefr @jeet-parekh I think according to the semantics of the REST PUT verb, the `Data` field should be updated according to what the client passes here. I guess the problem is that the current API doesn't prescribe a `data` field to be passed in the JSON payload, so if we update the API, a client that follows the old spec and doesn't pass `data` will overwrite the annotation data with `nil`.

Might this be a case where we want to increment the REST API version to avoid breaking clients?

Another alternative might be to require that the client passes a non-null value for this field in the JSON payload, to update the data value (i.e., to erase the data, pass `{}` instead of `null` in the payload). Then, old clients not passing the `data` field will be safe.

pkg/api/annotations.go
```diff
@@ -211,6 +213,10 @@ func PatchAnnotation(c *models.ReqContext, cmd dtos.PatchAnnotationsCmd) Respons
 		existing.EpochEnd = cmd.TimeEnd
 	}
 
+	if cmd.Data != nil {
```
aknuds1, line 216: Does one have to pass an empty JSON object in order to erase `Data`?

aknuds1 (commented):
Please see my comments regarding backwards compatibility. I also think the PR needs tests to cover new functionality.

pkg/api/annotations.go
jeet-parekh: I think it should be a JSON null. I'll add a test for this.

pkg/api/annotations.go
jeet-parekh: That alternative approach sounds good. Should I wait for a second approval before changing the implementation?

aknuds1: @marefr What do you think of my idea that the client has to pass an empty JSON object to delete this field, instead of `null` so we retain backwards compatibility?

jeet-parekh:
ping

pkg/api/dtos/annotations.go
```diff
@@ -13,19 +13,21 @@ type PostAnnotationsCmd struct {
...
-	Time    int64    `json:"time"`
-	TimeEnd int64    `json:"timeEnd,omitempty"` // Optional
-	Text    string   `json:"text"`
-	Tags    []string `json:"tags"`
+	Id      int64            `json:"id"`
+	Time    int64            `json:"time"`
+	TimeEnd int64            `json:"timeEnd,omitempty"` // Optional
+	Text    string           `json:"text"`
+	Tags    []string         `json:"tags"`
+	Data    *simplejson.Json `json:"data"`
```
kylebrandt, line 21: Maybe this could be an interface{} instead so it is more straight forward to unmarshal into types.

aknuds1:
Will try to have this reviewed during the week, sorry for the long wait.

pkg/api/annotations.go
wbrowne: Hey @jeet-parekh, do you still hope to add a test for this?

aknuds1: If a JSON `null` is passed, won't `cmd.Data` be `nil`? I think in that case I'm right and you have to pass and empty object?

pkg/api/dtos/annotations.go
aknuds1: @kylebrandt I don't really understand the benefit? Could you provide an example? The corresponding storage type field has the same type at least.

aknuds1 (approved):
AFAICT, this is good, but I think we should complement with a test. I would also like to hear @marefr's feedback. His old review is still blocking the PR after all.

pkg/api/annotations.go
wbrowne: >a client that follows the old spec and doesn't pass data will overwrite the annotation data with nil.

Data is `nullable: false` so technically could never be overwritten unless it's non-null. So I think it makes sense to have the requirement to expect `{}` to clear the annotation data. Then again, would need to check if this is even a valid state for an annotation to still exist.

pkg/api/annotations.go
jeet-parekh: @wbrowne, yes. I was waiting for a confirmation on the other thing. So I could push all the necessary changes together.

@aknuds1, you're right. Passing `null/nil` for `cmd.Data` won't do anything. Passing an empty object `{}` will clear the data.

jeet-parekh:
Question about writing the tests. I am thinking about adding a test case inside the function `TestAnnotationsApiEndpoint` in `pkg/api/annotations_test.go` with role `ROLE_EDITOR`. I'll send a few different PUT requests. But I'll also have to GET the annotation and check if they are correctly modified. So I'll add a function called `getAnnotationsScenario` in the same file. Does that sound good?

wbrowne:
> Question about writing the tests. I am thinking about adding a test case inside the function `TestAnnotationsApiEndpoint` in `pkg/api/annotations_test.go` with role `ROLE_EDITOR`. I'll send a few different PUT requests. But I'll also have to GET the annotation and check if they are correctly modified. So I'll add a function called `getAnnotationsScenario` in the same file. Does that sound good?

Sorry for the delay on this @jeet-parekh. Unfortunately, that won't test everything we need as those tests mock the repository. We just recently starting doing some work around API tests and I think this could be a good candidate for it potentially. I'll follow up in the new year and we'll get this thing closer to being merged. Thanks for your patience!

PR #25736, opened 2020-06-22
Teams: Add index for permission check

### LINKED ISSUE #25688: Dashboard Search/List is extremely slow for Non-Admins
**What happened**: Dashboard Search/List is extremely slow when we make heavy use of Dashboard and Folder level permissions, taking around 40s
In depth description down below

**What you expected to happen**: Grafana to still maintain decent response times, within the limit of a few seconds at max

**How to reproduce it (as minimally and precisely as possible)**: Have a setup with around 1000 users, 300 teams, and a few thousand dashboards, and then set permissions at dashboard level (or even Folder level) with around 10-15 teams per dashboard, with all permissions being for users/teams, and none for Viewer/Editor roles.

The folder/dashboard search query (and consequently, related tasks like dashboard listing) is taking a long time for non-admins on our setup (30-40s), but is super snappy for admins

Relevant Setup Information:
Grafana Backend Store: MySQL 5.6
Number of Dashboards: 6000, spread over 5 Folders (One Folder has 4600 dashboards, another has 1200 dashboards, rest are split among the remaining 3)
Number of Users: 1000
Number of Teams: 300
We are making heavy use of dashboard and folder level ACLs, configuring Viewer/Editor/Admin access per dashboard on the basis of teams. The only access on the basis of Roles is for Admins, none for Viewers and Editors

We've had a look at the Grafana backend database query time logs, and they are definitely the bottleneck
We noticed that in the query that is programmatically generated by Grafana's codebase the following snippet is present at the innermost level

`... team_member as ugm on ugm.team_id = da.team_id WHERE d.org_id = ? AND da.permission >= ? AND (da.user_id = ? OR ugm.user_id = ? OR da.role IN ...`

However, when we look at the indices created by Grafana on the table `team_member`, we find that there are just two indexes:
1. One composite index for `UQE_team_member_org_id_team_id_user_id` , with the order being org_id, team_id, user_id;
2. The second one is for `IDX_team_member_org_id`

This is problematic, as a MySQL composite index for columns (c1, c2, c3) is only useful in case of the following column combinations:
(c1)
(c1,c2)
(c1,c2,c3)

As we can see in the query, the first the query is trying to select rows by matching *only* on team_id and user_id from the team_member table, while the org_id is from the dashboard table's index, which can turn out to be very expensive in this case as the composite index won't be used

We manually added two more indices - `IDX_team_member_user_id` and `IDX_team_member_team_id`, which led to an decrease in query time from 30-40s to 5-6s, which is a significant decrease
This is still rather slow though, so we feel there's probably still room for improvement

We are not sure why these two columns weren't indexed, so we don't know if what we did is right
Also, it might even be possible to just change the generated query to better utilize the present index, we haven't been able to have a deep dive into this issue

If someone could please assist us, we'd be really grateful!

**Environment**:
- Grafana version: 6.7.3
- OS Grafana is installed on: RHEL6 
- User OS & Browser: Any linux variant, Chrome/Chromium
- Grafana plugins:
- Others:

sakjur:

For larger instances, since the dashboard permission call uses `team_id` and `user_id` for SQL querying (which do not have an index available) we risk search and opening dashboards to be extremely slow. This index was proposed by @fionn-mac in #25688 and is a low-hanging fruit for improving search/dashboard performance.

https://github.com/grafana/grafana/blob/b5b1b9d64e0e8772f86b624cf491e6b85d0f23b3/pkg/services/sqlstore/sqlbuilder.go#L49-L63

Fixes #25688


sakjur:
🤔 I thought that would automatically rotate in someone new in the schedule

aknuds1 (approved):
LGTM, but I don't know so much about the migrations. @marefr does this look good to you?

marefr (approved):
LGTM

Would do a sanity check using something like MySQL explain. Not really sure these new indices will be used, but maybe I'm wrong?

sakjur:
I'm not entirely convinced the `user_id` index wouldn't be useful for someone with a large database, but I don't have one right now. Both MySQL and Postgres use the `team_id` index from get-go to reduce cost.

marefr:
@sakjur For referenced query, probably not. Can maybe double check with issue author?

sakjur:
@marefr I'll merge this to make sure it gets into 7.1 (I'm away next week, so don't want to leave it hanging for too long) and will open a new PR with the user id as well if it turns out we needed it.

PR #25739, opened 2020-06-22
Panel edit: Clicking twice on a visualization closes the VizPicker

### LINKED ISSUE #25434: Panel Edit: double click on viz type should collapse the visualization picker
Prior 7.0 double-clicking on the element on the **Visualizations list** made the viz picker to collapse. We should bring back this behavior.

- [ ] Brind back double-click behaviour on the viz picker list
- [ ] Add tooltip with information about the behaviour when user clicks on the active element on the VizPicker
- [ ] Add information about this interaction to the panel edit documentation

peterholmberg:
Fixes an issue where clicking a selected visualization did not close the visualization picker.

Fixes #25434


public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
```diff
@@ -15,6 +15,7 @@ export interface OptionsGroupProps {
   nested?: boolean;
   persistMe?: boolean;
   onToggle?: (isExpanded: boolean) => void;
+  children: (toggleExpand?: any) => ReactNode;
```
torkelo, line 18: Think toggleExpanded should be typed, and should be optional so the type for children should be:
`children: ReactNode | (toggleExpand: () => void) => ReactNode`

Because most users of this component will not use the toggleExpand render prop so feels nice to make it optional

torkelo (changes requested):
The behavior feels great. Would like to make the children render prop callback optional so not all users need to make their children a function

public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
peterholmberg: Ah, yes.

dprokop: @torkelo for the sake of simplicity I would stick with API suggested by @peterholmberg , meaning render prop only. We have 9 usages of the `OptionsGroup` group currently, the implementation is hidden from the plugin authors as it's a core component, so the "easiness" of usage can be compromised IMO with the render prop. I think we are trying to optimize for something that is not necessarily a case here while introducing unnecessary complexity of conditional child rendering in a relatively naive component.

torkelo: yea it's close, but the conditional rendering is just a one-liner in one location, and makes all those 8 other uses a lot easier to read, render prop children is just so messy from the parent component so not having to wrap the elements in a nested function is a big plus in my view for readability in those 8 other consumers of this component

torkelo (approved):
Looks good!

Issue #25713, opened 2020-06-19
Long queue of requests when data loading time is longer than auto-refresh time


szerwi:
**What happened**:
When the data needs more time to load than the auto-refresh interval is set to, Grafana is making long queue of requests to database without cancelling old requests. Even when the old request finishes and returns data, the data is not being showed in the panel, because it is waiting for the response from the newest request.
With long requests queue page may freeze for a couple of minutes and the only solution is to refresh the page.
![1](https://user-images.githubusercontent.com/19311511/85131727-a89c9180-b237-11ea-83f2-e025b22be83d.jpg)
![2](https://user-images.githubusercontent.com/19311511/85131729-aa665500-b237-11ea-95c1-625663df2feb.jpg)
![3](https://user-images.githubusercontent.com/19311511/85131790-c964e700-b237-11ea-88cf-8335abbbc78e.jpg)

**What you expected to happen**:
The best option would be to set timeout time (for example 1 minute). If the data loading is not completed before next auto-refresh, it won't make another auto-refresh and wait for the timeout. After the timeout it should cancel old requests so it won't make long queue.
Another solution would be to add refresh interval per panel as described in #3514

**How to reproduce it (as minimally and precisely as possible)**:
Make a few panels in which the request times are longer than auto-refresh time.

**Environment**:
Grafana version: 7.0.3
Data source type & version: Influxdb 1.7.9
OS Grafana is installed on: Debian 4.9
User OS & Browser: Opera Gx

torkelo:
To fix this we need to change influx (and other data sources) to use the new fetch function (Observable) on backendSrv

Issue #21431, opened 2020-01-10
Explore cancel request


cyriltovena:
**What would you like to be added**:

I would like a way/button to abort a request.

**Why is this needed**:

now with frontend and single queries that can spawn many other queries (150 or more) it would be very nice to be able to cancel/abort a request, the request would then cancel all sub requests, in case of mistake this would be a nice way to recover request budget.

peterholmberg:
Solved via #25746

PR #25746, opened 2020-06-22
BackendSrv: Cancellable requests & Observable all the way  

torkelo:
We have a big problem with http request cancellation not working properly. 

Example:
* #21431
* #25713

BackendSrv only returns promises and promises have no way for the caller to cancel the request. So the way we have done this before is via a requestId and backendSrv keeping track of inflight requests and if the same requestId comes in we cancel the previous request with the same requestId. 

Number of problems with this
* Where and how to set requestId 
* If one call to DataSourceApi.query issues multiple requests (like prometheus does) there is no way merge & cancel them in the current design

This PR 
introduces backendSrv.fetch for a single unififed request processing path that returns an Observable 

Todo
* [x] Single unified request path
* [x] Single unified error handling path 
* [x] old backendSrv.request and backendSrv.datasourceRequest internally using new fetch call 
* [x] Moved out a lot of non generic dashboard & folder functions that had no business living in backendSrv
* [x] Automatically detect if a query is a datasource query (to not show app alerts on errors)
* [x] Updated logic for err.isHandled, added back support for old err.isHandled mechanic to limit scope, was too many files to update
* [x] Investigate how to handle the automatic error & success alerts 
* [x] Handle query inspector inspection with an rxjs Subject (broadcast) stream instead of old appEvents 
* [x] Updated unit tests
* [x] Update a few data source to use the new fetch function
  * [x] TestData

Related problem
* Problem with metricFindQuery returns a promise so no way to cancel template variable queries (To fix this we need to modify the DataSourceApi, which I am leaving to a different PR).


packages/grafana-runtime/src/services/backendSrv.ts
```diff
@@ -15,29 +21,95 @@ export type BackendSrvRequest = {
...
+   */
+  withCredentials?: boolean;
 };
 
+/**
+ * Response for fetch function in {@link BackendSrv}
+ *
+ * @public
+ */
+export interface FetchResponse<T = any> {
```
s-h-a-d-o-w, line 76: This should probably extend `Response` instead of redeclaring its props.

public/app/core/services/backend_srv.ts
```diff
@@ -84,100 +46,33 @@ export class BackendSrv implements BackendService {
...
-    }
-
-    throw data;
-  };
-
-  async request(options: BackendSrvRequest): Promise<any> {
-    // A requestId is a unique identifier for a particular query.
-    // Every observable below has a takeUntil that subscribes to this.inFlightRequests and
-    // will cancel/unsubscribe that observable when a new datasourceRequest with the same requestId is made
+  fetch<T>(options: BackendSrvRequest): Observable<FetchResponse<T>> {
```
s-h-a-d-o-w, line 55: I've considered this a matter of taste for a long time: Since `T` only occurs in the return, it's no better than asserting.

But it just struck me that a return type of `unknown` instead of a "dumb generic" (i.e. it's just passed through) is much more suitable because usually, generics indicate some sort of internal logic. But when used like this, there obviously is none.
`any` could of course be used as well instead of `unknown`, it would just depend on whether we want to force the user to assert the return value in order to access props on it.

As an aside - might it be worthwhile to call it `observableFetch` or `fetchObservable`, to highlight the difference to whatwg fetch?

s-h-a-d-o-w (commented):
I'll have to submit my comments in chunks, otherwise it would take days or longer - a lot to take in here. :wink:

public/app/core/services/backend_srv.ts
torkelo: > Since T only occurs in the return, it's no better than asserting.

You mean typecasting the result? Think this is cleaner. 

> As an aside - might it be worthwhile to call it observableFetch or fetchObservable, to highlight the difference to whatwg fetch?

yea, not sure, we have request already which is exactly the as fetch except returns Promise

packages/grafana-runtime/src/services/backendSrv.ts
torkelo: yea, think your right, but we are not returning the raw Response yet (it has a bunch more), so we can't do that right now without more refactoring and updating tests, but it's a good idea

public/app/core/services/backend_srv.ts
```diff
@@ -457,121 +152,211 @@ export class BackendSrv implements BackendService {
...
+      };
+    }
+
+    // If no message but got error string, copy to message prop
+    if (err.data && !err.data.message && typeof err.data.error === 'string') {
+      err.data.message = err.data.error;
+    }
+
+    // check if we should show an error alert
+    if (err.data.message && !isDataQuery(options.url) && options.showErrorAlert !== false) {
```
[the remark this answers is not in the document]
torkelo: no that is not the same, since if it's undefined we should show error alert. pretty poorly constructed boolean, should be named noErrorAlert and be set to true to disable default behavior but we had a showSuccessAlert already that was done like this. Will look at inverting that but could break some plugins that rely on it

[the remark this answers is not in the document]
s-h-a-d-o-w: I see. Well, if you decide to stick with as it is, then I think it would be a good idea to be more explicit about the behavior, so that people not familiar with it can't make the same mistake as I did. ;)
=> `options.showErrorAlert === undefined || options.showErrorAlert`

public/app/core/services/backend_srv.ts
s-h-a-d-o-w: > You mean typecasting the result?

It's not called a typecast in TS because it doesn't actually cast the type. (i.e. it only changes the type at compile time but doesn't affect run time)

> Think this is cleaner.

How come?

> we have request already which is exactly the as fetch except returns Promise

`observableRequest`? ;)

docs/sources/packages_api/runtime/backendsrv.md
```diff
@@ -26,19 +27,21 @@ import { BackendSrv } from '@grafana/runtime';
 
 ## Remarks
 
-By default Grafana will display an error message alert if the remote call fails. If you want to prevent this from happending you need to catch the error thrown by the BackendSrv and set the `isHandled = true` on the incoming error.
+By default Grafana will display an error message alert if the remote call fails. If you want to prevent this from happending you need to catch the error thrown by the BackendSrv and set the `showErrorAlert = true` on the request options object.
```
oddlittlebird, line 30:
```suggestion
By default, Grafana displays an error message alert if the remote call fails. To prevent this from happening, you need to catch the error thrown by the BackendSrv and set the `showErrorAlert = true` on the request options object.

```

docs/sources/packages_api/runtime/backendsrv.md
```diff
@@ -26,19 +27,21 @@ import { BackendSrv } from '@grafana/runtime';
 
 ## Remarks
 
-By default Grafana will display an error message alert if the remote call fails. If you want to prevent this from happending you need to catch the error thrown by the BackendSrv and set the `isHandled = true` on the incoming error.
+By default Grafana will display an error message alert if the remote call fails. If you want to prevent this from happending you need to catch the error thrown by the BackendSrv and set the `showErrorAlert = true` on the request options object.
+
+> In versions prior to v7.2 you disable the notification alert by setting isHandled on the caught error
```
oddlittlebird, line 32:
```suggestion
> In versions prior to v7.2, you turned off the notification alert by setting isHandled on the caught error.

```

public/app/core/services/backend_srv.ts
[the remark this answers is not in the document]
torkelo: Improved the logic for showErrorAlert and showSuccessAlert, now in the code false mean one thing and undefined means "auto mode"  (and true means force to true). Before the logic was only if false disable auto logic, so there was no way to enable it for cases where auto logic made it not show.

PR #25748, opened 2020-06-22
Notifications: use reply-to option for emails

AgnesToulet:
Use the `ReplyTo` field from the `SendEmailCommand` when sending emails from Grafana.

*Notes for reviewers:*
This is a parameter that can be set for the Reporting feature but we cannot use it without this fix.


marefr (approved):
LGTM

PR #25751, opened 2020-06-22
MSSQL: Fix parsing of uniqueidentifier type

### LINKED ISSUE #17084: MSSQL Datasource: Uniqueidentifier column type
What happened:
Uniqueidentifier column type not displayed correctly.

```
0:"�\�����̔�)�"
1:" �Jߠ6_�ᾒ&��"

```
What you expected to happen:
Response like
```
0:"BCED544A-EE3F-BB64-315A-61E46A285FC5"
1:"6230383E-2AE4-866C-649C-B4EA6D513AB9"
```

How to reproduce it (as minimally and precisely as possible):
Select from any table with this type

Environment:
- Grafana Version 6.1.6 (commit: cf9cb45, branch: HEAD)
- Linux msk1-mailgw01 3.2.0-4-amd64 #1 SMP Debian 3.2.96-2 x86_64 GNU/Linux
- Windows && Firefox 66.0.5 (64)

Temporaly its can be fix with cast

`CAST( columnName AS varchar(255)) as columnName`

edernucci:
This PR fixes one bug viewing columns with Uniqueidentifier datatype.

Closes #17084

I have tested with my own data, before and after the correction with good results.


marefr (changes requested):
Looks good. Would have hoped this could be resolved without using the uuid library, but doesn't look like that. Would like a test of this. Would recommend adding a unique identifier column to this [test](https://github.com/grafana/grafana/blob/d3299796f58fda40166cec689d0feaefac1c2e50/pkg/tsdb/mysql/mysql_test.go#L68).

kylebrandt:
> Looks good. Would have hoped this could be resolved without using the uuid library, but doesn't look like that.

In an old SQL dataframe branch I have:

```
var convertGUID = df.SQLStringConverter{
	Name:          "GUID to HEX String",
	InputScanKind: reflect.Slice,
	InputTypeName: "UNIQUEIDENTIFIER",
	ConversionFunc: func(in *string) (*string, error) {
		if in == nil {
			return nil, nil
		}
		var b [16]byte
		copy(b[:], *in)
		s := mssqldb.UniqueIdentifier(b).String()
		return &s, nil
	},
}
```

So can do it when we get around to dataframe conversion.

edernucci:
Hi @marefr. I'm afraid the test you've pointed are MySQL related. Is there any MSSQL test?

papagian:
> Hi @marefr. I'm afraid the test you've pointed are MySQL related. Is there any MSSQL test?

I was planning to add this test otherwise the modifications LGTM.

Anyway, this is the MSSQL [test](https://github.com/grafana/grafana/blob/d3299796f58fda40166cec689d0feaefac1c2e50/pkg/tsdb/mssql/mssql_test.go#L31)
Make sure to follow [the guidelines](https://github.com/grafana/grafana/blob/d3299796f58fda40166cec689d0feaefac1c2e50/pkg/tsdb/mssql/mssql_test.go#L21) for running this test.

edernucci:
> > Hi @marefr. I'm afraid the test you've pointed are MySQL related. Is there any MSSQL test?
> 
> I was planning to add this test otherwise the modifications LGTM.
> 
> Anyway, this is the MSSQL [test](https://github.com/grafana/grafana/blob/d3299796f58fda40166cec689d0feaefac1c2e50/pkg/tsdb/mssql/mssql_test.go#L31)
> Make sure to follow [the guidelines](https://github.com/grafana/grafana/blob/d3299796f58fda40166cec689d0feaefac1c2e50/pkg/tsdb/mssql/mssql_test.go#L21) for running this test.

Hi @papagian 
If you can add this test for me, I would appreciate it. The week started to get tight here with free time.

pkg/tsdb/mssql/mssql.go
```diff
@@ -130,6 +131,13 @@ func (t *mssqlQueryResultTransformer) TransformQueryResult(columnTypes []*sql.Co
 				} else {
 					t.log.Debug("Rows", "Error converting numeric to float", value)
 				}
+			case "UNIQUEIDENTIFIER":
+				v, err := uuid.FromBytes(value)
+				if err == nil {
+					values[i] = v.String()
+				} else {
+					t.log.Debug("Rows", "Error converting uniqueidentifier to string", value)
+				}
```
marefr, line 140: After @kylebrandt comment I think we actually can use the uniqueidentifier type in mssql lib to parse it instead of uuid lib. Something along these lines:
```go
uuid := &mssql.UniqueIdentifier{}
if err := uuid.Scan(value); err == nil {
  values[i] = uuid.String()
} else {
  t.log.Debug("Rows", "Error converting uniqueidentifier to string", value)
}
```

marefr (changes requested):
See comment. Let us know if you need help with this.

pkg/tsdb/mssql/mssql.go
edernucci: @marefr I have edited (and tested with my own data) as requested. Thanks for sharing your knowledge.

marefr (approved):
LGTM great!

I pushed the addition of test as well.

kylebrandt (approved):

marefr:

edernucci:
> Thank you for contributing to Grafana!

@marefr It is my pleasure to be able to contribute to this wonderful project.

Issue #21429, opened 2020-01-10
Move all custom metadata key/values to the custom section


ryantxu:
The current QueryResultMeta implementation allows any key/value pair:
```
export interface QueryResultMeta {
  [key: string]: any;
  ...
```
we also have a section that allows custom values (since #20854)
```
  // DatasSource Specific Values
  custom?: Record<string, any>;
}
```

Ideally we should have known/standard properties on the main element and than anything custom in custom.   Currently there are only a few things that use custom properties so I think we should move this before 7.0

The only things using values are:
```
    "ERROR TS2339: Property 'json' does not exist on type 'QueryResultMeta'.",
    "ERROR TS2339: Property 'transformations' does not exist on type 'QueryResultMeta'.",
    "ERROR TS2339: Property 'gmdMeta' does not exist on type 'QueryResultMeta'.",
    "ERROR TS2322: Type '{ query: any; }' is not assignable to type 'QueryResultMeta'.\n" +
    "ERROR TS2339: Property 'rawQuery' does not exist on type 'QueryResultMeta'.",
    "ERROR TS2339: Property 'alignmentPeriod' does not exist on type 'QueryResultMeta'.",
```

torkelo:
Wonder if we can have a generic list of stats as part of this that we can show in the panel inspector stats tab. 

Like backend query execution time, fetch time, processing/aggregation time etc (ie coming from the backend)

natellium:
@ryantxu is this still valid? And should it be part of the schema thingies?

PR #25755, opened 2020-06-23
Azure Logs: use executedQueryString in metadata

ryantxu:
In the slow quest to remove the one-off metadata keys: #21429


marefr (approved):
LGTM, but haven't tested it.

PR #25768, opened 2020-06-23
Instrumentation: Fix setting Jaeger tracing address through Grafana config

### LINKED ISSUE #25563: Not possible to set Jaeger tracing address through Grafana config
**What happened**:
Received error when enabled Jaeger tracing using Grafana configuration:
```
EROR[06-12|13:08:08] error when flushing the buffer: write udp 127.0.0.1:54672->127.0.0.1:6831: write: connection refused logger=jaeger
```

That Jaeger should use the address configured in Grafana and not log an error and by that function as expected.

Changed Grafana configuration to enable Jaeger tracing:
```
[tracing.jaeger]
# jaeger destination (ex localhost:6831)
address = 172.21.0.4:6831
```

If I set the environment variable `export JAEGER_AGENT_HOST=172.21.0.4` before starting Grafana Jaeger works as expected. Seems like the problem is that environment variable override in Jaeger client library overrides agent host or port to the default if they're not provided through environment variables. There's a fix merged, but still no official version tagged including those changes, ref https://github.com/jaegertracing/jaeger-client-go/pull/513. 

**Environment**:
- Grafana version: master / 7.0.3

marefr:
Upgrades jaeger-client-go to v2.24.0 to resolve problem with setting Jaeger tracing address through Grafana config.

Fixes #25563 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	github.com/teris-io/shortid v0.0.0-20171029131806-771a37caa5cf
 	github.com/timberio/go-datemath v0.1.1-0.20200323150745-74ddef604fff
 	github.com/ua-parser/uap-go v0.0.0-20190826212731-daf92ba38329
-	github.com/uber/jaeger-client-go v2.20.1+incompatible
+	github.com/uber/jaeger-client-go v2.24.0+incompatible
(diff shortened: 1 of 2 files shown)


aknuds1 (approved):
LGTM

PR #25770, opened 2020-06-23
Templating: Fixes query variable with ${__searchFilter} value selection not causing refresh & url update

### LINKED ISSUE #25671: Templating: When selecting variable using __searchFilter does not trigger refresh / url update
I was able to replicate on the playground, but it's not super easy.
First I had to edit the variable def for app to apps.${__searchFilter} and remove the "Multi-value" and "All" option.
Then when I type 'b' (country was the option) and let the autocomplete return and filter the options to backend and select that.
At this point, the URL is still ...&var-app=country&... and the dashboard doesn't refresh. When I click the refresh button in the timepicker, the panels refresh using the backend var, but the url doesn't change to var-app=backend

torkelo:
Fixes #25671

### DIFF
--- a/public/test/core/redux/reduxTester.ts
+++ b/public/test/core/redux/reduxTester.ts
@@ -50,10 +50,16 @@ export const reduxTester = <State>(args?: ReduxTesterArguments<State>): ReduxTes
   const debug = args?.debug ?? false;
   let store: EnhancedStore<State> | null = null;
 
+  const defaultMiddleware = getDefaultMiddleware<State>({
+    thunk: false,
+    serializableCheck: false,
+    immutableCheck: false,
+  } as any);
+
   const givenRootReducer = (rootReducer: Reducer<State>): ReduxTesterWhen<State> => {
     store = configureStore<State>({
       reducer: rootReducer,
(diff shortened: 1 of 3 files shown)


public/test/core/redux/reduxTester.ts
```diff
@@ -50,10 +50,16 @@ export const reduxTester = <State>(args?: ReduxTesterArguments<State>): ReduxTes
...
+    serializableCheck: false,
+    immutableCheck: false,
+  } as any);
+
   const givenRootReducer = (rootReducer: Reducer<State>): ReduxTesterWhen<State> => {
     store = configureStore<State>({
       reducer: rootReducer,
-      middleware: [...getDefaultMiddleware<State>(), logActionsMiddleWare, thunk] as [ThunkMiddleware<State>],
+      middleware: [...defaultMiddleware, logActionsMiddleWare, thunk] as [ThunkMiddleware<State>],
       preloadedState,
```
torkelo, line 63: This fixes the flooding of console log warnings about unserializable state in redux actions

mckn (approved):
It seems to work as expected! Great work!

PR #25777, opened 2020-06-23
Docs: Updated links

oddlittlebird:
Fixing links


robbymilo (approved):
This should work, thanks!

PR #25778, opened 2020-06-23
Postgres: Support Unix socket for host

### LINKED ISSUE #22119: Using a unix socket for postgres data source (invalid URL escape 2F (/))
Grafana 5.4.3 on OpenBSD 
Issues #12644 #2818 #4030 and #11055 may be related.

When setting up a datasource for postgresql and using /tmp as host for unix socket use, a metric request error is reported.

The logs show invalid URL escape \"%2F\".

Perhaps a socket check box could insert a grayed out / in the host field, if security is a concern?

aknuds1:
Rewrite the Postgres data source's logic for generating connection strings, so that Unix sockets are supported. Basically, I follow the Postgres convention that a host specifier starting with `/` is a Unix socket, otherwise it's expected to be a TCP host or TCP host/port pair.

I also use the normal [connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) format internally, instead of preparing a Postgres connection URL. When using a regular connection string, you don't have to URL encode and it becomes much easier to use a Unix socket pathname for the host. The format is generally much simpler, since it doesn't require URL encoding (only escaping of single quotes and backslashes).

I've read the XORM code, and it passes the connection string unchanged to [pq](https://github.com/lib/pq), so it should have the normal connection string semantics.

Fixes #22119.


pkg/tsdb/postgres/postgres.go
```diff
@@ -47,20 +48,53 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
 func generateConnectionString(datasource *models.DataSource, logger log.Logger) (string, error) {
+	logger.Debug("Trying to generate Postgres connection string")
+
 	sslMode := strings.TrimSpace(strings.ToLower(datasource.JsonData.Get("sslmode").MustString("verify-full")))
 	isSSLDisabled := sslMode == "disable"
 
-	// Always pass SSL mode
-	sslOpts := fmt.Sprintf("sslmode=%s", url.QueryEscape(sslMode))
+	reHost := regexp.MustCompile(`^([^/].*(?::\d+)?)|(/.*)$`)
+	ms := reHost.FindStringSubmatch(datasource.Url)
```
marefr, line 67: Sorry, but this solution feels a bit messy. Having hard time reading and understanding what the code actually does.  Any possibility to use same/similar approach as [mysql](https://github.com/grafana/grafana/blob/5d1508ed8c164cd9d5f054e02290cbcfd491f920/pkg/tsdb/mysql/mysql.go#L34-L45)?

marefr (commented):
Not sure this issue is really a bug since it maybe never have been supported. More of a feature request maybe. See comment

pkg/tsdb/postgres/postgres.go
aknuds1: @marefr It is a fairly uncomplicated regex to be honest, so I'm not really on board with it being messy. In my opinion the MySQL solution is actually more difficult to understand and verify since it encodes the connection string as a URL, and has to encode every parameter as a result (which led to the bug this PR is fixing in the first place).

I can rewrite the Unix socket detection though, by using strings.HasPrefix as in the MySQL code.

pkg/tsdb/postgres/postgres.go
```diff
@@ -47,20 +47,47 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+	logger.Debug("Trying to generate Postgres connection string")
+
 	sslMode := strings.TrimSpace(strings.ToLower(datasource.JsonData.Get("sslmode").MustString("verify-full")))
 	isSSLDisabled := sslMode == "disable"
 
-	// Always pass SSL mode
-	sslOpts := fmt.Sprintf("sslmode=%s", url.QueryEscape(sslMode))
+	var host string
+	var port int
+	if strings.HasPrefix(datasource.Url, "/") {
```
marefr, line 67: This is better. Just came to think about that we've probably missed to validate unix socket in add/update data source validation, [code](https://github.com/grafana/grafana/blob/615b000acd0ae301a159e24bdc6e32f870aaf974/pkg/api/datasources.go#L135).

pkg/tsdb/postgres/postgres.go
```diff
@@ -47,20 +47,47 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+		host = datasource.Url
+		logger.Debug("Generating connection string with Unix socket specifier", "socket", host)
+	} else {
+		sp := strings.SplitN(datasource.Url, ":", 2)
```
marefr, line 71: still think this can be simplified by moving the url check to the bottom, if unix socket return connectionstring without url encode, if not return `&url.URL{...}.String()` as before since it includes url encode.

pkg/tsdb/postgres/postgres.go
```diff
@@ -47,20 +47,47 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
 		logger.Debug("Failed connecting to Postgres", "err", err)
+		return nil, err
 	}
+
+	logger.Debug("Successfully connected to Postgres")
 	return endpoint, err
 }
 
 func generateConnectionString(datasource *models.DataSource, logger log.Logger) (string, error) {
+	logger.Debug("Trying to generate Postgres connection string")
```
marefr, line 60: I don't find this very useful. We already have https://github.com/grafana/grafana/blob/615b000acd0ae301a159e24bdc6e32f870aaf974/pkg/tsdb/postgres/postgres.go#L32-L34

pkg/tsdb/postgres/postgres.go
```diff
@@ -69,28 +96,22 @@ func generateConnectionString(datasource *models.DataSource, logger log.Logger)
...
 
-	u := &url.URL{
-		Scheme: "postgres",
-		User:   url.UserPassword(datasource.User, datasource.DecryptedPassword()),
-		Host:   datasource.Url, Path: datasource.Database,
-		RawQuery: sslOpts,
-	}
-
-	return u.String(), nil
+	logger.Debug("Generated Postgres connection string successfully")
```
marefr, line 113: I don't find this very useful. We already have https://github.com/grafana/grafana/blob/615b000acd0ae301a159e24bdc6e32f870aaf974/pkg/tsdb/postgres/postgres.go#L32-L34

pkg/tsdb/postgres/postgres_test.go
```diff
@@ -15,12 +15,85 @@ import (
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"xorm.io/xorm"
 
 	_ "github.com/lib/pq"
 	. "github.com/smartystreets/goconvey/convey"
 )
 
+// Test newPostgresQueryEndpoint.
+func TestNewPostgresQueryEndpoint(t *testing.T) {
```
marefr, line 27: Nit. This is fine, but could be simplified by testing the `generateConnectionString` function instead. The test of `newPostgresQueryEndpoint` is covered in integration tests.

marefr (changes requested):
Definitely better, but thinks this still can be more simplified.

pkg/tsdb/postgres/postgres.go
aknuds1: @marefr I don't understand the point of using a connection URL here at all, and respectfully disagree that we should use it if the host is not a socket. For one, there's no need to use the URL format (it makes things trickier, and the non-URL connection string is [more standard](https://pkg.go.dev/github.com/lib/pq?tab=doc)), two, using either the URL format or the normal connection string format depending on whether a socket is involved is more complicated logic, and makes for more difficult debugging.

pkg/tsdb/postgres/postgres.go
aknuds1: OK, removed it.

pkg/tsdb/postgres/postgres_test.go
aknuds1: I chose to test newPostgresQueryEndpoint on purpose, since by depending less on implementation details in tests, you ease the maintenance burden and cognitive load. While it may be easier to write tests for `generateConnectionString`, I would argue that maintenance will be more difficult by tying up implementation details in the test.

bergquist: I think this is more fragile since it depends on a mock. Since both functions take the same argument and `generateConnectionString` is smaller and very specific I think testing that is preferred.

pkg/tsdb/postgres/postgres_test.go
```diff
@@ -15,12 +15,98 @@ import (
...
+			database: "database",
+			expErr:   "invalid port in host specifier",
+		},
+		{
+			desc:       "Password with single quote and backslash",
+			host:       "host",
+			user:       "user",
+			password:   `p'\assword`,
+			database:   "database",
+			expConnStr: `user='user' password='p\'\\assword' host='host' dbname='database' sslmode='verify-full'`,
```
bergquist, line 86: :+1:

pkg/tsdb/postgres/postgres_test.go
```diff
@@ -15,12 +15,98 @@ import (
...
+	})
+	var gotDriver string
+	var gotConnStr string
+	sqleng.NewXormEngine = func(driverName string, connStr string) (*xorm.Engine, error) {
+		gotDriver = driverName
+		gotConnStr = connStr
+		return nil, fmt.Errorf("stop here")
+	}
+
+	testCases := []struct {
```
bergquist, line 40: A test with SSL params would be nice so we cover that part of connection string generation while touching this part of the codebase.

aknuds1: @bergquist Done.

pkg/tsdb/postgres/postgres_test.go
aknuds1: @bergquist I've done as you said and made the test for generateConnectionString instead. I think it's problematic however and would rather have kept testing newPostgresQueryEndpoint, since generateConnectionString is only a helper and:

* You lose test coverage of newPostgresQueryEndpoint
* The test becomes more fragile since it depends on the implementation detail of generateConnectionString being used
* Implementation becomes tied up to having the helper function generateConnectionString

bergquist (approved):
LGTM

marefr (approved):

PR #25785, opened 2020-06-23
InfluxDB: improve test response

ryantxu:
The flux datasource test currently gives a misleading error.  This makes it show the right thing

### DIFF
--- a/public/app/plugins/datasource/influxdb/datasource.ts
+++ b/public/app/plugins/datasource/influxdb/datasource.ts
@@ -309,16 +309,20 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
       return super
         .query(request)
         .toPromise()
-        .then((res: any) => {
-          const data: DataQueryResponse = res.data;
-          if (data && data.state === LoadingState.Done) {
-            const buckets = data.data[0].length;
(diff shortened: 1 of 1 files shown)


public/app/plugins/datasource/influxdb/datasource.ts
```diff
@@ -309,16 +309,20 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
...
           }
-          console.log('InfluxDB Error', data);
+          const first = res.data[0];
+          if (first && first.length) {
+            return { status: 'success', message: `${first.length} buckets found` };
+          }
+          console.log('InfluxDB Error', res);
           return { status: 'error', message: 'Error reading buckets' };
         })
         .catch((err: any) => {
```
srclosson, line 324: This line didn't even change, but I've been bit assuming there is a message field. Is the error always an particular type? Maybe we should actually use that type so it's more obvious, otherwise we throw an error on a useful error and loose the error..

srclosson (approved):
LGTM

public/app/plugins/datasource/influxdb/datasource.ts
ryantxu: previous problem is that it showd an error when it did not really exist

PR #25786, opened 2020-06-23
InfluxDB: add executedQueryString metadata for InfluxQL

ryantxu:
This is not totally necessary since the HTTP request shows the value -- however, I think it is nice to have it in the same place:

image

This also adds some more strict types for InfluxSeries


srclosson (approved):
Looks great!

PR #25789, opened 2020-06-24
Theme: Fixes bug in light theme for query editor part

torkelo:
Fixes issue with query function parts being white in editors 

Screenshot from 2020-06-24 08-58-15


tskarhed (approved):
Seems a bit dark to me, but it's your call.
Screenshot from 2020-06-24 11-20-20

torkelo:
There is no other color to choose from in the palette, so would have to introduce a new one

tskarhed:
bg2 fits with the other ones, but the whole point maybe is to have something different.


Screenshot from 2020-06-24 14-40-34

torkelo:
yea, maybe the function parts does not need to different background

mckn (approved):
I agree with @tskarhed and think it looks better with the same background as the other parts. I was thinking that also highlighting the text might be valid in this case since you can interact with it. But not sure.

Skärmavbild 2020-06-25 kl  15 25 03

PR #25792, opened 2020-06-24
Variables: Always Make timeRange Available in MetricFindQuery

### LINKED ISSUE #25715: Return TimeRange on Variable Refresh, no matter Refresh mode.
As a datasource plugin developer, whose plugin is not shipped with Grafana, I need to have Grafana send me the current time range, when I enter `metricFindQuery` during a variable refresh, no matter the refresh type. 

Right now the timeRange only appears in the `options` object  if the Refresh mode is `On time range change`. Otherwise, for the other modes, it is set to undefined.

I believe that `updateQueryVariableOptions` in `public/app/features/variables/query/actions.ts` controls this behavior, and it is only a matter of removing an if-check that keeps the timeRange undefined for the other modes.

**Why is this needed**:
When I enter `metricFindQuery` I need to perform a query via HTTP towards my data service to get my variable values, which always needs a timeRange. As my plugin is external to Grafana I don't have access to `timeSrv` and thus I cannot query for the timeRange myself. I am dependent upon the timeRange returned to me in the `options` object, regardless of which refresh mode has been selected.

If there is a reason why timeRange needs to be undefined in some cases, I'd like to know. Otherwise this seems like a rather simple fix :- )
I can submit it as a PR, I just need to know if that if appropriate.

AlexanderBrandborg:
As a plugin developer it makes things more difficult when the `metricFindQuery` method on my datasource only gets access to the current dashboard timeRange through the queryOptions object, when the variable is refreshed on timeRange change. If the refresh mode is set to `never` or `on dashboard load` I get an undefined object  instead of a timeRange, even when I still need access to the current timeRange in my case.

This PR removes what I understand to be the conditional that prevents the timeRange from being returned when using any other refresh mode than `on time range change`.

Fixes #25715

I admit that I haven't run this locally as it is a rather small fix that I expect doesn't have a big impact on other pieces of code. If this has any bigger impact, please let me know and I'll look a bit deeper into this.

### DIFF
--- a/public/app/features/variables/query/actions.ts
+++ b/public/app/features/variables/query/actions.ts
@@ -27,10 +27,7 @@ export const updateQueryVariableOptions = (
         dispatch(removeVariableEditorError({ errorProp: 'update' }));
       }
       const dataSource = await getDatasourceSrv().get(variableInState.datasource ?? '');
-      const queryOptions: any = { range: undefined, variable: variableInState, searchFilter };
-      if (variableInState.refresh === VariableRefresh.onTimeRangeChanged) {
-        queryOptions.range = getTimeSrv().timeRange();
-      }
+      const queryOptions: any = { range: getTimeSrv().timeRange(), variable: variableInState, searchFilter };
 
       if (!dataSource.metricFindQuery) {
         return;
--- a/public/app/features/variables/query/actions.ts
+++ b/public/app/features/variables/query/actions.ts
@@ -1,6 +1,6 @@
 import { AppEvents, DataSourcePluginMeta, DataSourceSelectItem } from '@grafana/data';
 import { validateVariableSelectionState } from '../state/actions';
-import { QueryVariableModel, VariableRefresh } from '../types';
+import { QueryVariableModel } from '../types';
 import { ThunkResult } from '../../../types';
 import { getDatasourceSrv } from '../../plugins/datasource_srv';
 import templateSrv from '../../templating/template_srv';


public/app/features/variables/query/actions.ts
```diff
@@ -27,10 +27,7 @@ export const updateQueryVariableOptions = (
         dispatch(removeVariableEditorError({ errorProp: 'update' }));
       }
       const dataSource = await getDatasourceSrv().get(variableInState.datasource ?? '');
-      const queryOptions: any = { range: undefined, variable: variableInState, searchFilter };
-      if (variableInState.refresh === VariableRefresh.onTimeRangeChanged) {
-        queryOptions.range = getTimeSrv().timeRange();
-      }
+      const queryOptions: any = { range: getTimeSrv().timeRange(), variable: variableInState, searchFilter };
```
torkelo, line 30: we need to add refresh: variableInState.refresh   here as well so data source knows if it should use current time range in variable name looks or not, this change could have some big breaking behavior so I am not 100% this is correct do to yet

AlexanderBrandborg: Hey, thanks for having a look at this :- )
Excuse a maybe oblivious question, but my understanding would be that variableInState.refresh is returned regardless of if this conditional is present?
But it seems the tests are failing and it is looking like this change is a bit more involved, as you say. I'm in no rush to push this through, but I was hoping to at least get the issue on your radar. I don't have the resources right now, to do a more involved fix, but I may look at it at a later date, unless someone else beats me to it.

PR #25403, opened 2020-06-05
Plugins: move jaeger trace type to grafana data

zoltanbedi:
**What this PR does / why we need it**:

We would like to access jaeger's trace type from x-ray data source plugin. Because of this I've moved trace.ts from jaeger-ui-components package to grafana/data.

PR #25810, opened 2020-06-24
Data: Prefix trace types in data package

zoltanbedi:
This is related to #25403
As @torkelo mentioned in slack these type names were too generic so a prefix was added.


torkelo (approved):
Looks good!

PR #25811, opened 2020-06-24
Arrow DataFrame: cast BigInt/UInt to numbers

### LINKED ISSUE #25809: Arrow DataFrame: int64 support is inconsistent across panels and usage
**What happened**:

When response data contains int64 results, it is often not handled well by panels.  The graph will *sometimes* show, but numeric displays are inconsistent.  The data is properly passed to the frontend, but arrow returns the value as a [BigInt](https://javascript.info/bigint) and fails when anything tries to calculate against it (Error: Cannot mix BigInt and other types)

The simplest solution would be to cast everything to Number before getting used

Show numeric values.

Use master influx datasource w/ flux and query against int64 data

ryantxu:
Fixes #25809 -- well "fix" may not be the right word... this will make it work as expected more of the time.

image

Without this change, we get:
image

Since this is just calculating stats, I can be more careful about the math and will will show values, but still fail with most manipulations


kylebrandt (approved):

packages/grafana-data/src/dataframe/ArrowDataFrame.ts
```diff
@@ -48,14 +49,19 @@ export function arrowTableToDataFrame(table: Table): ArrowDataFrame {
...
         case ArrowType.Decimal:
-        case ArrowType.Int:
         case ArrowType.FloatingPoint: {
           type = FieldType.number;
           break;
         }
+        case ArrowType.Int: {
+          type = FieldType.number;
+          values = new NumberColumn(col);
+          console.log('INT field', schema);
```
ryantxu, line 62: TODO: remove the console

PR #25815, opened 2020-06-24
Monaco: lazy loading code editor

ryantxu:
This is an alternative approach to #25251, it uses a [different library ](https://github.com/suren-atoyan/monaco-react) that avoids compiling with webpack.  This introduces a `lib` folder that is just a static copy from the monaco dist.  It is loaded only when the component first loads:

image

editor.main.js is 2mb ;(

My main question with this approch is how to import & use monaco types without having webpack do too much work.  My initial atempts have been unsuccesful.

Please ignore the other component properties for now (like AutoSizer...) we can sort that out after we decide on how to best load the resources


torkelo:
Why can't we use webpack async chunk loading with the react-monaco-editor (other PR approach) ? https://github.com/grafana/grafana/pull/25251/files#diff-d3ec0babb3902f81b01d6e16a4319020R13

dprokop:
@torkelo I think it's mainly because we are experimenting with getting the actual editor being loaded on demand. 

@ryantxu I'm not a fan of this solution. Even though it is somewhat easier, it forces us to manage the monaco dependency outside of other packages pipeline. I would love to avoid that. This will also force us to include this somehow in the grafana-ui, i.e. for the storybook build. 

When it comes to importing the types - think this is exactly the same problem as with first attempts in #25251- even though only types are imported from monaco, the entire package ends up in the vendors :/ Couldn't find any other way for fixing this than having monaco as split chunk.

PR #25818, opened 2020-06-25
Slack: Support use of chat.postMessage

### LINKED ISSUE #25817: Slack: Support chat.postMessage for posting notifications
Support  using [chat.postMessage](https://api.slack.com/methods/chat.postMessage) for the Slack notification URL, so long as an auth token is also provided.

**Why is this needed**:
Slack doesn't allow overriding channel for incoming Webhooks now.

https://api.slack.com/messaging/webhooks (custom integration webhooks are legacy and should be replaced with apps)
> You cannot override the default channel (chosen by the user who installed your app), username, or icon when you're using Incoming Webhooks to post messages. Instead, these values will always inherit from the associated Slack app configuration.

If user want to send slack message to multiple channel, user need one incoming Webhook per one channel.
It is hard to maintain these webhooks.
Grafana already has ability to use token, so let's use token to post message.

mtanda:
Support use of [chat.postMessage](https://api.slack.com/methods/chat.postMessage) to send message.

To test this feature, set url to `https://slack.com/api/chat.postMessage` and initialize the token field with a valid app OAuth token.

Fixes https://github.com/grafana/grafana/issues/25817


mtanda:
@marefr If you have time, please review this :-)

pkg/services/alerting/notifiers/slack.go
```diff
@@ -291,15 +291,17 @@ func (sn *SlackNotifier) Notify(evalContext *alerting.EvalContext) error {
...
-		"blocks": blocks,
+		"text": evalContext.GetNotificationTitle(),
 		"attachments": []map[string]interface{}{
 			attachment,
 		},
 		"parse": "full", // to linkify urls, users and channels in alert message.
 	}
 
 	//recipient override
+	if len(blocks) > 0 {
```
aknuds1, line 302: What is the practical effect of only including blocks if non-empty?

aknuds1: Also, how is this change relevant to the PR's topic?

mtanda: Without this change, `json.Marshal()` output `"blocks":null,` to POST body for `chat.postMessage` .
I'm not sure Slack ignore this noisy parameter, but it is safe to remove.

aknuds1: The comment `//recipient override` is also now in the wrong place, since it applies to line 305.

aknuds1 (changes requested):

aknuds1:
I've edited the PR title and description to accurately describe what the PR does.

aknuds1:
Before I can approve this PR, I'd like to see updates to our [documentation](https://grafana.com/docs/grafana/latest/alerting/notifications/#slack), that specifies that the URL field can also be _https://slack.com/api/chat.postMessage_ if the fields Recipient and Token are also correctly filled in(since the `channel` parameter is required by chat.postMessage and it also requires a token for auth).

pkg/services/alerting/notifiers/slack.go
mtanda: Sorry, I didn't care about comment. I move and simplify the code. Would you check?

docs/sources/alerting/notifications.md
```diff
@@ -110,6 +110,7 @@ Mention Channel | Optionally mention either all channel members or just active o
 Token | If provided, Grafana will upload the generated image via Slack's file.upload API method, not the external image destination.
 
 If you are using the token for a slack bot, then you have to invite the bot to the channel you want to send notifications and add the channel to the recipient field.
+To use `chat.postMessage` API to post message, you have to set `Url` as `https://slack.com/api/chat.postMessage`, and `Recipient` and `Token`.
```
oddlittlebird, line 113:
```suggestion
To use the `chat.postMessage` API to post message, set the following values: 
- `Url` - `https://slack.com/api/chat.postMessage` 
- `Recipient` -
- `Token` - 
```

oddlittlebird: Please add what the user needs to put in the Recipient and Token fields.

oddlittlebird (dismissed):

docs/sources/alerting/notifications.md
```diff
@@ -110,6 +110,7 @@ Mention Channel | Optionally mention either all channel members or just active o
 Token | If provided, Grafana will upload the generated image via Slack's file.upload API method, not the external image destination.
```
aknuds1, line 110:
```suggestion
Token | If provided, Grafana will upload the generated image via Slack's file.upload API method, not the external image destination. If you use the [chat.postMessage](https://api.slack.com/methods/chat.postMessage) API, this is also required.
```

aknuds1 (changes requested):
Looks better, but I think there's some room for improvement.

docs/sources/alerting/notifications.md
aknuds1: @oddlittlebird I think it's better to amend the _Url_, _Recipient_ and _Token_ field descriptions above, since they already state what the fields should contain. Basically, these fields need to be rethought a bit, since mtanda suggests another way of using this functionality.

I made a suggestion for the _Token_ field documentation, but GitHub won't allow the same for the other two fields. I can make those changes later, through another method.

docs/sources/alerting/notifications.md
oddlittlebird: Sounds good.

aknuds1:
@mtanda, we need some more changes to the docs. Do you mind if I push changes to your branch?

mtanda:
@aknuds1 If you improve the document, it is really helpful. Thanks lot :-)

aknuds1:
@mtanda OK, I will make some improvements.

aknuds1:
I pushed the changes I had in mind. I will try to test the PR now.

aknuds1 (approved):
LGTM, but have to test.

aknuds1:
Tested it and can confirm it works.

aknuds1:

mtanda:
Thanks for helping me and merge PR!

PR #8092, opened 2017-04-11
Proposed fix to #2524 by limiting number of returned measurements

sbhenderson:
This is a proposed fix to #2524 by just limiting the number of returned measurements.

Issue #2524, opened 2015-08-15
Slow 'select measurement' rendering for InfluxDB


JozoVilcek:
I am experimenting with grafana 2.1.1. When tying to populate graph from influxdb data source, I experience long lags on measurement selection. When I simply click in the box, grafana sends SHOW MEASUREMENTS query to influxdb. Response is returned in about 150ms. But for the drop down list to show up, I have to wait another 20s. Number of returned measurement is about 7k.

Is it supposed to be that slow?
If yes, is it possible to somehow set a LIMIT on select measurements queries?

torkelo:
This is bad, sorry for this. I will take a look at fixing this.  

The main issue is that if I use LIMIT on this query people will not able to use the the auto complete search to find the measurement they want. The fix for that would be to have a low limit, like 50 and issue repeated requests for each key entered, but do not think the InfluxDB 
SHOW MEASUREMENTS query supports regex match / filtering, or does it? ping @pauldix

One temporary work around is to pick tags before you pick measurement, this will limit the measurements to what exists for the tag filters selected.

bubenkoff:
+1 for solving this.
Show measurements does support where clause so with combination of limit it should suceed.

torkelo:
@bubenkoff do you have a syntax for that? I know it supports where clause for tags, but for the measurement name? haven't seen any example for such a query

bubenkoff:
you're right, existing where syntax doesn't help at all, if you only have measurements and no tags, which is my case
Looks like we have to request a change for it in influxdb

bubenkoff:
https://github.com/influxdb/influxdb/issues/3523

torkelo:
Its not much I can do here before InfluxDB implements the filtering.

JozoVilcek:
I have played with kairsdb backend, and grafana UI is much more responsive for roughly same number of metric names than with influxdb backend.

pauldix:
How many measurements do you have? If you're using InfluxDB, you really should be using tags. The whole reason we introduced tags was that searches in a flat measurement namespace didn't scale well if you had a large number of unique series.

We may introduce a feature to filter the measurement names, but I fear it will only be delaying the problem.

JozoVilcek:
It is about 7k measurements. It will grow and yes, long term solution is to for tags. 
Just wanted to point finger on different behavior in grafana when using influxdb a kairosdb backends.

pauldix:
[...]

PR #25824, opened 2020-06-25
Explore/InfluxDB: Remove limit on number of measurements listed in log explore

chrisfarms:
## What

* make callers of the influxdb query builder sets query limits explicitly
* removes the limit on the number of measurements loaded into the list in the influxdb logs explore pane.

## Why

in #8092 (a resolution to #2524) a hard-coded `LIMIT 100` was added to mitigate issues of slow running queries on databases with large numbers of measurements. This hard-coded limit is generally not noticed as the queries are mostly combined with a filter (such as the autocomplete query) so it is only "100 of the search query".

However because the value is hard-coded internally, when the query builder is been used elsewhere in contexts without such
a query filter, the limit of 100 is unexpected.

This behaviour can be seen on the "logs" explore view for influxdb, where if your influxdb database contained more than 100 measurements, and your "logs" measurement wasn't in the
first 100 returned, then there was no way to select your "logs" field, which was confusing (is my log shipping broken?)
and annoying (why can't I play with the handy explore!).

I believe the limit of 100 behaviour was not intentional on the logs explore view, as the component appears to have been
designed with the expectation that all measurements are returned.

Without the limit this page **is going to be slow for people with large numbers of metrics** (it performs N+1 requests!), so
I hope this change will:

1. make the page at least usable (even if slow for some)
2. highlight the larger problem with this component that we should not be making "show measurements" queries without filters or N+1 requests

### DIFF
--- a/public/app/plugins/datasource/influxdb/components/InfluxLogsQueryField.tsx
+++ b/public/app/plugins/datasource/influxdb/components/InfluxLogsQueryField.tsx
@@ -60,7 +60,7 @@ export class InfluxLogsQueryField extends React.PureComponent<Props, State> {
     const { datasource } = this.props;
     try {
       const queryBuilder = new InfluxQueryBuilder({ measurement: '', tags: [] }, datasource.database);
-      const measureMentsQuery = queryBuilder.buildExploreQuery('MEASUREMENTS');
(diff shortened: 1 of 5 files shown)


chrisfarms:
Rebased to resolve conflicts

PR #25825, opened 2020-06-25
LDAP: Get all groups for all group base search DNs

### LINKED ISSUE #17137: LDAP multiple group_search_base_dns not respected
**What happened**:

When using multiple group_search_base_dns in LDAP.toml, only the first entry in that array is used. 

I confirmed this by swapping the order of the entries. Whatever the order, only the first is considered.

After enabling the logging, it does show groups from only first entry.

Similar issue reported earlier in [this grafana community page](https://community.grafana.com/t/ldap-group-search-base-dns-multiple-dns/693)

**Environment**:
- Grafana version: 6.x
- User OS & Browser: Chrome

Annegies:
Signed-off-by: Annegies van 't Zand <ace.vtzand@gmail.com>

The LDAP configurations allows to search through multiple base DNs for groups, however currently once a base DN returns any group(s) it won't search the other base DN's specified.
This PR makes sure all groups in all the group base DNs specified are returned.

Fixes #17137

### DIFF
--- a/pkg/services/ldap/ldap.go
+++ b/pkg/services/ldap/ldap.go
@@ -538,7 +538,6 @@ func (server *Server) requestMemberOf(entry *ldap.Entry) ([]string, error) {
 					getAttribute(groupIDAttribute, group),
 				)
 			}
-			break
 		}
 	}
 


sbkg0002:
Any update on this? We also would like to have this merged!

xlson (approved):
Thanks for contributing and fixing the bug with searching for groups.

PR #25834, opened 2020-06-25
Chore: Consolidate on golangci-lint

aknuds1:
Consolidate on (latest) golangci-lint for our Go linting needs, meaning we don't need `go vet` or `gosec` separately.

Also upgrading revive, since we might as well?

Fixing also some issues in Go code that gosec pointed out, due to for loop variable aliasing.


ryantxu (approved):
I am all for more strict linting :)

marefr (approved):
LGTM

Great ❤️

PR #25835, opened 2020-06-25
Grafana-UI: Add FileUpload

### LINKED ISSUE #25794: Grafana-UI: Add FileUpload button
A button that triggers file upload popup and has predefined styles has a few use cases already and thus should be refactored into a separate component. [DashboardFileUpload](https://github.com/grafana/grafana/blob/master/public/app/features/manage-dashboards/components/DashboardFileUpload.tsx) should be used a base with customisable `title`, `accepts`, etc props.

Clarity-89:

Abstract file upload popup trigger functionality into separate component inside grafana/ui.

Fixes #25794 


packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -1,39 +1,43 @@
...
-import { css } from 'emotion';
+import { css, cx } from 'emotion';
+import { getFormStyles } from '../index';
+import { stylesFactory, useTheme } from '../../themes';
 
-interface Props {
+export interface Props {
   onFileUpload: (event: FormEvent<HTMLInputElement>) => void;
+  label?: string;
+  accept?: string;
```
tskarhed, line 10: Short description on what this does might be good

tskarhed:
Not sure what is doable here, but having the label change to `2 selected files` or something similar might be nice - maybe show the file name if it is just one. Or should this be manually implemented from case to case?
Screenshot from 2020-06-25 13-53-38

Clarity-89:
> Not sure what is doable here, but having the label change to `2 selected files` or something similar might be nice - maybe show the file name if it is just one. Or should this be manually implemented from case to case?
> Screenshot from 2020-06-25 13-53-38

`multiple={false}` - this is a single file upload. Not sure if we have use cases for multiple uploads.

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -1,39 +1,45 @@
...
-export const DashboardFileUpload: FC<Props> = ({ onFileUpload }) => {
+export const FileUpload: FC<Props> = ({ onFileUpload, className, label = 'Upload file', accept = '*' }) => {
   const theme = useTheme();
   const style = getStyles(theme);
 
   return (
-    <label className={style.button}>
-      Upload .json file
+    <label className={cx(style.button, className)}>
+      {label}
```
tskarhed, line 22: This might be an idea
```suggestion
      <Icon name="upload"/>
      {label}
```

tskarhed:
Okay, I get that. Just maybe something that just visibly changes when a file has been selected. Not quite sure what it could be.

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
Clarity-89: Good idea!

peterholmberg (commented):
Nice, my only question is. Should this be in grafana/ui? What cases/support to do we have for uploading files to Grafana?

@dprokop :point_up:

Clarity-89:
> Nice, my only question is. Should this be in grafana/ui? What cases/support to do we have for uploading files to Grafana?

We have at least one use case in enterprise with license uploads. Also I thought we're not separating between core and grafana/ui components anymore. Maybe can make this internal for now at least?

dprokop:
> Also I thought we're not separating between core and grafana/ui components anymore

Why wouldn't we? We are still going to have core, reusable components which will not make sense for plugin devs IMO.

>  Maybe can make this internal for now at least?

I think we don't need to. This is fairly basic functionality, better for people to reuse if needed rather than implementing their own. Also, I'm not sure if it will be wildly used :)

Clarity-89:
> Why wouldn't we? We are still going to have core, reusable components which will not make sense for plugin devs IMO

One of the reasons is that it makes it difficult to keep track and list the components we have. Another is that it's hard for UX team to keep styles in sync and update them. I think it was one of the points @jessover9000 raised at the last design-system meeting.

ryantxu:
> What cases/support to do we have for uploading files to Grafana?

@peterholmberg  lots!  many plugins have upload elements that are currently a bit jenkey.  A few places have you upload credentials files that then parse and and put them in the right place

Clarity-89:
> Okay, I get that. Just maybe something that just visibly changes when a file has been selected. Not quite sure what it could be.

@tskarhed In the current case (DashboardFileUpload) the view changes after file is uploaded, but I think it'd be good indeed to have an option to show the uploaded file name (could be under a prop `showFileName` maybe). How about this:

Screenshot 2020-06-26 at 8 51 47

In general, for the future, I think we'd also have a more robust solution, like [dropzone](https://www.dropzonejs.com/), so uploaded files are preview-able.

tskarhed:
@Clarity-89 I think it looks good. Some feedback from the UX people would be great though.

jessover9000:
The button itself looks fine. Some questions:

- is it always going to be a primary button or should we also have a secondary variant?
- is it just a "choose file" trigger that would need a separate upload button or does it upload directly upon choosing a file? If yes, why do we show the file name there? To show to the user what file they currently have uploaded?
- If the button triggers an upload directly, how does the user get feedback on whether their upload is working/the file is already uploaded?

Clarity-89:
> * is it always going to be a primary button or should we also have a secondary variant?

There's no use cases for that, but would probably make sense to have it customizable.

> * is it just a "choose file" trigger that would need a separate upload button or does it upload directly upon choosing a file? 
If yes, why do we show the file name there? To show to the user what file they currently have uploaded?

The only use case we have for this now is for dashboard import at `/dashboard/import`. There the button only triggers the popup and rest is handled outside of the component. We also gonna have smth similar in enterprise with license upload but there too the handing of uploaded files happens outside of the component. 
The file name is to show that smth actually happened, for example when the upload is used inside a form. 

> * If the button triggers an upload directly, how does the user get feedback on whether their upload is working/the file is already uploaded?

The button doesn't make the actual file upload, that is left to the consumer via `onFileUpload` callback.

jessover9000:
> The only use case we have for this now is for dashboard import at /dashboard/import.

Alright, I was missing this information. Looking at how it works there, I guess it makes sense. The "Import dashboard" view has its own UX fallacies, but not directly related to that button. General question then though, will the upload button always work like it does there – triggering a new screen with more upload options? I think this might be important information for the docs. :)

Clarity-89:
> General question then though, will the upload button always work like it does there – triggering a new screen with more upload options? I think this might be important information for the docs. :)

I'll add it to the docs. Also, as I wrote above, 

> In general, for the future, I think we'd also have a more robust solution, like [dropzone](https://www.dropzonejs.com/), so uploaded files are preview-able. 

But that'd probably be a different component and  we'd need designs for that :)

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { FC, FormEvent, useState } from 'react';
+import { GrafanaTheme } from '@grafana/data';
+import { css, cx } from 'emotion';
+import { getFormStyles, Icon } from '../index';
+import { stylesFactory, useTheme } from '../../themes';
+
+export interface Props {
+  onFileUpload: (event: FormEvent<HTMLInputElement>) => void;
+  /** Custom upload label text */
+  label?: string;
```
dprokop, line 10: Since this is button-ish component can we use `children` instead of `label` to keep the API similar?:

```
<FileUpload ...>
  Upload file...
</FileUpload>
```

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const delimiter = fileName.lastIndexOf('.');
+  const extension = fileName.substring(delimiter);
+  const file = fileName.substring(0, delimiter);
+
+  if (file.length < nameLength) {
+    return file;
+  }
+
+  return file.substring(0, nameLength) + '...' + extension;
+}
```
dprokop, line 27: Basic thing, but worth having a test for it

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const FileUpload: FC<Props> = ({ onFileUpload, className, label = 'Upload file', accept = '*' }) => {
+  const theme = useTheme();
+  const style = getStyles(theme);
+  const [fileName, setFileName] = useState('');
+
+  const onChange = (event: FormEvent<HTMLInputElement>) => {
```
dprokop, line 34: Use `useCallback` maybe?

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,77 @@
...
+  if (file.length < nameLength) {
+    return file;
+  }
+
+  return file.substring(0, nameLength) + '...' + extension;
+}
+
+export const FileUpload: FC<Props> = ({ onFileUpload, className, label = 'Upload file', accept = '*' }) => {
+  const theme = useTheme();
+  const style = getStyles(theme);
```
dprokop, line 31: `useStyles` instead :)

packages/grafana-ui/src/components/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const file = fileName.substring(0, delimiter);
+
+  if (file.length < nameLength) {
+    return file;
+  }
+
+  return file.substring(0, nameLength) + '...' + extension;
+}
+
+export const FileUpload: FC<Props> = ({ onFileUpload, className, label = 'Upload file', accept = '*' }) => {
```
dprokop, line 29: I wonder if this shouldn't belong to forms namespace maybe? Thoughts @tskarhed ?

public/app/features/manage-dashboards/DashboardImportPage.tsx
```diff
@@ -78,7 +77,7 @@ class DashboardImportUnConnected extends PureComponent<Props> {
     return (
       <>
         <div className={styles.option}>
-          <DashboardFileUpload onFileUpload={this.onFileUpload} />
+          <FileUpload label="Upload .json file" accept="application/json" onFileUpload={this.onFileUpload} />
```
dprokop, line 80:
```suggestion
          <FileUpload label="Upload JSON file" accept="application/json" onFileUpload={this.onFileUpload} />
```

dprokop (approved):
Looks good code-wise. Left few nits to consider :)

PR #25836, opened 2020-06-25
@grafana/e2e: added tests

stevenvachon:
See commit message(s).


ryantxu (approved):
the momemt upgrade is the only thing we should be sure to tipple check is OK

yarn.lock
```diff
@@ -10590,39 +10595,39 @@ cypress-file-upload@^4.0.7:
...
     is-ci "2.0.0"
-    is-installed-globally "0.1.0"
+    is-installed-globally "0.3.2"
     lazy-ass "1.6.0"
     listr "0.14.3"
     lodash "4.17.15"
     log-symbols "3.0.0"
     minimist "1.2.5"
-    moment "2.24.0"
+    moment "2.26.0"
```
ryantxu, line 10630: Is this necessary now?   what does it impact?

stevenvachon: Wow, you check lockfile changes :) Yarn would've done that automatically.

PR #25837, opened 2020-06-25
Add csp support and refactor security headers

### LINKED ISSUE #6820: Feature request: Set security headers
It would be useful if grafana could be configured to set security-related headers (`x-xss-protection`, `x-frame-options`, `content-security-policy`, etc). In my specific use case, I need to add these headers to comply with nist guidelines, but adding them would also add extra protection against xss and other attacks. Would you all be interested in using a library like https://github.com/unrolled/secure, or in adding the headers that it adds manually?

### LINKED ISSUE #16655: CSP Compliance
Grafana should be CSP compliant. This is pretty much a standard at this point.
What does that mean?
All `eval()` and `inline` script must be removed from the code base, this will allow us to remove the `unsafe-eval` and `unsafe-inline` from `script-src` that are currently needed to run it.

**Why is this needed**:
Security should be a priority at all time, CSP headers helps us prevent XSS and other attacks. Those are good practices. This was already raised in 2016 by https://github.com/grafana/grafana/issues/6820 and I think it should be made a priority.

iOrcohen:
Add CSP support to Grafana.
Following 2 open issues and our needs of Grafana to be secured (as much as possible) from XSS attacks, we added CSP compliance to Grafana.
We added https://github.com/unrolled/secure middleware for handling security headers, and default config for the CSP.

Fixes https://github.com/grafana/grafana/issues/16655
Fixes https://github.com/grafana/grafana/issues/6820

We add unsafe-eval option for the script-src because AngularJS force us to. Once you will finish the migrate away from Angular, this option could be removed.


orc1535:
Looking on some failed tests in backend.

conf/defaults.ini
```diff
@@ -209,6 +210,41 @@ x_content_type_options = true
...
+block_all_mixed_content = true
+
+#  Restricts the URLs which can be loaded using script interfaces
+connect_src = 'self' grafana.com
+
+# Specifies which manifest can be applied to the resource
+manifest_src = "'self'"
+
+# Specifies valid sources for loading media using the <audio> and <video>
+media_src = "'none'"
```
marefr, line 246: This is just adding too much configuration to an already huge configuration file. CSP should be considered a rather advanced use case and if you're using a reverse proxy you can add all of this there. But for us to accept this I would consider having a `content_security_policy` setting of string where you can input whatever you want and by default it should not be set.

marefr (commented):
Can you please split up embedding and CSP in two different PR's - that should increase the chance of getting them reviewed and accepted. Thanks

conf/defaults.ini
orc1535: Like this ? :
content_security_policy = script-src 'unsafe-eval' 'strict-dynamic' $NONCE; object_src = 'none'; font_src...

orc1535: It's not possible since semicolon in INI files starting a comment.
If you insist doing it that way, the only option is to change :
https://github.com/grafana/grafana/blob/master/pkg/setting/setting.go#L467 
https://github.com/grafana/grafana/blob/master/pkg/setting/setting.go#L510

from `ini.Load(filePath)` to `ini.LoadSources(LoadOptions{UnescapeValueCommentSymbols : true}, filePath)`, which allow using semicolon with escape, like this : `\;`.

Is that the an acceptable solution for you ?

iOrcohen:
Looks like tests are passing now.
Removed the embedded part of the PR and may open PR about it later.
Please reconsider the CSP config in the ini file, and let me know yours decision 😁
Thanks ahead.

conf/defaults.ini
marefr: You have to wrap it in triple quotes, like [here](https://grafana.com/docs/grafana/latest/administration/configuration/#password)

iOrcohen: Ok, changed that to work that way :)

aknuds1:
Would you mind please merging latest master into your branch? We have recently switched to Drone CI for our testing our PRs, and every PR branch needs to have .drone.yml in them :)

aknuds1:
Thanks :)

aknuds1:
There's a linting failure, due to a new inter we enabled since last you synced with master.

conf/defaults.ini
```diff
@@ -770,4 +776,4 @@ license_path =
 
 [feature_toggles]
 # enable features, separated by spaces
-enable =
+enable =
```
aknuds1, line 779: Please re-add the final newline of this file.

pkg/middleware/headers.go
```diff
@@ -12,3 +16,39 @@ func HandleNoCacheHeader() macaron.Handler {
 		ctx.SkipCache = ctx.Req.Header.Get(HeaderNameNoBackendCache) == "true"
 	}
 }
+
+func AddSeceureResponseHeaders() macaron.Handler {
```
aknuds1, line 20:
```suggestion
func AddSecureResponseHeaders() macaron.Handler {
```

pkg/middleware/headers.go
```diff
@@ -12,3 +16,39 @@ func HandleNoCacheHeader() macaron.Handler {
...
+	return func(res http.ResponseWriter, req *http.Request, c *macaron.Context) {
+
+		secureMiddleware := secure.New(createSecureOptions())
+
+		nonce, _ := secureMiddleware.ProcessAndReturnNonce(res, req)
```
aknuds1, line 24: Why is the error ignored here? I would handle it, although I'm not sure off the top of my head how to in these handlers.

pkg/middleware/middleware_test.go
```diff
@@ -567,6 +567,8 @@ func middlewareScenario(t *testing.T, desc string, fn scenarioFunc) {
 
 		sc.m.Use(OrgRedirect())
 
+		sc.m.Use(AddSeceureResponseHeaders())
```
aknuds1, line 570:
```suggestion
		sc.m.Use(AddSecureResponseHeaders())
```

pkg/api/http_server.go
```diff
@@ -323,6 +323,8 @@ func (hs *HTTPServer) addMiddlewaresAndStaticRoutes() {
 	}
 
 	m.Use(middleware.HandleNoCacheHeader())
+
+	m.Use(middleware.AddSeceureResponseHeaders())
```
aknuds1, line 327:
```suggestion
	m.Use(middleware.AddSecureResponseHeaders())
```

aknuds1 (dismissed):

pkg/middleware/headers.go
iOrcohen: I guess if an error occurred the function should just "return", like any other middleware behaviour that doesn't have reqContext.

pkg/middleware/headers.go
```diff
@@ -17,11 +17,16 @@ func HandleNoCacheHeader() macaron.Handler {
...
+func AddSecureResponseHeaders() macaron.Handler {
 	return func(res http.ResponseWriter, req *http.Request, c *macaron.Context) {
 		secureMiddleware := secure.New(createSecureOptions())
 
-		nonce, _ := secureMiddleware.ProcessAndReturnNonce(res, req)
+		nonce, err := secureMiddleware.ProcessAndReturnNonce(res, req)
+
+		if err != nil {
+			return
+		}
```
aknuds1, line 28:
```suggestion
		nonce, err := secureMiddleware.ProcessAndReturnNonce(res, req)
		if err != nil {
			return
		}
```

aknuds1 (commented):
Looks better now. I'm unfamiliar with the functionality, so will have to research a bit.

aknuds1:
I think the current CI failure is random BTW.

conf/defaults.ini
```diff
@@ -209,6 +209,12 @@ x_content_type_options = true
 # when they detect reflected cross-site scripting (XSS) attacks.
 x_xss_protection = true
 
+# Set Content Security Policy string to add the Content-Security-Policy header for your requests.
+# CSP allows to control resources the user agent is allowed to load and help to prevent XSS atacks.
+# Add $nonce to script-src to add random nonce.
```
aknuds1, line 214: Should it be $NONCE instead? That's how it's written in the actual string below.

aknuds1:
I wonder if another PR should be extracted from this, introducing the secure middleware _without_ CSP, since the PR replaces our current secure header writing with said secure middleware. So there are basically two changes we have to consider.

iOrcohen:
Well I got your point but I'm not sure I'm fully agree with.

The new [middleware](https://github.com/unrolled/secure) is do some pretty logic of managing the nonce in CSP (create it, replace it in the CSP string and etc..).
CSP in the basic form without nonce can be added just as string to the CSP header like other security headers you have, but won't be so relevant without the nonce. 

If eventually you will feel that this should be two different PR, Ill separate them, but this middleware, as is see it, is "meant to be" for CSP with NONCE so I think it should be together 😊

aknuds1:
The new middleware does more than just enable us to use CSP though, i.e. it replaces old logic in Grafana, so it represents a risk wrt. potentially introducing regressions. Can't you replicate the new middleware's CSP logic (producing nonce etc) without it?

In any case, we will have to discuss whether to add CSP support internally. A lot of people are on vacation now, so it'll have to be scheduled to August.

iOrcohen:
I can, but I'm not sure why if this package manage it so well.
Anyway, I'll wait for new updates about it.

aknuds1:
We're generally hesitant about adding new dependencies - we haven't even got to analyzing if this new module breaks things or not.

aknuds1:
@sakjur @xlson Might you have any opinion on this feature?

sakjur:
When it comes to CSP I think we should support it. I think this is a reasonable PR and I think this belongs in Grafana.

That said, I'm concerned about how we can do this in a way that contains the negative consequences of preventing XSS. Grafana is heavily customizable, every time we do something like this we break some legitimate stuff so we have to be aware that we're doing it and communicate it in a good way. On top of that I also echo @aknuds1 concerns wrt adding another library (and expand a little on that, since I've just spent some time reviewing the added library 😅).

Setting a default that applies XSS protection that we didn't have before that's not really compatible with what we had before is a breaking change. This is likely to break a number of people's custom panels and things like that.

When it comes to security, I think it's reasonable to apply a higher threshold to what is a tolerable and an intolerable breaking change, but I think we should have two configurations for CSP, one for enabled or disabled and one for what CSP header to send. This is because we're unable to convey the difference between "no configuration" and "default configuration".

That'd be done by adding back the original (but with another name).

```ini
content_security_policy = true
```

I think we might have to have it default to false for one or two releases so that we can try it out in production and see what consequences it might have to enable it globally. Breaking our user's Grafana instances is the best way of having people avoiding upgrading Grafana which'll make it harder for us to apply these changes in the future.

>> The new middleware does more than just enable us to use CSP though, i.e. it replaces old logic in Grafana, so it represents a risk wrt. potentially introducing regressions. Can't you replicate the new middleware's CSP logic (producing nonce etc) without it?
> I can, but I'm not sure why if this package manage it so well.

Dependencies are always security risks in themselves. Reviewing this library it looks legitimate and actively developed by people with a genuine interest.

While reviewing the library this is what it does with CSP:

1. It allows you to have CSP enabled or disabled with or without nonces.
2. The library generates a 16 byte (128 bits,  random string for the nonce using the standard library's `crypto/rand.Reader()`, that's great, but also trivial. However, the library panics if the nonce generation fails on [csp.go:40](https://github.com/unrolled/secure/blob/368ad3ae8f053cb99f4fd4c73987d191672082b3/csp.go#L40). Having libraries panicking randomly is _not_ ok. That's a huge minus, because now we're going to have to recover any panic the library throws.
3. The nonce is stored in the context of the request. :+1:
4. It translates `$NONCE` to `'nonce-%[1]s'`
5. The response header is written, either by directly copying the CSP header we're providing in options or by using `Sprintf` to add the aforementioned nonce into the CSP header.

These operations are all sensible, but the value-add of having a library doing it compared to implementing it directly is not sufficiently high imo.

The heavy part of implementing browser security for us is to stay compatible with plugins and what our users are using Grafana for, _not_ the technical implementation of the headers. To that end, this library removes fine-grained control and makes us subject to changes being made unbeknown to us (or having to keep vigilant track of the upstream). That might've been worth it if this library had an exceptionally nice API, but it's a pretty raw mapping from a Go struct to HTTP headers and that's just abstractions.

sakjur:
Adding to that: This will need some discussion and my recent comment are my thoughts on that discussion, not a review of the PR or a request for changes 🙂

Sytten:
Big fan of CSP and I originally requested that it be added. That being said, there totally needs to be a way to customize it via the configuration if people add frontend and/or backend plugins.

torkelo (commented):
Think this looks good

torkelo:
Adding additional TODO: this might need changes to webpack chunk loading and plugins loading

aknuds1:
Sorry for the long wait, but trying to resolve this again.

aknuds1:
Closing this PR since it requires too much reworking to be acceptable. I have made a [draft PR](https://github.com/grafana/grafana/pull/29740) instead, with the changes we would like. I'll make sure to credit @iOrcohen when merging that.

PR #25844, opened 2020-06-25
Fix: preventing the click outside wrapper from throwing errors due to listening to unmounted nodes.

### LINKED ISSUE #25749: Dashboard time range picker throws "Minified React error"
**What happened**:
This is thrown on each key stroke:
```
react-dom.production.min.js:52 Uncaught Error: Minified React error #188; visit https://reactjs.org/docs/error-decoder.html?invariant=188 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at react-dom.production.min.js:52
    at ot (react-dom.production.min.js:53)
    at Object.findDOMNode (react-dom.production.min.js:286)
    at n.onOutsideClick (ClickOutsideWrapper.tsx:43)
```

No error thrown.

Open a dashboard's absolute time range form and type into the "from" or "to" fields. It doesn't matter whether there're any panels or not.

This error causes problems with E2E tests.

**Environment**:
- Grafana version: master branch (7.1)
- Data source type & version: any data source, I think
- OS Grafana is installed on: nightly-master.plugins.grafana.net
- User OS & Browser: macOS 10.15.5, Chrome 83
- Grafana plugins:
- Others:

mckn:
Looks like we have a bug where we adds another event listener instead of removing it on unmount. This causes a lot of errors to appear in the console when triggering key events inside a ClickOutsideWrapper component.

Fixes https://github.com/grafana/grafana/issues/25749

I will also have a look if I can prevent the other error that @torkelo managed to trigger in a separate PR.

### DIFF
--- a/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
+++ b/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
@@ -35,7 +35,7 @@ export class ClickOutsideWrapper extends PureComponent<Props, State> {
   componentWillUnmount() {
     window.removeEventListener('click', this.onOutsideClick, false);
     if (this.props.includeButtonPress) {
-      window.addEventListener('keyup', this.onOutsideClick, false);
+      window.removeEventListener('keyup', this.onOutsideClick, false);
     }
   }
 


torkelo (approved):
LGTM

stevenvachon:
Thanks!

PR #25849, opened 2020-06-25
Azure: Application Insights metrics to Frame and support multiple query dimensions

kylebrandt:
- The Application Insights Service now returns a dataframe. This is a "wide" formatted dataframe with a single time index.
- Multiple "group by" dimensions may now be selected instead of just one with Application Insights.
- Some types are copied / slightly altered from the Azure Go SDK but that SDK is not imported at this time.


kylebrandt:
@ryantxu Can you change the frontend such that Dimensions allows you to select multiple instead of just one, and the dimensions property sent to the backend is an array of strings, where each string is dimension. If there are no dimensions, then an empty array should be sent.

pkg/tsdb/azuremonitor/applicationinsights-datasource.go
```diff
@@ -109,10 +110,13 @@ func (e *ApplicationInsightsDatasource) buildQueries(queries []*tsdb.Query, time
 		}
 		params.Add("aggregation", insightsJSONModel.Aggregation)
 
+		dims := []string{}
+
 		dimension := strings.TrimSpace(insightsJSONModel.Dimension)
```
kylebrandt, line 115: This part will be updated once the frontend sends dimensions as a string array

kylebrandt: as well as type insightsJSONQuery in types.go

ryantxu:
I updated the UI to send an array for `dimensions` it gets a bit hacky, and I am sure there is a better way :(

pkg/tsdb/azuremonitor/applicationinsights-datasource.go
```diff
@@ -24,25 +24,27 @@ import (
...
-// ApplicationInsightsDatasource calls the application insights query API's
+// ApplicationInsightsDatasource calls the application insights query API.
 type ApplicationInsightsDatasource struct {
 	httpClient *http.Client
 	dsInfo     *models.DataSource
 }
 
+// ApplicationInsightsQuery is the model that holds the information
+// need to make a metrics query to Application Insights, and information
+// used to parse the response.
```
aknuds1, line 35:
```suggestion
// ApplicationInsightsQuery is the model that holds the information
// needed to make a metrics query to Application Insights, and information
// used to parse the response.
```

pkg/tsdb/azuremonitor/applicationinsights-datasource_test.go
```diff
@@ -290,3 +210,21 @@ func TestAppInsightsPluginRoutes(t *testing.T) {
 	}
 
 }
+
+func TestInsightsDimensionsUnmarshalJSON(t *testing.T) {
+	var err error
+	a := []byte(`"foo"`)
+	b := []byte(`["foo"]`)
+	_ = b
```
aknuds1, line 218: What's the point of this? To stop the compiler from complaining about `b` being unused?

pkg/tsdb/azuremonitor/applicationinsights-metrics.go
```diff
@@ -0,0 +1,306 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/grafana/grafana-plugin-sdk-go/data"
+)
+
+// InsightsMetricsResultToFrame converts a MetricsResult (a Application Insights metrics query response) to a dataframe.
```
aknuds1, line 11:
```suggestion
// InsightsMetricsResultToFrame converts a MetricsResult (an Application Insights metrics query response) to a dataframe.
```

pkg/tsdb/azuremonitor/applicationinsights-metrics.go
```diff
@@ -0,0 +1,306 @@
...
+// Due to the dynamic nature of the MetricsResult object, the name of the metric, aggregation, and
+// and requested dimensions are used to determine the expected shape of the object.
+// This builds all series into a single data.Frame with one time index (a wide formatted time series frame).
```
aknuds1, line 14:
```suggestion
// Due to the dynamic nature of the MetricsResult object, the name of the metric, aggregation,
// and requested dimensions are used to determine the expected shape of the object.
// This builds all series into a single data.Frame with one time index (a wide formatted time series frame).
```

pkg/tsdb/azuremonitor/applicationinsights-metrics.go
```diff
@@ -0,0 +1,306 @@
...
+				dimension := dimensions[dimLen-1]
+				dimVal, err := dimValueFromAP(s.AdditionalProperties, dimension)
+				if err != nil {
+					return err
+				}
+				labels[dimension] = dimVal
+			}
+
+			if _, ok := fieldIdxMap[labels.String()]; !ok {
+				// When we find a new combintation of labels for the metric, a new Field is appended.
```
aknuds1, line 54:
```suggestion
				// When we find a new combination of labels for the metric, a new Field is appended.
```

pkg/tsdb/azuremonitor/applicationinsights-metrics.go
```diff
@@ -0,0 +1,306 @@
...
+	AdditionalProperties map[string]interface{} `json:""`
+	// Start - Start time of the metric segment (only when an interval was specified).
+	Start time.Time `json:"start,omitempty"`
+	// End - Start time of the metric segment (only when an interval was specified).
+	End time.Time `json:"end,omitempty"`
+	// Segments - Segmented metric data (if further segmented).
+	Segments *[]MetricsSegmentInfo `json:"segments,omitempty"`
+}
+
+// UnmarshalJSON is the custom unmarshaler for MetricsResultInfo struct.
```
aknuds1, line 189:
```suggestion
// UnmarshalJSON is the custom unmarshaler for MetricsSegmentInfo struct.
```

pkg/tsdb/azuremonitor/types.go
```diff
@@ -127,3 +130,41 @@ type logJSONQuery struct {
...
+			return err
+		}
+		*s = InsightsDimensions(sa)
+		return nil
+	}
+
+	var str string
+	err := json.Unmarshal(data, &str)
+	if err != nil {
+		return fmt.Errorf("could not parse as string or array: %v", s)
```
aknuds1, line 162: This makes more sense to me (doesn't make sense to output `s` in the error?):
```suggestion
		return fmt.Errorf("could not parse %q as string or array: %w", string(data), err)
```

aknuds1 (changes requested):
Looks mostly good to me, but found some spelling errors and things I'm wondering about.

pkg/tsdb/azuremonitor/applicationinsights-datasource_test.go
kylebrandt: Oopps, yes, will remove. Was while I was messing with the func (same reason I made a var err error)`.

aknuds1 (approved):
Looks good AFAICT

PR #25852, opened 2020-06-25
Docs: Add a query variable topic

oddlittlebird:
This is the first of several PRs to expand and improve the templates and variables documentation. Add a query variable will be the template I base the other add <type> variable tasks on. 

I also moved several long chunks of content out of the main Templates and variables topic into separate reference topics, which are linked where appropriate.

**Special notes for your reviewer**: Two options in Add a query variable need more explanation. There are notes in ALL CAPS to call attention to them.


docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -0,0 +1,131 @@
+# Filter variables with regex
+
+Using the Regex Query option, you filter the list of options returned by the variable query or modify the options returned.
+
+This page shows #### Using regex to filter/modify values in the Variable dropdown
```
daniellee, line 5: This looks weird

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+weight = 500
++++
+
+# Formatting multi-value variables
+
+Interpolating a variable with multiple values selected is tricky as it is not straight forward how to format the multiple values into a string that
+is valid in the given context where the variable is used. Grafana tries to solve this by allowing each data source plugin to
+inform the templating interpolation engine what format to use for multiple values.
+
+> **Note:** The **Custom all value** option on the variable ,ust be blank for Grafana to format all values into a single string.
```
daniellee, line 14: This is ust -> must. What this means is that if leave it blank then the Grafana will concatenate (add together) all the values together in the query. Something like `value1,value2,value3`. If a custom all value is used then instead it will be something like `*` or `all`.

daniellee (dismissed):
There is a lot of text that could be improved - still too hard to understand but as this is mostly a reorganization of existing text, I think that is better covered in a different PR in the future. I added some comments about the missing sections.

oddlittlebird:
@torkelo and @marefr, @daniellee is on vacation. Could you take a look at this PR?

marefr:
Yeah when I get some time over.

docs/sources/variables/add-query-variable.md
```diff
@@ -0,0 +1,99 @@
...
+   - **Variable -** No variable dropdown is displayed on the dashboard.
+
+## Enter Query Options
+
+1. In the **Data source** list, select the target data source for the query. For more information about data sources, refer to [Add a data source]({{< relref "../features/datasources/add-a-data-source.md" >}}).
+1. In the **Refresh** list, select when the variable should update options.
+   - **Never -** Variables queries are cached and values are not updated. This is fine if the values never change, but problematic if they are dynamic and change a lot.
+   - **On Dashboard Load -** Queries the data source every time the dashboard loads. This slows down dashboard loading, because the variable query needs to be completed before dashboard can be initialized. 
+   - **On Time Range Change -** Queries the data source when the dashboard time range changes. Only use this option if your variable options query contains a time range filter or is dependent on the dashboard time range.
+1. In the **Query** field, enter a query. If you need more room, hover your cursor over the lines in the lower right corner of the field and drag downward to expand.
```
marefr, line 44: Some data sources may have custom query variable editors that is not only a single input field. See for example [Google Cloud Monitoring](https://grafana.com/docs/grafana/latest/features/datasources/cloudmonitoring/#query-variable) that have drop downs etc. Maybe just needs a note that query field may differ depending on data source.

docs/sources/variables/add-query-variable.md
```diff
@@ -0,0 +1,99 @@
...
+## Query expressions
+
+Query expressions are different for each data source. For more information, refer to the documentation for your data source:
+
+- [Graphite templating queries]({{< relref "../features/datasources/graphite.md#templating" >}})
+- [Elasticsearch templating queries]({{< relref "../features/datasources/elasticsearch.md#templating" >}})
+- [InfluxDB templating queries]({{< relref "../features/datasources/influxdb.md#templating" >}})
+- [Prometheus templating queries]({{< relref "../features/datasources/prometheus.md#templating" >}})
+- [OpenTSDB templating queries]({{< relref "../features/datasources/opentsdb.md#templating" >}})
+- [AzureMonitor templating queries]({{< relref "../features/datasources/azuremonitor.md#templating-with-variables-for-the-azure-monitor-service" >}})
```
marefr, line 23: Should we link templating sections for all data sources here or why only a few? I guess due to historic reason, so maybe something that can be fixed in a separate PR?

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -0,0 +1,131 @@
...
+demo.robustperception.io:9090
+demo.robustperception.io:9093
+demo.robustperception.io:9100
+```examples of filtering on the following list of options:
+
+```text
+backend_01
+backend_02
+backend_03
+backend_04
```
marefr, line 76: This looks weird.

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -0,0 +1,131 @@
+# Filter variables with regex
+
+Using the Regex Query option, you filter the list of options returned by the variable query or modify the options returned.
+
+This page shows how to use regex to filter/modify values in the variable dropdown
+
+Using the Regex Query Option, you filter the list of options returned by the Variable query or modify the options returned.
```
marefr, line 7: A link to JavaScript regex or something could be useful I think. Not sure if https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions is a good source. Maybe @torkelo have any input on this?

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+type = "docs"
+[menu.docs]
+weight = 500
++++
+
+# Formatting multi-value variables
+
+Interpolating a variable with multiple values selected is tricky as it is not straight forward how to format the multiple values into a string that
+is valid in the given context where the variable is used. Grafana tries to solve this by allowing each data source plugin to
+inform the templating interpolation engine what format to use for multiple values.
```
marefr, line 12:
```suggestion
Interpolating a variable with multiple values selected is tricky as it is not straight forward how to format the multiple values into a string that is valid in the given context where the variable is used. Grafana tries to solve this by allowing each data source plugin to inform the templating interpolation engine what format to use for multiple values.
```

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+> **Note:** The **Custom all value** option on the variable must be blank for Grafana to format all values into a single string. If leave it blank, then the Grafana concatenates (adds together) all the values in the query. Something like `value1,value2,value3`. If a custom `all` value is used, then instead the value will be something like `*` or `all`.
+
+## Multi-value variables with a Graphite data source
+
+Graphite uses glob expressions. A variable with multiple values would, in this case, be interpolated as `{host1,host2,host3}` if
+the current variable value was *host1*, *host2*, and *host3*.
```
marefr, line 19:
```suggestion
Graphite uses glob expressions. A variable with multiple values would, in this case, be interpolated as `{host1,host2,host3}` if the current variable value was *host1*, *host2*, and *host3*.
```

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Multi-value variables with a Prometheus or InfluxDB data source
+
+InfluxDB and Prometheus use regex expressions, so the same variable
+would be interpolated as `(host1|host2|host3)`. Every value would also be regex escaped if not, a value with a regex control character would
+break the regex expression.
```
marefr, line 25:
```suggestion
InfluxDB and Prometheus use regex expressions, so the same variable would be interpolated as `(host1|host2|host3)`. Every value would also be regex escaped. If not, a value with a regex control character would break the regex expression.
```

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Multi-value variables with an Elastic data source
+
+Elasticsearch uses lucene query syntax, so the same variable would be formatted as `("host1" OR "host2" OR "host3")`. In this case, every value
+must be escaped so that the value only contains lucene control words and quotation marks.
```
marefr, line 30:
```suggestion
Elasticsearch uses lucene query syntax, so the same variable would be formatted as `("host1" OR "host2" OR "host3")`. In this case, every value must be escaped so that the value only contains lucene control words and quotation marks.
```

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Formatting troubles
+
+Automatic escaping and formatting can cause problems and it can be tricky to grasp the logic is behind it.
+Especially for InfluxDB and Prometheus where the use of regex syntax requires that the variable is used in regex operator context.
```
marefr, line 35:
```suggestion
Automatic escaping and formatting can cause problems and it can be tricky to grasp the logic behind it. Especially for InfluxDB and Prometheus where the use of regex syntax requires that the variable is used in regex operator context.
```

docs/sources/variables/formatting-multi-value-variables.md
```diff
@@ -0,0 +1,37 @@
...
+
+If you do not want Grafana to do this automatic regex escaping and formatting, then you must turn off the **Multi-value** or **Include All option** options.
```
marefr, line 37: Or I guess use [the raw variable format](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/#raw).

docs/sources/variables/templates-and-variables.md
```diff
@@ -54,196 +68,48 @@ documentation topic for details on value escaping during interpolation.
...
-*Refresh* | Controls when to update the variable option list (values in the dropdown). **On Dashboard Load** will slow down dashboard load as the variable query needs to be completed before dashboard can be initialized. Set this only to **On Time Range Change** if your variable options query contains a time range filter or is dependent on dashboard time range.
-*Query* | The data source specific query expression.
-*Regex* | Regex to filter or capture specific parts of the names return by your data source query. Optional.
-*Sort* | Define sort order for options in dropdown. **Disabled** means that the order of options returned by your data source query will be used.
-
-#### Using regex to filter/modify values in the Variable dropdown
+### Query
 
-Using the Regex Query Option, you filter the list of options returned by the Variable query or modify the options returned.
+Query variables allow you to write a data source query that might return a list of metric names, tag values, keys, server names, sensor IDs, or data centers.
```
marefr, line 79: Could basically return anything so would add etc?
```suggestion
Query variables allow you to write a data source query that might return a list of metric names, tag values, keys, server names, sensor IDs, or data centers etc.
```

docs/sources/variables/templates-and-variables.md
```diff
@@ -54,196 +68,48 @@ documentation topic for details on value escaping during interpolation.
...
-demo.robustperception.io:9093
-demo.robustperception.io:9100
-```
+### Data source
 
-### Query expressions
+Data source variables allow you to quickly change the data source for an entire dashboard. Useful if you have multiple instances of a data source in for example different environments.
 
-The query expressions are different for each data source.
+### Interval
```
marefr, line 99: Why draft of interval page, seems like this has the same content as the new page?

marefr (changes requested):
First round of review. Was a lot of changes. Let's start with this.

docs/sources/variables/filter-variables-with-regex.md
oddlittlebird: Check the current version. The earlier version had a formatting error.

docs/sources/variables/filter-variables-with-regex.md
oddlittlebird: I added this link. Please let me know if we need a different one.

docs/sources/variables/templates-and-variables.md
oddlittlebird: Because once I get the initial template figured out, there will be a separate task topic for each variable type.

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -0,0 +1,131 @@
...
+```regex
+/.*instance="([^"]*).*/
+```
+
+Result:
+
+```text
+demo.robustperception.io:9090
+demo.robustperception.io:9093
+demo.robustperception.io:9100
```
marefr, line 130: To me this feels like a repetition of content already written above this?

````suggestion
```
````

docs/sources/variables/formatting-multi-value-variables.md
marefr: Want to commit theis suggestions?

docs/sources/variables/formatting-multi-value-variables.md
marefr: Want to commit theis suggestions?

marefr (changes requested):
Some minor things left

marefr (approved):
LGTM but please merge with latest master to resolve build problems

PR #25853, opened 2020-06-26
WIP: ValueMappings: improve config UI and allow string inputs

ryantxu:
The current value mapping UI is difficult when doing anything complex:
* ValueMap should support strings (currently just numbers)
* hard to understand what is mapped when many items are entered
* rules should be sortable since they are evaluated in their sort order

image

For reference, things used to look kinda like this:
image

TODO:

- [x] Actually update the models
- [ ] React.FC vs PureComponent?
- [x] remove operator?
- [ ] id?  is that just the index?
- [ ] grid/flex sizing vs absolute?
- [ ] fix tests


dprokop:
String values mapping will be fixed with https://github.com/grafana/grafana/pull/25929

packages/grafana-data/src/types/valueMapping.ts
```diff
@@ -4,9 +4,9 @@ export enum MappingType {
 }
 
 interface BaseMap {
-  id: number;
-  operator: string;
-  text: string;
+  id: number; // Could/should be derived from the array index!!!
+  // operator: string; // ???????????????
```
dprokop, line 8: If TS does not complain then I think we can remove this

packages/grafana-ui/src/components/ValueMappingsEditor/RangeMappingRow.tsx
```diff
@@ -0,0 +1,73 @@
...
+  dragClass: string;
+  mapping: RangeMap;
+  index: number;
+  onChange: (index: number, mapping: ValueMapping) => void;
+}
+
+export class RangeMapRow extends PureComponent<RangeMapProps> {
+  constructor(props: RangeMapProps) {
+    super(props);
+  }
```
dprokop, line 15: This kind of constructors are not necessary :)

packages/grafana-ui/src/components/ValueMappingsEditor/ValueMappingRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+  dragClass: string;
+  mapping: ValueMap;
+  index: number;
+  onChange: (index: number, mapping: ValueMapping) => void;
+}
+
+export class ValueMapRow extends PureComponent<ValueMapProps> {
+  constructor(props: ValueMapProps) {
+    super(props);
+  }
```
dprokop, line 15:
```suggestion
```

torkelo:
This needs another UI update. This is just too inconsistent from the rest of the sidebar introducing a new mix of UI patterns, not following normal inline forms style nor the new forms. 

It's tricky, do not know what the solution is given the width constraints. But do not like the mix of inline forms labels and input prefixes (that causes misalignment). 

Screenshot from 2020-07-01 10-02-46

One idea is to use the collapsable row (that we use in transforms and query editors) to show a collapsed form of the mapping and expand it to show a full form

dprokop:
@torkelo I agree. I like the idea with the collapsible row too, gonna experiment with this.

packages/grafana-ui/src/components/ValueMappingsEditor/DraggableMappingRow.tsx
```diff
@@ -0,0 +1,40 @@
...
+        </div>
+        {children}
+      </>
+    );
+  }
+);
+const getDraggableRowStyles = (theme?: GrafanaTheme) => {
+  if (!theme) {
+    theme = {} as GrafanaTheme;
+  }
```
ryantxu, line 28: @dprokop -- obviously something funky here.  this was required for typings

`useTheme()` makes an optional theme?

ryantxu:
Collapsible is interesting -- it would require many clicks when trying to make a long list consistent though.  I was messing with trying to get flexbox to stack when there is not enough space to show in a single row... 

image

packages/grafana-ui/src/components/ValueMappingsEditor/DraggableMappingRow.tsx
dprokop: uhm, must have missed that. No need for that at all, useTheme assures theme object.

PR #25854, opened 2020-06-26
Chore: Start harmonizing linting with plugin SDK

aknuds1:
Start harmonizing with the stricter linting scheme of the plugin SDK. Fix issues arising from enabling a couple more linters, a lot of which are just due to files not being formatted with `goimports`.

Also moving golangci-lint configuration files to [TOML](https://github.com/toml-lang/toml) format, since we use that for the SDK and it's a less troublesome (i.e., far less ambiguous) language than YAML.


ryantxu (approved):

jessabe (approved):

pkg/infra/usagestats/usage_stats.go
```diff
@@ -181,9 +181,11 @@ func (uss *UsageStatsService) sendUsageStats(oauthProviders map[string]bool) {
 
 	client := http.Client{Timeout: 5 * time.Second}
 	go func() {
-		if _, err := client.Post(usageStatsURL, "application/json", data); err != nil {
+		resp, err := client.Post(usageStatsURL, "application/json", data)
+		if err != nil {
 			metricsLogger.Error("Failed to send usage stats", "err", err)
 		}
+		defer resp.Body.Close()
```
marefr, line 188: Defer needed here?

pkg/api/team_test.go
```diff
@@ -9,11 +9,12 @@ import (
 	"github.com/grafana/grafana/pkg/components/simplejson"
 	"github.com/grafana/grafana/pkg/models"
 
+	"net/http"
```
marefr, line 12: Not sure what the recommendation/standard is, but I have something is VS code that sorts imports when saving a file creating something like this:
```go
// ordered set of builtin Go packages

// ordered set of external Go packages
```

pkg/models/datasource_cache_test.go
```diff
@@ -224,13 +224,10 @@ func TestDataSourceProxyCache(t *testing.T) {
 			// 3. Send test request which should have the Authorization header set
 			req := httptest.NewRequest("GET", backend.URL+"/test-headers", nil)
 			res, err := transport.RoundTrip(req)
-			if err != nil {
-				log.Fatal(err.Error())
-			}
+			So(err, ShouldBeNil)
+			t.Cleanup(func() { res.Body.Close() })
```
marefr, line 228: Nit, defer should do it here? Already using defer for `backend.Close()` above

marefr (changes requested):
Great. Some minor comments/nits.

pkg/api/team_test.go
aknuds1: @marefr I don't have any control of it, I think it's what the `goimports` formatter does automatically.

pkg/infra/usagestats/usage_stats.go
aknuds1: Good catch, I also noticed you shouldn't try to close here if an error was returned.

kylebrandt:
I think `nakedret`'s default value of 5 is too aggressive:

"nakedret is a Go static analysis tool to find naked returns in functions greater than a specified function length." ... " If not specified, it defaults to 5."

That being said, I'm not to fond of this at all. I don't think it is going to make a big difference in code quality (named params in returns are rare anyways) and more likely just to be a annoying this randomly run into with PRs.

aknuds1:
What do you suggest regarding nakedret @kylebrandt? Disabling it? I have too little experience with it to say. It caught very little in Grafana at least, so not very annoying so far.

kylebrandt:
> What do you suggest regarding nakedret @kylebrandt? Disabling it? I have too little experience with it to say. It caught very little in Grafana at least, so not very annoying so far.

Yes disabling it. It catching little is kind of my point - that it will happen infrequently enough that it might feel like "random build failures", but not frequently enough that it is the sort of thing people will remember, or frequently enough to really change the code quality in a significant way.

Or to put it another way - I think that linter it is too nitpicky for this codebase :-)

aknuds1:
@kylebrandt OK I can disable nakedret.

kylebrandt (approved):

aknuds1:
@marefr Please have another look.

bergquist (approved):
LGTM

marefr (approved):
LGTM

PR #25857, opened 2020-06-26
Build: fixing failed master build adding missing named exports for lodash.

mckn:
When trying to publish the grafana-ui the rollup will fail due to missing named exports for lodash. This PR will resolve those issues.


dprokop (approved):
lovely!

PR #25858, opened 2020-06-26
Stackdriver: Deep linking from Grafana panels to the Metrics Explorer

### LINKED ISSUE #25669: Stackdriver: Deep linking from Grafana panels to the Metrics Explorer
Add the ability to link from a graph in Grafana and open the same query in Metric Explorer in the Stackdriver UI

papagian:
A new link `View in Metrics Exploer` is added in the graph panel which opens the Metrics explorer in the Monitoring and runs the query from the Grafana panel there.
cloud-moniitoring-deep-linking

Setup:
- Credentials can be found in the [private plugin-provisioning repo](https://github.com/grafana/plugin-provisioning/blob/master/plugins/stackdriver/provisioning/datasources/stackdriver.yml)
- Provision the YAML by placing it in the conf/provisioning/datasources and restarting your Grafana instance
- [Sample](https://gist.github.com/papagian/69a5fe2c1864204251e1c8c5684966af) dashboard

Fixes #25669

The link is only available for metrics queries (not for SLOs).
The user should be logged with a google account (otherwise is navigated to the login page first) that has access to the GCP project specified in the query (otherwise gets a permissions error).

A complementary PR adding a new image for the documentation will follow up.


docs/sources/features/datasources/stackdriver.md
```diff
@@ -323,3 +323,12 @@ datasources:
     jsonData:
       authenticationType: gce
 ```
+
+## Deep linking from Grafana panels to the Metrics Explorer in Gooogle Monitoring
```
marefr, line 327: Not sure if cloud should be included? 
```suggestion
## Deep linking from Grafana panels to the Metrics Explorer in Gooogle Cloud Monitoring
```

docs/sources/features/datasources/stackdriver.md
```diff
@@ -323,3 +323,12 @@ datasources:
...
+
+Only available in Grafana v7.1+.
+
+{{< docs-imagebox img="/img/docs/v71/cloud-moniitoring-deep-linking.png" max-width="500px" class="docs-image--right" caption="Google Cloud Monitoring deep linking" >}}
+
+Click on a time series in the panel to see a context menu with a link to View in Monitoring Metrics explorer. Clicking that link opens the Metrics explorer in the Monitoring and runs the query from the Grafana panel there.
```
marefr, line 333: Not sure about this either.
```suggestion
Click on a time series in the panel to see a context menu with a link to View in Cloud Monitoring Metrics explorer. Clicking that link opens the Metrics explorer in the Cloud Monitoring and runs the query from the Grafana panel there.
```

docs/sources/features/datasources/stackdriver.md
```diff
@@ -323,3 +323,12 @@ datasources:
...
+If you're not currently logged in to the Google console, then the link opens the login page. The provided link is valid for any account, but it only displays the query if your account has access to the GCP project specified in the query.
```
marefr, line 334: Maybe a note about deep linking for SLO queries not supported?

marefr (changes requested):
Great. Haven't tested it a lot, but expect you to do more of that. Added a couple comment/suggestions.  Please open website PR with image update so we can test docs before merging this .

Was a bit hard just clicking on link when you have multiple google accounts and it selected the wrong one when clicking on link. Had to copy it and append `&authuser=1` :smile:

papagian:
> Was a bit hard just clicking on link when you have multiple google accounts and it selected the wrong one when clicking on link. Had to copy it and append `&authuser=1` 😄

so do you suggest to include this in the url?
if we do, the user somehow should configure this; I don't think that selecting always the primary account would work for everyone.

docs/sources/features/datasources/stackdriver.md
papagian: I'm not completely sure about this; in the google console is referred only as Monitoring.

papagian:
> Please open website PR with image update so we can test docs before merging this .

https://github.com/grafana/website/pull/1724

docs/sources/features/datasources/stackdriver.md
```diff
@@ -323,3 +323,14 @@ datasources:
...
       authenticationType: gce
 ```
+
+## Deep linking from Grafana panels to the Metrics Explorer in Gooogle Monitoring
+
+Only available in Grafana v7.1+.
+
+{{< docs-imagebox img="/img/docs/v71/cloud-moniitoring-deep-linking.png" max-width="500px" class="docs-image--right" caption="Google Cloud Monitoring deep linking" >}}
+
+> Available only for Metric queries
```
oddlittlebird, line 333:
```suggestion
> **Note:** This feature is only available only for Metric queries.
```

daniellee (approved):
In the Metrics Explorer, when you create a new query, the first value you choose is the resource type. A list of resource types is not available in the public API and hence, we had to build it a different way in Grafana (Service + metric is the first choice). This means you often get this error message about a resource not have being selected:

image

If the user adds resource type to their filter then this error disappears:
image

Not sure if we can do anything about this. There could theoretically be multiple resource types so adding it automatically to the query is probably not a good idea. Maybe can mention it in the docs.

Otherwise, everything seems to work as expected. (Though I also have the problem with multiple accounts which I also have with Google Meet, Google Docs etc.)

docs/sources/features/datasources/stackdriver.md
marefr: That's fine. Probably me not understanding that you're referring to the console with "Gooogle Monitoring".

papagian:
> > Was a bit hard just clicking on link when you have multiple google accounts and it selected the wrong one when clicking on link. Had to copy it and append `&authuser=1` 😄
> 
> so do you suggest to include this in the url?
> if we do, the user somehow should configure this; I don't think that selecting always the primary account would work for everyone.

Following [this](https://github.com/grafana/grafana/pull/25874#issuecomment-651168276) comment's suggestion I found [this post](https://github.com/grafana/grafana/pull/25874#issuecomment-651168276) and I ended up with the solution to direct first to the AccountChooser (the `prompt` parameter - suggested in the post discussion - didn't work in this case because I suspect that it is only supported in OAuth2 authorization URLs).

papagian:
> In the Metrics Explorer, when you create a new query, the first value you choose is the resource type. A list of resource types is not available in the public API and hence, we had to build it a different way in Grafana (Service + metric is the first choice). This means you often get this error message about a resource not have being selected:
> 
> Not sure if we can do anything about this. There could theoretically be multiple resource types so adding it automatically to the query is probably not a good idea. Maybe can mention it in the docs.
> 

The resource type information for a given metric type is included in the Monitoring API response. Given a metric type having multiple resource types is considered a rare case we can enforce the first resource type found in the response to be included in the filter parameter (if no resource type is already provided in the query filters).

marefr (approved):
LGTM

Docs look good. It works good. 

It's a little bit weird that I have to go through the account chooser every time, but I can live with that since it's a much better experience than before. However, I wonder what the behavior is if you only have 1 google account - will it redirect directly to target link? Tried to find documentation about google account chooser, but came up empty, where did you find documentation?

papagian:
> It's a little bit weird that I have to go through the account chooser every time, but I can live with that since it's a much better experience than before. However, I wonder what the behavior is if you only have 1 google account - will it redirect directly to target link?

No, you still pass first from the account chooser. Another option would be to add a new configuration and only if it's enabled to enforce account chooser.

> Tried to find documentation about google account chooser, but came up empty, where did you find documentation?

There is a reference [here](https://developers.google.com/identity/toolkit/web/account-chooser#the_account_chooser) but I read about it in the stackoverflow post mentioned above.

daniellee (approved):
I tested with one account and it was smooth. Just one extra click that went really fast. Adding the resource type works well too.

LGTM

docs/sources/features/datasources/cloudmonitoring.md
```diff
@@ -324,3 +324,14 @@ datasources:
     jsonData:
       authenticationType: gce
 ```
+
+## Deep linking from Grafana panels to the Metrics Explorer in Gooogle Cloud Console
```
oddlittlebird, line 328:
```suggestion
## Deep linking from Grafana panels to the Metrics Explorer in Google Cloud Console
```

docs/sources/features/datasources/cloudmonitoring.md
```diff
@@ -324,3 +324,14 @@ datasources:
...
+
+Only available in Grafana v7.1+.
+
+{{< docs-imagebox img="/img/docs/v71/cloudmonitoring_deep_linking.png" max-width="500px" class="docs-image--right" caption="Google Cloud Monitoring deep linking" >}}
+
+> **Note:** This feature is only available for Metric queries.
+
+Click on a time series in the panel to see a context menu with a link to View in Metrics explorer in Google Cloud Console. Clicking that link opens the Metrics explorer in the Google Cloud Console and runs the query from the Grafana panel there.
+The link navigates the user first to the Google Account Chooser and after successfully selecting an account, the user is redirected to the metrics explorer. The provided link is valid for any account, but it only displays the query if your account has access to the GCP project specified in the query.
```
oddlittlebird, line 337: Inconsistent capitalization of "metrics explorer" here. If it is a proper noun (named in the UI), then it should be capitalized to match the UI. If it is not, then lowercase throughout. Either way, try to be consistent.

oddlittlebird (commented):
Got a couple suggestions, looks good otherwise.

docs/sources/features/datasources/cloudmonitoring.md
papagian, line 337:
```suggestion
Click on a time series in the panel to see a context menu with a link to View in Metrics Explorer in Google Cloud Console. Clicking that link opens the Metrics Explorer in the Google Cloud Console and runs the query from the Grafana panel there.
The link navigates the user first to the Google Account Chooser and after successfully selecting an account, the user is redirected to the Metrics Explorer. The provided link is valid for any account, but it only displays the query if your account has access to the GCP project specified in the query.
```

ericvolp12:
Hey, I've noticed that with a Stat panel type, there is no way to disable this form of deep linking and any click on the panel takes you to the deep link. We'd like to be able to link a stat to another dashboard in our Grafana for drill-downs but can't do that because every stat panel forcibly links to GCP and even with a Data Link configured for the panel, it gets overridden by the GCP deep link. I would love to just disable GCP deep links for a panel or even a whole dashboard if possible.

PR #24951, opened 2020-05-20
Chore: Toolkit fixes to use toolkit for release

srclosson:
These fixes were not merge earlier so that we could focus on the release. Not that is done, I'd like to get these fixes merged in.

They include:
- Utilities added to docker images
- Small change to allow grafana-toolkit to be run from a symbolic link 
- Fix for github publish in parsing errors with no 'data' object.

PR #25860, opened 2020-06-26
grafana/toolkit: fix binary path so toolkit is not run in linked mode when installed from npm

### LINKED ISSUE #25816: Plugin build fails when targeting canary
When I try to build a plugin with canary, it fails with a `tsconfig.conf` related issue.  I have seen the issue on a few plugins, here is the simplest way to reproduce:

https://github.com/grafana/simple-datasource/compare/with-canary

image

dprokop:
Fixes https://github.com/grafana/grafana/issues/25816

Problem was introduced by https://github.com/grafana/grafana/pull/24951. Made toolkit being executed in linked mode (meaning TS etc). The path to the js entry point was wrong, as there is no `dist` directory in the published version of toolkit. It's only used as a temporary directory for building the distribution.

We really need verification script for the canary to make sure the templates are working:0

### DIFF
--- a/packages/grafana-toolkit/bin/grafana-toolkit.js
+++ b/packages/grafana-toolkit/bin/grafana-toolkit.js
@@ -23,8 +23,9 @@ const isLinkedMode = () => {
 };
 
 const entrypoint = () => {
-  const resolvedJsDir = path.resolve(`${__dirname}/../dist/src/cli/index.js`);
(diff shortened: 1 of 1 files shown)


stevenvachon (approved):

stevenvachon:
Maybe add a quick `execa('node dist/bin/grafana-toolkit.js')` test to catch this for next time?

ryantxu (approved):

Issue #25445, opened 2020-06-08
Data Link - "Open in new tab" button status mismatch


mac3000:
**What happened**:
When adding Data Links to Graph, the "Open in new tab" button in the frame will always show "Off", regardless if you changed it to before.

**What you expected to happen**:
Button should reflect the actual status

**How to reproduce it (as minimally and precisely as possible)**:
1. Add a data link to a graph, and select to open in new Window. Apply, Save and confirm it works as expected (you can also Refresh page)
2. Go back to edit the Link, it will show Button in "Off" setting. If you change to "On" and test, it will open in same Window.

Basically, that button will always show "Off", regardless of what you changed it to before.

Grafana version: Grafana v7.0.0 (aee1438ff2)

Let me know if you want or need more details.

Thanks,
Jacob

torkelo:
thanks for reporting this

tskarhed:
DataLinkEditor could be refactored using the Form component instead. Should probably do that. I also noticed that it is in grafana-ui, which feels a little bit weird.

PR #25862, opened 2020-06-26
Switch: Deprecate checked prop in favor of value

tskarhed:
There is no implementation to support the checked value, which causes some weird behavior. To be consistent with our input components we prefer using the `value` prop instead. Checked prop is planned to be removed for 8.0.

 Long term fix for issues like #25445


Clarity-89 (approved):

jessover9000:
This is a code-sided change and doesn't have any UX implications, does it?

tskarhed:
@jessover9000 Yes, that's correct. You might get notifications for these kind of things, just if someone edits an MDX file.

jessover9000:
> @jessover9000 Yes, that's correct. You might get notifications for these kind of things, just if someone edits an MDX file.

Yes I know, most of the time I kind of get that it's not relevant for me but sometimes I'm just not sure :D

PR #9859, opened 2017-11-12
(prometheus) support POST for query and query_range

mtanda:
Support new POST endpoint. https://github.com/prometheus/prometheus/issues/3072

PR #7459, opened 2017-02-05
improve security of Prometheus datasource

mtanda:
I think it is better to Grafana proxy should block destructive request.
To acheive that, we should allow `GET` request and path should be `/api/.*`.
https://prometheus.io/docs/querying/api/

PR #25868, opened 2020-06-26
dsproxy: Allow Cortex DELETE/PUT/POST requests

s-h-a-d-o-w:

We need this in order to use Cortex' Ruler API.
To allow for that, @gotjosh recommended to simply remove this check and ask whether one of the original authors would object to this change.

### DIFF
--- a/pkg/api/pluginproxy/ds_proxy.go
+++ b/pkg/api/pluginproxy/ds_proxy.go
@@ -235,18 +235,6 @@ func (proxy *DataSourceProxy) validateRequest() error {
 		return errors.New("Target url is not a valid target")
 	}
 
-	if proxy.ds.Type == models.DS_PROMETHEUS {
-		if proxy.ctx.Req.Request.Method == "DELETE" {
-			return errors.New("Deletes not allowed on proxied Prometheus datasource")
-		}
-		if proxy.ctx.Req.Request.Method == "PUT" {
-			return errors.New("Puts not allowed on proxied Prometheus datasource")
-		}
-		if proxy.ctx.Req.Request.Method == "POST" && !(proxy.proxyPath == "api/v1/query" || proxy.proxyPath == "api/v1/query_range") {
-			return errors.New("Posts not allowed on proxied Prometheus datasource except on /query and /query_range")
-		}
-	}
-
 	if proxy.ds.Type == models.DS_ES {
 		if proxy.ctx.Req.Request.Method == "DELETE" {
 			return errors.New("Deletes not allowed on proxied Elasticsearch datasource")


gotjosh:
For context, the idea here is that the Prometheus datasource is used for any quasi-compatible Prometheus APIs in any system. In this particular case, Cortex uses this same datasource but supports additional endpoints under these verbs. 

We can either remove the check or do a filtering type of approach where we add the new routes we want to support. As @s-h-a-d-o-w mentioned we'd like to gather some insight on why do we have these checks in the first place.

mtanda:
To protect unexpected attack, I think Prometheus should check HTTP method and URL.
Worst case is losing all Prometheus data.
https://prometheus.io/docs/prometheus/latest/querying/api/#delete-series

We setup Prometheus with Grafana in 1 host, and allow local access from Grafana to Prometheus.
Current method check work like firewall for Prometheus.
If anyone who can access Grafana has privilege to remove Prometheus data, we need extra layer to protect the data. (e.g. setup nginx and check HTTP method, etc...)

I recommend to keep whitelist API in Grafana, and block others.

mtanda:
I found original commit.
https://github.com/grafana/grafana/pull/7459
I intend to limit to access config page like `http://demo.robustperception.io:9090/config` when Prometheus only support GET.

When Prometheus support POST, allow POST for specific endpoint.
https://github.com/grafana/grafana/pull/9859

s-h-a-d-o-w:
Thank you very much for the detailed explanation @mtanda !!
I suppose the way forward then is to add a whitelist for the specific Cortex end points that I need to use.

torkelo:
Not sure about this, this could have very serious security implications as it would open up these operations and api's to all users (even with viewer role)

s-h-a-d-o-w:
@torkelo Even with white-listing only the Ruler API end points? If so, then this sounds like it might be worthwhile discussing during an EasyStart Weekly?

gotjosh:
I agree with @s-h-a-d-o-w, I think a filtering approach would work. With the caveat that we use the user's role to limit these actions.

Interaction with the Cortex Ruler API has a limited set of endpoints, which are:

- Everything under `api/v1/rules/{namespace}/{group_name}`
- Or using the legacy prefix `rules/{namespace}{group_name}`

With the verbs POST, PUT, DELETE.

I'd say filtering to **only** allow these routes to users that are editors or admins would be an approach that can give us the flexibility that we need. What do you think @torkelo, @marefr and @mtanda?

marefr:
Please update PR with a filtering/whitelist approach so we can see what that look likes

torkelo:
We do not think a white list solves the problem either , this needs to use the app routes feature , because users will not want to use and admin token in their datasource setting that can modify things in their Cortex cluster.

Or we need to have special white lists and custom role checks just for these routes

s-h-a-d-o-w:
Are there docs on this app routes feature you're referring to?

s-h-a-d-o-w:
Ah, I guess you might mean this: https://grafana.com/docs/grafana/latest/developers/plugins/add-authentication-for-data-source-plugins/

torkelo:
> Ah, I guess you might mean this: https://grafana.com/docs/grafana/latest/developers/plugins/add-authentication-for-data-source-plugins/
yes

And we have appl plugin level routes as well that work the same way. 

but the biggest question mark for me is how are these API's protected in the cortex cluster or our hosted metrics API proxy?

Maybe we need to forward the Hosted Grafana auth cookie instead? But that will be problematic for any HG customer that have custom auth

marefr:
One idea I had yesterday was to use a Grafana.com API key in the app in a similar fashion to how Worldping works.

s-h-a-d-o-w:
@marefr I suppose [this](https://github.com/grafana/synthetic-monitoring-app/blob/master/src/datasource/plugin.json#L16) is the relevant part of the code where that key actually ends up being used?
If so - yeah, there's a bunch of stuff in that area I didn't realize, like also being able to use variables to target machines and so on.
If that ends up working similarly with the Ruler API, then there won't be a need to change the data source proxy.

marefr:
@s-h-a-d-o-w yes and [this](https://github.com/raintank/worldping-app/blob/master/src/plugin.json)

s-h-a-d-o-w:
We will use data source plugins to proxy calls to cortex services, so this won't be necessary any more.

PR #25889, opened 2020-06-29
Graphite: Fix tag value dropdowns not showing in query editor 

### LINKED ISSUE #25880: 7.0.4 : graphite tags have lost their values in the query editor
Hello,

I use tags with the graphite datasource.
Until v7.0.3, no problem, but with v7.0.4, the tag values are no longer shown on the query editor.

**What happened**:

Here is the same exact query in v7.0.3 and v7.0.4.

v7.0.3:
Sélection_084

v7.0.4 : 
Sélection_085

If I use the text editor mode, the full query is still there, so it's just a UI bug.

If I reinstall v7.0.3 with `sudo apt install grafana=7.0.3`, all is normal.

v7.0.3 behaviour is correct.

**Environment**:
- Grafana version: 7.0.4
- Data source type & version: graphite, v1.1.7
- OS Grafana is installed on: ubuntu 20.04
- User OS & Browser: chromium on ubuntu

torkelo:
Fixes #25880

This was caused by the jQuery Upgrade that breaks (changes) how jquery parses html templates, it no longer replaces self-closing tags with closing tags. 

The bug only happens if you have two self-closing tag angular directives after each other (as siblings). 

I searched code base and could not find any other case

### DIFF
--- a/public/app/plugins/datasource/graphite/partials/query.editor.html
+++ b/public/app/plugins/datasource/graphite/partials/query.editor.html
@@ -20,7 +20,8 @@
           css-class="query-segment-key"
           get-options="ctrl.getTags($index, $query)"
           on-change="ctrl.tagChanged(tag, $index)"
-        />
+        ></gf-form-dropdown>
+        
         <gf-form-dropdown
           model="tag.operator"
           label-mode="true"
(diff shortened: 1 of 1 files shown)


s-h-a-d-o-w:
Could you please provide an example query for `gdev-graphite`, so that I can confirm that this fix actually works?

torkelo:
Anything with a tag filter: 

seriesByTag('name=random-walk.count')

s-h-a-d-o-w (approved):
Thanks!
Can confirm that this solves the problem. :)

PR #25890, opened 2020-06-29
Explore: Unification of logs/metrics/traces user interface

aocenas:
Design doc: https://docs.google.com/document/d/1hZURlhvU5LVejMWXeQTaTMOLAC73FQjQR_EjgYtOV9c/edit#heading=h.sp7xjuf6unar


packages/grafana-data/src/types/data.ts
```diff
@@ -15,7 +15,7 @@ export enum LoadingState {
   Error = 'Error',
 }
 
-export type PreferredVisualisationType = 'graph' | 'table';
+export type PreferredVisualisationType = 'graph' | 'table' | 'logs' | 'trace';
```
aocenas, line 18: I think we can use this to figure out where to show the dataFrames. For data sources which won't use that we should just make sure the output is sensible (ie fallback to table table/graph)

public/app/features/explore/Explore.tsx
```diff
@@ -332,7 +338,8 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
               <SecondaryActions
                 addQueryRowButtonDisabled={isLive}
                 // We cannot show multiple traces at the same time right now so we do not show add query button.
-                addQueryRowButtonHidden={mode === ExploreMode.Tracing}
+                //TODO:unification
+                // addQueryRowButtonHidden={mode === ExploreMode.Tracing}
```
aocenas, line 342: I think we can just show one trace view per frame and deal with multiple traces that way.

aocenas:
We also need to check code like this one https://github.com/grafana/grafana/blob/72b8300571dc585041ffee6a35c82b79abeb686c/public/app/plugins/datasource/prometheus/datasource.ts#L221. It seems to be working but data sources should probably not be getting explore internal state and depend on that.

packages/grafana-data/src/types/data.ts
torkelo: :+1:

public/app/features/explore/utils/ResultProcessor.ts
```diff
@@ -16,23 +16,48 @@ import { getGraphSeriesModel } from 'app/plugins/panel/graph2/getGraphSeriesMode
...
+    for (const frame of this.dataFrames) {
+      if (isTimeSeries(frame, this.state.datasourceInstance?.meta.id)) {
+        if (shouldShowInVisualisationType(frame, 'graph')) {
+          this.graphFrames.push(frame);
+        }
+        if (shouldShowInVisualisationType(frame, 'table')) {
+          this.tableFrames.push(frame);
+        }
+      } else {
+        if (shouldShowInVisualisationType(frame, 'logs')) {
```
kaydelaney, line 43: Do we need to nest these if/else-if statements inside the else statement starting at line 42 or can we just use all else-if statements instead?

aocenas: the thing is time series by default are shown in both graph and table. So they need to be separate, other types are sort of priority based in the sense that when we think something should be logs we won't try to match this to anything else.

aocenas: Also this assumes we will start using the meta attributes so may not work at the moment for all cases, but as I mentioned if we can do some sensible default that should be fine.

kylebrandt:
For me this is a lot of JS :-) If I am to review (some concepts maybe?) I think I would need some context or a summary.

aocenas:
@kylebrandt seems like you ware added because the codeowners config and this touches some backend files. I think the automatic code owners assignment isn't that great in some cases,

public/app/features/explore/QueryRow.tsx
```diff
@@ -102,12 +100,13 @@ export class QueryRow extends PureComponent<QueryRowProps, QueryRowState> {
   };
 
   setReactQueryEditor = () => {
-    const { mode, datasourceInstance } = this.props;
+    const { datasourceInstance } = this.props;
     let QueryEditor;
 
-    if (mode === ExploreMode.Metrics && datasourceInstance.components?.ExploreMetricsQueryField) {
+    // TODO:unification
```
davkal, line 106: Looks like there is no other way than removing this field specialisation hand add a switcher to the field itself.

public/app/features/explore/state/actions.ts
```diff
@@ -451,23 +431,24 @@ export const runQueries = (exploreId: ExploreId): ThunkResult<void> => {
 
     // Some datasource's query builders allow per-query interval limits,
     // but we're using the datasource interval limit for now
-    const minInterval = datasourceInstance.interval;
+    const minInterval = datasourceInstance?.interval;
```
davkal, line 434: Why would there ever not be an instance?

public/app/features/explore/utils/ResultProcessor.ts
```diff
@@ -16,27 +15,51 @@ import { getGraphSeriesModel } from 'app/plugins/panel/graph2/getGraphSeriesMode
...
+      } else if (isTimeSeries(frame, this.state.datasourceInstance?.meta.id)) {
+        if (shouldShowInVisualisationType(frame, 'graph')) {
+          this.graphFrames.push(frame);
+        }
+        if (shouldShowInVisualisationType(frame, 'table')) {
+          this.tableFrames.push(frame);
+        }
+      } else {
+        // We fallback to table if we do not have any better meta info about the dataframe.
+        this.tableFrames.push(frame);
```
davkal, line 51: 👍

public/app/plugins/datasource/elasticsearch/datasource.ts
```diff
@@ -363,7 +363,7 @@ export class ElasticDatasource extends DataSourceApi<ElasticsearchQuery, Elastic
       let queryObj;
       if (target.isLogsQuery || queryDef.hasMetricOfType(target, 'logs')) {
         target.bucketAggs = [queryDef.defaultBucketAgg()];
-        target.metrics = [queryDef.defaultMetricAgg()];
+        target.metrics = [];
```
davkal, line 366: Ah ok, we're not requesting the aggregation anymore.

public/app/plugins/datasource/elasticsearch/module.ts
```diff
@@ -11,5 +10,4 @@ class ElasticAnnotationsQueryCtrl {
 export const plugin = new DataSourcePlugin(ElasticDatasource)
   .setQueryCtrl(ElasticQueryCtrl)
   .setConfigEditor(ConfigEditor)
-  .setExploreLogsQueryField(ElasticsearchQueryField)
```
davkal, line 14: The component file could be removed?

public/app/plugins/datasource/influxdb/module.ts
```diff
@@ -16,5 +15,4 @@ export const plugin = new DataSourcePlugin(InfluxDatasource)
   .setConfigEditor(ConfigEditor)
   .setQueryCtrl(InfluxQueryCtrl)
   .setAnnotationQueryCtrl(InfluxAnnotationsQueryCtrl)
-  .setExploreLogsQueryField(InfluxLogsQueryField)
```
davkal, line 19: Remove component?

public/app/plugins/datasource/jaeger/datasource.ts
```diff
@@ -49,6 +49,9 @@ export class JaegerDatasource extends DataSourceApi<JaegerQuery> {
                     values: response?.data?.data || [],
                   },
                 ],
+                meta: {
+                  preferredVisualisationType: 'trace',
```
davkal, line 53: I feel like these values should be constants from something in core grafana.

public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -93,30 +91,7 @@ export class LokiDatasource extends DataSourceApi<LokiQuery, LokiOptions> {
...
-                );
-              } else {
-                return dataQueryResponse;
-              }
-            })
-          )
-        )
-      );
-    }
+    filteredTargets.forEach(target => subQueries.push(this.runRangeQuery(target, options, filteredTargets.length)));
```
davkal, line 94: We're dropping the instant query here. Any way we can have it back for metric queries (or we just sent it anyway for logs and see what happens/deal with it in the response).

public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -150,7 +125,10 @@ export class LokiDatasource extends DataSourceApi<LokiQuery, LokiOptions> {
       filter((response: any) => (response.cancelled ? false : true)),
       map((response: { data: LokiResponse }) => {
         if (response.data.data.resultType === LokiResultType.Stream) {
-          throw new Error('Metrics mode does not support logs. Use an aggregation or switch to Logs mode.');
+          return {
+            data: [],
+            key: `${target.refId}_instant`,
```
davkal, line 130: Unclear what's happening here. Also, the `_instant` suffix suggests this has something to do with instant queries.

davkal (approved):
Awesome 🚀 

No blockers from my side.

Minor:

- Loki is missing the table result. Either add in this PR if simple, or add a comment TODO
- Elastic agg histogram from the datasource was nice. Is there no other way to reintroduce this?

kaydelaney:
Failing right now as it depends on https://github.com/grafana/grafana-plugin-sdk-go/pull/191 which needs to get merged first

packages/grafana-data/src/types/data.ts
```diff
@@ -47,6 +47,7 @@ export interface QueryResultMeta {
   searchWords?: string[]; // used by log models and loki
   limit?: number; // used by log models and loki
   json?: boolean; // used to keep track of old json doc values
+  instant?: boolean;
```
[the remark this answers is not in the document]
aocenas: Do we still need this? I thought that was solved by the preferredVis type?

public/app/features/explore/Explore.tsx
```diff
@@ -334,7 +340,8 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
               <SecondaryActions
                 addQueryRowButtonDisabled={isLive}
                 // We cannot show multiple traces at the same time right now so we do not show add query button.
-                addQueryRowButtonHidden={mode === ExploreMode.Tracing}
+                //TODO:unification
```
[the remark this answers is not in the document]
aocenas: My idea for simple solution would be to just show 1 trace view per each query. Would not be very practical from user perspective but would be easy to implement without any special casing here.

packages/grafana-data/src/types/data.ts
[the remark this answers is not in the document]
kaydelaney: Yeah seems like it's not used anymore

public/app/features/explore/state/actions.ts
kaydelaney: Good question, but it's defined as an optional property so the `?` is needed to suppress a strict null error.

NotAFile:
Is it intended that this design doc is private?

PR #25896, opened 2020-06-29
Chore: Enable scopelint Go linter

aknuds1:
Enable the scopelint Go linter as second step in harmonizing with plugin SDK linting.

The new linter catches the issue of closures using for loop variables, which can lead to race conditions.


marefr (approved):
LGTM

bergquist:
Im not sure I like this linter. Why is a scope value needed for those tests? they dont seem to start new go routines to me

aknuds1:
@bergquist I think the problem is the linter can't tell if those variables will be used in a synchronous way or not. Let's see if it's tweakable? scopelint is apparently unmaintained anyway, and the master revision of golangci-lint has a replacement called [exportloopref](https://github.com/kyoh86/exportloopref). Unfortunately exportloopref isn't available in the current release of golangci-lint, but maybe it will be more accurate about the cases it detects.

aknuds1:
Personally, I see value in catching this potential race condition, but maybe we need to tweak the linter.

bergquist:
Agree. 

How about we avoid it for test files? 
https://github.com/golangci/golangci-lint/pull/390

creating a local copy every time we use table testing seems less elegant. Using index position Is erroneous as well.

aknuds1:
@bergquist I can make a PR to exclude test files.

aknuds1:
Made PR #25923.

PR #25903, opened 2020-06-29
Chore: Enable whitespace linter

aknuds1:
As part of effort to harmonize the Go linting with the plugin SDK, enable whitespace linter.

This PR is pretty huge, but it's almost all whitespace changes. If you want I can divide it into several smaller PRs, but I might recommend using `git diff` locally to verify that it's almost all whitespace.


ryantxu (approved):

bergquist:
Borderline if this is valuable or annoying. 

Can we solve this with husky instead? https://github.com/grafana/grafana/blob/master/package.json#L58

aknuds1:
I guess it's down to personal preference, but I really like avoiding unnecessary newlines since I find them noisy when reading code.

bergquist:
A code formatting tool might be able to delete those lines. 

I don't mind getting rid of the white lines. But linting also comes with a cost of increased response time.

aknuds1:
@bergquist Once we've got rid of the extraneous blank lines in this PR, how will it slow us down in daily work? New ones will be caught quickly. If we have a formatter that removes them anyway, the linter will just be there to verify that the formatter did its job.

aknuds1:
@bergquist more importantly, which formatter will remove extra blank lines (corresponding to this linter)? AFAICT `gofmt -s` doesn't.

bergquist (approved):
I'm concerned about the feedback cycles of this linter. But let's give it a try. We can remove the linting step if it's annoying. The other changes are good :)

aknuds1:
OK thanks @bergquist - let's just disable it if it turns out to be annoying. I think it should result in quick feedback cycles though, since CI fails quickly on linting.

PR #25906, opened 2020-06-29
Explore: Add join transform for timeseries shown in table

aocenas:
In case we show timeseries in the table we were doing automatic join transform on Time column. This was changed in https://github.com/grafana/grafana/pull/25694 where we simplified the transform code. This reintroduces the behavior back to the data processing.


torkelo (approved):
Looks good!

PR #25908, opened 2020-06-29
@grafana/e2e: improvements

stevenvachon:
See commit message(s).

### DIFF
--- a/packages/grafana-e2e/src/flows/addDashboard.ts
+++ b/packages/grafana-e2e/src/flows/addDashboard.ts
@@ -15,7 +15,7 @@ export const addDashboard = (config?: Partial<AddDashboardConfig>): any => {
   const fullConfig = {
     timeRange: {
       from: '2020-01-01 00:00:00',
-      to: '2020-01-01 01:00:00',
+      to: '2020-01-01 06:00:00',
     },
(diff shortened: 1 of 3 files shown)


ryantxu:
This seems like a bad idea... maybe 6 hours max?

many timestream databases become very expensive for "large" (more than a few hours) and not recent data.  likely it does not matter much, but we should be very careful before promoting a larger time range query by defualt

stevenvachon:
The query editor seems to default the limit to 25 if that helps? I was just trying to get any data at all to display by default, mainly for ServiceNow at the moment.

ryantxu:
Can you pass in the date range easily?  changing the default seems odd since ServiceNow is a very very non standard datasource

stevenvachon:
> Can you pass in the date range easily?

Yup, it's a config option: `openDashboard({ timeRange: { from: '', to: '' } })`

I've made some new changes too; pushing now.

ryantxu (approved):
LGTM

in master/7.1+ the timezone is also configured in the same popup -- should it be moved to the same structure and set at the same time?

stevenvachon:
Oh, nice, I hadn't noticed that change. Maybe we can update the flows later when we have other flows for the settings window? In its current state, it at least tests that UI a bit.

PR #25917, opened 2020-06-29
Graph Panel: use FieldConfig for displayName and DataLinks

ryantxu:
The field config options for setting names is super important for DataSources that do not (and should not!) implement their own field name alias structure -- for example:  the new Azure Monitor support and flux in InfluxDB

This PR takes a small step towards using FieldConfig to define the series settings.

image

TODO:

- [x] Is this OK to do as a small step?  Is it too confusing to have "Field override" and "Series overrieds"
- [x] Better migration tests -- make sure FieldConfigs migrate OK

The other most obvious setting that we should *really* try to make work is units/decimals -- the details get complicated because existing settings/config is actually about the "axis"


public/app/features/dashboard/state/PanelModel.ts
```diff
@@ -334,7 +334,7 @@ export class PanelModel implements DataConfigSource {
   pluginLoaded(plugin: PanelPlugin) {
     this.plugin = plugin;
 
-    if (plugin.panel && plugin.onPanelMigration) {
+    if (plugin.onPanelMigration) {
```
ryantxu, line 337: The real change here is to allow migrations on angular panels *and* react panels

ryantxu:
@torkelo can you take a quick look at this?  Do you think we also need to figure out how to use `units`?  or is this enough for now?

Units is complicated because they are currently linked to the axis settings

torkelo (commented):
I took this for a spin and it works :) 

Sure a bit odd to only have these two settings available via Field & Overrides tab, that is the main concern. 

But I think it is fine.

ryantxu:
Sorry for the git history...  but the files should be OK now

ryantxu:
@torkelo -- can this get merged in beta1?

PR #25920, opened 2020-06-30
Monaco: option to hide line numbers

ryantxu:
For small editors, the line numbers may not be useful:
image

vs
image


packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
```diff
@@ -12,6 +12,7 @@ export interface CodeEditorProps {
 
   readOnly?: boolean;
   showMiniMap?: boolean;
+  hideLineNumbers?: boolean;
```
tskarhed, line 15: Add this as a knob in the story too :+1:

torkelo (approved):

packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
torkelo: wonder if we should invert it, so that it is showLineNumbers instead? 
Feels more naturual to hide something by setting it to false rather than true

torkelo (commented):
hideLineNumbers => showLineNumbers (can still default to true if we thinkt that will be the most commonly used setting)

packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
ryantxu: except that the "normal" case is to show them -- but I could go either way

dprokop (approved):
LGTM

PR #25921, opened 2020-06-30
Monaco: add suggestions for template variables

ryantxu:
WIP... posting here for early feedback

With many of the code editors, we need to support a standard language *and* apply template variables.  For example the text panel is HTML/markdown with and dashboard templates applied.  SQL etc, is standard SQL + a set of $__macro functions we define

image


packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
```diff
@@ -1,35 +1,113 @@
...
+      insertText: '${__field.name}',
+      range: range,
+      detail: 'Keword detail',
+    },
+  ];
+}
+function registerXXXX(language: string): monaco.IDisposable {
+  console.log('registerCompletionItemProvider', language);
+  return monaco.languages.registerCompletionItemProvider(language, {
+    triggerCharacters: ['$'],
```
ryantxu, line 56: Special key to open with `$`

packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
```diff
@@ -1,35 +1,113 @@
...
+      documentation: 'The field value name (...)',
+      insertText: '${__field.name}',
+      range: range,
+      detail: 'Keword detail',
+    },
+  ];
+}
+function registerXXXX(language: string): monaco.IDisposable {
+  console.log('registerCompletionItemProvider', language);
+  return monaco.languages.registerCompletionItemProvider(language, {
```
ryantxu, line 55: sadly -- it looks like completion needs to be added globally to the language :(  would be much nicer if it was linked to the editor instance.

It does look like you can register and destory the instances, but this has the not great side affect that really one suggest context is active for each language at a time.

NOTE: this works on top of the standard language completions :tada:

packages/grafana-ui/src/components/Monaco/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  Method = 0,
+  Function = 1,
+  Field = 3,
+  Variable = 4,
+  Property = 9,
+  Constant = 14,
+  Keyword = 17,
+}
+
+export interface CodeEditorSuggestionItem {
```
ryantxu, line 41: Very similar shape to https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/types/completion.ts#L34

packages/grafana-ui/src/components/Monaco/CodeEditor.tsx
torkelo: nice!

ryantxu:
@torkelo @dprokop -- this would be really great to get into beta1.  It solves key issues for timestream :)

You can test using the text panel on a page with template variables:
image

packages/grafana-ui/src/components/Monaco/types.ts
```diff
@@ -0,0 +1,77 @@
+export type CodeEditorChangeHandler = (value: string) => void;
+export type CodeEditorSuggestionProvider = () => CodeEditorSuggestionItem[];
```
torkelo, line 2: This type feels too simplistic, needs prefix & context so suggestions can be context specific. So we will have to break this api later

packages/grafana-ui/src/components/Monaco/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { VariableSuggestion } from '@grafana/data';
+import { CodeEditorSuggestionItem, CodeEditorSuggestionItemKind } from './types';
+
+/**
+ * @alpha
+ */
+export function variableSuggestionToCodeEditorSuggestion(sug: VariableSuggestion): CodeEditorSuggestionItem {
```
dprokop, line 7: Not a huge fan of this mapping. I mean we could probably try to unify the VariablesSuggestion and CodeEditorSuggestionItem to have one interface. I'll try to do that.

packages/grafana-ui/src/components/Monaco/types.ts
ryantxu: For complex things, shouldn't we just use monaco/lsp?  The goal here is to have simple support for the super common templates/variables.

Complex things can be supported by either loading custom languages:
https://github.com/microsoft/monaco-languages/tree/master/src

or customizing in JS:
https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-custom-languages

ryantxu: Note that these suggestions exist on top of the defined language.

packages/grafana-ui/src/components/Monaco/types.ts
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * A human-readable string that represents a doc-comment.
+   */
+  documentation?: string; // | IMarkdownString;
+
+  /**
+   * A string or snippet that should be inserted in a document when selecting
+   * this completion. When `falsy` the `label` is used.
+   */
+  insertText?: string;
```
dprokop, line 76: The label vs insertText. I think we can make the insertText not to be optional, to be more in line with label/value in VariableSuggestion interface. But then there's https://github.com/grafana/grafana/pull/25921/files#diff-f02ee9fdd9a1eb7c16cb24b321ddee2aR12 that assumes `$` for suggestions retrieval and it's going through the labels to find the matching suggestion. Think we need to update this to take the insert text into consideration when generating the suggestions list.

dprokop (approved):

PR #25923, opened 2020-06-30
Chore: Disable scopelint for tests

aknuds1:
As @bergquist [suggested](https://github.com/grafana/grafana/pull/25896#issuecomment-651329377), disable the scopelint Go linter for tests since it makes writing table driven tests a bit hairier. In the future, we should see if it's possible to tweak the linter so it doesn't flag such cases.


bergquist (approved):
Great! Thanks :)

<3

Issue #25904, opened 2020-06-29
Panel Inspect: Crash when applying transform makes the current dataframe disappear


tskarhed:
**What happened**:
1. Apply the Outer join transform
2. Open the panel inspector
3. Change dataframe to antyhing other than the first two
4. Crash
![Peek 2020-06-29 17-16](https://user-images.githubusercontent.com/1438972/86023855-86bbbf80-ba2c-11ea-96c3-3a5fde5534ed.gif)

**What you expected to happen**:
If the Data Frame does not exist after the transformation, it should default ot the seriesToColumns option.

dprokop:
Fixed with #25924

PR #25924, opened 2020-06-30
Panel Inspect: Fix crash when transform removes selected data frame

tskarhed:
Handles cases where the currently selected data frame is not in the newly transformed data. Also disables the Select when there aren't more than one data frame to select from.

Fixes #https://github.com/grafana/grafana/issues/25904


dprokop (changes requested):
Since the data tab is becoming more and more complex, can you add tests for this component?

tskarhed:
Looking at writing some e2e and maybe unit tests today.

dprokop (approved):
Works as expected!

PR #25929, opened 2020-06-30
Fix: Value mappings match against string values

### LINKED ISSUE #25351: Entering text in the Value field in the Value Mappings section
Value in the Value Mappings section of the New Table panel does not support text input, only numbers.
Grafana 7.0.3

In old table:
image

In new table:
image

peterholmberg:
There was a regression on the new Table introduced in 7.0 that removed the ability to match string values to valuemappings. 

Fixes #25351

### DIFF
--- a/packages/grafana-data/src/utils/valueMappings.test.ts
+++ b/packages/grafana-data/src/utils/valueMappings.test.ts
@@ -78,4 +78,15 @@ describe('Format value with value mappings', () => {
 
     expect(getMappedValue(valueMappings, value).text).toEqual('1-20');
   });
+
+  it('should map value text to mapping', () => {
+    const valueMappings: ValueMapping[] = [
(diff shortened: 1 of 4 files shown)


dprokop (changes requested):
Tested this, and found one fundamental issue. Let's say we have data like this:

"Time","Value"
1587459459769,hot
1587462040714,cold

Then adding value mapping `hot -> i.e. 42` does not have any effect until override is created. I think the issue is [here](https://github.com/grafana/grafana/blob/53b9f325c00281576cf823cfdf333f5769dbb1ce/packages/grafana-ui/src/utils/standardEditors.tsx#L162) as value mappings are processed only for numeric fields.

peterholmberg:
Ah, I only tried with applying an override

dprokop (approved):
Great!

PR #25934, opened 2020-06-30
TimeZonePicker: added possibility to toggle if internal time zones should be included or not.

mckn:
This updates the time zone picker so we don't include internal time zones if that prop is not set to `true`. Also make sure that we match on timeZone even if it is passed in different casing.


Clarity-89 (approved):
Looks good 👍

PR #25936, opened 2020-06-30
Docker: Upgrade to base on Alpine 3.12

aknuds1:
Upgrade our Dockerfile to base itself on Alpine 3.12, Node 12.18.1 and Go 1.14.4.


marefr (approved):
LGTM

PR #25937, opened 2020-06-30
Azure: Include Log Analytics error and ExecutedQueryString

kylebrandt:
When there is an error message in the HTTP response, this copies that error so it is relayed to the user by being shown in the frontend.

This also makes it so when there is an error the interpolated query string (ExecutedQueryString) is returned so users can see it in the query inspector.


ryantxu (approved):

marefr (approved):
LGTM

marefr:
Ohh it was already merged 👍

kylebrandt:
> Ohh it was already merged

Still appreciated <3 In a rush to get a couple things in for the beta with Azure. So if I merge something and you leave feedback I can always follow up with it in another PR.

PR #25942, opened 2020-06-30
Elastic: Internal data links

aocenas:
Add internal data links, similar to what we have in Loki to elastic config.

Right now only tracing data sources are supported as target data source.
Screenshot from 2020-06-30 16-42-18
Screenshot from 2020-06-30 16-42-37

We do not have good testing data for this case. In general though for manual testing the point is that the field value is correctly interpolated into the query of target data source and it is open when clicking on it so the actual value does not matter for the test.

### DIFF
--- a/docs/sources/features/datasources/elasticsearch.md
+++ b/docs/sources/features/datasources/elasticsearch.md
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
 - **Message field name:**  message
 - **Level field name:** fields.level
 
+### Data links
+
+Data links create a link from specified field that can be accessed in logs view in Explore.
+
+Each data link configuration consists of: 
+- **Field**: Name of the field used by the data link
+- **Url/Query**: If link is external, you fill in full url for the link. In case of internal link this input serves as query for the target data source. In both cases you can interpolate the value from the field with `${__value.raw }` macro.
+- **Internal link**: Switch to select if link is internal or external. In case of internal link, data source selector will be shown where you can select the target data source. At this moment only tracing data sources are supported.
+
 ## Metric Query editor
 
 ![Elasticsearch Query Editor](/img/docs/elasticsearch/query_editor.png)
--- a/docs/sources/features/datasources/elasticsearch.md
+++ b/docs/sources/features/datasources/elasticsearch.md
@@ -93,7 +93,7 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
 
 ### Data links
 
-Data links create a link from specified field that can be accessed in logs view in Explore.
+Data links create a link from a specified field that can be accessed in logs view in Explore.
 
 Each data link configuration consists of: 
 - **Field**: Name of the field used by the data link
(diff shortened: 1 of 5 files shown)


ivanahuckova (approved):
I have reviewed and tested this and everything looks good and works as expected. 
Also it has everything: 
✅ documentation
✅ test coverage
✅ useful comments
✅ readable code

Great work Andrej!🚀

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
 - **Message field name:**  message
 - **Level field name:** fields.level
 
+### Data links
```
oddlittlebird, line 94: This is going to be very confusing, considering we have a feature of the same name that works differently than this. @dprokop, thoughts?

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
...
+
+Data links create a link from specified field that can be accessed in logs view in Explore.
```
oddlittlebird, line 96:
```suggestion
Data links create a link from a specified field that can be accessed in logs view in Explore.
```

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
...
+
+Each data link configuration consists of: 
+- **Field**: Name of the field used by the data link
```
oddlittlebird, line 99:
```suggestion
- **Field -** Name of the field used by the data link.
```

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
...
+- **Url/Query**: If link is external, you fill in full url for the link. In case of internal link this input serves as query for the target data source. In both cases you can interpolate the value from the field with `${__value.raw }` macro.
```
oddlittlebird, line 100:
```suggestion
- **URL/query -** If the link is external, then enter the full link URL. If the link is internal link, then this input serves as query for the target data source. In both cases, you can interpolate the value from the field with `${__value.raw }` macro.
```

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -91,6 +91,15 @@ For example, if you're using a default setup of Filebeat for shipping logs to El
...
+- **Internal link**: Switch to select if link is internal or external. In case of internal link, data source selector will be shown where you can select the target data source. At this moment only tracing data sources are supported.
```
oddlittlebird, line 101:
```suggestion
- **Internal link -** Select if the link is internal or external. In case of internal link, a data source selectorallows you to select the target data source. Only tracing data sources are supported.
```

oddlittlebird (dismissed):
Got some doc changes. 

This is going to be awful for people searching for our regular data link feature.

docs/sources/features/datasources/elasticsearch.md
aocenas: This feature is actually fairly similar and we will probably unify it even more in the future so not sure if it make sense to change the name right now. There is a difference in with Loki where we do data link + create a new field that's why the naming there is different.

PR #25953, opened 2020-06-30
Suggestion: render documentation as markdown in suggestion's info

zoltanbedi:
The comment here https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/types/completion.ts#L57 says that this can be a markdown but it doesn't rendered as a markdown so I've added it.
Grafana - Google Chrome 2020-06-30 21_52_38


packages/grafana-ui/src/components/Typeahead/TypeaheadInfo.tsx
```diff
@@ -41,15 +41,15 @@ interface Props {
 export const TypeaheadInfo: React.FC<Props> = ({ item, height }) => {
   const visible = item && !!item.documentation;
   const label = item ? item.label : '';
-  const documentation = item && item.documentation ? item.documentation : '';
+  const documentation = renderMarkdown(item?.documentation);
```
kaydelaney, line 44: Should probably sanitize the produced html with `textUtil.sanitize()`

zoltanbedi: Done

kaydelaney (approved):

PR #25954, opened 2020-06-30
Panel: Cancel panel query via loading spinner

### LINKED ISSUE #25783: Support canceling a running query

Some datasources will start queries that may take a long (long!) time to run.  Often these will return cancelation tokens or jobIds that need to be passed back in order to kill the job.  (disconnect is not enough!)

We currently have no good method to cancel the running queries.  

**Why is this needed**:

Timestream and Splunk can both kick off jobs that take a LONG time to run.  we need to support killing these jobs

torkelo:
Closes #25783 

This 
* Updates the loading spinner to use the same refresh icon we use in refresh picker and run query button in explore. Had to add a new counter clock wise rotation for the spinner to work with this icon as the refresh icon arrows point in the counter clockwise direction. 
* Adds a tooltip and pointer cursor for the loading icon 
* Clicking on loading spinner will cancel the query 

Screen Shot 2020-06-30 at 21 58 16

This will cancel the http request where data source set requestId correctly in the backendSrv.datasourceRequest call. Many do not set requestId in a way where this works. For that we need https://github.com/grafana/grafana/pull/25746

### DIFF
--- a/public/app/features/dashboard/state/PanelQueryRunner.ts
+++ b/public/app/features/dashboard/state/PanelQueryRunner.ts
@@ -25,6 +25,7 @@ import {
   applyFieldOverrides,
   DataConfigSource,
   TimeZone,
+  LoadingState,
 } from '@grafana/data';
 
 export interface QueryRunnerOptions<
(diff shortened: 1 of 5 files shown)


public/app/features/dashboard/state/PanelQueryRunner.ts
```diff
@@ -198,6 +199,22 @@ export class PanelQueryRunner {
...
+      return;
+    }
+
+    this.subscription.unsubscribe();
+
+    // If we have an old result with loading state, send it with done state
+    if (this.lastResult && this.lastResult.state === LoadingState.Loading) {
+      this.subject.next({
+        ...this.lastResult,
+        state: LoadingState.Done,
```
ryantxu, line 213: not 100% convinced Done is the right status... but that is better than adding another status

ryantxu (approved):
now the real work to make sure they actually cancel properly!

public/app/features/dashboard/state/PanelQueryRunner.ts
torkelo: Can revisit after a PanelHeader UI make over (to show state icons more uniformly and extendable)

dprokop (approved):

mckn (approved):
LGTM, awesome work!!

PR #25960, opened 2020-07-01
Panel inspect: data tab e2e tests

dprokop:
Followup for https://github.com/grafana/grafana/pull/25924


packages/grafana-e2e-selectors/src/selectors/components.ts
```diff
@@ -63,6 +63,8 @@ export const Components = {
   PanelInspector: {
     Data: {
       content: 'Panel inspector Data content',
+      applyTransforms: 'Apply transforms to data',
+      applyFieldOverrides: 'Apply field overrides to data',
```
mckn, line 67: nit: Are these buttons? Maybe follow the naming convention as we have for the `actionButton`.

mckn (approved):
One small nit but I think it looks good!

PR #25964, opened 2020-07-01
Docs: monthly reports

xlson:

docs for the monthly reports schedule

Fixes #


docs/sources/enterprise/reporting.md
```diff
@@ -80,6 +80,16 @@ Weekly reports are generated once per week. All fields are required.
 * **Time -** Time of day in 24 hours format when the report should be sent.
 * **Time zone -** Time zone for the **Time** field.
 
+#### Monthly
+
+> Only available in Grafana Enterprise v7.1+.
+
+Monthly reports are generated once per month. All fields are required.
+
+* **Day in month -** Day of the month the when report should be sent. (it's possible to select `last` for reports that should go out on the last day of the month).
```
AgnesToulet, line 89:
```suggestion
* **Day in month -** Day of the month when the report should be sent (it's possible to select `last` for reports that should go out on the last day of the month).
```
Just noticed that I rewrote all dayInMonth variables to dayOfMonth and forgot to rename the field. I think it should be "Day of month" or "Day of the month"?

AgnesToulet (commented):
Great that you thought about this! Looks good except for a small typo and I think we should rename the first field in both the doc and the code.

docs/sources/enterprise/reporting.md
oddlittlebird: If it were regular writing, it would be "day of the month", but for variables, shorter seems better.

oddlittlebird: Remove the parenthesis and replace "it's possible to" with "You can".

docs/sources/enterprise/reporting.md
```diff
@@ -80,6 +80,16 @@ Weekly reports are generated once per week. All fields are required.
...
+* **Time -** Time of day in 24 hours format when the report should be sent.
```
oddlittlebird, line 90:
```suggestion
* **Time -** Time the report is sent, in 24-hour format.
```

oddlittlebird (dismissed):
A few suggestions.

docs/sources/enterprise/reporting.md
xlson: Thank you, I will incorporate the feeback!

docs/sources/enterprise/reporting.md
xlson: Thank you, I will incorporate the feeback!

AgnesToulet (approved):
LGTM!

PR #25973, opened 2020-07-01
Run strict typecheck on PRs

dprokop:


aknuds1:
Why shouldn't this be done as part of the normal front-end tests job?

dprokop:
> Why shouldn't this be done as part of the normal front-end tests job?

It could. I'll update this

dprokop:
Ah, wait - it couldn't becouse on master it would be performed twice - once for metrics reporting purposes and one just for check. On PRs we do not report metrics, so that's why I have separated it @aknuds1

aknuds1:
@dprokop I think it's better to modify the script to determine if it should report metrics or not. It can do that with the help of CircleCI environment variables.

aknuds1:
The script already only reports metrics if on the master branch, so you can just use it.

PR #25982, opened 2020-07-01
Fix: Build-in plugins failed to load in windows

papagian:

Fixes #

### DIFF
--- a/pkg/plugins/frontend_plugin.go
+++ b/pkg/plugins/frontend_plugin.go
@@ -3,6 +3,7 @@ package plugins
 import (
 	"net/url"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/grafana/grafana/pkg/setting"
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
LGTM

marefr (approved):
LGTM

Tested to run `fmt.Println(filepath.Base("C:\\test\\welcome"))` in the [example ](https://golang.org/pkg/path/filepath/#Base) and results in `C:\test\welcome` but I guess that's because Golang server is not running on Windows? 

Just to be sure, we should verify nightly works on Windows after this have been merged before doing a beta2.

papagian:
> Tested to run `fmt.Println(filepath.Base("C:\\test\\welcome"))` in the [example ](https://golang.org/pkg/path/filepath/#Base) and results in `C:\test\welcome` but I guess that's because Golang server is not running on Windows?

I experienced the same but I came to the same conclusion and it's actually makes sense because of this:
https://golang.org/src/path/filepath/path.go?s=13035:13064#L439

PR #25984, opened 2020-07-01
Alerting: New feature toggle for enabling standalone alerts

papagian:
- Introduce new feature toggle: `standaloneAlerts `


bergquist:
:fireworks: 
But I'm thinking we should use https://github.com/grafana/grafana/blob/master/pkg/setting/setting.go#L218 instead of introducing a new setting.

papagian:
> 🎆
> But I'm thinking we should use https://github.com/grafana/grafana/blob/master/pkg/setting/setting.go#L218 instead of introducing a new setting.

Ah that makes sense; I will change it then!

docs/sources/packages_api/data/featuretoggles.md
```diff
@@ -30,6 +30,7 @@ import { FeatureToggles } from '@grafana/data';
 |  [meta](#meta-property) | <code>boolean</code> |  |
 |  [newEdit](#newedit-property) | <code>boolean</code> |  |
 |  [reportGrid](#reportgrid-property) | <code>boolean</code> |  |
+|  [standaloneAlerts](#standalonealerts-property) | <code>boolean</code> |  |
```
marefr, line 33: Have you updated this manually? Think this file is generated after a stable release.

marefr (changes requested):
There are files updated that are automatically generated. I think this only happens after a stable release so shouldn't be needed?

Other than that LGTM

docs/sources/packages_api/data/featuretoggles.md
papagian: Not manually. I have generated it by running `scripts/ci-reference-docs-build.sh` but I'm also unsure if it's required at this phase. Let's see what the frontend people think otherwise I'll revert the [commit](https://github.com/grafana/grafana/pull/25984/commits/f6ca7e7ec96ac85b28622863ee8927423ea82a0d)

bergquist (approved):
Loogs good from my POV

@dprokop are the other changes here because the script havent been executed during the development of those features?

docs/sources/packages_api/data/featuretoggles.md
dprokop: @papagian no need for comiting these changes. We regenerate packages docs during the release anyways.

marefr (approved):
LGTM

PR #25990, opened 2020-07-01
PanelEdit: Remove field label for single item categories

torkelo:
Fixes this double section & item name we have for value mapppings and data links and thresholds 

Problem:
Screenshot from 2020-07-01 20-32-48

With fix: 

Screenshot from 2020-07-01 20-32-55

### DIFF
--- a/public/app/features/dashboard/components/PanelEditor/FieldConfigEditor.tsx
+++ b/public/app/features/dashboard/components/PanelEditor/FieldConfigEditor.tsx
@@ -178,7 +178,7 @@ export const DefaultFieldConfigEditor: React.FC<Props> = ({ data, onChange, conf
       );
 
       // hide label if there is only one item and category name is same as item, name
-      if (categoryItemCount === 1 && item.category?.length && item.category[0] === item.name) {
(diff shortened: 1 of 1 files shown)


public/app/features/dashboard/components/PanelEditor/FieldConfigEditor.tsx
```diff
@@ -171,12 +171,17 @@ export const DefaultFieldConfigEditor: React.FC<Props> = ({ data, onChange, conf
...
 
-      const label = (
+      let label: ReactNode | undefined = (
         <Label description={item.description} category={item.category?.slice(1)}>
           {item.name}
         </Label>
       );
 
+      // hide label if there is only one item and category name is same as item, name
+      if (categoryItemCount === 1 && item.category?.length && item.category[0] === item.name) {
```
Clarity-89, line 181:
```suggestion
      if (categoryItemCount === 1 && item.category?.[0] === item.name) {
```

Clarity-89 (approved):
Looks good! One minor change to consider.

PR #25998, opened 2020-07-02
Panel Loading: spin clockwise, not counter clockwise

ryantxu:
In #25954, the loading icon changed to match the refresh icon (`sync`) -- the arrows point left, so the spinner goes counter-clockwise.

My brain will totally explode if we have to watch time spinning counter-clockwise... call me old fashioned.

This flips the icon and spins it clockwise:
image

ideally the refresh icon would also be flipped, but that is not urgent.


torkelo (approved):
Ah, nice trick with the flip!

PR #26009, opened 2020-07-02
Feature: Dashboard Usage Tracking

### LINKED ISSUE #16700: Extend dashboard model with LastVisited attribute
I'd like to have possibility to check dashboard's last visited time.

**Why is this needed**:
This could help to detect old and forgotten dashboards which is essential for cases when Grafana is used by big amount of teams and users.

sinkingpoint:
In large grafana deployments, admins are often overwhelmed with
dashboards, harming discoverability and usability in general of Grafana.

In this commit we add in a "visited" column to the dashboard meta, along
side the similar "Created" and "Updated" timestamps. "Visited" gets
updated every time the "GetDashboard" api call is made (But does not
block this call, so as not to break read only deployments). This allows
us to track dashboards that have been abandoned or are no longer in use
and attempt to keep the cruft under control more easily

Fixes grafana#16700

**Notes**:

I've added this to the model interface - not sure whether this is the right place for this - perhaps it might be worthwhile adding a new "VisitDashboardCommand" to the Bus or similar? This was the path of least resistance, but happy to change this up if necessary

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -306,6 +306,7 @@ func (hs *HTTPServer) registerRoutes() {
 				dashIdRoute.Get("/versions", Wrap(GetDashboardVersions))
 				dashIdRoute.Get("/versions/:id", Wrap(GetDashboardVersion))
 				dashIdRoute.Post("/restore", bind(dtos.RestoreDashboardVersionCommand{}), Wrap(hs.RestoreDashboardVersion))
+				dashIdRoute.Post("/visit", Wrap(hs.VisitDashboard))
 
 				dashIdRoute.Group("/permissions", func(dashboardPermissionRoute routing.RouteRegister) {
 					dashboardPermissionRoute.Get("/", Wrap(GetDashboardPermissionList))
--- a/pkg/api/dashboard_visit.go
+++ b/pkg/api/dashboard_visit.go
@@ -1,0 +1,21 @@
+package api
+
+import (
+	"github.com/grafana/grafana/pkg/bus"
+	"github.com/grafana/grafana/pkg/models"
+)
+
+func (hs *HTTPServer) VisitDashboard(c *models.ReqContext) Response {
+	dashboardId := c.ParamsInt64(":dashboardId")
+	visitDashboardCommand := &models.VisitDashboardCommand{
+		UserId:      c.UserId,
+		OrgId:       c.OrgId,
+		DashboardId: dashboardId,
+	}
+
+	if err := bus.Dispatch(visitDashboardCommand); err != nil {
+		hs.log.Warn("Failed to mark dashboard as visited", "userId", c.UserId, "orgId", c.OrgId, "dashboardId", dashboardId, "err", err)
+	}
+
+	return Success("")
+}
--- a/pkg/cmd/grafana-server/server.go
+++ b/pkg/cmd/grafana-server/server.go
@@ -38,6 +38,7 @@ import (
 	_ "github.com/grafana/grafana/pkg/services/rendering"
 	_ "github.com/grafana/grafana/pkg/services/search"
 	_ "github.com/grafana/grafana/pkg/services/sqlstore"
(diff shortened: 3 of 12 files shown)


sinkingpoint:
Hummm. Build failure seems... unrelated? ```  1) Inspect drawer tests
 
       Tests various Inspect Drawer scenarios:
 
     Error: The following error originated from your application code, not from Cypress.
 

 
  > Cannot read property 'getText' of null
 

 
TypeError: Cannot read property 'getText' of null
 
    at Object.ai [as getFoldingRanges] (http://localhost:3001/public/build/monaco-json.worker.js:2:215257)
 
    at e.getFoldingRanges (http://localhost:3001/public/build/monaco-json.worker.js:2:221188)
 
    at e.fmr (http://localhost:3001/public/build/monaco-json.worker.js:2:116605)
 
    at e._handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7414)
 
    at Object.handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7018)
 
    at e._handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:5038)
 
    at e.handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:4606)
 
    at e.onmessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7097)
 
    at Tt.self.onmessage (http://localhost:3001/public/build/monaco-json.worker.js:2:117109)
  ```

marefr (commented):
Hey, thank you for contributing. Grafana Enterprise includes this in [Usage Insights](https://grafana.com/docs/grafana/latest/enterprise/usage-insights/) and are storing last visited timestamp for a dashboard asynchronously and not in the same database table as dashboards are stored. We don't want to have writes in the read path due to possible performance problems. That's why we cannot accept your currently proposed changes.

marefr:
Regarding build failure, think you need to sync with latest master

sinkingpoint:
Just to be clear, are you against the principle of this change (And people should upgrade to Grafana enterprise)? Or, if I work this into being async as is apparently done there would this be acceptable?

marefr:
> Just to be clear, are you against the principle of this change (And people should upgrade to Grafana enterprise)? 

Not really. But not sure that we in the future maybe make some parts of usage insights an OSS feature and if so maybe better to wait a bit. But the main is that your proposal is not async and are using the dashboard table for storing this, that's a lot of writes if you have a lot of users visiting the same dashboard. Your welcome to work on async solution and we would consider accepting it.

marefr:
You can convert PR to a draft until it's ready for review

sinkingpoint:
Thanks for the feedback @marefr :) Will work on this now

sinkingpoint:
@marefr I've done a bit more work on this - "visiting" a dashboard is now an extra API call done asynchronously with the fetching of the dashboard

torkelo:
Thanks for your contribution! It's a bit tricky how to handle this as it conflicts with the usage tracking in Grafana Enterprise (where we also record this information and have a generic usage tracking API). So this would result in duplicate tracking

edevil:
If the problem is the duplicate tracking, how about having this behind a feature flag that gets disabled when using the Enterprise version?

sinkingpoint:
CI tests pass now. I'm happy to add a feature flag guard if this is possible / desired

sinkingpoint:
@torkelo @marefr Any idea where we go from here?

jessabe:
Hej @sinkingpoint -- just a little update that this is being discussed internally right now across a few other teams. But things are a bit slow due to summer. Hopefully we'll get back to you soon on this. Apologies!

sinkingpoint:
Thanks @jessabe appreciate the update, and the politics that might be involved :)

mckn:
Thanks for opening this PR @sinkingpoint! Unfortunately in its current state we won't be able to approve it. It is a bit too far from being a complete feature and it has been stale too long. Feel free to help out on other parts of Grafana 🙏

PR #26017, opened 2020-07-02
Docker: Make sure to create default plugin provisioning directory

marefr:
Make sure that plugins provisioning directory are created in docker images to remove the following error message in startup:
```
EROR[07-02|12:18:24] Failed to read plugin provisioning files from directory logger=provisioning.plugins path=/etc/grafana/provisioning/plugins error="open /etc/grafana/provisioning/plugins: no such file or directory"
```


aknuds1 (approved):

bergquist (approved):
Let's reconsider error handling later

PR #24533, opened 2020-05-11
OrgUsersTable: Fix width when there are long cells in table

tskarhed:
**What this PR does / why we need it**:
We like good looking tables.
Screenshot from 2020-05-11 17-13-28

**Which issue(s) this PR fixes**:
Part of #23551 - Check the screenshot

**Special notes for your reviewer**:
Added ellipsis to enable text overflow.

PR #26019, opened 2020-07-02
AdminUsersTable: Fix width issues

tskarhed:
Solves the same problem as #24533, but for the users list under server admin.

Before: 
Screenshot from 2020-07-02 15-05-45
After:
Screenshot from 2020-07-02 15-58-30


dprokop (approved):
🥳

PR #26021, opened 2020-07-02
WIP: text with suggestions std editor (display name for start)

### LINKED ISSUE #22359: FieldOverrides: Generic component for displaying variables suggestions
For i.e. Title field in field config it's possible to use variables for the title to be interpolated. We need a component, similar to data link input` that given a suggestions list will allow picking up variable from a suggestions list dropdown.

dprokop:
Closes https://github.com/grafana/grafana/issues/22359
First attempt (really early draft to estimate work) to support variable suggestions in display name.  

Adds `textWithSuggestions` std editor, currently based on DataLinkInput, which needs a small refactor to be more generic (`TextInputWithSuggestions` ?).

- [ ] Should suggestions be configurable for the editor?
- [ ] What suggestions to expose in the field config editor?
- [ ] Add focus state to input with suggestions
- [ ] Data links insert `var-XXX={$XXX}` when selecting a template variable https://github.com/grafana/grafana/blob/c5385a02ac40cc0715351d14d661a13098c1eca5/packages/grafana-ui/src/components/DataLinks/DataLinkInput.tsx#L128 . The `TextInputWithSuggestions` needs probably some suggestion processors that is executed before the suggestion is saved to the editor value


packages/grafana-data/src/field/overrides/processors.ts
```diff
@@ -70,6 +77,10 @@ export interface StringFieldConfigSettings {
   rows?: number;
 }
 
+export interface StringFieldWithSuggestionsConfigSettings extends StringFieldConfigSettings {
+  getSuggestions: () => VariableSuggestion[];
```
dprokop, line 81: thoughts on this extension point @ryantxu?

packages/grafana-ui/src/components/OptionsUI/string.tsx
```diff
@@ -37,3 +43,23 @@ export const StringValueEditor: React.FC<FieldConfigEditorProps<string, StringFi
...
+  onChange,
+  item,
+  context,
+}) => {
+  const suggestions = useMemo(() => {
+    return context.getSuggestions ? context.getSuggestions(VariableSuggestionsScope.Values) : [];
+  }, [context.getSuggestions]);
+
+  return (
+    <DataLinkInput
```
dprokop, line 58: TODO: rename and make datalinks agnostic

public/app/app.ts
```diff
@@ -99,7 +99,11 @@ export class GrafanaApp {
     setMarkdownOptions({ sanitize: !config.disableSanitizeHtml });
 
     standardEditorsRegistry.setInit(getStandardOptionEditors);
-    standardFieldConfigEditorRegistry.setInit(getStandardFieldConfigs);
+    standardFieldConfigEditorRegistry.setInit(
+      getStandardFieldConfigs({
+        getVariables: getTemplateSrv().getVariables.bind(getTemplateSrv()),
```
dprokop, line 104: do the standard editor need access to template vars? Should we allow using those in i.e. display name?

public/app/features/dashboard/components/PanelEditor/FieldConfigEditor.tsx
```diff
@@ -190,7 +191,18 @@ export const DefaultFieldConfigEditor: React.FC<Props> = ({ data, onChange, conf
...
-              getSuggestions: (scope?: VariableSuggestionsScope) => getDataLinksVariableSuggestions(data, scope),
+              getSuggestions: (scope?: VariableSuggestionsScope) => {
+                if (item.settings?.getSuggestions) {
+                  return item.settings.getSuggestions(
+                    plugin,
+                    data,
+                    getTemplateSrv().getVariables.bind(getTemplateSrv()),
+                    scope
+                  );
+                }
```
dprokop, line 202: Moved suggestions providers(i.e. getDataLinksVariableSuggestions) to grafana-data so that the configuration can be configured via declarative options API.

@ryantxu @torkelo do you think we should provide some default suggestions if no config is provided? Should those maybe be merged with the ones provided with the API?

packages/grafana-data/src/types/suggestions.ts
```diff
@@ -0,0 +1,8 @@
+import { DataFrame, PanelPlugin, VariableModel, VariableSuggestion, VariableSuggestionsScope } from '..';
+
+export type SuggestionsProvider = (
```
dprokop, line 3: Introducing this type for having unified providers for panel/field option suggestions

packages/grafana-ui/src/components/OptionsUI/string.tsx
```diff
@@ -37,3 +43,21 @@ export const StringValueEditor: React.FC<FieldConfigEditorProps<string, StringFi
...
   );
 };
+
+export const StringWithSuggestionsValueEditor: React.FC<FieldConfigEditorProps<
+  string,
+  StringFieldWithSuggestionsConfigSettings
+>> = ({ value, onChange, item, context }) => {
+  const suggestions = useMemo(() => {
+    return context.getSuggestions ? context.getSuggestions(VariableSuggestionsScope.Values) : [];
+  }, [context.getSuggestions]);
```
dprokop, line 53: context.getSuggestions allows consuming the suggestions with the scope of the variables defined. item.settings.getSuggestions is the definition of the suggestion provider. Problem I see here is both of these are available for someone using the API. Any thoughts you have maybe @ryantxu?

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -35,23 +37,26 @@ import { UnitValueEditor } from '../components/OptionsUI/units';
...
     description: 'Change the field or series name',
-    editor: standardEditorsRegistry.get('text').editor as any,
-    override: standardEditorsRegistry.get('text').editor as any,
+    editor: standardEditorsRegistry.get('textWithSuggestions').editor as any,
+    override: standardEditorsRegistry.get('textWithSuggestions').editor as any,
     process: stringOverrideProcessor,
     settings: {
       placeholder: 'none',
       expandTemplateVars: true,
+      getSuggestions: getDataLinksVariableSuggestions,
```
dprokop, line 59: Needs a separate provider, curently using the data links provider.

packages/grafana-data/src/field/fieldOverrides.ts
```diff
@@ -1,25 +1,25 @@
 import {
-  DynamicConfigValue,
```
ryantxu, line 2: just alphabetizing changes here?   exclude this file?  I assume your IDE is doing it for you automatically... hugoh has the same thign

packages/grafana-data/src/types/suggestions.ts
```diff
@@ -0,0 +1,8 @@
...
+  plugin: PanelPlugin,
+  data: DataFrame[],
+  getTemplateVariables: () => VariableModel[],
```
ryantxu, line 6: hymm -- not wild about adding `getTemplateVariables` or templateSrv  -- maybe better if the response returns a more complex object like:

```
{
  includeTemplateVariables?: boolean;
  suggestions?: VariableSuggestion[];
}
```

packages/grafana-data/src/types/suggestions.ts
```diff
@@ -0,0 +1,8 @@
+import { DataFrame, PanelPlugin, VariableModel, VariableSuggestion, VariableSuggestionsScope } from '..';
+
+export type SuggestionsProvider = (
+  plugin: PanelPlugin,
```
ryantxu, line 4: do we need PanelPlugin?  it would be nice to have the same interface for DataSource query editors

packages/grafana-ui/src/components/OptionsUI/string.tsx
```diff
@@ -37,3 +43,21 @@ export const StringValueEditor: React.FC<FieldConfigEditorProps<string, StringFi
     />
   );
 };
+
+export const StringWithSuggestionsValueEditor: React.FC<FieldConfigEditorProps<
```
ryantxu, line 47: is it possible to avoid a new editor?  just use the existing text one?

In general, it should be the exception that we do *not* support template variables

packages/grafana-ui/src/components/OptionsUI/string.tsx
ryantxu: I don't understand why you need the memo here?  is it so that things are forced to `VariableSuggestionsScope.Values`?

packages/grafana-data/src/types/suggestions.ts
ryantxu: actually maybe `includeTemplateVariables` is a flag next to this since it likely does not depend on the state

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -35,23 +37,26 @@ import { UnitValueEditor } from '../components/OptionsUI/units';
...
     name: 'Display name',
     description: 'Change the field or series name',
-    editor: standardEditorsRegistry.get('text').editor as any,
-    override: standardEditorsRegistry.get('text').editor as any,
+    editor: standardEditorsRegistry.get('textWithSuggestions').editor as any,
+    override: standardEditorsRegistry.get('textWithSuggestions').editor as any,
     process: stringOverrideProcessor,
     settings: {
       placeholder: 'none',
       expandTemplateVars: true,
```
ryantxu, line 58: can we use `expandTemplateVars` as a key to add the standard variables suggestions to whatever else is added?  or can/should we have another parameter `showTemplateVarSuggestions`

public/app/features/dashboard/components/PanelEditor/FieldConfigEditor.tsx
ryantxu: Yes, default behavior for templates! -- ideally just a flag to enable/disable since the standard variables should always be the same

ryantxu (commented):
Apologies in the time to review -- I like it!  General thoughts:
* do we need a different option/editor/settings?  can we just use the existing text one?
* standard templateVar support should be the same for everywhere.  In addition, I think we should revisit some of these objects and maybe move them off the root so we can use a proxy object... but that is a different issue :)
* for the core suggestions API, lets avoid PanelPlugin if possible -- same components should be possible for datasource, pages, and apps

torkelo (commented):
This is looking awesome! needs more work, and need to think how generic we want this ability to be (an option on all text fields?). 

But the big questions is where does interpolation happen? Currently for displayName this is part fieldOveride processing but for other things like the text panel I think the intepolation is up to the panel implementation to use the replaceVariables function correct? Should this be more automatic, not sure

packages/grafana-ui/src/utils/standardEditors.tsx
dprokop: Ha, this is a very good point, completely forgot we added this property a long time ago. Since this enforces variables interpolation I think it's a good move to use it for enabling the input with suggestions.

dprokop:
@torkelo @ryantxu thanks a lot for all the feedback! This is great. 

@ryantxu 
> do we need a different option/editor/settings? can we just use the existing text one?
I agree with you, we can use the `expandTemplateVariables` property for that as it ensures variables replacement.

> for the core suggestions API, lets avoid PanelPlugin if possible -- same components should be possible for datasource, pages, and apps
Good point. Major consideration here is this piece of code: https://github.com/grafana/grafana/pull/26021/files/9e47fa2f7f467740aaa73034293f5d02ba90d639 where we want to avoid presenting data source variables when _panel_ do not support queries. I assume we can handle this by checking if data frames are provided or not.

> standard templateVar support should be the same for everywhere. In addition, I think we should revisit some of these objects and maybe move them off the root so we can use a proxy object... but that is a different issue :)

So, the standard variables to provide, what would those be?:

1. Template variables & `__all_variables`
2. Built-in variables: time range(+ from & to)
3. **If data provided**: series & field variables
4. **If data & value scope provided**: value variables

packages/grafana-data/src/types/suggestions.ts
dprokop: nope, we can rely on data being provided or not IMO

packages/grafana-data/src/field/fieldOverrides.ts
dprokop: yes, just organize change, ctrl+o in webstorm.

packages/grafana-ui/src/components/OptionsUI/string.tsx
dprokop: It's to make sure `getSuggestions` is not evaluated on every render. `useMemo` memorizes the return value, not the function. But good catch actually, this memoization can lead to weird bugs, as it's not invalidated when i.e. context data change (and can be input for suggestions...)

PR #26024, opened 2020-07-02
Table: JSON Cell should try to convert strings to JSON

ryantxu:
When showing JSON strings in the table, we should try to parse them.

Before:
image

After:
image


ryantxu:
I marked as 7.1 since this will be super helpful for timestream... but can get pushed if it seems risky

dprokop (approved):
I think this looks OK and should be safe to include in 7.1 @ryantxu

PR #26036, opened 2020-07-02
@grafana/e2e: close options panel before interacting with the query form

stevenvachon:
... it's logically better, but the real reason is to appease Cypress which was cause a consistent request error for a single plugin (datadog-datasource). An error which could not be reproduced manually.


ryantxu (approved):

ryantxu:
minor and low risk -- should be added to 7.1

dprokop (approved):

PR #26037, opened 2020-07-02
Graph panel: Move Stacking and null values before Hover tooltip options (#26035)

### LINKED ISSUE #26035: Stacking and null values's Stack toggle moves after being clicked
**What happened**:
[stack shifts itself.mov.zip](https://github.com/grafana/grafana/files/4866554/stack.shifts.itself.mov.zip)
The Stacking and null value area shifts up or down causing the `Stack` widget to move up or down away from my cursor.

Settings should only influence settings below themselves, not settings 

1. Create a dashboard
1. Add a panel
1. Click Stack

**Environment**:
- Data source type & version:
- OS Grafana is installed on: (Linux)
- User OS & Browser: macOS / Chrome 83.0.4103.116
- Grafana plugins:
- Others:

jsoref:

Moves the `Stacking and null value` section before the `Hover tooltip` to fix UI behavior of `Stack` in `Stacking and null value`.

Fixes #26035

### DIFF
--- a/docs/sources/panels/visualizations/graph-panel.md
+++ b/docs/sources/panels/visualizations/graph-panel.md
@@ -32,6 +32,17 @@ Use these settings to refine your visualization.
 - **Points -** Display points for values.
 - **Point radius -** Controls how large the points are.
 
+### Stacking and null value
+
+- **Stack -** Each series is stacked on top of another.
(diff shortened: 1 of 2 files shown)


oddlittlebird:
Docs look good.

dprokop (approved):
@jsoref thanks for the contribution. This is a great UX improvement and it works as expected!

PR #26044, opened 2020-07-03
Instrumentation for total count of dashboard versions and annotations in the database

bergquist:
this should make it easier to monitor annotations/dashboard versions
missusage in hosted grafana.

### DIFF
--- a/pkg/infra/metrics/metrics.go
+++ b/pkg/infra/metrics/metrics.go
@@ -159,6 +159,12 @@ var (
 	// StatsTotalDataSources is a metric total number of defined datasources, labeled by pluginId
 	StatsTotalDataSources *prometheus.GaugeVec
 
+	// StatsTotalAnnotations is a metric of total number of annotations stored in Grafana.
+	StatsTotalAnnotations prometheus.Gauge
+
+	// StatsTotalDashboardVersions is a metric of total number of dashboard versions stored in Grafana.
+	StatsTotalDashboardVersions prometheus.Gauge
+
 	// grafanaBuildVersion is a metric with a constant '1' value labeled by version, revision, branch, and goversion from which Grafana was built
 	grafanaBuildVersion *prometheus.GaugeVec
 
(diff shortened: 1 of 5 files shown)


bergquist:
Would be nice if we could get this into 7.1 so we can find instances with too many dashboard_versions // annotations before its a problem.

pkg/infra/metrics/metrics.go
```diff
@@ -483,6 +489,18 @@ func init() {
 		Help:      "A metric with a constant '1' value labeled by pluginId, pluginType and version from which Grafana plugin was built",
 		Namespace: ExporterName,
 	}, []string{"plugin_id", "plugin_type", "version"})
+
+	StatsTotalDashboardVersions = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name:      "stat_totals_dashboard_versions",
```
aknuds1, line 494: Why "stats_totals" instead of "stats_total"? stat_total would correspond to the variable name. Could it be a misplaced "s"?

pkg/infra/metrics/metrics.go
```diff
@@ -483,6 +489,18 @@ func init() {
...
+		Help:      "total amount of dashboard versions in the database",
+		Namespace: ExporterName,
+	})
+
+	StatsTotalAnnotations = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name:      "stat_totals_annotations",
```
aknuds1, line 500: Same question as above.

pkg/infra/metrics/metrics.go
marefr: Follows same convention as earlier metrics like stat_totals_datasource and stat_totals_active_admins

marefr (approved):
LGTM

pkg/infra/metrics/metrics.go
aknuds1: @marefr Yeah I saw it in retrospect. What's the reason for this naming convention differing from the variable names though?

bergquist: I dont think there is any thoughts about the convention. Just sticking to it to make sure it's the same. 

While I agree with you that it should be named `stats`, renaming old metrics is too much work, and not using the same pattern will misalign our graphite metrics. https://github.com/grafana/grafana/blob/master/pkg/infra/metrics/graphitebridge/graphite.go#L62

bergquist:
FYI @DanCech

pkg/infra/metrics/metrics.go
aknuds1: Thanks for clarifying @bergquist.

PR #23174, opened 2020-03-29
Theme:s Tooltip fixes 

torkelo:
Minor improvements to tooltip colors aligning the help & standard tooltip color to be the same (needs more cleanup & unification but will do later when we move styles to emotion) 

Fixes  #23166

PR #26048, opened 2020-07-03
Fix: Make the JSON in JSONCell Tooltip more visible

### LINKED ISSUE #25897: Table: JSONTooltip not following theme
**What happened**:
I hovered over a tags cell in a table. A tooltip opened, but the background was in dark mode while my Grafana is in light mode.

Screenshot from 2020-06-29 14-11-15

The tooltip to follow Grafanas theme.

Create a table panel, choose TestDB as datasource and Grafana API as scenario. Hover over the  tags.

**Environment**:
- Grafana version: Current master
- Data source type & version: TestDB
- OS Grafana is installed on: Ubuntu 20.04
- User OS & Browser: Ubuntu 20.04 / Chrome Version 83.0.4103.116 (Official Build) (64-bit)
- Grafana plugins:
- Others:

peterholmberg:
https://github.com/grafana/grafana/pull/23174 introduced some styling updates to tooltip. When this is being used in the JSONCell Tooltip, the text is barely visible since the JSONFormatter thinks the background is white. This is a "quick" workaround to have the JSON more visible in the tooltip. This should probably be revised to a more elegant solution in the future.

Fixes #25897

Before: 

Screenshot from 2020-07-02 15-58-27

After: 

Screenshot from 2020-07-03 12-07-50


torkelo (commented):
This did not really work well, (and not at all for dark theme). 

Better to add Tooltip theme that is dark in dark theme and light gray in light theme

peterholmberg:
There was some leftover style on one of the divs @torkelo

torkelo:
This still looks bad I think in light theme  (dark theme also a bit since the inner color)

torkelo:
But not sure what the best solution is, think we need a new theme for the tooltip that is dark in dark theme and light gray in light theme

peterholmberg:
Yeah, it's not optimal but it was the short term solution we discussed at stand up iirc.

torkelo:
In dark theme it should use normal tooltip bg & shadow so that it stands out more. 

In light theme the shadow is a bit much I feel 
Screenshot from 2020-07-07 15-51-34

packages/grafana-ui/src/components/Table/JSONViewCell.tsx
```diff
@@ -41,12 +42,15 @@ interface PopupProps {
 }
 
 const JSONTooltip: FC<PopupProps> = props => {
-  const clazz = css`
-    padding: 10px;
+  const theme = useTheme();
+  const className = css`
+    padding: ${theme.spacing.xs};
   `;
```
dprokop, line 48: `useStyles` maybe?

packages/grafana-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -4,7 +4,7 @@ import { Popover } from './Popover';
 import { PopoverController, UsingPopperProps } from './PopoverController';
 
 export interface TooltipProps extends UsingPopperProps {
-  theme?: 'info' | 'error';
+  theme?: 'info' | 'error' | 'code';
```
dprokop, line 7: I woke up with a thought today, that maybe instead of being very explicit with `code` name, we could have `info-alt`?

dprokop (approved):
This looks good implementation wise, left few nits to consider. Also, I agree with @torkelo about the bg and box-shadows. In dark theme, it should use dark theme's tooltip default box-shadow, and the light theme should have some more subtle shadow:)

packages/grafana-ui/src/components/Tooltip/Tooltip.tsx
peterholmberg: Hm, yeah maybe. Just need to add some documentation about that

dprokop: There is one already: https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md#styling-with-theme

dprokop (approved):

PR #25567, opened 2020-06-12
Auth:  Add support for forcing authentication in anonymous mode and modify SignIn to use it instead of redirect

papagian:
**What this PR does / why we need it**:

* Forbids additional redirect cookie values
* Introduces `forceLogin` query parameter for forcing authentication in anonymous mode
* Modifies SignIn to set `forceLogin` instead of `redirect` query
* Modifies Login not to follow redirect query parameter

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

Probably there is a more appropriate way for passing `forceLogin` query parameter.
I have tried [unsuccessfully](https://github.com/grafana/grafana/commit/cbb8733d77e102672f1a002f4e69930d271d672e) to use `updateLocation` for setting `forceLogin` query parameter in SignIn but even though the url is updated in the address bar, it does not trigger a new request (for being handled by the backend middleware).

PR #26049, opened 2020-07-03
Auth: Fix POST request failures with anonymous access

papagian:
This fix does not use `context.QueryBool()` for getting `forceLogin` url parameter.

Macaron `context.QueryBool()` seems to modify the request context that causes the POST (and PUT) requests to fail with:
`http: proxy error: net/http: HTTP/1.x transport connection broken: http: ContentLength=333 with Body length 0`

This modification is introduced [here](https://github.com/grafana/grafana/pull/25567) and affects anonymous requests.

How to reproduce the issue:
- enable anonymous access by setting `GF_AUTH_ANONYMOUS_ENABLED=true` and `GF_AUTH_ANONYMOUS_ORG_ROLE="Viewer"`
- start graphite:
`make devenv sources=graphite11`
- import [this](https://gist.github.com/papagian/3d6fe23a7a46f0a4e58fcebd2b365c6d) dashboard
- signout
- access the dashboard anonymously
- you should see a number of failures like:
`Request URL:http://localhost:3000/api/datasources/proxy/1/render
EROR[07-03|11:37:54] Data proxy error                         logger=data-proxy-log userId=0 orgId=1 uname= path=/api/datasources/proxy/1/render remote_addr=[::1] referer="http://localhost:3000/d/000000012/grafana-play-home?orgId=1" error="http: proxy error: net/http: HTTP/1.x transport connection broken: http: ContentLength=333 with Body length 0"`

Fixes #

An alternatively would be to allow `forceLogin` only for GETs.

### DIFF
--- a/pkg/middleware/auth.go
+++ b/pkg/middleware/auth.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/url"
+	"strconv"
 	"strings"
 
 	macaron "gopkg.in/macaron.v1"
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):

marefr (approved):
LGTM

PR #26056, opened 2020-07-03
ThemeContext: Make useStyles type-aware

s-h-a-d-o-w:
Simply a minor improvement to `useStyles` that occurred to me as I wanted to use it.
The returned styles object won't be `any` any more but instead has the type of the styles one creates. => Autocomplete :tada:


dprokop (approved):
Thank you @s-h-a-d-o-w! Seems like e2e failed, but I've just re-run to see if it's a flaky test.

PR #26062, opened 2020-07-04
Panel Inspect: Hide minimap in json view

torkelo:
I have personally never find vscode's minimap useful, and find it more distracting than good. And in the inspect drawer it's blocking content that I rather see.  

thoughts? 

Screenshot from 2020-07-04 12-25-25


ryantxu:
Can we increase the length where it turns on?  I find it super useful when scrolling through super long lists.  Things like this:

image

torkelo:
> Can we increase the length where it turns on? 

Can we make it more depend on screen width?

torkelo:
But I see now that for large responses and many series, you see a pattern in the minimap that I usually fail to see (in code)

torkelo:
think it's fine, changed my mind :)

PR #26075, opened 2020-07-06
Chore: Replace scopelint with exportloopref

aknuds1:
Replace the deprecated scopelint Go linter with its suggested replacement [exportloopref](https://github.com/kyoh86/exportloopref). Also removing scopelint exclude rule from tests, since this linter doesn't have those false positives apparently.


scripts/go/configs/.golangci.toml
```diff
@@ -44,10 +44,6 @@ enable = [
 # "gocyclo",
 # "unparam"
 
-[[issues.exclude-rules]]
-path = "_test\\.go"
-linters = ["scopelint"]
-
```
papagian, line 50: Why this is removed? Is that intentional?

aknuds1: @papagian Yes it's intentional, because scopelint gets replaced with exportloopref, and the latter doesn't trigger on our tests (hopefully it's just more accurate).

papagian (approved):

PR #26039, opened 2020-07-02
Chore: do not bundle rxjs with @grafana/data

ryantxu:
Using `7.1.0 beta2`, building plugins was failing with cryptic error:

```
import("/home/ryan/workspace/grafana/plugins/drone-app/node_modules/                           rxjs/internal/Observable").Observable<import("/home/ryan/workspace/grafana/plugins/drone-app/node_modules/@grafana/data/types/datasource").DataQueryResponse>' is not assignable to type '
import("/home/ryan/workspace/grafana/plugins/drone-app/node_modules/@grafana/data/node_modules/rxjs/internal/Observable").Observable<import("/home/ryan/workspace/grafana/plugins/drone-app/node_modules/@grafana/data/types/datasource").DataQueryResponse>'.
```

If I manually remove the embedded rxjs, the build is fine:
image

maybe this will help?

PR #26084, opened 2020-07-06
Chore: Bump rxjs to latest

dprokop:
Fixes issue with plugins not being built with grafana-toolkit, described here: https://github.com/grafana/grafana/pull/26039

`concurrently`, one package that we are using, has rxjs dependency set to ^6.3.3. our dependency on rxjs is currently 6.5.5. Few days ago 6.6.0 was released causing grafana-data being dependant on 6.5.5 while concurrently on 6.6.0. Hence, two versions was installed when using grafana-data and grafana-toolkit, causing some weird type mismatch.

Changes in this PR ar verified against local NPM repository set up with https://verdaccio.org/

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,7 +267,7 @@
     "regenerator-runtime": "0.13.3",
     "reselect": "4.0.0",
     "rst2html": "github:thoward/rst2html#990cb89",
-    "rxjs": "6.5.5",
+    "rxjs": "6.6.0",
     "search-query-parser": "1.5.4",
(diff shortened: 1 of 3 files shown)


torkelo (approved):
LGTM

ryantxu (approved):
Thank you!  Should we also add a resolutions entry that forces the same version?

dprokop:
> Should we also add a resolutions entry that forces the same version?

Yes, we can do that in toolkit, will add.

PR #26086, opened 2020-07-06
Fix: Icon and Tooltip on Variables editor

### LINKED ISSUE #25137: v7.0.1 Minor tooltip/icon errors on Variables Edit page
**What happened**:
Very minor issues but thought I would raise anyway:

On the Variables -> Edit page for `Query` type variables the Multi-value hover tooltip "i" icon shows the correct text. But the tooltip text is duplicated for the `Include All option` so both tooltips are the same.

Also I notice that "i" icons on those two values also differ from the other "i" icons on the page. So on the same page some are svg icons and others are fontawesome icons. 

**Environment**:
- Grafana version: 7.0.1

peterholmberg:
Use the Icon component in Switch to have uniform :information_source: icons. Fix tooltip text.

Fixes #25137

### Before:
Screenshot from 2020-07-06 15-33-24

### After: 
Screenshot from 2020-07-06 15-32-28


torkelo (approved):
looks good!

PR #26095, opened 2020-07-06
Plugin CI Dockerfile: Upgrade golangci-lint, pin Alpine version, don't use apk cache

aknuds1:
In the plugin Alpine based CI Dockerfile:

* Upgrade golangci-lint
* Pin Alpine version, so we know which one we get
* Don't use apk cache, as is the best practice in Dockerfiles

If needed, I can split the PR up, but I noticed the other issues simply when looking to upgrade golangci-lint.


srclosson (approved):
LGTM

PR #26098, opened 2020-07-06
Azure: Restore Insights Metrics alias feature

kylebrandt:
fix feature removed by earlier work - the Alias ability in application insights.

This also fixes a case sensitivity issue in related the azure monitor metrics alias feature.

### DIFF
--- a/pkg/tsdb/azuremonitor/applicationinsights-datasource.go
+++ b/pkg/tsdb/azuremonitor/applicationinsights-datasource.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/tsdb/azuremonitor/applicationinsights-datasource.go
```diff
@@ -250,3 +254,68 @@ func (e *ApplicationInsightsDatasource) getPluginRoute(plugin *plugins.DataSourc
...
 }
+
+// formatApplicationInsightsLegendKey builds the legend key or timeseries name
+// Alias patterns like {{metric}} are replaced with the appropriate data values.
+func formatApplicationInsightsLegendKey(alias string, metricName string, labels data.Labels) string {
+
+	// backwards support for "ToLower" (eww). Could be a problem if there were
+	// two keys that varied only in case, but I don't think that would
+	// happen with azure.
+	lowerLabels := labels.Copy()
```
aknuds1, line 265: Might it be easier/safer to create an empty data.Labels object, that you copy entries into with lowercased keys? Unsure if it's safe to mutate an array you're iterating over.

aknuds1 (commented):
Generally looks good to me, but left a question.

pkg/tsdb/azuremonitor/applicationinsights-datasource.go
kylebrandt: Mutating a map while iterating is fine in Go. That being said, your suggestion is cleaner - will make the change.

aknuds1 (approved):

Issue #22955, opened 2020-03-23
Grafana AMD64 Docker image named conflictingly with manifest


aknuds1:
**What happened**:
The Grafana AMD64 architecture Docker image is named the same as the Grafana Docker manifest, i.e. grafana/grafana. This makes it impossible for Docker to discern between the manifest and the image, and it becomes for example impossible to recreate a certain Grafana Docker manifest, since Docker thinks the AMD64 image is a manifest: 

```
✗ docker manifest create grafana/grafana:latest grafana/grafana:6.7.1 grafana/grafana-arm32v7-linux:6.7.1 grafana/grafana-arm64v8-linux:6.7.1
docker.io/grafana/grafana:6.7.1 is a manifest list
```
I think that to avoid this issue, we need to name AMD64 Docker images according to the same convention as the ARM images, i.e. grafana/grafana-<arch>-linux, in order to have a unique name versus the manifest.

@xlson What do you think?

**NB:** 
After fixing this, revert #24280.

PR #26107, opened 2020-07-07
CircleCI: Check errors from publishing master Docker manifests

aknuds1:
In CircleCI, since fixing #22955, once again check errors from publishing of Docker manifests on master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -667,8 +667,7 @@ jobs:
               # This is a release
               /tmp/grabpl publish-docker --jobs 4 --edition << parameters.edition >> --ubuntu=<< parameters.ubuntu >>
             else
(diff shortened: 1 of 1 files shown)


sakjur (approved):
:+1:

papagian (approved):
LGTM, I'm not sure though which is preferred: the build to finish (not fail) and somebody publish the docker images manually or the build to succeed.
In the latter if one doesn't check the CI output can believe that the docker images are ok but all subsequent steps will be completed.
While in the former the docker image publishing and all the subsequent jobs should be completed manually (unless there are no other subsequent jobs).

aknuds1:
@papagian We would only ignore errors from publishing master manifests anyway. The manifest publishing problem should be gone now, since we fixed the underlying issue (#22955).

PR #26110, opened 2020-07-07
Grafana npm packages: Add guide and tooling for local packages registry setup

dprokop:
This PR adds guide and tooling to setup local npm registry and publish @grafana/* packages to it. Hopefully, this will help us avoid situations when broken packages are pushed to npm. Also, this approach makes it way easier to debug packages locally as you can test the actually published dist without the need to publish to npm.

I'm gonna present this approach during front-end weekly


torkelo (approved):
Nice!

mckn (approved):
Great stuff!

Issue #25842, opened 2020-06-25
Can't remove 5s database auto-refresh interval


Tomica-G:
**What happened**:
The 5s auto-refresh interval is still available for the dashboard after removing it from the dashboard configuration. All other default intervals can be removed.

This is happening on the 7.0.3 version. The version 6.7.4 works fine. It's been reproduced on play.grafana.org as well.

**What you expected to happen**:
That the 5S auto-refresh interval is no longer available after being deleted.

**How to reproduce it (as minimally and precisely as possible)**:
1. Open the dashboard settings screen
2. Remove the 5s interval from the Auto-refresh box in the General settings tab
3. Save the dashboard
4. Reload the dashboard and click on the auto-refresh interval picker. 5s is still available there.

**Anything else we need to know?**:
N/A

**Environment**:
- Grafana version: 7.0.3
- Data source type & version: N/A
- OS Grafana is installed on: CentOS7
- User OS & Browser: N/A
- Grafana plugins: N/A
- Others: N/A

marefr:
Please configure https://grafana.com/docs/grafana/latest/administration/configuration/#min-refresh-interval

PR #26126, opened 2020-07-07
Do not break dashboard settings UI when time intervals end with trailing comma

### LINKED ISSUE #26097: Auto-refresh interval trailing comma breaks Time Options
**What happened**:
Saving a dashboard with a trailing comma in Settings --> General --> Time Options --> Auto-refresh leads to the Time Options no longer displaying. The Time Options can be restored by removing the empty string and preceding comma from the JSON Model.

time_options_bug

Grafana should check to make sure the Auto-refresh entry is valid before saving.

1. Settings --> General --> Time Options --> Auto-refresh: 10s,30s,1m,5m
2. Enter a comma after 5m so Auto-refresh is now: 10s,30s,1m,5m;
3. Save dashboard
4. Enter dashboard settings and Time Options is no longer shown.

The following console errors are shown when loading the dashboard settings:

image

**Environment**:
- Grafana version: v7.0.4
- Data source type & version: Test Data DB (Random Walk) & InfluxDB
- OS Grafana is installed on: RHEL/CentOS
- User OS & Browser: MacOS & Chrome 83
- Grafana plugins: n/a
- Others: n/a

dprokop:
Fixes https://github.com/grafana/grafana/issues/26097

Done:
1. Time intervals input is controlled now and change is propagated on blur
2. SettingsCtrl makes sure saved refresh intervals do not contain empty values

### DIFF
--- a/public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
+++ b/public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
@@ -255,7 +255,7 @@ export class SettingsCtrl {
   };
 
   onRefreshIntervalChange = (intervals: string[]) => {
-    this.dashboard.timepicker.refresh_intervals = intervals;
+    this.dashboard.timepicker.refresh_intervals = intervals.filter(i => i.trim() !== '');
   };
 
   onNowDelayChange = (nowDelay: string) => {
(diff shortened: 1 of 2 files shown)


torkelo (approved):
Looks good, we have some other strange problems with the interval options, like
https://github.com/grafana/grafana/issues/25842

dprokop:
> Looks good, we have some other strange problems with the interval options, like #25842

@torkelo yeah, that one is really strange and I got confused as well:D But at the same time it's a consequence of the configuration, so the question is - do we want the user-configured intervals to override the config? Currently, it's an inclusive relation - whatever is configured in the `min_refresh_interval` lands in the refresh picker ¯\_(ツ)_/¯ 

I would expect and I think that whatever the user put into the refresh intervals settings should take precedence over any other settings.

PR #26137, opened 2020-07-07
Templating: Fixed recursive queries triggered when switching dashboard settings view

### LINKED ISSUE #25975: Uncaught (in promise) RangeError: Maximum call stack size exceeded in 7.0.4/7.05
I have complex dashboard containing large amount of template variables, tens of rows where each row might contain number amount of different panels, sometime with repeating rows. This worked perfectly in 6.4.3 and 7.0.3.

When same dashboard is run in 7.0.4 and I switch time range from last 1h to last 3h as example the Maximum call stack size exceeded error is thrown and dashboard completely freeze and/or never recover.

From Console I can one or error like below. 
runtime.js:423 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at new j (runtime.js:423)
    at Object.u [as wrap] (runtime.js:23)
    at Object.e (adapter.ts:29)
    at app.7b506073f46893460c0d.js:2
    at new Promise (<anonymous>)
    at Object.<anonymous> (app.7b506073f46893460c0d.js:2)
    at Object.setValueFromUrl (adapter.ts:29)
    at actions.ts:459
    at c (runtime.js:45)
    at Generator._invoke (runtime.js:271)

To narrow down the problem I removed all the rows and panels. Thus only template variables are left and it still failed when time range is changed. 
To further narrowed the issue down I removed half of the template variables and once I remove "variable type" prometheus adhoc" the time range change started to work.

But once AdHoc was removed the the first time range change works but when going to settings/variables result similar errors (seems that templates are updated) result.

runtime.js:263 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at Generator._invoke (runtime.js:263)
    at Generator.forEach.e.<computed> [as next] (runtime.js:97)
    at Ot (app.7b506073f46893460c0d.js:2)
    at i (app.7b506073f46893460c0d.js:2)
    at app.7b506073f46893460c0d.js:2
    at new Promise (<anonymous>)
    at Object.<anonymous> (app.7b506073f46893460c0d.js:2)
    at Object.setValue (adapter.ts:24)
    at actions.ts:239
    at c (runtime.js:45)

With full dashboard all queries will after second error like

tti-polyfill.js:4 GET http://address:port/api/v1/query?query=removedquery&time=1593601626&timeout=90s net::ERR_INSUFFICIENT_RESOURCES

Sounds that the is some major regression between 7.0.3 and 7.0.4 in template variable population which make Grafana 7.04 -> unusable.

**Environment**:
- Grafana version: 7.0.4
- User OS & Browser: Chrome
- Grafana plugins: None
- Others:

### LINKED ISSUE #26101: When viewing template variables, Grafana can easily get caught in a rapid, endless loop of trying to refresh every panel on the dashboard
This is 100% reproduceable on the live demo site/any dashboard with template variables. For example: https://play.grafana.org/d/000000002/influxdb-templated

**What happened**: When trying to edit template variables, Grafana started spamming out thousands of queries in an attempt to refresh every panel many times over (despite none of them being visible at the time).

**What you expected to happen**: For no refresh call to be sent out when in the dashboard edit view, the panels to register as not visible, and (failing all of that) for each panel to only try to refresh once.

**How to reproduce it (as minimally and precisely as possible)**: After changing a template variable value and seeing the URL update to reflect that change, go into the dashboard options page, which defaults you to "general", and then switch over to view "Variables". At this point the page will appear to hang, Dev Tools will reveal hundreds, or even thousands of requests being made to your datasource in rapid succession.

**Anything else we need to know?**: This is very easy to reproduce and seems very, very bad. Not only does it hang the browser, it hammers your datasources.

**Environment**:
- Grafana version: 7.0.5
- Data source type & version: Seen with ES and InfluxDB
- OS Grafana is installed on: Windows and whatever the Live demo site runs on

torkelo:
Fixes #26101 

Triggered by multiple issues
* The code that detects variable url changes with an ad-hoc variable had an empty array value  (which the angular state did not have)
* The URL links generated for the dashboard edit side nav incorrectly formatted query parameters, triggering template variable updates 

Might also fix #25975


public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
```diff
@@ -121,8 +120,12 @@ export class SettingsCtrl {
...
 
     for (const section of this.sections) {
-      const sectionParams = _.defaults({ editview: section.id }, params);
-      section.url = getConfig().appSubUrl + url + '?' + $.param(sectionParams);
+      section.url =
+        getConfig().appSubUrl +
+        urlUtil.renderUrl(url, {
+          ...params,
+          editview: section.id,
+        });
```
dprokop, line 128: Maybe you could use `locationUtil.assureBaseUrl` for this?

dprokop (approved):
works as expected, couldn't reproduce the issue anymore.

Left one nit to consider :)

public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
torkelo: yes, Updated!

PR #26142, opened 2020-07-08
Docs: use relative paths

jsoref:
This makes Markdown paths relative to where they are in the repository:

```shell
files=$(git ls-files |grep 'md$'|xargs grep -l '](/')
for a in $files; do dir=$(dirname $a); up=$(realpath . --relative-to=$dir); perl -pi -e 's{]\(/}{]\('$up'/}g' $a; done
```

Without this, local Markdown aware applications, such as **MacDown.app** can't follow the links -- because no one clones a foreign git repository into `/` on their computer.

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ We want to make Grafana accessible to even more people. Submit an [accessibility
 
 If you don't have the knowledge or time to code, consider helping with _issue triage_. The community will thank you for saving them time by spending some of yours.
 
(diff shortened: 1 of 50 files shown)


oddlittlebird:
Unfortunately, this will not work in our website. In the docs folder, the README instructs people to [use relrefs](https://github.com/grafana/grafana/tree/master/docs#using-relref-for-internal-links) in the documentation. This is because the documentation for this project is imported into another repo, where it is then built as part of the website. Adding a `.` would break our current functionality.

However, if you want to build locally, there are instructions for that in the README I linked above.

PR #26143, opened 2020-07-08
Provisioning: Remove provisioned dashboards without parental reader

### LINKED ISSUE #15716: Removing files for provisioned folders and dashboards does not delete them from Grafana's database
5.4.2
InfluxDb and MySQL
CentOS 7
### What did you do?
Provisioned dashboards using providers on `/etc/grafana/provisioning/dashboards` and creating the directories with JSON files inside them on the path configured in the YAML.
Restarted Grafana
Opened Grafana and saw that the dashboards were correctly provisioned.
Deleted the YAML file on `/etc/grafana/provisioning/dashboards` and the files on the configured path.
Restarted Grafana.
Dashboards and folders are no longer in Grafana
### What happened instead?
They were still there.

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

This PR adds deletion of provisioned dashboards, which have no link with provision reader configuration (or you can say, that they are orphan).

The deletion process starts on the Grafana load and occurs on every dashboard provision configuration reloading (requesting `/api/admin/provisioning/dashboards/reload` endpoint).

Fixes #15716

There are two problems:
* This realization ignores `providers[].disableDeletion` flag, because there is no information about previous file readers' settings in the database. In my opinion, to solve this problem we need to add a column to the `dashboard_provisioning` table and store the information about dashboards deletion possibility in it.
* We don't delete folders, because there is no information either folder provisioned or not.

I also wanted to mention that I tried to reuse existed code as much as possible.

### DIFF
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -74,7 +74,7 @@ func (provider *Provisioner) Provision() error {
 	return nil
 }
 
-// CleanUpOrphanedDashboards delete provisioned dashboards, which have no linked reader
+// CleanUpOrphanedDashboards deletes provisioned dashboards missing a linked reader.
 func (provider *Provisioner) CleanUpOrphanedDashboards() {
 	currentReaders := make([]string, len(provider.fileReaders))
 
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/grafana/grafana/pkg/bus"
 	"github.com/grafana/grafana/pkg/infra/log"
+	"github.com/grafana/grafana/pkg/models"
 	"github.com/grafana/grafana/pkg/util/errutil"
 )
 
@@ -16,6 +18,7 @@ type DashboardProvisioner interface {
 	PollChanges(ctx context.Context)
 	GetProvisionerResolvedPath(name string) string
 	GetAllowUIUpdatesFromConfig(name string) bool
+	CleanUpOrphanedDashboards()
 }
 
 // DashboardProvisionerFactory creates DashboardProvisioners based on input
@@ -69,6 +72,19 @@ func (provider *Provisioner) Provision() error {
 	}
 
 	return nil
+}
+
+// CleanUpOrphanedDashboards delete provisioned dashboards, which have no linked reader
+func (provider *Provisioner) CleanUpOrphanedDashboards() {
+	currentReaders := make([]string, len(provider.fileReaders))
+
+	for index, reader := range provider.fileReaders {
+		currentReaders[index] = reader.Cfg.Name
+	}
+
+	if err := bus.Dispatch(&models.DeleteOrphanedProvisionedDashboardsCommand{ReaderNames: currentReaders}); err != nil {
+		provider.log.Warn("Failed to delete orphaned provisioned dashboards", "err", err)
+	}
 }
 
 // PollChanges starts polling for changes in dashboard definition files. It creates goroutine for each provider
--- a/pkg/services/provisioning/provisioning.go
+++ b/pkg/services/provisioning/provisioning.go
@@ -139,14 +139,16 @@ func (ps *provisioningServiceImpl) ProvisionDashboards() error {
 		return errutil.Wrap("Failed to create provisioner", err)
 	}
 
+	// If we fail to provision with the new provisioner, mutex will unlock and the polling we restart with the
+	// old provisioner as we did not switch them yet.
 	ps.mutex.Lock()
 	defer ps.mutex.Unlock()
 
 	ps.cancelPolling()
+	dashProvisioner.CleanUpOrphanedDashboards()
 
-	if err := dashProvisioner.Provision(); err != nil {
-		// If we fail to provision with the new provisioner, mutex will unlock and the polling we restart with the
-		// old provisioner as we did not switch them yet.
+	err = dashProvisioner.Provision()
+	if err != nil {
 		return errutil.Wrap("Failed to provision dashboards", err)
 	}
 	ps.dashboardProvisioner = dashProvisioner
--- a/pkg/services/provisioning/provisioning.go
+++ b/pkg/services/provisioning/provisioning.go
@@ -139,8 +139,6 @@ func (ps *provisioningServiceImpl) ProvisionDashboards() error {
 		return errutil.Wrap("Failed to create provisioner", err)
 	}
 
-	// If we fail to provision with the new provisioner, the mutex will unlock and the polling will restart with the
-	// old provisioner as we did not switch them yet.
 	ps.mutex.Lock()
 	defer ps.mutex.Unlock()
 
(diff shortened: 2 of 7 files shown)


bergquist:
Nice! :) :fireworks:

I won't have time to review this as I'll be on vacation for a while but I'll make sure to help someone catching up on provisioning to be able to review this.

aknuds1:
Do you mind merging this with master, in order to include Drone CI configuration?

pkg/services/provisioning/dashboards/dashboard.go
```diff
@@ -71,6 +74,19 @@ func (provider *Provisioner) Provision() error {
 	return nil
 }
 
+// CleanUpOrphanedDashboards delete provisioned dashboards, which have no linked reader
```
aknuds1, line 77:
```suggestion
// CleanUpOrphanedDashboards deletes provisioned dashboards missing a linked reader.
```

pkg/services/provisioning/provisioning.go
```diff
@@ -136,14 +136,16 @@ func (ps *provisioningServiceImpl) ProvisionDashboards() error {
 		return errutil.Wrap("Failed to create provisioner", err)
 	}
 
+	// If we fail to provision with the new provisioner, mutex will unlock and the polling we restart with the
```
aknuds1, line 139:
```suggestion
	// If we fail to provision with the new provisioner, the mutex will unlock and the polling will restart with the
```

aknuds1: Why are you moving this comment? It looks to me as if it makes since in the old location.

pkg/services/sqlstore/dashboard_provisioning.go
```diff
@@ -88,3 +89,28 @@ func UnprovisionDashboard(cmd *models.UnprovisionDashboardCommand) error {
...
+		}
+
+		err := sess.NotIn("name", convertedReaderNames...).Find(&result)
+		if err != nil {
+			return err
+		}
+
+		for _, deleteDashCommand := range result {
+			err := deleteDashboard(&models.DeleteDashboardCommand{Id: deleteDashCommand.DashboardId}, sess)
+			if err != nil && err != models.ErrDashboardNotFound {
```
aknuds1, line 109: I think it's better to detect an error type with [errors.Is](https://golang.org/pkg/errors/#Is), since it will handle wrapped errors:
```suggestion
			if err != nil && !errors.Is(err, models.ErrDashboardNotFound) {
```

pkg/services/sqlstore/dashboard_provisioning.go
```diff
@@ -10,6 +10,7 @@ func init() {
 	bus.AddHandler("sql", SaveProvisionedDashboard)
 	bus.AddHandler("sql", GetProvisionedDataByDashboardId)
 	bus.AddHandler("sql", UnprovisionDashboard)
+	bus.AddHandler("sql", DeleteOrphanedProvisionedDashboard)
```
aknuds1, line 13:
```suggestion
	bus.AddHandler("sql", DeleteOrphanedProvisionedDashboards)
```

pkg/services/sqlstore/dashboard_provisioning.go
```diff
@@ -88,3 +89,28 @@ func UnprovisionDashboard(cmd *models.UnprovisionDashboardCommand) error {
 	}
 	return nil
 }
+
+func DeleteOrphanedProvisionedDashboard(cmd *models.DeleteOrphanedProvisionedDashboardsCommand) error {
```
aknuds1, line 93:
```suggestion
func DeleteOrphanedProvisionedDashboards(cmd *models.DeleteOrphanedProvisionedDashboardsCommand) error {
```

pkg/services/sqlstore/dashboard_provisioning_test.go
```diff
@@ -54,6 +54,43 @@ func TestDashboardProvisioningTest(t *testing.T) {
...
+				err := SaveProvisionedDashboard(anotherCmd)
+				So(err, ShouldBeNil)
+
+				query := &models.GetDashboardsQuery{DashboardIds: []int64{anotherCmd.Result.Id}}
+				err = GetDashboards(query)
+				So(err, ShouldBeNil)
+				So(query.Result, ShouldNotBeNil)
+
+				deleteCmd := &models.DeleteOrphanedProvisionedDashboardsCommand{ReaderNames: []string{"default"}}
+				So(DeleteOrphanedProvisionedDashboard(deleteCmd), ShouldBeNil)
```
aknuds1, line 84:
```suggestion
				So(DeleteOrphanedProvisionedDashboards(deleteCmd), ShouldBeNil)
```

aknuds1 (changes requested):
Looks generally good AFAICT, but I saw some things that should be fixed.

pkg/services/sqlstore/dashboard_provisioning.go
nabokihms: You right, it looks better.

pkg/services/provisioning/provisioning.go
nabokihms: Unfortunately, I can't remember now. Looks like it's better to move it to the old place.

nabokihms:
@aknuds1 I applied your suggestions, moved commentary, and released the current branch on the master. The drone pipeline passed.

aknuds1:
Thanks @nabokihms.

aknuds1 (approved):
Looks good AFAICT, but I'll let @bergquist have his say since I'm new to the functionality.

marefr (approved):
Tested it and works as expected. LGTM

pkg/services/sqlstore/dashboard_provisioning.go
```diff
@@ -88,3 +91,28 @@ func UnprovisionDashboard(cmd *models.UnprovisionDashboardCommand) error {
...
+			convertedReaderNames[index] = readerName
+		}
+
+		err := sess.NotIn("name", convertedReaderNames...).Find(&result)
+		if err != nil {
+			return err
+		}
+
+		for _, deleteDashCommand := range result {
+			err := deleteDashboard(&models.DeleteDashboardCommand{Id: deleteDashCommand.DashboardId}, sess)
```
bergquist, line 110: Would be great if we could move this deletion loop outside the `inTransaction` wrapper. This could create very large transactions without any clear upsides. I suggest we query for potential dashboards to delete without a transaction and then delete each dashboard in its own transition.

bergquist (commented):
Very sorry for the slow response. I dropped this completely. Things have been upside down since I got back from vacation :) Any chance you could address the minor comment in this PR? Otherwise I'll do it before 7.2 stable. 

Cheers for your great work!

pkg/services/sqlstore/dashboard_provisioning.go
nabokihms: Yes, it makes sense. I have changed the `DeleteOrphanedProvisionedDashboards` function behavior according to your suggestions. To make CI works, I also rebased the branch onto master.

bergquist (approved):
LGTM! 

The folder still says around in the database but that can be solved in another PR. Thank you for contributing and your patience :)

PR #26147, opened 2020-07-08
DataLinks: Fixed interpolation of repeated variables used in Graph data links

### LINKED ISSUE #25881: 7.0.4 Data Link repeated variable is not interpolated for the current value but instead shows "All"
I have a chart panel repeated for some variable.
I use data-links to navigate to some drill down dashboard.
When the variable shows "All" (so that the panel is repeated), the data-link interpolates the $variable to "all". Only when I chose a specific value for that variable (and then the "repeat" is "off") the variable is interpolated correctly for the data-link.

I also tried to repeat a "row" and store the panel as non-repeated in the row, and still I get the same results.

torkelo:
Fixes #25881


dprokop (approved):
LGTM

PR #26150, opened 2020-07-08
Dashboard: Allow removing min refresh interval from refresh options (5s or other)

### LINKED ISSUE #25842: Can't remove 5s database auto-refresh interval

torkelo:
Fixes #25842

### DIFF
--- a/public/app/features/dashboard/services/TimeSrv.ts
+++ b/public/app/features/dashboard/services/TimeSrv.ts
@@ -79,12 +79,7 @@ export class TimeSrv {
       return intervals;
     }
 
-    const validIntervals = intervals.filter(str => str !== '').filter(this.contextSrv.isAllowedInterval);
-
-    if (validIntervals.indexOf(this.contextSrv.minRefreshInterval) === -1) {
-      validIntervals.unshift(this.contextSrv.minRefreshInterval);
-    }
-    return validIntervals;
+    return intervals.filter(str => str !== '').filter(this.contextSrv.isAllowedInterval);
   }
 
   private parseTime() {


public/app/features/dashboard/services/TimeSrv.ts
```diff
@@ -79,12 +79,7 @@ export class TimeSrv {
       return intervals;
     }
 
-    const validIntervals = intervals.filter(str => str !== '').filter(this.contextSrv.isAllowedInterval);
-
-    if (validIntervals.indexOf(this.contextSrv.minRefreshInterval) === -1) {
-      validIntervals.unshift(this.contextSrv.minRefreshInterval);
```
torkelo, line 85: not sure why this logic was added here, @marefr do you know?

dprokop (approved):
This looks OK

public/app/features/dashboard/services/TimeSrv.ts
marefr: No, not sure. If I understand it correctly this adds the min interval to valid intervals if it doesn't exists? Yeah that sounds wrong to me.

marefr (approved):
LGTM

PR #26158, opened 2020-07-08
Grafana-UI: Add time range input

Clarity-89:

This PR adds a separate `TimeRangeInput` component, which uses existing `TimeRangePicker` as a base. The idea is to repurpose `TimeRangePicker` for use in forms with minimal changes.

The difference with `TimeRangePicker`:
- Removed range controls and zoom buttons
- Form input styles have been applied to the range label
- Position of the picker dropdown has been adjusted
- Time range history has been removed (via `hideHistory` prop for the `TimeRangeContent`)

Changes to the `TimeRangePicker`:
- Interacting with the picker and selecting the ranges doesn't cause page reload (due to the internal form submit)
- The selected values of the TimeRangeForm have been synchronized with the `value` prop for immediate feedback to the absolute range picker when selecting a relative range

Screenshots:

Screenshot 2020-07-08 at 14 31 00
Screenshot 2020-07-08 at 14 30 49


packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { getFocusStyle } from '../Forms/commonStyles';
+import { TimePickerButtonLabel } from './TimeRangePicker';
+import { TimePickerContent } from './TimeRangePicker/TimePickerContent';
+import { otherOptions, quickOptions } from './rangeOptions';
+
+export interface Props {
+  value: TimeRange;
+  timeZone?: TimeZone;
+  onChange: (timeRange: TimeRange) => void;
+  onChangeTimeZone: (timeZone: TimeZone) => void;
```
dprokop, line 17: Wonder if the input should support time zone change by default. Is this something that's needed in your use case? It feels to me like the time zone change feature is intentional rather than default, and possibility to change it was drive by certain needs for quick access in dashboard.

dprokop:
I see a small UI issue when the "input" is short:

image

Caret icon is overlapping the label.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  const onRangeChange = (timeRange: TimeRange) => {
+    onClose();
+    onChange(timeRange);
+  };
+
+  const styles = useStyles(getStyles);
+  const theme = useTheme();
+  const inputStyles = getInputStyles({ theme, invalid: false });
```
dprokop, line 40: Two things: 
1. Can we keep hooks at the top level of the function?
2. Can the `inputStyles` become part of the getStyles function?

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+    content: css`
+      margin-left: 0;
+    `,
+    pickerInput: css`
+      display: flex;
+      align-items: center;
+      justify-content: space-between;
+      cursor: pointer;
+
+      :focus {
```
dprokop, line 88:
```
      :focus {
```


hmm, don't you have to do `&:focus` for this to work?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker.tsx
```diff
@@ -225,21 +191,23 @@ const TimePickerTooltip = ({ timeRange, timeZone }: { timeRange: TimeRange; time
   );
 };
 
-const TimePickerButtonLabel = memo<Props>(({ hideText, value, timeZone }) => {
-  const theme = useTheme();
-  const styles = getLabelStyles(theme);
+export const TimePickerButtonLabel = memo<Pick<Props, 'hideText' | 'value' | 'timeZone'>>(
```
dprokop, line 194: Can you definee the picked type outside of the memo's generic type?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -134,6 +134,8 @@ interface Props {
   quickOptions?: TimeOption[];
   otherOptions?: TimeOption[];
   history?: TimeRange[];
+  hideHistory?: boolean;
```
dprokop, line 137: Think it's more relevant to have `showHistory` instead, as it's relevant in the dashboard context, and most probably not so much in other parts of the UI.

dprokop (approved):
Few nits, but other than that it looks great!

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
Clarity-89: Yeah, makes sense to make it optional.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
Clarity-89: Good catch! The `focus` is actually redundant here since it comes from `getFocusStyle`.

PR #26159, opened 2020-07-08
Update TextArea.mdx

oanamangiurea:
Add additional information in docs to textarea component

Fixes #


peterholmberg (approved):
Looks good

Clarity-89 (approved):

jessover9000:
@oanamangiurea  Nice work! Am I understanding it right that without the "textArea" part you cannot have form validation? because if that's true, it should be more explicit in the text.

oanamangiurea:
@jessover9000 Good point. You can not have form validation only if it's a required field, like for input. I'll update the text here and for input as well to make more sense.

PR #26161, opened 2020-07-08
Explore: Run query on splitOpen action

aocenas:
When opening a split with some existing query, rung that query so that when clicking on internal link you already see what you want instead of having to run the query manually.

ie state before when clicking on link
Screenshot from 2020-07-08 15-34-26

This seems to be a side effect of https://github.com/grafana/grafana/pull/26033.

### DIFF
--- a/public/app/features/explore/ExploreToolbar.tsx
+++ b/public/app/features/explore/ExploreToolbar.tsx
@@ -91,7 +91,7 @@ type Props = StateProps & DispatchProps & OwnProps;
 
 export class UnConnectedExploreToolbar extends PureComponent<Props> {
   onChangeDatasource = async (option: { value: any }) => {
-    this.props.changeDatasource(this.props.exploreId, option.value);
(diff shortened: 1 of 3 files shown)


ivanahuckova (approved):
LGTM! I really like your solution with adding `options?: { importQueries: boolean }` to make `changeDatasource` function more understandable (what is going on inside) and to prevent it from importing queries for split screen.

zoltanbedi (approved):
This looks good to me. Just one thing that comes to my mind. Is awaiting dispatch makes sense here? https://github.com/grafana/grafana/pull/26161/files#diff-a76fddf8e7a5790d979f2319911933f2R158

aocenas:
@zoltanbedi tbh I have no idea right now. I see it sparkled over the explore action and not sure if they solve some race condition issues or not. In general I think it does something ie thunk middleware just returns the result of the action function so you would get promise out of dispatch you can await for. So using await on dispatch should not no op imho.

But yeah I also think that is a bit of code smell that's one of the reasons I asked the question about explore refactor in the slack channel.

PR #26176, opened 2020-07-08
BarGauge: Fix space bug in single series mode

torkelo:
Fixes bug introduced by the stat text mode option PR. 

In single series mode the Bar Gauge still made space for title


dprokop (approved):
great!

PR #26178, opened 2020-07-08
Explore/Loki: POC for toggling parsed fields in the list view

fredr:
This is a working POC and a proposal on toggling what parsed fields to display in the Explore/Loki view. But it is only meant as a discussion starter.

This pr adds a toggle button in the Log details view, where its is possible to toggle one or more parsed fields to be displayed instead of the raw log message.

Our logs are pretty big json blobs (50ish fields), and json log rows are quite hard to read in general, so this would improve the usability a lot.

This was [discussed briefly on slack](https://grafana.slack.com/archives/CSKQV9Y5B/p1594137083079000?thread_ts=1594137083.079000&cid=CSKQV9Y5B) today between @carlpett, @jessover9000 and @aocenas but I'm guessing more people could be interested in this

toggle-parsed-fields

### DIFF
--- a/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
+++ b/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
@@ -18,11 +18,11 @@ export interface Props extends Themeable {
   isLabel?: boolean;
   onClickFilterLabel?: (key: string, value: string) => void;
   onClickFilterOutLabel?: (key: string, value: string) => void;
-  onClickShowParsedField?: (key: string) => void;
-  onClickHideParsedField?: (key: string) => void;
   links?: Array<LinkModel<Field>>;
   getStats: () => LogLabelStatsModel[] | null;
   showParsedFields?: string[];
+  onClickShowParsedField?: (key: string) => void;
+  onClickHideParsedField?: (key: string) => void;
 }
 
 interface State {
--- a/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
+++ b/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
@@ -20,6 +20,7 @@ export interface Props extends Themeable {
   onClickFilterOutLabel?: (key: string, value: string) => void;
   links?: Array<LinkModel<Field>>;
   getStats: () => LogLabelStatsModel[] | null;
+  showParsedFields: Array<string>;
 }
 
 interface State {
@@ -52,6 +53,20 @@ class UnThemedLogDetailsRow extends PureComponent<Props, State> {
     showFieldsStats: false,
     fieldCount: 0,
     fieldStats: null,
+  };
+
+  showField = () => {
+    const { onClickShowParsedField, parsedKey } = this.props;
+    if (onClickShowParsedField) {
+      onClickShowParsedField(parsedKey);
+    }
+  };
+
+  hideField = () => {
+    const { onClickHideParsedField, parsedKey } = this.props;
+    if (onClickHideParsedField) {
+      onClickHideParsedField(parsedKey);
+    }
   };
 
   filterLabel = () => {
@@ -87,14 +102,21 @@ class UnThemedLogDetailsRow extends PureComponent<Props, State> {
   }
 
   render() {
-    const { theme, parsedKey, parsedValue, isLabel, links } = this.props;
+    const { theme, parsedKey, parsedValue, isLabel, links, showParsedFields } = this.props;
     const { showFieldsStats, fieldStats, fieldCount } = this.state;
     const styles = getStyles(theme);
     const style = getLogRowStyles(theme);
+    const toggleFieldButton =
+      !isLabel && showParsedFields.includes(parsedKey) ? (
+        <IconButton name="eye-slash" title="Hide field" onClick={this.hideField} />
+      ) : (
+        <IconButton name="eye" title="Show field" onClick={this.showField} />
+      );
+
     return (
       <tr className={cx(style.logDetailsValue, { [styles.noHoverBackground]: showFieldsStats })}>
         {/* Action buttons - show stats/filter results */}
-        <td className={style.logsDetailsIcon} colSpan={isLabel ? undefined : 3}>
+        <td className={style.logsDetailsIcon} colSpan={isLabel ? undefined : 2}>
           <IconButton name="signal" title={'Ad-hoc statistics'} onClick={this.showStats} />
         </td>
 
@@ -106,6 +128,12 @@ class UnThemedLogDetailsRow extends PureComponent<Props, State> {
             <td className={style.logsDetailsIcon}>
               <IconButton name="search-minus" title="Filter out value" onClick={this.filterOutLabel} />
             </td>
+          </>
+        )}
+
+        {!isLabel && (
+          <>
+            <td className={style.logsDetailsIcon}>{toggleFieldButton}</td>
           </>
         )}
 
--- a/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
+++ b/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
@@ -18,8 +18,8 @@ export interface Props extends Themeable {
   isLabel?: boolean;
   onClickFilterLabel?: (key: string, value: string) => void;
   onClickFilterOutLabel?: (key: string, value: string) => void;
-  onClickShowParsedField?: (key: String) => void;
-  onClickHideParsedField?: (key: String) => void;
+  onClickShowParsedField?: (key: string) => void;
+  onClickHideParsedField?: (key: string) => void;
   links?: Array<LinkModel<Field>>;
   getStats: () => LogLabelStatsModel[] | null;
   showParsedFields?: string[];
--- a/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
+++ b/packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
@@ -18,9 +18,11 @@ export interface Props extends Themeable {
   isLabel?: boolean;
   onClickFilterLabel?: (key: string, value: string) => void;
   onClickFilterOutLabel?: (key: string, value: string) => void;
+  onClickShowParsedField?: (key: String) => void;
+  onClickHideParsedField?: (key: String) => void;
   links?: Array<LinkModel<Field>>;
   getStats: () => LogLabelStatsModel[] | null;
-  showParsedFields: Array<string>;
+  showParsedFields?: string[];
 }
 
 interface State {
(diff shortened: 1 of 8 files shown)


leahneukirchen:
Pretty-printing multiline JSON strings would be very relevant to me. (Think backtraces.)

aocenas:
This looks pretty neat. Thanks for the contribution @fredr. This needs a bit of a discussion and there are lots of vacation going on so it may be slow.

For me personally this looks great and still seems "ad-hoc enough" to be matching the explore usecase without any heavy configuration that user would probably hate to loose. We may need some smaller adjustments like some indicator that message is filtered and also figure out what to do in dashboards (the detail view is not there when using logs in dashboard panel).

@jessover9000 @davkal what do you think?

fredr:
> This looks pretty neat. Thanks for the contribution @fredr. This needs a bit of a discussion and there are lots of vacation going on so it may be slow.

Thanks for the feedback, and no worries on slow replies.

I'm happy to continue on this pr if you guys decide this should be implemented, or if you rather implement it some other way I would not take any offense if you scraped it. Just let me know!

:beers:

jessover9000:
Hi! Thank for this effort, it's simple but effective and already looking pretty good. I agree with Andrej that the interface needs feedback that tells the user that they filtered something. I think it would be nice to be able to quick-remove/unhide these hidden fields again. So I quickly threw together what it could look like in Figma: https://www.figma.com/file/7ZHtMoaK4RxBz3b6m6U0cW/Show-more-logs-Sorting?node-id=136%3A278

Regarding dashboards, I'm not sure whether in a dashboard logs panel you would want to toggle showing/hiding fields a lot, or whether you'd just like to hide some once and keep them hidden for easy reading. I don't know how these filters work on the technical side. I wonder if it would be useful if the filters were part of the query and not just visual? because then you could share queries with pre-hidden stuff, and I imagine the hidden fields would automatically be applied in a dashboard view. If we can't do that, we need something in the panel edit to tweak the parsed fields. Maybe as a Transformation or in the "Display" options area?

aocenas:
> Regarding dashboards, I'm not sure whether in a dashboard logs panel you would want to toggle showing/hiding fields a lot, or whether you'd just like to hide some once and keep them hidden for easy reading....

My 2 cents to this: We still do not show the row details in dashboard, not sure if that's going to change near term or not (no immediate plans as far as I know). I would probably be against having it in the query itself, as it's a feature of the logs panel not the datasource (should not be a problem to make it generic for all logging datasources with parseable log lines) so for me in case of dashboard it would make more sense to have it in the visualisation options.

fredr:
> Hi! Thanks for this effort, it's simple but effective and already looking pretty good. 

Thanks for giving feedback!

> I agree with Andrej that the interface needs feedback that tells the user that they filtered something. I think it would be nice to be able to quick-remove/unhide these hidden fields again.

I agree, that would be very useful, because right now, if you change your query while having toggled fields, and the new query doesn't have those fields, you wont be able to remove the toggles and get back to the raw log row.

> Regarding dashboards, I'm not sure whether in a dashboard logs panel you would want to toggle showing/hiding fields a lot, or whether you'd just like to hide some once and keep them hidden for easy reading. I don't know how these filters work on the technical side. I wonder if it would be useful if the filters were part of the query and not just visual? because then you could share queries with pre-hidden stuff, and I imagine the hidden fields would automatically be applied in a dashboard view. If we can't do that, we need something in the panel edit to tweak the parsed fields. Maybe as a Transformation or in the "Display" options area?

I haven't used the dashboard logs panel yet, so I don't have a good idea of how that should play into this. 

But currently, in this POC, the toggles are only saved in the local state, so if you copy the link and open it in an other browser, they will get lost.

IMO it would be a good idea to add them to the url query as you suggest, so that it will be preserved when sending a log explore link to someone. This is something we do a lot whenever there is an incident, or when we want to share some insight. I think it might be confusing if they are not preserved.

I guess this way it could also be saved to the query history, which I think would also be expected.

fredr:
I've updated the POC with a indicator that there are filters, and also a clear filter button

davkal (changes requested):
There are some changes to LogQL coming that should make it easer to "unpack" certain values within JSON log lines. But I think this is a great feature nonetheless. I suggest a couple of things:

Blocker:

- `parseMessage` that is used in `LogRowMessageParsed` is the same as in `LogDetails`, please reuse that one (there was already a fix done in `LogDetails` recently that would have not made it into this new feature if we didnt reuse the code).
- The eye metaphor is not quite fitting. It's not show/hide, but rather select/deselect, plus a base state where nothing is selected which means that everything is shown unparsed. Select/deselect are usually done with checkbox icons, but I think the eye is not so bad after all if we consider this as setting field visibility. I think what I don't like is the crossed eye (hide) semantic. I'd rather do something like a blue eye, ie if a field was selected its eye gets blue, and then it can be unselected to turn it normal gray again. And the tooltips should be adapted to accurately describe what's happening: Something like "Show this field instead of the message" to activate and "Hide this field" to deselect.
- the status message about the selection needs to be cleaned up, either as part of the controls (A), or as part of the meta section where it would make sense to also specify which fields are being shown (B)

A)
Screenshot 2020-08-14 at 17 14 44

B) 
Screenshot 2020-08-14 at 17 18 52


Minor:

- IMO empty rows should not be rendered.

Otherwise, awesome work! 🚀 
Let me know how I can help.

fredr:
> There are some changes to LogQL coming that should make it easer to "unpack" certain values within JSON log lines. But I think this is a great feature nonetheless. I suggest a couple of things:

Thanks for this feedback!

>     * `parseMessage` that is used in `LogRowMessageParsed` is the same as in `LogDetails`, please reuse that one (there was already a fix done in `LogDetails` recently that would have not made it into this new feature if we didnt reuse the code).

I've moved the `parseMessage` to a common place, so we only have one 

>     * The eye metaphor is not quite fitting. It's not show/hide, but rather select/deselect, plus a base state where nothing is selected which means that everything is shown unparsed. Select/deselect are usually done with checkbox icons, but I think the eye is not so bad after all if we consider this as setting field visibility. I think what I don't like is the crossed eye (hide) semantic. I'd rather do something like a blue eye, ie if a field was selected its eye gets blue, and then it can be unselected to turn it normal gray again. And the tooltips should be adapted to accurately describe what's happening: Something like "Show this field instead of the message" to activate and "Hide this field" to deselect.

I changed it to be blue when selected, I did not know what blue I should use, so I used blue95, but is there a better way of doing that, or a more general value to set it to? 

>     * the status message about the selection needs to be cleaned up, either as part of the controls (A), or as part of the meta section where it would make sense to also specify which fields are being shown (B)

I've now added them to the meta section, as I think its nice to display what fields are selected there

>     * IMO empty rows should not be rendered.

I think its kind of nice to be able to expand the rows that did not include any of the selected fields to be able to check what that was, and potentially add fields from those as well (thats kind of how it works in kibana which is what I use today). But if you feel strongly about it, I'll change it. I guess I would have to parse the rows in `LogRows` then instead

> Otherwise, awesome work! rocket
> Let me know how I can help.

I'm not really a frontend developer, so feel free to be super nitpicky on this pr, I've probably done some things weirdly

packages/grafana-ui/src/components/Logs/LogDetailsRow.tsx
```diff
@@ -44,6 +47,9 @@ const getStyles = stylesFactory((theme: GrafanaTheme) => {
       label: wordBreakAll;
       word-break: break-all;
     `,
+    showingField: css`
+      color: ${theme.palette.blue95};
```
davkal, line 51: @jessover9000 what's a good color here to highlight an icon as active (I've said blue) and how should the color be referenced?

davkal: Example: Screenshot 2020-08-18 at 15 14 57

public/app/features/explore/Logs.tsx
```diff
@@ -290,6 +323,25 @@ export class Logs extends PureComponent<Props, State> {
           />
         )}
 
+        {showParsedFields && showParsedFields.length > 0 && (
+          <MetaInfoText
+            metaItems={[
+              {
+                label: 'Showing only parsed labels',
```
davkal, line 330: This should be "parsed fields".

davkal (commented):
This is really shaping up. I've demoed this internally and got very positive feedback. Most users also prefer to see the empty rows even when they don't have any selected field to show, so let's leave this as is.
I'm trying to find out how to best apply the color (it's something we're not doing anywhere else yet). Probably ready to merge tomorrow.

public/app/features/explore/Logs.tsx
fredr: good catch!

fredr:
> This is really shaping up. I've demoed this internally and got very positive feedback. Most users also prefer to see the empty rows even when they don't have any selected field to show, so let's leave this as is.
> I'm trying to find out how to best apply the color (it's something we're not doing anywhere else yet). Probably ready to merge tomorrow.

Excellent, glad to hear! Let me know if I should change anything, or squash the commits or something.

davkal (approved):

davkal:
Awesome contribution @fredr!! 🚀

fredr:
🍻 Cheers! Thanks for merging

gordonjb:
@fredr I just built and checked that PR, and yes it does now work on my parsed fields! Great work btw, we were contemplating bodging something like this ourselves and you put this PR up about 2 days later :)

fredr:
> @fredr I just built and checked that PR, and yes it does now work on my parsed fields! Great work btw, we were contemplating bodging something like this ourselves and you put this PR up about 2 days later :)

Thanks for testing, glad it worked! And thank you :) We are in the process of switching over to loki/grafana from es/kibana, and this was the biggest blocker for us

cloudcafetech:
@fredr 

I am using Jaeger HotROD application in Kubernetes ...

My logs are as follows ...

image

and Derived fields configuration as follows ..

image

But I am not able to find ```trace_id``` in **parsed fields** section (in first image) loki explorer when querying.

Am I missing anything or do I need any additional instrumentation ?

fredr:
@cloudcafetech Your regex expects the format `trace_id=id` or `traceID=id` and in the log message that I can see in the screenshot it is in the format `"trace_id": "id"`, so maybe you need to change your regex? Something like `"(?:trace_id|traceID)": "(\w+)"`

cloudcafetech:
@fredr 

Thanks, Its working ....

Able to setup using GUI but ..
Facing little challenges when doing configmap

```
datasources.yaml: |
    apiVersion: 1
    datasources:  
    - access: browser
      editable: false
      isDefault: false
      name: Jaeger
      type: jaeger
      uid: jaeger
      url: http://simplest-query.tracing:16686
    - access: proxy
      editable: false
      isDefault: false
      jsonData:
          maxLines: 1000
          derivedFields:
            - datasourceUid: jaeger
              matcherRegex: "(?:traceID|trace_id)=(\\w+)"
              name: TraceID
              url: "$${__value.raw}"
      name: Loki
      type: loki
      url: http://loki.logging:3100
```

Can you please help me ? how it should be in ```matcherRegex ```

fredr:
> Just quick Question does it support in Dashboard ?

The toggle parsed fields functionality has only been added for the Explore view. For dashboards it would need some more discussion with the grafana team of how/if that should work. Maybe via a transform. I would suggest starting a conversation about it on slack, or create a separate issue about it on github

cloudcafetech:
Ok, Anyway its a Great work !!

💯

PR #26179, opened 2020-07-08
Flux: use monaco query editor

ryantxu:
Early feedback from people using the beta is that writing flux and using templates is hard -- not too surprising.

This PR does two things:
* uses the monaco editor.  -- kind of a mixed bag because we can not (yet!) easily use flux language specification in our build, but it is better than just a text box
* adds suggestions for the supported macros and any template variables in context

 `ctrl+space` and `$` will open suggestions :tada: 
image

#### Open questions:

- [x] fix the reload issue
- [x] can/should we include this in 7.1?  I vote yes even though it is kinda big and past the beta


dprokop:
> the monaco editor is not shown 🤯 -- I think this is caused by some angular binding issues, but I am stumped on where/how to fix it. That same behavior in other react based components using <CodeEditor load fine when you go back

Similar issue but with code formatting- https://github.com/react-monaco-editor/react-monaco-editor/issues/89

I couldn't get around it any other way than adding 
```
setTimeout(() => {
      editor.layout();
}, 100);
```
in CodeEditor's editorDidMount :/

ryantxu:
thank you @dprokop !  updated with the delayed `layout()` and it seems to work well now.

ivanahuckova (approved):
Works great. It also looks great and is more user friendly. I really like autocomplete and highlighting. 
LGTM! 🚀  

Styling nit picks, maybe personal preference - not relevant to this PR:
1. Add marginTop to `<div className="gf-form-inline">` to give is some nice space. 
2. Remove   `<div className="gf-form gf-form--grow"><div className="gf-form-label gf-form-label--grow"></div></div>` as it looks slightly off with it. 

Something like: 
image

PR #26181, opened 2020-07-08
Docs: Use current spelling for macOS

jsoref:
Most spellings of `macOS` in the code are as such, this fixes the two stray instances (the remaining case is `reducers.test.ts` and that's related to how Chrome reports macOS...

Fixes #


oddlittlebird (approved):
LGTM. Thanks for contributing to the docs!

PR #26188, opened 2020-07-09
Chore: Upgrade go-mssqldb

### LINKED ISSUE #26163: Security: Upgrade indirect Go dependency apache/thrift
From [CVE-2019-0205](https://nvd.nist.gov/vuln/detail/CVE-2019-0205):
> In Apache Thrift all versions up to and including 0.12.0, a server or client may run into an endless loop when feed with specific input data. Because the issue had already been partially fixed in version 0.11.0, depending on the installed version it affects only certain language bindings.

Not sure how or if this actually affects Grafana, but we should investigate/try upgrading go dependencies that are dependent on `github.com/apache/thrift v0.12.0`..

aknuds1:
Upgrade go-mssqldb in order to get rid of indirect dependency on apache/thrift v0.12.0.

Fixes #26163.

Unfortunately, our xorm dependency (which is non-current, due to breaking changes in newer versions) pulls in an older version of go-mssqldb, so I had to use a `replace` directive in go.mod to get the newest version. The only alternative I can see is to fork xorm - maybe this is something we want to do in any case, since it breaks compatibility in new versions?


go.mod
```diff
@@ -89,3 +91,6 @@ require (
 	xorm.io/core v0.7.3
 	xorm.io/xorm v0.8.1
 )
+
+// Override xorm's outdated go-mssqldb dependency
```
kylebrandt, line 95: Is there an issue in xorm for this?  If not can we make one and then leave the issue as a comment here (or reference existing one).

Then there is a link to see when we might be able to remove this.

aknuds1: @kylebrandt The problem is that we are on an old version on xorm, and if we upgrade to their latest it will most likely break for us.

kylebrandt: Ah, can you add that to the comment then? append 'since we are on an old version of xorm and not ready to upgrade' - and that it is about getting a newer version of thrift?

Basically the information future people will need to know if they ask themselves "Can I remove this?"

aknuds1: Yeah thanks for the feedback @kylebrandt, I can do that. Just have to deal with more urgent stuff.

marefr: Could we maybe move replace directive above require directive? Makes it easier to spot/not miss it has a replace directive.

marefr (approved):
LGTM, ran the mssql integration tests

PR #26189, opened 2020-07-09
Chore: Enable PR testing in Drone

aknuds1:
Enable PR testing in Drone, along with required fixes. Especially e2e and database integration tests needed tweaking to support running under Drone. CircleCI is untouched, so that will work as before.

The Drone configuration is written in .drone.star (Drone has built-in support for the Starlark language). However, since Starlark processing is currently disabled in our Drone server (due to conflicting with the Jsonnet extension), we have to convert to .drone.yml locally. So basically ignore .drone.yml as it's auto-generated.

The Drone configuration consists of two pipelines for testing PRs: OSS and enterprise. Had to break it into two pipelines since Drone will only distribute pipelines across machines, and not steps within them.

The command to convert .drone.star into .drone.yml is 

```
drone starlark convert
```

Drone builds of my Grafana fork are [here](https://drone.grafana.net/aknuds1/grafana).


marefr:
Are you planning to enable drone as check here so that we can verify it works before merging? Regarding starlark, seems like you have both a starlark file and a drone.yaml file - what's the plan for migrating to jsonnet?

aknuds1:
@marefr the idea is to try to use Drone instead of Circle for PR testing, but making the switch would have to be discussed with leadership first. 

Regarding the .drone.yml, I explain in the PR description that the .drone.yml file has to be auto-generated and committed since our Drone server had to disable the Starlark extension.

It hasn't been decided at all if we should use Jsonnet or not for the templating here, but at least while prototyping I way prefer Starlark since it's much easier to write and read.

marefr:
>  the idea is to try to use Drone instead of Circle for PR testing, but making the switch would have to be discussed with leadership first. As a first step we should get builds running on Drone, quite simply. 

Okay. But can't we run them in parallel for a while where CircleCI continues to be required?  Less risk of accepting and merging this then. Can you at least link to a drone build using this branch?

> Regarding the .drone.yml, I explain in the PR description that the .drone.yml file has to be auto-generated and committed since our Drone server had to disable the Starlark extension.

Sorry missed that.

> It hasn't been decided at all if we should use Jsonnet or not for the templating here, but at least while prototyping I way prefer Starlark since it's much easier to write and read.

Got it. But I got the impression jsonnet have been decided from infra perspective since they disabled starlark and enabled jsonnet.  Probably personal taste here, but I find it more disrupting learning a new language and harder to read this. Maybe because I haven't used python? Not important for now, but when company have agreed on starlark or jsonnet we should use that. One thing I'm currently lacking is how you actually generating the drone.yml file.

aknuds1:
@marefr Yes, we can run them in parallel. This PR isn't doing anything to remove Circle.

To convert the file: `drone starlark convert`.

ryantxu (approved):

package.json
```diff
@@ -13,7 +13,7 @@
     "e2e:debug": "./e2e/start-and-run-suite debug",
     "e2e:dev": "./e2e/start-and-run-suite dev",
     "jest": "jest --notify --watch",
-    "jest-ci": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit --maxWorkers 2",
+    "jest-ci": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit -w ${TEST_MAX_WORKERS:-100%}",
```
dprokop, line 16: `"jest-ci": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit -w ${TEST_MAX_WORKERS:-100%}",`

What does the -100% mean?

aknuds1: @dprokop The bash syntax with `${SOMETHING:-<default>}` means to use the default value if the environment variable isn't set. So it defaults to "100%" for max workers.

packages/grafana-e2e/src/flows/login.ts
```diff
@@ -13,11 +13,10 @@ export const login = (username: string = 'admin', password: string = 'admin') =>
   e2e()
     .url()
     .then(url => {
-      if (/^https?:\/\/localhost/.test(url)) {
-        e2e.pages.Login.skip()
-          .should('be.visible')
-          .click();
-      }
+      // TODO: Detect if we're on the page for setting a new password
```
dprokop, line 16: I assume we need to update the condition to match env's HOST variable? guess we can do this with `e2e.env` method

aknuds1: @dprokop Might it work instead to test if we're on the page for setting a new password? It would be more direct than assuming that we are on this page, if on a certain host.

package.json
dprokop: kk, gotcha. Thanks!

packages/grafana-e2e/src/flows/login.ts
dprokop: Yes, you are right. IMO the assertion below:
```
e2e.pages.Login.skip()
        .should('be.visible')
```

already covers the check if we are on the skip password page. Not sure what was the motivation for the host check.

aknuds1: @dprokop Thanks, great that it's actually this simple. Then I can just drop the TODO comment :)

dprokop (approved):
LGTM

marefr (approved):
LGTM

PR #26191, opened 2020-07-09
Templating: Fix recursive loop of template variable queries when changing ad-hoc-variable 

### LINKED ISSUE #26101: When viewing template variables, Grafana can easily get caught in a rapid, endless loop of trying to refresh every panel on the dashboard

torkelo:
Fixes #26101

### DIFF
--- a/packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
+++ b/packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
@@ -31,7 +31,6 @@ export function SegmentAsync<T>({
         onClick={async () => {
           setSelectPlaceholder('Loading options...');
           const opts = await loadOptions();
-          console.log('loading', opts);
           setLoadedOptions(opts);
           setSelectPlaceholder(opts.length ? '' : 'No options found');
         }}
--- a/packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
+++ b/packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
@@ -29,8 +29,10 @@ export function SegmentAsync<T>({
     return (
       <Label
         onClick={async () => {
+          console.log('loading');
           setSelectPlaceholder('Loading options...');
           const opts = await loadOptions();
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
```diff
@@ -29,8 +29,10 @@ export function SegmentAsync<T>({
     return (
       <Label
         onClick={async () => {
+          console.log('loading');
```
dprokop, line 32:
```suggestion
```

packages/grafana-ui/src/components/Segment/SegmentAsync.tsx
```diff
@@ -31,6 +31,7 @@ export function SegmentAsync<T>({
         onClick={async () => {
           setSelectPlaceholder('Loading options...');
           const opts = await loadOptions();
+          console.log('loading', opts);
```
dprokop, line 34:
```suggestion
```

PR #26200, opened 2020-07-09
ThemeContext: Fix useStyles memoization

s-h-a-d-o-w:
Upon working more with this hook, I realized that memoization simply doesn't work currently. This is because on every call to `stylesFactory()`, the memoized function is recreated. We obviously need to have it stick around and this PR achieves that.

Also includes a docs update to inform the user that the passed function instance needs to remain the same (or - you know, use `useCallback` with appropriate dependencies if necessary) if they want to make use of memoization.

I should also note that I of course tried keeping the memoized functions around locally by using `useState` instead of a map but that would cause an infinite loop if the user passes a function that changes on every render (function triggers setState, setState triggers an update, a new function instance is created, triggers setState, etc.).

### DIFF
--- a/packages/grafana-ui/src/themes/ThemeContext.test.tsx
+++ b/packages/grafana-ui/src/themes/ThemeContext.test.tsx
@@ -1,23 +1,41 @@
-import React from 'react';
 import { config } from '@grafana/runtime';
+import { renderHook } from '@testing-library/react-hooks';
 import { css } from 'emotion';
 import { mount } from 'enzyme';
+import React from 'react';
 import { useStyles } from './ThemeContext';
 
 describe('useStyles', () => {
-  it('passes in theme and returns style object', () => {
+  it('memoizes the passed in function', () => {
+    const stylesCreator = () => ({});
+
+    const { rerender, result } = renderHook(() => useStyles(stylesCreator));
+    const storedReference = result.current;
+    rerender();
+    expect(storedReference).toBe(result.current);
+  });
+
+  it('does not memoize if the passed in function changes every time', () => {
+    const { rerender, result } = renderHook(() => useStyles(() => ({})));
+    const storedReference = result.current;
+    rerender();
+    expect(storedReference).not.toBe(result.current);
+  });
+
+  it('passes in theme and returns style object', done => {
     const Dummy: React.FC = function() {
       const styles = useStyles(theme => {
         expect(theme).toEqual(config.theme);
 
         return {
           someStyle: css`
-            color: ${theme?.palette.critical};
+            color: ${theme.palette.critical};
           `,
         };
       });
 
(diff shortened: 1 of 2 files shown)


packages/grafana-ui/src/themes/ThemeContext.test.tsx
```diff
@@ -1,23 +1,52 @@
-import React from 'react';
 import { config } from '@grafana/runtime';
+import { renderHook } from '@testing-library/react-hooks';
 import { css } from 'emotion';
 import { mount } from 'enzyme';
-import { useStyles } from './ThemeContext';
+import React from 'react';
```
peterholmberg, line 5: Maybe this is just my ocd, but why are you moving this import?

peterholmberg (commented):
Just a question re imports.

Otherwise it seems to make sense, guess this something we need to iterate over.

packages/grafana-ui/src/themes/ThemeContext.test.tsx
s-h-a-d-o-w: I rarely manage imports manually these days, must've been vscode. :)
But you're absolutely right - it's ugly and I fixed it.

s-h-a-d-o-w:
@peterholmberg What is missing here to get approval?

torkelo:
added @dprokop to review list as he wrote the original styleFactory :)

packages/grafana-ui/src/themes/ThemeContext.tsx
```diff
@@ -39,9 +41,23 @@ export function useTheme(): GrafanaTheme {
...
-  return stylesFactory(getStyles)(useTheme());
+  const theme = useTheme();
+
+  let memoizedStyleCreator = memoizedStyleCreators.get(getStyles);
+  if (!memoizedStyleCreator) {
+    memoizedStyleCreator = stylesFactory(getStyles);
+    memoizedStyleCreators.set(getStyles, memoizedStyleCreator);
+  }
+
+  return memoizedStyleCreator(theme);
```
dprokop, line 60: This looks good @s-h-a-d-o-w . WDYT about removing the memorized styles getter from the `memoizedStyleCreators` Map, when the component unmounts?

dprokop (approved):
Goodie, thanks @s-h-a-d-o-w !

packages/grafana-ui/src/themes/ThemeContext.tsx
s-h-a-d-o-w: I think that's a great catch and something I should really work out because it would constitute a memory leak otherwise. ;)

s-h-a-d-o-w: Done.

PR #26203, opened 2020-07-09
Queries: Move explore state out of query targets

### LINKED ISSUE #26202: CloudWatch: Various internal variables leaked with query requests
**What happened**:
Certain variables that should be internal to grafana/explore such as `liveStreaming` and `showingGraph` are sent along as part of the query request.
image

For only the data relevant to the query request to be sent.

- Navigate to Explore
- Select a CloudWatch Datasource
- Make a CloudWatch metrics query
- Investigate the query request in your browser's devtools

The unwanted variables in question are attached to the queries as part of `buildQueryTransaction` (in public/app/core/utils/explore.ts). This isn't a problem in and of itself but when the actual query is sent we should use lodash pick or similar to send only the relevant parameters.

kaydelaney:
Before being provided to a data source's query method, query targets are populated with various pieces of explore state
in `buildQueryTransaction`. Since this data isn't query-specific, it makes more sense to move it out of the targets and up a level as part of the `DataQueryRequest` object

Closes #26202

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/datasource.ts
+++ b/public/app/plugins/datasource/cloudwatch/datasource.ts
@@ -71,6 +71,25 @@ const displayCustomError = (title: string, message: string) =>
 export const MAX_ATTEMPTS = 8;
 const POLLING_TIMES = [100, 200, 500, 1000];
 
+const metricsParams = [
+  'queryMode',
+  'id',
+  'region',
+  'namespace',
+  'expression',
+  'metricName',
+  'dimensions',
+  'statistics',
+  'period',
+  'alias',
+  'matchExact',
+  'datasource',
+  'queryType',
+  'key',
+  'hide',
+  'refId',
+];
+
 export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWatchJsonData> {
   type: any;
   proxyUrl: any;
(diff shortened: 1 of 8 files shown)


public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -199,12 +218,14 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
             throw { message: 'Invalid extended statistics' };
           }
 
+          const reducedItem = _.pick(item, metricsParams);
```
papagian, line 221: I have a small concern how we can be sure that the above list is complete therefore I would prefer to filter out the internal to explorer variables instead if this is feasible.

public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -44,7 +44,10 @@ import LanguageProvider from './language_provider';
 import { serializeParams } from '../../../core/utils/fetch';
 import { RowContextOptions } from '@grafana/ui/src/components/Logs/LogRowContextProvider';
 
-export type RangeQueryOptions = Pick<DataQueryRequest<LokiQuery>, 'range' | 'intervalMs' | 'maxDataPoints' | 'reverse'>;
+export type RangeQueryOptions = Pick<
+  DataQueryRequest<LokiQuery>,
+  'range' | 'intervalMs' | 'maxDataPoints' | 'reverse' | 'liveStreaming' | 'showingGraph' | 'showingTable'
```
torkelo, line 49: This RangeQueryOptions type feels very mess, is there an easier way to define it?

kaydelaney: You're right. I'll take a look and see if I can improve it

aocenas (approved):

PR #26204, opened 2020-07-09
Chore: Fix all Typescript strict null errors 

torkelo:


public/app/core/components/Select/MetricSelect.tsx
```diff
@@ -7,7 +7,7 @@ import { Variable } from 'app/types/templates';
 const { Select } = LegacyForms;
 
 export interface Props {
-  onChange: (value: string) => void;
+  onChange: (value: string | undefined) => void;
```
kaydelaney, line 10: Could this just be
```suggestion
  onChange: (value?: string) => void;
```

torkelo: no, it's not the same. The argument is not optional, but can be undefined

kaydelaney: Cool, makes sense

kaydelaney (approved):

tsconfig.json
```diff
@@ -10,7 +10,8 @@
     },
     "rootDirs": ["public/"],
     "typeRoots": ["node_modules/@types", "public/app/types"],
-    "allowJs": true
+    "allowJs": true,
+    "strictNullChecks": true
```
dprokop, line 14: <3

dprokop (approved):
Plenty to digest but haven't found anything suspicious. Great work @torkelo!

PR #26207, opened 2020-07-09
Backend: use latest go plugin sdk (0.74.0) to sort wide frames

### LINKED ISSUE #22937: Azure Application Insights: feature for sorting the series in groupby
**Why is this needed**:

- Grafana version: 6.7.0
- Datasource: Application Insights

Hi. I will start of with a screenshot showing three metrics in each their own graph, all grouped by `cloud/roleName`. Role name is a set of six different executables in our solution. Pay attention to the order of the series.

image

image

There is no correlation of series ordering and coloring between the graphs/data tables. It is quite hard for the human brain to accurately comprehend the information.

There is no way to alter the behavior in Application Insights itself without resolving to query mode, which would be a tradeoff to larger delays in the delivery of the metrics as well as being more complex to define. Adding a sorting option will resolve this in a nice and clean way.

kylebrandt:

This fetches the latest Go plugin sdk which adds support for sorting the field of a wide time series frame.

This makes it so results are more stable between refreshes of data. data.LongToWide now sorts, so the two Azure Analytics APIs will be updated to sort based on that. This also calls the functions in Insights Metrics will builds a wide frame directly. 

Fixes #22937


marefr (approved):
LGTM

ryantxu (approved):

papagian (commented):
[`TestFrameToSeriesSlice`](https://github.com/grafana/grafana/blob/3651a8e976acc7bfc57d481660d27e204a384356/pkg/tsdb/frame_util_test.go) needs to be updated too

kylebrandt:
Putting this back to WIP.

When running through some tests I wasn't fond of the sort order. There are cases where data is hierarchical, consider Country,City (A Country has cities, a city doesn't have countries (I think, probably some fun geography fact that proves me wrong)).  However, alphabetically, City will sort first.

So want to redo the sort such that one can specify the label keys that have sort order preference. With things like Kusto/SQL, this would be implied by the order of columns in the initial response. Assuming one would naturally select Country,City, or if not, can change the order. If the response does not provide it, it is possible a query editor can knowing the order based on the request - so multiple options for different situations.

So will take another stab at the sorting tomorrow. That being said, if it doesn't go well can go forward with this, and change the sort later. This is because being sorted is more of an issue that the sort order itself - but it would be nice to get it correct.

PR #26214, opened 2020-07-09
Failed to load application files message improvement

### LINKED ISSUE #26128: Improve "failed to load" message when serving at subpath
If grafana fails with subpath, it would be nice to show a hint about `serve_from_sub_path`.

**What happened**:

When serving grafana at subpath (from kube-prometheus with ingress), the following message is shown in the browser. There are no 404s or anything like that:

```
If you're seeing this Grafana has failed to load its application files

[...]

2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath

[...]
```

A message like the following would save a lot of time in fixing the issue:

```
If you're seeing this Grafana has failed to load its application files

[...]

2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath,
and that serve_from_sub_path is set unless there is a rewriting proxy in front of grafana.

[...]
```

papagian:

Fixes #26128


dprokop:
@papagian is this going to be included in 7.1?

papagian:
> @papagian is this going to be included in 7.1?

not sure, I've assigned it to 7.2 since it's just an improvement.

public/views/index-template.html
```diff
@@ -199,9 +199,10 @@
         </p>
         <p>
           1. This could be caused by your reverse proxy settings.<br /><br />
-          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath<br />
+          2. Otherwise, if you host grafana under subpath make sure your grafana.ini root_url setting includes subpath
```
marefr, line 202: Otherwise to what? Using a reverse proxy?

public/views/index-template.html
```diff
@@ -199,9 +199,10 @@
...
+          and that serve_from_sub_path is set.<br />
```
marefr, line 203: serve_from_sub_path shouldn't be needed when using a reverse proxy, but could. You use serve_from_sub_path if you want to host under a sub path but don't use a reverse proxy.

marefr (changes requested):
Some minor comments/concerns

public/views/index-template.html
papagian: that's I have used `otherwise`. Should I change it to: `I you don't use a reverse proxy...`?

marefr: I'm confused by the whole issue request. If you're not using a reverse proxy why would you end up with this problem?

papagian: I guess the [issue](https://github.com/grafana/grafana/issues/26128) is for adding an indication for making sure that `serve_from_sub_path` is set in addition to setting `root_url` properly  if not using a reverse proxy, no?

public/views/index-template.html
marefr: @papagian what about this?

```suggestion
          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath. If not using a reverse proxy make sure to set serve_from_sub_path to true.
```

public/views/index-template.html
papagian, line 203:
```suggestion
```

public/views/index-template.html
```diff
@@ -199,7 +199,7 @@
         </p>
         <p>
           1. This could be caused by your reverse proxy settings.<br /><br />
-          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath<br />
+          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath. If not using a reverse proxy make sure to set serve_from_sub_path to true.
```
aknuds1, line 202: Is there a reason why the `<br />` at the end of the line is removed? Seems that other list items end with to `<br />`.

papagian: good catch; it was by mistake

aknuds1 (approved):
LGTM

hugohaggmark (approved):
LGTM!

papagian:
@marefr Can you have a fresh look next week? thanks!

marefr (approved):
LGTM

PR #26216, opened 2020-07-09
[WIP] App Plugin: Allow plugin header to show/hide based on content

besartberisha:
This PR includes two simple changes needed for App Plugin development:

-  Show header only if the data is passed to it;
-  Allow the app plugin to remove parent styling wrapper to be able for more customization such as fullscreen layout;

### DIFF
--- a/packages/grafana-data/src/types/app.ts
+++ b/packages/grafana-data/src/types/app.ts
@@ -17,7 +17,12 @@ export interface AppRootProps<T = KeyValue> {
   /**
    * Pass the nav model to the container... is there a better way?
    */
-  onNavChanged: (nav: NavModel) => void;
+  onNavChanged: (nav: NavModel | null) => void;
+
+  /**
+   * Pass a boolean to either show or hide the default styling that the parent App Wrapper has.
+   */
+  removeParentStyling?: (removeParentStyling: boolean) => void;
 }
 
 export interface AppPluginMeta<T = KeyValue> extends PluginMeta<T> {
--- a/public/app/core/components/Page/Page.tsx
+++ b/public/app/core/components/Page/Page.tsx
@@ -14,6 +14,7 @@ import { Branding } from '../Branding/Branding';
 interface Props {
   children: React.ReactNode;
   navModel: NavModel;
+  isLoading?: boolean;
 }
 
 class Page extends Component<Props> {
@@ -44,12 +45,12 @@ class Page extends Component<Props> {
   }
 
   render() {
-    const { navModel } = this.props;
+    const { navModel, isLoading } = this.props;
     return (
       <div className="page-scrollbar-wrapper">
         <CustomScrollbar autoHeightMin={'100%'} className="custom-scrollbar--page">
           <div className="page-scrollbar-content">
-            <PageHeader model={navModel} />
+            {!isLoading && navModel && <PageHeader model={navModel} />}
             {this.props.children}
             <Footer />
           </div>
--- a/public/app/core/components/Page/PageContents.tsx
+++ b/public/app/core/components/Page/PageContents.tsx
@@ -6,14 +6,19 @@ import PageLoader from '../PageLoader/PageLoader';
 
 interface Props {
   isLoading?: boolean;
+  removeStyling?: boolean;
   children: React.ReactNode;
 }
 
 class PageContents extends Component<Props> {
   render() {
-    const { isLoading } = this.props;
+    const { isLoading, removeStyling } = this.props;
 
-    return <div className="page-container page-body">{isLoading ? <PageLoader /> : this.props.children}</div>;
+    return (
+      <div className={`${!removeStyling ? 'page-container page-body' : ''}`}>
+        {isLoading ? <PageLoader /> : this.props.children}
+      </div>
+    );
   }
 }
 
--- a/public/app/features/plugins/AppRootPage.tsx
+++ b/public/app/features/plugins/AppRootPage.tsx
@@ -24,6 +24,7 @@ interface State {
   loading: boolean;
   plugin?: AppPlugin | null;
   nav: NavModel;
+  removeStyling?: boolean;
 }
 
 export function getAppPluginPageError(meta: AppPluginMeta) {
@@ -45,6 +46,7 @@ class AppRootPage extends Component<Props, State> {
     this.state = {
       loading: true,
       nav: getLoadingNav(),
+      removeStyling: false,
     };
   }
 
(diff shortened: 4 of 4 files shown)


packages/grafana-data/src/types/app.ts
```diff
@@ -17,7 +17,12 @@ export interface AppRootProps<T = KeyValue> {
   /**
    * Pass the nav model to the container... is there a better way?
    */
-  onNavChanged: (nav: NavModel) => void;
+  onNavChanged: (nav: NavModel | null) => void;
+
+  /**
+   * Pass a boolean to either show or hide the default styling that the parent App Wrapper has.
+   */
+  removeParentStyling?: (removeParentStyling: boolean) => void;
```
ryantxu, line 25: rather than a boolean flag..  maybe we could have a enum?  To start with, it would just be two states... with and without header?

torkelo:
I do am not sure I can allow this, Grafana needs to own the styling so we can update look and feel without breaking all apps. If the app needs a new style option / mode, (full width etc?) we should have an option for that instead of removing the page classes. 

What page style is needed here?

packages/grafana-data/src/types/app.ts
torkelo: Should this not be on the page level instead of the whole app plugin? And yes an enum like PageStyle { Centered | Full width }

public/app/core/components/Page/Page.tsx
```diff
@@ -44,12 +45,12 @@ class Page extends Component<Props> {
...
 
   render() {
-    const { navModel } = this.props;
+    const { navModel, isLoading } = this.props;
     return (
       <div className="page-scrollbar-wrapper">
         <CustomScrollbar autoHeightMin={'100%'} className="custom-scrollbar--page">
           <div className="page-scrollbar-content">
-            <PageHeader model={navModel} />
+            {!isLoading && navModel && <PageHeader model={navModel} />}
```
torkelo, line 53: navModel is not nullable

public/app/core/components/Page/PageContents.tsx
```diff
@@ -6,14 +6,19 @@ import PageLoader from '../PageLoader/PageLoader';
 
 interface Props {
   isLoading?: boolean;
+  removeStyling?: boolean;
```
torkelo, line 9: is is named like a verb/action, not a boolean :) 

think better option here is pageStyle?: PageStyle  (and have a Centered + FullWidth)

public/app/core/components/Page/Page.tsx
torkelo: there is loading nav shown during loading state

public/app/features/plugins/AppRootPage.tsx
```diff
@@ -71,20 +73,30 @@ class AppRootPage extends Component<Props, State> {
...
+      <Page navModel={nav} isLoading={loading}>
+        <Page.Contents isLoading={loading} removeStyling={removeStyling}>
           {plugin && plugin.root && (
-            <plugin.root meta={plugin.meta} query={query} path={path} onNavChanged={this.onNavChanged} />
+            <plugin.root
+              meta={plugin.meta}
+              query={query}
+              path={path}
+              onNavChanged={this.onNavChanged}
+              removeParentStyling={this.removeStyling}
```
torkelo, line 98: not sure why this callback exists, seems like a pageStyle property on the app is enough 

Or does the app want to change page style on dynamically? For that I think it's better to either expose seperate page components, or expose Page & PageHeader & NavModel to plugins so they can compose pages themselves. 

thoughts @ryantxu ?

torkelo (commented):
Sorry for lots of comments, Apps needs a lot of refinements in the API & capabilities. 

In the future apps will probably get their own react router, and then use Page & PageHeader components directly. 

so things like onNavChanged should not be needed. 

In the meantime not sure what is best to do.

packages/grafana-data/src/types/app.ts
ryantxu: > export interface AppRootProps<T = KeyValue> {

This is the props for the root (and only) page that an app has control of.  It is `/a/${pluginId}/*`

public/app/features/plugins/AppRootPage.tsx
besartberisha: Hi @torkelo thank you for the feedback I really appreciate it, since it's all tied together I thought I will use this thread to try and explain my changes.

So the app plugin we are working on easystart requires full width no header layout, but also has other pages that have header with same layout as other plugins. 

Now what I did is: 
First I used the `removeParentStyling` to pass a boolean, whenever we wanted to remove the classes of the app which are responsible for centering the content, so by default that would always be as it was before, but only when we need the fullscreen. 

And second for the header I used the `onNavChanged` to check when no data is there to not show the header, in this case the other plugins where data is passed to header it wouldn't affect them. 

Here is a screenshot of how the wizard will look like and maybe we could add two other conditional props, one for showing or not showing the header and the other one as you suggested with enums for styling?

wizard-intro

ryantxu: as implemented, the only way to tell the container about your settings (nav) is with callbacks.  

I hope we can revist this when we are using react router nativly

besartberisha: Does it make sense to have two callbacks? e.g.

1.) `headerVisibility` which would either be `visible` or `hidden` 
2.) `pageStyles` which would have `fullWidth` `centered` 

For me to achieve the designs of the app plugin, I think these two could be enough since I can then control when I need to hide the header and when I need to have a fullscreen, but also gives me the flexibility to have the `normal` app plugin layout which we will probably need to manage the plugin.

besartberisha:
Will close this PR in favor of the #26247 which is cleaner and simple solution for our custom app plugin.

public/app/features/plugins/AppRootPage.tsx
dprokop: @torkelo @ryantxu @besartberisha instead of altering the Grafana DOM, what if Grafana had a Portal that app plugins can render to? Similar idea as we do with modals: https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Modal/ModalsContext.tsx

torkelo: not sure that helps any, think exposing a Router component, and Page & Page header components makes the most long term sense:
https://github.com/grafana/grafana/pull/26245

PR #26218, opened 2020-07-09
Docs: Add chained variables

oddlittlebird:
This topic adds in-depth information about chained variables. I attempted to make it accessible to users of all levels.

**Special notes for your reviewer**: @scoren-gl, please read through this from the perspective of an inexperienced user. I am trying to make this accessible to all levels.

All reviewers, please pay special attention to my queries. My multi-value queries probably need correcting.

Feel free to expand or add information anywhere that you think would be useful.


docs/sources/variables/chained-variables.md
```diff
@@ -0,0 +1,174 @@
++++
+title = "Chained variables"
+keywords = ["grafana", "templating", "variable", "nested", "chained", "linked"]
+type = "docs"
+[menu.docs]
+weight = 600
```
marefr, line 6: Shouldn't be needed any longer since we have the meny.yaml
```suggestion
```

docs/sources/variables/chained-variables.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### server variable
+
+The query for this variable basically says, "Give me all servers for the currently chosen application." 
+
+```
+apps.$app.*
+```
+
+By itself, this query is nonsense, because `$app` has no value. However, if the user selects `backend`, then the query changes to:
```
marefr, line 56: I wouldn't call it nonsense because the query will be `apps.{backend,country,fakesite}.*` and can also be useful depending on use case. For example, if you only want to filter on server, but not app given that you have same server value for multiple apps.

docs/sources/variables/chained-variables.md
```diff
@@ -0,0 +1,174 @@
...
+apps.$app.$server.cpu.*
+```
+
+This query basically says, "Show me the CPU metrics for the selected server."
+
+Depending on what variable options the user selects, you could get queries like:
+
+```
+apps.backend.backend_01.cpu.*
+apps.backend.backend_02+backend_03.cpu.*
```
marefr, line 88: If I'm not mistaken:

```suggestion
apps.{backend.backend_02,backend_03}.cpu.*
```

docs/sources/variables/chained-variables.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### host variable
+
+The query for this variable basically says, "Give me all hosts for the currently chosen data center." 
+
+```
+SHOW TAG VALUES WITH KEY = "hostname" WHERE "datacenter" =~ /^$datacenter$/
+```
+
+By itself, this query is nonsense, because `$datacenter$` has no value. However, if the user selects `America`, then the query changes to:
```
marefr, line 116: Don't really understand the nonsense part. Why doesn't `$datacenter` has a value. Since multi value is not selected the first value returned from data center variable will be selected.

```suggestion
By itself, this query is nonsense, because `$datacenter` has no value. However, if the user selects `America`, then the query changes to:
```

docs/sources/variables/chained-variables.md
```diff
@@ -0,0 +1,174 @@
...
+
+- Chaining variables create parent/child dependencies. You can envision them as a ladder or a tree.
+- The easiest way to create a new chained variable is to copy the variable that you want to base the new one on. In the variable list, click the **Duplicate variable** icon to the right of the variable entry to create a copy. You can then add on to the query for the parent variable.
+- New variables created this way appear at the bottom of the list. You might need to drag it to a different position in the list to get it into a logical order.
+
+### Variable order
+
+You can change the orders of variables in the dashboard variable list by clicking the up and down arrows on the right side of each entry. Grafana lists variable dropdowns left to right according to this list, with the variable at the top on the far left.
+
+- List variables that do not have dependencies at the top.
```
marefr, line 166: Personally this depends, if I have a variable that I'm not changing that often like interval, I would put that at the end because I would like to have the drop downs I use the most starting at the left in dashboard.

marefr (changes requested):
Some minor things, see comments. Are you planning to link to this page from the variables overview/index page?

docs/sources/variables/chained-variables.md
oddlittlebird: We want to get rid of menu.yaml and go back to the standard Hugo method of creating the side menu, which is based on folder structure and weight.

oddlittlebird:
> Some minor things, see comments. Are you planning to link to this page from the variables overview/index page?

From the overview, yes. Now that the other variables PR is merged, that is. :)

docs/sources/variables/chained-variables.md
oddlittlebird: I based this on recommendations from @torkelo . This list of best practices is specifically for chained variables, but this would be a great suggestion for a more general list of variable best practices. Got any others? :)

docs/sources/variables/chained-variables.md
marefr: Ohh great 👍

marefr (approved):
LGTM

PR #26224, opened 2020-07-10
Chore: Enable Go linter gocritic

aknuds1:
As part of ongoing effort to harmonize Go linting with grafana-plugin-sdk-go, enable also the gocritic linter and fix issues it points out.

The linter mostly finds needlessly complicated code, and in my opinion it does make the code better/more homogeneous. Interested in feedback however, with regards to other people's POV.

We've decided to delay the merging until v7.1.0, just in case it trips people's PRs up.


pkg/services/alerting/rule.go
```diff
@@ -89,7 +89,7 @@ func getTimeDurationStringToSeconds(str string) (int64, error) {
 
 	matches := valueFormatRegex.FindAllString(str, 1)
 
-	if len(matches) <= 0 {
+	if len(matches) == 0 {
```
papagian, line 92: From the [FindAllString()](https://golang.org/pkg/regexp/#Regexp.FindAllString) docs: `A return value of nil indicates no match.`
so maybe this should check if the `matches` is `nil` instead.

papagian (approved):
LGTM! I have a small suggestion.

pkg/services/alerting/rule.go
aknuds1: @papagian I think it's actually considered OK to check the length of a nil slice in Go, see [this example](https://tour.golang.org/moretypes/12#:~:text=The%20zero%20value%20of%20a,and%20has%20no%20underlying%20array.&text=fmt.,-Println(%22nil!). 

We can check against `nil` instead, but I don't think there's any requirement calling for it :)

papagian: yes, let's keep it as is.

pkg/api/dtos/models.go
```diff
@@ -73,12 +73,7 @@ func GetGravatarUrlWithDefault(text string, defaultText string) string {
 		return GetGravatarUrl(text)
 	}
 
-	reg, err := regexp.Compile("[^a-zA-Z0-9]+")
-
-	if err != nil {
-		return ""
-	}
-
+	reg := regexp.MustCompile("[^a-zA-Z0-9]+")
```
kylebrandt, line 76: Put into package init, that way a panic it is caught at package init time. 

and also only need to compile once.

pkg/plugins/dashboard_importer.go
```diff
@@ -125,7 +125,7 @@ func (this *DashTemplateEvaluator) findInput(varName string, varType string) *Im
 func (this *DashTemplateEvaluator) Eval() (*simplejson.Json, error) {
 	this.result = simplejson.New()
 	this.variables = make(map[string]string)
-	this.varRegex, _ = regexp.Compile(`(\$\{.+\})`)
+	this.varRegex = regexp.MustCompile(`(\$\{.+\})`)
```
kylebrandt, line 128: same here with the regex

pkg/services/alerting/notifiers/telegram.go
```diff
@@ -125,16 +125,16 @@ func (tn *TelegramNotifier) buildMessageLinkedImage(evalContext *alerting.EvalCo
 
 	ruleURL, err := evalContext.GetRuleURL()
 	if err == nil {
-		message = message + fmt.Sprintf("URL: %s\n", ruleURL)
+		message += fmt.Sprintf("URL: %s\n", ruleURL)
```
kylebrandt, line 128: string builder would be better but out of scope

kylebrandt (approved):

kylebrandt:
Approved, but would be good to fix those regexes before merging to avoid the possible panic.

aknuds1:
Thanks @kylebrandt, will take care of it.

marefr (approved):
LGTM

jessabe (commented):
LGTM but as discussed in our weekly yesterday it might be best to hold off merging this until after 7.1

ryantxu (approved):

PR #26239, opened 2020-07-10
SignIn button - use correct url

### LINKED ISSUE #26228: Sign In / Login in  ignoring root_url option in v7.1.0-beta2 and 7.0.6
**What happened**:
Signing in from any area of Grafana does not work if grafana is running from a root URL containing a longer path. Sign Out and all other links within Grafana seem to work fine.

Sign In button should contain the link including the root_url option from the grafana.cfg file.

Add a root URL to grafana.ini similar to this:
root_url = https://server.onmydomain.com/grafana

Go to sign into Grafana. Hovering over the Sign In link shows:
https://server.onmydomain.com/d/nsFkdjs3f/dashboard?orgid=1

**Environment**:
- Grafana version: 7.1.0-beta2
- User OS & Browser:Win10/Chrome83
- Others: Nginx as a reverse proxy

dprokop:
Fixes https://github.com/grafana/grafana/issues/26228

Providing base url with subpath to [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) removes the subpath. We need to provide it as part of the URL in order to generate correct URL.


papagian (commented):
Thanks for fixing it. I hadn't released that for the subpath by reading the `URL()` docs. However, in retrospect seems quite obvious.
I have tested w/o subpath and with reverse proxy and it works as expected.
I have opened a [PR](https://github.com/grafana/grafana/pull/26266) for fixing/improving the failing test.

public/app/core/components/sidemenu/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { getConfig } from 'app/core/config';
+
+export const getForcedLoginUrl = (url: string) => {
+  const urlObj = new URL(`${getConfig().appSubUrl}${url}`, getConfig().appUrl);
```
torkelo, line 4: why does this link include appUrl? Should only need touse appSubUrl, any reason why this link is absolute? Not sure any other link is absolute

torkelo (commented):
@papagian wondering why https://github.com/grafana/grafana/pull/25567  made login link absolute (ie include app url) 

related 
https://github.com/grafana/grafana/issues/26268

papagian:
> @papagian wondering why #25567 made login link absolute (ie include app url)
> 
> related
> #26268

The intention was to use [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) following [this](https://github.com/grafana/grafana/pull/25567#discussion_r439932330) suggestion.

papagian (approved):
LGTM, I don't know though if there is a better way to add the `forceLogin` query parameter without changing the link to absolute URL.

torkelo:
Ok, think this should not be an absolute url, should just use appSubUrl , we never use appUrl in frontend (no need), only for server side absolute links & redirects

dprokop:
> Ok, think this should not be an absolute url, should just use appSubUrl , we never use appUrl in frontend (no need), only for server side absolute links & redirects

dprokop:
@papagian updated not to use abs URL. Do you mind taking a look?

public/app/core/components/sidemenu/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { getConfig } from 'app/core/config';
+
+export const getForcedLoginUrl = (url: string) => {
+  return `${getConfig().appSubUrl}${url}&forceLogin=true`;
```
papagian, line 4: This can be a problem if the url does not contain any other query parameters eg. if the `url` is `http://localhost:3000/grafana/dashboards` the signin url wil become: `http://localhost:3000/grafana/dashboards&forceLogin=true`.

An alternative (but ugly approach) for solving this is [here](https://github.com/grafana/grafana/pull/25567/commits/ed6d2216a3a8b3e4dab25f455aa6ebf49fc6005e#diff-4fe11550fd8c741c221f27c3e5db5737R7).

dprokop: you are absolutely right, missed this case :/

dprokop: Fixed it in a simpler way in 4d4937d @papagian

papagian (approved):
LGTM, works as expected.

PR #26247, opened 2020-07-10
AppPlugin: give full control to page layout when navigation is missing

ryantxu:
This is an altenative to https://github.com/grafana/grafana/pull/26245 and https://github.com/grafana/grafana/pull/26216

In 7.2, we should fix this so that we pass a react-router to apps and give them the ability to structure however they need.  However, we need something simpler to get in 7.1

This PR changes the behavior without changing any methods.  

If a navModel is defined, the container will include it, otherwise the page has full control of the layout

### DIFF
--- a/public/app/features/plugins/AppRootPage.tsx
+++ b/public/app/features/plugins/AppRootPage.tsx
@@ -9,9 +9,9 @@ import { AppEvents, AppPlugin, AppPluginMeta, NavModel, PluginType, UrlQueryMap 
 import Page from 'app/core/components/Page/Page';
 import { getPluginSettings } from './PluginSettingsCache';
 import { importAppPlugin } from './plugin_loader';
-import { getLoadingNav } from './PluginPage';
 import { getNotFoundNav, getWarningNav } from 'app/core/nav_model_srv';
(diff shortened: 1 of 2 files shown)


public/app/features/plugins/AppRootPage.tsx
```diff
@@ -80,6 +79,14 @@ class AppRootPage extends Component<Props, State> {
       return <div>No Root App</div>;
     }
 
+    // When no naviagion is set, give full control to the app plugin
+    if (!nav) {
+      if (plugin && plugin.root) {
+        return <plugin.root meta={plugin.meta} query={query} path={path} onNavChanged={this.onNavChanged} />;
+      }
+      return <PageLoader />;
```
ryantxu, line 87: @besartberisha is this loading behavior what you want?

torkelo (approved):

public/app/features/plugins/AppRootPage.tsx
besartberisha: Hey @ryantxu yes exactly, far more cleaner solution rather than exposing different callbacks, thank you.

PR #26248, opened 2020-07-10
InfluxDB: include sample response data for each test case

ryantxu:
This PR includes a golden response for each sample flux response.


papagian (approved):
LGTM!

PR #26257, opened 2020-07-11
Removes erroneous $ while deleting organisation

### LINKED ISSUE #26256: Random "$" while deleting organisation
**What happened**:
image

Try to delete an organisation

**Environment**:
- Grafana version: 7.2.0-pre
- User OS & Browser: Ubuntu 16.04, Chrome

KamalGalrani:
Removes erroneous $ while deleting organisation

Fixes #26256


Clarity-89 (approved):
Good catch! thanks for fixing this 🎉

PR #26277, opened 2020-07-13
Admin: Update org list after delete

Clarity-89:

Currently the list of organisations is not updated after an organisation has been deleted, unless the page is refreshed. This PR makes sure that `fetchOrgs` api call happens after the `deleteOrg` call completed.


dprokop:
Seems OK, but long term I think we should consider altering the local state first and then re-fetching the list.  Sort of optimistic update. With the proposed solution I can see weird behaviour i.e. when the connection is slow:

Kapture 2020-07-13 at 14 05 39

Clarity-89:
> Seems OK, but long term I think we should consider altering the local state first and then re-fetching the list. Sort of optimistic update. With the proposed solution I can see weird behaviour i.e. when the connection is slow:
> 

Yeah, makes sense. But then we'll need to be able to revert the state if the api call has errors. Do we have any functionality like this available?

dprokop:
> Do we have any functionality like this available?

nothing reusable unfortunately

dprokop (approved):

PR #26282, opened 2020-07-13
Packaging: Support arbitrary plugin URLs in custom docker builds

robx:
This modifies the custom Dockerfiles to support

GF_INSTALL_PLUGINS=\<plugin-url\>;\<plugin-name\>

just like the normal docker image entrypoint.


robx:
CI failure seems unrelated.

robx:
This isn't an important change for me, I've since moved to installing plugins in a different way. Still seems like a good change to me to provide more consistency.

marefr:
Thank you, but since you no longer need this and there are no referenced issue I will close this. I find the use case quite advanced and if you need it you could easily build your own docker image extending the existing one. Should someone express a need of this we'll consider reopen this.

PR #26288, opened 2020-07-13
implicit import symbol-observable to avoid inconsistent symbol

### LINKED ISSUE #26264: Need help: rxjs error in datasource plugin
**What happened**:
Got error `TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable` when return observable from datasource plugin.

No error, observable is processed correctly, returned data is displayed in graph panel.

Use observable in third-party plugin.

My plugin is mixed datasource.
The Grafana builtin mixed datasource is work without error.
So I copy builtin datasource code and use it as third-party plugin, but it doesn't work.
https://github.com/mtanda/grafana-target-label-rewrite-datasource/tree/same_as_mixed_datasource

I suspect the issue might be same as https://github.com/ReactiveX/rxjs/issues/5409 .
https://ncjamieson.com/how-to-use-interop-observables/

**Environment**:
- Grafana version: f4b6abb05e39a5a4fcc45c5897eeb313a1a19418
- Data source type & version:
- User OS & Browser: Firefox
- Grafana plugins:
- Others:

mtanda:
Avoid inconsistent observable symbol issue.

Fixes https://github.com/grafana/grafana/issues/26264

This could be blocker to migrate datasource plugin to Grafana 7.x.
Because, grafana-runtime use Observable in datasource base class.
https://github.com/grafana/grafana/blob/v7.0.6/packages/grafana-runtime/src/utils/DataSourceWithBackend.ts#L105

### DIFF
--- a/public/app/app.ts
+++ b/public/app/app.ts
@@ -1,3 +1,4 @@
+import 'symbol-observable';
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 


dprokop (approved):
@mtanda thanks for detailed investigation in #26264!

PR #26292, opened 2020-07-13
Chore: Upgrade build pipeline tool; refactor Drone config

aknuds1:
Chore PR to upgrade to latest build pipeline tool, and also refactor Drone configuration to a shared module, scripts/lib.star. This module is shared with grafana-enterprise. Basically, scripts/lib.star gets vendored in the grafana-enterprise repo so it can be used there also.

Please ignore .drone.yml, since it's auto-generated. Eventually you can read it to verify that it does what it should.


marefr (approved):
LGTM

PR #26293, opened 2020-07-13
AdminUsers: reset page to zero on query change

### LINKED ISSUE #26209: user search works only on page 1
**What happened**:
searching a user (/admin/users) will only work if you are on page 1. Even if you search for an user present on page 2 it will not work unless you are on page 1.

user search should work no matter which page are you in.

just search for an user in other page except 1

**Environment**:
- Grafana version: 7.0.3
- Data source type & version: no need for this case
- OS Grafana is installed on: Windows 2019
- User OS & Browser: Windows 10 / Chrome
- Grafana plugins:
- Others:

hshoff:

- fixes user search on `/admin/users` to work no matter what page you are on (https://github.com/grafana/grafana/issues/26209)

Before if you searched for a user while on page 2 you would see empty results. This happens because page state does not get reset when query changes. The fix is to reset page to `0` whenever the query changes.

before
search-page-before

after
search-page-fixed

Fixes #26209

### DIFF
--- a/public/app/features/admin/state/reducers.test.ts
+++ b/public/app/features/admin/state/reducers.test.ts
@@ -10,8 +10,10 @@ import {
   userMappingInfoLoadedAction,
   userProfileLoadedAction,
   userSessionsLoadedAction,
+  userListAdminReducer,
+  queryChanged,
 } from './reducers';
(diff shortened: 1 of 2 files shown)


zoltanbedi (approved):
Thank you @hshoff for your contribution, this looks good to me.

PR #26302, opened 2020-07-13
Generic OAuth: Enable Hybrid Flow/AcrValues

dcbroad3:

This PR is intended to allow the Generic OAuth login setting to perform the Hybrid OIDC flow, which is a type of OIDC flow that is well described [here](https://auth0.com/docs/api-auth/tutorials/hybrid-flow). At OSIsoft we have implemented this flow with our identity server built on [Identity Server](https://identityserver.io/) to enable login to [OSIsoft Cloud Services](https://cloud.osisoft.com/). We currently have published a [Grafana sample](https://github.com/osisoft/OSI-Samples-OCS/tree/master/advanced_samples/Grafana) that can log in to our services using Client Credentials, but would like to enable users to log in with their own credentials instead. Hybrid flow is our best option, as in our services this allows for a refresh token which Grafana can use. However, this flow does require that the identity server does a form_post back as a response instead of a traditional response uri, which required some changes to api.go and login_oauth.go to handle the response.
In addition, I have also included an optional `acr_values` parameter which our system uses to specify what organization (tenant) the user belongs to. 

N/A

I am open to instead including a new OAuth provider for OSIsoft Cloud Services specifically that uses Hybrid flow without adding the `use_hybrid_flow` and `acr_values` parameters to the Generic OAuth section. I would need to discuss this internally before submitting a new PR but I do consider this to be an option if it's not desirable to modify the Generic OAuth code. 

I am also open to renaming `use_hybrid_flow`, or removing `acr_values` as the latter is not technically required to log in to our services, it simply allows us to bypass our browser prompt for the user's organization.


marefr:
Thank you for your contribution. It's hard for us to accept this currently given your description and that there are no referenced feature requests nor comments/upvotes of it. OAuth in Grafana is already hard to maintain with all special cases. To accept changes we would need something that we identify as useful for many users. I suggest you open a new feature request (if there's not one already for this) explaining your use case and why you have/want/need to use that. We're closing this for now. We'll consider reopen this if the feature request gains traction from many users and/or if we misunderstood something.

> new OAuth provider for OSIsoft Cloud Services specifically

Sorry, but we cannot maintain code for every provider.

PR #26311, opened 2020-07-13
DashboardLinks: do not over-query search endpoint

### LINKED ISSUE #26289: Dashboard links: tag searches being done per panel instead of per dashboard
**What happened**: When a dashboard configured with dashboard links to list dashboards by tag, when dashboard being accessed/refreshed, it should really just run the search to "/api/search?tag=XYZ&limit=100" once, but it's running numerous times(one for each panel), this cause lots of requests stalled while loading a dashboard with many panels, due to Chrome's "6 connections per origin" limit.

**What you expected to happen**: The search should be run once only

**How to reproduce it (as minimally and precisely as possible)**: Create a dashboard links by tag in the dashboard with 10 panels, load the dashboard and through web inspector you will see the tag searches being run 10+ times. 

**Environment**: 
- Grafana version: Grafana 7.0.5
- Data source type & version: Unrelated, but we are using Metrictank
- OS Grafana is installed on: RHEL 7.6
- User OS & Browser: MacOS, Chrome or Safari
- Grafana plugins: 
- Others:

torkelo:
Tries to fix the dashboard links issue that looks pretty big: 
Fixes #26289 

Dashboard to replicate issue: 
https://gist.github.com/torkelo/34628b4904befe0ba0bac6df9fbee232

Problem
* 1) SubMenu is connected to redux but is using a getSubMenuVariables selector that filters and creates a new array every redux update which causes SubMenu to render on every redux update (and as you scroll down a big dashboard there are redux updates as panels get's loaded)
* 2) DashboardLinks build linkInfo models every render, and DashboardLinksDashboards component will search for new update every time linkInfo model changed  => lots of calls to dashboard search api 

Solution
1) Memoize the call to getSubMenuVariables , was not easy as it takes in state and not the thing we want to memomize on (state variables object), had to do a bit of a hacky soliution without changing all get variable selector functions 
2) Update DashboardLinksDashboards way of detecting changes 

New Problem (why this is WIP):
1) Now DashboardLinksDashboards does not update when the link definition updates (need to make the link definition update mutable so a new link object is created), or switch to lodash deep equals in componentDidUpdate

### DIFF
--- a/public/app/features/dashboard/components/DashLinks/DashLinksEditorCtrl.ts
+++ b/public/app/features/dashboard/components/DashLinks/DashLinksEditorCtrl.ts
@@ -65,7 +65,7 @@ export class DashLinksEditorCtrl {
   }
 
   saveLink() {
-    this.dashboard.links = [...this.dashboard.links];
+    this.dashboard.links = _.cloneDeep(this.dashboard.links);
     this.backToList();
   }
 
--- a/public/app/features/dashboard/components/DashLinks/DashLinksEditorCtrl.ts
+++ b/public/app/features/dashboard/components/DashLinks/DashLinksEditorCtrl.ts
@@ -54,7 +54,7 @@ export class DashLinksEditorCtrl {
   };
 
   addLink() {
-    this.dashboard.links.push(this.link);
+    this.dashboard.links = [...this.dashboard.links, this.link];
     this.mode = 'list';
     this.dashboard.updateSubmenuVisibility();
   }
@@ -65,6 +65,7 @@ export class DashLinksEditorCtrl {
   }
 
   saveLink() {
+    this.dashboard.links = [...this.dashboard.links];
     this.backToList();
   }
 
--- a/public/app/features/dashboard/components/SubMenu/SubMenu.tsx
+++ b/public/app/features/dashboard/components/SubMenu/SubMenu.tsx
@@ -7,6 +7,7 @@ import { DashboardModel } from '../../state';
 import { DashboardLinks } from './DashboardLinks';
 import { Annotations } from './Annotations';
 import { SubMenuItems } from './SubMenuItems';
+import { DashboardLink } from '../../state/DashboardModel';
 
 interface OwnProps {
   dashboard: DashboardModel;
@@ -14,6 +15,7 @@ interface OwnProps {
 
 interface ConnectedProps {
   variables: VariableModel[];
+  links: DashboardLink[];
 }
 
 interface DispatchProps {}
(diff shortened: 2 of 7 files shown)


torkelo:
@dprokop I am off tomorrow if you have time to look at this that would be great

public/app/features/dashboard/components/SubMenu/SubMenu.tsx
```diff
@@ -66,9 +68,17 @@ class SubMenuUnConnected extends PureComponent<Props> {
...
-});
+const mapStateToProps: MapStateToProps<ConnectedProps, OwnProps, StoreState> = state => {
+  return {
+    variables: getSubMenuVariables(state.templating.variables),
+
+    /** Haven't found any other way to force links being passed to the links menu. Not spreading into new array makes
+     * the DashboardLinks not re-render as it gets reference to the same array in the model.
+     * Ref: https://github.com/grafana/grafana/issues/26289
+     */
+    links: [...state.dashboard.getModel()?.links],
```
dprokop, line 79: @torkelo I haven't found any other "quick" way to fix the links rendering issue and it feels hacky too. Think the mutable dashboard is problematic, as the DashboardLinks is not re-rendered, as it gets reference to the same dashboard object always. I think this is OK to force the re-render - `links` prop is used only for this and the rendering of the actual link happens based on the links from the dashboard model passed to the DashboarLinks component.

torkelo: This Will cause re render on every redux update, exactly why I did add the memorization for variables array.

Much better make parent send in links as normal prop just dashboard.links 

Then have editor create a new array in any links change

dprokop: will try this approach too

torkelo: But maybe tricky , as editor is still angular ?

dprokop: yes, unfortunately

dprokop: Another approach, with replacing links array in the model: https://github.com/grafana/grafana/pull/26311/commits/a2bb30550ce30216b22199a948171d309f693f7a

torkelo: Yea, think that is better

public/app/features/dashboard/components/DashLinks/DashLinksEditorCtrl.ts
```diff
@@ -65,6 +65,7 @@ export class DashLinksEditorCtrl {
   }
 
   saveLink() {
+    this.dashboard.links = [...this.dashboard.links];
```
torkelo, line 68: I think this needs to do a deep clone as the DashboardLinksDashboard checks link equality to figure out out if another search for dashboards should be triggered.

You could test this by changing the tag in a link definition and then go back to dashboard, does list of links change?

dprokop: You are right, fixed

dprokop (approved):

PR #26315, opened 2020-07-14
Admin: Fix updating organization name not updating configuration subtitle

### LINKED ISSUE #26314: Updating organization name fails to update Configuration subtitle
Gif of the issue:
Grafana Org Name Update Bug

**What happened**:
Updating the organization name failed to update the Configuration subtitle until the user refreshes

I expected all instances of the organization name to be renamed to the updated organization name.

1. Go to `/org` (Configuration -> Preferences) 
2. Change the organization name and click `Update organization name`
3. Notice the configuration subtitle still contains the old organization name

I currently have a branch that is almost PR-ready that will fix the issue, but the fix is a bit involved since it requires updating the `navIndex` state in redux (`cfg` plus all of the children nodes for `cfg`). Is that an acceptable solution or is there a better alternative? It looks like the original configuration subtitle comes back hardcoded with the current organization name from the server side, so there may be a better way to fix this server side that I'm not privy to. All suggestions/feedback on the solution are welcome. I intend on opening a PR to fix this via a new Redux action in the next day or two if I don't hear anything back. 

**Environment**:
- Grafana version: Current `master` branch (747bdaf2e78d74e2d596f24fe049abe0747f0f57) Running locally
- OS Grafana is installed on: Mac
- User OS & Browser: macOS Catalina Version 10.15.5; Brave Browser Version 1.10.97 Chromium: 83.0.4103.116 (Official Build) (64-bit)
- Grafana plugins: N/A
- Others: N/A

KrammerJake:
Updates the configuration subtitle to reflect the updated organization name without the user having to refresh first.

Fixes #26314 

### DIFF
--- a/public/app/core/reducers/navModel.test.ts
+++ b/public/app/core/reducers/navModel.test.ts
@@ -50,7 +50,7 @@ describe('navModelReducer', () => {
         users: { ...initialState.users, parentItem: newCfgState },
         teams: { ...initialState.teams, parentItem: newCfgState },
         plugins: { ...initialState.plugins, parentItem: newCfgState },
-        ['org-settings']: { ...initialState['org-settings'], parentItem: newCfgState },
+        'org-settings': { ...initialState['org-settings'], parentItem: newCfgState },
         apikeys: { ...initialState.apikeys, parentItem: newCfgState },
       };
 
--- a/public/app/core/reducers/navModel.test.ts
+++ b/public/app/core/reducers/navModel.test.ts
@@ -1,5 +1,5 @@
 import { reducerTester } from '../../../test/core/redux/reducerTester';
-import { getItemWithNewSubTitle, navIndexReducer, updateNavIndex, updateConfigurationSubtitle } from './navModel';
+import { navIndexReducer, updateNavIndex, updateConfigurationSubtitle } from './navModel';
 import { NavIndex } from '@grafana/data';
 
 describe('navModelReducer', () => {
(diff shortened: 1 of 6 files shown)


public/app/core/reducers/navModel.test.ts
```diff
@@ -38,4 +38,26 @@ describe('applicationReducer', () => {
...
+      const orgName = 'New Org Name';
+      const newCfgState = { ...initialState.cfg, subTitle: `Organization: ${orgName}` };
+      const expectedState = {
+        ...initialState,
+        cfg: newCfgState,
+        datasources: { ...initialState.datasources, parentItem: newCfgState },
+        users: { ...initialState.users, parentItem: newCfgState },
+        teams: { ...initialState.teams, parentItem: newCfgState },
+        plugins: { ...initialState.plugins, parentItem: newCfgState },
+        ['org-settings']: { ...initialState['org-settings'], parentItem: newCfgState },
```
atymchuk, line 53:
```suggestion
        'org-settings': { ...initialState['org-settings'], parentItem: newCfgState },
```
Square brackets are normally used for calculated properties, which is not the case.

ademidoff (changes requested):

public/app/features/org/state/actions.ts
```diff
@@ -18,5 +19,6 @@ export function updateOrganization(): ThunkResult<any> {
     await getBackendSrv().put('/api/org', { name: organization.name });
 
     dispatch(loadOrganization());
+    dispatch(updateConfigurationSubtitle(organization.name));
```
hugohaggmark, line 22: Skilled work on this, but the order between these two actions seems wrong. We should update first and then load the org. WDYT @KrammerJake ? Also, I would like to add this new behaviour in a test, let me know if I can help out with that.

public/app/core/selectors/navModel.ts
```diff
@@ -1,6 +1,6 @@
 import { NavModel, NavModelItem, NavIndex } from '@grafana/data';
 
-function getNotFoundModel(): NavModel {
+const getNotFoundModel = (): NavModel => {
```
hugohaggmark, line 3: Nice work on this refactor ⭐

public/app/core/reducers/navModel.ts
```diff
@@ -60,6 +62,19 @@ export const navIndexReducer = (state: NavIndex = initialState, action: AnyActio
...
+    return {
+      ...state,
+      cfg: newCfgState,
+      datasources: { ...state.datasources, parentItem: newCfgState },
+      users: { ...state.users, parentItem: newCfgState },
+      teams: { ...state.teams, parentItem: newCfgState },
+      plugins: { ...state.plugins, parentItem: newCfgState },
+      'org-settings': { ...state['org-settings'], parentItem: newCfgState },
+      apikeys: { ...state.apikeys, parentItem: newCfgState },
+    };
```
hugohaggmark, line 77: It looks like we're replacing the complete parentItem every line here. I think something like this is the safer way to mutate this state, wdyt @KrammerJake 
```suggestion
         const subTitle = `Organization: ${action.payload}`;
     
         return {
      ...state,
      cfg: { ...state.cfg, subTitle },
      datasources: getItemWithNewSubTitle(state.datasources, subTitle),
      users: getItemWithNewSubTitle(state.users, subTitle),
      teams: getItemWithNewSubTitle(state.teams, subTitle),
      plugins: getItemWithNewSubTitle(state.plugins, subTitle),
      'org-settings': getItemWithNewSubTitle(state['org-settings'], subTitle),
      apikeys: getItemWithNewSubTitle(state.apikeys, subTitle),
    };

    };
    
    // add this outside the reducer
    const getItemWithNewSubTitle = (item: NavModelItem, subTitle: string): NavModelItem => {
  return { ...item, parentItem: { ...item.parentItem, text: item.parentItem?.text ?? '', subTitle } };
};
```

hugohaggmark (commented):
Thank you for contributing to Grafana. This is looking good. There are some issues left that I mention in my comments. Let me know if something is unclear.

public/app/features/org/state/actions.ts
KrammerJake: I agree, the order should be flipped. I made an attempt at adding a test for this behavior in my latest commit (`public/app/features/org/state/actions.test.ts`) based on what I found elsewhere in the codebase. Let me know if I made any mistakes or if you'd like me to extend it to test additional behavior.

public/app/core/reducers/navModel.ts
KrammerJake: I agree, your way seems safer and more precise.

public/app/core/reducers/navModel.test.ts
```diff
@@ -38,4 +37,123 @@ describe('applicationReducer', () => {
...
+      const subTitle = `Organization: ${orgName}`;
+      const expectedState = {
+        ...initialState,
+        cfg: { ...initialState.cfg, subTitle },
+        datasources: getItemWithNewSubTitle(initialState.datasources, subTitle),
+        users: getItemWithNewSubTitle(initialState.users, subTitle),
+        teams: getItemWithNewSubTitle(initialState.teams, subTitle),
+        plugins: getItemWithNewSubTitle(initialState.plugins, subTitle),
+        'org-settings': getItemWithNewSubTitle(initialState['org-settings'], subTitle),
+        apikeys: getItemWithNewSubTitle(initialState.apikeys, subTitle),
```
hugohaggmark, line 150: Significant work adding tests but I think the expect part of any test should be raw and never use parts of the code it tests. I mean you shouldn't use `getItemWithNewSubTitle` in this expect but create the expected state yourself. Does that make sense @KrammerJake

public/app/features/org/state/actions.ts
```diff
@@ -17,6 +18,7 @@ export function updateOrganization(): ThunkResult<any> {
 
     await getBackendSrv().put('/api/org', { name: organization.name });
```
hugohaggmark, line 19: nitpick: you might want to move out `getBackendSrv` as a dependency like we do here https://github.com/grafana/grafana/blob/5a58f2b4c51b38c2fc03c1c5acf30b977c66271b/public/app/features/variables/state/actions.ts#L465 
This will make it easier to test `updateOrganization` without using jest.mock that can be limiting in some cases.

public/app/core/reducers/navModel.test.ts
KrammerJake: That makes perfect sense. I'll remove the usage of `getItemWithNewSubTitle` in my next commit.

torkelo: this initial state seems to big for what we are testing, is all this needed?

KrammerJake: @torkelo I reduced the size of the initial state in my latest commit and only kept the necessities. Don't know what I was thinking. 😄

public/app/features/org/state/actions.ts
KrammerJake: I have addressed this in my latest commit. 

I also went ahead and added `getBackendSrv` as a dependency to `loadOrganization` as well, since it seemed to be causing my test to fail. 

One potential concern I want to call out is that I pass the `dependencies` parameter from `updateOrganization` to `loadOrganization`, which I am not sure is correct. Removing this caused me to get a type error when running my test that said `get` was `undefined`. Adding the `dependencies` parameter and passing it in from `updateOrganization` resolved that issue. Is there a better solution?

hugohaggmark: Aha, good catch didn't notice that chain. I think it's fine as it is the same dependency. So you could declare one type `type OrganizationDependencies = { getBackendSrv: typeof getBackendSrv };` and use that in both thunks.

Another alternative is to remove the `loadOrganization` call in `updateOrganization` and add that call after this line https://github.com/grafana/grafana/blob/eed313f3129d2e1bb321ce28c6e859e623045f8b/public/app/features/org/OrgDetailsPage.tsx#L30

 I think this is the last comment I've left, outstanding work @KrammerJake  🌟

KrammerJake: Thank you! 🙇 I appreciate the helpful feedback! 

I implemented your first suggestion and consolidated the organization dependency type. I liked your second suggestion of moving the `loadOrganization` dispatch outside of `updateOrganization` thunk that you mentioned above; however, I was worried I potentially might introduce a race condition since we weren't `await`ing the thunk dispatches in `onUpdateOrganization` and the `get` request might return before the `put` finishes? I was thinking if we moved the `loadOrganization` call, we would likely want to rewrite `onUpdateOrganization` like this:

```js
  onUpdateOrganization = async (orgName: string) => {
    this.props.setOrganizationName(orgName);
    await this.props.updateOrganization();
    await this.props.loadOrganization();
  };
```

Let me know if you'd like me to make that change.

hugohaggmark (approved):
Thank you for contributing @KrammerJake , this looks great and works as expected ⭐

Issue #25967, opened 2020-07-01
Alerting: Alert notifiers should store sensitive settings encrypted 


marefr:
**What would you like to be added**:
Following #25113 which adds the basic support for storing sensitive settings encrypted we need to update all alert notifiers so that they support storing sensitive information encrypted.

Alert notifiers:
- [x] Slack (url and token) #25114
- [x] Webhook (password) #26164
- [x] Pagerduty (integration key) #26319
- [x] Alertmanager (Basic Auth Password) #26318
- [x] Opsgenie (API Key) #27074
- [x] Sensu (Password, possibly) #27131
- [x] Telegram (BOT API Token) #27122
- [x] LINE (token) #27157
- [x] Pushover (API Token, User key(s)?) #27159
- [x] Threema Gateway (API Secret) #27160

The following notifiers have not been converted to store sensitive settings encrypted because we don't believe they have any sensitive information and/or encrypting it would make it hard for administrators to understand what configuration is used for certain notifiers:
- [ ] VictorOps (Url?) 
- [ ] Google Hangouts Chat (Url)
- [ ] Microsoft Teams (URL)
- [ ] Discord (Webhook URL)
- [ ] DingDing (url)

mtanda:
I start to work Alertmanager.

marefr:
Fixed by #25114
Fixed by #26164
Fixed by #26319
Fixed by #26318
Fixed by  #27074
Fixed by #27131
Fixed by #27122
Fixed by #27157
Fixed by #27159
Fixed by #27160

PR #26318, opened 2020-07-14
Alerting: Store sensitive settings encrypted for alertmanager notifier

mtanda:
Store basic auth user/password in secure setting.

https://github.com/grafana/grafana/issues/25967


marefr (changes requested):
Thanks. Looks good. Suggests we encrypt basic auth password, but not basic auth username. In #26170 we decided to skip encrypting username because we don't do that for data sources and in addition could be hard for users to understand what account is used when not seeing the username.

mtanda:
I revert basicAuthUser change. Please check :-)

marefr (approved):
LGTM

marefr:

PR #26320, opened 2020-07-14
Grafana-UI: Enable empty time range

Clarity-89:
Add a possibility to use `TimeRangeInput` in an empty state, without initial date value as well as enabling clearing the input.

Screenshot 2020-07-14 at 12 40 10

Screenshot 2020-07-14 at 12 40 18

### DIFF
--- a/packages/grafana-data/src/datetime/moment_wrapper.ts
+++ b/packages/grafana-data/src/datetime/moment_wrapper.ts
@@ -5,7 +5,7 @@ export interface DateTimeBuiltinFormat {
   __momentBuiltinFormatBrand: any;
 }
 export const ISO_8601: DateTimeBuiltinFormat = moment.ISO_8601;
-export type DateTimeInput = Date | string | number | Array<string | number> | DateTime; // null | undefined
+export type DateTimeInput = Date | string | number | Array<string | number> | DateTime | null; // | undefined;
 export type FormatInput = string | DateTimeBuiltinFormat | undefined;
 export type DurationInput = string | number | DateTimeDuration;
 export type DurationUnit =
--- a/packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
+++ b/packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
@@ -17,7 +17,7 @@ export const defaultTimeRange: TimeRange = {
   raw: { from: 'now-6h', to: 'now' },
 };
 
-interface InputTimeRange {
+export interface InputTimeRange {
   from: DateTime | string;
   to: DateTime | string;
   raw: RawTimeRange;
--- a/packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
+++ b/packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
@@ -132,7 +132,7 @@ const getStyles = (theme: GrafanaTheme) => {
     clearIcon: css`
       margin-right: ${theme.spacing.xs};
       &:hover {
-        color: ${theme.palette.white};
+        color: ${theme.colors.linkHover};
       }
     `,
     placeholder: css`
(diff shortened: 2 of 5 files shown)


packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -45,19 +66,34 @@ export const TimeRangeInput: FC<Props> = ({
     onChange(timeRange);
   };
 
+  const onRangeClear = (event: MouseEvent<HTMLDivElement>) => {
+    event.stopPropagation();
+    const from = '';
+    const to = '';
+    onChange({ from, to, raw: { from, to } });
```
dprokop, line 73: I think we should allow `dateTime(null)`. moment  accepts the null values producing date object that is invalid (isValid === false). Ref https://stackoverflow.com/a/57817236

dprokop (commented):
Left a few comments. I'm mostly concerned with `InputTimeRange` interface that feels unnecessary and would love to get rid of it.

dprokop:
UX/UI wise this looks great!

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -9,13 +9,32 @@ import { getFocusStyle } from '../Forms/commonStyles';
...
+  from: dateTime().subtract(6, 'hour'),
+  to: dateTime(),
+  raw: { from: 'now-6h', to: 'now' },
+};
+
+export interface InputTimeRange {
+  from: DateTime | string;
+  to: DateTime | string;
+  raw: RawTimeRange;
+}
```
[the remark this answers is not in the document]
Clarity-89: We need to have some kind of empty input value for the time range. For example in the report form if range is not specified that means that custom time range isn't used. The current `TimeRange` interface doesn't allow empty time ranges, so I had to make a separate on for the input, where an empty string in the range would signify no value. 

Agree, it doesn't seem like the cleanest solution but I couldn't figure out a better way to do it.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
Clarity-89: Yeah, that seems like a good option. It's just the api often stores time values in a string and returns empty strings when there's no value.

Clarity-89: Ok, now I see what you mean, indeed seems like `dateTime(null)` is working, so I can get rid of `InputTimeRange` interface. Just need to mention in the docs that `dateTime(null)` should be provided for time range values.

packages/grafana-data/src/datetime/moment_wrapper.ts
```diff
@@ -5,7 +5,7 @@ export interface DateTimeBuiltinFormat {
   __momentBuiltinFormatBrand: any;
 }
 export const ISO_8601: DateTimeBuiltinFormat = moment.ISO_8601;
-export type DateTimeInput = Date | string | number | Array<string | number> | DateTime; // null | undefined
+export type DateTimeInput = Date | string | number | Array<string | number> | DateTime | null; // | undefined;
```
Clarity-89, line 8: Had to uncomment this to allow null values, no idea why it was commented out in the first place.

dprokop: Yeah, no idea neither, but it should be OK

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
dprokop: perfect!

dprokop (approved):
@Clarity-89 looks great, awesome that you got rid of that specific interface for clearable values. Awesome job!

PR #26321, opened 2020-07-14
Drone: Tighten grafana-enterprise access to token

aknuds1:
In Drone shared module, tighten access to GitHub token for enterprise builds as much as possible (only `git` basically). No practical effects for this repository's builds, except I make the build-docs-website step depend on the initialize step (necessary for enterprise builds).

**NB:** .drone.yml is auto-generated from .drone.star, and is what is actually used to configure our Drone CI. You may read it to see how the actual configuration changes (very little for this PR).


jessabe (approved):
LGTM

PR #26329, opened 2020-07-14
InfluxDB: flux query result can cause crash

### LINKED ISSUE #26040: InfluxDB Flux plugin - use of 'map' causes internal server error
**What happened**:
I was trying to create an aggregate of values across multiple InfluxDB 2.0 tables and graph them. The query used to do this works in InfluxDB 2.0's data explorer (showing tables and graphs), but it causes an internal server error after the Flux function `map` is used, with no graph appearing (it says there's an error log, but I'm using grafana via a Docker container. Opening the container's CLI and going to var/log/grafana yields no results.)

image

With data looking like this: 
image

I was able to create a mean aggregate using this:
```
from(bucket: "MyBucket")
  |> range(start: -1d, stop: 1d)
  |> filter(fn: (r) => r["_measurement"] == "AvgData")
  |> filter(fn: (r) => r["_field"] == "Mean Information")
  |> group(columns: ["_time", "_field"], mode:"by")
  |> mean()
  |> map(fn: (r) => ({
  time: r._time,
  MeanAggregate: r._value
  }))
```
Which can create a graph in InfluxDB 2.0's data explorer:

image

Create a set of example data set up like the image above, then try entering the flux query into both the InfluxDB data explorer and Grafana.

I'm guessing this is either because `map` functionality is not implemented yet, or because there's something strange occurring with the use of names here.

**Environment**:
- Grafana version: v7.1.0-beta2
- Data source type & version: InfluxDB, specifically using the Flux query language
- OS Grafana is installed on: Installed as Docker Container within Ubuntu Virtual Machine
- User OS & Browser: Metrics page Viewed using Microsoft Edge (Chromium version) on Windows 10
- Grafana plugins: None other than default
- Others:

ryantxu:
Fixes: #26040

Currently this just has test import that can cause the panic/crash.

With https://github.com/grafana/grafana-plugin-sdk-go/pull/204 the panic will be replaced with an error


ryantxu:
hymmm -- changes to the test structure will not merge cleanly with 7.1, but this fix should be in 7.1.... ideas?

kylebrandt:
I tagged 0.75.0 for the error instead of panic - should I make a new PR or do you want to include it in this?

ryantxu:
@kylebrandt -- I'll include it here

kylebrandt:
> hymmm -- changes to the test structure will not merge cleanly with 7.1, but this fix should be in 7.1.... ideas?

I'm not sure what you mean here. Changes to the way tests are being done, are the data generated from the tests?

ryantxu:
The 'golden file' stuff was added in 7.2, so I think the test will not merge cleanly with 7.1... I think we should merge this, and I will make another one that can cleanly be cherry picked to 7.1

kylebrandt (approved):

kylebrandt:
If doing another PR for the CP, be sure to remove the CP label from this issue.

PR #26331, opened 2020-07-14
Influx: fix flux panic.  cherry picks for #26329

ryantxu:
The go and test files in #26329 do not cleanly cherry pick to 7.1

This is a PR with explicit changes against 7.1.  I'm not sure the best process here :man_shrugging:


ryantxu:
@dprokop -- I added the `cherry-pick` label here since it needs to be added to 7.1, BUT note this pr is agains the 7.1.x branch since #26329 will not cherry pick cleanly.

What is the best practice here?

papagian (commented):
LGTM

The CI fails to build enterprise because it targets the v7.1.x branch and tries to checkout an invalid OSS branch.
@dprokop We need to change it to target the cherry-picks PR once it will be created and merge it together with the others.

dprokop:
@papagian haven't we agreed to cherry pick https://github.com/grafana/grafana/pull/26329 instead?

papagian:
> @papagian haven't we agreed to cherry pick #26329 instead?X

we cannot because the `verifyGoldenResponse()` that is called by the test does not exist in #26329 (it is introduced in another commit that it [will not](https://github.com/grafana/grafana/pull/26329#issuecomment-658328198) be included in 7.1)

dprokop:
OK, then i'll merge this one to the cherry picks branch 👍

pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -273,3 +285,12 @@ func TestBuckets(t *testing.T) {
 		fmt.Println("----------------------")
 	})
 }
+
+func TestGoldenFiles(t *testing.T) {
+	t.Run("Renamed", func(t *testing.T) {
+		_, err := verifyGoldenResponse("renamed")
+		if err != nil {
+			t.Fatal(err.Error())
+		}
```
aknuds1, line 294: require.NoError(t, err)

aknuds1 (approved):
LGTM, with a suggestion.

pkg/tsdb/influxdb/flux/executor_test.go
papagian: no objection; only that this is already merged in master in #26329 if we accept the suggestion here we need to change it also in the master.

aknuds1: Good point @papagian. Might as well not change this.

PR #26339, opened 2020-07-15
Database: Set 0640 permissions on SQLite database file

### LINKED ISSUE #8283: [Bug] Wrong permissions in grafana package for grafana.db
It looks like in https://github.com/grafana/grafana/issues/2126 there were plans to lock down the sqlite DB to 0600. While grafana.ini did get locked down, the DB did not. Unless this is no longer believed to be necessary I could try to send a PR.

- What Grafana version are you using? 4.2.0
- What datasource are you using? InfluxDB
- What OS are you running grafana on? CentOS 7.3.1611
- What did you do? Logged into grafana.db as a regular user 
- What was the expected result? Permission denied
- What happened instead? Logged in without issue, able to query salt and password for users
 
```
[vagrant@monitoring ~]$ less /etc/grafana/grafana.ini 
/etc/grafana/grafana.ini: Permission denied
[vagrant@monitoring ~]$ sqlite3 /var/lib/grafana/grafana.db 
SQLite version 3.7.17 2013-05-20 00:56:22
Enter ".help" for instructions
Enter SQL statements terminated with a ";"
sqlite> select salt, password from user;
SALT|PW
[vagrant@noc-monitoring ~]$ sudo yum info grafana
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: centos.chi.host-engine.com
 * extras: mirror.sigmanet.com
 * updates: mirror.sigmanet.com
Installed Packages
Name        : grafana
Arch        : x86_64
Version     : 4.2.0
Release     : 1
Size        : 128 M
Repo        : installed
From repo   : grafana
Summary     : Grafana
URL         : https://grafana.com
License     : "Apache 2.0"
Description : Grafana
```

aknuds1:
#8283 points out that the we should restrict access to the SQLite database file, so only the owner can use it. This PR ensures the file has permissions 0640 (read/write for owner, read for group) when it gets created. If the file already exists and has broader permissions, a warning will be logged.

@wz2b @austinbutler does this look right to you?

Fixes #8283.


papagian (approved):
I have tested both with existing and not existing sqlite database file and with mysql database and it works as expected.

aknuds1:
> I have tested both with existing and not existing sqlite database file and with mysql database and it works as expected.

Thanks @papagian - do you have any opinion on whether we should change the permissions of existing SQLite files or not? Maybe just log a warning to the administrator?

papagian:
> Thanks @papagian - do you have any opinion on whether we should change the permissions of existing SQLite files or not? Maybe just log a warning to the administrator? I'm just worried in case there are valid scenarios where the permissions (or owner?) should be changed by the operator.

I didn't think about it. I'm not sure. Maybe the admin could overwrite the permissions by the configuration.

wz2b:
> > Thanks @papagian - do you have any opinion on whether we should change the permissions of existing SQLite files or not? Maybe just log a warning to the administrator? I'm just worried in case there are valid scenarios where the permissions (or owner?) should be changed by the operator.

> I didn't think about it. I'm not sure. Maybe the admin could overwrite the permissions by the configuration.

I think you should.  Not only is it the user table but also the data_source table.  In theory plugins will store passwords encrypted in secureJsonData but they don't all.  Just an an example, the Zabbix datasource does store in secureJsonData now, but it didn't always, and if you upgrade the datasource it doesn't clear out the plaintext 'password' column that was there previously (even though it doesn't use it any more).  What that means is if somebody didn't set up a zabbix service account with the right permissions, you can steal this password from grafana.db and do bad things to zabbix.  We don't want somebody to be able to point to some exploit and have the reason be that grafana allowed the password to escape.

kylebrandt:
> > Thanks @papagian - do you have any opinion on whether we should change the permissions of existing SQLite files or not? Maybe just log a warning to the administrator? I'm just worried in case there are valid scenarios where the permissions (or owner?) should be changed by the operator.
> 
> I didn't think about it. I'm not sure. Maybe the admin could overwrite the permissions by the configuration.

One may want to give a group of users access to the sqlite db. So forcing group permissions to 0 could be problematic.

wz2b:
> One may want to give a group of users access to the sqlite db. So forcing group permissions to 0 could be problematic.

The Ubuntu package (maybe all debian) yields this:

  -rw-r--r-- 1 grafana grafana 260056 July 15 13:29 /var/lib/grafana/grafana.db

so it creates a 'grafana' user and a 'grafana' group.  So I'd be fine with the permissions being 640 (u+rw,g+r if I did that right).  Normally there's nobody in the 'grafana' group other than the 'grafana' user.  So, this way, if somebody did have some weird reason to add a second user to the 'grafana' group (like say for backing it up) that would still work.

Regarding changing an existing file's permissions if it exists: I think this is a good idea, but it's not something I would do on the go side.  I would have grafana set 0640 if it creates the file; but as for changing a file that already exists, in thinking about this, it's really something I would put into the postinst script.  If you want to be nice you'd even make it prompt the user whether or not they wanted you to do it.

kylebrandt:
I don't know enough about the various files the Grafana process can create, or the packaging ... but my fuzzy old knowledge says that maybe we should be setting the umask for the process so all things created are only accessible by grafana user/group - but I don't know if that might cause problems - and don't think it should be a blocker getting more secure specific DB perms in place.

marefr:
Think logging a warning if permissions are to "broad" on sqlite file should be enough. Then handle the migration thru documentation because older Grafana instances can apply that change manually without upgrading, which is good from a security perspective.

aknuds1:
I'll change the permissions to 0640 give the group read access. That sound alright to you @papagian @kylebrandt @marefr?

@papagian @kylebrandt Do you agree with @marefr about just logging a warning if the file exists with broader permissions?

papagian:
> I'll change the permissions to 0640 give the group read access. That sound alright to you @papagian @kylebrandt @marefr?

Being created with the `grafana` group is not always the case. For instance if you are in development mode then the file is created with the current user/group. However, I think that restricting read access to the group can have implications so I would say yes.

> @papagian @kylebrandt Do you agree with @marefr about just logging a warning if the file exists with broader permissions?

yes, I think that is much better.

aknuds1:
I made a change to only log a warning if the SQLite file permissions are broader than expected (0640).

aknuds1:
@marefr @kylebrandt @papagian Please have another look.

pkg/services/sqlstore/sqlstore.go
```diff
@@ -265,6 +266,32 @@ func (ss *SqlStore) getEngine() (*xorm.Engine, error) {
...
+		}
+
+		const perms = 0640
+		if !exists {
+			ss.log.Info("Creating SQLite database file", "path", ss.dbCfg.Path)
+			f, err := os.OpenFile(ss.dbCfg.Path, os.O_CREATE|os.O_RDWR, perms)
+			if err != nil {
+				return nil, errutil.Wrapf(err, "failed to create SQLite database file %q", ss.dbCfg.Path)
+			}
+			f.Close()
```
kylebrandt, line 282: Should maybe check for the error on close and return it? I'm not sure if close already being called is the only possible error case or not.

Also maybe this can be its own function

kylebrandt (approved):
Two possible code considerations for checking an error and moving this to its own function, but leave those to your judgement.

I agree with the logging choice.

pkg/services/sqlstore/sqlstore.go
aknuds1: Good point @kylebrandt on reporting error on close. I actually avoided checking it on purpose, but realize in retrospect that a closing error here could mean it didn't get created.

papagian:
I've tested it again and works as expected in my development environment.

PR #26342, opened 2020-07-15
Azure: Handle real type nan/inf values in Log/Insights Analytics

kylebrandt:
This handle NaN/-Inf/Inf real values when converting to a dataframe.  Before this, if the user were to divide by 0.0, "Infinity would be returned in the result and the user would get an error: "unexpected type, expected json.Number but got string".

Now those values are properly set as NaN/Inf values.

cc @wardbekker


papagian (approved):
LGTM; I have tested and works as expected!

aknuds1 (approved):

PR #26349, opened 2020-07-15
Explore: parse queryType from explore url

zoltanbedi:

In explore view with x-ray data source we persist the query type in the url. For it to work properly we need to parse it from the url. This way when we reload the page we will get back the right query type.

### DIFF
--- a/public/app/core/utils/explore.ts
+++ b/public/app/core/utils/explore.ts
@@ -246,8 +246,7 @@ export function parseUrlState(initial: string | undefined): ExploreUrlState {
   };
   const datasource = parsed[ParseUrlStateIndex.Datasource];
   const parsedSegments = parsed.slice(ParseUrlStateIndex.SegmentsStart);
-  const metricProperties = ['expr', 'expression', 'target', 'datasource', 'query', 'queryType'];
-  const queries = parsedSegments.filter(segment => isSegment(segment, ...metricProperties));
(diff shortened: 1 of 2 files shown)


ivanahuckova (approved):
I am just able to review the code and added test (and not test it with x-ray data source). But code and test look good. LGTM!  🚀

public/app/core/utils/explore.ts
```diff
@@ -246,7 +246,7 @@ export function parseUrlState(initial: string | undefined): ExploreUrlState {
   };
   const datasource = parsed[ParseUrlStateIndex.Datasource];
   const parsedSegments = parsed.slice(ParseUrlStateIndex.SegmentsStart);
-  const metricProperties = ['expr', 'expression', 'target', 'datasource', 'query'];
+  const metricProperties = ['expr', 'expression', 'target', 'datasource', 'query', 'queryType'];
```
aocenas, line 249: Ugh it took me a while to realize what is going on here. This is pretty brittle. I don't think we can count on these properties in any way and there is no type safety or anything here. Ideally we would change the url a bit but that's probably going to be a bit more invasive. What I think we can do here is instead get all the segments we know the structure of so `ui` and `originPanelId` and then just treat all that are left as queries. There does not seem to be any code path where there would be something that we just need to ignore. This will prevent this kind of issues for datasources which will want to structure their query differently and I think should be safe in general (but please test a bit if that's true.)

aocenas (approved):
Looks good nice work

PR #26360, opened 2020-07-15
Drone: Rename publish-storybook to build-storybook

aknuds1:
Rename Drone step publish-storybook to build-storybook, since we won't publish for PR builds (confirmed with front-end team).


sakjur (approved):
I like that you refactored it into a function as well 👍

PR #26367, opened 2020-07-16
Drone: Enable master branch testing

aknuds1:
Enable testing of master branch in Drone.

Publishing steps are included, but only simulated until we disable CircleCI's master pipeline.

**NB:** .drone.yml is auto-generated from .drone.star, but you might want to read it just to verify the Drone configuration.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -60,9 +60,7 @@ steps:
   image: grafana/build-container:1.2.21
   commands:
   - rm -rf $(go env GOCACHE) && cp -r go-cache $(go env GOCACHE)
-  - ./bin/grabpl build-backend --edition $${GF_EDITION} --build-id $DRONE_BUILD_NUMBER --variants linux-x64,linux-x64-musl,osx64,win64 --no-pull-enterprise
-  environment:
-    GF_EDITION: oss
+  - ./bin/grabpl build-backend --edition oss --build-id $DRONE_BUILD_NUMBER --variants linux-x64,linux-x64-musl,osx64,win64 --no-pull-enterprise
   depends_on:
   - initialize
   - lint-backend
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -233,5 +225,282 @@ services:
...
+  - test-frontend
+
+- name: build-windows-installer
+  image: grafana/wix-toolset-ci:v3
+  commands:
+  - echo ./scripts/build/ci-msi-build/ci-msi-build-oss.sh
+  depends_on:
+  - package
+
+- name: release-next-npm-packages
```
marefr, line 460: This is already done by CircleCI master build, correct? If so, need to be removed for now.

.drone.yml
```diff
@@ -233,5 +225,282 @@ services:
...
+  - npx lerna bootstrap
+  - echo //registry.npmjs.org/:_authToken=$${NPM_TOKEN} >> ~/.npmrc"
+  - echo ./scripts/circle-release-next-packages.sh
+  environment:
+    NPM_TOKEN:
+      from_secret: npm_token
+  depends_on:
+  - end-to-end-tests
+
+- name: publish-packages
```
marefr, line 472: This is already done by CircleCI master build, correct? If so, need to be removed for now.

marefr (changes requested):
I was going to approve, but seems like it will run publish steps that's already in CircleCI, see comments.

.drone.yml
aknuds1: @marefr No need to worry, the publishing is only simulated (see `echo` at line 465).

.drone.yml
aknuds1: @marefr No need to worry, the publishing is only simulated (see `echo` at line 475).

marefr (approved):
Oops :+1: LGTM

PR #26375, opened 2020-07-16
Prometheus: parse value as number when label is le

### LINKED ISSUE #22403: Explore Table sort numeric value

When you're trying to visualize an histogram by it's count per bucket and you try to sort bucket `le` this happens:

image

I think they are sorted by string value but not numeric value which would be awesome here.

**Why is this needed**:

I this would help understand better histogram metrics.

zoltanbedi:

Fixes #22403


ivanahuckova:
Nice work Zoltan!  LGTM! 🚀 I would maybe also include a test where `le = +Inf`.

aocenas:
There seems to be a small issue with the infinite values and table filtering. In the table you can filter in/out values:
Screenshot from 2020-07-17 15-39-09

But with infinity right now it produces wrong query 
Screenshot from 2020-07-17 15-36-47

There needs to be some way to decouple the value of the label with the text value inserted into the query so in the query it will still use `+Inf` string.

davkal (approved):
Nice work. The trade-off here is that we gain sortability on the le column, but it will no longer render as "+Inf", but as the infinity symbol. I'm happy with this though. 

I wonder if could do this across more columns and detect numerical content in the string (not as easy though, since `isNan` also let's strings like `01` and `1.` pass as numbers).

PR #26376, opened 2020-07-16
Chore: Enable goprintffuncname and nakedret linters

aknuds1:
Enable goprintffuncname and nakedret Go linters (already enabled for the plugin SDK), fix resulting issues. I don't think I had to make any changes to appease nakedret, but goprintffuncname requires appending an "f" to logging functions that take format strings. I think the latter requirement makes total sense, since it can be tricky to figure out right now if a logging function takes a format strings or key-value pairs. I'm open to other views on this though.


kylebrandt:
I'm confused about the point of infra/log/interface.go's Logger interface and how it, if it all, relates to these other methods.

aknuds1:
> I'm confused about the point of infra/log/interface.go's Logger interface and how it, if it all, relates to these other methods.

I have no idea either @kylebrandt. Carl and I talked about consolidating Grafana's logging at some point, also with an eye towards improving performance.

kylebrandt (approved):

PR #26404, opened 2020-07-17
Chore: Reduce settings Go code complexity

aknuds1:
As part of enabling gocyclo linter, make Go code to load settings less complex (by breaking huge functions into more manageable ones).

I've done basic manual testing and it works.

Probably easiest diffed locally with a GUI tool.


kylebrandt (approved):

ryantxu (approved):
I am all for it... and trust you that this is better ;)

pkg/setting/setting.go
```diff
@@ -661,73 +661,264 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 		return err
 	}
 	cfg.ProvisioningPath = makeAbsolute(provisioning, HomePath)
-	server := iniFile.Section("server")
-	AppUrl, AppSubUrl, err = parseAppUrlAndSubUrl(server)
+	if err := readServerSettings(iniFile, cfg); err != nil {
+		return err
```
marefr, line 665: Nit: Maybe return some context when returning errors from reading certain sections/settings by wrapping error in "failed to read <section/setting> settings" or similar?

marefr (approved):
LGTM. Added a nit comment

pkg/setting/setting.go
aknuds1: @marefr Isn't that out of scope for the PR, since all it does is refactor?

marefr: Sure

PR #26408, opened 2020-07-17
Drone: Fix StoryBook build

aknuds1:
Fix StoryBook build, by adding a separate publish step using correct Docker image on master.


marefr (approved):
LGTM

PR #26412, opened 2020-07-17
Bring back scripts evaluation in TextPanel

### LINKED ISSUE #26406: Custom JS tabs menu  not responding since move to 7.1.0 (from 6.6.2)
**What happened**:
Upgrading to 7.1 has disabled JavaScript query execution in custom text panel.

Simple "Tabs" menu to cycle between "Dashboard explanations" for my multi-visualisation dashboards no longer responding to the button presses.

The below is an example html object I would use. **I operate with disable_sanitize_html=True**

https://www.w3schools.com/howto/howto_js_tabs.asp
We essentially use a customised version of the above, with no change to the JS itself.

Worked fine on 6.6.2. Upgrade to 7.1.0 this morning and noted this change.

**Environment**:
- Grafana version: 7.1.0
- OS Grafana is installed on: Ubuntu 18.0.4
- User OS & Browser: Windows 10 Chrome,Internet Explorer and Edge
- Grafana plugins: Text Plugin
- Others:

dprokop:
Fixes https://github.com/grafana/grafana/issues/26406

When html sanitization is disabled we need to process html against any inline script tag occurrences and eval those scripts to have an effect. React's `dangerouslySetInnerHTML` uses `innerHTML` which does not evaluate scripts: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

@ryantxu I couldn't get the test for new logic to pass, any ideas? Seems like the parsed text is undefined in the test env, no idea why. Maybe some jsdom issue?

### DIFF
--- a/public/app/plugins/panel/text/TextPanel.test.tsx
+++ b/public/app/plugins/panel/text/TextPanel.test.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { TextPanel } from './TextPanel';
+import { dateMath, dateTime, LoadingState, FieldConfigSource } from '@grafana/data';
+import { TextOptions } from './types';
+import { updateConfig } from 'app/core/config';
+
+describe('TextPanel', () => {
+  describe('when html sanitization disabled', () => {
+    it('should evaluate provided scripts', done => {
+      jest.spyOn(window, 'alert').mockImplementation(() => {});
+
+      updateConfig({
+        disableSanitizeHtml: true,
+      });
+
+      expect(window.alert).not.toBeCalled();
+
+      createTextPanel({
+        series: [],
+        timeRange: {
+          from: dateMath.parse('now-6h') || dateTime(),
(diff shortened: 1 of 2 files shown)


dprokop:
Alternatively we could use https://github.com/christo-pr/dangerously-set-html-content

ryantxu:
ugg -- maybe we can take the opportunity to move script to an explicit script field?

dprokop:
> ugg -- maybe we can take the opportunity to move script to an explicit script field? 

Well, we could, but this will still be breaking change. And following this pattern we would have to provide style field too. I think with the Monaco editor having all the things in one editor is just fine.

ryantxu:
for sure the approach in https://github.com/christo-pr/dangerously-set-html-content is better :)

```
const slotHtml = document.createRange().createContextualFragment(html)
````

dprokop:
yes, i'm testing this lib as we speak

dprokop:
Closing in favour of https://github.com/grafana/grafana/pull/26413

PR #26426, opened 2020-07-17
Tracing: upstream fix for hovering on log lines

zoltanbedi:

See this pr for the issue https://github.com/jaegertracing/jaeger-ui/pull/599

### DIFF
--- a/packages/jaeger-ui-components/src/TraceTimelineViewer/SpanTreeOffset.tsx
+++ b/packages/jaeger-ui-components/src/TraceTimelineViewer/SpanTreeOffset.tsx
@@ -43,7 +43,7 @@ export const getStyles = createStyle((theme: Theme) => {
       /* The size of the indentGuide is based off of the iconWrapper */
       padding-right: calc(0.5rem + 12px);
       height: 100%;
-      border-left: 1px solid transparent;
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui-components/src/TraceTimelineViewer/SpanTreeOffset.tsx
```diff
@@ -53,12 +53,9 @@ export const getStyles = createStyle((theme: Theme) => {
     `,
     indentGuideActive: css`
       label: indentGuideActive;
-      padding-right: calc(0.5rem + 11px);
-      border-left: 0px;
+      border-left: 3px solid darkgrey;
```
s-h-a-d-o-w, line 56: Yup that looks identical to the upstream fix. But I can't help but wonder whether @everett980 is right about the fact that you only need to specify the color because `3px` and `solid` are already covered by `indentGuide`.

zoltanbedi: Yep, that's correct. Using border-color is enough.

s-h-a-d-o-w (approved):
LGTM now :)

PR #26427, opened 2020-07-17
Tracing: Add stack trace to span detail row

zoltanbedi:

This pr add a new row to jaeger tracing. To show stack traces.


aocenas:
Code looks good but can we change the formatting:

Screenshot from 2020-07-20 18-23-10

The new lines don't seem to be respected and also I think it would be better to have a textarea with internal horizontal scroll instead of wrapping the lines.

aocenas (approved):
Nice, looks good :+1:

PR #26429, opened 2020-07-17
Azure: convert azure units to grafana units

### LINKED ISSUE #26396: 7.1.0: Stat displays "percent" instead of % symbol with Azure Monitor datasource
**What happened**:
Upgraded Grafana from 7.0.3 to 7.1.0 and Stat visualization with Unit percent started displaying "100 percent" instead of "100%"
image

I expect no change and the Unit percent should display the % symbol

Create Stat visuzalization with numeric data and select Unit percent in the Field tab

**Environment**:
- Grafana version: 7.1.0
- Data source type & version:
- User OS & Browser:
- Grafana plugins:
- Others:

ryantxu:
Fixes #26396

In 7.1, we started returning the Azure unit along with the result.  But "Percent" != "percent" so the display is off

### DIFF
--- a/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
+++ b/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
@@ -293,11 +293,9 @@ func (e *AzureMonitorDatasource) parseResponse(queryRes *tsdb.QueryResult, amr A
 		dataField := frame.Fields[1]
 		dataField.Name = amr.Value[0].Name.LocalizedValue
 		dataField.Labels = labels
-		if amr.Value[0].Unit != "" {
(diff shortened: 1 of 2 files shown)


kylebrandt:
I think we should add a package with unit constants to the SDK.

The the conversion func with return gunit.Percent (or whatever package name), and whatever string that constant is can be changed in the future if need be.

Kind of like https://github.com/bosun-monitor/bosun/blob/8a7cd2d2f5830948135b0d076b5da2c35b992780/metadata/metadata.go#L44

ryantxu:
@kylebrandt sounds good, we should make it easy to accurately use "known" units.  Not sure we should update the sdk for 7.1.1 though

kylebrandt (approved):

PR #26440, opened 2020-07-18
Provisioning: Validate that datasource access field equals to direct or proxy

### LINKED ISSUE #25866: Data source provisioning access value error not caught by validator
**What happened**:
Using a data source provisioning yaml with a certain incorrect value for the "access" key will still add the data source. The data source however, is not usable.

Example to incorrect data source provisioning yaml:
```
    apiVersion: 1
    datasources:
        - access: proxy,
          editable: true
          name: prometheus
          isDefault: true
          orgId: 1
          type: prometheus
          url: localhost:9090
          version: 1
```

Note the error on the "access" key property: "proxy," (a comma at the end). Grafana will parse this data source and add it without any warnings or errors. When trying to create a dashboard using data from the data source a hard-to-parse UI error will occur: "Failed to fetch"

Receive an error of incorrect value on key "access".

- Create a data source provisioning file using the example given above, eg: /etc/grafana/provisioning/datasources/prom.yaml
- Start Grafana (note no error message of incorrect syntax)
- Try to add a panel in a new dashboard with any data from the data source (eg; PromQL: prometheus_build_info)
- Note the exclamation point in the corner of the panel, hovering over the exclamation point shows the error: "Failed to fetch"

**Environment**:
- Grafana version: 7.0.4
- User OS & Browser: Ubuntu 18.04
- Grafana plugins: None
- Others: N/A

nabokihms:
This PR adds validation of datasource configuration that `access` field should equal to proxy or direct. If it's not, we will change the value to 'proxy' (default value) and will warn about it into the console log. 

Fixes #25866

Nothing special. There was a difficult choice between error and warning. I chose the second because returning an error will break Grafana startup for users with invalid value in access filed (which makes this change a breaking change).

### DIFF
--- a/pkg/services/provisioning/datasources/config_reader_test.go
+++ b/pkg/services/provisioning/datasources/config_reader_test.go
@@ -22,6 +22,7 @@ var (
 	brokenYaml                      = "testdata/broken-yaml"
 	multipleOrgsWithDefault         = "testdata/multiple-org-default"
 	withoutDefaults                 = "testdata/appliedDefaults"
+	invalidAccess                   = "testdata/invalid-access"
 
 	fakeRepo *fakeRepository
 )
@@ -147,6 +148,12 @@ func TestDatasourceAsConfig(t *testing.T) {
 			reader := &configReader{}
 			_, err := reader.readConfig(brokenYaml)
 			So(err, ShouldNotBeNil)
+		})
+
+		Convey("invalid access should warn about invalid value and return 'proxy'", func() {
+			reader := &configReader{log: logger}
+			configs, _ := reader.readConfig(invalidAccess)
+			So(configs[0].Datasources[0].Access, ShouldEqual, models.DS_ACCESS_PROXY)
 		})
 
 		Convey("skip invalid directory", func() {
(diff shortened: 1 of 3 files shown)


pkg/services/provisioning/datasources/config_reader_test.go
```diff
@@ -149,6 +150,12 @@ func TestDatasourceAsConfig(t *testing.T) {
 			So(err, ShouldNotBeNil)
 		})
 
+		Convey("invalid access should warn about invalid value and return 'proxy'", func() {
+			reader := &configReader{log: logger}
+			configs, _ := reader.readConfig(invalidAccess)
```
aknuds1, line 155: Why do you ignore the error here, instead of verifying it's nil?

nabokihms: My bad. I will fix it.

papagian (approved):
LGTM; thank you for contributing to grafana and fixing this!

aknuds1 (approved):

aknuds1:

PR #26454, opened 2020-07-20
Grafana-UI: Pass timeZone to TimePickerButtonLabel in TimeRangeInput

Clarity-89:
Currently `TimePickerButtonLabel` doesn't receive `timeZone` prop from `TimeRangeInput` and as a result, the timezone is always displayed from default settings even if different time zone is passed to `TimeRangeInput`.


AgnesToulet (approved):
LGTM!

PR #26458, opened 2020-07-20
Units: Allow re-scaling nanoseconds up to days

### LINKED ISSUE #26428: Units: nanosecond unit to scale to hours
**What happened**:

Have a metric in nanoseconds, select unit: ```Time\ nanoseconds (ns)```. It will display as minutes even though the value is more than 1 hour.

image
image

I expect it to show as hours when it's more than 1 hour and even days when it's more than 1 day.

**Environment**:
- Grafana version: 7.1.0 Alpine Docker
- Data source type & version: InfuxDB 1.8.1

kaydelaney:
Previously nanoseconds could only be re-scaled up to minutes. This PR allows nanoseconds to be scaled up to days.

Closes #26428


torkelo (approved):
Looks good!

PR #26462, opened 2020-07-20
Drone: Remove netrc at beginning of steps

aknuds1:
In Drone, remove /root/.netrc at the beginning of every step.


aknuds1:
Closing since shouldn't be necessary after all.

Issue #18763, opened 2019-08-29
Time range variable $__from and $__to should report epoch time as seconds instead of milliseconds


jtorkkel:
Currently epoch time is shown as millisecond  resulting a mismatch when panel links are used to launch other tools which mandate time range as seconds or seconds.milliseconds format. (1566971076499 --> 1566971076.499 or 1566971076, 'Wed Aug 28 08:44:36 2019' )

My understanding is that "standard" use seconds and optional decimal milliseconds depending on representation.

Due mismatch custom redirecting converting proxy would be required to divide epoch values with 1000

Extract from documentation
Time range variables
Grafana has two built in time range variables in $__from and $__to. They are currently always interpolated as epoch milliseconds. These variables are only available in Grafana v6.0 and above.

Possible options:
- change to $__from and $__to from milliseconds to second.millisecond decimal format, would still represent milliseconds
- change $__from and $__to seconds and add $__from_ms and $__to_ms to provide millisecond granularity
- keep $__from and $__to as milliseconds and add seconds representation $__from_s and $__to_s

**Environment**:
- Grafana version: 6.3.2
- Data source type & version: Prometheus
- OS Grafana is installed on: Windows and Linux
- User OS & Browser: Chrome

retzkek:
I'd definitely like to see other time formats supported for the range variables, ISO8601 being another important one. Maybe [variable format options](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/) could be used for that? Otherwise a standard naming scheme like `$__from_{format}` sounds reasonable for me, e.g. `$__from_s`, `$__from_iso`, etc.

retzkek:
I think this can be closed now, what do you say @jtorkkel

jtorkkel:
yes, [26466](https://github.com/grafana/grafana/pull/26466) fix it

PR #26466, opened 2020-07-20
TemplateSrv: Formatting options for ${__from} and ${__to}, unix seconds epoch, ISO 8601/RFC 3339 

### LINKED ISSUE #15903: Support $__from and $__to in seconds
When linking to other tools it's good to support passing the current time range you are using in Grafana. 
Which is why we have `$__to` and `$__from`. But those are in milliseconds and some expect them to be in seconds. 

## Just add another variable
One solution for this would be to add `$__to_s` and `$__from_s`. The downside to this would be that we might end up adding more units down the road. Ex `$__to_date_us` Upside being that its a tiny change. 

## Templating interpolation with datesupport 
An other solution would be to add support for transforming this value using the [templating interpolation feature](http://docs.grafana.org/reference/templating/#pipe). ex `${__to:format:epoch:seconds}` `${__to:format:HHmmss}`
The downside for this solution is that it's far more work and in one of our most complex areas of the  codebases. It's also a matter of how far we would like to push such a feature. I think there are tons of good string manipulations we could add on top of this. 

We most likly have an issue for the second solution allready. The reason I wrote up this is because Im not sure if we should go with the first or the second solution to solve the original request :)

torkelo:
Closes #15903 

Introduces an extended templating syntax format that can have 1 or more arguments like this `${variable:<formatName>:<optionalArg1>:<optionalArg2>}` 

Adds date format and args
* `${__from:date}`  no args, defaults to iso/rfc (example 2020-07-13T20:19:09.254Z) 
* `${__from:date:iso}` same as above
* `${__from:date:ms}` unix millisecond epoch same as no date format `${__from}`
* `${__from:date:seconds}` unix seconds epoch
* `${__from:date:YYYY-MM}` custom date format

TODO
* [x] Get feedback on naming and syntax 
* [x] Update docs

In the upcoming PR:
* [x] Possibly refactor formats to use a new format registry using Registry class (could be useful for syntax IntelliSense in the future)

### DIFF
--- a/docs/sources/variables/global-variables.md
+++ b/docs/sources/variables/global-variables.md
@@ -77,7 +77,7 @@ The `$timeFilter` variable returns the currently selected time range as an expre
 
 This is used in several places, including:
 
-- The WHERE clause for the InfluxDB data source. Grafana adds it automatically to InfluxDB queries when in Query Editor Mode. It has to be added manually in Text Editor Mode: `WHERE $timeFilter`.
+- The WHERE clause for the InfluxDB data source. Grafana adds it automatically to InfluxDB queries when in Query Editor mode. You can add it manually in Text Editor mode: `WHERE $timeFilter`.
 - Log Analytics queries in the Azure Monitor data source.
 - SQL queries in MySQL, Postgres, and MSSQL
 - The `$__timeFilter` variable is used in the MySQL data source.
--- a/docs/sources/variables/global-variables.md
+++ b/docs/sources/variables/global-variables.md
@@ -13,20 +13,23 @@ weight = 200
 Grafana has global built-in variables that can be used in expressions in the query editor. This topic lists them in alphabetical order and defines them.
 
 ## $__dashboard
-> Only available in Grafana v6.7+
 
-This variable is the UID of the current dashboard.
-`${__dashboard.name}` is the name of the current dashboard.
+- `${__dashboard.name}` is the name of the current dashboard
+- `${__dashboard.uid}` is the UID of the current dashboard (used in url)
 
 ## $__from and $__to
 
-> Only available in Grafana v6.0+
+Grafana has two built in time range variables: `$__from` and `$__to`. They are currently always interpolated as epoch milliseconds by default but you can control formatting using a special formating syntax.
 
-Grafana has two built in time range variables: `$__from` and `$__to`. They are currently always interpolated as epoch milliseconds. 
+> This special formatting syntax is only available in Grafan a 7.1.2+
+
+| Syntax              | Example result | Description |
+| ------------------- | ------------- | ----------- |
+| `${__from:date}` | 2020-07-13T20:19:09.254Z | Defaults to ISO 8601/RFC 3339 |
 
 ## $__interval
 
-The `$__interval` variable can be used as a parameter to group by time (for InfluxDB, MySQL, Postgres, MSSQL), Date histogram interval (for Elasticsearch) or as a *summarize* function parameter (for Graphite).
+The `$__interval` variable can be used as a parameter to group by time (for InfluxDB, MySQL, Postgres, MSSQL), Date histogram interval (for Elasticsearch) or as a _summarize_ function parameter (for Graphite).
 
 Grafana automatically calculates an interval that can be used to group by time in queries. When there are more data points than can be shown on a graph then queries can be made more efficient by grouping by a larger interval. It is more efficient to group by 1 day than by 10s when looking at 3 months of data and the graph will look the same and the query will be faster. The `$__interval` is calculated using the time range and the width of the graph (the number of pixels).
 
@@ -48,8 +51,6 @@ This variable is only available in the Singlestat panel and can be used in the p
 
 ## $__org
 
-> Only available in Grafana v6.7+
-
 This variable is the ID of the current organization.
 `${__org.name}` is the name of the current organization.
 
@@ -62,8 +63,6 @@ This variable is the ID of the current organization.
 
 ## $__range
 
-> Only available in Grafana v5.3+
-
 Currently only supported for Prometheus data sources. This variable represents the range for the current dashboard. It is calculated by `to - from`. It has a millisecond and a second representation called `$__range_ms` and `$__range_s`.
 
 ## $timeFilter or $__timeFilter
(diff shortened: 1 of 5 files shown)


ryantxu:
a new templating syntax!  :exploding_head: 

When I actually look at the code, it looks great -- totally reasonable.  I just get scared by adding more powerful hidden complex features to templateSrv.

+1 to refactor the formats into a Registry that we can more clearly promote with IntelliSense etc.

public/app/features/templating/template_srv.ts
```diff
@@ -154,6 +154,19 @@ export class TemplateSrv implements BaseTemplateSrv {
       return format(value, variable, this.formatValue);
     }
 
+    if (!format) {
+      format = 'glob';
```
kaydelaney, line 158: Could a default parameter be used instead? Like
```typescript
formatValue(value: any, format: any = 'glob', variable: any) {
```

jtorkkel:
There is similar need for "to" like ${__to:date:ms}, see https://github.com/grafana/grafana/issues/18763

public/app/features/templating/template_srv.ts
torkelo: don't think you can use a default value for the non-optional argument (difference between ? and being undefined)

torkelo:
of course this also handled ${__to}

torkelo:
Ok, ready for merge. Updated docs. 

The refactor to registry will come in new PR as we need to cherry-pick this to 7.1.2 so want to keep this PR as small as possible, so moving the refactoring to a new PR.

docs/sources/variables/global-variables.md
```diff
@@ -10,23 +10,32 @@ weight = 200
 
 # Global variables
 
-Grafana has global built-in variables that can be used in expressions in the query editor. This topic lists them in alphabetical order and defines them.
+Grafana has global built-in variables that can be used in expressions in the query editor. This topic lists them in alphabetical order and defines them. These variables are useful in both queries, dashboard links, panel links & data links.
```
oddlittlebird, line 13:
```suggestion
Grafana has global built-in variables that can be used in expressions in the query editor. This topic lists them in alphabetical order and defines them. These variables are useful in queries, dashboard links, panel links, and data links.
```

oddlittlebird: "Both" means you only have two items in your list. Also, please do not use ampersands as an abbreviation for "and" in prose.

docs/sources/variables/global-variables.md
```diff
@@ -10,23 +10,32 @@ weight = 200
...
 
-This variable is the UID of the current dashboard.
-`${__dashboard.name}` is the name of the current dashboard.
+- `${__dashboard.name}` is the name of the current dashboard
+- `${__dashboard.uid}` is the UID of the current dashboard (used in url)
 
 ## $__from and $__to
 
-> Only available in Grafana v6.0+
+Grafana has two built in time range variables: `$__from` and `$__to`. They are currently always interpolated as epoch milliseconds by default but you can control date formatting.
```
oddlittlebird, line 22:
```suggestion
Grafana has two built-in time range variables: `$__from` and `$__to`. By default, they areinterpolated as epoch milliseconds, but you can change date formatting.
```

docs/sources/variables/global-variables.md
```diff
@@ -10,23 +10,32 @@ weight = 200
...
+
+| Syntax                   | Example result           | Description |
+| ------------------------ | ------------------------ | ----------- |
+| `${__from}`              | 1594671549254            | Unix millisecond epoch |
+| `${__from:date}`         | 2020-07-13T20:19:09.254Z | No args, defaults to ISO 8601/RFC 3339 | 
+| `${__from:date:iso}`     | 2020-07-13T20:19:09.254Z | ISO 8601/RFC 3339 |
+| `${__from:date:seconds}` | 1594671549               | Unix seconds epoch |
+| `${__from:date:YYYY-MM}` | 2020-07                  | Any custom [date format](https://momentjs.com/docs/#/displaying/) |
+
+The above syntax of course works with `${__to}` as well.
```
oddlittlebird, line 34:
```suggestion
The above syntax oworks with `${__to}` as well.  
```

docs/sources/variables/global-variables.md
```diff
@@ -10,23 +10,32 @@ weight = 200
...
+| `${__from:date:iso}`     | 2020-07-13T20:19:09.254Z | ISO 8601/RFC 3339 |
+| `${__from:date:seconds}` | 1594671549               | Unix seconds epoch |
+| `${__from:date:YYYY-MM}` | 2020-07                  | Any custom [date format](https://momentjs.com/docs/#/displaying/) |
+
+The above syntax of course works with `${__to}` as well.  
 
 ## $__interval
 
-The `$__interval` variable can be used as a parameter to group by time (for InfluxDB, MySQL, Postgres, MSSQL), Date histogram interval (for Elasticsearch) or as a *summarize* function parameter (for Graphite).
+The `$__interval` variable can be used as a parameter to group by time (for InfluxDB, MySQL, Postgres, MSSQL), Date histogram interval (for Elasticsearch) or as a _summarize_ function parameter (for Graphite).
```
oddlittlebird, line 38:
```suggestion
You can use the `$__interval` variable as a parameter to group by time (for InfluxDB, MySQL, Postgres, MSSQL), Date histogram interval (for Elasticsearch), or as a _summarize_ function parameter (for Graphite).
```

docs/sources/variables/global-variables.md
```diff
@@ -62,16 +69,15 @@ This variable is the ID of the current organization.
...
 The `$timeFilter` variable returns the currently selected time range as an expression. For example, the time range interval `Last 7 days` expression is `time > now() - 7d`.
 
 This is used in several places, including:
-* The WHERE clause for the InfluxDB data source. Grafana adds it automatically to InfluxDB queries when in Query Editor Mode. It has to be added manually in Text Editor Mode: `WHERE $timeFilter`.
-* Log Analytics queries in the Azure Monitor data source.
-* SQL queries in MySQL, Postgres, and MSSQL
-* The `$__timeFilter` variable is used in the MySQL data source.
\ No newline at end of file
+
+- The WHERE clause for the InfluxDB data source. Grafana adds it automatically to InfluxDB queries when in Query Editor Mode. It has to be added manually in Text Editor Mode: `WHERE $timeFilter`.
```
oddlittlebird, line 80:
```suggestion
- The WHERE clause for the InfluxDB data source. Grafana adds it automatically to InfluxDB queries when in Query Editor mode. You can add it manually in Text Editor mode: `WHERE $timeFilter`.
```

oddlittlebird (approved):
Good to go aside from some grammar suggestions in the docs.

ehenara:
I've upgraded to 7.1.5; but still not able to use the Formatting option

ex: Panel title "${__from:date}" result is seconds not date format

image

PR #26486, opened 2020-07-21
Azure: Insights metrics - do not sort if no data

### LINKED ISSUE #26473: 7.1.0 Error "field sorting for a wide time series frame called on a series that is not a wide frame" when no data to display
**What happened**:
An error happen when there is no any data in the time period selected.
image

The visuals can normally display "0" when no data, instead of a error.

1. Create a source of Azure Monitor, and config to connect to Application Insights
2. Add a "Stat" and set the query as below.
image
3. Change the time period to make sure there is no data to display

**Environment**:
- Grafana version: v7.1.0
- Data source type & version: Azure Monitor (0.3.0) - Application Insights
- User OS & Browser: Win10 & Microsoft Edge
- Grafana plugins: Azure Monitor (0.3.0)
- Others:

kylebrandt:

If you zoom into a narrow time window with no data, it will not be a "wide series" due to lack of information in the response. Currently this returns an error when sorting.

So we return before sorting if this is the case, an now the UI shows "No data" without any error.

Fixes #26473


aknuds1 (approved):

papagian (approved):
LGTM

PR #24009, opened 2020-04-28
Dashboard: Refactor container css to use flex for height instead of calc 

torkelo:
This refactors and simplifies css a bit for dashboard container to not have to do calc  to determine height

PR #26495, opened 2020-07-21
Dashboard: Fix empty panels after scrolling on Safari/iOS

### LINKED ISSUE #26442: Dashbard: Lazy loading panels does not work in Safari / iOS
**What happened**:
Not sure if this has been reported before: I upgraded my grafana instance from 7.0.6 to 7.1.0 and lost 3 or my 8 panels from the dashboard. I can see that they are actually still there but are just completely blacked out. Reverted back to 7.0.6 and they become visisble again.

No disappearance in panel display.

Upgrade to 7.1.0 on a large graphite database dashboard with more than 5 panels

**Environment**:
- Grafana version: 7.1.0
- Data source type & version: graphite
- OS Grafana is installed on: ubuntu 18.04
- User OS & Browser: macOS Safari, iOS Safari
- Grafana plugins: N/A

torkelo:
Fixes #26442 

This one was tricky as we have made no change to scrollbar / lazy loading, nor any lib upgrade. 

This was caused by a css change for class dashboard-scroll (change to flex box, so we do not need absolute css height calc math). https://github.com/grafana/grafana/pull/24009 

Somehow this change broke firefox, scrolling still worked but we got no scroll events. 

Fixed it by subtle css changes.

### DIFF
--- a/public/sass/components/_navbar.scss
+++ b/public/sass/components/_navbar.scss
@@ -5,6 +5,7 @@
   padding: 0 16px 0 60px;
   display: flex;
   flex-grow: 0;
+  flex-shrink: 0;
   border-bottom: 1px solid transparent;
   transition-duration: 350ms;
   transition-timing-function: ease-in-out;
--- a/public/sass/pages/_dashboard.scss
+++ b/public/sass/pages/_dashboard.scss
@@ -13,6 +13,7 @@
   width: 100%;
   flex-grow: 1;
   min-height: 0;
+  display: flex;
 }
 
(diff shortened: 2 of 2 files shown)


Clarity-89 (approved):
Works for me on the latest safari.

ivanahuckova (approved):
Works for me on latest safari as well!

Paraphraser:
Confirming 7.1.1 fixes issue reported in #26572, apparently a duplicate of #26442. Thanks.

PR #26498, opened 2020-07-21
API: Recognize dashboard errors when saving a folder

### LINKED ISSUE #26478: Error message on folder creation error
**What happened**:
When trying to create a folder with a name already used by a dashboard, we get a 500 error, with {"message":"Folder API error"}. 

Getting a 400 error with an explanatory message

1. Create a dashboard named foo by hand (in the “General” folder)
2. POST to /api/folders { “title”:”foo”} will fail with a 500
3. Delete or rename the dashboard
4. POST /api/folders will succeed

**Environment**:
- Grafana version: 7.0.3
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

aknuds1:
Recognize dashboard errors when saving a folder via the API, for example that a dashboard by the same name already exists. Return a suitable error code instead of 500.

Fixes #26478.

### DIFF
--- a/pkg/api/folder.go
+++ b/pkg/api/folder.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/grafana/grafana/pkg/api/dtos"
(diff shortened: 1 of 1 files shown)


aknuds1:
Closing since GitHub created duplicate PRs for me :/

PR #26501, opened 2020-07-21
Bump caniuse lite

simPod:
When building frontend part, there were warnings about outdated caniuse deps. This fixes it.

>> Browserslist: caniuse-lite is outdated. Please run the following command: `yarn upgrade`


torkelo (approved):
Looks good, thanks! ❤️

PR #26504, opened 2020-07-21
Docs: Add data source, interval, and ad hoc variable tasks

oddlittlebird:
Adds task topics for data source, interval, and ad hoc variables. Also added links to the new topics.

Fixes #

**Special notes for your reviewer**: If you see places to extend or improve the content, please do so. I am not an expert on variables.


jessabe (approved):
LGTM!

PR #26507, opened 2020-07-21
Email: change the year in templates

adeptg:
It's a weird year, and I'd like to skip it too :) But anyway


ivanahuckova (approved):
😂 LGTM!

PR #26510, opened 2020-07-22
Chore: Upgrade build pipeline tool

aknuds1:
Upgrade the build pipeline tool, to run back-end tests with a number of retries since these have been flaky under Drone.


jessabe (approved):
LGTM

PR #26514, opened 2020-07-22
HTTPServer: add possibility to use additional middlewares

AgnesToulet:
This PR allows to easily add new middlewares on the HTTP server macaron instance.

**Special notes for your reviewer**: Will be used in Grafana Enterprise for the auditing feature.


pkg/api/http_server.go
```diff
@@ -82,6 +83,10 @@ func (hs *HTTPServer) Init() error {
 	return nil
 }
 
+func (hs *HTTPServer) AddMiddleware(middleware macaron.Handler) {
```
xlson, line 86: My only concern here is if this is enough for what we need to do. Is it good enough to always run the auditing middleware last? What if a previous middleware shortcircuits the chain (I'm not completely sure but I presume we can do this for auth for example).

The alternative would be to support some kinda of priority system and make sure to add all the middlewares to the list in order using the priorities.

xlson (approved):

pkg/api/http_server.go
AgnesToulet: The previous middlewares are middlewares that run for all routes so it generally doesn't shortcircuit the chain (except for the `OrgRedirect` one but I don't think it's a problem in our case). They often add things to the context (like the authentication middlewares) and we need these extra information to know which user is behind the request. The middleware that checks for user authorization runs later.

PR #26515, opened 2020-07-22
CloudWatch: Fix passing of namespace for getting custom metrics

aknuds1:
Fix passing of namespace when getting CloudWatch custom metrics. AFAICT, the namespace wasn't correctly passed on before.

Also make sure a couple of errors are wrapped, so we don't lose original context.


jessabe (approved):
LGTM

papagian (approved):
LGTM

PR #26531, opened 2020-07-22
Fields: __field.name as field name and __field.displayName as displayName

ryantxu:
Template variables currently set `__field.name` as the display name, not the field name.  This makes it impossible to configure the display name using the field name :exploding_head: :turtle: :turtle: :turtle: 

This PR updates the syntax so that `__field.name` is simply the name and `__field.displayName` is the full rendered name

This uses a proxy object to define the behavior when values are requested, rather than when we initialize the scopedVars.  This makes it possible to support:

`__field.labels` = the string representation of the labels
`__field.labels.__values` = the string representation of the label values

With these changes, it is possible to configure the azure responses more reasonably


packages/grafana-data/src/field/fieldOverrides.test.ts
```diff
@@ -127,11 +127,7 @@ describe('applyFieldOverrides', () => {
         Object {
           "__field": Object {
             "text": "Field",
-            "value": Object {
-              "formattedLabels": "",
-              "labels": undefined,
-              "name": "A message",
-            },
+            "value": Object {},
```
ryantxu, line 130: the real value is the proxy object

kylebrandt:
@ryantxu seems to be missing `${__field.labels.__values}` based on my manual testing and looking at the code.

ryantxu:
dooh missing git push!

kylebrandt:
It seems if you have an instance where there is a label with a key, but its value is an empty string, then field.name is displayed when it should just be empty:

image

torkelo:
This looks good

need to update the docs as well: https://github.com/grafana/grafana/blob/master/docs/sources/panels/field-configuration-options.md#display-name 

 Worried about potentially breaking many dashboards, wonder if we should include this in a patch release or wait for 7.2 but then we will likely break more dashboards :( 

7.1.1 is going out tomorrow, need to make sure this bug free :)

torkelo:
ah, no I see 7.1.2 milestone, that seems more reasonable

torkelo:
What is `${__field.labels.__values}` ?? that looks strange (with the double __ prefix the middle)

packages/grafana-data/src/field/templateProxies.ts
```diff
@@ -26,6 +26,9 @@ export function getTemplateProxyForField(field: Field, frame?: DataFrame, frames
           }
           return {
             ...field.labels,
+            __values: Object.values(field.labels)
+              .sort()
+              .join(', '),
```
torkelo, line 31: what is the use case for this? vs toString?

kylebrandt:
> What is `${__field.labels.__values}` ?? that looks strange (with the double __ prefix the middle)

This displays only the value component of each label. Mainly to be able to reproduce the default display behavior (when there is only one label, only the value is displayed).

Since one can use labels.X to access arbitrary label keys, the idea is `__values` is less likely to collide with an actual label key would (like values).

torkelo:
Can we make the root field.labels object toString default to returning the same as __values ?

kylebrandt:
> Can we make the root field.labels object toString default to returning the same as __values ?

I don't think that works out, because since labels are "k=v", one wants to be able the full label set, so it makes sense to me for labels to always do that. labels.__values is probably more often useful in the in case where there is just one label.

Started to draft this up ( https://gist.github.com/kylebrandt/f19d258e984c120506387cb2fbc2666c ) since we might be going in circles here. Can possibly work on it more Monday - also maybe better as a gdoc.

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
 
 Lets you set the display title of all fields. You can use [variables]({{< relref "../variables/templates-and-variables.md" >}}) in the field title.
 
-When multiple stats are shown, this field controls the title in each stat. By default this is the series name and field name. You can use expressions like ${__series.name} or ${**field.name} to use only series name or field name in title or \${**cell_2} to refer to other fields (2 being field/column with index 2).
+When multiple stats, fields, or series are shown, this field controls the title in each stat. You can use expressions like `${__field.name}` to use only the series name or the field name in title.
+
+Given a field with a name of Temp, and labels of {"Loc"="PBI", "Sensor"="3"}
+
+| Expression Syntax            | Example                | Renders to                        | Explanation |
```
oddlittlebird, line 146:
```suggestion
| Expression syntax            | Example                | Renders to                        | Explanation |
```

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
+| ---------------------------- | ---------------------- | --------------------------------- | ----------- |
+| `${__field.displayName}`     | Same as Syntax         | `Temp {Loc="PBI", Sensor="3"}`    | Displays the field's name, and labels in `{}` if they are present. If there is only one label key in the response, for the label portion, only the value of the label will be displayed only without the enclosing braces |
```
oddlittlebird, line 148:
```suggestion
| `${__field.displayName}`     | Same as syntax         | `Temp {Loc="PBI", Sensor="3"}`    | Displays the field name, and labels in `{}` if they are present. If there is only one label key in the response, then for the label portion, Grafana displays the value of the label without the enclosing braces. |
```

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
+| `${__field.name}`            | Same as Syntax          | `Temp`                           | Displays the name of the field (without labels) |
```
oddlittlebird, line 149:
```suggestion
| `${__field.name}`            | Same as syntax          | `Temp`                           | Displays the name of the field (without labels). |
```

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
+| `${__field.labels}`          | Same as Syntax          | `Loc="PBI", Sensor="3"`          | Displays the labels without the name |
```
oddlittlebird, line 150:
```suggestion
| `${__field.labels}`          | Same as syntax          | `Loc="PBI", Sensor="3"`          | Displays the labels without the name. |
```

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
+| `${__field.labels.X}`        | `${__field.labels.Loc}` | `PBI`                            | Displays the value of the specified label key |
```
oddlittlebird, line 151:
```suggestion
| `${__field.labels.X}`        | `${__field.labels.Loc}` | `PBI`                            | Displays the value of the specified label key. |
```

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
+| `${__field.labels.__values}` | Same as Syntax          | `PBI, 3`                         | Displays the values of the labels separated by a comma (without label keys) |
```
oddlittlebird, line 152:
```suggestion
| `${__field.labels.__values}` | Same as Syntax          | `PBI, 3`                         | Displays the values of the labels separated by a comma (without label keys). |
```

oddlittlebird (changes requested):

docs/sources/panels/field-configuration-options.md
```diff
@@ -139,7 +139,19 @@ For more information and instructions, refer to [Data links]({{< relref "../link
...
 
-When multiple stats are shown, this field controls the title in each stat. By default this is the series name and field name. You can use expressions like ${__series.name} or ${**field.name} to use only series name or field name in title or \${**cell_2} to refer to other fields (2 being field/column with index 2).
+When multiple stats, fields, or series are shown, this field controls the title in each stat. You can use expressions like `${__field.name}` to use only the series name or the field name in title.
+
+Given a field with a name of Temp, and labels of {"Loc"="PBI", "Sensor"="3"}
+
+| Expression syntax            | Example                | Renders to                        | Explanation |
+| ---------------------------- | ---------------------- | --------------------------------- | ----------- |
+| `${__field.displayName}`     | Same as syntax         | `Temp {Loc="PBI", Sensor="3"}`    | Displays the field name, and labels in `{}` if they are present. If there is only one label key in the response, then for the label portion, Grafana displays the value of the label without the enclosing braces. |
```
ryantxu, line 148: note:  this should eventually link to whatever is settled on in https://github.com/grafana/grafana/pull/26656

oddlittlebird (approved):

oddlittlebird:
I approved the docs. I recommend holding on merging until someone else approves the code.

ryantxu:
Discussed the [case above](https://github.com/grafana/grafana/pull/26531#issuecomment-662617026) with @kylebrandt , for now lets solve that with documentatoin.

This is only a small part of the coherent naming strategy, but I think gives good clear progress towards explicitly setting reasonable values when the default behavior is undesirable

ryantxu:
@kylebrandt / @torkelo -- i think this is ready to merge and should be included in 7.1.2

for back compatibility -- i don't have a great answer other than the longer we live with an incoherent model the harder it is to change.  I don't have a great sense of how widely used explicitly setting name vs name+labels is yet, but these changes are required to support reasonable behavior when the [automatic](https://github.com/grafana/grafana/pull/26656) one does not work

packages/grafana-data/src/field/templateProxies.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * This object is created often, and only used when tmplates exist.  Using a proxy lets us delay
+ * calculations of the more complex structures (label names) until they are actually used
+ */
+export function getTemplateProxyForField(field: Field, frame?: DataFrame, frames?: DataFrame[]): any {
+  return new Proxy(
+    {} as any, // This object shows up in test snapshots
+    {
+      get: (obj: Field, key: string, reciever: any) => {
```
dprokop, line 13: @ryantxu I'm not super familiar with Proxy objects, but shouldn't the target here be the `field` passed to the `getTemplateProxyForField` function call? 🤯 Then the handler would use its first argument(`obj`) to reference the proxied object.

PR #26537, opened 2020-07-22
Live: include a streaming event manager

ryantxu:
This PR adds the https://github.com/centrifugal/centrifuge library to manage websocket streams and channels.  In this initial implementation, there is no authentication and any channel can broadcast to everyone.  This should move to having the channel behavior more strictly controlled by a plugin.  

The requires the feature flag "live"
```
[feature_toggles]
enable = live
```

With that enabled, the backend will send random streams to the channels:
* [random-2s-stream](https://github.com/grafana/grafana/pull/26537/files#diff-06cc78b0d87adc91d3c4b7c27db1b05dR88) -- updates every 2s
* [random-flakey-stream](https://github.com/grafana/grafana/pull/26537/files#diff-06cc78b0d87adc91d3c4b7c27db1b05dR89) -- updates intermitentanly

From the [/admin/live](http://localhost:3000/admin/live) page, you can see the latest message on a channel:
image

If you pick the 'chat' channel, and connect from multiple browsers, you can use the "publish" button to broadcast to everyone:
image


torkelo:
Looks promising!

go.mod
```diff
@@ -55,9 +55,9 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.3.0
-	github.com/prometheus/client_model v0.1.0
-	github.com/prometheus/common v0.7.0
+	github.com/prometheus/client_golang v1.6.0
+	github.com/prometheus/client_model v0.2.0
+	github.com/prometheus/common v0.9.1
```
ryantxu, line 60: importing centrifuge updates these libraries... is this a problem?  @bergquist @marefr

public/app/features/live/live.ts
```diff
@@ -0,0 +1,113 @@
...
+import { KeyValue } from '@grafana/data';
+
+interface Channel<T = any> {
+  subject: Subject<T>;
+  subscription?: Centrifuge.Subscription;
+}
+
+class CentrifugeSrv implements GrafanaLiveSrv {
+  centrifuge: Centrifuge;
+  channels: KeyValue<Channel> = {};
```
ryantxu, line 14: Currently this is a *super* naive approach, and should be improved.  In particular, this will hold on to subscriptions after everyone has unsubscribed

ryantxu:
@torkelo @aknuds1 -- I would love some feedback on this.  I hope it is ready to merge soon.  This is enough in the core to really exercise how we send/receive messages in plugins.

public/app/features/live/live.ts
ryantxu: In this approach plugins do not have direct access to `centrifuge` and this server intercepts all messages and manages rxjs subscriptions.  Alternatively, we could expose centrifuge and give plugins full control/responsibility.

This initial approach feels like the least risky, but we will could always expose centrifuge if it becomes too difficult to wrap everything

packages/grafana-runtime/src/services/live.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Configure a channel with the given setup
+   */
+  initChannel<T>(channel: string, handler: ChannelHandler<T>): void;
+
+  /**
+   * Subscribe to activity on a given channel
+   */
+  subscribe<T>(channel: string, observer: PartialObserver<T>): Unsubscribable;
```
torkelo, line 45: Is there a reason for passing in an observer here? instead of returning an Observable? returning an Observable (I think) will make this easier to use as you can use it together with rxjs operators etc

public/app/features/live/live.ts
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  //----------------------------------------------------------
+  // Exported functions
+  //----------------------------------------------------------
+
+  /**
+   * Is the server currently connected
+   */
+  isConnected = (): boolean => {
```
torkelo, line 59: would make all functions here live on the prototype instead (ie not define them using equal signs & arrow function), a bit allergic to this function definition style, would only use it as an exception for functions passed to other components

public/app/features/live/live.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  };
+
+  subscribe = <T>(path: string, observer: PartialObserver<T>): Unsubscribable => {
+    let c = this.channels[path];
+    if (!c) {
+      this.initChannel(path, standardChannelHandler);
+      c = this.channels[path];
+    }
+    return c!.subject.subscribe(observer);
```
torkelo, line 94: I think this should not be named subscribe, but getChannel and return an Observable (like getData in PanelQueryRunner), you could then just return the subject (it is an Observable)

torkelo (commented):
Looks good, main improvement idea I have is subscribe => getChannel and have it return an Observable

packages/grafana-runtime/src/services/live.ts
ryantxu: updated to return Observable -- yes this makes more sense :+1:

public/app/features/live/live.ts
ryantxu: "a bit allergic"  :rofl:   HA -- yes, the functions as properties voodo is awkward.  

I updated the things that feel like methods to be real methods, but kept the functions that get passed as callbacks as properties -- IIUC, the property syntax helps with making sure `this` refers to the right scope without needing to bind

public/app/features/live/live.ts
ryantxu: done

pkg/setting/setting.go
```diff
@@ -1199,3 +1199,12 @@ func IsExpressionsEnabled() bool {
...
 	return v
 }
+
+// IsLiveEnabled returns if grafana live should be enabled
+func IsLiveEnabled() bool {
+	v, ok := FeatureToggles["live"]
+	if !ok {
+		return false
+	}
+	return v
```
torkelo, line 1209: can you move this to Cfg struct instead and pass this to the Live manager? trying to get rid of global settings variables, somehow FeatureToggles was added to globals 
https://github.com/grafana/grafana/pull/26613

torkelo (approved):
Exiting to make progress on this!

pkg/services/live/channels.go
```diff
@@ -0,0 +1,49 @@
+package live
+
+import (
+	"encoding/json"
+	"math/rand"
+	"time"
+)
+
+// ChannelInfo holds metadata about each channel and is retruned on connection.
+// Eventually each plugin should control exactly what is in this structure
```
aknuds1, line 10:
```suggestion
// ChannelInfo holds metadata about each channel and is returned on connection.
// Eventually each plugin should control exactly what is in this structure.
```

pkg/services/live/channels.go
```diff
@@ -0,0 +1,49 @@
...
+type ChannelInfo struct {
+	Description string
+}
+
+type randomWalkeMessage struct {
```
aknuds1, line 15: I guess it should be "randomWalkMessage"?
```suggestion
type randomWalkMessage struct {
```

pkg/services/live/channels.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// RunRandomCSV just for an example
+func RunRandomCSV(broker *GrafanaLive, channel string, speedMillis int, dropPercent float64) {
+	spread := 50.0
+
+	walker := rand.Float64() * 100
+	ticker := time.NewTicker(time.Duration(speedMillis) * time.Millisecond)
+
+	line := randomWalkeMessage{}
```
aknuds1, line 29:
```suggestion
	line := randomWalkMessage{}
```

pkg/services/live/channels.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		delta := rand.Float64() - 0.5
+		walker += delta
+
+		line.Time = t.UnixNano() / int64(time.Millisecond)
+		line.Value = walker
+		line.Min = walker - ((rand.Float64() * spread) + 0.01)
+		line.Max = walker + ((rand.Float64() * spread) + 0.01)
+
+		bytes, _ := json.Marshal(&line)
```
aknuds1, line 43: Should we panic or `continue` on error here, instead of failing in some unpredictable way (in case an error _is_ returned)?

pkg/services/live/live.go
```diff
@@ -0,0 +1,187 @@
...
+	loggerCF = log.New("live.centrifuge")
+)
+
+// GrafanaLive pretends to be the server
+type GrafanaLive struct {
+	node    *centrifuge.Node
+	Handler interface{} // handler func
+}
+
+// InitalizeBroker the broker and start listening for requests
```
aknuds1, line 24:
```suggestion
// InitalizeBroker initializes the broker and starts listening for requests.
```

pkg/services/live/live.go
```diff
@@ -0,0 +1,187 @@
...
+	// Set SubscribeHandler to react on every channel subscription attempt
+	// initiated by client. Here you can theoretically return an error or
+	// disconnect client from server if needed. But now we just accept
+	// all subscriptions to all channels. In real life you may use a more
+	// complex permission check here.
+	node.OnSubscribe(func(c *centrifuge.Client, e centrifuge.SubscribeEvent) (centrifuge.SubscribeReply, error) {
+		info := &ChannelInfo{
+			Description: fmt.Sprintf("channel: %s", e.Channel),
+		}
+		bytes, _ := json.Marshal(&info)
```
aknuds1, line 68:
```suggestion
		bytes, err := json.Marshal(&info)
		if err != nil {
			return centrifuge.SubscribeReply{}, err
		}
```

pkg/services/live/live.go
```diff
@@ -0,0 +1,187 @@
...
+		logger.Debug("client subscribes on channel", "channel", e.Channel, "info", string(bytes))
+
+		return centrifuge.SubscribeReply{
+			ExpireAt:    0, // does not expire
+			ChannelInfo: bytes,
+		}, nil
+	})
+
+	node.OnUnsubscribe(func(c *centrifuge.Client, e centrifuge.UnsubscribeEvent) {
+		s, _ := node.PresenceStats(e.Channel)
```
aknuds1, line 78: The error should be handled here.

pkg/services/live/channels.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/json"
+	"math/rand"
+	"time"
+)
+
+// ChannelInfo holds metadata about each channel and is retruned on connection.
+// Eventually each plugin should control exactly what is in this structure
+type ChannelInfo struct {
```
aknuds1, line 11: Can we unexport ChannelInfo here, since it's not used outside of the package? In general, the less exported the better.

pkg/services/live/live.go
```diff
@@ -0,0 +1,187 @@
...
+	Handler interface{} // handler func
+}
+
+// InitalizeBroker the broker and start listening for requests
+func InitalizeBroker() (*GrafanaLive, error) {
+	// We use default config here as starting point. Default config contains
+	// reasonable values for available options.
+	cfg := centrifuge.DefaultConfig
+
+	// cfg.LogLevel = centrifuge.LogLevelDebug
```
aknuds1, line 30: Should we keep this commented out code?

aknuds1 (commented):
Looks generally pretty good, but left some comments.

pkg/services/live/live.go
ryantxu: Keeping it as debug is too much when channels are publishing fast, but it is useful to have at debug to understand which events get hit in which order

pkg/services/live/live.go
```diff
@@ -0,0 +1,192 @@
...
+		return centrifuge.SubscribeReply{
+			ExpireAt:    0, // does not expire
+			ChannelInfo: bytes,
+		}, nil
+	})
+
+	node.OnUnsubscribe(func(c *centrifuge.Client, e centrifuge.UnsubscribeEvent) {
+		s, err := node.PresenceStats(e.Channel)
+		if err != nil {
+			logger.Warn("unable to get presence stats", "channel", e.Channel, "error", err)
```
aknuds1, line 83: You should return here, since `s` will be invalid.

PR #26556, opened 2020-07-23
Dashboard: Fix for viewer can enter panel edit mode by modifying url (but cannot not save anything) 

### LINKED ISSUE #26555: Dashboard: Viewer can enter panel edit mode by modifying url (but not save anything)

torkelo:
Fixes #26555

Regression in 7.0 with the change to edit mode

### DIFF
--- a/public/app/features/dashboard/containers/DashboardPage.tsx
+++ b/public/app/features/dashboard/containers/DashboardPage.tsx
@@ -117,6 +117,12 @@ export class DashboardPage extends PureComponent<Props, State> {
     // entering edit mode
     if (!editPanel && urlEditPanelId) {
       this.getPanelByIdFromUrlParam(urlEditPanelId, panel => {
+        // if no edit permission show error
+        if (!dashboard.canEditPanel(panel)) {
+          this.props.notifyApp(createErrorNotification('Permission to edit panel denied'));
+          return;
+        }
+
         this.setState({ editPanel: panel });
       });
     }
--- a/public/app/features/dashboard/containers/DashboardPage.test.tsx
+++ b/public/app/features/dashboard/containers/DashboardPage.test.tsx
@@ -139,6 +139,20 @@ describe('DashboardPage', () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


public/app/features/dashboard/containers/DashboardPage.tsx
```diff
@@ -117,6 +117,12 @@ export class DashboardPage extends PureComponent<Props, State> {
     // entering edit mode
     if (!editPanel && urlEditPanelId) {
       this.getPanelByIdFromUrlParam(urlEditPanelId, panel => {
+        // if no edit permission show error
+        if (!dashboard.canEditPanel(panel)) {
+          this.props.notifyApp(createErrorNotification('Permission to edit panel denied'));
+          return;
```
stevenvachon, line 123: An `else` would be easier to notice than a nested `return`.

torkelo: the alternative is 

```ts
if (dashboard.canEditPanel(panel)) {
    this.setState({ editPanel: panel });
} else {
   this.props.notifyApp(createErrorNotification('Permission to edit panel denied'))
}
```

But not sure why, just find this more readable and clean:

```ts
if (!dashboard.canEditPanel(panel)) {
   this.props.notifyApp(createErrorNotification('Permission to edit panel denied')); 
   return; 
} 

this.setState({ editPanel: panel });
```

stevenvachon: I disagree, heh, but it's probably not important enough.

stevenvachon (approved):

public/app/features/dashboard/containers/DashboardPage.tsx
torkelo: no, but fun to talk about code readability. I think it's very close between these two. Just I find limiting nesting to help readability,  and having edge cases & error handling at the top and the normal flow follow after.  This being an inner callback function makes me question the use of early return pattern a bit more than usual (ie side with your view)

PR #26560, opened 2020-07-23
fix: return missing exports

### LINKED ISSUE #26554: missing ValueMappingsEditor and ThresholdsEditor export
**What happened**:

in `@grafana/ui@6.7` there is export ValueMappingsEditor and ThresholdsEditor  from `@grafana/ui/components`. in `@grafana/ui@7` this exports are missing.
I have checked that code exists in bundle `7.0.x` and `7.1.0`, but exports is missing.

export ValueMappingsEditor and ThresholdsEditor

Will be pleased if storybook will be plugin developers friendly and ccomponents will be exports starting from `@grafana/ui` specifier.
Снимок экрана 2020-07-23 в 20 27 28

This actually solves problem above since story will not work as expected (export is missing)

Tip: enhance webpack storybook config with
```
{
    resolve: {
            alias: {
                  "@grafana/ui": path.resolve(__dirname, "<relative path to grafana-ui dir>")
            }
    }
}
```

with webpack@5 you could define [`exports` field](https://github.com/webpack/webpack/pull/10953) and use self-referencing by default 

**Environment**:
- Grafana version: 7
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

vankop:

described in #26554 

fixes #26554 

### DIFF
--- a/packages/grafana-ui/src/components/index.ts
+++ b/packages/grafana-ui/src/components/index.ts
@@ -108,6 +108,8 @@ export { DataLinksContextMenu } from './DataLinks/DataLinksContextMenu';
 export { SeriesIcon } from './Legend/SeriesIcon';
 export { InfoBox } from './InfoBox/InfoBox';
 export { FeatureInfoBox } from './InfoBox/FeatureInfoBox';
+export { ThresholdsEditor } from './ThresholdsEditorNew/ThresholdsEditor';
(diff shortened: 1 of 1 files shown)


torkelo:
The ThresholdsEditor is more stable maybe is something we want to expose outside field configs usage. But the value mappings editor is not stable and should not really be used standalone (outside field config)

PR #26573, opened 2020-07-24
Drone: Identify Drone runner for each pipeline

aknuds1:
On the recommendation of @captncraig start Drone pipelines with a step identifying the runner, for aid in debugging.


papagian (approved):
LGTM

PR #26577, opened 2020-07-24
Generic OAuth: customize login and id_token attributes

alexanderzobnin:

This PR closes few RA requests related to generic OAuth:

- Allow to use custom field for user login.
- Able to set empty `client_secret` (customer does not use client secret)
- Able to use token other than `id_token` (`access_token` in customer case) for extracting user information.

Original issue is in the Enterprise repo.


DanCech:
Looks reasonable to me

sakjur (commented):
This looks really nice 👍 Is there anything else needed for this to not be in draft mode anymore?

Also, I have a similar PR up on #25744, but yours is better (I didn't figure out the testing) so I'll close mine 🙂

alexanderzobnin:
@sakjur Just waiting for customer feedback (not 100% sure if they want to use a custom field for login or username).

DanCech (approved):

docs/sources/auth/generic-oauth.md
```diff
@@ -59,6 +59,10 @@ Check for the presence of a role using the [JMESPath](http://jmespath.org/exampl
 
 See [JMESPath examples](#jmespath-examples) for more information.
 
+> Only available in Grafana v7.2+.
+
+Customize user login using `login_attribute_path` configuration option. Grafana will attempt to evaluate the `login_attribute_path` JMESPath expression against the ID token first and then, if no value found, against the HTTP response obtained from querying the UserInfo endpoint specified via the `api_url` configuration option. The attribute name which will be used for extracting ID token from the returned OAuth token can be customized with `id_token_attribute_name` option.
```
aknuds1, line 64:
```suggestion
Customize user login using `login_attribute_path` configuration option. Grafana will attempt to evaluate the `login_attribute_path` JMESPath expression against the ID token first and then, if no value is found, against the HTTP response obtained by querying the UserInfo endpoint specified via the `api_url` configuration option. The attribute name which will be used for extracting the ID token from the returned OAuth token can be customized with the `id_token_attribute_name` option.
```

pkg/login/social/generic_oauth_test.go
```diff
@@ -332,3 +332,97 @@ func TestUserInfoSearchesForEmailAndRole(t *testing.T) {
...
+
+		for _, test := range tests {
+			provider.loginAttributePath = test.LoginAttributePath
+			t.Run(test.Name, func(t *testing.T) {
+				response, err := json.Marshal(test.APIURLResponse)
+				require.NoError(t, err)
+				ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.WriteHeader(http.StatusOK)
+					w.Header().Set("Content-Type", "application/json")
+					_, err = io.WriteString(w, string(response))
```
aknuds1, line 410: Isn't this more straightforward?
```suggestion
					_, err = io.Copy(w, response)
```

aknuds1 (commented):
Just spotted some nits.

docs/sources/auth/generic-oauth.md
oddlittlebird: Best practice is to keep sentences at 25 words or less. Please break up the second sentence.

Also, try rewriting this in present tense. "Grafana attempts" instead of "Grafana will attempt" and so on.

pkg/login/social/generic_oauth_test.go
alexanderzobnin: Hm, in this case `response` should be `io.Reader`, not a `[]byte`.

aknuds1: @alexanderzobnin Ah, I didn't notice that difference.

aknuds1: In that case, wouldn't the following work?

```suggestion
_, err = w.Write(response)
```

docs/sources/auth/generic-oauth.md
alexanderzobnin: @oddlittlebird could you take a quick look at the changes?

pkg/login/social/generic_oauth_test.go
alexanderzobnin: To be honest, I just copied it from the test above :). But I'll check it out.

aknuds1: Aha, well unless I'm mistaken, going via `io.WriteString` is pointless.

docs/sources/auth/generic-oauth.md
```diff
@@ -59,6 +59,10 @@ Check for the presence of a role using the [JMESPath](http://jmespath.org/exampl
 
 See [JMESPath examples](#jmespath-examples) for more information.
 
+> Only available in Grafana v7.2+.
+
+Customize user login using `login_attribute_path` configuration option. Grafana attempts to evaluate the `login_attribute_path` JMESPath expression against the ID token. If no value is found, Grafana uses for the path lookup the HTTP response obtained by querying the UserInfo endpoint specified via the `api_url` configuration option. The attribute name which will be used for extracting the ID token from the returned OAuth token can be customized with the `id_token_attribute_name` option.
```
oddlittlebird, line 64:
```suggestion
Customize user login using `login_attribute_path` configuration option. Order of operations is as follows:

1. Grafana evaluates the `login_attribute_path` JMESPath expression against the ID token. 
1. If Grafana finds no value, then Grafana queries the UserInfo endpoint and uses the HTTP response for the path lookup. The UserInfo endpoint is specified in the `api_url` configuration option.

You can customize the attribute name used to extract the ID token from the returned OAuth token with the `id_token_attribute_name` option.
```

oddlittlebird: I took a stab at trying to reorganize this paragraph a bit.

It looks like this information could be organized either in steps (as I did) or by configuration options involved.

pkg/login/social/generic_oauth_test.go
aknuds1: I made a [PR](https://github.com/grafana/grafana/pull/26679) to change the existing test to use `w.Write`.

oddlittlebird (approved):
LGTM!

PR #26578, opened 2020-07-24
CloudWatch: Fix a few API status codes

aknuds1:
For CloudWatch, fix a few API status codes to be 400 instead of 500 for what are client and not server errors, plus use `errors` package to support wrapped errors. Additionally some code cleanup and helpful debug logging (that I needed while writing an API test).

### DIFF
--- a/pkg/services/sqlstore/sqlstore.go
+++ b/pkg/services/sqlstore/sqlstore.go
@@ -80,7 +80,7 @@ func (ss *SqlStore) Init() error {
 	x = engine
 	dialect = ss.Dialect
 
-	migrator := migrator.NewMigrator(x)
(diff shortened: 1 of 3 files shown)


pkg/services/sqlstore/sqlstore.go
```diff
@@ -80,7 +80,7 @@ func (ss *SqlStore) Init() error {
 	x = engine
 	dialect = ss.Dialect
 
-	migrator := migrator.NewMigrator(x)
+	migrator := migrator.NewMigrator(engine)
```
aknuds1, line 83: Use `engine` since it's a local variable, instead of `x`which is a global variable we want to get rid of.

papagian (approved):
LGTM!

PR #26582, opened 2020-07-24
Drone: Reduce parallelism of certain steps to conserve memory

aknuds1:
Reduce parallelism of some Drone steps in order to avoid running out of memory. @captncraig suspects some builds are killed due to OOM, let's see if we can avoid it.


papagian (approved):
LGTM!

PR #26595, opened 2020-07-24
TablePanel: Add support for basic gauge as a cell display mode

### LINKED ISSUE #25704: Add the "Basic" single color based display mode for bar guages in tables
Currently the bar guages in table cells can only be displayed as the Retro LCD style or the Gradient style. Adding the option of the single colored "Basic" display mode, which is available for bar guage panels, would also be a good option.

**Why is this needed**:
The Basic display mode is IMO more helpful in recognizing the anomalies in a series. Values which are below or above a certain threshold will be displayed in a completely different color.

jutley:

This PR adds support for the Basic Gauge type as a table cell display mode. 

Fixes #25704


torkelo (approved):
LGTM

PR #20650, opened 2019-11-25
Units: Remove SI prefix symbol from new milli/microSievert(/h) units

zegelin:
**What this PR does / why we need it**: 
Changes the definition of the new _milli/microSievert_ and _milli/microSievert/h_ units to no-longer include the SI prefix symbol (i.e, µ) in the definition.

The incorrect definition resulted in the display of duplicate prefix symbols (such as µµSv/h).

**Which issue(s) this PR fixes**:

Fixes #20614

Issue #20614, opened 2019-11-24
microSievert(/hour) shows as µµSv(/h)


zegelin:
**What happened**:

When graphing µSv/h values (from a geiger counter) using the new _microSievert/hour_ unit (#20144), Grafana displays non-fractional values (e.g, 5.65) as µµSv/h -- note the double µ.

One has to scale the values by 1E-6 for the correct unit to display.

**What you expected to happen**:

The raw values are already in µSv/h, hence they should be displayed as µSv/h when the _microSievert/hour_ unit is selected. They should not need to be scaled.

Compare this to any other unit that uses the `decimalSIPrefix` function.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a panel and use the test datasource.
1. Set the unit for the Y-axis to _microSievert/hour_.
1. Observe that the Y-axis displays all values as µµSv/h.
1. Set the unit for the Y-axis to _millimeter (mm)_ (another unit that uses the `decimalSIPrefix` function)
1. Observe that the values are correctly displayed as mm.

**Environment**:
- Grafana version: 6.5.0-beta1
- Data source type & version: Prometheus
- OS Grafana is installed on: Arch Linux
- User OS & Browser: Windows & FireFox
- Grafana plugins: _none_
- Others: _n/a_

PR #26598, opened 2020-07-25
Units: Remove duplicate SI prefix from mSv and µSv

tofurky:

This removes duplicate SI prefixes from the sievert units. Otherwise they show up as e.g. µµSv instead of µSv.

The associated issue for the time-based units was #20614, but this specific case was not mentioned. That was addressed by #20650, but only for mSv/h and µSv/h.

I gave the rest of categories.ts a once-over and didn't see any other decimalSIPrefix()'s that had the issue.


torkelo (approved):
LGTM

PR #26600, opened 2020-07-25
Dashboard: Add Duplicate dashboard links button to links list

Hmerac:
**What this PR does / why we need it**: This PR allows users to duplicate individual `Dashboard Links` to work faster. I wish this feature was already implemented when I was creating `Dashboard Links `for all regions that we have deployments in. Pressing `New` button and copy-pasting each textbox was quite a burden. With this feature, users can easily duplicate the links and change only the needed/required areas. For example: region parts of the URL, title...

Screenshot 2020-07-25 at 17 21 47

**Which issue(s) this PR fixes**: _None of the existing ones_

**Special notes for your reviewer**: First contribution :) ~I wasn't really sure which button type to use. I've decided `repeat` is the most appropriate, although `copy` also looks appropriate.~ Changed to `copy`.

### DIFF
--- a/docs/sources/linking/dashboard-links.md
+++ b/docs/sources/linking/dashboard-links.md
@@ -64,8 +64,8 @@ To change or update an existing dashboard link, follow this procedure.
 
 ## Duplicate a dashboard link
 
-To duplicate an existing dashboard link, click the blue **⇄** next to the existing link that you want to duplicate.
+To duplicate an existing dashboard link, click the duplicate icon next to the existing link that you want to duplicate.
 
 ## Delete a dashboard link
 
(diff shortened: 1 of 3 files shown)


torkelo:
Please use same icons & buttons as the variables list

torkelo (changes requested):
Needs to look the same as variables list (That also has duplicate button)

Hmerac:
> Needs to look the same as variables list (That also has duplicate button)

I believe the button types in the screenshot is what you meant:

Screenshot 2020-07-26 at 20 46 12

torkelo (approved):
Looks good, thanks for contributing ❤️

PR #26607, opened 2020-07-26
TemplateSrv: Refactor out all variable formatting to separate file and introduce a formatRegistry

torkelo:
builds on #26466

### DIFF
--- a/public/app/features/templating/formatRegistry.ts
+++ b/public/app/features/templating/formatRegistry.ts
@@ -1,0 +1,212 @@
+import kbn from 'app/core/utils/kbn';
+import { Registry, RegistryItem, VariableModel, textUtil, dateTime } from '@grafana/data';
+import { map, isArray, replace } from 'lodash';
+
+export interface FormatRegistryItem extends RegistryItem {
+  formatter(value: any, args: string[], variable: VariableModel): string;
+}
+
+export const formatRegistry = new Registry<FormatRegistryItem>(() => {
+  const formats: FormatRegistryItem[] = [
+    {
+      id: 'lucene',
+      name: 'Lucene',
+      description: 'Values are lucene escaped and multi-valued variables generate an OR expression',
+      formatter: value => {
+        if (typeof value === 'string') {
+          return luceneEscape(value);
+        }
+
+        if (value instanceof Array && value.length === 0) {
+          return '__empty__';
+        }
+
+        const quotedValues = map(value, (val: string) => {
+          return '"' + luceneEscape(val) + '"';
+        });
+
+        return '(' + quotedValues.join(' OR ') + ')';
+      },
+    },
+    {
+      id: 'regex',
+      name: 'Regex',
+      description: 'Values are regex escaped and multi-valued variables generate a (<value>|<value>) expression',
+      formatter: value => {
+        if (typeof value === 'string') {
+          return kbn.regexEscape(value);
+        }
+
+        const escapedValues = map(value, kbn.regexEscape);
+        if (escapedValues.length === 1) {
+          return escapedValues[0];
+        }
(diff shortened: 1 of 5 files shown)


public/app/plugins/datasource/grafana-azure-monitor-datasource/query_ctrl.ts
```diff
@@ -581,13 +581,9 @@ export class AzureMonitorQueryCtrl extends QueryCtrl {
...
 
   /* Application Insights Section */
-
   getAppInsightsAutoInterval() {
-    const interval = this.templateSrv.getBuiltInIntervalValue();
-    if (interval[interval.length - 1] === 's') {
-      return '1m';
-    }
-    return interval;
+    // TODO no way to get this currently from angular
```
ryantxu, line 585: can we remove the whole function?  lots has changed in azure now that it is mostly a backend plugin, but the frontend still has many function around

public/app/features/templating/formatRegistry.ts
```diff
@@ -0,0 +1,212 @@
+import kbn from 'app/core/utils/kbn';
+import { Registry, RegistryItem, VariableModel, textUtil, dateTime } from '@grafana/data';
+import { map, isArray, replace } from 'lodash';
+
+export interface FormatRegistryItem extends RegistryItem {
+  formatter(value: any, args: string[], variable: VariableModel): string;
+}
+
+export const formatRegistry = new Registry<FormatRegistryItem>(() => {
```
ryantxu, line 9: This looks great and is a HUGE improvement over no registry (with no metadata)

Should the namespace include template? [valueFormats](https://github.com/grafana/grafana/tree/master/packages/grafana-data/src/valueFormats) are often referred to as formats/formatter so may changing this to `templateFormatregistry` would help?

ryantxu (approved):
This looks great and will help expose the possible formats to future autocomplete wizards.  It is in the internal package, so it is not a big deal, but maybe we should prefix the exported names with `template`?

public/app/features/templating/formatRegistry.ts
torkelo: maybe variableFormatRegistry? It's not part of any @grafana/data namespace yet (only an internal app thing) so easy to change

torkelo: or longer: interpolationFormatRegistry

public/app/plugins/datasource/grafana-azure-monitor-datasource/query_ctrl.ts
torkelo: Think it's still used form UI, but this functionality has been broken since v6

torkelo: But can see if I can update the angular template that calls this

public/app/features/templating/formatRegistry.ts
ryantxu: `variableFormatRegistry` is ok -- I'm not sure how well communicated variables vs template variables vs "templating" is.  For sure gets confusing when compared to macros often used in backend

But I agree this is not an urgent name change as it is internal

public/app/features/templating/formatRegistry.ts
```diff
@@ -0,0 +1,212 @@
...
+  const formats: FormatRegistryItem[] = [
```
simPod, line 10: personally I'd sort it alphabetically by format id for better maintenance

simPod (commented):
Nice improvement, this broke `${__cell_0:raw}` though

public/app/features/templating/formatRegistry.ts
dprokop: Great stuff with the registry!

PR #26611, opened 2020-07-27
fix template variable not loading on dashboard loading

### LINKED ISSUE #25757: ${__user.login} template not loaded properly on variable Init

mohamednizar:
When the dashboard loading template variable not loading 

#25757 

Fixes #25757 

- The bug is due to some missing practices of  Grafana variable implementation.

### DIFF
--- a/public/app/features/variables/state/actions.ts
+++ b/public/app/features/variables/state/actions.ts
@@ -27,7 +27,7 @@ import {
   resolveInitLock,
   setCurrentVariableValue,
 } from './sharedReducer';
-import { toVariableIdentifier, toVariablePayload, VariableIdentifier } from './types';
+import { AddVariable, toVariableIdentifier, toVariablePayload, VariableIdentifier } from './types';
 import { appEvents } from 'app/core/core';
 import { contextSrv } from 'app/core/services/context_srv';
 import templateSrv from '../../templating/template_srv';
@@ -78,14 +78,21 @@ import isEqual from 'lodash/isEqual';
 
 export const initDashboardTemplating = (list: VariableModel[]): ThunkResult<void> => {
   return (dispatch, getState) => {
-    let orderIndex = 0;
     for (let index = 0; index < list.length; index++) {
       const model = list[index];
       if (!variableAdapters.getIfExists(model.type)) {
         continue;
       }
 
-      dispatch(addVariable(toVariablePayload(model, { global: false, index: orderIndex++, model })));
+      const global = false;
+      // const index = getNewVariabelIndex(getState());
+      const identifier: VariableIdentifier = { type: 'adhoc', id: model.id };
+
+      dispatch(
+        addVariable(
+          toVariablePayload<AddVariable>(identifier, { global, model, index })
+        )
+      );
     }
 
     templateSrv.updateTimeRange(getTimeSrv().timeRange());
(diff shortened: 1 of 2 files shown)


public/app/features/variables/state/actions.ts
```diff
@@ -78,14 +78,21 @@ import isEqual from 'lodash/isEqual';
...
-      dispatch(addVariable(toVariablePayload(model, { global: false, index: orderIndex++, model })));
+      const global = false;
+      // const index = getNewVariabelIndex(getState());
+      const identifier: VariableIdentifier = { type: 'adhoc', id: model.id };
+
+      dispatch(
+        addVariable(
+          toVariablePayload<AddVariable>(identifier, { global, model, index })
+        )
+      );
```
torkelo, line 95: This change does not make any sense to mee, whats the point of it? what's the point of global variable?  almost every line changed here looks very strange

public/app/features/variables/state/actions.ts
```diff
@@ -555,10 +562,12 @@ export const initVariablesTransaction = (dashboardUid: string, dashboard: Dashbo
     dispatch(variablesInitTransaction({ uid: dashboardUid }));
 
     dispatch(initDashboardTemplating(dashboard.templating.list));
-    await dispatch(processVariables());
+
     dispatch(completeDashboardTemplating(dashboard));
 
     dispatch(variablesCompleteTransaction({ uid: dashboardUid }));
+
+    await dispatch(processVariables());
```
torkelo, line 570: this needs to where it is, as the two actions above rely on variable queries having been executed

torkelo (changes requested):
Thanks for trying to fix this! But this looks like it will introduce many more bugs

public/app/features/variables/state/actions.ts
mohamednizar: Yes, agreed this will not make sense to the bug.

public/app/features/variables/state/actions.ts
mohamednizar: `processVariables` needs to be dispatched after `completeDashboardTemplating` . than only it's refreshing the variable on dashboard load 
shall we call the function after completeDashboardTemplating  ?

mohamednizar:
@torkelo  The build error exists in the master too.

torkelo:
Opened another PR with I think a correct fix, https://github.com/grafana/grafana/pull/26637

Thanks for contributing and identifying the problem!

PR #26613, opened 2020-07-27
Settings: Remove global variable

torkelo:
Somehow we have added more global setting variables (everything should move to Cfg struct).


ryantxu (approved):

papagian (approved):
LGTM

PR #26616, opened 2020-07-27
StatPanel: Fix stat panel display name not showing when explicitly set

### LINKED ISSUE #26583: Stat Panel: display name not showing when text mode set to Auto and explicitly set in field config
**What happened**: 
The StatPanel doesn't render the display name anymore, since the 7.1

See display name in the panel

* setup a new dashboard, with a stat panel
* set a display name

Everything was fine with 7.0.x

**Environment**:
- Grafana version: 7.1.1
- Data source type & version: default, stackdriver (and probably all others datasources)
- OS Grafana is installed on: kubernetes via the official helm chart
- User OS & Browser: mac os / chrome

torkelo:
Fixes #26583

A bit conflicted on how to solve this or if it should be solved. with Text mode set to Auto should stat panel show display name if set via Field tab default? what about via override? I started on a fix like that, to check in StatPanel.tsx if fieldConfig.defaults.displayName was set and then switch text mode to value and name (passed to BigValue). But not sure, maybe Auto should just be if count === 1 then do not show name, if you want to show name if count === 1 you have to change display name. and to fix broken old dashboards instead of adding logic in StatPanel.tsx we can do a migration that for panels with default displayName set we change textMode. Only downside with that users will try to change Display name and it won't have any effect when text mode set to Auto.

### DIFF
--- a/packages/grafana-data/src/types/panel.ts
+++ b/packages/grafana-data/src/types/panel.ts
@@ -69,6 +69,8 @@ export interface PanelProps<T = any> {
   onChangeTimeRange: (timeRange: AbsoluteTimeRange) => void;
   /** @internal */
   renderCounter: number;
+  /** Panel title */
+  title: string;
 }
 
 export interface PanelEditorProps<T = any> {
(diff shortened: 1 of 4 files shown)


ryantxu:
This approach seems OK.  We may also want to check if the panel title is set?  In the example on #26583, the missing panel title makes the auto behavior less than great

ryantxu (approved):

ryantxu:
> what about via override?

I don't think so -- this feels a little better and should avoid most weird cases

torkelo:
@ryantxu added panel title to PanelProps so the auto logic can also take that into consideration

PR #26621, opened 2020-07-27
CloudWatch Logs: Adjusts CloudWatch Logs timeout logic

kaydelaney:
Previously CloudWatch Logs queries would time out if, after a total number of attempts, a response was received with no additional data.
This commit changes the behavior so that a _consecutive_ number of requests yielding no additional data must be made before we cancel the query.

### DIFF
--- a/public/app/features/explore/state/reducers.ts
+++ b/public/app/features/explore/state/reducers.ts
@@ -13,6 +13,7 @@ import {
   ExploreMode,
   LogsDedupStrategy,
   sortLogsResult,
+  DataQueryErrorType,
 } from '@grafana/data';
 import { RefreshPicker } from '@grafana/ui';
 import { LocationUpdate } from '@grafana/runtime';
@@ -510,12 +511,14 @@ export const processQueryResponse = (
   const { request, state: loadingState, series, error } = response;
 
   if (error) {
-    if (error.cancelled) {
+    if (error.type === DataQueryErrorType.Timeout) {
       return {
         ...state,
         queryResponse: response,
         loading: loadingState === LoadingState.Loading || loadingState === LoadingState.Streaming,
       };
+    } else if (error.type === DataQueryErrorType.Cancelled) {
+      return state;
     }
 
     // For Angular editors
--- a/public/app/features/explore/state/reducers.ts
+++ b/public/app/features/explore/state/reducers.ts
@@ -511,7 +511,11 @@ export const processQueryResponse = (
 
   if (error) {
     if (error.cancelled) {
-      return state;
+      return {
+        ...state,
+        queryResponse: response,
+        loading: loadingState === LoadingState.Loading || loadingState === LoadingState.Streaming,
+      };
     }
 
     // For Angular editors
--- a/public/app/plugins/datasource/cloudwatch/datasource.ts
+++ b/public/app/plugins/datasource/cloudwatch/datasource.ts
@@ -234,7 +234,7 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
       };
     });
 
-    const dataFrames = increasingInterval(100, 1000, 300).pipe(
+    const dataFrames = increasingInterval({ startPeriod: 100, endPeriod: 1000, step: 300 }).pipe(
       concatMap(_ => this.makeLogActionRequest('GetQueryResults', queryParams)),
       repeat(),
       share()
--- a/public/app/plugins/datasource/cloudwatch/datasource.ts
+++ b/public/app/plugins/datasource/cloudwatch/datasource.ts
@@ -241,20 +241,23 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
       share()
     );
 
-    let prevRecordsMatched: Record<string, number> = {};
     const consecutiveFailedAttempts = dataFrames.pipe(
-      scan((acc, frames) => {
-        let newAcc = acc + 1;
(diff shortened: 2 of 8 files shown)


public/app/features/explore/state/reducers.ts
```diff
@@ -511,7 +511,11 @@ export const processQueryResponse = (
 
   if (error) {
     if (error.cancelled) {
-      return state;
+      return {
```
kaydelaney, line 514: Not 100% sure that this is okay to do, but seemed like the simplest way to be able to keep fetched results while still displaying an error message

aocenas: Hmm not sure about this. If I get this correctly this works for states when it times out but still got some response data. So like CloudWatch logs where some data can come through even though some may timeout.

For other datsources this will probably do the opposite. If the response times out before we preserved any previous data but now we will replace it with the response that probably does not have any data, is that right?

public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -234,64 +233,74 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
         statsQuery: (param.statsGroups?.length ?? 0) > 0 ?? false,
       };
     });
+
+    const dataFrames = increasingInterval(100, 1000, 300).pipe(
```
aocenas, line 237: This would probably be more readable as named params/object instead ie `{startPeriod: 100, endPeriod: 1000, step: 300}`

public/app/plugins/datasource/cloudwatch/utils/rxjs/increasingInterval.ts
```diff
@@ -0,0 +1,30 @@
+import { SchedulerLike, Observable, SchedulerAction, Subscriber, asyncScheduler } from 'rxjs';
+
+export const increasingInterval = (
```
aocenas, line 3: Needs some comment to quickly understand what this does.

public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -234,64 +233,74 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
...
+      concatMap(_ => this.makeLogActionRequest('GetQueryResults', queryParams)),
+      repeat(),
```
aocenas, line 239: I am wondering, aren't we duplicating the retry timeout logic between frontend and beckend part? I remember having some sort of retry/timeout in Go.

public/app/features/explore/state/reducers.ts
kaydelaney: Ah, you're right, seems like some more work will have to go into this. What do you think of adding a different `DataQueryError` prop like `timedOut`, or perhaps better, refactoring the interface so that there's an `errorType` prop that could be one of `ErrorType.Cancelled`, `ErrorType.TimeOut` etc

public/app/plugins/datasource/cloudwatch/datasource.ts
kaydelaney: Fair enough, will change

public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -234,64 +234,77 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
...
-                  const noProgressMade = i >= MAX_ATTEMPTS - 2 && !moreRecordsMatched;
-                  if (noProgressMade) {
-                    for (const frame of frames) {
-                      _.set(frame, 'meta.custom.Status', CloudWatchLogsQueryStatus.Complete);
-                    }
-                  }
+        return newAcc;
+      }, 0),
+      share()
+    );
```
hugohaggmark, line 259: Outstanding work with this! It's been a while since I used `scan` but couldn't you pass a more complex object then `0` as the seed? That would eliminate the use of `let prevRecordsMatched: Record<string, number> = {};` I think

kaydelaney:
```typescript
    const consecutiveFailedAttempts = dataFrames.pipe(
      scan(
        ({ failures, prevRecordsMatched }, frames) => {
          failures++;
          for (const frame of frames) {
            const recordsMatched = frame.meta?.custom?.['Statistics']['RecordsMatched'];
            if (recordsMatched > (prevRecordsMatched[frame.refId!] ?? 0)) {
              failures = 0;
            }
            prevRecordsMatched[frame.refId!] = recordsMatched;
          }

          return { failures, prevRecordsMatched };
        },
        { failures: 0, prevRecordsMatched: {} as Record<string, number> }
      ),
      map(({ failures }) => failures),
      share()
    );
```

Something like this?

hugohaggmark: Yeah, that looks excellent ⭐

public/app/features/explore/state/reducers.ts
aocenas: So checking now and it seems like in case of errors we do not keep the previous response data anyway. I mean it may be confusing. Right now you can get response with data + error like in this case and it would be confusing if there were cases where we left the previous data cause than you wouldn't know if that is old data or you get new response but it's looking the same. So I think this approach is fine.

aocenas (approved):

marefr:
@kaydelaney what's the status of this. Have 7.1.2 milestone?

PR #26628, opened 2020-07-27
Docs: Add preferences and password docs

### LINKED ISSUE #21365: Docs: Document changing user passwords
**What would you like to be added**: Instructions so that users can passwords in the Grafana UI. Right now, we only have written instructions for changing the Admin password using Grafana CLI.

**Why is this needed**: Because right now a Google search for "grafana change user password" brings up a bunch of other companies, then the Grafana docs instructions for changing the Admin password. We do not seem to have explicit instructions to help users or user admins change passwords.

oddlittlebird:
Added basic topics about Grafana preferences and password information.

I wrote these with an eye towards answering questions that users are likely to Google.

Fixes: #21365

### DIFF
--- a/docs/sources/administration/_index.md
+++ b/docs/sources/administration/_index.md
@@ -10,4 +10,4 @@ weight = 2
 
 # Administration
 
-This section provides resources for Grafana administrators.
+This section provides resources for Grafana administrators and users performing administrative tasks.
--- a/docs/sources/administration/change-your-password.md
+++ b/docs/sources/administration/change-your-password.md
@@ -15,7 +15,7 @@ You can change your password in the Change Password tab.
 
 ## Change your password
 
-1. Hover your mouseover your user icon in the lower left corner of the screen.
+1. Hover your mouse over your user icon in the lower left corner of the screen.
 1. Click **Change Password**. Grafana opens the Change Password tab.
 1. Enter your **Old password** to authorize the change.
 1. Enter your **New password** and then **Confirm password**.
--- a/docs/sources/administration/change-your-password.md
+++ b/docs/sources/administration/change-your-password.md
@@ -11,7 +11,7 @@ weight = 100
 
 You can change your password in the Change Password tab.
 
-> **Note:** If your Grafana instance uses single sign-on (SSO), then you might not be able to change your password. Contact your Grafana administrator for more information.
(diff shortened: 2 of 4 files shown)


adeverteuil:
I don't think this belongs in the Administration section. The Administration section is for the server side of things. I would place those pages under Manage Users.

I see that there is a link to a page 'change-home-dashboard.md'. This should probably be moved under Manage Users too. We could create a Manage Organizations section but orgs are being deprecated and will be removed in a future version.

Admin user management resources: there is also a grafana-cli command to reset the admin user's password if it is lost 'administration/cli.md#reset-admin-password'.

oddlittlebird:
"Administration" does not actually state anywhere that it is for the server side of things. I eventually plan to rewrite and reorganize the entire section, but for now, it is the most appropriate place for these topics.

"Manage users" is not an appropriate place for these topics. The content does not match the section title.

sakjur (commented):
I agree with Alexandre that "Administration" is not the right place for this. Maybe we could rename "Manage users" to "Users"? That'd match the mix of administrative and end-user documentation divided by resource which is the case for alerts, data sources, dashboards, and so on. Other than that and the "mouseover" typo I've highlighted I'd give an approval, the rest are loose suggestions 🙂

oddlittlebird:
So, funny thing about the "this section is for Grafana administrators" assumption, we don't actually say that anywhere except for one brief sentence in the _index file of the Administration section, and that file is not browsable.

This is the best place for that content until I get around to refactoring the entire "Administration" section into something better. No, it does not belong in "Manage users," because that section very specifically IS for Grafana admins and people with an Admin or higher role.

sakjur (approved):
👍 LGTM. Looking forward to the reorganization

adeverteuil:
Grafana "Admins" and "Server Admins" are often the same people, but not necessarily. App administration and server administration are two different tasks. And user profile/preferences management is yet another task.

I don't think "this section is for Grafana administrators" is a funny assumption for a section called Administration that also contains only content on server administration.

I agree that there is currently no better place to put the files for now, and I re-read the Mange Users overview and that is also not the right place for it. I look forward to the refactoring of the Administration section, I just want to make sure there is a common understanding that this introduces an inconsistency until the section is better organized.

Let's go! :+1:

PR #26633, opened 2020-07-27
Explore: Simplifies rich history drawer animation logic

kaydelaney:
Changes the logic of the rich history drawer animation to use CSS animations rather than a timer and react state.


ivanahuckova (approved):
LGTM! 🙂

PR #26634, opened 2020-07-28
InfluxDB: support annotations in flux and DataSourceWithBackend

### LINKED ISSUE #26357: InfluxDB: support flux queries in template variables
When using InfluxDB with flux query support, we should also be able to author template variable queries in flux 

**Why is this needed**:
to support template variables ;)

ryantxu:
Fixes #26357

still WIP


packages/grafana-runtime/src/components/AnnotationResultMapper.tsx
```diff
@@ -0,0 +1,201 @@
+import React, { PureComponent } from 'react';
```
torkelo, line 1: why is this in runtime package?

public/app/plugins/datasource/influxdb/components/AnnotationQueryEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+    this.runQuery();
+  }
+
+  componentDidUpdate(oldProps: Props) {
+    if (this.props.annotation !== oldProps.annotation) {
+      this.runQuery();
+    }
+  }
+
+  runQuery = async () => {
```
torkelo, line 40: should this really be in the editor, can't we move this to Grafana app/annotation editor view?  (I know it's in angular and a lot of work to do there to migrate and rethink that)

ryantxu: should it be?  NO!

ideally this would be the default behavior if you do not specify anything... but given our setup that is a much larger project.  My plan here is to make things in runtime as generic as possible so that anything that extends DataSourceWithBackend can just use them, and when we put the effort towards annotations in general we can reuse this component/path

packages/grafana-runtime/src/components/AnnotationResultMapper.tsx
ryantxu: I don't want to add it to `@grafana/ui` since it is a stop-gap until this can just be default behavior.  My goal is to get *something* so that I can easily use this same pattern/component for the immediate projects I am working on:
* influxdb
* adx
* github
* snowflake

ryantxu:
Closing in favor of #27375

PR #26636, opened 2020-07-28
Dashboard: Fixed disable draggable panels on correct breakpoint

torkelo:
Dashboard switches to responsive single column layout at medium breakpoint (769px) so DasbboardGrid needs to disable dragging at same point, noticed this as if you click on panel headers in responsive mode the panels get dragged to the bottom by just that click

### DIFF
--- a/public/app/features/dashboard/dashgrid/DashboardGrid.tsx
+++ b/public/app/features/dashboard/dashgrid/DashboardGrid.tsx
@@ -64,8 +64,9 @@ function GridWrapper({
   /*
     Disable draggable if mobile device, solving an issue with unintentionally
      moving panels. https://github.com/grafana/grafana/issues/18497
+     theme.breakpoints.md = 769
   */
-  const draggable = width <= 420 ? false : isDraggable;
+  const draggable = width <= 769 ? false : isDraggable;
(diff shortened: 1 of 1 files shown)


public/app/features/dashboard/dashgrid/DashboardGrid.tsx
```diff
@@ -64,8 +64,9 @@ function GridWrapper({
   /*
     Disable draggable if mobile device, solving an issue with unintentionally
      moving panels. https://github.com/grafana/grafana/issues/18497
+     theme.breakpoints.md = 769
   */
-  const draggable = width <= 420 ? false : isDraggable;
+  const draggable = width <= 769 ? false : isDraggable;
```
kaydelaney, line 69: Could `theme.breakpoints.md` be referred to or does this have to be hardcoded here?

torkelo: sadly theme.breakpoints.md cannot be used as it's a string, and do not want to parse it here. We should change most variables in theme to be numbers so they can more easily be used in code

Clarity-89 (approved):

PR #26650, opened 2020-07-28
LDAP: users without org mappings are marked as disabled

xlson:

This PR corrects the mapping of `extUser` so that it knows that a user that does not have org mappings (where org mappings are expected by the ldap config) is disabled.

This will make it so that LDAP Debug view will correctly show disabled for users that don't have any valid mappings and also make it so that the LDAP background sync disables users that no longer have valid org mappings.

Fixes #

### DIFF
--- a/pkg/services/ldap/ldap.go
+++ b/pkg/services/ldap/ldap.go
@@ -427,7 +427,7 @@ func (server *Server) buildGrafanaUser(user *ldap.Entry) (*models.ExternalUserIn
 		}
 	}
 
-	// If there are groups org mappings configured but no matching mappings
+	// If there are group org mappings configured, but no matching mappings,
 	// the user will not be able to login and will be disabled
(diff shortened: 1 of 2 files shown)


pkg/services/ldap/ldap.go
```diff
@@ -427,6 +427,12 @@ func (server *Server) buildGrafanaUser(user *ldap.Entry) (*models.ExternalUserIn
 		}
 	}
 
+	// If there are groups org mappings configured but no matching mappings
```
aknuds1, line 430: I think "group org mappings" is correct?
```suggestion
	// If there are group org mappings configured, but no matching mappings,
```

aknuds1 (approved):
LGTM with a language suggestion.

Could any test be written for this?

alexanderzobnin (approved):
LGTM!

xlson:
> LGTM with a language suggestion.
> 
> Could any test be written for this?

Good point. I've added tests to cover this behaviour.

PR #26656, opened 2020-07-28
DataFrame: POC FieldNaming strategy

### LINKED ISSUE #25212: v7 Stat & Gauge panel not respecting "Display Name" for graphite queries
**What happened**:
A dashboard imported/migrated from v6 contains stat panels where I would set the Panel Title (under general) to nothing and set the "Title" in the visualiation settings to a string.  
In v7 this setting is now called "Display Name" but any string I enter is ignored and the full metric target string is used as the title, which generally makes it unreadable when not in edit view. If the display name field is set to none then nothing is shown, but otherwise it only ever shows the metric target.  
e.g. here I typed "uptime" in the display name field, but the panel is using maxSeries(logstash.stats.nodes.ops_logstash_1.jvm.uptime)
image
This seems to only be broken for graphite datasource.

I expected the display name to be used as the panel text rather than the metric target string. 

Create a Stat panel with a graphite datasource. Add a metric query. Under the panel options enter text in the "Display Name" field. The panel then updates with the metric target string rather than the diaplay name entered. 

**Environment**:
- Grafana version: 7.0.1
- Data source type & version: graphite 9.x
- User OS & Browser: MacOS 10.15.4
- Grafana plugins: 
- Others:

torkelo:
In v7.0 we added label display name logic where Grafana uses the labels to render the field name as well as some initial smart behavior of removing duplicate info (duplicate metric names or field labels). This smart logic would be a big breaking change for Prometheus and graphite  and with no time to do a UI option to control or disable this smart naming we opted to disable the smart behavior by setting field.config.displayName from Prometheus and graphite data sources as setting displayName from data source overrides the smart naming. 

However setting field.config.displayName from data source makes the panel Field tab (field config defaults) to have no effect. 

Creating bugs like #25212 for both Graphite and Prometheus 

To fix this the idea with this PR is to introduce a naming strategy instead.

* Smart (automatically remove duplicate info)
* Complete (Include all info)
* FrameName (For data sources that have legend formats and naming schemes and want to prebake the name but we still want to be able to override it via defaults.displayName, so this is the one we use to solve the problem above). 

Fixes #25212


packages/grafana-data/src/types/data.ts
```diff
@@ -17,13 +17,24 @@ export enum LoadingState {
...
   /** DatasSource Specific Values */
   custom?: Record<string, any>;
 
   /** Stats */
   stats?: QueryResultMetaStat[];
 
+  /**
+   * Field naming strategy
+   */
+  namingStrategy?: FieldNamingStrategy;
```
torkelo, line 36: Open question is should this live on field instead? we do not have field meta, so would be on FieldConfig

ryantxu: and importantly, what happens after a transform.  For example if a prometheus field wants the name from the frame... what happens after a transform?  do we make sure to render it first?

packages/grafana-data/src/types/data.ts
```diff
@@ -17,13 +17,24 @@ export enum LoadingState {
 
 export type PreferredVisualisationType = 'graph' | 'table' | 'logs' | 'trace';
 
+export enum FieldNamingStrategy {
+  Smart = 'smart',
```
ryantxu, line 21: I think @kylebrandt document calls this "short"

packages/grafana-data/src/types/data.ts
ryantxu: the "frame-name" strategy really only makes sense with time/value frames (simple timeseries)

torkelo: yea, agree so this naming strategy only makes sense for time series pre-transform, which is why I think it can live on frame. but the other two (short/smart & complete) makes sense on field, so maybe this is a field config option.

torkelo:
Tricky with the FrameName strategy, it only works for time series per data frame, so after a transform, we need to change / remove that naming strategy but that means transforms that merge / join many data frames needs to be aware of this

kylebrandt:
> FrameName (For data sources that have legend formats and naming schemes and want to prebake the name but we still want to be able to override it via defaults.displayName, so this is the one we use to solve the problem above).

I don't think I understand this.  A data source could set the displayName on the field config, and then I assume a user could use over ride this. So I'm not understanding the use case? (Think I'm missing something).

torkelo:
> I don't think I understand this. A data source could set the displayName on the field config, and then I assume a user could use over ride this. So I'm not understanding the use case? (Think I'm missing something).

The problem is if a datasource set's displayName like Graphite & Prometheus does today to sidestep the current smart/short naming we have currently makes the displayName under Field tab unusable as it only sets defaults and they do not override data source settings.

kylebrandt:
> The problem is if a datasource set's displayName like Graphite & Prometheus does today to sidestep the current smart/short naming we have currently makes the displayName under Field tab unusable as it only sets defaults and they do not override data source settings.

1. Wouldn't the Field Overrides tab solve that case?

2. If the answer to "1" is that the override tab is meant to target more specific fields, that what about something like:

 - Populated Field's Tab displayName from the datasource
 - Have UX indicator that states that it comes from the datasource
 - Have small UX button to restore what comes from the datasource if someone has put something else in there
 - (Detail, but probably would want an undo with that restore button).

torkelo:
1) Yes overrides override data source field config options.  But this is not a very good solution as it would make the default unusable would be very unintuitive for this option to never work like for other data sources. 

2) 
> Populated Field's Tab displayName from the datasource

What does that solve if changing has no effect? 

> Have UX indicator that states that it comes from the datasource

This is a good idea and something we have actually designed the UI to handle so user can see what options are overridden by data source. But still not a good option I feel as it would still always force an override rule and the displayName option under Field tab would still be unusable. 

The idea here is to have a naming strategy solve this, so that Graphite & Prometheus can return a full prebuilt names and labels and have Grafana not rebuilt it add labels to a name that already contains it

dprokop:
```
Smart (automatically remove duplicate info)
Complete (Include all info)
FrameName (For data sources that have legend formats and naming schemes and want to prebake the name but we still want to be able to override it via defaults.displayName, so this is the one we use to solve the problem above).
```

What's the point of introducing the `complete` strategy @torkelo? Is there any benefit of including the duplicate info that I cannot see?

Since we want the strategy to be executed "prior" the transformations, it feels OK to me to have this config on a frame level. I guess display consistency would be the major argument for it.

torkelo:
Closing this in favor of https://github.com/grafana/grafana/pull/27186 for now

PR #26663, opened 2020-07-28
Provisioning: setting configuration files path per type

### LINKED ISSUE #24631: split provisioning conf

It would be nice if there were multiple provisioning conf paths, perhaps (a subset of?):

/etc/grafana/conf/provisioning/datasources
/etc/grafana/conf/provisioning/notifiers
/etc/grafana/conf/provisioning/dashboards
/usr/share/grafana/conf/provisioning/datasources
/usr/share/grafana/conf/provisioning/notifiers
/usr/share/grafana/conf/provisioning/dashboards
/var/lib/grafana/conf/provisioning/datasources
/var/lib/grafana/conf/provisioning/notifiers
/var/lib/grafana/conf/provisioning/dashboards

Each with its own conf var. I can try to come up with suitable names, if needed, although naming isn't my strong side... e.g.:
conf_provisioning_datasources_path
conf_provisioning_notifiers_path
conf_provisioning_dashboards_path
sys_... (or default_..., or static_..., etc.)
var_... (or runtime_..., etc.)

**Why is this needed**:

In order to allow flexibility in having both packaged, static data (stuff that you get from an installed package when you install it, and do not change manually later), user-configurable data (which is commonly in /etc, and there are systems that (are trying to) keep /usr read-only), etc. Not sure what's the exact use-case of keeping also in /var, but in principle this is for program-managed data (not manually user-managed).

I am writing this while working on integration of oVirt DWH and grafana. I'd like to be able to supply a datasource (which is changeable, as it contains credentials, so should be in /etc) and some dashboards (which should be static, thus should be in /usr). For my own need, it might be enough to just have everything in /etc, or have conf/provisioning there and make dashboards a symlink to /usr. But that's a big ugly, IMO.

nabokihms:
This PR adds the ability to specify different paths for different types of provisioning configuration files (datasources, dashboards, etc.) via grafana config.

Fixes #24631

### DIFF
--- a/conf/defaults.ini
+++ b/conf/defaults.ini
@@ -26,6 +26,17 @@ plugins = data/plugins
 # folder that contains provisioning config files that grafana will apply on startup and while running.
 provisioning = conf/provisioning
 
+#################################### Provisioning ###############################
+[provisioning]
+# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
+datasources_path =
+# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
+dashboards_path =
+# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
+notifiers_path =
+# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
+plugins_path =
+
 #################################### Server ##############################
 [server]
 # Protocol (http, https, h2, socket)
--- a/conf/defaults.ini
+++ b/conf/defaults.ini
@@ -28,13 +28,13 @@ provisioning = conf/provisioning
 
 #################################### Provisioning ###############################
 [provisioning]
-# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
+# folder that contains data source provisioning config files, the default is paths.provisioning/datasources
 datasources_path =
-# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
+# folder that contains dashboard provisioning config files, the default is paths.provisioning/dashboards
 dashboards_path =
-# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
+# folder that contains alert notification channel provisioning config files, the default is paths.provisioning/notifiers
 notifiers_path =
-# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
+# folder that contains plugin provisioning config files, the default is paths.provisioning/plugins
 plugins_path =
 
 #################################### Server ##############################
--- a/conf/sample.ini
+++ b/conf/sample.ini
@@ -28,13 +28,13 @@
 
 #################################### Provisioning ###############################
 [provisioning]
-# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
+# folder that contains data source provisioning config files, the default is paths.provisioning/datasources
 ;datasources_path =
-# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
+# folder that contains dashboard provisioning config files, the default is paths.provisioning/dashboards
 ;dashboards_path =
-# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
+# folder that contains alert notification channel provisioning config files, the default is paths.provisioning/notifiers
 ;notifiers_path =
-# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
+# folder that contains plugin provisioning config files, the default is paths.provisioning/plugins
 ;plugins_path =
 
 #################################### Server ####################################
--- a/conf/sample.ini
+++ b/conf/sample.ini
@@ -26,6 +26,17 @@
 # folder that contains provisioning config files that grafana will apply on startup and while running.
 ;provisioning = conf/provisioning
 
+#################################### Provisioning ###############################
+[provisioning]
+# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
+;datasources_path =
+# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
+;dashboards_path =
+# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
+;notifiers_path =
+# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
+;plugins_path =
+
 #################################### Server ####################################
 [server]
 # Protocol (http, https, h2, socket)
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -666,22 +666,22 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 		return err
 	}
 	provisioningConfig := iniFile.Section("provisioning")
-	provisioningDatasourcesPath, err := valueAsString(provisioningConfig, "datasources_path", path.Join(provisioning, "datasources"))
+	provisioningDatasourcesPath, err := valueAsString(provisioningConfig, "datasources_path", filepath.Join(provisioning, "datasources"))
 	if err != nil {
 		return err
 	}
 	cfg.ProvisioningDatasourcesPath = makeAbsolute(provisioningDatasourcesPath, HomePath)
-	provisioningDashboardsPath, err := valueAsString(provisioningConfig, "dashboards_path", path.Join(provisioning, "dashboards"))
+	provisioningDashboardsPath, err := valueAsString(provisioningConfig, "dashboards_path", filepath.Join(provisioning, "dashboards"))
 	if err != nil {
 		return err
 	}
 	cfg.ProvisioningDashboardsPath = makeAbsolute(provisioningDashboardsPath, HomePath)
-	provisioningNotifiersPath, err := valueAsString(provisioningConfig, "notifiers_path", path.Join(provisioning, "notifiers"))
+	provisioningNotifiersPath, err := valueAsString(provisioningConfig, "notifiers_path", filepath.Join(provisioning, "notifiers"))
 	if err != nil {
 		return err
 	}
 	cfg.ProvisioningNotifiersPath = makeAbsolute(provisioningNotifiersPath, HomePath)
-	provisioningPluginsPath, err := valueAsString(provisioningConfig, "plugins_path", path.Join(provisioning, "plugins"))
+	provisioningPluginsPath, err := valueAsString(provisioningConfig, "plugins_path", filepath.Join(provisioning, "plugins"))
 	if err != nil {
 		return err
 	}
@@ -1187,7 +1187,7 @@ func (cfg *Cfg) LogConfigSources() {
 	cfg.Logger.Info("Path Data", "path", cfg.DataPath)
 	cfg.Logger.Info("Path Logs", "path", cfg.LogsPath)
 	cfg.Logger.Info("Path Plugins", "path", PluginsPath)
-	cfg.Logger.Info("Path Datasources Provisioning", "path", cfg.ProvisioningDatasourcesPath)
+	cfg.Logger.Info("Path Data Sources Provisioning", "path", cfg.ProvisioningDatasourcesPath)
 	cfg.Logger.Info("Path Dashboards Provisioning", "path", cfg.ProvisioningDashboardsPath)
 	cfg.Logger.Info("Path Notifiers Provisioning", "path", cfg.ProvisioningNotifiersPath)
 	cfg.Logger.Info("Path Plugins Provisioning", "path", cfg.ProvisioningPluginsPath)
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -239,10 +239,15 @@ type Cfg struct {
 	Packaging string
 
 	// Paths
-	ProvisioningPath   string
 	DataPath           string
 	LogsPath           string
 	BundledPluginsPath string
+
+	// Provisioning
+	ProvisioningDatasourcesPath string
+	ProvisioningDashboardsPath  string
+	ProvisioningNotifiersPath   string
+	ProvisioningPluginsPath     string
 
 	// SMTP email settings
 	Smtp SmtpSettings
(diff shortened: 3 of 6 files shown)


pkg/setting/setting.go
```diff
@@ -655,11 +660,33 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
...
 		return err
 	}
-	cfg.ProvisioningPath = makeAbsolute(provisioning, HomePath)
+	provisioningConfig := iniFile.Section("provisioning")
+	provisioningDatasourcesPath, err := valueAsString(provisioningConfig, "datasources_path", path.Join(provisioning, "datasources"))
+	if err != nil {
+		return err
+	}
+	cfg.ProvisioningDatasourcesPath = makeAbsolute(provisioningDatasourcesPath, HomePath)
+	provisioningDashboardsPath, err := valueAsString(provisioningConfig, "dashboards_path", path.Join(provisioning, "dashboards"))
```
papagian, line 674:
```suggestion
	provisioningDashboardsPath, err := valueAsString(provisioningConfig, "dashboards_path", filepath.Join(provisioning, "dashboards"))
```

pkg/setting/setting.go
```diff
@@ -655,11 +660,33 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
...
+	if err != nil {
+		return err
+	}
+	cfg.ProvisioningDashboardsPath = makeAbsolute(provisioningDashboardsPath, HomePath)
+	provisioningNotifiersPath, err := valueAsString(provisioningConfig, "notifiers_path", path.Join(provisioning, "notifiers"))
```
papagian, line 679:
```suggestion
	provisioningNotifiersPath, err := valueAsString(provisioningConfig, "notifiers_path", filepath.Join(provisioning, "notifiers"))
```

pkg/setting/setting.go
```diff
@@ -655,11 +660,33 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
...
+	if err != nil {
+		return err
+	}
+	cfg.ProvisioningNotifiersPath = makeAbsolute(provisioningNotifiersPath, HomePath)
+	provisioningPluginsPath, err := valueAsString(provisioningConfig, "plugins_path", path.Join(provisioning, "plugins"))
```
papagian, line 684:
```suggestion
	provisioningPluginsPath, err := valueAsString(provisioningConfig, "plugins_path", filepath.Join(provisioning, "plugins"))
```

papagian (dismissed):

You have to use operating system aware [`filepath`](https://golang.org/pkg/path/filepath/) instead (in case grafana is running on Windows).

Also, from the issue I understand that the goal is to have multiple provisioning paths for example for datasources at the same time. I'll discuss it further with the team and I'll come back.

conf/defaults.ini
```diff
@@ -26,6 +26,17 @@ plugins = data/plugins
 # folder that contains provisioning config files that grafana will apply on startup and while running.
 provisioning = conf/provisioning
 
+#################################### Provisioning ###############################
+[provisioning]
+# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
```
aknuds1, line 31: The style guide dictates it should be ["data source"](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-style-guide.md#data-source):
```suggestion
# folder that contains data source provisioning config files, the default is paths.provisioning/datasources
```

conf/defaults.ini
```diff
@@ -26,6 +26,17 @@ plugins = data/plugins
...
+datasources_path =
+# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
```
aknuds1, line 33:
```suggestion
# folder that contains dashboard provisioning config files, the default is paths.provisioning/dashboards
```

conf/defaults.ini
```diff
@@ -26,6 +26,17 @@ plugins = data/plugins
...
+dashboards_path =
+# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
```
aknuds1, line 35:
```suggestion
# folder that contains alert notification channel provisioning config files, the default is paths.provisioning/notifiers
```

conf/defaults.ini
```diff
@@ -26,6 +26,17 @@ plugins = data/plugins
...
+notifiers_path =
+# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
```
aknuds1, line 37:
```suggestion
# folder that contains plugin provisioning config files, the default is paths.provisioning/plugins
```

conf/sample.ini
```diff
@@ -26,6 +26,17 @@
 # folder that contains provisioning config files that grafana will apply on startup and while running.
 ;provisioning = conf/provisioning
 
+#################################### Provisioning ###############################
+[provisioning]
+# folder that contains datasource provisioning config files, default is paths.provisioning/datasources
```
aknuds1, line 31:
```suggestion
# folder that contains data source provisioning config files, the default is paths.provisioning/datasources
```

conf/sample.ini
```diff
@@ -26,6 +26,17 @@
...
+;datasources_path =
+# folder that contains dashboards provisioning config files, default is paths.provisioning/dashboards
```
aknuds1, line 33:
```suggestion
# folder that contains dashboard provisioning config files, the default is paths.provisioning/dashboards
```

conf/sample.ini
```diff
@@ -26,6 +26,17 @@
...
+;dashboards_path =
+# folder that contains alert notifications channels provisioning config files, default is paths.provisioning/notifiers
```
aknuds1, line 35:
```suggestion
# folder that contains alert notification channel provisioning config files, the default is paths.provisioning/notifiers
```

conf/sample.ini
```diff
@@ -26,6 +26,17 @@
...
+;notifiers_path =
+# folder that contains plugins provisioning config files, default is paths.provisioning/plugins
```
aknuds1, line 37:
```suggestion
# folder that contains plugin provisioning config files, the default is paths.provisioning/plugins
```

pkg/setting/setting.go
```diff
@@ -655,11 +660,33 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
...
 	PluginsPath = makeAbsolute(plugins, HomePath)
 	cfg.BundledPluginsPath = makeAbsolute("plugins-bundled", HomePath)
+
 	provisioning, err := valueAsString(iniFile.Section("paths"), "provisioning", "")
 	if err != nil {
 		return err
 	}
-	cfg.ProvisioningPath = makeAbsolute(provisioning, HomePath)
+	provisioningConfig := iniFile.Section("provisioning")
+	provisioningDatasourcesPath, err := valueAsString(provisioningConfig, "datasources_path", path.Join(provisioning, "datasources"))
```
aknuds1, line 669:
```suggestion
	provisioningDatasourcesPath, err := valueAsString(provisioningConfig, "datasources_path", filepath.Join(provisioning, "datasources"))
```

pkg/setting/setting.go
```diff
@@ -1160,7 +1187,11 @@ func (cfg *Cfg) LogConfigSources() {
 	cfg.Logger.Info("Path Data", "path", cfg.DataPath)
 	cfg.Logger.Info("Path Logs", "path", cfg.LogsPath)
 	cfg.Logger.Info("Path Plugins", "path", PluginsPath)
-	cfg.Logger.Info("Path Provisioning", "path", cfg.ProvisioningPath)
+	cfg.Logger.Info("Path Datasources Provisioning", "path", cfg.ProvisioningDatasourcesPath)
```
aknuds1, line 1190:
```suggestion
	cfg.Logger.Info("Path Data Sources Provisioning", "path", cfg.ProvisioningDatasourcesPath)
```

aknuds1 (changes requested):
Generally it looks about right to me, but I saw some things that should be fixed.

nabokihms:
Thanks for the quick review!
Filepath looks better, you are right. I have used path only because of [previous realization](https://vk.com/away.php?to=https%3A%2F%2Fgithub.com%2Fgrafana%2Fgrafana%2Fblob%2F642e9040485f803db4f8da5f5b7b61ece8cf77e9%2Fpkg%2Fservices%2Fprovisioning%2Fprovisioning.go%23L118&cc_key=).

And it seems that I also need to check the style guide one more time :)
I will fix everything according to your commentaries.

aknuds1:
@nabokihms Thanks :) I don't know for sure about the functional changes, although they _look_ fine to me, will have to consult with my colleagues and hear what they think.

aknuds1:
Do you know if there are any tests related to provisioning that could be augmented, to test your new behaviour?

nabokihms:
I think there are no specific provisioning tests to cover this case. However, I thought about adding a test case to the loading settings test [here](https://github.com/grafana/grafana/blob/master/pkg/setting/setting_test.go#L23).

It's better to cover this code by a test, so I will add it.

aknuds1:
Thanks, a test can't hurt.

pkg/setting/setting_test.go
```diff
@@ -296,6 +296,35 @@ func TestLoadingSettings(t *testing.T) {
 			So(value, ShouldEqual, "")
 		})
 	})
+
+	Convey("Test reading provisioning config", t, func() {
+		Convey("By default it should use provisioning folder setting", func() {
```
aknuds1, line 301:
```suggestion
		Convey("By default it should use the provisioning folder setting", func() {
```

pkg/setting/setting_test.go
```diff
@@ -296,6 +296,35 @@ func TestLoadingSettings(t *testing.T) {
...
+				HomePath: "../../",
+			})
+			So(err, ShouldBeNil)
+			So(cfg.ProvisioningDashboardsPath, ShouldEqual, "../../conf/provisioning/dashboards")
+			So(cfg.ProvisioningDatasourcesPath, ShouldEqual, "../../conf/provisioning/datasources")
+			So(cfg.ProvisioningNotifiersPath, ShouldEqual, "../../conf/provisioning/notifiers")
+			So(cfg.ProvisioningPluginsPath, ShouldEqual, "../../conf/provisioning/plugins")
+		})
+
+		Convey("With settings in provisioning section it should override defaults", func() {
```
aknuds1, line 313:
```suggestion
		Convey("With settings in the provisioning section it should override defaults", func() {
```

pkg/setting/setting_test.go
```diff
@@ -296,6 +296,35 @@ func TestLoadingSettings(t *testing.T) {
...
+			cfg := NewCfg()
+			err := cfg.Load(&CommandLineArgs{
+				HomePath: "../../",
+				Config:   filepath.Join(HomePath, "pkg/setting/testdata/provisioning.ini"),
+			})
+			So(err, ShouldBeNil)
+			So(cfg.ProvisioningDashboardsPath, ShouldEqual, "/etc/absolute/path")
+			So(cfg.ProvisioningDatasourcesPath, ShouldEqual, "../../not/absolute/path")
+			// There are no entries in mock provisioning config for plugins and notifiers
+			// They should equal default values
```
aknuds1, line 323:
```suggestion
			// There are no entries in the test provisioning config for plugins and notifiers
			// They should equal the defaults
```

aknuds1: Why aren't you configuring also datasources_path and dashboards_path in this test, so you can test also those settings? I think that would make more sense for a test case where you want to test non-defaults.

aknuds1 (changes requested):
Nice touch with the tests! I left some suggestions for improvement though, and I'm also wondering why the non-defaults test case only uses non-defaults for half of the settings.

pkg/setting/setting_test.go
nabokihms: I tried to cover two cases in the one test. 
1. All options are specified. 
2. Some options are specified. 
I think I will follow your suggestion because I was not 100% sure about this decision.

kylebrandt:
Should we consider making these "paths"? so you can have an array of paths for the same thing.

aknuds1:
We have discussed this internally and we think we want to wait for @bergquist to come back from vacation (beginning of August more or less) and provide his opinion on the PR.

aknuds1:
@sakjur Do you have experience with provisioning?

pkg/setting/setting_test.go
aknuds1: @nabokihms Thanks for clarifying that, I figured it was your thinking. Personally, if I wanted to test the other case (of a mixture of defaults and non-defaults), I would make second test case. I don't think I want to ask you to do it though, since it should work and the PR already has good test coverage :)

pkg/setting/setting_test.go
```diff
@@ -296,6 +296,33 @@ func TestLoadingSettings(t *testing.T) {
 			So(value, ShouldEqual, "")
 		})
 	})
+
+	Convey("Test reading provisioning config", t, func() {
```
aknuds1, line 300: I thought about your use of `HomePath`, and figure it's probably a better idea to not use the global `HomePath` variable (we want to get rid of the globals in any case). I also think it's wise to use an absolute path instead of "../..", since it can be really helpful when debugging failures (a relative path leaves you wondering what it actually resolves to).

What do you think of using a `homePath` variable as follows in the tests?
```suggestion
	Convey("Test reading provisioning config", t, func() {
			homePath, err := filepath.Abs("../..")
			So(err, ShouldBeNil)

```

aknuds1:
Adding @ryantxu as a reviewer as I think he might have some insight to share.

sakjur:
@aknuds1 I don't, but this looks really nice 😄

ryantxu:
It makes more sense to me if we just changed the existing config to be an array -- however it looks like there is a special type of loader for each type that assumes a folder structure.

A better solution would be similar to pluin scanning:
1. walk all configured folders
2. when you find a provisioning yaml, register it as somethign from that type
3. actually run the provisioning in the required order (all datasources first etc)

The approach suggested here is the easiest given the current implementation, but I'm not sure that is where we want to be eventually.  Thoughts @bergquist ?

pkg/setting/setting_test.go
nabokihms: For me, getting rid of `HomePath` looks a bit off this PR subject. Tell me whether I should do it here or open another PR.

However, I have no objections to correct the test.

aknuds1: @nabokihms Sorry, didn't mean for you to get rid of HomePath in general, I was just referring to your tests :) I saw they could be written without using the global Homepath.

aknuds1 (approved):

aknuds1 (approved):
Approving as the code looks good to me, but we need a round of internal discussion before we can decide if this is the right solution or not.

aknuds1 (approved):
Looks as if GitHub duplicated my approval...

torkelo:
This does not look backward compatible

Close to zero demand for the feature request (#24631) to warrant the complexity of these many new settings. 

And it does not look to solve the main request in #24631 , having multiple paths so you can have internal (prepackaged) provisioned files and user provisioned files

aknuds1:
@torkelo Might sound like what @kylebrandt and @ryantxu also wanted, i.e. arrays instead of single provisioning paths.

bergquist:
> Close to zero demand for the feature request (#24631) to warrant the complexity of these many new settings.

I think this implementation makes sense if we want to support multiple paths. But I agree with Torkel about not introducing more settings unless we have to. Convention over configuration.

torkelo:
Closing this PR as there seems to be little demand/need for this option

PR #26671, opened 2020-07-29
Alerting: Remove dashboard dependency (WIP)

papagian:
It modifies implementation where alert rules are expected to have one dashboard to support 0 or more associated dashboards.
- It modifies `/api/alerts` (GetAlerts API) to enlist standalone alerts when they are enabled.

Fixes #

### DIFF
--- a/pkg/services/sqlstore/alert.go
+++ b/pkg/services/sqlstore/alert.go
@@ -83,7 +83,11 @@ func deleteAlertByIdInternal(alertId int64, reason string, sess *DBSession) erro
 func HandleAlertsQuery(query *models.GetAlertsQuery) error {
 	builder := SqlBuilder{}
 
-	builder.Write(`SELECT
+	joinType := "INNER"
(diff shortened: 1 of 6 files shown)


pkg/services/sqlstore/alert.go
```diff
@@ -83,7 +83,11 @@ func deleteAlertByIdInternal(alertId int64, reason string, sess *DBSession) erro
 func HandleAlertsQuery(query *models.GetAlertsQuery) error {
 	builder := SqlBuilder{}
 
-	builder.Write(`SELECT
+	joinType := "INNER"
+	if query.StandaloneAlertsEnabled {
+		joinType = "LEFT OUTER"
```
torkelo, line 88: Is this logic really needed? can't we always do an outer join?

papagian: I guess we can. I differentiated only for keeping the existing behaviour as is and not fetching additional rows if there are not needed.

PR #26673, opened 2020-07-29
Chore: use Wrap for all admin API

AgnesToulet:
Standardize the API to use the `Wrap` function for all `/api/admin` endpoints.

Having a standardized API makes it easier to add feature across the whole API like auditing.

### DIFF
--- a/pkg/api/admin.go
+++ b/pkg/api/admin.go
@@ -9,7 +9,7 @@ import (
 	"github.com/grafana/grafana/pkg/setting"
 )
 
-func AdminGetSettings(c *models.ReqContext) {
+func AdminGetSettings(c *models.ReqContext) Response {
 	settings := make(map[string]interface{})
 
 	for _, section := range setting.Raw.Sections() {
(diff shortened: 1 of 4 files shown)


kylebrandt:
I think ... *maybe* ... the test is failing because of `ServerError(err)`?

```
func Wrap(action interface{}) macaron.Handler {
	return func(c *models.ReqContext) {
		var res Response
		val, err := c.Invoke(action)
		if err == nil && val != nil && len(val) > 0 {
			res = val[0].Interface().(Response)
		} else {
			res = ServerError(err)
		}

		res.WriteTo(c)
	}
}
```

```
var (
	NotFound = func() Response {
		return Error(404, "Not found", nil)
	}
	ServerError = func(err error) Response {
		return Error(500, "Server error", err)
	}
)
```

aknuds1:
@kylebrandt Interesting. At least the tests are doing their job :D

aknuds1:
@AgnesToulet Can you look into why the tests are failing?

aknuds1 (changes requested):
Tests need fixing.

aknuds1 (approved):
Looks great AFAICT.

PR #26682, opened 2020-07-29
Inspect: Fix link to error tab from panel header

### LINKED ISSUE #26665: Clicking on red triangle error does not open the error
When I click on the red error triangle, it opens the inspector, but not to the error tab.

The error is in the url:
image

torkelo:
Fixes #26665


ryantxu (approved):
thank you!

PR #26696, opened 2020-07-30
ThemeContext: Fix useStyles return type

s-h-a-d-o-w:
Used a type override in my project, so I didn't notice this. :facepalm: 

Autocomplete still didn't work correctly as originally intended [here](https://github.com/grafana/grafana/pull/26056) because `useStyles` still returned `any`.


torkelo (approved):
Looks good!

PR #26698, opened 2020-07-30
Explore: Adds query inspector drawer to explore

kaydelaney:
Adds the query inspector drawer available in dashboards to Explore, letting a user see stats related to their queries as well as investigating the request and response JSON.

### DIFF
--- a/packages/grafana-ui/src/components/TabbedContainer/TabbedContainer.tsx
+++ b/packages/grafana-ui/src/components/TabbedContainer/TabbedContainer.tsx
@@ -2,8 +2,8 @@ import React, { useState } from 'react';
 import { css } from 'emotion';
 
 import { SelectableValue, GrafanaTheme } from '@grafana/data';
-import { stylesFactory, withTheme } from '../../themes';
-import { IconName, Themeable, TabsBar, Tab, IconButton, CustomScrollbar, TabContent } from '../..';
+import { stylesFactory, useTheme } from '../../themes';
+import { IconName, TabsBar, Tab, IconButton, CustomScrollbar, TabContent } from '../..';
 
 export interface TabConfig {
   label: string;
@@ -12,7 +12,7 @@ export interface TabConfig {
   icon: IconName;
 }
 
-export interface TabbedContainerProps extends Themeable {
+export interface TabbedContainerProps {
   tabs: TabConfig[];
   defaultTab?: string;
   closeIconTooltip?: string;
@@ -49,7 +49,7 @@ const getStyles = stylesFactory((theme: GrafanaTheme) => {
   };
 });
 
-function UnThemedTabbedContainer(props: TabbedContainerProps) {
+export function TabbedContainer(props: TabbedContainerProps) {
   const [activeTab, setActiveTab] = useState(
     props.tabs.some(tab => tab.value === props.defaultTab) ? props.defaultTab : props.tabs?.[0].value
   );
@@ -58,7 +58,8 @@ function UnThemedTabbedContainer(props: TabbedContainerProps) {
     setActiveTab(item.value!);
   };
 
-  const { tabs, theme, onClose, closeIconTooltip } = props;
+  const { tabs, onClose, closeIconTooltip } = props;
+  const theme = useTheme();
   const styles = getStyles(theme);
 
   return (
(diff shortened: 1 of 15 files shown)


packages/grafana-ui/src/components/TabbedContainer/TabbedContainer.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <IconButton className={styles.close} onClick={onClose} name="times" title={closeIconTooltip ?? 'Close'} />
+      </TabsBar>
+      <CustomScrollbar className={styles.scrollbar}>
+        <TabContent className={styles.tabContent}>{tabs.find(t => t.value === activeTab)?.content}</TabContent>
+      </CustomScrollbar>
+    </div>
+  );
+}
+
+export const TabbedContainer = withTheme(UnThemedTabbedContainer);
```
torkelo, line 85: Since it's a functional component can't use use useTheme hook? instead of withTheme

public/app/features/explore/Explore.tsx
```diff
@@ -281,6 +284,14 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
     });
   };
 
+  toggleShowQueryInspector = () => {
```
ivanahuckova, line 287: Are we thinking about either:
- closing history tab when inspector is opened and vice versa
- moving one over the other depending on which was open more recently

packages/grafana-ui/src/components/TabbedContainer/TabbedContainer.tsx
kaydelaney: You're right, thanks for pointing this out

ivanahuckova (commented):
Amazing job Kay! 🚀 I've added 1 little comment/question. And 1 more question, are there going to be more tabs as in panel inspect (e.g. error, data)?

public/app/features/explore/Explore.tsx
kaydelaney: That's one of the remaining questions for sure. On the one hand showing the history drawer closing and the inspector drawer opening (or vice-versa) would be more clear in showing what's happening, but on the other hand I worry that such a sequence of animations would be too much or take too long.
The trouble with just showing, say, the inspector drawer opening over the history drawer is that it's not exactly clear that the history drawer is then closed, so I worry about that a bit

hugohaggmark:
Outstanding work @kaydelaney ⭐  
I was testing this out and wonder on thing. How is the QueryInspector supposed to work in split view? Is it supposed to specific to the split or for the entire Explore?

kaydelaney:
> I was testing this out and wonder on thing. How is the QueryInspector supposed to work in split view? Is it supposed to specific to the split or for the entire Explore?
Really good question! I suspect from a user perspective having the inspector reflect the queries specific to the split they're in would be preferable. Any thoughts on this @davkal @torkelo ?

torkelo:
To me this could cover both splits but I do not have a strong opinion

public/app/features/explore/Explore.tsx
ivanahuckova: Yeah, I agree. I have tested it and I think it looks good as you have implemented it now. Also, I don't think that users will be switching between these two very often.

ivanahuckova:
> Really good question! I suspect from a user perspective having the inspector reflect the queries specific to the split they're in would be preferable. 

I totally agree with this as well. I think that panes are pretty much separate views, they can have separate data sources, queries, ... and therefore it makes sense to have separate query inspectors for both. Also this is consistent with how rich history works. I also really like, that you can have for example query inspector on first pane and query history on second.
image

ivanahuckova (approved):
Amazing job and very useful feature for Explore! I really like your work around creating reusable components that are used in panel inspect and rich history. ✅ Tests coverage. LGTM 🕺

torkelo:
Added a followup issue to address the duplication here: 
https://github.com/grafana/grafana/issues/26893

torkelo:
Feel this was merged too quickly, at least without directly starting unifying the query inspectors https://github.com/grafana/grafana/issues/26893 
As we now have to identical query inspectors that we have to maintain

PR #26701, opened 2020-07-30
Fix language in backend.md

aknuds1:
Just a tiny language fix in the back-end style guide doc.

### DIFF
--- a/contribute/style-guides/backend.md
+++ b/contribute/style-guides/backend.md
@@ -1,6 +1,6 @@
 # Backend style guide
 
-Grafanas backend has been developed for a long time with a mix of code styles.  This guide explains how we want to write Go code in the future.
+Grafana's backend has been developed for a long time with a mix of code styles.  This guide explains how we want to write Go code in the future.
 
 Unless stated otherwise, use the guidelines listed in the following articles:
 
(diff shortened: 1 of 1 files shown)


contribute/style-guides/backend.md
```diff
@@ -1,6 +1,6 @@
 # Backend style guide
 
-Grafanas backend has been developed for a long time with a mix of code styles.  This guide explains how we want to write Go code in the future.
+Grafana's backend has been developed for a long time with a mix of code styles.  This guide explains how we want to write Go code in the future.
```
oddlittlebird, line 3:
```suggestion
Grafana's backend has been developed for a long time with a mix of code styles. This guide explains how we want to write Go code in the future.
```

oddlittlebird (approved):
Good catch! Added one minor edit.

contribute/style-guides/backend.md
aknuds1: Thanks @oddlittlebird, I also noticed the two spaces, but wasn't sure if we have standardized on a single space after a period. Can we add to the style guide that there should be a single space after a period? I know people come from different traditions on this, so it would be great to standardize.

PR #26702, opened 2020-07-30
Convert CloudWatch to use dataframes

aknuds1:
Convert CloudWatch to use data frames.


pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -95,62 +103,84 @@ func parseGetMetricDataTimeSeries(metricDataResults map[string]*cloudwatch.Metri
...
-					float64(t.Unix())*1000))
+				val := metricDataResult.Values[j]
+				plog.Debug("Handling timestamp", "timestamp", t, "value", *val)
+				timestamps = append(timestamps, float64(t.Unix()*1000))
+				points = append(points, val)
+			}
+
+			fields := []*data.Field{
+				data.NewField("timestamp", nil, timestamps),
+				data.NewField("value", nil, points),
```
kylebrandt, line 168: I think maybe tags should be a data.Labels{}, and passed as the second argument to the value field. Part of the change to dataframes is that labels are actually part of our data model now.

aknuds1: @kylebrandt Aha, thanks, I was totally unsure about the tags.

kylebrandt (approved):

PR #26703, opened 2020-07-30
Grafana-UI: Make TimeOfDayPicker value optional

Clarity-89:
Make `TimeOfDayPicker` value optional so it can be used inside `InputControl` which doesn't pass explicit `value` for controlled components.


AgnesToulet (approved):
LGTM!

PR #26716, opened 2020-07-31
OAuth: Add some debug logs

aknuds1:
Add some debug logging to OAuth syncing logic, so it's easier to diagnose what's happening. Also use `errors.Is`, instead of checking for direct error equality.

### DIFF
--- a/pkg/services/login/login.go
+++ b/pkg/services/login/login.go
@@ -207,7 +207,7 @@ func syncOrgRoles(user *models.User, extUser *models.ExternalUserInfo) error {
 			deleteOrgIds = append(deleteOrgIds, org.OrgId)
 		} else if extRole != org.Role {
 			// update role
-			cmd := &models.UpdateOrgUserCommand{OrgId: org.OrgId, UserId: user.Id, Role: extUser.OrgRoles[org.OrgId]}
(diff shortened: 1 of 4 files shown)


pkg/services/login/login.go
```diff
@@ -197,9 +202,10 @@ func syncOrgRoles(user *models.User, extUser *models.ExternalUserInfo) error {
 	for _, org := range orgsQuery.Result {
 		handledOrgIds[org.OrgId] = true
 
-		if extUser.OrgRoles[org.OrgId] == "" {
+		extRole := extUser.OrgRoles[org.OrgId]
+		if extRole == "" {
 			deleteOrgIds = append(deleteOrgIds, org.OrgId)
-		} else if extUser.OrgRoles[org.OrgId] != org.Role {
+		} else if extRole != org.Role {
```
papagian, line 208: minor remark: you can use the same variable `extRole` for constructing the `cmd` [below](https://github.com/grafana/grafana/pull/26716/files#diff-6db13569d0b1313cbb06a85d465729eaR210)

papagian (approved):
LGTM

pkg/services/login/login.go
aknuds1: Thanks @papagian - having a look.

aknuds1: Done.

PR #26718, opened 2020-07-31
Chore: Drop xerrors dependency

aknuds1:
Drop dependency on legacy [xerrors](https://pkg.go.dev/golang.org/x/xerrors?tab=doc) Go module, since it's replaced by standard library functionality (in errors and fmt packages). Also clean up some error strings according to Go standards (no capitalization or punctuation).


jessabe (approved):
Are we moving towards lowercase and not sentencing? Is this a style choice? :)

aknuds1:
@jessabe You mean not capitalizing error strings? They shouldn't be capitalized, as per official Go [style](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

jessabe:
> @jessabe You mean not capitalizing error strings? They shouldn't be capitalized, as per official Go [style](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

👍

pkg/cmd/grafana-cli/utils/grafana_path.go
```diff
@@ -23,7 +23,7 @@ func GetGrafanaPluginDir(currentOS string) string {
 func getGrafanaRoot() (string, error) {
 	ex, err := os.Executable()
 	if err != nil {
-		return "", xerrors.New("Failed to get executable path")
+		return "", errutil.Wrap("failed to get executable path", err)
```
papagian, line 26: It is already checked that the `err != nil` so I think there is no reason to use `errutil.Wrap`; `fmt.Errorf` should be enough.

aknuds1: @papagian I think there must be a misunderstanding. `errutil.Wrap` doesn't do any checking, it wraps `err` with another message ("failed to get executable path"). Now, we can wrap also with `fmt.Errorf`, but so far we have been using `errutil.Wrap` in Grafana code. Shall we stop using `errutil,.Wrap`?

pkg/util/errutil/errors.go
```diff
@@ -2,23 +2,18 @@ package errutil
...
 	"fmt"
-
-	"golang.org/x/xerrors"
 )
 
-// Wrap is a simple wrapper around Errorf that is doing error wrapping. You can read how that works in
-// https://godoc.org/golang.org/x/xerrors#Errorf but its API is very implicit which is a reason for this wrapper.
-// There is also a discussion (https://github.com/golang/go/issues/29934) where many comments make arguments for such
-// wrapper so hopefully it will be added in the standard lib later.
+// Wrap is a simple wrapper around fmt.Errorf that wraps errors.
```
papagian, line 7: I would suggest keeping the reference to the golang issue so if such a wrapper is introduced at some point we can remove this one.

aknuds1: @papagian The wrapper is introduced actually, it's `fmt.Errorf("wrapping message: %w", err)`. I am open to replacing `errutil.Wrap*`, but I got the impression so far we're keeping it. Maybe we should schedule a team discussion wrt. potentially replacing it?

pkg/cmd/grafana-cli/utils/grafana_path.go
papagian: It's not this [one](https://github.com/grafana/grafana/blob/f8d4d3fd07bd78003445908c484a012d82bf6b58/pkg/util/errutil/errors.go#L8)?

aknuds1: @papagian Yes, but I don't see the relevance? Why exactly do you think wrapping isn't necessary?

aknuds1: Ah, you're referring to that `errutil.Wrap` checks if err is `nil` at the beginning, but that isn't its real functionality.

pkg/util/errutil/errors.go
papagian: ok, then I don't see why not removing it but I'm open to discuss it with the rest of the team.

aknuds1: @papagian Shall we try to have a discussion when everyone is back from vacation then?

papagian: It sounds good

pkg/cmd/grafana-cli/utils/grafana_path.go
papagian: yes, my point is that this could be replaced with:
`fmt.Errorf("failed to get executable path: %w", err)`

So actually it's the same discussion whether there is need to keep `errutil.Wrap` at all.

papagian (approved):

PR #26724, opened 2020-07-31
Tests: Change some GoConvey tests to use testify

sakjur:

> Tests must use the standard library, testing. For assertions, prefer using testify.
>
> The majority of our tests uses GoConvey but that's something we want to avoid going forward.

from https://github.com/grafana/grafana/blob/master/contribute/style-guides/backend.md#testing

Getting rid of GoConvey would be a nice long term goal. I'll try to pick a few "easy targets" where we don't use the advanced functionality of GoConvey to convert to Testify based tests. I'm also making sure we're not ignoring errors and other general quality of life improvements to tests. Will structure PR so that when I move it out of draft each package will be a single commit (or a sequence of commits, if more appropriate) to make it easier to review.

### DIFF
--- a/pkg/util/url_test.go
+++ b/pkg/util/url_test.go
@@ -4,69 +4,43 @@ import (
 	"net/url"
 	"testing"
 
-	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestUrl(t *testing.T) {
-	Convey("When joining two urls where right hand side is empty", t, func() {
-		result := JoinURLFragments("http://localhost:8080", "")
-
-		So(result, ShouldEqual, "http://localhost:8080")
-	})
-
-	Convey("When joining two urls where right hand side is empty and lefthand side has a trailing slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080/", "")
-
-		So(result, ShouldEqual, "http://localhost:8080/")
-	})
-
-	Convey("When joining two urls where neither has a trailing slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080", "api")
-
-		So(result, ShouldEqual, "http://localhost:8080/api")
-	})
-
-	Convey("When joining two urls where lefthand side has a trailing slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080/", "api")
-
-		So(result, ShouldEqual, "http://localhost:8080/api")
-	})
-
-	Convey("When joining two urls where righthand side has preceding slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080", "/api")
-
-		So(result, ShouldEqual, "http://localhost:8080/api")
-	})
-
-	Convey("When joining two urls where righthand side has trailing slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080", "api/")
-
-		So(result, ShouldEqual, "http://localhost:8080/api/")
-	})
-
-	Convey("When joining two urls where lefthand side has a trailing slash and righthand side has preceding slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080/", "/api/")
-
-		So(result, ShouldEqual, "http://localhost:8080/api/")
-	})
+	tests := []struct {
+		base     string
+		path     string
+		expected string
+	}{
+		{base: "http://localhost:8080", path: "", expected: "http://localhost:8080"},
+		{base: "http://localhost:8080/", path: "", expected: "http://localhost:8080/"},
+		{base: "http://localhost:8080", path: "api", expected: "http://localhost:8080/api"},
+		{base: "http://localhost:8080/", path: "api", expected: "http://localhost:8080/api"},
+		{base: "http://localhost:8080/", path: "api/", expected: "http://localhost:8080/api/"},
+		{base: "http://localhost:8080/", path: "/api/", expected: "http://localhost:8080/api/"},
+	}
+	for _, testcase := range tests {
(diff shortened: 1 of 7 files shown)


aknuds1 (commented):
Looks really good so far, thanks!

sakjur:
I tried to add some other tests to this batch, but they were complex so I'll take those separately 🙂

pkg/util/url_test.go
```diff
@@ -4,69 +4,43 @@ import (
...
-		So(result, ShouldEqual, "http://localhost:8080/api/")
-	})
-
-	Convey("When joining two urls where lefthand side has a trailing slash and righthand side has preceding slash", t, func() {
-		result := JoinURLFragments("http://localhost:8080/", "/api/")
-
-		So(result, ShouldEqual, "http://localhost:8080/api/")
-	})
+	tests := []struct {
+		base     string
```
papagian, line 13: I would suggest adding a description field and wrapping each assertion [below](https://github.com/grafana/grafana/blob/6760a43ff9ffedea95bd39bd8ce696c8df49a4b4/pkg/util/url_test.go#L25) in a separate `t.Run` block.

pkg/util/validation_test.go
```diff
@@ -3,19 +3,15 @@ package util
...
-		result := IsEmail("abcdef.com")
-
-		So(result, ShouldEqual, false)
-	})
+	emails := map[string]bool{
+		"abc@def.com": true,
+		"abcdef.com":  false,
+	}
+	for input, expected := range emails {
+		assert.Equal(t, expected, IsEmail(input))
```
papagian, line 15: I would like to keep the test descriptions; I would suggest wrapping in a `t.Run` block

papagian (approved):
Looks good! I've two small suggestions for improvement.

sakjur:
@papagian Thank you, I was on the fence about keeping the descriptions or not. I'll add them in 🙂

aknuds1:
Having a look.

pkg/util/url_test.go
aknuds1: Agree with @papagian to use `t.Run`.

pkg/util/validation_test.go
aknuds1: Agree with using `t.Run`.

aknuds1 (approved):
Thanks for taking care of this!! LGTM, but see comments.

pkg/util/validation_test.go
sakjur: I added back some test descriptions, they're not _exactly_ as they were (mostly because GoConvey neatly shows descriptions and the Go test runner insists on formatting them with `_` instead of ` `).

I also added a few more test cases here because it felt relevant.

pkg/util/url_test.go
sakjur: ✅ The description are not verbatim exactly as they were (see https://github.com/grafana/grafana/pull/26724#discussion_r469344449 ) but they should say the same thing. I also changed the name of the test. The output ends up being

```
    --- PASS: TestJoinURLFragments/where_RHS_is_empty (0.00s)
    --- PASS: TestJoinURLFragments/where_RHS_is_empty_and_LHS_has_trailing_slash (0.00s)
    --- PASS: TestJoinURLFragments/where_neither_has_trailing_slash (0.00s)
    --- PASS: TestJoinURLFragments/where_LHS_has_trailing_slash (0.00s)
    --- PASS: TestJoinURLFragments/where_LHS_and_RHS_has_trailing_slash (0.00s)
    --- PASS: TestJoinURLFragments/where_LHS_has_trailing_slash_and_RHS_has_preceding_slash (0.00s)
```

which combines brevity and readability imo

pkg/util/url_test.go
```diff
@@ -8,28 +8,63 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestUrl(t *testing.T) {
+func TestJoinURLFragments(t *testing.T) {
+	t.Parallel()
```
aknuds1, line 12: What's the exact benefit of calling `t.Parallel`? I haven't seen this in action before, so could use some insight into why it's applied here.

aknuds1 (approved):
LGTM, just wondering about the use of t.Parallel.

pkg/util/url_test.go
sakjur: @aknuds1 It tells the Go test runner that this test is safe to run in parallel with other tests also marked with `t.Parallel`. In theory, if we did this for all tests which are truly units (i.e. testing pure functions or contained environments) we should be able to run the tests several times faster.

aknuds1: @sakjur Interesting. AFAICT sub-tests are also run in parallel (in their own goroutines). Will Go run all (non-sub-)tests in the test suite in serial, unless you use `t.Parallel`?

papagian (approved):

pkg/util/url_test.go
sakjur: Yes! I chose not to do that optimization here since these validations are nearly instantaneous anyways so running the top in parallel *should* net similar benefits with less overhead, but they are serial by default.

sakjur: That is to say sub-tests run serially by default, but in their own Goroutines. There's a great post about this in the Go blog, I'll try to dig it up

sakjur: https://blog.golang.org/subtests

I think my subtests should be made parallel _as well_ for this to really work 🤔

aknuds1: Thanks @sakjur - good reading. So sub-tests will run in serial in relation to each other, even if they are in separate goroutines? I'm slightly confused.

PR #26732, opened 2020-07-31
CloudWatch Logs: Move query response stats to appropriate FrameMeta property

kaydelaney:
Moves query stats out of `meta.custom` into `meta.stats` so that they show up in the query inspector.

### DIFF
--- a/pkg/tsdb/cloudwatch/log_query.go
+++ b/pkg/tsdb/cloudwatch/log_query.go
@@ -95,24 +95,45 @@ func logsResultsToDataframes(response *cloudwatchlogs.GetQueryResultsOutput) (*d
 	}
 
 	frame := data.NewFrame("CloudWatchLogsResponse", newFields...)
+
+	bytesScanned := -1.0
+	recordsScanned := -1.0
+	recordsMatched := -1.0
+	if response != nil && response.Statistics != nil {
+		if response.Statistics.BytesScanned != nil {
+			bytesScanned = *response.Statistics.BytesScanned
+		}
+
+		if response.Statistics.RecordsScanned != nil {
+			recordsScanned = *response.Statistics.RecordsScanned
+		}
+
+		if response.Statistics.RecordsMatched != nil {
+			recordsMatched = *response.Statistics.RecordsMatched
+		}
+	}
+
 	frame.Meta = &data.FrameMeta{
 		Stats: []data.QueryStat{
 			{
 				FieldConfig: data.FieldConfig{DisplayName: "Bytes scanned"},
-				Value:       *response.Statistics.BytesScanned,
+				Value:       bytesScanned,
 			},
 			{
 				FieldConfig: data.FieldConfig{DisplayName: "Records scanned"},
-				Value:       *response.Statistics.RecordsScanned,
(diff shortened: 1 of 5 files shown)


pkg/tsdb/cloudwatch/log_query.go
```diff
@@ -96,9 +96,22 @@ func logsResultsToDataframes(response *cloudwatchlogs.GetQueryResultsOutput) (*d
 
 	frame := data.NewFrame("CloudWatchLogsResponse", newFields...)
 	frame.Meta = &data.FrameMeta{
+		Stats: []data.QueryStat{
+			{
+				FieldConfig: data.FieldConfig{DisplayName: "Bytes scanned"},
+				Value:       *response.Statistics.BytesScanned,
```
kylebrandt, line 102: If any of these floats are nil, then de-referencing them will cause a panic.  So it needs to be checked that that value is not nil first.

Other than that looks good to me.

kylebrandt (approved):
1 Change requested to avoid possible panic, other than that LGTM.

pkg/tsdb/cloudwatch/log_query.go
aknuds1: Agree with Kyle.

aknuds1 (commented):
LGTM as well except I agree with @kylebrandt that those pointers should be nil-checked before they are dereferenced.

pkg/tsdb/cloudwatch/log_query.go
```diff
@@ -95,24 +95,45 @@ func logsResultsToDataframes(response *cloudwatchlogs.GetQueryResultsOutput) (*d
 	}
 
 	frame := data.NewFrame("CloudWatchLogsResponse", newFields...)
+
+	bytesScanned := -1.0
+	recordsScanned := -1.0
+	recordsMatched := -1.0
```
aknuds1, line 101: Why -1.0 as the default, as opposed to 0? I don't know the answer, asking as I'd like to find out :)

kaydelaney: I suppose my thinking was that a negative number makes it clear it's not a "valid" value to a user, but if this breaks some established convention I can change it to 0.

kylebrandt: If the statistic is not present, it should be omitted according to https://github.com/grafana/grafana-plugin-sdk-go/pull/191#discussion_r450278962

Or we could making a breaking SDK change to the QueryStat struct and make it a pointer (Which I am not against). The behavior as viewed by the frontend could remain the same, because this is in json so we could omitempty (which would omit when nil, but allow 0).

aknuds1 (approved):
LGTM, although you might want to do as @kylebrandt suggests regarding statistic default values.

kaydelaney:
> LGTM, although you might want to do as @kylebrandt suggests regarding statistic default values.

You mean the breaking SDK change?

aknuds1:
@kaydelaney No, I meant

> If the statistic is not present, it should be omitted according to grafana/grafana-plugin-sdk-go#191 (comment)

PR #26735, opened 2020-07-31
Flux: Ensure connections to InfluxDB are closed

### LINKED ISSUE #26731: Flux: Connections to the InfluxDB Flux datasource are never closed
**What happened**:

When using the InfluxDB datasource with Flux, anytime I refresh dashboard  page with uses this datasource, or move between dashboards, a handful of new TCP connections are opened between Grafana and InfluxDB, the stay in ESTABLISHED state and never close unless either InfluxDB or Grafana are restarted.  Over time this can lead to an exhaustion of file handles.

I would expect either for connections to be reused, or for them to be closed after use.

1) Create a datasource which connects to InfluxDB using Flux.
2) Create a dashboard which queries the datasource.
3) Monitor the number of connections on the InfluxDB server

    netstat -atn | grep ":8086"
    netstat -atn | grep ":8086" | wc -l

4) Refresh the dashboard page, or navigate between 2 dashboards.  The number of ESTABLISHED TCP connections will constantly increase and never decrease

I compared the code for InfluxQL compared to the Flux and I can see that with InfluxQL the connection is always closed, but there is no Close() method called in the Flux code (except in the test).

Under - /grafana/pkg/tsdb/influxdb
[root@apps3 influxdb]# grep -Rn "Close()" *
flux/executor_test.go:47:	defer server.Close()
influxdb.go:86:	defer resp.Body.Close()

I am not a Go programmer, or familiar with this codebase, but if I add the following to the Query function, it seems to resolve the issue.

defer runner.client.Close()

Happy to submit the pull request if this seems sensible.

**Environment**:
- Grafana version: 7.1.1
- Data source type & version: InfluxDB (build in)
- OS Grafana is installed on: CentOS 7.8
- User OS & Browser: MacOS 10.15
- Grafana plugins: Just the built in plugins
- Others:

paul-sneddon:
What this PR does / why we need it:
Ensures TCP connections created between the Grafana Flux datasource and InFluxDB are closed after they have been used.

Which issue(s) this PR fixes:
Fixes grafana#26731

Special notes for your reviewer:

### DIFF
--- a/pkg/tsdb/influxdb/flux/flux.go
+++ b/pkg/tsdb/influxdb/flux/flux.go
@@ -41,6 +41,7 @@ func Query(ctx context.Context, dsInfo *models.DataSource, tsdbQuery *tsdb.TsdbQ
 
 		tRes.Results[query.RefId] = backendDataResponseToTSDBResponse(&res, query.RefId)
 	}
+	defer runner.client.Close()
 	return tRes, nil
 }
 
--- a/pkg/tsdb/influxdb/flux/flux.go
+++ b/pkg/tsdb/influxdb/flux/flux.go
@@ -30,7 +30,7 @@ func Query(ctx context.Context, dsInfo *models.DataSource, tsdbQuery *tsdb.TsdbQ
 		return nil, err
 	}
 	defer runner.client.Close()
-	
+
 	for _, query := range tsdbQuery.Queries {
 		qm, err := GetQueryModelTSDB(query, tsdbQuery.TimeRange, dsInfo)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/tsdb/influxdb/flux/flux.go
```diff
@@ -41,6 +41,7 @@ func Query(ctx context.Context, dsInfo *models.DataSource, tsdbQuery *tsdb.TsdbQ
 
 		tRes.Results[query.RefId] = backendDataResponseToTSDBResponse(&res, query.RefId)
 	}
+	defer runner.client.Close()
```
aknuds1, line 44: Please move this to line 32.

aknuds1 (changes requested):
Looks mostly good to me, a minor adjustment is needed though.

aknuds1 (approved):
Thanks, nice catch!

aknuds1:

PR #26738, opened 2020-07-31
Docs: Organize variables

oddlittlebird:
This PR organizes variable content a bit.

- Moved the variable types into a table
- Moved some of the variable topics into a subfolder
- Updated links to reflect new locations
- Added aliases to topics that moved

@scoren-gl , something for you to be aware of

### DIFF
--- a/docs/sources/menu.yaml
+++ b/docs/sources/menu.yaml
@@ -256,26 +256,27 @@
     - link: /variables/variable-examples/
       name: Variable examples
     - name: Variable types
-      link: /variables/variable-types/  
-      - link: /variables/add-query-variable/
+      link: /variables/variable-types/
+      children:
+      - link: /variables/variable-types/add-query-variable/
         name: Add query variable
-      - link: /variables/add-custom-variable/
+      - link: /variables/variable-types/add-custom-variable/
         name: Add custom variable
-      - link: /variables/add-text-box-variable/
+      - link: /variables/variable-types/add-text-box-variable/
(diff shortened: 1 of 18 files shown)


docs/sources/menu.yaml
```diff
@@ -255,28 +255,33 @@
...
-      name: Add data source variable
-    - link: /variables/add-interval-variable/
-      name: Add interval variable
-    - link: /variables/add-ad-hoc-filters/
-      name: Add ad hoc filters
+    - name: Variable types
+      link: /variables/variable-types/
+      children:
+      - link: /variables/variable-types/add-query-variable/
+        name: Add query variable
```
torkelo, line 262: Do we need Add in the name for every page here? I can get it's ok to have it in the page title for SEO reasons but in the nav it feels a bit superfluous

torkelo (approved):
Looks good! 

only think the nav menu wording could be simplified

docs/sources/menu.yaml
oddlittlebird: Yes, we do. We don't know which variable a user wants to know about, so we want to make this content as discoverable as possible for the user.

oddlittlebird: @scoren-gl will let me know if I am wrong. :)

oddlittlebird: Also, the verb in the title signals to the reader that this is a TASK topic with instructions on HOW to do the task, in this case adding the variable.

PR #26742, opened 2020-07-31
Provisioning: Improve validation by validating across all dashboard providers

### LINKED ISSUE #22777: Multiple dashboards with the same UID in different folders creates race condition
Given there are two dashboards with the same `uid` Grafana will first create one dashboards but then update the same dashboards in the dashboard with the second dashboard json file from disk. Each time the provisioning feature runs it will flipp between the dashboard json files and delete/insert one of them into Grafanas database. 

This behavior is quite hard to understand for any operator who is not an expert on provisioning Grafana dashboards and something we need to help operators to identify. Given how unusually this is I think error logging should be enough.

nabokihms:
This PR changes the behavior of the sanity checker. Previously, we threw warnings only about dashboards provisioned by the same provider. Now we tell users about problems across all providers.

Fixes #22777

I have also added a test for the sanity checker because the test can't harm :)

### DIFF
--- a/pkg/services/provisioning/dashboards/config_reader.go
+++ b/pkg/services/provisioning/dashboards/config_reader.go
@@ -110,7 +110,7 @@ func (cr *configReader) readConfig() ([]*config, error) {
 
 	for uid, times := range uidUsage {
 		if times > 1 {
-			cr.log.Error("the same folder UID is used more than once", "folderUid", uid)
+			cr.log.Error("the same folderUID is used more than once", "folderUID", uid)
 		}
 	}
 
--- a/pkg/services/provisioning/dashboards/config_reader.go
+++ b/pkg/services/provisioning/dashboards/config_reader.go
@@ -110,7 +110,7 @@ func (cr *configReader) readConfig() ([]*config, error) {
 
 	for uid, times := range uidUsage {
 		if times > 1 {
-			cr.log.Error("the same folderUID is used more than once", "folderUID", uid)
+			cr.log.Error("the same folder UID is used more than once", "folderUid", uid)
 		}
 	}
 
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -27,9 +27,10 @@ type DashboardProvisionerFactory func(string, dashboards.Store) (DashboardProvis
 
 // Provisioner is responsible for syncing dashboard from disk to Grafana's database.
 type Provisioner struct {
-	log         log.Logger
-	fileReaders []*FileReader
-	configs     []*config
+	log           log.Logger
+	fileReaders   []*FileReader
+	configs       []*config
+	sanityChecker *sanityChecker
 }
 
 // New returns a new DashboardProvisioner
@@ -41,15 +42,17 @@ func New(configDirectory string, store dashboards.Store) (DashboardProvisioner, 
 		return nil, errutil.Wrap("Failed to read dashboards config", err)
 	}
 
-	fileReaders, err := getFileReaders(configs, logger, store)
+	checker := newSanityChecker()
+	fileReaders, err := getFileReaders(configs, logger, checker)
 	if err != nil {
 		return nil, errutil.Wrap("Failed to initialize file readers", err)
 	}
 
 	d := &Provisioner{
-		log:         logger,
-		fileReaders: fileReaders,
-		configs:     configs,
+		log:           logger,
+		fileReaders:   fileReaders,
+		configs:       configs,
+		sanityChecker: checker,
 	}
 
 	return d, nil
@@ -70,6 +73,7 @@ func (provider *Provisioner) Provision() error {
 		}
 	}
 
+	provider.sanityChecker.logWarnings(provider.log)
 	return nil
 }
 
@@ -92,6 +96,8 @@ func (provider *Provisioner) PollChanges(ctx context.Context) {
 	for _, reader := range provider.fileReaders {
 		go reader.pollChanges(ctx)
 	}
+
+	go provider.sanityChecker.startLogWarningsLoop(ctx, provider.log)
 }
 
 // GetProvisionerResolvedPath returns resolved path for the specified provisioner name. Can be used to generate
@@ -115,14 +121,13 @@ func (provider *Provisioner) GetAllowUIUpdatesFromConfig(name string) bool {
 	return false
 }
 
-func getFileReaders(configs []*config, logger log.Logger, store dashboards.Store) ([]*FileReader, error) {
+func getFileReaders(configs []*config, logger log.Logger) ([]*FileReader, error) {
 	var readers []*FileReader
 
 	for _, config := range configs {
 		switch config.Type {
 		case "file":
-			fileReader, err := NewDashboardFileReader(config, logger.New("type", config.Type, "name", config.Name),
-				store)
+			fileReader, err := NewDashboardFileReader(config, logger.New("type", config.Type, "name", config.Name))
 			if err != nil {
 				return nil, errutil.Wrapf(err, "Failed to create file reader for config %v", config.Name)
 			}
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -27,10 +27,10 @@ type DashboardProvisionerFactory func(string, dashboards.Store) (DashboardProvis
 
 // Provisioner is responsible for syncing dashboard from disk to Grafana's database.
 type Provisioner struct {
-	log           log.Logger
-	fileReaders   []*FileReader
-	configs       []*config
-	sanityChecker *sanityChecker
+	log                log.Logger
+	fileReaders        []*FileReader
+	configs            []*config
+	duplicateValidator *duplicateValidator
 }
 
 // New returns a new DashboardProvisioner
@@ -42,17 +42,16 @@ func New(configDirectory string, store dashboards.Store) (DashboardProvisioner, 
 		return nil, errutil.Wrap("Failed to read dashboards config", err)
 	}
 
-	checker := newSanityChecker()
-	fileReaders, err := getFileReaders(configs, logger, checker)
+	fileReaders, err := getFileReaders(configs, logger)
 	if err != nil {
 		return nil, errutil.Wrap("Failed to initialize file readers", err)
 	}
 
 	d := &Provisioner{
-		log:           logger,
-		fileReaders:   fileReaders,
-		configs:       configs,
-		sanityChecker: checker,
+		log:                logger,
+		fileReaders:        fileReaders,
+		configs:            configs,
+		duplicateValidator: newDuplicateValidator(fileReaders),
 	}
 
 	return d, nil
@@ -73,7 +72,7 @@ func (provider *Provisioner) Provision() error {
 		}
 	}
 
-	provider.sanityChecker.logWarnings(provider.log)
+	provider.duplicateValidator.logWarnings(provider.log)
 	return nil
 }
 
@@ -97,7 +96,7 @@ func (provider *Provisioner) PollChanges(ctx context.Context) {
 		go reader.pollChanges(ctx)
 	}
 
-	go provider.sanityChecker.startLogWarningsLoop(ctx, provider.log)
+	go provider.duplicateValidator.startLogWarningsLoop(ctx, provider.log)
 }
 
 // GetProvisionerResolvedPath returns resolved path for the specified provisioner name. Can be used to generate
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -30,7 +30,7 @@ type Provisioner struct {
 	log                log.Logger
 	fileReaders        []*FileReader
 	configs            []*config
-	duplicateValidator *duplicateValidator
+	duplicateValidator duplicateValidator
 }
 
 // New returns a new DashboardProvisioner
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -42,7 +42,7 @@ func New(configDirectory string, store dashboards.Store) (DashboardProvisioner, 
 		return nil, errutil.Wrap("Failed to read dashboards config", err)
 	}
 
-	fileReaders, err := getFileReaders(configs, logger)
+	fileReaders, err := getFileReaders(configs, logger, store)
 	if err != nil {
 		return nil, errutil.Wrap("Failed to initialize file readers", err)
 	}
@@ -51,7 +51,7 @@ func New(configDirectory string, store dashboards.Store) (DashboardProvisioner, 
 		log:                logger,
 		fileReaders:        fileReaders,
 		configs:            configs,
-		duplicateValidator: newDuplicateValidator(fileReaders),
+		duplicateValidator: newDuplicateValidator(logger, fileReaders),
 	}
 
 	return d, nil
@@ -72,7 +72,7 @@ func (provider *Provisioner) Provision() error {
 		}
 	}
 
-	provider.duplicateValidator.validate(provider.log)
+	provider.duplicateValidator.validate()
 	return nil
 }
 
@@ -96,7 +96,7 @@ func (provider *Provisioner) PollChanges(ctx context.Context) {
 		go reader.pollChanges(ctx)
 	}
 
-	go provider.duplicateValidator.Run(ctx, provider.log)
+	go provider.duplicateValidator.Run(ctx)
 }
 
 // GetProvisionerResolvedPath returns resolved path for the specified provisioner name. Can be used to generate
@@ -120,13 +120,14 @@ func (provider *Provisioner) GetAllowUIUpdatesFromConfig(name string) bool {
 	return false
 }
 
-func getFileReaders(configs []*config, logger log.Logger) ([]*FileReader, error) {
+func getFileReaders(configs []*config, logger log.Logger, store dashboards.Store) ([]*FileReader, error) {
 	var readers []*FileReader
 
 	for _, config := range configs {
 		switch config.Type {
 		case "file":
-			fileReader, err := NewDashboardFileReader(config, logger.New("type", config.Type, "name", config.Name))
+			fileReader, err := NewDashboardFileReader(config, logger.New("type", config.Type, "name", config.Name),
+				store)
 			if err != nil {
 				return nil, errutil.Wrapf(err, "Failed to create file reader for config %v", config.Name)
 			}
--- a/pkg/services/provisioning/dashboards/dashboard.go
+++ b/pkg/services/provisioning/dashboards/dashboard.go
@@ -72,7 +72,7 @@ func (provider *Provisioner) Provision() error {
 		}
 	}
 
-	provider.duplicateValidator.logWarnings(provider.log)
+	provider.duplicateValidator.validate(provider.log)
 	return nil
 }
 
@@ -96,7 +96,7 @@ func (provider *Provisioner) PollChanges(ctx context.Context) {
 		go reader.pollChanges(ctx)
 	}
 
-	go provider.duplicateValidator.startLogWarningsLoop(ctx, provider.log)
+	go provider.duplicateValidator.Run(ctx, provider.log)
 }
 
 // GetProvisionerResolvedPath returns resolved path for the specified provisioner name. Can be used to generate
--- a/pkg/services/provisioning/dashboards/file_reader.go
+++ b/pkg/services/provisioning/dashboards/file_reader.go
@@ -31,14 +31,16 @@ var (
 type FileReader struct {
 	Cfg                          *config
 	Path                         string
+	FoldersFromFilesStructure    bool
 	log                          log.Logger
 	dashboardProvisioningService dashboards.DashboardProvisioningService
-	sanityChecker                *sanityChecker
-	FoldersFromFilesStructure    bool
+
+	mux          sync.RWMutex
+	usageTracker *usageTracker
 }
 
 // NewDashboardFileReader returns a new filereader based on `config`
-func NewDashboardFileReader(cfg *config, log log.Logger, checker *sanityChecker) (*FileReader, error) {
+func NewDashboardFileReader(cfg *config, log log.Logger) (*FileReader, error) {
 	var path string
 	path, ok := cfg.Options["path"].(string)
 	if !ok {
@@ -61,7 +63,7 @@ func NewDashboardFileReader(cfg *config, log log.Logger, checker *sanityChecker)
 		log:                          log,
 		dashboardProvisioningService: dashboards.NewProvisioningService(store),
 		FoldersFromFilesStructure:    foldersFromFilesStructure,
-		sanityChecker:                checker,
+		usageTracker:                 newUsageTracker(),
 	}, nil
 }
 
@@ -112,7 +114,10 @@ func (fr *FileReader) walkDisk() error {
 		return err
 	}
 
-	fr.sanityChecker.Set(fr.Cfg.Name, usageTracker)
+	fr.mux.Lock()
+	defer fr.mux.Unlock()
+
+	fr.usageTracker = usageTracker
 	return nil
 }
 
@@ -412,6 +417,13 @@ func (fr *FileReader) resolvedPath() string {
 	return path
 }
 
+func (fr *FileReader) getUsageTracker() *usageTracker {
+	fr.mux.RLock()
+	defer fr.mux.RUnlock()
+
+	return fr.usageTracker
+}
+
 type provisioningMetadata struct {
 	uid      string
 	identity dashboardIdentity
@@ -444,116 +456,5 @@ func (t *usageTracker) track(pm provisioningMetadata) {
 	}
 	if pm.identity.Exists() {
 		t.titleUsage[pm.identity]++
-	}
-}
-
-type set struct {
-	data map[string]struct{}
-}
-
-func newSet() *set {
-	return &set{make(map[string]struct{})}
-}
-
-func (s *set) Add(entry string) {
-	s.data[entry] = struct{}{}
-}
-
-func (s *set) AsSlice() []string {
-	entries := make([]string, 0, len(s.data))
-
-	for entry := range s.data {
-		entries = append(entries, entry)
-	}
-
-	return entries
-}
-
-type duplicate struct {
-	Readers *set
-	Sum     uint8
-}
-
-func newDuplicate() *duplicate {
-	return &duplicate{Readers: newSet()}
-}
-
-type duplicateEntries struct {
-	Title map[dashboardIdentity]*duplicate
-	UID   map[string]*duplicate
-}
-
-func newDuplicateEntries() *duplicateEntries {
-	return &duplicateEntries{Title: make(map[dashboardIdentity]*duplicate), UID: make(map[string]*duplicate)}
-}
-
-type sanityChecker struct {
-	mux      sync.RWMutex
-	trackers map[string]*usageTracker
-}
-
-func newSanityChecker() *sanityChecker {
-	return &sanityChecker{trackers: make(map[string]*usageTracker)}
-}
-
-func (c *sanityChecker) Set(readerName string, tracker *usageTracker) {
-	c.mux.Lock()
-	defer c.mux.Unlock()
-
-	c.trackers[readerName] = tracker
-}
-
-func (c *sanityChecker) getDuplicates() *duplicateEntries {
-	duplicates := newDuplicateEntries()
-
-	for readerName, tracker := range c.trackers {
-		for uid, times := range tracker.uidUsage {
-			if _, ok := duplicates.UID[uid]; !ok {
-				duplicates.UID[uid] = newDuplicate()
-			}
-			duplicates.UID[uid].Sum += times
-			duplicates.UID[uid].Readers.Add(readerName)
-		}
-
-		for id, times := range tracker.titleUsage {
-			if _, ok := duplicates.Title[id]; !ok {
-				duplicates.Title[id] = newDuplicate()
-			}
-			duplicates.Title[id].Sum += times
-			duplicates.Title[id].Readers.Add(readerName)
-		}
-	}
-
-	return duplicates
-}
-
-func (c *sanityChecker) logWarnings(log log.Logger) {
-	c.mux.Lock()
-	defer c.mux.Unlock()
-
-	duplicates := c.getDuplicates()
-
-	for uid, usage := range duplicates.UID {
-		if usage.Sum > 1 {
-			log.Error("the same 'uid' is used more than once", "uid", uid, "times", usage.Sum, "providers", usage.Readers.AsSlice())
-		}
-	}
-
-	for id, usage := range duplicates.Title {
-		if usage.Sum > 1 {
-			log.Error("dashboard title is not unique in folder", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", usage.Readers.AsSlice())
-		}
-	}
-}
-
-func (c *sanityChecker) startLogWarningsLoop(ctx context.Context, log log.Logger) {
-	ticker := time.NewTicker(30 * time.Second)
-	for {
-		select {
-		case <-ticker.C:
-			c.logWarnings(log)
-		case <-ctx.Done():
-			return
-		}
 	}
 }
--- a/pkg/services/provisioning/dashboards/file_reader.go
+++ b/pkg/services/provisioning/dashboards/file_reader.go
@@ -137,7 +137,8 @@ func (fr *FileReader) isDatabaseAccessRestricted() bool {
 }
 
 // storeDashboardsInFolder saves dashboards from the filesystem on disk to the folder from config
-func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo, dashboardRefs map[string]*models.DashboardProvisioning, usageTracker *usageTracker) error {
+func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo, 
+	dashboardRefs map[string]*models.DashboardProvisioning, usageTracker *usageTracker) error {
 	folderID, err := getOrCreateFolderID(fr.Cfg, fr.dashboardProvisioningService, fr.Cfg.Folder)
 	if err != nil && !errors.Is(err, ErrFolderNameMissing) {
 		return err
@@ -158,7 +159,8 @@ func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.Fil
 
 // storeDashboardsInFoldersFromFilesystemStructure saves dashboards from the filesystem on disk to the same folder
 // in grafana as they are in on the filesystem
-func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk map[string]os.FileInfo, dashboardRefs map[string]*models.DashboardProvisioning, resolvedPath string, usageTracker *usageTracker) error {
+func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk map[string]os.FileInfo,
+	dashboardRefs map[string]*models.DashboardProvisioning, resolvedPath string, usageTracker *usageTracker) error {
 	for path, fileInfo := range filesFoundOnDisk {
 		folderName := ""
 
@@ -263,7 +265,12 @@ func (fr *FileReader) saveDashboard(path string, folderID int64, fileInfo os.Fil
 			Updated:    resolvedFileInfo.ModTime().Unix(),
 			CheckSum:   jsonFile.checkSum,
 		}
-		_, err = fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp)
+		if _, err := fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp); err != nil {
+			return provisioningMetadata, err
+		}
+	} else {
+		fr.log.Warn("Not saving new dashboard due to restricted database access", "provisioner", fr.Cfg.Name, 
+			"file", path, "folderId", dash.Dashboard.FolderId)
 	}
 
 	return provisioningMetadata, err
--- a/pkg/services/provisioning/dashboards/file_reader.go
+++ b/pkg/services/provisioning/dashboards/file_reader.go
@@ -137,7 +137,7 @@ func (fr *FileReader) isDatabaseAccessRestricted() bool {
 }
 
 // storeDashboardsInFolder saves dashboards from the filesystem on disk to the folder from config
-func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo, 
+func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo,
 	dashboardRefs map[string]*models.DashboardProvisioning, usageTracker *usageTracker) error {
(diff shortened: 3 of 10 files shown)


pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+type duplicateEntries struct {
+	Title map[dashboardIdentity]*duplicate
+	UID   map[string]*duplicate
+}
+
+func newDuplicateEntries() *duplicateEntries {
+	return &duplicateEntries{Title: make(map[dashboardIdentity]*duplicate), UID: make(map[string]*duplicate)}
+}
+
+type sanityChecker struct {
```
sakjur, line 482: Since you're already improving the name maybe you can take it a step further and change it to either something more descriptive such as `duplicateChecker` or something more succinct such as `validator`?

And then we can break this out to `checker.go` or `validator.go` because it really doesn't logically belong to FileReader anymore.

sakjur (dismissed):
Thank you for this contribution! I've added some feedback on this related to the Go style rather than the feature in itself. I think this could be made easier to follow with some minor tweaks.

NB: I won't make the final approval of this PR since that falls outside of my domain.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
 }
 
 func (d *dashboardIdentity) Exists() bool {
-	return len(d.title) > 0 && d.folderID > 0
+	return len(d.title) > 0
```
aknuds1, line 416: Please illuminate me, why is `&& d.folderID > 0` no longer necessary?

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+	Sum     uint8
+}
+
+func newDuplicate() *duplicate {
+	return &duplicate{Readers: newSet()}
+}
+
+type duplicateEntries struct {
+	Title map[dashboardIdentity]*duplicate
+	UID   map[string]*duplicate
```
aknuds1, line 475: Why are these fields not named in plural, since they are collections?
```suggestion
	Titles map[dashboardIdentity]*duplicate
	UIDs   map[string]*duplicate
```

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -248,6 +250,42 @@ func TestDashboardFileReader(t *testing.T) {
 				So(folderCount, ShouldEqual, 2)
 				So(dashCount, ShouldEqual, 2)
 			})
+
+			Convey("Sanity checker should collect info about duplicate uids and folder/title's", func() {
```
aknuds1, line 254: You mean folders/titles?
```suggestion
			Convey("Sanity checker should collect info about duplicate UIDs and folders/titles", func() {
```

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -248,6 +250,42 @@ func TestDashboardFileReader(t *testing.T) {
...
+				folderID, err := getOrCreateFolderID(cfg, fakeService, "sanity-check-folder")
+				So(err, ShouldBeNil)
+
+				identity := dashboardIdentity{folderID: folderID, title: "Grafana"}
+
+				cfg1 := &config{Name: "first", Type: "file", OrgID: 1, Folder: "sanity-check-folder", Options: map[string]interface{}{"path": containingUID}}
```
aknuds1, line 260: This is more readable.
```suggestion
				cfg1 := &config{
					Name: "first", Type: "file", OrgID: 1, Folder: "sanity-check-folder", 
					Options: map[string]interface{}{"path": containingUID},
				}
```

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -248,6 +250,42 @@ func TestDashboardFileReader(t *testing.T) {
...
+				cfg2 := &config{Name: "second", Type: "file", OrgID: 1, Folder: "sanity-check-folder", Options: map[string]interface{}{"path": containingUID}}
```
aknuds1, line 261:
```suggestion
				cfg2 := &config{
					Name: "second", Type: "file", OrgID: 1, Folder: "sanity-check-folder", 
					Options: map[string]interface{}{"path": containingUID},
				}
```

aknuds1 (changes requested):
I'll have to review this in several iterations, since I'm new to the functionality, but left some suggestions.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+		index++
+	}
+
+	return entries
+}
+
+type duplicate struct {
+	Readers *set
+	Sum     uint8
+}
```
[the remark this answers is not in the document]
aknuds1: I like the idea to structure less, as @sakjur suggests. It makes things less complex (a good thing usually!).

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+	}
+
+	return duplicates
+}
+
+func (c *sanityChecker) logWarnings(log log.Logger) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	duplicates := c.getDuplicates()
```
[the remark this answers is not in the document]
aknuds1: @sakjur Might we want to keep also the `defer c.mux.RUnlock()` call just in case we either forget to unlock after `c.getDuplicates()` or the latter panics (and gets recovered)?

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+
+	for uid, usage := range duplicates.UID {
+		if usage.Sum > 1 {
+			log.Error("the same 'uid' is used more than once", "uid", uid, "times", usage.Sum, "providers", usage.Readers.AsSlice())
+		}
+	}
+
+	for id, usage := range duplicates.Title {
+		if usage.Sum > 1 {
+			log.Error("the same 'folder/title' is used more than once", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", usage.Readers.AsSlice())
```
aknuds1, line 536: What is a folder/title? Is this supposed to be a folder title? I honestly don't know, which is why I have to ask.
```suggestion
			log.Error("the same 'folder/title' is used more than once", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", usage.Readers.AsSlice())
```

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -412,42 +413,139 @@ type dashboardIdentity struct {
...
+func (s *set) AsSlice() []string {
+	entries := make([]string, len(s.data))
+	index := 0
+
+	for entry := range s.data {
+		entries[index] = entry
+		index++
+	}
+
+	return entries
```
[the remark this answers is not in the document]
aknuds1: Agree with @sakjur.

pkg/services/provisioning/dashboards/file_reader.go
sakjur: If I understand this correctly, maybe something like this would be clearer?

```suggestion
			log.Error("dashboard title is not unique in folder", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", usage.Readers.AsSlice())
```

aknuds1: Same as @sakjur, if it's actually about dashboard title's uniqueness within folder, this is a much better error message.

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
nabokihms: Yeah, this is here by accident. I was trying different approaches while working on this PR. `Set` is the part of my previous realization, thus I will get rid of it.

pkg/services/provisioning/dashboards/dashboard.go
```diff
@@ -38,16 +39,17 @@ func New(configDirectory string) (*Provisioner, error) {
 		return nil, errutil.Wrap("Failed to read dashboards config", err)
 	}
 
-	fileReaders, err := getFileReaders(configs, logger)
-
+	checker := newSanityChecker()
+	fileReaders, err := getFileReaders(configs, logger, checker)
```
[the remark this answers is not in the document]
nabokihms: Sounds good, thank you! I will try it.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: Yep, I felt like we need to move validation logic to another `.go` file. Thank you for guiding me on the right path.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: My bad, I will fix it.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: I will use the one you provided. Thanks.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: It's a bug. folderID `0` is an ID of the General folder. We need to remove this clause to track duplicates in the General folder too.

aknuds1: Thanks for clarifying that!

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
sakjur: @aknuds1 That's a good point, moving the mutex logic into `.getDuplicates()` should be straightforward enough 🙂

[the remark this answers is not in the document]
aknuds1: @sakjur An even better point :D

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
nabokihms: Done.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: Fixed.

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
nabokihms: It was removed.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: The new name of this entity is `duplicateValidator`. I have moved the code to `validator.go` and tests to `validator_test.go`.

pkg/services/provisioning/dashboards/dashboard.go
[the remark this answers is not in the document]
nabokihms: I added a new realization that iterates over `[]*FileReader`. Please, take a look when you have time.

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
nabokihms: I have moved lock closer to the data it protects (to the FileReader structure).

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type duplicateEntries struct {
+	Titles map[dashboardIdentity]*duplicate
+	UIDs   map[string]*duplicate
+}
+
+func newDuplicateEntries() *duplicateEntries {
+	return &duplicateEntries{Titles: make(map[dashboardIdentity]*duplicate), UIDs: make(map[string]*duplicate)}
+}
```
aknuds1, line 26: For simplicity, can we just remove this function and inline the call? It's only used in one place and I think inlining makes for less code.

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return duplicates
+}
+
+func (c *duplicateValidator) logWarnings(log log.Logger) {
+	duplicates := c.getDuplicates()
+	for uid, usage := range duplicates.UIDs {
+		if usage.Sum > 1 {
+			log.Error("the same uid is used more than once", "uid", uid, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```
aknuds1, line 66: It's preferable IMO to standardize on writing acronyms in uppercase:
```suggestion
			log.Error("the same UID is used more than once", "uid", uid, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		err = reader1.startWalkingDisk()
+		require.NoError(t, err)
+
+		err = reader2.startWalkingDisk()
+		require.NoError(t, err)
+
+		duplicates := duplicateValidator.getDuplicates()
+
+		require.Equal(t, duplicates.UIDs["Z-phNqGmz"].Sum, uint8(2))
```
aknuds1, line 60: In `require.Equal`, the expectation comes first:
```suggestion
		require.Equal(t, uint8(2), duplicates.UIDs["Z-phNqGmz"].Sum)
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,70 @@
...
+		uidUsageReaders := keysToSlice(duplicates.UIDs["Z-phNqGmz"].InvolvedReaders)
+		sort.Strings(uidUsageReaders)
+		require.Equal(t, uidUsageReaders, []string{"first", "second"})
```
aknuds1, line 63: In `require.Equal`, the expectation comes first:
```suggestion
		require.Equal(t, []string{"first", "second"}, uidUsageReaders)
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+		require.Equal(t, duplicates.Titles[identity].Sum, uint8(2))
```
aknuds1, line 65:
```suggestion
		require.Equal(t, uint8(2), duplicates.Titles[identity].Sum)
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,70 @@
...
+		titleUsageReaders := keysToSlice(duplicates.Titles[identity].InvolvedReaders)
+		sort.Strings(titleUsageReaders)
+		require.Equal(t, titleUsageReaders, []string{"first", "second"})
```
aknuds1, line 68:
```suggestion
		require.Equal(t, []string{"first", "second"}, titleUsageReaders)
```

aknuds1 (changes requested):
Looks better now, but left some more suggestions.

sakjur:
:+1: I really like the reorganization! Looks a lot cleaner now, thank you! 🙏

nabokihms:
Thanks for the review! I fixed everything according to suggestions.

P.S. There is a problem with frontend dependencies according to the CI pipeline log, even after rebasing on the master branch. Guess the only opportunity is to wait for the fix.

aknuds1:
Yeah sorry, CI issues is due to the NPM registry. Will just have to wait it out.

aknuds1:
Looks like NPM registry issues are over.

aknuds1 (approved):

aknuds1:
Approving since it looks good AFAICT, but will discuss with my colleagues to get their POV also.

aknuds1:
@bergquist Is this the solution you want?

bergquist (approved):
LGTM! :)

aknuds1:
We're going to test this next week, to make sure it works well, before considering merging it.

marefr:
@nabokihms in general I would suggest that Grafana would return an error when you have a collision of uids since that could lead to overloading the database. Either treat the error as not processing/including the affected providers in the provisioning or either make the Grafana process stop  Do you have any input/feedback on this?

nabokihms:
@marefr good suggestion. I think, stop the Grafana process is the easiest way, but it looks like the worst variant. Even copying of the dashboard inside the provisioning folder using copy/paste hotkeys will crush the Grafana, what is unwanted behavior for sure.

According to code from this PR, we record information about dashboards usage in tracker only after saving them into the database. The duplicates tracker knows every reader. Maybe we can tell the reader, that because of duplicates he can't write to the database anymore. The reader will be able to continue provisioning (but without write permissions), and the duplicates tracker will continue warning users with log messages. 

After the "duplicates" issue is resolved, the duplicates tracker will allow the reader to write to the database again. What do you think about it? This looks simple and useful.

aknuds1:
Your suggestion sounds like it could work @nabokihms.

nabokihms:
Ok, I will work on implementation then.

aknuds1:
That's just my opinion though, I don't know how @marefr feels.

marefr:
@nabokihms I agree, sounds like your suggestion could work.

nabokihms:
I implemented the feature I had been talking about and rebased the branch onto the master to activate CI.

marefr:
@nabokihms thanks. We'll try to review your changes soon..

aknuds1:
> /keepalive
I already removed the stale label :)

nabokihms:
@aknuds1 @marefr Happy new year! How is testing going? Can I help with something?

aknuds1:
Hi there, happy new year to you as well! I will try to look at this next week, January's been busy so far!

aknuds1 (approved):
Have reviewed the PR in depth now, although not actually tested it yet, and it looks good to me :+1: However, I'd like to merge #30396 first since it contains cleanup I made while reviewing.

aknuds1:
Can you merge with master and fix conflicts?

pkg/services/provisioning/dashboards/config_reader.go
```diff
@@ -110,7 +110,7 @@ func (cr *configReader) readConfig() ([]*config, error) {
 
 	for uid, times := range uidUsage {
 		if times > 1 {
-			cr.log.Error("the same folder UID is used more than once", "folderUid", uid)
+			cr.log.Error("the same folderUID is used more than once", "folderUID", uid)
```
aknuds1, line 113: I don't see any reason for this change, the original is fine.
```suggestion
			cr.log.Error("the same folder UID is used more than once", "folderUid", uid)
```

pkg/services/provisioning/dashboards/dashboard.go
```diff
@@ -26,9 +26,10 @@ type DashboardProvisionerFactory func(string) (DashboardProvisioner, error)
 
 // Provisioner is responsible for syncing dashboard from disk to Grafana's database.
 type Provisioner struct {
-	log         log.Logger
-	fileReaders []*FileReader
-	configs     []*config
+	log                log.Logger
+	fileReaders        []*FileReader
+	configs            []*config
+	duplicateValidator *duplicateValidator
```
aknuds1, line 32: Is there any need for this field to be a pointer? From my testing it can be a normal value type.
```suggestion
	duplicateValidator duplicateValidator
```

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -98,25 +104,39 @@ func (fr *FileReader) walkDisk() error {
...
+	fr.mux.RLock()
+	defer fr.mux.RUnlock()
+
+	return fr.dbWriteAccessRestricted
+}
+
 // storeDashboardsInFolder saves dashboards from the filesystem on disk to the folder from config
-func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo,
-	dashboardRefs map[string]*models.DashboardProvisioning, sanityChecker *provisioningSanityChecker) error {
+func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo, dashboardRefs map[string]*models.DashboardProvisioning, usageTracker *usageTracker) error {
```
aknuds1, line 139: It's better if you don't make the line too long, since it's then difficult to diff.
```suggestion
func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.FileInfo, 
	dashboardRefs map[string]*models.DashboardProvisioning, usageTracker *usageTracker) error {
```

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -130,15 +150,14 @@ func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.Fil
...
 	}
 	return nil
 }
 
 // storeDashboardsInFoldersFromFilesystemStructure saves dashboards from the filesystem on disk to the same folder
-// in Grafana as they are in on the filesystem.
-func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk map[string]os.FileInfo,
-	dashboardRefs map[string]*models.DashboardProvisioning, resolvedPath string, sanityChecker *provisioningSanityChecker) error {
+// in grafana as they are in on the filesystem
+func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk map[string]os.FileInfo, dashboardRefs map[string]*models.DashboardProvisioning, resolvedPath string, usageTracker *usageTracker) error {
```
aknuds1, line 160: It's better if you don't make the line too long, since it's then difficult to diff.
```suggestion
func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk map[string]os.FileInfo,
	dashboardRefs map[string]*models.DashboardProvisioning, resolvedPath string, usageTracker *usageTracker) error {
```

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -235,15 +254,17 @@ func (fr *FileReader) saveDashboard(path string, folderID int64, fileInfo os.Fil
...
-		CheckSum:   jsonFile.checkSum,
+	if !fr.isDatabaseAccessRestricted() {
+		fr.log.Debug("saving new dashboard", "provisioner", fr.Cfg.Name, "file", path, "folderId", dash.Dashboard.FolderId)
+		dp := &models.DashboardProvisioning{
+			ExternalId: path,
+			Name:       fr.Cfg.Name,
+			Updated:    resolvedFileInfo.ModTime().Unix(),
+			CheckSum:   jsonFile.checkSum,
+		}
+		_, err = fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp)
```
aknuds1, line 265: It's better that you return the error here and change line 268 to return `nil` for error, since if the `if` statement doesn't execute, it's difficult for the reader to ensure that `err` is nil on line 268.
```suggestion
		if _, err := fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp); err != nil {
			return provisioningMetadata, err
		}
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, duplicate := range d.UIDs {
+		if duplicate.Sum <= 1 {
+			continue
+		}
+
+		for readerName := range duplicate.InvolvedReaders {
+			if _, ok := involvedReaders[readerName]; !ok {
+				involvedReaders[readerName] = struct{}{}
+			}
```
aknuds1, line 35: There shouldn't be any practical benefit to only set this key if it's not already been set, so this will read more easily:
```suggestion
			involvedReaders[readerName] = struct{}{}
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, duplicate := range d.Titles {
+		if duplicate.Sum <= 1 {
+			continue
+		}
+
+		for readerName := range duplicate.InvolvedReaders {
+			if _, ok := involvedReaders[readerName]; !ok {
+				involvedReaders[readerName] = struct{}{}
+			}
```
aknuds1, line 47: There shouldn't be any practical benefit to only set this key if it's not already been set, so this will read more easily:
```suggestion
			involvedReaders[readerName] = struct{}{}
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+	return involvedReaders
+}
+
+type duplicateValidator struct {
+	readers []*FileReader
+}
+
+func newDuplicateValidator(readers []*FileReader) *duplicateValidator {
+	return &duplicateValidator{readers: readers}
+}
```
aknuds1, line 60:
```suggestion
func newDuplicateValidator(readers []*FileReader) duplicateValidator {
	return duplicateValidator{readers: readers}
}
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+	for _, reader := range c.readers {
+		readerName := reader.Cfg.Name
+		tracker := reader.getUsageTracker()
+
+		for uid, times := range tracker.uidUsage {
+			if _, ok := duplicates.UIDs[uid]; !ok {
+				duplicates.UIDs[uid] = newDuplicate()
+			}
+			duplicates.UIDs[uid].Sum += times
+			duplicates.UIDs[uid].InvolvedReaders[reader.Cfg.Name] = struct{}{}
```
aknuds1, line 77:
```suggestion
			duplicates.UIDs[uid].InvolvedReaders[readerName] = struct{}{}
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return &duplicates
+}
+
+func (c *duplicateValidator) logWarnings(log log.Logger, duplicates *duplicateEntries) {
+	for uid, usage := range duplicates.UIDs {
+		if usage.Sum > 1 {
+			log.Error("the same UID is used more than once", "uid", uid, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```
aknuds1, line 95: Should this really be logged as an error? It doesn't match the function name which says it will log warnings.

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	for id, usage := range duplicates.Titles {
+		if usage.Sum > 1 {
+			log.Error("dashboard title is not unique in folder", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```
aknuds1, line 101: Should this really be logged as an error? It doesn't match the function name which says it will log warnings.

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (c *duplicateValidator) takeAwayWritePermissions(log log.Logger, duplicates *duplicateEntries) {
+	involvedReaders := duplicates.InvolvedReaders()
+	for _, reader := range c.readers {
+		_, isReaderWithDuplicates := involvedReaders[reader.Cfg.Name]
+		// We restrict reader permissions to write to the database here to prevent overloading
+		reader.changeWritePermissions(isReaderWithDuplicates)
+
+		if isReaderWithDuplicates {
+			log.Error("dashboards provisioning provider has no database write permissions because of duplicates", "provider", reader.Cfg.Name)
```
aknuds1, line 114: Should this really be logged as an error? I wonder if it should be a warning instead.

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,123 @@
...
+		require.Equal(t, []string{"first", "second"}, uidUsageReaders)
+
+		require.Equal(t, uint8(2), duplicates.Titles[identity].Sum)
+		titleUsageReaders := keysToSlice(duplicates.Titles[identity].InvolvedReaders)
+		sort.Strings(titleUsageReaders)
+		require.Equal(t, []string{"first", "second"}, titleUsageReaders)
+
+		duplicateValidator.validate(logger)
+		require.Equal(t, true, reader1.isDatabaseAccessRestricted())
+		require.Equal(t, true, reader2.isDatabaseAccessRestricted())
```
aknuds1, line 75:
```suggestion
		require.True(t, reader1.isDatabaseAccessRestricted())
		require.True(t, reader2.isDatabaseAccessRestricted())
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,123 @@
...
+		require.Equal(t, []string{"first"}, uidUsageReaders)
+
+		require.Equal(t, uint8(2), duplicates.Titles[identity].Sum)
+		titleUsageReaders := keysToSlice(duplicates.Titles[identity].InvolvedReaders)
+		sort.Strings(titleUsageReaders)
+		require.Equal(t, []string{"first"}, titleUsageReaders)
+
+		duplicateValidator.validate(logger)
+		require.Equal(t, true, reader1.isDatabaseAccessRestricted())
+		require.Equal(t, false, reader2.isDatabaseAccessRestricted())
```
aknuds1, line 121:
```suggestion
		require.True(t, reader1.isDatabaseAccessRestricted())
		require.False(t, reader2.isDatabaseAccessRestricted())
```

aknuds1 (changes requested):
Spotted some issues.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -235,15 +254,17 @@ func (fr *FileReader) saveDashboard(path string, folderID int64, fileInfo os.Fil
...
 	}
```
aknuds1, line 266: How about adding

```suggestion
	} else {
		fr.log.Warn("Not saving new dashboard due to restricted database access", "provisioner", fr.Cfg.Name, 
			"file", path, "folderId", dash.Dashboard.FolderId)
	}
```

pkg/services/provisioning/dashboards/validator.go
nabokihms: I think this is not necessary. I assume that I took it from the previous realization.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -235,15 +256,22 @@ func (fr *FileReader) saveDashboard(path string, folderID int64, fileInfo os.Fil
...
+		if _, err := fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp); err != nil {
+			return provisioningMetadata, err
+		}
+	} else {
+		fr.log.Warn("Not saving new dashboard due to restricted database access", "provisioner", fr.Cfg.Name,
+			"file", path, "folderId", dash.Dashboard.FolderId)
 	}
 
-	_, err = fr.dashboardProvisioningService.SaveProvisionedDashboard(dash, dp)
 	return provisioningMetadata, err
```
aknuds1, line 275: Can you change this to return nil for error, so the reader can be sure of this?

```suggestion
	return provisioningMetadata, nil
```

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -130,15 +151,15 @@ func (fr *FileReader) storeDashboardsInFolder(filesFoundOnDisk map[string]os.Fil
...
 
-		sanityChecker.track(provisioningMetadata)
+		usageTracker.track(provisioningMetadata)
 	}
 	return nil
 }
 
 // storeDashboardsInFoldersFromFilesystemStructure saves dashboards from the filesystem on disk to the same folder
-// in Grafana as they are in on the filesystem.
+// in grafana as they are in on the filesystem
```
aknuds1, line 160: Please undo this change.
```suggestion
// in Grafana as they are in on the filesystem.
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	return &duplicates
+}
+
+func (c *duplicateValidator) logWarnings(log log.Logger, duplicates *duplicateEntries) {
+	for uid, usage := range duplicates.UIDs {
+		if usage.Sum > 1 {
+			log.Warn("the same UID is used more than once", "uid", uid, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```
aknuds1, line 91:
```suggestion
			log.Warn("the same UID is used more than once", "uid", uid, "times", usage.Sum, "providers",
				keysToSlice(usage.InvolvedReaders))
```

pkg/services/provisioning/dashboards/validator.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	for id, usage := range duplicates.Titles {
+		if usage.Sum > 1 {
+			log.Warn("dashboard title is not unique in folder", "title", id.title, "folderID", id.folderID, "times", usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```
aknuds1, line 97:
```suggestion
			log.Warn("dashboard title is not unique in folder", "title", id.title, "folderID", id.folderID, "times", 
				usage.Sum, "providers", keysToSlice(usage.InvolvedReaders))
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,123 @@
...
+		titleUsageReaders := keysToSlice(duplicates.Titles[identity].InvolvedReaders)
+		sort.Strings(titleUsageReaders)
+		require.Equal(t, []string{"first", "second"}, titleUsageReaders)
+
+		duplicateValidator.validate(logger)
+		require.True(t, reader1.isDatabaseAccessRestricted())
+		require.True(t, reader2.isDatabaseAccessRestricted())
+	})
+
+	t.Run("Duplicates validator should restrict write access for only reader with duplicates", func(t *testing.T) {
```
aknuds1, line 78:
```suggestion
	t.Run("Duplicates validator should restrict write access only for readers with duplicates", func(t *testing.T) {
```

aknuds1 (changes requested):
Some more nits.

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,123 @@
...
+		require.NoError(t, err)
+
+		identity := dashboardIdentity{folderID: folderID, title: "Grafana"}
+
+		cfg1 := &config{
+			Name: "first", Type: "file", OrgID: 1, Folder: "duplicates-validator-folder",
+			Options: map[string]interface{}{"path": dashboardContainingUID},
+		}
+		cfg2 := &config{
+			Name: "second", Type: "file", OrgID: 1, Folder: "duplicates-validator-folder",
```
aknuds1, line 43:
```suggestion
		const folderName = "duplicates-validator-folder"
		folderID, err := getOrCreateFolderID(cfg, fakeService, folderName)
		require.NoError(t, err)

		identity := dashboardIdentity{folderID: folderID, title: "Grafana"}

		cfg1 := &config{
			Name: "first", Type: "file", OrgID: 1, Folder: folderName,
			Options: map[string]interface{}{"path": dashboardContainingUID},
		}
		cfg2 := &config{
			Name: "second", Type: "file", OrgID: 1, Folder: folderName,
```

pkg/services/provisioning/dashboards/validator_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+		err = reader1.walkDisk()
+		require.NoError(t, err)
+
+		err = reader2.walkDisk()
+		require.NoError(t, err)
+
+		duplicates := duplicateValidator.getDuplicates()
+
+		folderID, err := getOrCreateFolderID(cfg, fakeService, "duplicates-validator-folder")
```
aknuds1, line 104:
```suggestion
		folderID, err := getOrCreateFolderID(cfg, fakeService, cfg1.Folder)
```

aknuds1 (commented):
Some suggestions for simplification of the tests.

aknuds1:
I will try to test this manually.

aknuds1 (approved):
Tested and LGTM! @marefr have you concluded yet?

marefr:
Planning to have a look

nabokihms:
@marefr Hello! Do you have any news about this PR?

nabokihms:
@marefr just a friendly reminder to check this PR. If you need something from me to move this PR forward, I'm always ready.

nabokihms:
@marefr Hello! How is testing going?

nabokihms:
@marefr Hello, this is just a friendly reminder about this PR. I assume that the team is busy with the new release, but maybe you can find a window in your schedule to review this. Thanks.

marefr (commented):
Very sorry. I actually started to review a while ago, but somehow forgot to submit it :cry: I'll try to have another look just to verify I haven't missed anything. Still planning to test this out as well.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -98,25 +104,40 @@ func (fr *FileReader) walkDisk() error {
...
 	}
 	if err != nil {
 		return err
 	}
 
-	sanityChecker.logWarnings(fr.log)
+	fr.mux.Lock()
+	defer fr.mux.Unlock()
 
+	fr.usageTracker = usageTracker
```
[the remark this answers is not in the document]
nabokihms: This is a kind of implementation of a non-blocking algorithm. I thought that resetting a tracker will cost time and block the entire process.

About concurrent access - yes, it can be accessed concurrently. Duplicates validator checks that there are no duplicates across all readers, and it requires access to usage trackers of the readers.
Corresponding link [here](https://github.com/grafana/grafana/pull/26742/files/1f0a9f0d40154eb7b3d1443b12be53f5478bebc8#diff-40968223800ec29c8675bd383f59f205f60d58c17284eb2abd72f5d4e916333aR65-R66).

nabokihms:
@marefr I rebased this PR onto main and fixed merge conflicts.

marefr:
@nabokihms Great thanks :+1:  I'm planning to review this today if time allows or tomorrow.

pkg/services/provisioning/dashboards/file_reader.go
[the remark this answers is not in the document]
marefr: That link doesn't work any longer since you force pushed. 

The `getUsageTracker()` does a `RLock` followed by a `RUnlock`, but I do notice you do a `Lock` followed by `Unlock` above the mention code above so please disregard my comment.

pkg/services/provisioning/dashboards/file_reader_linux_test.go
```diff
@@ -25,7 +25,9 @@ func TestProvisionedSymlinkedFolder(t *testing.T) {
 	}
 
 	reader, err := NewDashboardFileReader(cfg, log.New("test-logger"), nil)
-	require.NoError(t, err)
+	if err != nil {
+		t.Error("expected err to be nil")
+	}
```
marefr, line 30: I think the existing code makes more sense. Any reason you changed it?

```suggestion
	require.NoError(t, err)
```

marefr (approved):
This looks good. Tested it out a bit and seems to work as expected when having multiple dashboards with same uid in both same configuration and different configurations.

Sorry this has taken really long time for me to wrap my head around. 

LGTM

pkg/services/provisioning/dashboards/file_reader_linux_test.go
wbrowne: Will merge as is for now

PR #26746, opened 2020-07-31
InfluxDB/Flux: Increase series limit for Flux datasource

### LINKED ISSUE #26632: InfluxDB Flux: Grafana truncates flux queries to 10 tables (dataframes) maximum
**What happened**:
When writing queries which return more than 10 tables/series, only the first 10 are retrieved

Grafana should not be truncating the number of tables/series which are able to be returned by flux queries

Create a database `testing` in influxdb and write the following data to it:
[test.txt](https://github.com/grafana/grafana/files/4984748/test.txt)

Open up explore in grafana and perform the following flux query:
```
from(bucket: "testing/autogen")
  |> range(start: 2020-05-20T18:40:00Z, stop: 2020-05-20T18:42:00Z)
```

Change the time range in grafana to be absolute:
From: `2020-05-20 18:40:00`
To: `2020-05-20 18:42:00`
With the UTC timezone

Observe that only 10 series are there, and 2 series (namely 8 and 9) are missing from the query results

Performing the same flux query above with chronograf shows that there are actually 12 tables returned by this query

**Screenshots**:
Grafana Query
image

image

image

Chronograf Query:
image

**Environment**:
- Grafana version: Grafana v7.1.1 (3039f9c3bd)
- Data source type & version: InfluxDB v1.8.1 using flux
- OS Grafana is installed on: Linux Ubuntu 20.04

paul-sneddon:
What this PR does / why we need it:
Increases the limit for the number of series / tables that can be displayed for a dashboard using flux. The increase is from 10 to 50.

Which issue(s) this PR fixes:

Fixes #26632

Special notes for your reviewer:

### DIFF
--- a/pkg/tsdb/influxdb/flux/flux.go
+++ b/pkg/tsdb/influxdb/flux/flux.go
@@ -37,7 +37,7 @@ func Query(ctx context.Context, dsInfo *models.DataSource, tsdbQuery *tsdb.TsdbQ
 			continue
 		}
 
-		res := ExecuteQuery(context.Background(), *qm, runner, 10)
+		res := ExecuteQuery(context.Background(), *qm, runner, 50)
 
 		tRes.Results[query.RefId] = backendDataResponseToTSDBResponse(&res, query.RefId)
 	}


ryantxu (approved):
We should likely make this configurable, but bumping to 50 is better than 10

PR #26771, opened 2020-08-03
Provisioning: describe which folder caused an error while provisioning from files structure

nabokihms:
This PR adds a description about which folder caused an error while provisioning folders from files structure.

If we receive an error about folder while provisioning, it will look like this:
```
EROR[08-03|23:57:48] Failed to provision dashboard            logger=provisioning error="Failed to provision dashboards: Failed to provision config default: A folder with that name already exists"
```
This is ok for an ordinary dashboard provisioning. We can easily find a folder by the provisioning config name. But, using the `foldersFromFilesStructure` option makes it harder to determine a problematic folder.
We know which folder caused the problem, and, in my opinion, we need to share it with users.

Besides, I noticed that it's not clear from the documentation how to provision dashboards to the General folder. So I have added some notes about it.

Also, this PR:
* makes some variables constant because, as was mentioned before, we want to avoid globals in Grafana
* fixes an issue with receiver names for the dashboard struct
(it's just because I touched them while researching)

### DIFF
--- a/docs/sources/administration/provisioning.md
+++ b/docs/sources/administration/provisioning.md
@@ -275,6 +275,8 @@ providers:
 
 When Grafana starts, it will update/insert all dashboards available in the configured path. Then later on poll that path every **updateIntervalSeconds** and look for updated json files and update/insert those into the database.
 
+> **Note.** Dashboards will be provisioned to the `General` folder if the `folder` option is missing or empty.
+
 #### Making changes to a provisioned dashboard
 
 It's possible to make changes to a provisioned dashboard in the Grafana UI. However, it is not possible to automatically save the changes back to the provisioning source.
@@ -332,6 +334,8 @@ providers:
 `server` and `application` will become new folders in Grafana menu.
 
 > **Note.** `folder` and `folderUid` options should be empty or missing to make `foldersFromFilesStructure` work.
+
+> **Note.** To provision dashboards to the `General` folder, you should store them in the root of your `path`.
 
 ## Alert Notification Channels
 
--- a/docs/sources/administration/provisioning.md
+++ b/docs/sources/administration/provisioning.md
@@ -275,7 +275,7 @@ providers:
 
 When Grafana starts, it will update/insert all dashboards available in the configured path. Then later on poll that path every **updateIntervalSeconds** and look for updated json files and update/insert those into the database.
 
-> **Note.** Dashboards will be provisioned to the `General` folder if the `folder` option is missing or empty.
+> **Note:** Dashboards are provisioned to the General folder if the `folder` option is missing or empty.
 
 #### Making changes to a provisioned dashboard
 
@@ -335,7 +335,7 @@ providers:
 
 > **Note.** `folder` and `folderUid` options should be empty or missing to make `foldersFromFilesStructure` work.
 
-> **Note.** To provision dashboards to the `General` folder, you should store them in the root of your `path`.
+> **Note:** To provision dashboards to the General folder, store them in the root of your `path`.
 
 ## Alert Notification Channels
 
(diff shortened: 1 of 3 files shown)


docs/sources/administration/provisioning.md
```diff
@@ -275,6 +275,8 @@ providers:
 
 When Grafana starts, it will update/insert all dashboards available in the configured path. Then later on poll that path every **updateIntervalSeconds** and look for updated json files and update/insert those into the database.
 
+> **Note.** Dashboards will be provisioned to the `General` folder if the `folder` option is missing or empty.
```
oddlittlebird, line 278:
```suggestion
> **Note:** Dashboards are provisioned to the General folder if the `folder` option is missing or empty.
```

docs/sources/administration/provisioning.md
```diff
@@ -333,6 +335,8 @@ providers:
 
 > **Note.** `folder` and `folderUid` options should be empty or missing to make `foldersFromFilesStructure` work.
 
+> **Note.** To provision dashboards to the `General` folder, you should store them in the root of your `path`.
```
oddlittlebird, line 338:
```suggestion
> **Note.:** To provision dashboards to the General folder, store them in the root of your `path`.
```

aknuds1:
Having a look.

pkg/models/dashboards.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/grafana/grafana/pkg/util"
 )
 
+const RootFolderName = "General"
```
aknuds1, line 15: I think this makes sense, because it's better to use `const` when a variable is actually a constant.

pkg/models/dashboards.go
```diff
@@ -141,7 +142,7 @@ func (d UpdatePluginDashboardError) Error() string {
 	return "Dashboard belong to plugin"
 }
 
-var (
+const (
```
aknuds1, line 145: Great, can you please also remove `DashTypeJson` and `DashTypeScript` though? They're not in use.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -148,7 +146,7 @@ func (fr *FileReader) storeDashboardsInFoldersFromFileStructure(filesFoundOnDisk
 
 		folderID, err := getOrCreateFolderID(fr.Cfg, fr.dashboardProvisioningService, folderName)
 		if err != nil && err != ErrFolderNameMissing {
-			return err
+			return fmt.Errorf("can't provision %s folder from files structure: %w", folderName, err)
```
aknuds1, line 149: I think it's better to quote the folder name in case it's got whitespace, and it should probably be "file system structure":
```suggestion
			return fmt.Errorf("can't provision folder %q from file system structure: %w", folderName, err)
```

aknuds1 (changes requested):
LGTM with some suggestions.

pkg/services/provisioning/dashboards/file_reader.go
nabokihms: Yep, file, you are right. 
As for quote marks, I tried it because you had mentioned it in one of my previous PRs, and quote marks were looking like this.
```
EROR[08-03|23:33:23] A service failed                         logger=server err="Failed to provision dashboards: Failed to provision config default: can't provision \"General\" folder from files structure: A folder with that name already exists"
```
But maybe it's better to use them instead of thinking about spaces.

aknuds1: @nabokihms Yeah double quotes will get escaped in this logging format, it's no big deal. It's better than being confused by spaces :)

aknuds1 (approved):
LGTM! Thanks!

oddlittlebird:

PR #26777, opened 2020-08-04
Stats: Stop counting the same user multiple times

### LINKED ISSUE #26744: Improve user stats
We need to rethink how we query for the role based user counts:

https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/stats.go#L83 

and 

https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/stats.go#L195 

We count a user role more than once if he is Admin in 2 orgs, and Editor in 3 orgs we count that as 3 editors and 2 admins, but if it's the same user we should only count this as 1 Admin  (the highest role in any org). So a single user id should only be counted once and not be included as part of the number of any of the other stats. 

So Total users =  admins +  editors + viewers
Same for active users. 

To fix this we might have to abandon the current combined mega SQL stats query and make this a separate query that does this calculation or figure out an in-expensive way to do it in current method.

sakjur:

Fixes #26744


pkg/services/sqlstore/stats.go
```diff
@@ -192,23 +184,118 @@ func GetSystemUserCountStats(ctx context.Context, query *models.GetSystemUserCou
...
+
+				if user.LastSeenAt.After(activeUserDeadline) {
+					activeMap[user.ID] = user.Role
+				}
+				totalMap[user.ID] = user.Role
+			}
+
+			memo.active = mapToStats(memo.active, activeMap)
+			memo.total = mapToStats(memo.total, totalMap)
+		}
```
papagian, line 275: We could have a debug message if last `len(usersSlice) == userStatsPageSize` that most probably the `userStatsPageLimit` has been reached?

pkg/services/sqlstore/stats.go
```diff
@@ -192,23 +184,118 @@ func GetSystemUserCountStats(ctx context.Context, query *models.GetSystemUserCou
 	})
 }
 
-func GetActiveUserStats(query *models.GetActiveUserStatsQuery) error {
-	activeUserDeadlineDate := time.Now().Add(-activeUserTimeLimit)
-	sb := &SqlBuilder{}
+func GetUserStats(query *models.GetUserStatsQuery) error {
+	err := updateUserRoleCountsIfNecessary(query.MustUpdate)
```
papagian, line 188: I wonder if there is a use case that the  `GetAdminStats` has to force an update similar to what happens here.

papagian (dismissed):
LGTM

As a minor improvement, even though 5 million users is a large number, I would suggest adding a debug message if the this limit is reached.

pkg/services/sqlstore/stats.go
sakjur: I'll add that!

pkg/services/sqlstore/stats_test.go
```diff
@@ -5,67 +5,129 @@ import (
...
+	})
+}
+
+func TestGetUserStats(t *testing.T) {
+	InitTestDB(t)
+
+	cmd := &models.CreateUserCommand{
+		Email:   fmt.Sprintf("admin@test.com"),
+		Name:    fmt.Sprintf("Admin"),
+		Login:   fmt.Sprintf("admin"),
```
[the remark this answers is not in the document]
sakjur: Haha, I have this exact commit locally 🙈

pkg/services/sqlstore/stats.go
sakjur: That's a good question! Maybe?

kylebrandt:
I'm not too familiar with this code so I might be following it wrong - but my main concern would be load against the DB. Are we scanning through the whole table? If so, perhaps it is better to do the counting in SQL? (But perhaps I'm reading this wrong).

papagian:
> I'm not too familiar with this code so I might be following it wrong - but my main concern would be load against the DB. Are we scanning through the whole table? If so, perhaps it is better to do the counting in SQL? (But perhaps I'm reading this wrong).

I'm afraid that this is not possible because if the user has several roles then we want to count the higher one (if it's admin and editor we would count the user only in the admins). Things would be better if the role type was a number (with correct ordering) instead of string in the database.

kylebrandt:
> I'm afraid that this is not possible because if the user has several roles then we want to count the higher one (if it's admin and editor we would count the user only in the admins). Things would be better if the role type was a number (with correct ordering) instead of string in the database.

Maybe this is particular to MSSQL? But I think I recall doing CASE statements for stuff like this - would that not work here? https://stackoverflow.com/a/14189418/107156

sakjur:
@kylebrandt @papagian You gave me an idea, it works for SQLite, but I guess I might have to tweak the logic for the other dialects (particularly MySQL has a tendency of being different from SQLite syntax).

pkg/services/sqlstore/stats.go
```diff
@@ -192,23 +182,108 @@ func GetSystemUserCountStats(ctx context.Context, query *models.GetSystemUserCou
...
+                   ELSE 1
+                   END AS 'role',
+               CASE
+                WHEN u.last_seen_at>? THEN ` + dialect.BooleanStr(true) + `
+                ELSE ` + dialect.BooleanStr(false) + `
+               END AS active
+           FROM user AS u LEFT JOIN org_user ON org_user.user_id = u.id
+           GROUP BY u.id, u.last_seen_at, org_user.role)
+     GROUP BY active, id)
+GROUP BY active, role;`
```
papagian, line 241: Smart! I think with a few adjustments works against all the databases (sqlite, postgres, mysql)
```suggestion
	userTable := "user"
	if x.Dialect().DriverName() == migrator.POSTGRES {
		userTable = `"user"`
	}
	query := fmt.Sprintf(`
SELECT role AS bitrole, active, COUNT(role) AS count FROM
    (SELECT active, SUM(role) AS role
     FROM (SELECT
               u.id,
               CASE org_user.role
                   WHEN 'Admin' THEN 4
                   WHEN 'Editor' THEN 2
                   ELSE 1
                   END AS role,
               CASE
                WHEN u.last_seen_at>? THEN `+dialect.BooleanStr(true)+`
                ELSE `+dialect.BooleanStr(false)+`
               END AS active
           FROM %s AS u LEFT JOIN org_user ON org_user.user_id = u.id
           GROUP BY u.id, u.last_seen_at, org_user.role) AS t2
     GROUP BY active, id) AS t1
GROUP BY active, role;`, userTable)
```

pkg/services/sqlstore/stats.go
```diff
@@ -2,6 +2,7 @@ package sqlstore
 
 import (
 	"context"
+	"strconv"
```
papagian, line 5:
```suggestion
	"fmt"
	"strconv"
	
	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
```

sakjur: :( Augh, again with the "oh, I thought I had pushed that commit". I went for a slightly different route, I think this should work over the different SQLs without needing `migrator`. I'll move this back to being a draft PR.

papagian: I agree that your approach with `dialect.Quote` is much more preferable

pkg/services/sqlstore/stats.go
```diff
@@ -192,23 +182,108 @@ func GetSystemUserCountStats(ctx context.Context, query *models.GetSystemUserCou
...
+               u.id,
+               CASE org_user.role
+                   WHEN 'Admin' THEN 4
+                   WHEN 'Editor' THEN 2
+                   ELSE 1
+                   END AS role,
+               CASE
+                WHEN u.last_seen_at>? THEN ` + dialect.BooleanStr(true) + `
+                ELSE ` + dialect.BooleanStr(false) + `
+               END AS active
```
sakjur, line 237: Note to self, this should be the same thing:

```suggestion
               u.last_seen_at>? AS active
```

xlson (approved):
Great work on a hairy beast!

PR #26779, opened 2020-08-04
Chore: Simplify generic OAuth code

aknuds1:
In order to lay the groundwork for OAuth org membership mapping, I'm simplifying the generic OAuth code, so it's more straightforward and deterministic (helps in extending the code with more functionality). Also adding more debug logging and tweaking some of the corresponding tests (was helpful in figuring out what was going on while refactoring).

As far as I can tell, we are currently lacking a standard way of providing a debug enabled logger during tests, so I added a function to create one (`log.NewWithLevel`). Please point it out if there is already another such facility. I basically need to see some logs while testing and figuring out bugs.

### DIFF
--- a/pkg/login/social/generic_oauth.go
+++ b/pkg/login/social/generic_oauth.go
@@ -82,6 +82,7 @@ type UserInfoJson struct {
 	Upn         string              `json:"upn"`
 	Attributes  map[string][]string `json:"attributes"`
 	rawJSON     []byte
+	source      string
 }
 
 func (info *UserInfoJson) String() string {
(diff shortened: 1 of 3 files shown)


pkg/login/social/generic_oauth.go
```diff
@@ -91,64 +92,97 @@ func (info *UserInfoJson) String() string {
...
+			if data.Name != "" {
+				s.log.Debug("Setting user info name from name field")
+				userInfo.Name = data.Name
+			} else if data.DisplayName != "" {
+				s.log.Debug("Setting user info name from display name field")
+				userInfo.Name = data.DisplayName
+			}
+		}
+
+		if userInfo.Login == "" {
```
alexanderzobnin, line 117: Isn't better to extract this into separate function like `extractLogin(data *UserInfoJson)`?

alexanderzobnin (approved):
Looks good! Maybe it makes sense to split `UserInfo()` func into few smaller functions to increase readability.

pkg/login/social/generic_oauth.go
aknuds1: @alexanderzobnin I think it depends on personal preference; I follow the old rule of thumb of generally not making functions longer than the height of a screen, and until that point it's usually preferable to inline, because the more indirection you have the more difficult it's to review if the code is correct.

alexanderzobnin: It's probably not a big deal in this case because function is quite straightforward and it's easy to follow the code even if it's more than one screen.

aknuds1: @alexanderzobnin Thanks, I take it you don't mind if I merge then? As you say, the code is relatively straightforward after all and not that difficult to follow.

aknuds1: I'll merge it now - if you want, we can make another PR to refactor the function a little :)

PR #26785, opened 2020-08-04
SQL: Fix missing org error messages

aknuds1:
Fix error messages (a log message and an actual error) when an org can't be created.


kylebrandt (approved):

papagian (approved):
LGTM

PR #26788, opened 2020-08-04
Test datasource: fix query editor html partial

ryantxu:
The query editor for test datasource is only showing options for *some* of the query types.  This PR fixes some invalid html elements that cause unpredictable html rendering


torkelo (approved):

kylebrandt (approved):
Thank you :-)

Issue #26762, opened 2020-08-03
DataLinks from data sources override user defined data link


LukeDearden:
**What happened**: Updated from 7.0.6 to 7.1.1 and DataLinks from Stat Panels on our dashboard using Log Analytics queries all reverted to "View in Azure Portal" ignoring the existing Datalink

**What you expected to happen**: It should render the Datalink correctly and open the correct link in a new tab

**How to reproduce it (as minimally and precisely as possible)**: Create a Panel that uses an Azure Log Analytics query and try set and follow the Datalink

**Anything else we need to know?**: Panel Links still work fine

**Environment**:
- Grafana version: v7.1.1 (3039f9c3bd)
- Data source type & version: Azure Log Analytics
- OS Grafana is installed on: Ubuntu 16.04.6 LTS
- User OS & Browser: Edge Chromium/FireFox 
- Grafana plugins: Azure Monitor 0.3.0

Example
![datalinks](https://user-images.githubusercontent.com/6813601/89162080-35977200-d57c-11ea-84f7-b2432b039d0b.png)

Actual link
![datalinks2](https://user-images.githubusercontent.com/6813601/89162547-ebfb5700-d57c-11ea-8951-537b23b07b72.png)

- Others:

aknuds1:
@kylebrandt Is this related to your work?

kylebrandt:
I can replicate this. I assume this is because we are sending the Azure portal link as a Datalink from the backend, and this situation runs into this issue.

There are two different places to set data links:
 1. The "Field" Tab
 2. The "Field Overrides tab"

In the Field tab, adding a new datalink seems to have no impact as described in the issue. 

In the Field Overrides tab, setting the datalink replaces the "View in Azure Portal" datalink that is returned from the backend ( @LukeDearden this might provide with a workaround - although I imagine not ideal ).

The override behavior seems like it good be okay to me, but since data links is plural (and is an array) - I think that when the Field tab, adding a data link should append to any existing data links provided by the backend. Are you able to look into this @ryantxu ?

nickjacobsen:
I believe the underlying issue is lines 173 - 179
of
public/app/plugins/datasource/grafana-azure-monitor-datasource/azure_log_analytics/azure_log_analytics_datasource.ts

Where rather than appending to the links array it just clobbers it.

   field.config.links = [
                  {
                    url: url,
                    title: 'View in Azure Portal',
                    targetBlank: true,
                  },
                ];

[...]

PR #26791, opened 2020-08-04
Overrides: support an append mode

ryantxu:
when applying field overrides to links, we really want to keep both values.

This PR explores how we can have an append mode.  In this iteration it is configured globally for the field, but it would be better if this were part of the UI

See https://github.com/grafana/grafana/issues/26762
image

### DIFF
--- a/packages/grafana-data/src/field/fieldOverrides.ts
+++ b/packages/grafana-data/src/field/fieldOverrides.ts
@@ -36,6 +36,7 @@ import { formatLabels } from '../utils/labels';
 import { getFrameDisplayName, getFieldDisplayName } from './fieldState';
 import { getTimeField } from '../dataframe/processDataFrame';
 import { mapInternalLinkToExplore } from '../utils/dataLinks';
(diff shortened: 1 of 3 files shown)


torkelo (commented):
Think this looks ok, to solve the UI option to replace the default/data source array items. Tricky, first thought was an option on the first item "replace", but that only works if you want a datalink / something. If you just want to clear the array that won't work. Hm.. maybe we can have an item property like:

```ts
override: 'replace' | 'clear' | 'append' 
```

The UI would be a bit wonky, as it's options on an item. I don't know, tricky.

kylebrandt:
For https://github.com/grafana/grafana/issues/26762

dprokop:
image

I think something like this could work for the UI to expose the option to replace/append. Not sure about the choice of the icons, but the radio button feels like the right element to use.

kylebrandt:
Just to tie this back to this initial reported issue. If we were to change the behavior to append, and that was the only change, that would probably be "good enough" to unbreak users adding datalinks when the datasource also sets one. Although given 7.2, I imagine there is time to do these things, but if it turns out there is not, please consider changing the default behavior to append as a start.

Issue #26470, opened 2020-07-20
Data column name is now part of field name in 7.1 with Azure Log Analytics


djryanj:
**What happened**: 

In Grafana 6.7 with Azure Log Analytics (Azure Monitor) as a data source, the data table column name that Grafana used was transparently removed from the UI with no extra effort, which resulted in clean looking legend labels like this:

![image](https://user-images.githubusercontent.com/33379396/87989725-c9a70b00-ca9f-11ea-8063-d39a8eaca225.png)

Part of the labels have been removed in this image but are identical; for example, whole names might be APP-01 and APP-02. **Note that the fact that the query returns multiple data sets is not relevant.** The same occurs if selecting even a single VM (see sample KQL below).

In 7.1 (possibly 7.0), the move to dataframes has resulted in the labels now including the metric name, like this:

![image](https://user-images.githubusercontent.com/33379396/87989799-ed6a5100-ca9f-11ea-9d5b-85081c460024.png)

Again, the part after the "avg_CounterValue" would be identical like "APP" as in the first image above.

In Azure's native Log Analytics tool, the data table returned looks like this:

![image](https://user-images.githubusercontent.com/33379396/87989933-2b677500-caa0-11ea-8296-2eae8d4db0ea.png)

This is consistent with the data format in the query inspector in 6.7:

![image](https://user-images.githubusercontent.com/33379396/87990080-8600d100-caa0-11ea-900f-be5a70a0623b.png)

But in the query inspector in 7.1, we see:

![image](https://user-images.githubusercontent.com/33379396/87990166-b3e61580-caa0-11ea-847c-4719315a4107.png)

And we can see where the series labels are coming from. This "intelligent assembly" of the dataframe data has removed flexibility in how we can display metric labels on panels.

Using a field selector in the display name options does not work, as using ${__field.name} only returns the exact same column name as before, and ${__series.name} results in labels like "Series (A)", and ${__cell_N} or variants like ${**cell_N} do not resolve.

**What you expected to happen**:

No change to the label names from 6.7.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a Windows VM in Azure and ensure that it is logging performance counters to Log Analytics workspace
2. Connect Grafana 7.1 Azure Monitor (0.3.0) datasource to that Log Analytics workspace
3. In a graph panel, using Log Analytics as a data source, use the following KQL (replace "testVM" as needed to match the VM name created/available in step 1):

```
Perf  
[...]

PR #26804, opened 2020-08-05
Docs: Azure logs - cross reference field configuration display name

kylebrandt:
Cross reference Field configuration options added in 7.1.2

Documentation updates for https://github.com/grafana/grafana/issues/26470

### DIFF
--- a/docs/sources/features/datasources/azuremonitor.md
+++ b/docs/sources/features/datasources/azuremonitor.md
@@ -268,7 +268,7 @@ Perf
 | order by TimeGenerated asc
 ```
 
-> **Tip**: In the above query, the Kusto syntax `Samples=count()` and `["Avg Value"]=...` is used to rename those columns — the second syntax allowing for the space. This will change the name of the metric that Grafana uses, and as a result, things like series legends and table columns will match what you specify. In this case, `Samples` will be displayed instead of `_count`.
+> **Tip**: In the above query, the Kusto syntax `Samples=count()` and `["Avg Value"]=...` is used to rename those columns — the second syntax allowing for the space. This changes the name of the metric that Grafana uses, and as a result, things like series legends and table columns will match what you specify. Here `Samples` is displayed instead of `_count`.
 
 {{< docs-imagebox img="/img/docs/azuremonitor/logs_multi-value_multi-dim.png" class="docs-image--no-shadow" caption="Azure Logs query with multiple values and multiple dimensions" >}}
 
--- a/docs/sources/features/datasources/azuremonitor.md
+++ b/docs/sources/features/datasources/azuremonitor.md
@@ -263,10 +263,12 @@ You can also select additional number value columns (with, or without multiple d
 ```kusto
 Perf
 | where $__timeFilter(TimeGenerated)
-| summarize Samples=count(), AvgValue=avg(CounterValue)
+| summarize Samples=count(), ["Avg Value"]=avg(CounterValue)
     by bin(TimeGenerated, $__interval), Computer, CounterName, InstanceName
 | order by TimeGenerated asc
 ```
+
+> **Tip**: In the above query, the Kusto syntax `Samples=count()` and `["Avg Value"]=...` is used to rename those columns — the second syntax allowing for the space. This will change the name of the metric that Grafana uses, and as a result, things like series legends and table columns will match what you specify. In this case, `Samples` will be displayed instead of `_count`.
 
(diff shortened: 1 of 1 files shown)


docs/sources/features/datasources/azuremonitor.md
```diff
@@ -281,6 +281,16 @@ AzureActivity
 | order by TimeGenerated desc
 ```
 
+### Format legend keys with aliases for Log Analytics
+
+The default legend formatting is:
+
+`metricName{dimensionName=dimensionValue,dimensionTwoName=DimensionTwoValue}`
+
+This can be customized by using the [display name field configuration option]({{< relref "../../panels/field-configuration-options.md#display-name" >}}).
```
oddlittlebird, line 290:
```suggestion
You can customize the legend formatting by setting the [display name field configuration option]({{< relref "../../panels/field-configuration-options.md#display-name" >}}).
```

docs/sources/features/datasources/azuremonitor.md
```diff
@@ -281,6 +281,16 @@ AzureActivity
...
+
+In order to change the metric name, you can change your query to specify a specific name. For example, `summarize avg(Active) by...` will return metrics with the name `avg_Active`. This can be changed to `Active` by changing that part of the query to be `summarize ["Active"]=avg(Active) by ...`.
```
oddlittlebird, line 292: Where are they changing the metric name? In the visualization? The legend? Both?

Also, instead of "this can be...", try writing "You can <do this>." It is both more active and often uses fewer words.

oddlittlebird (changes requested):

docs/sources/features/datasources/azuremonitor.md
kylebrandt: The metric name is being changed in the data source query (a text field style query editor). So this will change the results returned from query, and therefore impact everything upstream.  I am not sure of a better way to say that than "your query"?

kylebrandt: So by SQL analogy, this would be similar to changing `Select avg(someColumn) FROM ...` to `SELECT avg(someColumn) AS Active FROM ...`

oddlittlebird: Try to write in present tense. "returns" rather than "will return", for example

oddlittlebird: For the query, I am trying to put myself in the user's shoes. As a user, why would I want to do this? Where would I see this change reflected?

Am I trying to change the variable name in the query so that my visualization is easier to understand, perhaps?

kylebrandt: This is more of a tip about the query language. So I have removed it. Also the title was wrong (not an alias) so hopefully this removes some of the confusion.

oddlittlebird: I think it's a great tip, just confusing where you had it. Consider adding a note or brief subsection somewhere else in this document, because I see this question A LOT in the community forums for various data sources.

oddlittlebird (approved):
Looks good. Please consider my suggestion about adding the note elsewhere in the document.

kylebrandt:
> Looks good. Please consider my suggestion about adding the note elsewhere in the document.

Added it back, I think this is a better place?

docs/sources/features/datasources/azuremonitor.md
```diff
@@ -263,11 +263,13 @@ You can also select additional number value columns (with, or without multiple d
 ```kusto
 Perf
 | where $__timeFilter(TimeGenerated)
-| summarize Samples=count(), AvgValue=avg(CounterValue)
+| summarize Samples=count(), ["Avg Value"]=avg(CounterValue)
     by bin(TimeGenerated, $__interval), Computer, CounterName, InstanceName
 | order by TimeGenerated asc
 ```
 
+> **Tip**: In the above query, the Kusto syntax `Samples=count()` and `["Avg Value"]=...` is used to rename those columns — the second syntax allowing for the space. This will change the name of the metric that Grafana uses, and as a result, things like series legends and table columns will match what you specify. In this case, `Samples` will be displayed instead of `_count`.
```
oddlittlebird, line 271:
```suggestion
> **Tip**: In the above query, the Kusto syntax `Samples=count()` and `["Avg Value"]=...` is used to rename those columns — the second syntax allowing for the space. This changes the name of the metric that Grafana uses, and as a result, things like series legends and table columns will match what you specify. Here `Samples` is displayed instead of `_count`.
```

oddlittlebird (approved):
Looks great!

oddlittlebird:
One minor grammar suggestion, then good to go!

PR #23661, opened 2020-04-17
OAuth: Support mapping generic OAuth users to Grafana organizations

treydock:
**What this PR does / why we need it**:
Add `orgs_attribute_path` to generic oauth configuration similar to `role_attribute_path` but for mapping a user to one or many Grafana organizations.

**Which issue(s) this PR fixes**:

Fixes #23473  and would potentially make #22605 a non-issue for me.

**Special notes for your reviewer**:

Currently using `auto_assign_org_id` doesn't work with OAuth when the desire is to have people login in to an org that's set with `auto_assign_org_id`, and then assigning them after the fact to a different org (#22605). To have maximum flexibility of putting people in multiple organizations I opted to make the values comma delimited.

PR #26806, opened 2020-08-05
WIP: OAuth: Support mapping from claims to org memberships

aknuds1:
WIP PR to discuss an alternate method of mapping OAuth claims to org memberships.

Based on #23661, but exists to explore an alternative way of doing it.


aknuds1:
Dup.

PR #26810, opened 2020-08-05
Add guide on adding support for variable queries

marcusolsson:

Adds docs for how to adding support for variable queries in a data source plugin.

### DIFF
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -59,9 +59,9 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
    }
    ```
 
-## Add support for variable queries to your data source
+## Add support for query variables to your data source
 
-[Query variables]({{< relref "../../variables/variable-types/add-query-variable.md" >}}) is a type of variable that allows you to query a data source for the values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
+[Query variables]({{< relref "../../variables/variable-types/add-query-variable.md" >}}) is a type of variable that allows you to query a data source for the values. By adding support for query variables to your data source plugin, users can create dynamic dashboards based on data from your data source.
 
 Let's start by defining a query model for the variable query.
 
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -77,7 +77,7 @@ For a data source to support query variables you need to override the [`metricFi
 ```ts
 async metricFindQuery(query: MyVariableQuery, options?: any) {
   // Retrieve DataQueryResponse based on query.
-  const response = await this.doQuery(query);
+  const response = await this.fetchMetricNames(query.namespace, query.rawQuery);
 
   // Convert query results to a MetricFindValue[]
   const values = response.data.map(frame => ({ text: frame.name }));
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -61,7 +61,7 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
 
 ## Add support for variable queries to your data source
 
-[Query variables]({{< relref "../../variables/variable-types/add-query-variable/" >}}) is a type of variable that allows you to query a data source for the values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
+[Query variables]({{< relref "../../variables/variable-types/add-query-variable.md" >}}) is a type of variable that allows you to query a data source for the values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
 
 Let's start by defining a query model for the variable query.
 
@@ -152,3 +152,5 @@ Let's create a custom query editor to allow the user to edit the query model.
      .setQueryEditor(QueryEditor)
      .setVariableQueryEditor(VariableQueryEditor);
    ```
+
+That's it! You can now try out the plugin by adding a [query variable]({{< relref "../../variables/variable-types/add-query-variable.md" >}}) to your dashboard.
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -72,7 +72,7 @@ export interface MyVariableQuery {
 }
 ```
 
-For a data source to support query variables you need to override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`.
+For a data source to support query variables, you must override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`:
 
 ```ts
 async metricFindQuery(query: MyVariableQuery, options?: any) {
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -63,10 +63,19 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
 
 [Query variables]({{< relref "../../variables/variable-types/add-query-variable/" >}}) is a type of variable that allows you to query a data source for the values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
 
-For a data source to support query variables you need to override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your data source class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`.
+Let's start by defining a query model for the variable query.
 
 ```ts
-async metricFindQuery(query: any, options?: any) {
+export interface MyVariableQuery {
+  namespace: string;
+  rawQuery: string;
+}
+```
+
+For a data source to support query variables you need to override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`.
+
+```ts
+async metricFindQuery(query: MyVariableQuery, options?: any) {
   // Retrieve DataQueryResponse based on query.
   const response = await this.doQuery(query);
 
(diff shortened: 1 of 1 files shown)


ryantxu (approved):
Can you add the return type [Promise<MetricFindValue[]>](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/datasource.ts#L222)?

The options and query are the real trick and I still don't totally understand them :cry:

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,21 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+
+```ts
+async metricFindQuery(query: any, options?: any) {
+  // Retrieve data based on query.
+  const result = await this.doQuery(query);
+
+  // Return the name of each series as the value.
+  const values = result.map(series => ({ text: series.name }));
+
+  return values;
```
ryantxu, line 76: maybe something more pseudo code?  since results is hopefully a DataQueryResponse?

```suggestion
  const result = await this.doQuery(query);
  
  // Convert query results to a MetricFindValue[]
  return result.map(series => ({ text: series.name }));
```

marcusolsson:
> The options and query are the real trick and I still don't totally understand them 😢

Me neither. Since `query` is of `any` type, I assume you can pass more advanced queries, but I've only seen examples of strings. I've seen it mentioned somewhere that you can create a custom query editor, but haven't been able to find any examples of that.

marcusolsson:
Found some resources! Will update this PR as I learn more!

oddlittlebird:
Cool beans, @marcusolsson ! Please tag me when you're ready for my review.

marcusolsson:
@ryantxu I've added instructions on how to use a custom query editor. Could you take another look?

marcusolsson:
@oddlittlebird Think this is ready for you now!

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+  rawQuery: string;
+}
+```
+
+For a data source to support query variables you need to override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`.
+
+```ts
+async metricFindQuery(query: MyVariableQuery, options?: any) {
+  // Retrieve DataQueryResponse based on query.
+  const response = await this.doQuery(query);
```
alexanderzobnin, line 80: Maybe we can highlight using `query` as an object, because in case of string query we don't need custom editor.
```suggestion
  const response = await this.fetchMetricNames(query.namespace, query.rawQuery);
```

alexanderzobnin (approved):
Looks good! Just a few potential things to think on.

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: I like it!

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+Let's start by defining a query model for the variable query.
+
+```ts
+export interface MyVariableQuery {
+  namespace: string;
+  rawQuery: string;
+}
+```
+
+For a data source to support query variables you need to override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`.
```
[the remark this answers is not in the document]
marcusolsson: That's interesting indeed. I don't think we should add it here until we've updated the interface. Would you mind creating an issue and linking it to this PR?

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
      return { data };
    }
    ```
+
+## Add support for variable queries to your data source
```
oddlittlebird, line 62: Do you mean add support for _query variables_ to your data source? Or for using variables in queries? Unclear.

Also, will this be just template variables (defined by the user) or does this include global variables as well?

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+
+[Query variables]({{< relref "../../variables/variable-types/add-query-variable/" >}}) is a type of variable that allows you to query a data source for the values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
```
oddlittlebird, line 64:
```suggestion
[Query variables]({{< relref "../../variables/variable-types/add-query-variable/" >}}) are allows you to query a data source for values. By adding support for variable queries to your data source plugin, users can create dynamic dashboards based on data from your data source.
```

oddlittlebird: Do we want to mention that this would be necessary for Repeated panels/rows and perhaps link to that topic as well?

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird, line 75:
```suggestion
For a data source to support query variables, you must override the [`metricFindQuery`]({{< relref "../../packages_api/data/datasourceapi.md#metricfindquery-method" >}}) in your `DataSourceApi` class. `metricFindQuery` returns an array of [`MetricFindValue`]({{< relref "../../packages_api/data/metricfindvalue.md" >}}) which has a single property, `text`:
```

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+  const response = await this.fetchMetricNames(query.namespace, query.rawQuery);
+
+  // Convert query results to a MetricFindValue[]
+  const values = response.data.map(frame => ({ text: frame.name }));
+
+  return values;
+}
+```
+
+> **Note:** By default, Grafana provides a default query model and editor for simple text queries. If that's all you need, you can leave the query type as `string`.
```
oddlittlebird, line 89:
```suggestion
> **Note:** By default, Grafana provides a default query model and editor for simple text queries. If that's all you need, then you can leave the query type as `string`.
```

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+>
+> ```ts
+> async metricFindQuery(query: string, options?: any)
+> ```
+
+Let's create a custom query editor to allow the user to edit the query model.
+
+1. Create a `VariableQueryEditor` component.
```
oddlittlebird, line 97: I'm a little confused by your ending punctuation on the steps and sentences preceding code blocks. Either a colon or period could be correct, but I thought you mostly used colons in previous topics.

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+           <input name="rawQuery" className="gf-form-input" onBlur={saveQuery} onChange={handleChange} value={state.rawQuery} />
+         </div>
+       </>
+     );
+   };
+   ```
+
+   Grafana saves the query model whenever one of the text fields loses focus (`onBlur`) and then previews the values returned by `metricFindQuery`.
+
+   The second argument to `onChange` allows you to set a text representation of the query which will appear next to the name of the variable in the variables list.
```
oddlittlebird, line 144: I recommend putting the explanation for what the code block does above the code block or as code comments in the code block.

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -58,3 +58,97 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+
+1. Finally, configure your plugin to use the query editor.
+
+   ```ts
+   import { VariableQueryEditor } from './VariableQueryEditor';
+
+   export const plugin = new DataSourcePlugin<DataSource, MyQuery, MyDataSourceOptions>(DataSource)
+     .setQueryEditor(QueryEditor)
+     .setVariableQueryEditor(VariableQueryEditor);
+   ```
```
oddlittlebird, line 154: Do we want to add a testing step? How does the user know they are done and that they have been successful?

Something like, "Open or restart Grafana. You should now be able to use query variables in queries to your data source."

Is there anything that might go wrong that we might want to warn the user about?

One big question I have as a user is, "What about the other types of variables? Do I need to do something special to make them work in my data source?"

oddlittlebird (changes requested):
This is great stuff. Query variables are super powerful, so I imagine a lot of plugin developers will be delighted to see this.

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: I was going back and forth on this. The feature is called Query variables, but from the plugin's perspective it's about adding support for "Query variables queries".

This is only for creating template variables. I don't believe plugins can create global variables themselves.

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: Hm, I don't think this piece needs to explain the different use cases for variables. 

Did you mean `Query variables are _variables that_ allow you ...`?

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: I put them before initially, but since they are more notes/comments, I felt they wasn't valuable until the reader has read the code block.

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: I'll look at adding a closing step to verify it works.

The previous sections in this document explain how to use variables set by the user, no matter the type, and interpolate them in your data source. 

This section explains how to make a data source populate a variable that they can use for _other_ panels and data sources.

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: For the other plugin guides, I've used a period. If you find a colon, I'll fix it.

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird: I actually meant "Query variables allow you to..." but your way works as well.

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird: A good rule is to explain to the reader what they are reading before they read it, not after.

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird: How about, "Add support for query variables to your data source"?

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird: Sounds good, thanks for clarifying. 

We should add this rule to the style guide of when to colon vs. when not to colon.

oddlittlebird (approved):
One suggestion left for you to consider, then good to go!

docs/sources/developers/plugins/add-support-for-variables.md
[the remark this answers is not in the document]
sebbo2002: I just stumbled upon this. Is there a follow-up issue on this? Didn't find one through the search.

PR #26817, opened 2020-08-05
Grafana Toolkit: add an option for signing plugins as admin

leventebalogh:
### The problem
When the `grafana-toolkit plugin:ci-package` command is called it is communicating with the `grabpl` executable under the hood in order to get a signed MANIFEST.txt for the plugin. `grabpl` has an option for exceptional cases - it can sign a plugin without having it in the plugin database by providing an extra flag. Unfortunately this flag is not "proxied" by the toolkit at this point.

### The solution
Add a `--signing-admin` flag to the `grafana-toolkit plugin:ci-package` command.

### The future
As we discussed with @ryantxu the Toolkit is probably not the best place for things that are closely CI related, so we are planning to lift it out in the future.

### DIFF
--- a/packages/grafana-toolkit/src/cli/index.ts
+++ b/packages/grafana-toolkit/src/cli/index.ts
@@ -194,9 +194,12 @@ export const run = (includeInternalScripts = false) => {
 
   program
     .command('plugin:ci-package')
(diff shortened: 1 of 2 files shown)


ryantxu (approved):
Looks good -- the better solution is to remove all the `ci-*` tasks from toolkit and have them in the build tool... but lets save that for another day

PR #26829, opened 2020-08-06
Templating: Adds -- remove filter -- back to incomplete AdHoc filters

### LINKED ISSUE #26104: "-- remove filter --" option is not present for ad hoc variables unless looking at a "complete" adhoc variable
**What happened**: Began creating a new ad hoc variable, but stopped part way through. When attempting to clean up this variable, the "--remove filter--" option is not present, with no indication of why that would be the case.

**What you expected to happen**: "--remove filter--" should always be there, which I believe is how it used to work.

**How to reproduce it (as minimally and precisely as possible)**: Create the "left half" of an adhoc variable, then click on that left half again and attempt to remove it.

**Anything else we need to know?**: This probably sounds like a trivial issue, but I am surprised how often users are running into it after we upgraded to 7. They have gotten used to using ad hoc variables as a way of exploring the data, and do not necessarily always intend to actually filter data with the result. Or, they thought they wanted to apply a filter, but found that the options in the "right half" did not have what they were looking for.

However users get into this state, once you are there, it is very confusing. How to get rid of your incomplete ad hoc filter is not at all intuitive at this point (either a full page refresh, or writing some value in the right half of the variable, applying a filter you did not want applied).

If you don't know how to fix it, it is very frustrating to have this item that simply can't be cleared. If you know how to fix it, it's a little bit of extra friction injected into your workflow.

**Environment**:
- Grafana version: 7.0.5
- User OS & Browser: Chrome

hugohaggmark:
For details, look in issue #26104. 

I did a medium refactoring pushing the data fetching down to the components `AdHocFilterKey` and `AdHocFilterValue`. I also introduced `AdHocFilterRenderer` because the code was almost identical in `AdHocPickerUnconnected` and `AdHocFilterBuilder`. 

Probably went crazy with the refactoring but I'm sure the reviewers will let me know :)

Fixes #26104

### DIFF
--- a/public/app/features/variables/adhoc/picker/AdHocFilterBuilder.tsx
+++ b/public/app/features/variables/adhoc/picker/AdHocFilterBuilder.tsx
@@ -1,77 +1,63 @@
-import React, { FC, ReactElement, useState } from 'react';
-import { Icon, SegmentAsync } from '@grafana/ui';
-import { OperatorSegment } from './OperatorSegment';
+import React, { FC, useCallback, useState } from 'react';
 import { AdHocVariableFilter } from 'app/features/variables/types';
 import { SelectableValue } from '@grafana/data';
+import { AdHocFilterKey, REMOVE_FILTER_KEY } from './AdHocFilterKey';
+import { AdHocFilterRenderer } from './AdHocFilterRenderer';
 
 interface Props {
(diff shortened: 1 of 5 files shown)


torkelo (approved):
Tested this and works great!

PR #26835, opened 2020-08-06
CloudWatch: Fix sorting of metrics results

### LINKED ISSUE #25494: Cloudwatch SORT metrics math function is not respected for more than 10 results

aknuds1:
Fix sorting of CloudWatch metrics query results.

Fixes #25494.

Not 100% sure if this is correct, since I'm new to this part of the code base, but I think it makes sense.


kylebrandt:
FYI we do have https://pkg.go.dev/github.com/grafana/grafana-plugin-sdk-go/data?tab=doc#Fields which sorts the fields of a wideframe to be predictable, but that doesn't preserve order. I created it for a case where the order returned from the datasource was not consistent.

aknuds1:
@kylebrandt Is that something we can use here at all, considering we're not switching to data frames as part of this PR?

pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -23,25 +24,27 @@ func (e *cloudWatchExecutor) parseResponse(metricDataOutputs []*cloudwatch.GetMe
 		}
 
 		for _, r := range mdo.MetricDataResults {
```
kylebrandt, line 26: I would insert something like:

```
id := *r.id
lablel := *r.Label
```

as the start of the inside of the loop. That should make it easier to read, and also if you need to do a nil check to make sure you are not dereferencing a nil pointer -  that would be the place to do it.

pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 func (e *cloudWatchExecutor) parseResponse(metricDataOutputs []*cloudwatch.GetMetricDataOutput, queries map[string]*cloudWatchQuery) ([]*cloudwatchResponse, error) {
-	mdr := make(map[string]map[string]*cloudwatch.MetricDataResult)
+	mdrs := make(map[string]map[string]*cloudwatch.MetricDataResult)
```
kylebrandt, line 16: I find helpful to have comment saying what the strings are on map like this, e.g. id -> label -> result.

kylebrandt (approved):
Looks okay to me, but maybe make a couple assignments to clean up all the `*` to make it more readable, also be good to be safe and make sure things are not nil before referencing them

kylebrandt:
> @kylebrandt Is that something we can use here at all, considering we're not switching to data frames as part of this PR?
> 
> I guess we can do things differently after the switch to data frames, but until then, isn't this an acceptable solution?

Ah I didn't realize that ...

aknuds1:
> Looks okay to me, but maybe make a couple assignments to clean up all the `*` to make it more readable, also be good to be safe and make sure things are not nil before referencing them

OK thanks for the suggestion @kylebrandt, will get right on it.

pkg/tsdb/cloudwatch/response_parser.go
aknuds1: Thanks @kylebrandt, does make sense indeed and I'll take care of it.

pkg/tsdb/cloudwatch/response_parser_test.go
```diff
@@ -60,7 +61,7 @@ func TestCloudWatchResponseParser(t *testing.T) {
 				Period: 60,
 				Alias:  "{{LoadBalancer}} Expanded",
 			}
-			series, partialData, err := parseGetMetricDataTimeSeries(resp, query)
+			series, partialData, err := parseGetMetricDataTimeSeries(mdrs, labels, query)
```
papagian, line 64: It would be nice to test also that `labels` are having the respected order; therefore I would suggest breaking `parseResponse` in two separate functions.
This would be a safeguard in case the code that building labels would change in the future and would break this behaviour but this only a suggestion.

papagian (approved):
LGTM, I have a small suggestion.

pkg/tsdb/cloudwatch/response_parser_test.go
aknuds1: That should be covered through the API tests I'm adding, and things will be changed around radically anyway when the conversion to data frames is over anyway.

Issue #25166, opened 2020-05-27
@grafana/e2e and e2e-selectors API redesign


stevenvachon:
## *Un*wrap `Cypress` and `cy` globals
We're using Cypress, so we should be explicit about it in our code instead of wrapping it to look like something else. We currently need to learn two _overlapping_ APIs which causes grind. If we were to switch to a different tool, everything—assertions, etc—would have to change with it, so there's no benefit.

### Before
```js
import { e2e } from '@grafana/e2e'
e2e().get('something')
```

### After
```js
cy.get('something')
```

## Use convenience functions in a flat module structure
This solves multiple problems:
* Time wasted when referencing _@grafana/e2e-selectors_' source code as it's not a 1:1 relation with _@grafana/e2e_'s API.
* Importing everything and having to follow a deeply nested structure.
* Component selectors within "pages", separate from "components".
* Magic/confusing (function with no arguments) syntax to use real CSS selectors instead of `aria-label`. Generally, we want to use `aria-label` whenever we can, but they're often not applicable to dependencies created with different methodologies.

### Before
```js
export const Components = {
  PanelEditor: {
    OptionsPane: {
      Group: {
        Visualization: {
          header: 'Options group Panel type'
        }
      }
    }
  }
}
```
```js
export const Pages = {
  Dashboard: {
    url: (uid: string) => `/d/${uid}`,
    someComponent: 'Some label'
  }
}
```
```js
import e2e from '@grafana/e2e'
e2e.pages.Dashboard.visit('abc123')
e2e.pages.Dashboard.someComponent()
e2e.components.PanelEditor.OptionsPanel.Group.Visualization.header()
e2e.flows.addPanel()
```

### After
```js
export default {
  dashboard: {
    someComponent: '[aria-label="Some label"]'
  },
  panelEditor: {
    optionsPane: {
      group: {
        visualization: {
          header: '[aria-label="Options group Panel type"]'
        }
      }
    }
  }
}
```
```js
export default {
  dashboard: (uid: string) => `/d/${uid}`
  // `someComponent` moved
}
```
```js
import { addPanel, getRoute, getSelector } from '@grafana/e2e'
cy.visit(getRoute('dashboard', 'abc123'))
cy.get(getSelector('dashboard.someComponent'))
cy.get(getSelector('panelEditor.optionsPanel.group.visualization.header'))
addPanel()
```

[...]

PR #26838, opened 2020-08-06
@grafana/e2e: improvements

stevenvachon:


packages/grafana-e2e/src/flows/addDataSource.ts
```diff
@@ -32,7 +50,39 @@ export const addDataSource = (config?: Partial<AddDataSourceConfig>): any => {
 
   e2e.pages.DataSource.name().clear();
   e2e.pages.DataSource.name().type(name);
+
+  if (basicAuth) {
+    e2e()
+      .contains('label', 'Basic auth')
```
torkelo, line 56: can we use aria-label for this like for other selectors

torkelo (approved):
the password check looks good, but the data source auth we usually use aria-label e2e constant selectors right?

packages/grafana-e2e/src/flows/addDataSource.ts
stevenvachon: Yes, there are many places that need such updates. I've been postponing it for a while now to do them "all" at once. Probably after the [API redesign](https://github.com/grafana/grafana/issues/25166) to benefit from compound selectors.

Issue #26828, opened 2020-08-06
Azure Monitor values 


migel0:
**What happened**:
 
After update from 6.x version to 7.x ( 7.0 , 7.1 and 7.2 tested ) some  value data from Azure Monitor Datasource are detected as  Unspecified. Specifically data from Telegraf agent.

The picture: Telegraf Agent sending metric to Azure Monitor  from VM and VMScaleset, then using Azure Monitor Datasource for  show the data.
Over Grafana prior versions, 6x this not happens.

![grafana_value2](https://user-images.githubusercontent.com/42206588/89525967-531c4400-d7e7-11ea-9165-445100955f92.png)
![grafana_value](https://user-images.githubusercontent.com/42206588/89525981-56173480-d7e7-11ea-8fd5-9536c9257b61.png)

Another inconvenient is that overwrite the value in  "yaxes" context, field context , or "overwrite" context  dont work.

- On the y axes section the values arent saved:

```json
  "yaxes": [
    {
      "label": null,
      "show": true,
      "logBase": 1,
      "min": null,
      "max": null,
      "format": "Unspecified",  // Cant save this value, dont work
      "$$hashKey": "object:829"
    },

```

-  Field context:
Overwrite the format value here dont work:
```json
    "fieldConfig": {
      "defaults": {
        "custom": {},
        "unit": "percent"
      }
```
- Overwrite section 
The instances on the scaleset are ephimeral so the matcher section work if i put  a value but dont work even with regex expression.

```json
      "overrides": [
                {
        "matcher": {
          "id": "byName",
          "options": "used" //   try regex here
        },
        "properties": [
          {
            "id": "unit",
            "value": "percent"
          }
        ]
      }]
```

**How to reproduce it (as minimally and precisely as possible)**:

-Send metrics with telegraf agent to Azure monitor
- Try  to visualize them in Grafana using Azure Monitor Datasource

**Anything else we need to know?**:

- Grafana version: 7.1
- OS Grafana is installed on: Ubuntu
- Grafana plugins: Default core instalation plugings

kylebrandt:
If I understand correctly:

 1. I will change the datasource so it doesn't set the unit if it is "unspecified" so overrides would not be need in that case particular case.
 2. For overrides to be useful, we really need multi-field matcher (a regex would likely suffice for now): https://github.com/grafana/grafana/issues/26831

migel0:
You have understood perfectly. thanks you.

[...]

Issue #26831, opened 2020-08-06
Field Overrides: Multi-Field Matching


kylebrandt:
**What would you like to be added**:
When using Overrides, The "Matcher > Filter by field" only lets you select a single field. We need at least a regex matcher to apply overrides to multiple fields. (Numeric / String / Time field matchers might be useful as well, but a regex would move us forward I think)

**Why is this needed**:
A response can have many series (e.g. United States, group by State). Additionally, the data may be very dynamic so the fields are not know ahead of time. 

Now that we are returning  data with Frames, users depend on overrides to get their display right in certain circumstances.

PR #26839, opened 2020-08-06
Azuremonitor: do not set unit if literal "Unspecified"

kylebrandt:
When the unit returned from azure is `Unspecified`, then we do not add that to frame.

part of #26828

### DIFF
--- a/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
+++ b/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
@@ -293,9 +293,11 @@ func (e *AzureMonitorDatasource) parseResponse(queryRes *tsdb.QueryResult, amr A
 		dataField := frame.Fields[1]
 		dataField.Name = amr.Value[0].Name.LocalizedValue
 		dataField.Labels = labels
-		dataField.SetConfig(&data.FieldConfig{
-			Unit: toGrafanaUnit(amr.Value[0].Unit),
-		})
+		if amr.Value[0].Unit != "Unspecified" {
+			dataField.SetConfig(&data.FieldConfig{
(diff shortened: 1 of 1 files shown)


ryantxu (approved):
Maybe better is to have "Unspecified" return an empty string from toGrafanaUnit()?

kylebrandt:
> Maybe better is to have "Unspecified" return an empty string from toGrafanaUnit()?

I considered that, but does add a unit with a value of an empty string. While that won't get sent to the front because of the json omitempty on the struct, I felt it was clearer to show that in that case we are not setting the unit.

pkg/tsdb/azuremonitor/azuremonitor-datasource.go
```diff
@@ -293,9 +293,11 @@ func (e *AzureMonitorDatasource) parseResponse(queryRes *tsdb.QueryResult, amr A
...
 		dataField.Name = amr.Value[0].Name.LocalizedValue
 		dataField.Labels = labels
-		dataField.SetConfig(&data.FieldConfig{
-			Unit: toGrafanaUnit(amr.Value[0].Unit),
-		})
+		if amr.Value[0].Unit != "Unspecified" {
+			dataField.SetConfig(&data.FieldConfig{
+				Unit: toGrafanaUnit(amr.Value[0].Unit),
+			})
+		}
```
papagian, line 300: in addition to this, what about adding a warning if unit is "unspecified"?

kylebrandt: Adding the unit is new, so I think of it more of a nice to have. So I would worry about making a thing that would be more likely to annoy the user than do anything useful. In particular since adding the unit has resulted in other issues that still need to be resolved ( https://github.com/grafana/grafana/issues/26831 )

PR #21835, opened 2020-01-30
POC: try uPlot for a fast simple graphing library

ryantxu:
This adds the uPlot library and a simple component that can show a dataframe.

This has strong potential to support a simple but super common set of desired graphs.  

This PR adds a toggle to the `graph2` (alpha React Graph) panel and will let you swith to the uPlot version.

ugly bits...  CSS.  I took their CSS and converted to .scss and then just changed the background color.  Things look OK on light theme, but terrible in the dark theme.

image

This should be able to support nano-date too ;)  with some work of couse

PR #26844, opened 2020-08-06
GraphPanel: Add Ordinal option for X-Axis Mode

### LINKED ISSUE #2418: Feature: Condense/exclude time on the x-axis for given time windows
We are currently investigating graphing stock market data using grafana, and as such it only has certain periods where we actually receive data to graph. it would be really good if we could define windows in which we expect to see data and then the graph just collapses the x-axis up so it just show 1 continuous line with a dashed line (or something) where the windows starts and stops

eg 
just show data between 9am - 3pm && 6pm - 9pm

cogentParadigm:

This X-Axis Mode will represent data sequentially on the X-axis and the X values will only be labelled as timestamps rather than being on a time scale. This allows you to do things like exclude weekends from the X-axis rather than having lines that span across them.

Fixes #2418

### DIFF
--- a/docs/sources/panels/visualizations/graph-panel.md
+++ b/docs/sources/panels/visualizations/graph-panel.md
@@ -125,6 +125,7 @@ Options are identical for both Y-axes.
     - **Buckets -** The number of buckets to group the values by. If left empty, then Grafana tries to calculate a suitable number of buckets.
     - **X-Min -** Filters out values from the histogram that are under this minimum limit.
     - **X-Max -** Filters out values that are greater than this maximum limit.
+  - **Ordinal -** The data is represented sequentially on the X-axis and the X values are only labelled as timestamps rather than being on a time scale.
 	
 ## Legend
 
--- a/docs/sources/panels/visualizations/graph-panel.md
+++ b/docs/sources/panels/visualizations/graph-panel.md
@@ -125,7 +125,7 @@ Options are identical for both Y-axes.
     - **Buckets -** The number of buckets to group the values by. If left empty, then Grafana tries to calculate a suitable number of buckets.
     - **X-Min -** Filters out values from the histogram that are under this minimum limit.
     - **X-Max -** Filters out values that are greater than this maximum limit.
-  - **Ordinal -** The data is represented sequentially on the X-axis and the X values are only labelled as timestamps rather than being on a time scale.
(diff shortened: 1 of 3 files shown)


docs/sources/panels/visualizations/graph-panel.md
```diff
@@ -125,6 +125,7 @@ Options are identical for both Y-axes.
     - **Buckets -** The number of buckets to group the values by. If left empty, then Grafana tries to calculate a suitable number of buckets.
     - **X-Min -** Filters out values from the histogram that are under this minimum limit.
     - **X-Max -** Filters out values that are greater than this maximum limit.
+  - **Ordinal -** The data is represented sequentially on the X-axis and the X values are only labelled as timestamps rather than being on a time scale.
```
oddlittlebird, line 128:
```suggestion
  - **Ordinal -** Data is represented sequentially on the X-axis, and the X values are only labelled as timestamps rather than being on a time scale.
```

torkelo:
Thanks for contributing. 

We are currently exploring new graph panels as the current one is built on our old angular panel architecture. So not sure we want to add new features like this to the graph panel. Will check with @ryantxu on how this could be supported in the coming graph panels

cogentParadigm:
Hi @torkelo. Thanks for the response. Two questions:

1. Does that mean you are declining this PR, or are you saying you are not sure if you can accept it yet?
2. Can you point me to any work/issues on new graph panels?

ryantxu:
Check https://github.com/grafana/grafana/pull/21835

I totally get the problem and we need a solution.  The float graph and structure is super difficult for us and I'm nervous about making to the visualization that can do everything

cogentParadigm:
I haven't tried out the uPlot branch, but It looks like uPlot has built-in support for this type of functionality based on this: https://github.com/leeoniya/uPlot/issues/261

I'm fine with looking at the uPlot panel and ensuring this type functionality is appropriately exposed there, but it does look like that has been in progress for some time and is still in draft stages. Are you expecting to release it relatively soon? and if not, are you opposed to adding this functionality in both places?

mckn:
@cogentParadigm great work on this! But I am sorry to say that we will not approve this PR. Starting next week we will start a project where we will migrate the Graph panel into its next version and we don't want to add more functionality to the existing one. Its responsibility is already way to big and we don't want to increase it even more.

Please feel free to help out with other issues, feature requests or in the Graph panel project!

PR #26845, opened 2020-08-06
Dashboard Links: Keep links in sync with dashboard time range

### LINKED ISSUE #26694: Time range not updated in dashboard link
**What happened**:
Time range parameter doesn't get updated in dashboard links

Time ranged generated with the dashboard link to be updated right after changing the time range of the dashboard.
The time range passed to the dashboard link is only updated after saving the dashboard (with the time range checkbox), closing the dashboard and reopen it.

Create dashboard link from dashboard settings, checking 'time range' switch button, go to dashboard, change time range, open the link and see that time range is didn't update and use previous time range instead.

Notice the time range in the generated link on the bottom left corner
Kapture 2020-07-30 at 13 55 04

**Environment**:
- Grafana version: 7.1.1

kaydelaney:
Updates dashboard links that make use of the current time range when that time range is updated.

Closes #26694

### DIFF
--- a/public/app/features/dashboard/components/SubMenu/DashboardLinks.tsx
+++ b/public/app/features/dashboard/components/SubMenu/DashboardLinks.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, useReducer } from 'react';
 import { Icon, IconName, Tooltip } from '@grafana/ui';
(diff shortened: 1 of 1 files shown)


torkelo (approved):
Looks ok, if dashboard time range was in a recoil js state, this would have been easy :) 
https://recoiljs.org/

PR #26847, opened 2020-08-07
TextPanel: Fixes issue when interpolation of variables stops working

### LINKED ISSUE #26083: global built-in variables: __dashboard and __dashboard.name are not available in the text panel
I want to use the relatively new global variable __dashboard to open git hub issues.
When I try to use the variable in the panel it is not being replaced, I've tried both the `$` and the `[[]]` substitution none worked

How I just reproduce it:

* Edit a text panel and add ${__dashboard.name} in the text - Should work
* Exit the edit panel
* Edit the same panel again, remove and add the ${__dashboard.name} now it does not work

hugohaggmark:
I couldn't really repro the issue as described in issue #26083. But I could get the interpolation of variables to stop working as we were transforming the content provided by the user to markdown before interpolating variables. This PR reverses the order, we interpolate variables before transforming the content to markdown.

I also updated the `marked` library and did some small refactoring, removing the remains of the `text` mode for instance.

Fixes #26083 (couldn't repro the exact issue)


torkelo (approved):
Looks good!

hugohaggmark:
@torkelo I needed to change the TextPanel from markdown to html because I guessed that we want to show the `raw` interpolated value and not a markdownified value. Let me know if this makes sense 🙏

torkelo:
>  I needed to change the TextPanel from markdown to html because I guessed that we want to show the raw interpolated value and not a markdownified value. Let me know if this makes sense pray 

Not sure I understand. You mean in the test dashboard / e2e test?

devenv/dev-dashboards/feature-templating/global-variables-and-interpolation.json
```diff
@@ -34,8 +34,8 @@
...
+        "mode": "html"
```
torkelo, line 38: Any reason for this change?  is not the above rendering to this (but with interpolated variables) anyway?

hugohaggmark: Sorry for being unclear. Let's see the upgraded markdown will transform an interpolated string like this `(A'A"A|BB\\B|CCC)` into this `(A'A"A|BB\B|CCC)`. 

Much like GitHub markdown renderer does with the same string below
(A'A"A|BB\\B|CCC) is in fact `(A'A"A|BB\\B|CCC)`
* `Server:regex` = (A'A"A|BB\\B|CCC)
* `Server:lucene` = ("A'A\"A" OR "BB\\B" OR "CCC")
* `Server:glob` = {A'A"A,BB\B,CCC}

So I thought that we wanted the e2e test to verify the correct string, the raw output from interpolation and not the markdownified version. Does this make sense? Hard to explain.

Just realized I should be able to surround the variables with ` and change it back to markdown.

hugohaggmark: Reverted back to markdown but I think there was a bug in the e2e test for the `:html` format so I had to change the test.

torkelo: ah, now I understand. now that interpolation happens before markdown markdown is impacting the interpolated value. 
In that case I agree html mode  is probably better :)

hugohaggmark: Hehehe I've already reverted back to markdown surrounding all values with `. Please see my latest commit :)

torkelo: ah, smart!

Issue #24588, opened 2020-05-12
Support for multiple organizations


torkelo:
Organizations (orgs) were originally created to support a multi-tenancy hosting service. Meaning, to use a single Grafana server to provide a fully isolated Grafana experience to multiple companies. 

But mainly used orgs to help split a large installation up and create some separation between users & teams within a single company. In Grafana v5 we added teams & dashboard permissions to help manage a large single organization within Grafana so that use case is no longer recommended. 

Multi-org support is currently a big usability problem, especially around user management. Removing multi-org support would make Grafana a lot easier to manage and setup for single org setups. But we realize this could be very disruptive and want to gather all feedback on use cases first. 

Whatever we decide, we plan to have some migration story ready when we remove multi-org support.

uwereiter:
We use the multi-org feature to host multiple independent customers and let them access their dashboards separately. So multi-org feature is mandatory for us.

kdandersen:
Saw the deprecation warning notice.
And would like to add that we also use the Org. feature to create tenancy separations in a shared Grafana set-up. Thus we would be **very sorry** to see this feature dissapear.

ajaxous:
Hi,

I'm also using the multi-org feature to give customers views to just their data -- the data source is Zabbix and there are specific users created for use with each orgs' Grafana Zabbix data source.  So hosts are partitioned by customer in Zabbix.  Special users are created for each customer (these are internal sources, not accessible by the customer) in Zabbix.  These special users are used for each associate org's Zabbix data source.  So in each org, there is no possible way an org's members could see another customers hosts and data even if given admin previleges.

This is an awkward way to do things, but has proven very effective.  I could not have achieved the same results with the Teams feature.  I do use that feature for people within an org to par down what certain people have access to.

If the org concept went away without another equivelant method of managing distinct and seperate data sources, each with their own user lists, I think it would be a great loss and likely prevent Grafana from being used as a solution in my particular case.

I do agree that user management could be improved and moving dashboards from org to org could also be optimized.

Thank you for the heads up about the potential change and for giving serious consideration to other multi-tenancy options that might offer similar capabilities to what exists today.

bergquist:
> We use the multi-org feature to host multiple independent customers and let them access their dashboards separately. So multi-org feature is mandatory for us.

Would it be possible to run one instance per customer? Grafana is a fairly lightweight process.

Codelica:
[...]

PR #26853, opened 2020-08-07
Orgs: Remove org deprecation notice as we have decided to preserve multi-org support 

torkelo:
After an overwhelmingly negative reaction to our proposal to remove orgs and streamline user management (#24588) we have decided to keep support for multiple orgs.


hugohaggmark (approved):
LGTM!

PR #26854, opened 2020-08-07
Drone/CircleCI: Upgrade build pipeline tool

### LINKED ISSUE #23567: Convert our musl builds to static linking
I think we should convert our musl builds to statically link musl, for robustness reasons. The size increase is probably minimal.

**Why is this needed**:
After solving #23500, that our master Docker images were broken on ARM v7, I realized that the issue arose from the fact that Go builds targeting musl v1.2.x are incompatible with musl v1.1.x (f.ex. Alpine 3.10/3.11). The incompatibility is because v1.2.x switches to 64-bit time types, and the Go binaries have references to these types, which are not present in musl v1.1.x, and a dynamic linking error happens as a result.

Another thing worth mentioning, is that we are now stuck at musl v1.1.x, in order to not break compatibility with Alpine.

aknuds1:
Upgrade to new version of the build pipeline tool, with static musl libc linking. Should make our musl (Alpine Linux compatible) binaries more portable, since they become independent of the system's version of musl (1.2 breaks compatibility with 1.1 on 32-bit archs).
 
Fixes #23567.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ commands:
       - run:
           name: "Install Grafana build pipeline tool"
           command: |
-            VERSION=0.4.25
+            VERSION=0.5.0
             curl -fLO https://grafana-downloads.storage.googleapis.com/grafana-build-pipeline/v${VERSION}/grabpl
             chmod +x grabpl
             mv grabpl /tmp
--- a/.drone.yml
+++ b/.drone.yml
@@ -27,7 +27,7 @@ steps:
   - cp -r $(yarn cache dir) yarn-cache
   environment:
     DOCKERIZE_VERSION: 0.6.1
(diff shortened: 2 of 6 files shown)


wbrowne (commented):
Not sure if relevant but there's some [other references](https://github.com/grafana/grafana/blob/master/packages/grafana-toolkit/docker/grafana-plugin-ci-alpine/scripts/deploy.sh#L46) to `grabpl` installs in the grafana-toolkit. Should they be bumped in this PR too?

aknuds1:
@wbrowne I wasn't really aware of that. I guess I might as well bump it?

aknuds1:
@wbrowne Thanks for finding additional references, bumped them all.

wbrowne (approved):
🕺

PR #26858, opened 2020-08-07
SQL: Run tests as integration tests

sakjur:
Ensures we're running the SQL store test suite as integration tests since we're excluding non-integration tagged tests from running.

### DIFF
--- a/pkg/services/sqlstore/alert_notification_test.go
+++ b/pkg/services/sqlstore/alert_notification_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 package sqlstore
 
 import (
--- a/pkg/services/sqlstore/alert_test.go
+++ b/pkg/services/sqlstore/alert_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 package sqlstore
 
 import (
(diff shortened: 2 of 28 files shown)


aknuds1 (approved):
LGTM assuming they should all be integration tests.

papagian (approved):

aknuds1:
The integration tests are insanely slow.

torkelo:
> Ensures we're running the SQL store test suite as integration tests since we're excluding non-integration tagged tests from running. 

What do you mean we are excluding non-integration tests?

sakjur:
@torkelo With the build pipeline we changed the behavior around running integration tests such that tests which have already been run as unit tests aren't also being run as integration tests. But we also hadn't marked the SQL tests as integration tests, so the MySQL/Postgres test runners aren't running all of those atm.

But we also have a huge performance problem when enabling them, so we can't solve it by simply flicking a switch.

sakjur:
(cc @bergquist, who had some thoughts about 👆 when we talked a while back)

bergquist:
My suggestion is to improve the `CleanDb` function for each database. 
Rather than dropping and recreating the database, I think we should just delete the content of each table and test the DB migration separately. I _think_ that could speed up the integration tests signcitly. 

Ideally, all tests should delete the content they created in the teardown, but for now I think this idea might be worth exploring.

papagian:
Closing this in favour of #28265

PR #26866, opened 2020-08-07
Templating: Fixes renaming a variable using special characters or same name

### LINKED ISSUE #26562: Can't remove invalid character when creating variable
**What happened**:
I was creating a variable and accidentally typed an invalid character (@) in the name. An error indicated a validation error since only word and digit characters are allowed in a name. I tried to delete the incorrect character using backspace but the cursor stopped at that character and would not delete the invalid character. I had to leave the Add Variable screen to clear the field.

I subsequently found that if you type characters before the invalid character you can move the cursor to the beginning of the input field and use delete to delete all characters, including the invalid one.
grafana variable name bug

Entering an invalid character should display a validation error then allow the user to backspace to delete that invalid character.

1) From a dashboard, click dashboard settings, Variables, Add Variable
2) In the Name field type "@". Validation error displays (expected)
3) Press backspace. Nothing happens.

Screencap GIF demonstrates this. This can be reproduced at play.grafana.org 

**Environment**:
- Grafana version: 7.1.0  (8101355285)
- Data source type & version: N/A - This occurs before specifying a source
- OS Grafana is installed on: Docker container from official Docker Hub. Host OS for Docker is AWS Linux
- User OS & Browser: Windows 10 20H2 (Build 19042.388) - Beta, Browser Chrome 84.0.4147.89, also reproducible on current versions of FIrefox and Edge Chromium
- Grafana plugins: N/A
- Others:

lucasra1:
You couldn't delete an invalid character after typing it into the name-input field.
While investigating, the issue turned out to be bigger, as there was a problem with
valid characters too. (See test scenarios below)

The fix seems to be, to remove an unnecessary check in the `changeVariableName`
action. There is theoretically now the possibility, that the `changeVariableName`
action is called with the same name, as the variable is already, but practically
there seems no possibility, that this could happen. A test, which checks that, had
to be removed too.

Test scenarios:
* 1st Scenario
    1. Type "@"
    2. Try deleting it
* 2nd Scenario
    1. Type "w"
    2. delete "w"
    3. Try typing "w" again

Fixes #26562 

@hugohaggmark I removed a check and a test, that you created. This check seemed unnecessary to me, so I removed it to fix the bug. Please correct me if I am wrong. Maybe you had your reasons to implement it that way.

### DIFF
--- a/public/app/features/variables/editor/actions.ts
+++ b/public/app/features/variables/editor/actions.ts
@@ -95,6 +95,10 @@ export const completeChangeVariableName = (identifier: VariableIdentifier, newNa
   getState
 ) => {
   const originalVariable = getVariable(identifier.id, getState());
+  if (originalVariable.name === newName) {
+    dispatch(changeVariableNameSucceeded(toVariablePayload(identifier, { newName })));
+    return;
+  }
   const model = { ...cloneDeep(originalVariable), name: newName, id: newName };
   const global = originalVariable.global;
   const index = originalVariable.index;
--- a/public/app/features/variables/editor/actions.ts
+++ b/public/app/features/variables/editor/actions.ts
@@ -57,11 +57,6 @@ export const onEditorAdd = (identifier: VariableIdentifier): ThunkResult<void> =
 
 export const changeVariableName = (identifier: VariableIdentifier, newName: string): ThunkResult<void> => {
   return (dispatch, getState) => {
-    const variableInState = getVariable(identifier.id, getState());
-    if (newName === variableInState.name) {
-      return;
-    }
-
     let errorText = null;
     if (!newName.match(/^(?!__).*$/)) {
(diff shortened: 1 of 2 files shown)


hugohaggmark (changes requested):
Nice catch @LucasRaschek and I don't remember why I added the check but now when I try the second scenario as described in the issue the `w` variable is removed.
1. Type "w" (OK)
2. delete "w", I see the `Only word and digit characters are allowed in variable names` message (OK)
3. Try typing "w" again, the renamed variable is removed. (NOT OK)

LucasRaschek:
@hugohaggmark maybe this was the reason, why you implemented that check :laughing: To fix this issue I moved your check now. I modified and added your test again too. Should work now, please check

hugohaggmark (approved):
Thank you for contributing @LucasRaschek this looks and works great now 🎉

PR #26868, opened 2020-08-07
Don't require HTTPS for HSTS headers to be emmitted

### LINKED ISSUE #26770: Enable HSTS when serving HTTP only (SSL is offloaded)

Per the [documentation about the `strict_transport_security` settings](https://grafana.com/docs/grafana/latest/administration/configuration/#strict_transport_security), they are only in effect when/if HTTPS is set to true. Would it be possible to have those options not depend on other options? After all, they are just HTTP headers.

**Why is this needed**:

Using grafana's standard docker images, we have a setup in which the container only serves through HTTP and SSL is offloaded to a transparent reverse proxy in front of it. This makes it a lot easier to handle certificates used to serve grafana, but it is a downside that we can't enable this very simple security setting that should be completely independent.

dnmvisser:
Fixes https://github.com/grafana/grafana/issues/26770

### DIFF
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -475,7 +475,7 @@ mitigate the risk of [Clickjacking](https://www.owasp.org/index.php/Clickjacking
 
 ### strict_transport_security
 
-Set to `true` if you want to enable HTTP `Strict-Transport-Security` (HSTS) response header. This is only sent when HTTPS is enabled in this configuration. HSTS tells browsers that the site should only be accessed using HTTPS.
+Set to `true` if you want to enable HTTP `Strict-Transport-Security` (HSTS) response header. Only use this when HTTPS is enabled in your configuration, or when there is another upstream system that ensures your application does HTTPS (like a frontend load balancer). HSTS tells browsers that the site should only be accessed using HTTPS.
 
 ### strict_transport_security_max_age_seconds
 
--- a/pkg/middleware/middleware.go
+++ b/pkg/middleware/middleware.go
@@ -300,7 +300,7 @@ func AddDefaultResponseHeaders() macaron.Handler {
 
 // AddSecurityHeaders adds various HTTP(S) response headers that enable various security protections behaviors in the client's browser.
 func AddSecurityHeaders(w macaron.ResponseWriter) {
-	if (setting.Protocol == setting.HTTPS || setting.Protocol == setting.HTTP2) && setting.StrictTransportSecurity {
+	if setting.Protocol == setting.HTTP2 && setting.StrictTransportSecurity {
 		strictHeaderValues := []string{fmt.Sprintf("max-age=%v", setting.StrictTransportSecurityMaxAge)}
 		if setting.StrictTransportSecurityPreload {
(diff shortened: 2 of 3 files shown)


docs/sources/administration/configuration.md
```diff
@@ -475,7 +475,7 @@ mitigate the risk of [Clickjacking](https://www.owasp.org/index.php/Clickjacking
 
 ### strict_transport_security
 
-Set to `true` if you want to enable HTTP `Strict-Transport-Security` (HSTS) response header. This is only sent when HTTPS is enabled in this configuration. HSTS tells browsers that the site should only be accessed using HTTPS.
+Set to `true` if you want to enable HTTP `Strict-Transport-Security` (HSTS) response header. Only use this when HTTPS is enabled in your configuration, or when there is another upstream system that ensures your application does HTTPS (like a frontend load balancer). HSTS tells browsers that the site should only be accessed using HTTPS.
```
sakjur, line 478: I think it's more effective to just explain the effect of HSTS, to keep the description brief.

@oddlittlebird WDYT?

```suggestion
Set to `true` if you want to enable HTTP `Strict-Transport-Security` (HSTS) response header. HSTS tells browsers that the site should only be accessed using HTTPS.
```

pkg/middleware/middleware.go
```diff
@@ -300,7 +300,7 @@ func AddDefaultResponseHeaders() macaron.Handler {
 
 // AddSecurityHeaders adds various HTTP(S) response headers that enable various security protections behaviors in the client's browser.
 func AddSecurityHeaders(w macaron.ResponseWriter) {
-	if (setting.Protocol == setting.HTTPS || setting.Protocol == setting.HTTP2) && setting.StrictTransportSecurity {
+	if setting.Protocol == setting.HTTP2 && setting.StrictTransportSecurity {
```
sakjur, line 303: Hm, this doesn't look right. We want to go from:

|                   | HTTP | HTTPS | HTTP2 |
| ---------- | ------ | ------- | ------- |
| HSTS ON  | N         | Y          | Y           |
| HSTS OFF | N        |  N          | N          |

to

|                   | HTTP | HTTPS | HTTP2 |
| ---------- | ------ | ------- | ------- |
| HSTS ON  | Y        | Y          | Y           |
| HSTS OFF | N        |  N          | N          |

But this looks like:

|                   | HTTP | HTTPS | HTTP2 |
| ---------- | ------ | ------- | ------- |
| HSTS ON  | N         | N          | Y           |
| HSTS OFF | N        |  N          | N          |

Am I missing something?

sakjur: In case I'm getting this right, the fix would be

```suggestion
	if setting.StrictTransportSecurity {
```

aknuds1 (commented):
Can this PR also include tests for the HSTS behaviour?

docs/sources/administration/configuration.md
aknuds1: I like @sakjur's edit, brief and to the point.

oddlittlebird: Agreed. However, it seems like the "when to use" information might be important to include.

Issue #26523, opened 2020-07-22
Graph component improvement


vankop:
**Why is this component needed**:

Will be great if Graph component will allow to render points with info from data source. This is especially useful for tracing/logging and integrating this in one all-in-one-observability tool. Since logs and traces can not be presented as a timeline (line/area) they can be displayed as a separate points on board
___
 - [x] Is/could it be used in more than one place in Grafana?

**Where is/could it be used?**:
<img width="1343" alt="Снимок экрана 2020-07-22 в 16 04 21" src="https://user-images.githubusercontent.com/10380560/88180105-87173700-cc35-11ea-9a66-d8d906523cc7.png">

___
- [x] Post screenshots possible.
- [ ] It has a single use case.
- [x] It is/could be used in multiple places.

**Implementation** (Checklist meant for the person implementing the component)

- [ ] Component has a story in Storybook.
- [ ] Props and naming follows [our style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/frontend.md).
- [ ] It is extendable (rest props are spread, styles with className work, and so on).
- [ ] Uses [theme for spacing, colors, and so on](https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md).
- [ ] Works with both light and dark theme.

**Documentation**

- [ ] Properties are documented.
- [ ] Use cases are described.
- [ ] Code examples for the different use cases.
- [ ] Dos and don'ts.
- [ ] Styling guidelines, specific color usage (if applicable).

torkelo:
We will add something like this soon with the coming exemplars and improved annotation support, to follow progress on new graph component  https://github.com/grafana/grafana/issues/27848

Issue #25312, opened 2020-06-03
Bump Storybook to 6.x


dprokop:
[Bumping Storybook](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-53x-to-60x) gives us these improvements:
- Zero config typescript support
- Possibly build performance
- [Args](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#args-passed-as-first-argument-to-story) gives us actions and knobs for free

Issues to solve:

- [x] Get branch building (currently the tests are failing)
- [x] Improve performance in dev using ts-fork-checker-webpack-plugin and transpileOnly flag
- [x] [Correct globs in main.js](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#correct-globs-in-mainjs)
- [x] [Usage of deprecated addParameters and addDecorator in preview.js](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addparameters-and-adddecorator)
- [x] [use of addon.setConfig](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-immutable-options-parameters)
- [x] [Check story sorting is correct](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-pass-denormalized-parameters-to-storysort)
- [x] ~~[Is storybook centered decorator a replacement for our own?](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#removed-addon-centered)~~ The layout parameter centers a story vertically _and_ horizontally.
- [x] ~~Check theming is functioning correctly and [using latest config options]~~ https://github.com/storybookjs/storybook/pull/11352(https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/theming.md#storybook-theming)
- [x] Migrate a story to use addon-controls
- [x] Update grafana-ui readme with story migration path for making the most of v6 addons and contributions

PR #26873, opened 2020-08-07
Improve storybook build performance

vankop:

I wanted to improve Graph component in our fork regarding to https://github.com/grafana/grafana/issues/26523 and found out that developing using current storybook setup veeeery slow.. So I decided to improve build a little bit:

- upgrade loaders versions (`expose-loader` improve perf a lot)
- use react-docgen-typescript-loader only for production

no issue

Fixes #

|       | cold start | incremental build |
|----|-----------|------------------|
| `master` https://github.com/grafana/grafana/commit/398bc045fc0094d9a1482f040d7ef334c6a8034b | 1.73 min for manager and 3.1 min for preview | 5-12s depending on changes |
|  PR           | 56s , 1.38min | 5-11.5s depending on changes |
| PR + transpileOnly | 25s, 45s | <4.5s |

This PR improves mostly only cold start. To improve incremental build I strongly recommend to move out of `ts-loader` and use babel instead or solve `ts-loader` issues. Also shipping es6+ in dev allow to use `options.devtool = false`

Profile of HMR update below
Снимок экрана 2020-08-08 в 1 00 15


packages/grafana-ui/src/components/Select/utils.ts
```diff
@@ -20,19 +20,29 @@ export const cleanValue = (value: any, options: SelectOptions): SelectableValue[
...
 export const findSelectedValue = (value: string | number, options: SelectOptions): SelectableValue | null => {
-  for (const option of options) {
-    if ('options' in option) {
-      let found = findSelectedValue(value, option.options);
-      if (found) {
-        return found;
-      }
-    } else if ('value' in option && option.value === value) {
-      return option;
+  if (Array.isArray(options)) {
```
vankop, line 40: found out this when was playing increasing `tsconfig.json` target to es2017

public/app/plugins/panel/gettingstarted/types.ts
```diff
@@ -24,6 +24,6 @@ export interface SetupStep {
   subheading: string;
   title: string;
   info: string;
-  cards: (Card | TutorialCardType)[];
+  cards: Array<Card | TutorialCardType>;
```
vankop, line 27: yarn lint

aknuds1:
The front-end tests [fail](https://drone.grafana.net/grafana/grafana/1070/1/12).

vankop:
> The front-end tests fail.

hm.. whatever, I have reverted utils changes

aknuds1:
The StoryBook build looks really fast with your changes, but I'll leave it to the front-end team to look them over.

torkelo:
Awesome work!  Big thanks from the frontend platform squad! :heart: we will look at this asap in the coming 2-3 days

packages/grafana-ui/.storybook/webpack.config.js
```diff
@@ -7,30 +7,39 @@ module.exports = ({ config, mode }) => {
...
-          loader: require.resolve('react-docgen-typescript-loader'),
-          options: {
-            tsconfigPath: path.resolve(__dirname, 'tsconfig.json'),
-            // https://github.com/styleguidist/react-docgen-typescript#parseroptions
-            // @ts-ignore
-            propFilter: prop => {
-              if (prop.parent) {
-                return !prop.parent.fileName.includes('node_modules/@types/react/');
-              }
+      use: isProductionBuild
```
dprokop, line 10: Good call, but this will result in the props table not being rendered in the dev mode's preview. I suggest maybe use env variables instead of the production assumption only (https://storybook.js.org/docs/configurations/env-vars/) and introduce another storybook script in the grafana-ui package.json that would allow opting in to render the docs tables too? like `yarn storybook:docs`?

packages/grafana-ui/.storybook/webpack.config.js
```diff
@@ -65,16 +74,10 @@ module.exports = ({ config, mode }) => {
...
-      },
-      {
-        loader: 'expose-loader',
-        query: '$',
-      },
-    ],
+    loader: 'expose-loader',
+    options: {
+      exposes: ['jQuery', '$'],
+    },
```
dprokop, line 80: 👍

packages/grafana-ui/.storybook/webpack.config.js
```diff
@@ -89,7 +92,6 @@ module.exports = ({ config, mode }) => {
           test: /[\\/]node_modules[\\/].*[jt]sx?$/,
           chunks: 'initial',
           priority: -10,
-          reuseExistingChunk: true,
```
dprokop, line 92: Why is this removed?

packages/grafana-ui/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "baseUrl": "node_modules/@types",
-    "declarationDir": "dist",
```
dprokop, line 4: This will make the dist package not include declaration files, so it can not be removed unless we programmatically move all *.d.ts files when the package is bundled.

dprokop (changes requested):
Thanks for the contribution @vankop. It's indeed faster, but it also introduces a couple of regressions that I've mentioned in my comments.

packages/grafana-ui/.storybook/webpack.config.js
vankop: yes good idea.

aknuds1:
:crossed_fingers: this work can be merged, since StoryBook is really slowing down CI at the moment :)

vankop:
I am busy right now, feel free to pick it up + will be great to see incremental build improvments

aknuds1:
Unfortunately I'm not involved in the front-end development, so I don't know how to tweak the build :/

zoltanbedi:
Thanks for contributing @vankop. Let me share my 2c here. I think we should upgrade storybook to v6 first and then revisit the build performance. I know that the main problem is the props table generator because you disabled it for development it means that the ci is still going to be slow. All in all I like the idea to make this faster but I think we should upgrade to v6 first.

aknuds1:
What @zoltanbedi says sounds sensible.

mckn:
A quick update: @peterholmberg is currently looking into upgrading storybook 6 and we will see what bits and pieces from this PR we can apply after that is done.

I love this initiative and great that we could speed up things some more.

mckn:
As I mentioned earlier we are looking into upgrading Storybook to version 6 so we will close this PR. We would love to get some help of speeding up the upgraded version of Storybook once that is in place.

Issue: https://github.com/grafana/grafana/issues/25312

Thanks again for opening this PR 🙌

PR #19723, opened 2019-10-09
Graph-Panel: Center option for bar charts

clamber:
This PR adds a 'center'-option for bar charts.
We already used different visualization tools and we really missed this option in grafana.
So we decided to add it by ourselves and contribute.

The 'center'-option looks very nice if you have daily aggregated data, so the bar is shown right above the date.

Here are 2 Images to visualize the impact:

Bar-Chart - not centered:
Graph-Panel

Bar-Chart - centered:
Graph-Panel-Center

PR #26876, opened 2020-08-09
Graph: Revert bar centering to fix tooltip issue

### LINKED ISSUE #22611: Tooltip bug: value transition inside bar
**What happened**:
When using bars, tooltip displays value of previous data point when hovering over left half of a bar and displays correct value, when hovering over right half of bar.

Tooltip shows value of current data point regardless where I hover over the bar. Transitions of value ("previous <-> current" and "current <-> next") should happen between bars.

- Create new Graph panel (we can use the default random generated data)
- Change draw mode to bar only
- Zoom in (several times) until you can comfortable move your cursor left and right inside a bar
- Move cursor slowly from left border of the bar to the right. Ovserve the tooltip value change when crossing the mid of the bar

This also happens, when plotting wider bars (e.g. group by 1 day) so I don't think it has anything to to with the zoom level of above example...

**Environment**:
- Grafana version: Grafana v6.6.2 (3fa63cfc34)
- Data source type & version: Default random data (also tried on MS-SQL data)
- OS Grafana is installed on: Official grafana container on ubuntu (inside VM)
- User OS & Browser: Windows 10 & Firefox 72.0.2 (64-bit)
- Grafana plugins: Graph

- Others:

image

image

torkelo:
Reverts https://github.com/grafana/grafana/pull/19723 as it's causing to many issues with the All series tooltip (Default tooltip mode). Mostly because this tooltip mode is really crap at detecting data points your hovering over. The code is messy and soon to be replaced so opting to revert this instead.

Fixes #22611  (And many duplicates as this has been reported many times)

### DIFF
--- a/public/app/plugins/panel/graph/graph.ts
+++ b/public/app/plugins/panel/graph/graph.ts
@@ -516,7 +516,6 @@ class GraphElement {
       }
       default: {
         options.series.bars.barWidth = this.getMinTimeStepOfSeries(this.data) / 1.5;
-        options.series.bars.align = 'center';
         this.addTimeAxis(options);
         break;
       }


hugohaggmark (approved):
LGTM!

shija03:
@torkelo  - basically when looking at bar charts, users will just need to be initially confused due to the fact that the value changes on first half of the bar to next half.  You mentioned that there is a new release coming out that has all this fixed?

ryantxu:
@shija03 -- this was merged a week ago, you can test with the latest nightly build:
https://grafana.com/grafana/download
(click the nighly build link)

jrgilman:
@ryantxu it doesn't seem the issue has been corrected? https://github.com/grafana/grafana/issues/22611#issuecomment-677886265

Peek 2020-08-20 15-26

PR #26879, opened 2020-08-09
Table: Support date unit formats on string values 

### LINKED ISSUE #25884: Grafana 7.x cannot change the format date by setting the unit in
**What happened**:
image
image
I try to change the format like the captures, but it doesn't work.
work as the described in the release note.

**Environment**:
- Grafana version: 7.0.4
- Data source type & version: Postgresql 10
- User OS & Browser: Chrome
- Grafana plugins: table
- Others:

torkelo:
Fixes #25884 

So many ways to address this. 

I think the display processor should be smarter here and identify that a date unit has been selected and try to parse strings to dates.

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.ts
+++ b/packages/grafana-data/src/field/displayProcessor.ts
@@ -7,7 +7,7 @@ import { GrafanaTheme } from '../types/theme';
 import { DisplayProcessor, DisplayValue, DecimalCount, DecimalInfo } from '../types/displayValue';
 import { getValueFormat } from '../valueFormats/valueFormats';
 import { getMappedValue } from '../utils/valueMappings';
-import { DEFAULT_DATE_TIME_FORMAT } from '../datetime';
+import { dateTime } from '../datetime';
 import { KeyValue, TimeZone } from '../types';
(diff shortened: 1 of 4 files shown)


dprokop (approved):
looks good

packages/grafana-data/src/field/displayProcessor.ts
```diff
@@ -30,25 +30,27 @@ export function getDisplayProcessor(options?: DisplayProcessorOptions): DisplayP
...
+  let unit = config.unit;
+  let hasDateUnit = unit && (timeFormats[unit] || unit.startsWith('time:'));
 
-  if (field.type === FieldType.time) {
-    if (config.unit && timeFormats[config.unit]) {
-      // Currently selected unit is valid for time fields
-    } else if (config.unit && config.unit.startsWith('time:')) {
-      // Also OK
-    } else {
-      config.unit = `time:${DEFAULT_DATE_TIME_FORMAT}`;
```
torkelo, line 42: this is the modify the config object bug

ryantxu (approved):
LGTM -- good that we do not mutate anymore

PR #26884, opened 2020-08-09
Transformations: Added new transform for excluding and including rows based on their values.

### LINKED ISSUE #25416: Transformation: Filter out zero or null values
Hiding rows with values that we do not need in table. For instance rows with 0 values.

**Why is this needed**:
OpenTSDB, query to exclude 0 values is not possible

Example:
We have a sorted table in ascending order that has many 0 values. We would like to hide these rows.

Possible solution 1: Add option under Value Mappings to hide row if value is 0. This solution would make it possible to hide any value.
Possible solution 2: Use Transformations -> if All Zeros equals true then hide row

Totalus:

Add a transformation for filtering data by value after the query. There is currently no easy way to filter out some data out of a query result (that I know of). Not all queries/data source allow for proper data filtering options, thus a way to filter out data directly in Grafana is useful when we don't have enough control over the query results.

I did not create a specific issue for this PR, but it will fix a few existing issues.
Fixes #25416, probably part of #25357.

~~This is currently a draft. I'd like to get comments on the way I'm planning to implement it.~~
[EDIT] This is pretty close to final. Take a loot at my other comment.

It currently looks like this:
Snapshot

The idea is that we can include/exclude some data (rows) that match the filters in the transform. The way it processes is it runs all the listed filters on each row, in order (top to bottom). If the row matches the filter, it flags the corresponding row for include/exclude (depending on the filter type). The default flag for each row is the opposite of the first filter type. At the end of the transform, it returns the rows that are flagged to be included. Simple yet very flexible. The transform can be chained more than once to achieve the desired filtering.

It currently only supports Regex test for filtering the data, but I plan on adding multiple test types, namely:
- Greater/GreaterOrEqual
- Lower/LowerOrEqual
- NotNull/Null
- Equal (though Regex)
- Any other ideas ?


Totalus:
Ok, I build the thing. Hope its good enough. Here's a preview.

**Example**
filter2

**Multiple type of tests implemented**
snapshot

**Highlight invalid configuration**
snapshot

Tests and documentation are also there.

There is one thing I would improve. As it is right now, it is possible to apply a few filter types on a `time` field, but the filtering will be confusing for the user. The time is displayed in an text format (ISO or other), but the filtering is applied on the numeric (epoch) value.

I see two possibilities:

- The first is we remove the filtering on `time` fields. There might not be much use cases for filtering on time since we specify the time range in the dashboard anyway, so its more than often part of the query and results are returned accordingly.

- The second possibility is to convert the value to ISO format before filtering (for filter like Regex, or equal) or in other cases (for filter like greater), convert a filter expression to epoch time. This way the user could say `Include` `Time` `Greater Than` `2020-08-22`, and they would get the desired result. To do that however I need to put the finger on the functions that are used to format the time values for display. If someone can point them out, I would investigate this second possibility.


Also, there are a few remaining errors.
- Two of them are related to `==` and `!=` that are intentionally used. I don't know how to ignore those errors.
- The third I'm not sure to understand why there is an error. I used similar code elsewhere and there were no complaints. There must be something I don't see.

oddlittlebird:
For the UI, please use sentence case for field names and titles.

Per the [Documentation style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-style-guide.md), all UI text should be bold in the docs.

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
 | 2020-07-07 09:30:57 | Humidity    | 33    |
 | 2020-07-07 09:30:05 | Temperature | 19    |
 
+
+## Filter By Value
+
+This transformation allows you to filter your data directly in Grafana. This allows you to remove some of the data points from your query result. You can either include or exclude some of your data points (rows) from a field that match the filter test (`Equal`, `Greater`, `Lower`, `isNull`, `Regex`, `Equal` and a few more).
```
oddlittlebird, line 265:
```suggestion
This transformation allows you to filter your data in Grafana and remove some data points from your query result. You can either include or exclude data points (rows) from a field that match the filter test: (`Equal`, `Greater`, `Lower`, `isNull`, `Regex`, `Equal` and a few more).
```

oddlittlebird: List the filters in a table or an unordered (bulleted) list. Define each one.

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
...
+|---------------------|--------------|
+| 2020-07-07 11:34:23 | 32    |
+| 2020-07-07 11:34:22 | 28    |
+| 2020-07-07 11:34:21 | 26   |
+| 2020-07-07 11:34:20 | 23    |
+| 2020-07-07 10:32:24 | 31    |
+| 2020-07-07 10:31:22 | 20    |
+| 2020-07-07 09:30:57 | 19    |
+
+For instance, you can `Include` all data points that have a **Temperature** lower than 30°C, you would get the following result :
```
oddlittlebird, line 279: What would this look like in the UI or field? You have shown me what I can do, but not how to do it.

Also, why is Include in code format but Temperature is bold?

oddlittlebird, line 279:
```suggestion
For instance, you can `Include` all data points that have a **Temperature** lower than 30°C, you would get the following result:
```

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
...
+
+| Time                | Temperature |
+|---------------------|--------------|
+| 2020-07-07 11:34:22 | 28    |
+| 2020-07-07 11:34:21 | 26   |
+| 2020-07-07 11:34:20 | 23    |
+| 2020-07-07 10:31:22 | 20    |
+| 2020-07-07 09:30:57 | 19    |
+
+You can also add more than one filter. For instance you could want to `Exclude` the temperature below 20°C, so you would add the corresponding filter.
```
oddlittlebird, line 289: Again, show me how to do this. Don't just tell us what is possible.

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
...
+
+It is important to understand how the filtering process works. All the filters added to the transformation are applied on the rows that match and, at the end of the transformation only, the rows marked as excluded are removed. This means one filter may overwrite the behavior of a previous filter if both match the same row. The filters are applied in order they appear in the transformation. You can chain the _Filter By Value_ transformation more than once to achieve the desired filtering if it is hard to achieve within the same transformation.
```
oddlittlebird, line 291: These three paragraphs need to be moved up under the first paragraph in this section. Also, rewrite them to address "you" and be less passive. Review the style guide linked earlier.

General format should be:

1st paragraph: Explain what the transformation does.
Next paragraphs: Go into more detail about it, including defining the filters and how multiple filters work. You might need subheadings to help your readers.
Example: Start with a subheading introducing the example, then show what the user can do and how.

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
...
+
+If your first filter is an `Include` filter, by default all the data is excluded, then the filters are applied. If your first filter is and `Exclude` filter, then by default all you data is included, then the filters are applied.
+
+Filters that are invalid or incompletely configured are ignored in the transformation (will not be applied).
```
oddlittlebird, line 295: Avoid parenthetical statements. Rewrite this without parentheses.

oddlittlebird (dismissed):

docs/sources/panels/transformations.md
Totalus: What I intend to do is to make simple sentences that translate to a configuration in the UI:

> You can `Include` all data points that have a **Temperature** lower than 30°C, you would get the following result

This translates to `Include` `Temperature` `Lower Than` `30` in the UI.

Maybe I should put all the parts in inline code `like this` instead, or all in bold (but bold you won't see the separation between `temperature` and `lower than`.

Totalus:
> 
> 
> For the UI, please use sentence case for field names and titles.
> 
> Per the [Documentation style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-style-guide.md), all UI text should be bold in the docs.

Where do you actually read that in the doc ? I searched `bold`, `field` or `UI` in the page of you linked and can't find anything about how the Field names and UI text should be in the documentation. Is it on another page ?

docs/sources/panels/transformations.md
oddlittlebird: Not in this case, since it is a table heading. The bold is intended to set UI text apart from other text in sentences.

oddlittlebird: I like what you're doing, but I skew towards less formatting rather than more. In this example, I would say, "...you can **Include** all data points that have a temperature lower than 30°C. You would get the following result:

oddlittlebird:
I apologize, I was unclear. Handling of UI text is covered in depth in the Microsoft style guide, particularly [Formatting text in instructions](https://docs.microsoft.com/en-us/style-guide/procedures-instructions/formatting-text-in-instructions). 

In our style guide, Line 15: For all items not covered in this guide, refer to the Microsoft Style Guide and the Chicago Manual of Style.

mckn (commented):
Wow, great job @Totalus! You are on fire when it comes to open PRs.

Currently we are adding functionality to filter out values in the Table visualization (in an excel-ish way) so I'm not sure this transformation is needed.

I guess the benefits of having a transformation like this would be that you can filter values for other types of visualizations in scenarios where you can't filter the result in the datasource query.

mckn:
Okay, so I had a chat with the rest of the team and it would be a great addition to the transforms. So lets try to get this mergeable. I will review it during the day.

docs/sources/panels/transformations.md
```diff
@@ -259,6 +259,41 @@ Here is the result after applying the `Series to rows` transformation.
 | 2020-07-07 09:30:57 | Humidity    | 33    |
 | 2020-07-07 09:30:05 | Temperature | 19    |
 
+
+## Filter By Value
```
mckn, line 263:
```suggestion
## Filter by value
```

packages/grafana-data/src/transformations/transformers/filterByValue.ts
```diff
@@ -0,0 +1,116 @@
...
+  filterType: ValueFilterID;
+}
+
+export interface FilterByValueTransformerOptions {
+  valueFilters: ValueFilter[];
+}
+
+export const filterByValueTransformer: DataTransformerInfo<FilterByValueTransformerOptions> = {
+  id: DataTransformerID.filterByValue,
+  name: 'Filter by Value',
```
mckn, line 21:
```suggestion
  name: 'Filter by value',
```

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+          }}
+        />
+      </>
+    );
+  }
+
+  return (
+    <div className="gf-form-inline">
+      <div className="gf-form gf-form-spacing">
+        <div className="gf-form-label width-8">Filter Type</div>
```
mckn, line 71:
```suggestion
        <div className="gf-form-label width-8">Filter type</div>
```

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+          value={config.type}
+          onChange={option => {
+            // console.log('onChange filterType', option.value);
+            onConfigChange({ ...config, type: option.value });
+          }}
+          menuPlacement="bottom"
+        />
+      </div>
+      <div className="gf-form gf-form-spacing">
+        <div className="gf-form-label width-8">Filter on Field</div>
```
mckn, line 87:
```suggestion
        <div className="gf-form-label width-8">Filter on field</div>
```

mckn:
I have been playing around a bit with the transformer and it seems to work really well! I found a couple of small things but nothing major.

I like the UI but would suggest that you select the first value as default instead of marking the field it as invalid. Feels less aggressive against the user and it follows the behavior of other forms in Grafana.

Skärmavbild 2020-08-28 kl  13 50 31

The labels being display in the UI should be written with first letter capital. An example is `Filter type` instead of `Filter Type`. I tried some different labels in the screen shot below:

Skärmavbild 2020-08-28 kl  14 04 56

Also - when you apply several filters the behavior isn't working as I expected. In the scenario above I would have assumed that we only keep rows where values are between 50 and 53.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+          invalid={filterExpressionInvalid}
+          defaultValue={config.filterExpression}
+          placeholder={filterInfo.placeholder}
+          onBlur={event => {
+            onConfigChange({ ...config, filterExpression: event.currentTarget.value });
+          }}
+        />
+      </>
+    );
+  }
```
mckn, line 66: You could do the check where you want to render it instead of doing it this way. The benefit of doing the check where you want to render it is that you don't have to jump up and down when you read the code.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+          options={filterTypeOptions}
+          value={config.filterType}
+          onChange={value => {
+            // console.log('onChange test', value);
+            onConfigChange({ ...config, filterType: value.value });
+          }}
+          menuPlacement="bottom"
+        />
+      </div>
+      <div className="gf-form gf-form--grow gf-form-spacing ">{filterOptionsInput}</div>
```
mckn, line 121:
```suggestion
      <div className="gf-form gf-form--grow gf-form-spacing ">
        {filterInfo.placeholder && (
          <Input
            className="flex-grow-1"
            invalid={filterExpressionInvalid}
            defaultValue={config.filterExpression}
            placeholder={filterInfo.placeholder}
            onBlur={event => {
              onConfigChange({ ...config, filterExpression: event.currentTarget.value });
            }}
          />
        )}
      </div>
```

I would suggest to do it this way.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { Select, Button, Input } from '@grafana/ui';
+
+import {
+  FilterByValueTransformerOptions,
+  ValueFilter,
+} from '@grafana/data/src/transformations/transformers/filterByValue';
+
+import { valueFiltersRegistry, ValueFilterID } from '@grafana/data/src/transformations/valueFilters';
+
+function FilterSelectorRow(props: any) {
```
mckn, line 21:
```suggestion
interface RowProps {
  fieldNameOptions: Array<SelectableValue<string>>;
  onDelete: () => void;
  onConfigChange: (config: ValueFilter) => void;
  config: ValueFilter;
  fieldType: FieldType;
}

const FilterSelectorRow: React.FC<RowProps> = props => {
```

I would change it to this so you get proper typings on your child component.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+  let filterTypeOptions = valueFiltersRegistry
+    .list()
+    .filter(element => {
+      if (element.supportedFieldTypes) {
+        return !element.supportedFieldTypes ? true : element.supportedFieldTypes.includes(fieldType);
+      } else {
+        return true;
+      }
+    })
+    .map(item => ({ value: item.id, label: item.name, description: item.description }));
```
mckn, line 36:
```suggestion
const filterTypeOptions = useMemo(() => {
    return valueFiltersRegistry
      .list()
      .filter(element => {
        if (!Array.isArray(element?.supportedFieldTypes)) {
          return true;
        }
        return element.supportedFieldTypes.includes(fieldType);
      })
      .map(item => ({
        value: item.id,
        label: item.name,
        description: item.description,
      }));
  }, [fieldType]);
```

If you use the `react hook` called `useMemo` like this we will only recalculate the filterTypeOptions if the `fieldType` change instead of doing it on every render.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  let filterInfo = valueFiltersRegistry.get(config.filterType);
+  let filterValid = filterInfo.getInstance({
+    filterExpression: config.filterExpression,
+    fieldType: fieldType,
+  }).isValid;
+
+  let fieldNameInvalid =
+    config.fieldName === null ||
+    fieldNameOptions.filter((item: Record<string, any>) => item.value === config.fieldName).length === 0;
```
mckn, line 46:
```suggestion
const fieldNameInvalid = config.fieldName === null || !fieldNameOptions.find(item => item.value === config.fieldName);
```

If you use the `.find` function it will stop looking in the list on first match instead of looping the whole list for all values matching the filter.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+  let filterTypeInvalid =
+    !fieldNameInvalid && filterInfo.supportedFieldTypes && !filterInfo.supportedFieldTypes.includes(fieldType);
+  let filterExpressionInvalid = !fieldNameInvalid && !filterTypeInvalid && !filterValid;
```
mckn, line 49:
```suggestion
  const filterExpressionInvalid = !fieldNameInvalid && !filterTypeInvalid && !filterValid;
```

I would suggest using `const` instead of `let` on all the places where the value of the variable never change.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+          options={fieldNameOptions}
+          value={config.fieldName}
+          invalid={fieldNameInvalid}
+          onChange={value => {
+            // console.log('onChange fieldName', value);
+            if (value === null) {
+              onConfigChange({ ...config, fieldName: null });
+            } else {
+              onConfigChange({ ...config, fieldName: value.value });
+            }
```
mckn, line 100:
```suggestion
           onConfigChange({
              ...config,
              fieldName: value?.value ?? null,
            })
```

Here is a smart trick to make the code a bit more compact. I am using the `value?.value` that is a safe way of accessing properties on an object when it might be undefined. If `value` if undefined it will return undefined instead of throwing an exception. I combine this together with the `??` that will return the value to the right if the value to the left is undefined so it will then be null.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+export const FilterByValueTransformerEditor: React.FC<TransformerUIProps<FilterByValueTransformerOptions>> = ({
+  input,
+  options,
+  onChange,
+}) => {
+  const fieldNames = useMemo(() => getAllFieldNamesFromDataFrames(input), [input]);
+  const fieldNameOptions = fieldNames.map((item: string) => ({ label: item, value: item }));
+
+  const onAddFilter = () => {
+    let valueFilters = options.valueFilters.map(filter => ({ ...filter })); // Deep copy
```
mckn, line 138: If you want to perform a deep clone you can easily do that with the lodash [cloneDeep](https://lodash.com/docs/4.17.15#cloneDeep) function.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+      filterType: ValueFilterID.regex,
+    });
+
+    onChange({
+      ...options,
+      valueFilters,
+    });
+  };
+
+  const onDeleteFilter = (index: number) => () => {
```
mckn, line 152: All of these `callback` functions that are defined in the render function should be wrapped in a [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) react-hook to prevent being created on each render.

mckn (changes requested):
Again, great work @Totalus! I have added some feedback (mostly on the editor UI). Please ping me when you feel that this PR is ready for some more testing.

Totalus:
**About the PR**
> Okay, so I had a chat with the rest of the team and it would be a great addition to the transforms. So lets try to get this mergeable. I will review it during the day.

Yes, I saw the other PR as well. They are indeed complementary and cover different use cases.

**About the UI**
> I like the UI but would suggest that you select the first value as default instead of marking the field it as invalid. Feels less aggressive against the user and it follows the behavior of other forms in Grafana.

It turns out to be much simple to just remove the red highlight when the field name is unspecified. Adding by default the first field is kind of tricky, but I just thought of a clean enough way to do it if you really want a default field selected.

I updated the different labels, I like your suggestion of `Rows with` and `Matching`.

**About the transformation behavior**

> Skärmavbild 2020-08-28 kl 14 04 56
> 
> Also - when you apply several filters the behavior isn't working as I expected. In the scenario above I would have assumed that we only keep rows where values are between 50 and 53.

Well yes and no. That is actually how I programmed it. I agree it is not intuitive if you don't previously read the doc and understand how the transformation works, but there is a reason that I made it that way. I don't know if there is a better way.

If we wanted to remove the rows before the end of the transformation (between each filter) like you implicitly suggest, we wouldn't be able to cover some use cases that imply filter overlapping (ex: I could not exclude the data between 50 and 53, whatever filter combination I use would end up excluding all the data).

If you want to achieve your desired filtering, you can do it two ways:

- With a single transformation

snapshot

- By chaining twice the transformation

snapshot

mckn:
> **About the PR**
> 
> > Okay, so I had a chat with the rest of the team and it would be a great addition to the transforms. So lets try to get this mergeable. I will review it during the day.
> 
> Yes, I saw the other PR as well. They are indeed complementary and cover different use cases.
> 
> **About the UI**
> 
> > I like the UI but would suggest that you select the first value as default instead of marking the field it as invalid. Feels less aggressive against the user and it follows the behavior of other forms in Grafana.
> 
> It turns out to be much simple to just remove the red highlight when the field name is unspecified. Adding by default the first field is kind of tricky, but I just thought of a clean enough way to do it if you really want a default field selected.
> 
> I updated the different labels, I like your suggestion of `Rows with` and `Matching`.
> 
> **About the transformation behavior**
> 
> > Skärmavbild 2020-08-28 kl 14 04 56
> > Also - when you apply several filters the behavior isn't working as I expected. In the scenario above I would have assumed that we only keep rows where values are between 50 and 53.
> 
> Well yes and no. That is actually how I programmed it. I agree it is not intuitive if you don't previously read the doc and understand how the transformation works, but there is a reason that I made it that way. I don't know if there is a better way.
> 

I think most users will not read the docs to start with. They will probably try to configure something on their own and if they get stuck they will look in the docs (best case scenario). Usually people open an issue if something doesn't work as they expected.

> If we wanted to remove the rows before the end of the transformation (between each filter) like you implicitly suggest, we wouldn't be able to cover some use cases that imply filter overlapping (ex: I could not exclude the data between 50 and 53, whatever filter combination I use would end up excluding all the data).
> 
I'm not really following here. Could you clarify a bit more? My initial thought was to build this so we verify that all rules is true for each row. Otherwise we should consider it a false and it should be excluded.

So basically loop through all the values. For each value check that all the filter rules is true for that value. If not, exclude it. Does this make sense?

> If you want to achieve your desired filtering, you can do it two ways:
> 
> * With a single transformation
> 
> snapshot
> 
> * By chaining twice the transformation
> 
> snapshot

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,263 @@
...
+      fieldName: null,
+      filterExpression: null,
+      filterType: ValueFilterID.regex,
+    });
+
+    onChange({
+      ...options,
+      valueFilters,
+    });
+  }, [options]);
```
mckn, line 158:
```suggestion
  }, [onChange , options]);
```
All the variables (mutable) that are from the outer scope should be part of the dependency list.

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const onDeleteFilter = useCallback(
+    (index: number) => () => {
+      let valueFilters = cloneDeep(options.valueFilters);
+      valueFilters.splice(index, 1);
+      onChange({
+        ...options,
+        valueFilters,
+      });
+    },
+    [options]
```
mckn, line 169:
```suggestion
    [options, onChange]
```

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const onConfigChange = useCallback(
+    (index: number) => (config: ValueFilter) => {
+      let valueFilters = cloneDeep(options.valueFilters);
+      valueFilters[index] = config;
+      onChange({
+        ...options,
+        valueFilters: valueFilters as ValueFilter[],
+      });
+    },
+    [options]
```
mckn, line 181:
```suggestion
    [options, onChange]
```

mckn:
I synced this branch with master and also fixed so we ignore the loos equality checks in the valueFilter.

Totalus:
>I think most users will not read the docs to start with. They will probably try to configure something on their own and if they get stuck they will look in the docs (best case scenario). Usually people open an issue if something doesn't work as they expected.

@mckn you are soooooo right about that.

>>    If we wanted to remove the rows before the end of the transformation (between each filter) like you implicitly suggest, we wouldn't be able to cover some use cases that imply filter overlapping (ex: I could not exclude the data between 50 and 53, whatever filter combination I use would end up excluding all the data).
>
> I'm not really following here. Could you clarify a bit more? My initial thought was to build this so we verify that all rules is true for each row. Otherwise we should consider it a false and it should be excluded.

If we verify all the rules, it's like we AND all of them, but then if you want to OR something, you're screwed. Its not exactly that, and its not exactly true, but it illustrates my point. Anyway, you're right, it's not intuitive enough.

> So basically loop through all the values. For each value check that all the filter rules is true for that value. If not, exclude it. Does this make sense?

It's another way. However, if we only include the values that match all rules, we won't be able to handle all use cases (ex: Including values between 0-10 and 20-30, I think this would be impossible. It would be possible if give the option to exclude, but not very intuitive to configure).

You gave me a good idea though. Here's what I suggest.

We `Include` or `Exclude` the values that match a set of rules. We also give the option to choose if we want to `Match all` or `Match any` of the rules. This should give enough flexibility to configure I think pretty much all of the use cases in an intuitive way.

We could also add a `Range` type filter that would match for a given range that we could input as an interval with braces (ex: `]1,10.2]` would match x for `1 < x <= 10.2`). This would make it even more straightforward for the user to include or exclude ranges and allow to filter multiple ranges within one single transformation.

How does that sound ? I'll look into it.

mckn:
> > I think most users will not read the docs to start with. They will probably try to configure something on their own and if they get stuck they will look in the docs (best case scenario). Usually people open an issue if something doesn't work as they expected.
> 
> @mckn you are soooooo right about that.
> 
> > > If we wanted to remove the rows before the end of the transformation (between each filter) like you implicitly suggest, we wouldn't be able to cover some use cases that imply filter overlapping (ex: I could not exclude the data between 50 and 53, whatever filter combination I use would end up excluding all the data).
> > 
> > 
> > I'm not really following here. Could you clarify a bit more? My initial thought was to build this so we verify that all rules is true for each row. Otherwise we should consider it a false and it should be excluded.
> 
> If we verify all the rules, it's like we AND all of them, but then if you want to OR something, you're screwed. Its not exactly that, and its not exactly true, but it illustrates my point. Anyway, you're right, it's not intuitive enough.
> 
> > So basically loop through all the values. For each value check that all the filter rules is true for that value. If not, exclude it. Does this make sense?
> 
> It's another way. However, if we only include the values that match all rules, we won't be able to handle all use cases (ex: Including values between 0-10 and 20-30, I think this would be impossible. It would be possible if give the option to exclude, but not very intuitive to configure).
> 
> You gave me a good idea though. Here's what I suggest.
> 
> We `Include` or `Exclude` the values that match a set of rules. We also give the option to choose if we want to `Match all` or `Match any` of the rules. This should give enough flexibility to configure I think pretty much all of the use cases in an intuitive way.
> 
> We could also add a `Range` type filter that would match for a given range that we could input as an interval with braces (ex: `]1,10.2]` would match x for `1 < x <= 10.2`). This would make it even more straightforward for the user to include or exclude ranges and allow to filter multiple ranges within one single transformation.
> 
> How does that sound ? I'll look into it.

I think this sounds really great! I'm a bit worried about creating an easy-to-use UI for this transform things but lets give it a try 💪

Totalus:
Updated transform to `Include/Exclude` and `Match all/Match any` over some conditions:

capture

Adding Range condition to match ranges:

capture

I think it makes it much more user friendly and intuitive, while keeping it flexible and simple to process. What do you think ?

mckn:
> Updated transform to `Include/Exclude` and `Match all/Match any` over some conditions:
> 
> capture
> 
> Adding Range condition to match ranges:
> 
> capture
> 
> I think it makes it much more user friendly and intuitive, while keeping it flexible and simple to process. What do you think ?

Great job trying out that concept! But after playing around a bit with the current version of the filter I think we might take a couple of steps back and simplify things a bit. The UI feels a bit too complex when having so many possible combinations.

After doing a demo on this I got an interesting suggestion on moving forward:

Simplify the transformation so we only allow one filter-row per added transform. Then we could remove the "matching" select and "Add condition" button. If you wan't to apply multiple filters you then need to add more transforms. The drawback is that it might not be possible to do a "or" filtering in this scenario.

Also, to make the `range` condition a bit more foolproof it would be nice to have two inputs. One for the start value and another one for the stop value.

What do you think about the suggestion above?

packages/grafana-data/src/transformations/transformers/filterByValue.ts
```diff
@@ -0,0 +1,131 @@
...
+                includeThisRow[row] = includeRow;
+              } else if (filterIndex === 0) {
+                includeThisRow[row] = !includeRow;
+              }
+            }
+          }
+        }
+
+        // Create the skeleton of the new data, copy original field attributes
+        let filteredFields: Field[] = [];
```
mckn, line 96:
```suggestion
        let filteredFields: MutableField[] = [];
```

If you change from a `Field` to a `MutableField` you will resolve the typing error that currently exists in this branch. You also need to import the `MutableField` by adding `import { MutableField } from '../../dataframe';` to the top of the file.

mckn (changes requested):
When we have decided on the final UI I will do a full review of the `filterByValue.ts` but apart of the typing error I think it looks good.

Totalus:
> Great job trying out that concept! But after playing around a bit with the current version of the filter I think we might take a couple of steps back and simplify things a bit. The UI feels a bit too complex when having so many possible combinations.
> 
> After doing a demo on this I got an interesting suggestion on moving forward:
> 
> Simplify the transformation so we only allow one filter-row per added transform. Then we could remove the "matching" select and "Add condition" button. If you wan't to apply multiple filters you then need to add more transforms. The drawback is that it might not be possible to do a "or" filtering in this scenario.

To me, this seems like a **major** drawback. Isn't it ? It would make some use cases hard or maybe even impossible to handle. 

Lets say you want to only to include a few string values. It translates to *`Include` if `Equals this` OR `Equals that` OR `Equals that`*. But since you can't *OR* conditions, to achieve that, you would have to exclude ALL of the other values. If you have 100 different values in this field, you're screwed:  you need to add one transform per value to exclude each of the 97 values you don't want to see in your data. And if you expect new values to come in for the same field eventually, you're definitely screwed.

> Also, to make the `range` condition a bit more foolproof it would be nice to have two inputs. One for the start value and another one for the stop value.

That is possible. The drawback here is you would not be able to specify if you want the min and max values to be included or not. You could be able to handle this with another condition of another filter to exclude the boundaries assuming we include them by default. Or you can be smart and put 1.00001 as min value instead of 1 to exclude the 1, but I find this solution more like a patch to a problem than a real solution.

mckn:
> > Great job trying out that concept! But after playing around a bit with the current version of the filter I think we might take a couple of steps back and simplify things a bit. The UI feels a bit too complex when having so many possible combinations.
> > After doing a demo on this I got an interesting suggestion on moving forward:
> > Simplify the transformation so we only allow one filter-row per added transform. Then we could remove the "matching" select and "Add condition" button. If you wan't to apply multiple filters you then need to add more transforms. The drawback is that it might not be possible to do a "or" filtering in this scenario.
> 
> To me, this seems like a **major** drawback. Isn't it ? It would make some use cases hard or maybe even impossible to handle.
> 
> Lets say you want to only to include a few string values. It translates to _`Include` if `Equals this` OR `Equals that` OR `Equals that`_. But since you can't _OR_ conditions, to achieve that, you would have to exclude ALL of the other values. If you have 100 different values in this field, you're screwed: you need to add one transform per value to exclude each of the 97 values you don't want to see in your data. And if you expect new values to come in for the same field eventually, you're definitely screwed.

Hmm, you are correct and I agree. Doesn't feel optimal. What do you think about moving back to your initial suggestion? Feels like it supports all the different scenarios that you pointed out.

> 
> > Also, to make the `range` condition a bit more foolproof it would be nice to have two inputs. One for the start value and another one for the stop value.
> 
> That is possible. The drawback here is you would not be able to specify if you want the min and max values to be included or not. You could be able to handle this with another condition of another filter to exclude the boundaries assuming we include them by default. Or you can be smart and put 1.00001 as min value instead of 1 to exclude the 1, but I find this solution more like a patch to a problem than a real solution.

I see your point and I would probably create two different options with two inputs each.

1. Range with input from - to.
2. Between with input from - to.

Totalus:
> What do you think about moving back to your initial suggestion? Feels like it supports all the different scenarios that you pointed out.

It does, but it it is much less intuitive to the user in my opinion. Even I had to think a bit more to implement the filtering properly, and you thought the filter wasn't working. The current implementation much more intuitive and easier to use while having all the required flexibility to cover any use cases easily.

> I see your point and I would probably create two different options with two inputs each.
>
>  1. Range with input from - to.
>  2. Between with input from - to.

For a start lets just keep `Range` and include the boundaries. I think it will be enough, otherwise it gets confusing as of what is the difference between the two. If the user really wants to exclude the boundaries, he can use a second condition for that.

mckn:
> > What do you think about moving back to your initial suggestion? Feels like it supports all the different scenarios that you pointed out.
> 
> It does, but it it is much less intuitive to the user in my opinion. Even I had to think a bit more to implement the filtering properly, and you thought the filter wasn't working. The current implementation much more intuitive and easier to use while having all the required flexibility to cover any use cases easily.
> 

OK ok, I am currently playing around with different ways of structuring the UI to make it super easy to use. Looking on similar examples in the Grafana project and going through our [design system](https://developers.grafana.com/ui/latest/index.html?path=/story/*) for some inspiration.

Feel free to pitch in any ideas!

> > I see your point and I would probably create two different options with two inputs each.
> > 
> > 1. Range with input from - to.
> > 2. Between with input from - to.
> 
> For a start lets just keep `Range` and include the boundaries. I think it will be enough, otherwise it gets confusing as of what is the difference between the two. If the user really wants to exclude the boundaries, he can use a second condition for that.

Totalus:
> OK ok, I am currently playing around with different ways of structuring the UI to make it super easy to use. Looking on similar examples in the Grafana project and going through our [design system](https://developers.grafana.com/ui/latest/index.html?path=/story/*) for some inspiration.

I'm not finding the UI overloaded or particularly hard to use. What makes you suggest that it requires a rework ?

We could use radio buttons for the **Include/Exclude** and **Match Any/Match All** options, there is space for that. To me it does not make a big difference though.

mckn:
> > OK ok, I am currently playing around with different ways of structuring the UI to make it super easy to use. Looking on similar examples in the Grafana project and going through our [design system](https://developers.grafana.com/ui/latest/index.html?path=/story/*) for some inspiration.
> 
> I'm not finding the UI overloaded or particularly hard to use. What makes you suggest that it requires a rework ?
> 
> We could use radio buttons for the **Include/Exclude** and **Match Any/Match All** options, there is space for that. To me it does not make a big difference though.

Well, we need to do a UI that is easy to use and consistent with the rest of the UI. So we would probably need to tweak this a bit.

mckn:
> > Well, we need to do a UI that is easy to use and consistent with the rest of the UI. So we would probably need to tweak this a bit.
> 
> Ok, but I'm not sure to understand what makes it consistent vs inconsistent, or hard to use, and what you mean by the _rest of the UI_, so I'm kind of clueless on how you want it to change...

Sorry, that wasn't my intention. I'm not sure how to structure it either but I am trying different ways and will hopefully get back to you with a suggestion that might work.

Totalus:
- I've added two inputs for the Range filter:

snapshot

- Now the **Match All / Match Any** option only appears when there is more than one condition (not relevant when only one condition).

- I've tested with radio buttons, but I'm not it's better:

radio-button

Totalus:
I've changed the top config for *ButtonSelect* components, and integrated the labels that were in between, I think it's cleaner this way:

Snapshot

mckn:
@Totalus what do you think about the following approach?

Skärmavbild 2020-09-08 kl  16 18 54

I tried to group the conditions together and I added a field that acts as a header with "match all/any" in it so it feels like it is connected to that group.

Also some feedback on the naming of the filter:

`Greater` -> `Greater than`
`Lower` -> `Lower than`
`Greater or Equal` -> `Greater than or equal`
`Lower or Equal` -> `Lower than or equal`
`Different` -> `Not equal`
`Is Null` -> `Is null`
`Is Not Null` -> `Is not null`

When I am testing the `Greater` is seems to work the same way as the `Greater than or equal` so I think we have a bug somewhere around that one.

Totalus:
> 
> 
> @Totalus what do you think about the following approach?
> 
> Skärmavbild 2020-09-08 kl 16 18 54
> 
> I tried to group the conditions together and I added a field that acts as a header with "match all/any" in it so it feels like it is connected to that group.

That's very nice !

> Also some feedback on the naming of the filter:
> 
> `Greater` -> `Greater than`
> `Lower` -> `Lower than`
> `Greater or Equal` -> `Greater than or equal`
> `Lower or Equal` -> `Lower than or equal`
> `Different` -> `Not equal`
> `Is Null` -> `Is null`
> `Is Not Null` -> `Is not null`

Sure, no problem with that

> When I am testing the `Greater` is seems to work the same way as the `Greater than or equal` so I think we have a bug somewhere around that one.

I've double checked the code, but it seems ok and the corresponding tests are passing. If you are playing with floats, you probably won't see the difference. Also the floats are rounded by default in the table view, so you don't actually see the real value with all decimals which can sometimes cause some confusion.

Can you merge/push the UI modifications ?

mckn:
> > @Totalus what do you think about the following approach?
> > Skärmavbild 2020-09-08 kl 16 18 54
> > I tried to group the conditions together and I added a field that acts as a header with "match all/any" in it so it feels like it is connected to that group.
> 
> That's very nice !
> 
> > Also some feedback on the naming of the filter:
> > `Greater` -> `Greater than`
> > `Lower` -> `Lower than`
> > `Greater or Equal` -> `Greater than or equal`
> > `Lower or Equal` -> `Lower than or equal`
> > `Different` -> `Not equal`
> > `Is Null` -> `Is null`
> > `Is Not Null` -> `Is not null`
> 
> Sure, no problem with that
> 
> > When I am testing the `Greater` is seems to work the same way as the `Greater than or equal` so I think we have a bug somewhere around that one.
> 
> I've double checked the code, but it seems ok and the corresponding tests are passing. If you are playing with floats, you probably won't see the difference. Also the floats are rounded by default in the table view, so you don't actually see the real value with all decimals which can sometimes cause some confusion.
> 
> Can you merge/push the UI modifications ?

Yes, I will try to do it later today.

mckn:
I will do one more round of reviewing and see what's left for this to get merge.

Totalus:
I updated the tests. Is there anything missing from my end ?

public/app/core/components/TransformersUI/FilterByValueTransformerEditor.tsx
```diff
@@ -0,0 +1,303 @@
...
+  input,
+  options,
+  onChange,
+}) => {
+  const styles = getEditorStyles();
+  const fieldsInfo = useMemo(() => getAllFieldInfoFromDataFrames(input), [input]);
+  const fieldNameOptions = useMemo(
+    () =>
+      fieldsInfo
+        .filter((item: Record<string, any>) => item.type !== FieldType.time)
```
mckn, line 147: @Totalus just so I understans. Why do we want to exclude the time field?

Totalus: I commented on that before:

>There is one thing I would improve. As it is right now, it is possible to apply a few filter types on a time field, but the filtering will be confusing for the user. The time is displayed in an text format (ISO or other), but the filtering is applied on the numeric (epoch) value.
>
> I see two possibilities:
>
> The first is we remove the filtering on time fields. There might not be much use cases for filtering on time since we specify the time range in the dashboard anyway, so its more than often part of the query and results are returned accordingly.
>
>  The second possibility is to convert the value to ISO format before filtering (for filter like Regex, or equal) or in other cases (for filter like greater), convert a filter expression to epoch time. This way the user could say Include Time Greater Than 2020-08-22, and they would get the desired result. To do that however I need to put the finger on the functions that are used to format the time values for display. If someone can point them out, I would investigate this second possibility.

There are also more subtleties to discuss if we want to implement filtering for the time field. Things like :

- In which format the user be able to input the time ?
- If the user inputs the time in ISO, should we keep the ISO format in the filter expression or should we convert to epoch so it becomes independent of the time zone ? Because otherwise it would filter differently in different time zones.
- When you enter a Regex, does it filter on the displayed value ? I guess the displayed, but then you are affected by the time zone again because your Regex will not be converted for each time zone.

So since I didn't get any comment on that and I didn't want to slow down the merge with additional discussion, I figured for now I would just remove the time option and keep that for a future PR. We could create a new issue or wait for a user to ask for its support, see if it's really a need.

mckn: Got it! I agree - lets wait with this and open a separate PR for this if it is needed.

mckn:
> I updated the tests. Is there anything missing from my end ?

Awesome work! I have limited time to spend on this. Have another project I need to give some attention but I think we need to change the way we have structured the ValueFilter and the UI for those. The solution that we have added a expression2 is not optimal and I think we can do some small changes to make it a bit easier to extend in the future.

Totalus:
> 
> 
> > I updated the tests. Is there anything missing from my end ?
> 
> Awesome work! I have limited time to spend on this. Have another project I need to give some attention but I think we need to change the way we have structured the ValueFilter and the UI for those. The solution that we have added a expression2 is not optimal and I think we can do some small changes to make it a bit easier to extend in the future.

What would you suggest ? @mckn 

Those are the current structures:
```js
export interface ValueFilter {
  fieldName: string | null; // Corresponding field name
  filterExpression: string | null; // The filter expression / value
  filterExpression2: string | null;
  filterType: ValueFilterID;
}

export interface FilterByValueTransformerOptions {
  valueFilters: ValueFilter[];
  type: string; // 'include' or 'exclude'
  match: string; // 'all' or 'any'
}
```

`FilterByValueTransformerOptions` seems ok to me. I agree that `ValueFilter` is not super elegant, but I ain't sure how we should improve it. Here are some possibilities I can think of:

- We could use make *filterExpression* an Array to store an arbitrary number of filter expressions. I'm not sure if there are much use cases for more than 2 or 3 *filterExpressions*, so I'm wondering if it really worth using an Array.
- We could somehow encode multiple filter expression all in the same *filterExpression* field as string, but I don't think that's a very smart solution. It might cause some issues depending on what the user inputs.
- We could maybe have *filterExpression* as an object containing different fields for each type of filter expression (min, max, regex, equal, etc.). This seems to me like a bad solution as it would mean handling each of them individually in the UI somehow. Not sure it would be pretty or easily manageable. And there might be issues with ordering as well (handling in which order each of the filterExpression input field appears).

After this thought, I think the first option (array) would be the most decent solution, from the ones I can think of.

Do you have any other ideas ?

mckn:
> ```js
> export interface ValueFilter {
>   fieldName: string | null; // Corresponding field name
>   filterExpression: string | null; // The filter expression / value
>   filterExpression2: string | null;
>   filterType: ValueFilterID;
> }
> 
> export interface FilterByValueTransformerOptions {
>   valueFilters: ValueFilter[];
>   type: string; // 'include' or 'exclude'
>   match: string; // 'all' or 'any'
> }
> ```

hmm, what do you think about the following:

```ts
export interface ValueFilter {
  fieldName: string | null; // Corresponding field name
  filterArgs: Record<string, any> | null; // to store the filter arguments
  filterType: ValueFilterID;
}

export interface FilterByValueTransformerOptions {
  valueFilters: ValueFilter[];
  type: string; // 'include' or 'exclude'
  match: string; // 'all' or 'any'
}

export interface ValueFilterInfo extends RegistryItem {
  getInstance: ValueFilterInstanceCreator;
  supportedFieldTypes?: FieldType[]; // If defined, support only those field types
  component: React.ReactNode; // the UI for the filter that will take filterArgs, getInstance, and the other props for updating/deleting etc.
}
```

Then in the UI we could basically just render the component. The component can then fetch the values by it's name from the filterArgs and use the getInstance.test to verify those.

mckn:
> > ```js
> > export interface ValueFilter {
> >   fieldName: string | null; // Corresponding field name
> >   filterExpression: string | null; // The filter expression / value
> >   filterExpression2: string | null;
> >   filterType: ValueFilterID;
> > }
> > 
> > export interface FilterByValueTransformerOptions {
> >   valueFilters: ValueFilter[];
> >   type: string; // 'include' or 'exclude'
> >   match: string; // 'all' or 'any'
> > }
> > ```
> 
> hmm, what do you think about the following:
> 
> ```ts
> export interface ValueFilter {
>   fieldName: string | null; // Corresponding field name
>   filterArgs: Record<string, any> | null; // to store the filter arguments
>   filterType: ValueFilterID;
> }
> 
> export interface FilterByValueTransformerOptions {
>   valueFilters: ValueFilter[];
>   type: string; // 'include' or 'exclude'
>   match: string; // 'all' or 'any'
> }
> 
> export interface ValueFilterInfo extends RegistryItem {
>   getInstance: ValueFilterInstanceCreator;
>   supportedFieldTypes?: FieldType[]; // If defined, support only those field types
>   component: React.ReactNode; // the UI for the filter that will take filterArgs, getInstance, and the other props for updating/deleting etc.
> }
> ```
> 
> Then in the UI we could basically just render the component. The component can then fetch the values by it's name from the filterArgs and use the getInstance.test to verify those.

Okay, let's do like this to keep this PR in an okay scope.

The reason that we have the `filterExpression2` is that we added the range filter. So let's take a couple of steps back. Remove the range filter operator (just to keep it simple), clean up the code and get this PR merged.

When that is merged. Open a new PR where we add the range filter functionality and also do some refactoring of the code structure to support that scenario. We have similar pattern on other places of the code base so this is probably a pattern that we want to introduce as a concept on more places.

Sorry for all the back and forth but I think it is good to keep this as small as possible.

Totalus:
> hmm, what do you think about the following:
> 
> ```ts
> export interface ValueFilter {
>   fieldName: string | null; // Corresponding field name
>   filterArgs: Record<string, any> | null; // to store the filter arguments
>   filterType: ValueFilterID;
> }
> 
> export interface FilterByValueTransformerOptions {
>   valueFilters: ValueFilter[];
>   type: string; // 'include' or 'exclude'
>   match: string; // 'all' or 'any'
> }
> 
> export interface ValueFilterInfo extends RegistryItem {
>   getInstance: ValueFilterInstanceCreator;
>   supportedFieldTypes?: FieldType[]; // If defined, support only those field types
>   component: React.ReactNode; // the UI for the filter that will take filterArgs, getInstance, and the other props for updating/deleting etc.
> }
> ```
> 
> Then in the UI we could basically just render the component. The component can then fetch the values by it's name from the filterArgs and use the getInstance.test to verify those.

Ouh, nice one ! Let me see if I can pull it off today. If we change to *filterArgs* sooner we won't be stuck with a *filterExpression* string later.

Totalus:
Are we actually ok with moving some UI code to the `grafana-data` package ? Putting JSX in the `valueFilters.ts` forces us to rename it `valuFilter.tsx`. Is this against some Grafana design rules somewhere that tells us keep the UI code not in some part of the code structure, not in others ? I only see one other (non-test) .tsx file in all the `grafana-data` package and it does not actually contain any JSX nor UI code.

Totalus:
> 
> 
> Are we actually ok with moving some UI code to the `grafana-data` package ? Putting JSX in the `valueFilters.ts` forces us to rename it `valuFilter.tsx`. Is this against some Grafana design rules somewhere that tells us keep the UI code not in some part of the code structure, not in others ? I only see one other (non-test) .tsx file in all the `grafana-data` package and it does not actually contain any JSX nor UI code.

Is it possible that the code from each package (packages/*) must be independent from one another ?


If yes than there's an issue with this design pattern. We can't link any UI related code in the *grafana-data* package (where the `valueFitlers.ts` file belongs) because any UI code will end up depending on the *grafana-ui* package.

A work around would be to create a component registry that will point to react nodes. The UI would retrieve those nodes corresponding to each filter type on render so the corresponding UI code would not reside in the *grafana-data* package anymore.

Totalus:
There you go. The options structure is definitely nicer with `filterArgs`. The design pattern seems fairly good. It splits a little bit the code around but it's not so bad. Check it out.

> So let's take a couple of steps back. Remove the range filter operator (just to keep it simple), clean up the code and get this PR merged.

Feel free to do that if you that if you prefer and think there should be more rethinking/rework.

mckn:
> There you go. The options structure is definitely nicer with `filterArgs`. The design pattern seems fairly good. It splits a little bit the code around but it's not so bad. Check it out.
> 
> > So let's take a couple of steps back. Remove the range filter operator (just to keep it simple), clean up the code and get this PR merged.
> 
> Feel free to do that if you that if you prefer and think there should be more rethinking/rework.

As you highlighted there are a couple of issues with adding this to the @grafana/data and we have the issue that I mentioned about the way we structure the filter components ui.

Both of those things could be solved if we simplify and exclude the range filter in this PR and open another PR for that which then will include a better design pattern.

Totalus:
> As you highlighted there are a couple of issues with adding this to the @grafana/data and we have the issue that I mentioned about the way we structure the filter components ui.

I worked that out as I highlighted, but anyway.

> Both of those things could be solved if we simplify and exclude the range filter in this PR and open another PR for that which then will include a better design pattern.

I reverted it.

Totalus:
What's up ?

mckn:
> What's up ?

Hi there! Sorry but I have been busy with another project. Will have a look at this asap.

Totalus:
If we are to postpone this to 7.4, I suggest we consider directly the version before the revert (commit 26884) with the Range filter and the different UI design pattern. It is more featured and has a better option structure. I find it more final and closer to what we want than the reverted (current) version.

Also can you inform on why it is being postpone to 7.4 ? Apart for final code review, there is not much work left to be done on this PR I believe.

achatterjee-grafana (approved):
LGTM

mckn:
@Totalus I will start to look into this starting today. I needed to focus on a couple of other projects but I have some things we need to change here before we can merge this PR.

1. We need to sync this PR with master.

2. I think we should change the `ValueFilters` a bit since they are located within @grafana/data and will be available for other plugin developers. So I think we need to do some changes to how that is done. I will get back to you with some suggestions.

Totalus:
> 
> 
> @Totalus I will start to look into this starting today. I needed to focus on a couple of other projects but I have some things we need to change here before we can merge this PR.
> 
>1. We need to sync this PR with master.
> 
> 2. I think we should change the `ValueFilters` a bit since they are located within @grafana/data and will be available for other plugin developers. So I think we need to do some changes to how that is done. I will get back to you with some suggestions.

Sure, let me know your suggestions.

Also can we consider commit 26884 as I highlighted previously ? Reference:
> 
> 
> If we are to postpone this to 7.4, I suggest we consider directly the version before the revert (commit 26884) with the Range filter and the different UI design pattern. It is more featured and has a better option structure. I find it more final and closer to what we want than the reverted (current) version.
>

mckn:
> > @Totalus I will start to look into this starting today. I needed to focus on a couple of other projects but I have some things we need to change here before we can merge this PR.
> > 
> > 1. We need to sync this PR with master.
> > 2. I think we should change the `ValueFilters` a bit since they are located within @grafana/data and will be available for other plugin developers. So I think we need to do some changes to how that is done. I will get back to you with some suggestions.
> 
> Sure, let me know your suggestions.
> 
> Also can we consider commit 26884 as I highlighted previously ? Reference:
> 
> > If we are to postpone this to 7.4, I suggest we consider directly the version before the revert (commit 26884) with the Range filter and the different UI design pattern. It is more featured and has a better option structure. I find it more final and closer to what we want than the reverted (current) version.

I can't find that commit. Could you link it again?

Totalus:
Sorry, the correct commit is dab000a71095f4bd8ef88c2ee8c5078d2de116be.

mckn:
> > I can't find that commit. Could you link it again?
> 
> @mckn Sorry, the correct commit is [dab000a](https://github.com/grafana/grafana/commit/dab000a71095f4bd8ef88c2ee8c5078d2de116be).

I will push a branch with some suggestions tomorrow. Stay tuned 🎵

Totalus:
Read the comments from https://github.com/grafana/grafana/pull/26884#issuecomment-692903630 up to Sept 16th to understand the changes that were made. It was from your own suggestion @mckn.

mckn:
@Totalus I have open a small PR with some suggestions on how to do this:

https://github.com/Totalus/grafana/pull/1/files

Totalus:
> 
> 
> @Totalus I have open a small PR with some suggestions on how to do this:
> 
> https://github.com/Totalus/grafana/pull/1/files

Thanks, I took a quick look and understand, however it implies more rework than I expected and I'm on other projects at the moment. I suggest you take it from here if that is ok with you.

mckn:
> > @Totalus I have open a small PR with some suggestions on how to do this:
> > https://github.com/Totalus/grafana/pull/1/files
> 
> Thanks, I took a quick look and understand, however it implies more rework than I expected and I'm on other projects at the moment. I suggest you take it from here if that is ok with you.

Ok ok, I'm not sure I have permissions to push to this branch anymore. Have you changed it @Totalus?

Totalus:
> Ok ok, I'm not sure I have permissions to push to this branch anymore. Have you changed it @Totalus?

@mckn I didn't change anything. You should have write permission. If it does not work, just make a PR and I'll merge it.

mckn:
@Totalus can you sync your branch with master?

docs/sources/panels/transformations.md
```diff
@@ -340,6 +340,67 @@ Here is the result after applying the `Series to rows` transformation.
...
+|---------------------|-------------|----------
+| 2020-07-07 11:34:23 | 32          | 101
+| 2020-07-07 11:34:22 | 28          | 125
+| 2020-07-07 11:34:21 | 26          | 110
+| 2020-07-07 11:34:20 | 23          | 98
+| 2020-07-07 10:32:24 | 31          | 95
+| 2020-07-07 10:31:22 | 20          | 85
+| 2020-07-07 09:30:57 | 19          | 101
+
+If you **Include** the data points that have a temperature lower than 30°C. The configuration will look as followed:
```
osg-grafana, line 376: If you **Include** the data points that have a temperature lower than 30°C. The configuration will look as followed:

-->

If you **Include** the data points that have a temperature below 30°C, the configuration will look as follows:

Note: there is the word "below" on line 381, which is why I changed "lower than" to "below"; this way, you are using parallel language.

docs/sources/panels/transformations.md
```diff
@@ -340,6 +340,67 @@ Here is the result after applying the `Series to rows` transformation.
...
+
+| Time                | Temperature | Altitude
+|---------------------|-------------|----------
+| 2020-07-07 11:34:22 | 28          | 125
+| 2020-07-07 11:34:21 | 26          | 110
+| 2020-07-07 11:34:20 | 23          | 98
+| 2020-07-07 10:31:22 | 20          | 85
+| 2020-07-07 09:30:57 | 19          | 101
+
+You can add more than one condition to the filter. We could want to include the data only if the altitude is higher than 100. To do that we can add a condition to get the following configuration:
```
osg-grafana, line 392: You can add more than one condition to the filter. We could want to include the data only if the altitude is higher than 100. To do that we can add a condition to get the following configuration:

-->

You can add more than one condition to the filter. For example, you might want to include the data only if the altitude is greater than 100. To do so, add that condition to the following configuration:

mckn:
@Totalus feel free to have a look at the current structure. Any feedback or thoughts are welcome 🙏

packages/grafana-data/src/transformations/transformers/filterByValue.ts
```diff
@@ -0,0 +1,133 @@
...
+
+            const matching = matchAll
+              ? filters.every(checkIfFilterIsMatchingRow)
+              : !!filters.find(checkIfFilterIsMatchingRow);
+
+            if (matching) {
+              rows.add(index);
+            }
+          }
+        }
```
Totalus, line 84: In the original implementation I was looping over the `filters` array first, and then on the rows. The reason was to avoid having to re-create the matcher function (given by `getValueMatcher()` here) for each row.

When you check each row for one filter at the time, you only have to call `getValueMatcher()` function once per filter (so `fiilter.length` times total). If you loop over the rows first like you are doing here, you have to call `filters.length` times the `getValueMatcher()` function **per row** (so `filter.length` x `rows.length` total), which seems much less optimal speed wise. Not sure how much of a difference it makes, but it certainly adds quite a few operation to execute inside the loop.

torkelo (commented):
Did a quick initial pass through the code, the value matcher stuff looks really great! Will do some more UI testing now

packages/grafana-data/src/transformations/transformers/filterByValue.ts
mckn: Yeah, agree! I will look into making this a bit more performant.

mckn:
> Would it make sense to name this `upload-packages-enterprise` (and on line 2462 `upload-packages-oss`) so it is clear that that flag is turned on? It took me several seconds to spot the difference.

Maybe, but this PR does not change that file so it should probably be done in some other PR 👍

packages/grafana-data/src/transformations/transformers/filterByValue.ts
mckn: I have pushed some changes regarding this @torkelo @Totalus. Please have a look.

torkelo:
Made some UX fixes and improvements to alignment and fixed issue where gf-form was nested (was added again in the matcher editors) creating double margins 

Screenshot from 2020-11-27 10-27-28

torkelo (approved):
Awesome work @Totalus and @mckn on this one, think this should be good to merge

mckn (approved):

PR #26892, opened 2020-08-10
CI: Upgrade golangci-lint and base image

aknuds1:
For our CI build image, upgrade golangci-lint and the base image (Debian Stretch). Also avoid the anti-pattern of upgrading Debian packages when building Docker images.


scripts/build/ci-build/Dockerfile
```diff
@@ -76,28 +71,31 @@ RUN cd /tmp && \
...
-FROM debian:stretch-20200607
+FROM debian:stretch-20200803
 
 ENV GOVERSION=1.14.7 \
     PATH=/usr/local/go/bin:$PATH \
     GOPATH=/go \
     NODEVERSION=12.18.3-1nodesource1 \
     YARNVERSION=1.22.4-1
 
+# Use ARG so as not to persisten environment variable in image
```
papagian, line 92: I don't fully understand this comment; especially because such an environmental variable is defined [above](https://github.com/grafana/grafana/pull/26892/files#diff-c5eb1190a4c6100cfa6e45d2452b0ec8R9). Do we need both?

aknuds1: @papagian It means that `ARG` is preferable to `ENV` if you don't want to persist an environment variable in the image. The problem with `ENV` is that for example `DEBIAN_FRONTEND` becomes also an environment variable in the image, and containers executed from this image will have DEBIAN_FRONTEND=noninteractive in their environment. We don't want to set that (it's only relevant while building the image), and it could have a bad effect for users of the image. Did it become clear now?

papagian: that makes sense, thanks for the explanation!

papagian (approved):
LGTM

sakjur (approved):

scripts/build/ci-build/Dockerfile
aknuds1: You're welcome :)

Issue #26912, opened 2020-08-11
Flux datasource leaking TCP connections


coussej:
**What happened**:
When running a flux query, 2 tcp connection to influxdb are opened and not closed. Over time, this results in grafana not accepting tcp connections with the following error: 
`http: Accept error: accept tcp [::]:3000: accept4: too many open files; retrying in 1s`

I the following GIF, I refresh a single flux query a few times, resulting in 2 extra open connections to influx each time (command:    `netstat -atn | grep ":8086" | wc -l`

![Aug-11-2020 09-07-20](https://user-images.githubusercontent.com/9944663/89868257-cb518380-dbb2-11ea-912d-343fc2c566d6.gif)

**What you expected to happen**:
Grafana not to crash every other day :-) 

**How to reproduce it (as minimally and precisely as possible)**:
Set up a panel with flux query and monitor TCP connections on port 8086.

**Anything else we need to know?**:
This was already mentioned in #26731, and a fix was released in 7.1.3, but the issue persists.
 
**Environment**:
- Grafana version: 7.1.3
- Data source type & version: InfluxDB 1.8.1 with flux.
- OS Grafana is installed on: Ubuntu
- User OS & Browser: Any
- Grafana plugins: /
- Others:

aknuds1:
You mean Grafana version 7.1.3?

coussej:
@aknuds1 yes of course 🤦🏻‍♂️ Updated.

aknuds1:
I've made a PR that will hopefully help, by closing idle HTTP connections: https://github.com/grafana/grafana/pull/26917.

aknuds1:
In the interest of reproducing, can you specify more clearly how you configure your query?

coussej:
So, query is run agains the influx 1.8.1 flux compatibility endpoint. Just replace the series below with anything valid for your system. Run the query a few times while running `watch 'netstat -atn | grep ":8086" | wc -l'` on your grafana host.

```
f1 = from(bucket: "simulation_tanksystem")
  |> range(start: v.timeRangeStart, stop:v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "WP1_fill_level")
  |> aggregateWindow(every: 30s, timeSrc: "_start", fn: mean, createEmpty:false)

f2 = from(bucket: "simulation_tanksystem")
  |> range(start: v.timeRangeStart, stop:v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "WP2_fill_level")
[...]

PR #26917, opened 2020-08-11
InfluxDB: Close Flux query results

aknuds1:
Our InfluxDB integration seems to leak connections in Flux mode, according to #26912. I'm attempting to fix this by calling `Close` on query results from the InfluxDB Go client. This should result in HTTP response bodies getting closed. Not 100% sure if it fixes the issue, but ought to help.

Also perform some general cleanup.


wbrowne (approved):
LGTM! Nice work on the cleanup too 👍 

squeaky clean

bergquist:
Hm. I'm not sure we want to force idle connections. I think we should leave connection management to the httpclient. AFAIK we don't do this anyway else in Grafana.

I think we just have a bug where connections are not reused. https://stackoverflow.com/questions/17948827/reusing-http-connections-in-golang

aknuds1:
@bergquist Thanks, will look into it!

bergquist:
> do you actually mean the HTTP transport? The client is ephemeral

Yes I do :)

aknuds1:
@bergquist Doesn't [this code](https://github.com/golang/go/blob/master/src/net/http/transfer.go#L996) mean that your pattern of discarding the request body before close is now automatic? There's at least one comment in the go-nuts thread you posted indicating the same.

aknuds1:
I figure that the next step might be to try to reproduce #26912, to see for ourselves what might cause connection leaks. I've asked the author to provide more details on how to reproduce.

papagian (approved):
LGTM!

pkg/tsdb/influxdb/flux/executor.go
```diff
@@ -32,6 +33,7 @@ func ExecuteQuery(ctx context.Context, query QueryModel, runner queryRunner, max
 		dr.Frames = append(dr.Frames, metaFrame)
 		return
 	}
+	defer tables.Close()
```
coussej, line 36: Shouldn't this be before the error check? In case of an error the tables won't be closed.

aknuds1: @coussej No, I don't think that would be a good idea, because if an error happened, you can expect the non-error return value to be invalid (a nil pointer in this case).

papagian: the defer statement will run immediately before the `executeQuery` returns

coussej: Yes my bad! Ignore my comment.

PR #26932, opened 2020-08-11
API: Support paging in the admin orgs list API

### LINKED ISSUE #26171: API: Support paging in the admin orgs list api
Right now the org query has a hardcoded limit of 1000. https://github.com/grafana/grafana/blob/3223d4b4190b0b2b25d75d9dbc54e2cbbb30bbdb/pkg/api/org.go#L167

It would be nice if that (and other query limits) were configurable.

**Why is this needed**:
To allow queries to return more than 1000 orgs if there are more than 1000 orgs; or other queryable object (like the page limit on users.
https://github.com/grafana/grafana/blob/3223d4b4190b0b2b25d75d9dbc54e2cbbb30bbdb/pkg/api/user.go#L286

benjaminjb:
Allow users to set page limits and page for org list.

Fixes #26171

I tried to follow the pattern set by the `/users` route (since it allows pagination and page limits), but found that setting the page to 1 as the default skipped entries that were present on page 0.

### DIFF
--- a/pkg/services/sqlstore/org_test.go
+++ b/pkg/services/sqlstore/org_test.go
@@ -36,37 +36,31 @@ func TestAccountDataAccess(t *testing.T) {
 		})
 
 		Convey("Given we have organizations, we can limit and paginate search", func() {
-			var err error
-			var cmd *models.CreateOrgCommand
-			ids := []int64{}
-
 			for i := 1; i < 4; i++ {
-				cmd = &models.CreateOrgCommand{Name: fmt.Sprint("Org #", i)}
-				err = CreateOrg(cmd)
+				cmd := &models.CreateOrgCommand{Name: fmt.Sprint("Org #", i)}
+				err := CreateOrg(cmd)
 				So(err, ShouldBeNil)
-
-				ids = append(ids, cmd.Result.Id)
 			}
 
-			Convey("Should be able to limit search", func() {
-				query := &models.SearchOrgsQuery{Limit: 1}
-				err = SearchOrgs(query)
-
-				So(err, ShouldBeNil)
-				So(len(query.Result), ShouldEqual, 1)
-			})
-
-			Convey("Should be able to paginate search", func() {
-				query := &models.SearchOrgsQuery{Page: 0}
-				err = SearchOrgs(query)
+			Convey("Should be able to search with defaults", func() {
+				query := &models.SearchOrgsQuery{}
+				err := SearchOrgs(query)
 
 				So(err, ShouldBeNil)
 				So(len(query.Result), ShouldEqual, 3)
 			})
 
+			Convey("Should be able to limit search", func() {
+				query := &models.SearchOrgsQuery{Limit: 1}
+				err := SearchOrgs(query)
+
+				So(err, ShouldBeNil)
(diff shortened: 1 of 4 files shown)


pkg/services/sqlstore/org_test.go
```diff
@@ -35,6 +35,44 @@ func TestAccountDataAccess(t *testing.T) {
 			So(len(query.Result), ShouldEqual, 3)
 		})
 
+		Convey("Given we have organizations, we can limit and paginate search", func() {
+			var err error
```
aknuds1, line 39: I would just remove this variable and define the error variable within the individual test case functions, since we're not going to use the variable outside of them. I see it's copied from the old test, but I think it could be better.

pkg/services/sqlstore/org_test.go
```diff
@@ -35,6 +35,44 @@ func TestAccountDataAccess(t *testing.T) {
...
+			var cmd *models.CreateOrgCommand
```
aknuds1, line 40: Same goes for this variable, please just define it (with `:=`) on line 44.

pkg/services/sqlstore/org_test.go
```diff
@@ -35,6 +35,44 @@ func TestAccountDataAccess(t *testing.T) {
...
+			ids := []int64{}
```
aknuds1, line 41: What's the point of this variable, since you don't actually use it for anything?

pkg/services/sqlstore/org_test.go
```diff
@@ -35,6 +35,44 @@ func TestAccountDataAccess(t *testing.T) {
...
+			Convey("Should be able to limit search", func() {
+				query := &models.SearchOrgsQuery{Limit: 1}
+				err = SearchOrgs(query)
+
+				So(err, ShouldBeNil)
+				So(len(query.Result), ShouldEqual, 1)
+			})
+
+			Convey("Should be able to paginate search", func() {
+				query := &models.SearchOrgsQuery{Page: 0}
```
aknuds1, line 60: Is there any point to this test case, since it's just using the default page and limit parameters?

aknuds1 (changes requested):
I think the test code can be cleaned up a bit, if you see my comments.

pkg/services/sqlstore/org_test.go
benjaminjb: Thanks for the review! With this test, I was looking to test the defaults (and really should've dropped the `Page: 0` part). Should I remove it all together? 

Or should I make >1000 orgs to test that the default is still 1000 (just to make sure that previous behavior is still maintained)? That test seems a little extreme, but it does test for what I want to verify.

aknuds1 (approved):
This looks better to me, thanks! I'm going to confer with colleagues about this feature.

papagian (approved):

The difference with the `/users` route is that in that case we [subtract one](https://github.com/grafana/grafana/blob/826184c9480b8e5e1faf71af353bffc7b4b433f9/pkg/services/sqlstore/user.go#L478) from the `page`
while in the `/orgs` route [we don't](https://github.com/grafana/grafana/blob/826184c9480b8e5e1faf71af353bffc7b4b433f9/pkg/services/sqlstore/org.go#L40):

PR #26935, opened 2020-08-11
Docs: Use relref links for data sources in whats new docs

zoltanbedi:

Some what's new docs in our docs site had wrong links for data sources. This pr fixes those.


oddlittlebird (approved):
LGTM. Nice catch!

PR #26936, opened 2020-08-11
Improve validation for grafanaDependency

marcusolsson:

Improves the validation of `grafanaDependency` to more closely follow [node-semver](https://github.com/npm/node-semver).

Test corpus (all listed are accepted):

```
=1.0.0
1.0.0
~1.0.0
^1.0.0
1.x.0
0.0.x
1.*.0
>=1.0.0
>=1.0.0 <2.0.0
```

Test here: https://regex101.com/r/4G0OUp/2

Not that it's regexp, and not magic. It won't actually validate that ranges are correct. It will still accept impossible ranges like `<=6.0.0 >7.0.0`.

### DIFF
--- a/docs/sources/developers/plugins/plugin.schema.json
+++ b/docs/sources/developers/plugins/plugin.schema.json
@@ -131,8 +131,8 @@
         },
         "grafanaDependency": {
           "type": "string",
-          "description": "Required Grafana version for this plugin, e.g. `>=7.0.0` to denote plugin requires Grafana 7.0.0 or later.",
-          "pattern": "^(>=|<|>|<=)?([0-9]+)(\\.[0-9]+)?(\\.[0-9])?$"
+          "description": "Required Grafana version for this plugin. Validated using https://github.com/npm/node-semver.",
+          "pattern": "^(<=|>=|<|>|=|~|\\^)?([0-9]+)(\\.[0-9x\\*]+)(\\.[0-9x\\*])?(\\s(<=|>=|<|=>)?([0-9]+)(\\.[0-9x]+)(\\.[0-9x]))?$"
         },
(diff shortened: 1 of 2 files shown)


docs/sources/developers/plugins/plugin.schema.json
```diff
@@ -127,8 +127,8 @@
         },
         "grafanaDependency": {
           "type": "string",
-          "description": "Required Grafana version for this plugin, e.g. `>=7.0.0` to denote plugin requires Grafana 7.0.0 or later",
-          "pattern": "^(>=|<|>|<=)?([0-9]+)(\\.[0-9]+)?(\\.[0-9])?$"
+          "description": "Required Grafana version for this plugin. Validated using https://github.com/npm/node-semver.",
+          "pattern": "^(<=|>=|<|>|=|~|\\^)?([0-9]+)(\\.[0-9x\\*]+)(\\.[0-9x\\*])?(\\s(<=|>=|<|=>)?([0-9]+)(\\.[0-9x]+)(\\.[0-9x]))?$"
```
papagian, line 131: This regular expression does not accept ranges like: `1.2.7 || >=1.2.9 <2.0.0` but I guess it is highly improbable to have gaps between the supported versions.

oddlittlebird (approved):
LGTM!

docs/sources/developers/plugins/plugin.schema.json
marcusolsson: Maybe we can expand the expression if it turns out to be too restrictive?

papagian: It depends if we want to support such cases, otherwise it's fine.

PR #26937, opened 2020-08-11
Prometheus: add $__rate_interval variable

### LINKED ISSUE #21417: Prometheus data source: Add `$__rate_interval`

A variable `$__xinterval` (“extended interval”) similar to the already existing `$__interval`. It's value is the value of `$__interval` plus a configurable duration, which should default to the _Scrape interval_ configured in the Prometheus data source __not multiplied by anything like the resolution factor!!!__

To configure the additional duration explicitly, we need a new per-query input field in the query editor. It's tempting to call it _Scrape interval_, too, because that's what it will usually be (see below), but perhaps it's better to be more explicit about what it's doing and call it _Interval extension_ or something. I'll leave that to the UX experts.

**Why is this needed**:

Wider context: [this RP blog post](https://www.robustperception.io/what-range-should-i-use-with-rate).

Key paragraph: “Another consideration is that if you're using query_range, such as in graphing, then the range should be at least the size of the step. Otherwise you'll skip over some data. Grafana's `$__interval` can be useful here.”

To actually not skip over data, the range needs to be slightly larger than `$__interval`, ideally to create an overlap of one sample. (This is not true if there are already recording rules in the game, and then `avg_over_time` is used to extend the result to longer time spans, which is also described in the [RP blog post](https://www.robustperception.io/what-range-should-i-use-with-rate). In this case, the current behavior works just fine.)

This “not skipping over data at all” was such a deep desire for some, that they proposed to change how PromQL handles range selectors specifically for `rate` expressions. What came out of that is a fork with something called `xrate`. This approach has issues of very different kinds. It would be great if we could solve it in Grafana, which then separates concerns quite nicely.

That's where this suggestion is coming from: Add a new variable just for the Prometheus data source `$__xinterval` which is (usually) one scrape interval longer than `$__interval`. Since the scrape interval for a given time series cannot be inquired by Grafana, we need to make it configurable per query (with the default being the scrape interval configured in the data source). This also opens the door for some other, more fringe use cases, where people wanted to use range selectors with a somewhat modified `$__xinterval`. (We are only offering adding a duration here, but should legitimate use cases come up where a factor is required, we could introduce that without a breaking change.)

Note: Other solutions to this problem were discussed, but Grafana has an obligation to not break existing dashboards. That's why the introduction of a separate variable was inevitable.

zoltanbedi:

Fixes #21417

## Todo

- [x] Add documentation possible to Prometheus data source


public/app/plugins/datasource/prometheus/datasource.test.ts
```diff
@@ -1601,9 +1601,24 @@ describe('PrometheusDatasource', () => {
...
       });
     });
   });
+
+  describe('The __rate_interval variable', () => {
+    it('should be 4 times the scrape interval if interval + scrape interval is lower', () => {
+      const { __rate_interval } = ds.getRateIntervalScopedVariable(15, 15);
+      expect(__rate_interval.value).toBe('60s');
+    });
+    it('should fall back to 15s if interval is 0', () => {
```
beorn7, line 1617: s/15s/60s/

public/app/plugins/datasource/prometheus/datasource.test.ts
```diff
@@ -1601,9 +1601,24 @@ describe('PrometheusDatasource', () => {
...
+          text: '75s',
+          value: '75s',
+        },
       });
     });
   });
+
+  describe('The __rate_interval variable', () => {
+    it('should be 4 times the scrape interval if interval + scrape interval is lower', () => {
+      const { __rate_interval } = ds.getRateIntervalScopedVariable(15, 15);
```
beorn7, line 1614: Perhaps use a "weirder" value here than 15s? With 15s, the expected result is 60s, which happens to be the same as if 0s is used.

beorn7 (commented):
Looks good so far. I have a gut feeling we should test more edge cases, but I'm not familiar enough with the typescript here to come up with concrete code suggestions. Perhaps I should simply play with a Grafana instance built from this code to see if I can find unintended behavior, which we can then fix and guard by a test case.

public/app/plugins/datasource/prometheus/datasource.test.ts
zoltanbedi: Done

oddlittlebird:
Doc looks fine.

public/app/plugins/datasource/prometheus/datasource.test.ts
zoltanbedi: I'm not sure what you mean by that.

zoltanbedi:
> Looks good so far. I have a gut feeling we should test more edge cases, but I'm not familiar enough with the typescript here to come up with concrete code suggestions. Perhaps I should simply play with a Grafana instance built from this code to see if I can find unintended behavior, which we can then fix and guard by a test case.

Sounds good to me.

public/app/plugins/datasource/prometheus/datasource.test.ts
beorn7: Perhaps I misread the test, but it appears the whole sentence is "The __rate_interval variable should fall back to 15s if interval is 0" but the __rate_interval is in fact 60s, not 15s, see line 1619.

zoltanbedi: I see, you are right.

beorn7 (approved):
OK, let's try it out. (Am I right to assume that this code will be used in the ops-tools1 Grafana instance once it is merged?)

torkelo (approved):
Looks good! Not using any new UI editor option (which was my main concern when this was discussed earlier this year)

davkal (approved):
Works great.

beorn7:
Now that I can play with it in practice, I found a few edge cases. I'll follow up in #21417 to discuss.

PR #26945, opened 2020-08-12
grafana-cli: Fix installing of plugins missing directory entries in zip

adrianlzt:
If a plugin zip file does not have the base directory placed in the first position, this command will fail trying to create a file in
a directory that does not exists. Same problem if the zip file does not contain the parent directory of files/directories.

To create an unordered zip file (directory not the first one):

```
zip -r file.zip foo/bar foo/foobar foo/
```

Zip file without the base dir:

```
zip -r file.zip foo/bar foo/foobar
```

Fixes #

### DIFF
--- a/pkg/cmd/grafana-cli/commands/install_command.go
+++ b/pkg/cmd/grafana-cli/commands/install_command.go
@@ -233,6 +233,12 @@ func extractFiles(archiveFile string, pluginName string, filePath string, allowS
 				return fmt.Errorf(permissionsDeniedMessage, newFile)
 			}
 		} else {
+			// Create needed directories to extract file
+			err := os.MkdirAll(filepath.Dir(newFile), 0755)
+			if err != nil {
+				return errutil.Wrap("failed to create directory to extract plugin files", err)
+			}
+
 			if isSymlink(zf) {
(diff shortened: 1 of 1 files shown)


pkg/cmd/grafana-cli/commands/install_command.go
```diff
@@ -233,6 +233,9 @@ func extractFiles(archiveFile string, pluginName string, filePath string, allowS
 				return fmt.Errorf(permissionsDeniedMessage, newFile)
 			}
 		} else {
+			// Create needed directories to extract file
+			os.MkdirAll(filepath.Dir(newFile), 0755)
```
aknuds1, line 237: The error should be checked, linting fails at the moment because it isn't.

aknuds1 (changes requested):
I found a problem, but will look more closely.

pkg/cmd/grafana-cli/commands/install_command.go
adrianlzt: Fixed, thanks!

aknuds1 (approved):
LGTM!

aknuds1:

PR #26947, opened 2020-08-12
BackendSrv: Queues data source requests but passes through api requests

### LINKED ISSUE #25713: Long queue of requests when data loading time is longer than auto-refresh time

### LINKED ISSUE #25859: Dashboard: Data source requests in-flight putting save dashoard request and other requests in pending state
**What happened**:
Saving a dashboard and save dashboard modal hangs for a while. Not feedback is provided to the user what's going on. This also sometimes leads to user being logged out.

Save dashboard request should happen instantly.and not interfere with data sources queries taking long time to complete.

1. Save a copy of the gdev dashboard: Panel tests - Slow Queries & Annotations
2. Open saved copy
3. While panels are loading, save the dashboard (opens save modal) and click Save button.
4. Save modal hangs and as can be seen in animation the save dashboard request are sent directly but put in pending state waiting for other requests to complete:
save_dashboard_slow

Regarding that users can get logged out in this scenario I think that could happen if user auth token are rotated in data source query after save dashboard request have been put in pending state.

There are other things happening if for example going into dashboard settings when multiple lengthy dashboard queries are in progress.

Folder picker is spinning until queued get folders request is finished:
image

Permissions view show dashboard doesn't have any permissions until queued get permission request is finished:
image

**Environment**:
- Grafana version: master (10715a1c79) and 7.0.4
- User OS & Browser: Chrome v83

hugohaggmark:
This PR introduces a queue for all fetch calls in backendSrv. The `FetchQueueWorker` handles all the logic and makes sure that Grafana can't have over 5 parallel data source related http requests at any point. 

> The limit of 5 parallel data source requests is for a Grafana instance configured with HTTP or HTTPS as described in [https://grafana.com/docs/grafana/latest/administration/configuration/#protocol](https://grafana.com/docs/grafana/latest/administration/configuration/#protocol) for a Grafana instance configured with HTTP2 the limit is symbolically set to 1000.

`FetchQueueWorker` also makes sure that any calls to Grafana's internal api have precedence and will always be executed before any data source related calls.

Fixes #25859
Fixes #25713

This PR probably needs a lot of testing because it affects all data fetching. I added this to the 7.2 milestone but that might be optimistic :)


public/app/core/services/FetchQueue.ts
```diff
@@ -0,0 +1,92 @@
+import { Observable, Subject } from 'rxjs';
+
+import { BackendSrvRequest } from '@grafana/runtime';
+
+export interface QueueState extends Record<string, { state: FetchStatus; options: BackendSrvRequest }> {}
+
+export enum FetchStatus {
+  NotStarted,
+  Started,
+  Ended,
```
dprokop, line 10: Pending, InProgress, Done maybe? :)

dprokop (commented):
The fetch/response queues look reasonable. Left a few questions that bind my mind a little bit:)

Great work!

public/app/core/services/FetchQueue.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export interface FetchQueueUpdate {
+  noOfStarted: number;
+  noOfNotStarted: number;
```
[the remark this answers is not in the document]
hugohaggmark: Yes, we could leave it up to the consumer to count this every time we push an update on the stream. I had it like that first, but my thought was it was cleaner for the queue to publish this. I could easily change it back if you think it's important.

public/app/core/services/backend_srv.ts
```diff
@@ -53,6 +64,25 @@ export class BackendSrv implements BackendService {
...
+          subscriptions.add(result.observable.subscribe(observer));
+        })
+      );
+
+      this.fetchQueue.add(id, options);
+
+      return function unsubscribe() {
+        subscriptions.unsubscribe();
+      };
+    });
```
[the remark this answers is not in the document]
hugohaggmark: 1. The subscriptions is a collection of subscriptions that when unsubscribed will call underlying subscriptions `unsubscribe` function. We want to make sure we unsubscribe (complete) existing subscriptions when the returned Observable itself is unsubscribed (completed) for whatever reason. Does this makes sense? I could have created two variables like the snippet below but then I would have to introduce a `let` and that's not fun:
```ts
let innerSubscription: Subscription;
const outerSubscription = this.responseQueue.getResponses<T>(id).subscribe(result => {
  innerSubscription = result.observable.subscribe(observer);
});

this.fetchQueue.add(id, options);

return function unsubscribe() {
  innerSubscription.unsubscribe();
  outerSubscription.unsubscribe();
};
```

2. Yes, the magic of RxJs :) The "inner subscription" (result.observable.subscribe(observer)) will push any updates to the `observer` and the `observer` is then subscribed too from the outside. This is hard to explain, does that makes sense to you?

public/app/core/services/FetchQueue.ts
hugohaggmark: Good suggestions, will fix that ⭐  🙏

torkelo:
Sorry for slow review from me, this week has been a bit busy with re-org,

hugohaggmark:
> Sorry for slow review from me, this week has been a bit busy with re-org,

No worries, this one will need some testing and lots of reviews :)

public/app/core/services/FetchQueue.ts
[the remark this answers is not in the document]
dprokop: Nope, it isn't, after reading the code all got clear :)

marefr:
Just my 2 cents. Using HTTP2 will normally allow more concurrent requests. Whenever you have a lot of queries it would be good if it used the maximum concurrency offered by the browser, no? I think the number of concurrent requests should/need to be configurable.

hugohaggmark:
> Just my 2 cents. Using HTTP2 will normally allow more concurrent requests. Whenever you have a lot of queries it would be good if it used the maximum concurrency offered by the browser, no? I think the number of concurrent requests should/need to be configurable.
> 
> Maybe possible to "guess"/estimate the maximum amount of concurrent data source requests possible?

Good suggestion, I'll look into how we can accomplish that!

hugohaggmark:
> Just my 2 cents. Using HTTP2 will normally allow more concurrent requests. Whenever you have a lot of queries it would be good if it used the maximum concurrency offered by the browser, no? I think the number of concurrent requests should/need to be configurable.
> 
> Maybe possible to "guess"/estimate the maximum amount of concurrent data source requests possible?

@marefr not sure what you mean by configurable, can you elaborate? Do we need a new section in the configuration file? 

Ideally we could automatically figure out if Grafana is running HTTP/2 and if Grafana is using HTTP/1 then we could figure out which browser the user is using and change the concurrent requests accordingly. Not sure what the configuration would be.

hugohaggmark:
@marefr added `http2Enabled` to BootData and check for that in code. Skipped specific browser version detection for now, but we could implement that moving forward, WDYT?

marefr:
That could work, but HTTP2 can be enabled in a reverse proxy where Grafana still uses HTTP where this approach fails unfortunately. This is tricky. If I understood correct HTTP2 uses one connection per domain and multiplexes that for virtually unlimited amount of concurrent requests. HTTP1/1 6-8 concurrent requests per domain. 

Would all of this be simplified if all non-data source requests always cancel all pending data source requests?

hugohaggmark:
> That could work, but HTTP2 can be enabled in a reverse proxy where Grafana still uses HTTP where this approach fails unfortunately. This is tricky. If I understood correct HTTP2 uses one connection per domain and multiplexes that for virtually unlimited amount of concurrent requests. HTTP1/1 6-8 concurrent requests per domain.
> 
> Would all of this be simplified if all non-data source requests always cancel all pending data source requests?

I see, not sure if we should optimize for a special case like the one you mentioned. Would you say the reverse proxy case is a common case in the community? 

Also, I think cancelling all data requests when using an api request doesn't seem like a very nice experience for our users. But I can be convinced otherwise :)

pkg/api/frontendsettings.go
```diff
@@ -226,6 +226,7 @@ func (hs *HTTPServer) getFrontendSettingsMap(c *models.ReqContext) (map[string]i
 		},
 		"featureToggles":    hs.Cfg.FeatureToggles,
 		"rendererAvailable": hs.RenderService.IsAvailable(),
+		"http2Enabled":      setting.Protocol == "h2",
```
aknuds1, line 229: Could you merge with master and use `hs.Cfg.Protocol == "h2"` here instead? We are trying to stop using `setting.*` global variables, and I just merged a PR to add `hs.Cfg.Protocol`.

aknuds1 (approved):
LGTM, but please see my comment.

pkg/api/frontendsettings.go
hugohaggmark: Of course :)

public/app/core/services/FetchQueue.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const entriesWithoutOptions = Object.keys(update.state).reduce((all, key) => {
+      const entry = { id: key, state: update.state[key].state };
+      all.push(entry);
+      return all;
+    }, [] as Array<{ id: string; state: FetchStatus }>);
+
+    console.log('FetchQueue noOfStarted', update.noOfInProgress);
+    console.log('FetchQueue noOfNotStarted', update.noOfPending);
+    console.log('FetchQueue state', entriesWithoutOptions);
```
torkelo, line 92: Nitpick: Is there a reason these are instance variable functions instead of prototype functions?  When I see arrow functions like this I sort of expect that these functions are used from outside somehow and need the instance arrow binding.

public/app/core/services/ResponseQueue.ts
```diff
@@ -0,0 +1,50 @@
...
+        ),
+      });
+    });
+  }
+
+  add = (id: string, options: BackendSrvRequest): void => {
+    this.queue.next({ id, options });
+  };
+
+  getResponses = <T>(id: string): Observable<FetchResponsesEntry<T>> =>
```
torkelo, line 48: Nitpick: Same as above, any reason for arrow functions here?

public/app/core/services/backend_srv.ts
```diff
@@ -53,6 +64,38 @@ export class BackendSrv implements BackendService {
...
+      const subscriptions: Subscription = new Subscription();
+
+      // We're using the subscriptions.add function to add the subscription implicitly returned by this.responseQueue.getResponses<T>(id).subscribe below.
+      subscriptions.add(
+        this.responseQueue.getResponses<T>(id).subscribe(result => {
+          // The one liner below can seem magical if you're not accustomed to RxJs.
+          // Firstly, we're subscribing to the result from the result.observable and we're passing in the outer observer object.
+          // By passing the outer observer object then any updates on result.observable are passed through to any subscriber of the fetch<T> function.
+          // Secondly, we're adding the subscription implicitly returned by result.observable.subscribe(observer).
+          subscriptions.add(result.observable.subscribe(observer));
```
torkelo, line 83: black magic! good that you added the comment here :) 
Can't say I understand this 100% yet :)

torkelo (approved):
Awesome work!  

Tested this with this dashboard http://localhost:3000/d/xtY_uCAiz/panel-tests-slow-queries-and-annotations?orgId=1  (gdev dashboard), 

and it has a huge impact, now you can search and open another dashboard!

torkelo:
> Would all of this be simplified if all non-data source requests always cancel all pending data source requests?

@marefr this would be sort of bad, opening search or saving a dashboard would cancel all running queries

public/app/core/services/FetchQueue.ts
hugohaggmark: When using arrow functions in a class you never need to bind `this` or worry about `this`. But I can change it to the normal function declaration instead, to keep everything is consistent.

hugohaggmark: https://www.typescriptlang.org/play?q=414#example/this

hugohaggmark: But these function are also `private` so I can see your point. I'll change

hugohaggmark:
> LGTM, but please see my comment.

Sorry to bother you @aknuds1 but your suggested change breaks the logic for HTTP2 this is what I get when I write the two settings to the log:

```log
WARN[08-20|10:38:20] hs.Cfg.Protocol:http setting.Protocol:h2 
```

Any idea what's going on here?

torkelo:
this line needs to be moved down, https://github.com/grafana/grafana/blob/master/pkg/setting/setting.go#L676 

the PR that added this looks incomplete, it did not remove the global setting (it can be removed only used in rendering service and that has access to Cfg struct) 
https://github.com/grafana/grafana/pull/27071/files

hugohaggmark:
I've changed `setting.go` to fix the issue I mentioned earlier

public/app/core/services/FetchQueue.ts
hugohaggmark: I thought this was in backendSrv, I think I'm going to keep it like this for now.

public/app/core/services/backend_srv.ts
hugohaggmark: Thank you for reviewing this, I tried to comment as much as possible as I know how hard it is to reason about RxJs when you're not an expert on RxJs :)

public/app/core/services/backend_srv.ts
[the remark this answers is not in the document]
dprokop: > This is hard to explain, does that makes sense to you?

I think it does:) Thank you for the explanation!

dprokop:
>>Would all of this be simplified if all non-data source requests always cancel all pending data source requests?

>@marefr this would be sort of bad, opening search or saving a dashboard would cancel all running queries

Agree with this, it would be really strange is all queries are suddenly cancelled

marefr:
Great work. I still think that limit to 5 concurrent requests if you have HTTP2 enabled in a reverse proxy will lead to a bad experience and something to consider moving forward. This will affect Grafana Cloud for example (I think).

torkelo:
@marefr do you have any ideas for how to else fix https://github.com/grafana/grafana/issues/25859 ? 

We could add a backend option to disable the limit

marefr:
@torkelo yeah a backend setting would do

PR #26952, opened 2020-08-12
Dashboard: Adds folder name and link to the dashboard overview on the homepage

### LINKED ISSUE #26869: Folder names are not shown in the "Recently viewed Dashboards" view at grafana home
**Why is this component needed**:
We have several dashboards with the same name for different environments. They are separated in different "Folders". in the _Recently viewed dashboards_ view, I can see only the name of the dashboards, but since they all have the same name across different folders, it would be nice to have the folder name displayed somewhere, so we can know from what environment is a specific folder, without the need to go to the /dashboards endpoint.
___
 - [ ] Is/could it be used in more than one place in Grafana?

**Where is/could it be used?**:
home page
___
- [ ] Post screenshots possible.
- [ ] It has a single use case.
- [ ] It is/could be used in multiple places.

**Implementation** (Checklist meant for the person implementing the component)

- [ ] Component has a story in Storybook.
- [ ] Props and naming follows [our style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/frontend.md).
- [ ] It is extendable (rest props are spread, styles with className work, and so on).
- [ ] Uses [theme for spacing, colors, and so on](https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md).
- [ ] Works with both light and dark theme.

**Documentation**

- [ ] Properties are documented.
- [ ] Use cases are described.
- [ ] Code examples for the different use cases.
- [ ] Dos and don'ts.
- [ ] Styling guidelines, specific color usage (if applicable).

michelengelen:
Adds the `folderTitle` of the Dashboard to the Dashboard overview on the Grafana home.

fixes #26869

### DIFF
--- a/public/app/plugins/panel/dashlist/module.html
+++ b/public/app/plugins/panel/dashlist/module.html
@@ -9,7 +9,7 @@
           <span class="dashlist-title">
             {{dash.title}}
           </span>
-          <span class="dashlist-folder">
+          <span ng-if="dash.folderTitle" class="dashlist-folder">
             ({{dash.folderTitle}})
           </span>
           <span class="dashlist-star" ng-click="ctrl.starDashboard(dash, $event)">
--- a/public/app/plugins/panel/dashlist/module.html
+++ b/public/app/plugins/panel/dashlist/module.html
@@ -9,6 +9,9 @@
           <span class="dashlist-title">
             {{dash.title}}
           </span>
+          <span class="dashlist-folder">
+            ({{dash.folderTitle}})
+          </span>
           <span class="dashlist-star" ng-click="ctrl.starDashboard(dash, $event)">
             <icon name="dash.isStarred ? 'favorite':'star'" type="dash.isStarred ? 'mono':'default'"></icon>
           </span>
--- a/public/app/plugins/panel/dashlist/module.html
+++ b/public/app/plugins/panel/dashlist/module.html
@@ -10,7 +10,8 @@
             {{dash.title}}
           </span>
           <span ng-if="dash.folderTitle" class="dashlist-folder">
-            ({{dash.folderTitle}})
+            <icon name="'folder'" type="mono" size="'xs'"></icon>
+            {{dash.folderTitle}}
           </span>
           <span class="dashlist-star" ng-click="ctrl.starDashboard(dash, $event)">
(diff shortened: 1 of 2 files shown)


public/app/plugins/panel/dashlist/module.html
```diff
@@ -9,6 +9,9 @@ <h6 class="dashlist-section-header" ng-show="ctrl.panel.headings">
           <span class="dashlist-title">
             {{dash.title}}
           </span>
+          <span class="dashlist-folder">
+            ({{dash.folderTitle}})
+          </span>
```
hugohaggmark, line 14: Maybe we should ignore empty folder names as it looks kind of weird:
image

hugohaggmark (commented):
Thank you for contributing @michelengelen but I left a small comment for you to consider. Also what do you think about this feature @torkelo ?

public/app/plugins/panel/dashlist/module.html
michelengelen: Hi @hugohaggmark 

I have added a `ng-if` directive to the span. 
Now it does not display the empty folder anymore, when no `folderTitle` property is present.

It should be fine now.

WDYT?

dprokop:
@hugohaggmark @torkelo I think (feature-wise) this is OK. I understand the need here in a case where there are dashboards with the same names across different folders. One thing to consider is to consult with @jessover9000 if she doesn't have other ideas how to solve this issue.

torkelo:
Think this is fine, but long term we should use the same card as search results I think (but need to rewrite this panel to react first)

jessover9000:
I think it's a good idea to show the folder names! However, from a UX perspective putting the folder name in brackets doesn't seem like an established pattern to indicate a file location. If I looked at that view without explanation, I would probably wonder what that text in the brackets means – I'm not sure whether I could recognize it's a folder name, especially if I don't know what folders there are in the system. So maybe instead of showing the brackets, you could just add a little folder icon like I quickly mocked up in the attached screenshot?

image

michelengelen:
The solution with the folder icon is way better ... I agree! I will update the PR in the evening (I am on vacation atm 🌴)

@torkelo if you need help migrating the panel to react I would be glad to help in this regard as well. Just add me to the issue when it is ready!

PR #26954, opened 2020-08-12
grafana-cli: Fix file path processing, returning of errors

aknuds1:
Fix grafana-cli so it uses `filepath.Join` for filesystem path joining, so it works on Windows. Also fix some returning of errors.

### DIFF
--- a/pkg/cmd/grafana-cli/commands/install_command.go
+++ b/pkg/cmd/grafana-cli/commands/install_command.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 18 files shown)


papagian (approved):
LGTM :heart:

I wonder how we haven't encountered any issues on Windows earlier. I guess not a lot of users use customised plugins in Windows?

pkg/cmd/grafana-cli/commands/install_command.go
```diff
@@ -225,27 +224,26 @@ func extractFiles(archiveFile string, pluginName string, filePath string, allowS
...
-				err = extractSymlink(zf, newFile)
-				if err != nil {
+				if err := extractSymlink(zf, newFile); err != nil {
 					logger.Errorf("Failed to extract symlink: %v \n", err)
 					continue
 				}
 			} else {
-				err = extractFile(zf, newFile)
-				if err != nil {
+				if err := extractFile(zf, newFile); err != nil {
```
papagian, line 246: lint prefers `else if` here

aknuds1: @papagian Ah argh.

aknuds1:
> I wonder how we haven't encountered any issues on Windows earlier. I guess not a lot of users use customised plugins in Windows?

@papagian Wondering the same...

pkg/cmd/grafana-cli/commands/install_command.go
aknuds1: Will fix.

hugohaggmark:
> > I wonder how we haven't encountered any issues on Windows earlier. I guess not a lot of users use customised plugins in Windows?
> 
> @papagian Wondering the same...

What was the issue on Windows?

papagian:
> What was the issue on Windows?

Assuming `filePath` is `C:\grafana\data\plugins\plugin` and `newFile` is `something` then `path.Join(filePath, newFileName)` becomes `C:\grafana\data\plugins\plugin/something` which is not valid path for Windows. Check [here](https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go#answer-39182128) for more details.

aknuds1:
@hugohaggmark @papagian is correct. We can't use `path.Join` for filesystem paths, since it assumes the path separator is '/'. The `path` package is meant for URI paths, not filesystem ones.

hugohaggmark:
Thank you @papagian @aknuds1 🙏

PR #26957, opened 2020-08-12
Chore: Fix tests for Go 1.15

aknuds1:
Some tests won't compile with Go 1.15, due to programming mistakes (naive conversion of ints to strings). Here are the fixes.

### DIFF
--- a/pkg/components/gtime/gtime_test.go
+++ b/pkg/components/gtime/gtime_test.go
@@ -23,7 +23,7 @@ func TestParseInterval(t *testing.T) {
 		{interval: "1M", duration: now.Sub(now.AddDate(0, -1, 0))},
 		{interval: "1y", duration: now.Sub(now.AddDate(-1, 0, 0))},
 		{interval: "5y", duration: now.Sub(now.AddDate(-5, 0, 0))},
-		{interval: "invalid-duration", err: regexp.MustCompile(`^time: invalid duration \"invalid-duration\"$`)},
+		{interval: "invalid-duration", err: regexp.MustCompile(`^time: invalid duration "?invalid-duration"?$`)},
 	}
 
 	for i, tc := range tcs {
--- a/pkg/components/gtime/gtime_test.go
+++ b/pkg/components/gtime/gtime_test.go
@@ -2,6 +2,7 @@ package gtime
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


aknuds1:
Ugh, looks like there's some behaviour wrt. quoting of "invalid duration" in the error strings that differs between Go 1.14 and 1.15 :/

pkg/components/gtime/gtime_test.go
```diff
@@ -14,25 +15,26 @@ func TestParseInterval(t *testing.T) {
...
+		err      *regexp.Regexp
 	}{
 		{interval: "1d", duration: now.Sub(now.AddDate(0, 0, -1))},
 		{interval: "1w", duration: now.Sub(now.AddDate(0, 0, -7))},
 		{interval: "2w", duration: now.Sub(now.AddDate(0, 0, -14))},
 		{interval: "1M", duration: now.Sub(now.AddDate(0, -1, 0))},
 		{interval: "1y", duration: now.Sub(now.AddDate(-1, 0, 0))},
 		{interval: "5y", duration: now.Sub(now.AddDate(-5, 0, 0))},
-		{interval: "invalid-duration", err: "time: invalid duration invalid-duration"},
+		{interval: "invalid-duration", err: regexp.MustCompile(`^time: invalid duration \"invalid-duration\"$`)},
```
papagian, line 26: I would expect this to fail given [the error starts with `ParseInterval` and the input interval is not quoted](https://github.com/grafana/grafana/blob/b2a08c2a53f8ce038a819662d08549c6738e9082/pkg/components/gtime/gtime.go#L35).

aknuds1: @papagian I don't know why Go 1.15 is different, but it has quotes. I'm in the middle of preparing a fix that uses a regexp.

aknuds1: Should be fixed now.

kylebrandt: I noticed this in the go 1.15 release notes. It is actually the very last thing :-P https://golang.org/doc/go1.15#time 

> When returning an error, ParseDuration now quotes the original value.

aknuds1: Thanks @kylebrandt for clearing that up :+1: I've fixed it using regexps.

papagian: So in-fact this error is returned [here](https://github.com/grafana/grafana/blob/b2a08c2a53f8ce038a819662d08549c6738e9082/pkg/components/gtime/gtime.go#L17).

aknuds1:
All the tests are fixed now.

kylebrandt (approved):
<3

PR #26958, opened 2020-08-12
Users: Improve conflict error handling

### LINKED ISSUE #26934: Adding a new user with an already-existing name returns 500 instead of 409
**What happened**:
I sent an api request to create a new user whose name already exists. Grafana replied with a 500 error code.

Grafana send a 409 HTTP code, like it does for new teams for example.

Create a user. Try creating a user with the same name again.

There may be other api endpoints with the same issue. Those endpoints should send a meaningful http status code instead of a 500.

**Environment**:
- Grafana version: 7.1.3

joanlopez:

Currently, when an admin tries to create a new user with an already existing email / login, it returns a `500 Internal Server Error` with the ORM error text. So, the idea of this PR is to handle these situations properly and return a meaningful error.

Fixes #26934

- Even though the issue's author requested to return a `409 Conflict`, I preferred to return the `412 Precondition Failed` because we're already returning that HTTP status code for very similar cases (i.e. trying to invite a user with an already existing email / login), so IMHO I'd try to keep the coherency among all of these cases.
- Both cases (already existing email / login) are handled as the same error because IMHO it does not worth it to have them divided (I prefer to keep it as simple as possible).

### DIFF
--- a/pkg/api/admin_users.go
+++ b/pkg/api/admin_users.go
@@ -33,7 +33,7 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 
 	if err := bus.Dispatch(&cmd); err != nil {
 		if errors.Is(err, models.ErrOrgNotFound) {
-			return Error(400, models.ErrOrgNotFound.Error(), nil)
+			return Error(400, err.Error(), nil)
 		}
 
 		if errors.Is(err, models.ErrUserAlreadyExists) {
--- a/pkg/api/admin_users.go
+++ b/pkg/api/admin_users.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/grafana/grafana/pkg/api/dtos"
@@ -31,11 +32,11 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 	}
 
 	if err := bus.Dispatch(&cmd); err != nil {
-		if err == models.ErrOrgNotFound {
+		if errors.Is(err, models.ErrOrgNotFound) {
 			return Error(400, models.ErrOrgNotFound.Error(), nil)
 		}
 
-		if err == models.ErrUserAlreadyExists {
+		if errors.Is(err, models.ErrUserAlreadyExists) {
 			return Error(412, fmt.Sprintf("User with email '%s' or username '%s' already exists", form.Email, form.Login), err)
 		}
 
--- a/pkg/api/admin_users.go
+++ b/pkg/api/admin_users.go
@@ -36,7 +36,7 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 		}
 
 		if err == models.ErrUserAlreadyExists {
-			return Error(412, fmt.Sprintf("User with email %s and/or login %s already exists", form.Email, form.Login), err)
+			return Error(412, fmt.Sprintf("User with email '%s' or username '%s' already exists", form.Email, form.Login), err)
 		}
 
 		return Error(500, "failed to create user", err)
--- a/pkg/api/admin_users.go
+++ b/pkg/api/admin_users.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"fmt"
+
 	"github.com/grafana/grafana/pkg/api/dtos"
 	"github.com/grafana/grafana/pkg/bus"
 	"github.com/grafana/grafana/pkg/infra/metrics"
@@ -31,6 +33,10 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 	if err := bus.Dispatch(&cmd); err != nil {
 		if err == models.ErrOrgNotFound {
 			return Error(400, models.ErrOrgNotFound.Error(), nil)
+		}
+
+		if err == models.ErrUserAlreadyExists {
+			return Error(412, fmt.Sprintf("User with email %s and/or login %s already exists", form.Email, form.Login), err)
 		}
 
 		return Error(500, "failed to create user", err)
--- a/pkg/api/admin_users_test.go
+++ b/pkg/api/admin_users_test.go
@@ -260,7 +260,7 @@ func TestAdminApiEndpoint(t *testing.T) {
 		})
 	})
 
-	Convey("When a server admin attempts to create a user with an already existing email / login", t, func() {
+	Convey("When a server admin attempts to create a user with an already existing email/login", t, func() {
 		bus.AddHandler("test", func(cmd *models.CreateUserCommand) error {
 			return models.ErrUserAlreadyExists
 		})
--- a/pkg/api/admin_users_test.go
+++ b/pkg/api/admin_users_test.go
@@ -259,6 +259,26 @@ func TestAdminApiEndpoint(t *testing.T) {
 			})
 		})
 	})
+
+	Convey("When a server admin attempts to create a user with an already existing email / login", t, func() {
+		bus.AddHandler("test", func(cmd *models.CreateUserCommand) error {
+			return models.ErrUserAlreadyExists
+		})
+
+		createCmd := dtos.AdminCreateUserForm{
+			Login:    TestLogin,
+			Password: TestPassword,
+		}
+
+		adminCreateUserScenario("Should return an error", "/api/admin/users", "/api/admin/users", createCmd, func(sc *scenarioContext) {
+			sc.fakeReqWithParams("POST", sc.url, map[string]string{}).exec()
+			So(sc.resp.Code, ShouldEqual, 412)
+
+			respJSON, err := simplejson.NewJson(sc.resp.Body.Bytes())
+			So(err, ShouldBeNil)
+			So(respJSON.Get("error").MustString(), ShouldEqual, "User already exists")
+		})
+	})
 }
 
 func putAdminScenario(desc string, url string, routePattern string, role models.RoleType, cmd dtos.AdminUpdateUserPermissionsForm, fn scenarioFunc) {
--- a/pkg/services/sqlstore/user_test.go
+++ b/pkg/services/sqlstore/user_test.go
@@ -541,28 +541,6 @@ func TestUserDataAccess(t *testing.T) {
 					}
 				})
 			})
-
-			Convey("When trying to create a new user with an existing email, an error is returned", func() {
-				createUserCmd := &models.CreateUserCommand{
-					Email:        "user2@test.com",
-					Name:         "user2",
-					Login:        "user2",
-					SkipOrgSetup: true,
-				}
-				err := CreateUser(context.Background(), createUserCmd)
-				So(err, ShouldEqual, models.ErrUserAlreadyExists)
-			})
-
-			Convey("When creating a new user with an already existing login returns error", func() {
-				createUserCmd := &models.CreateUserCommand{
-					Email:        "user-2@test.com",
-					Name:         "user2",
-					Login:        "loginuser2",
-					SkipOrgSetup: true,
-				}
-				err := CreateUser(context.Background(), createUserCmd)
-				So(err, ShouldEqual, models.ErrUserAlreadyExists)
-			})
 		})
 
 		Convey("Given one grafana admin user", func() {
@@ -588,6 +566,39 @@ func TestUserDataAccess(t *testing.T) {
 				So(getUserError, ShouldBeNil)
 
 				So(query.Result.IsAdmin, ShouldEqual, true)
+			})
+		})
+
+		Convey("Given one user", func() {
+			var err error
+			createUserCmd := &models.CreateUserCommand{
+				Email: fmt.Sprint("user", "@test.com"),
+				Name:  "user",
+				Login: "user",
+			}
+			err = CreateUser(context.Background(), createUserCmd)
+			So(err, ShouldBeNil)
+
+			Convey("When trying to create a new user with the same email, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        fmt.Sprint("user", "@test.com"),
+					Name:         "user2",
+					Login:        "user2",
+					SkipOrgSetup: true,
+				}
+				err := CreateUser(context.Background(), createUserCmd)
+				So(err, ShouldEqual, models.ErrUserAlreadyExists)
+			})
+
+			Convey("When trying to create a new user with the same login, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        fmt.Sprint("user2", "@test.com"),
+					Name:         "user2",
+					Login:        "user",
+					SkipOrgSetup: true,
+				}
+				err := CreateUser(context.Background(), createUserCmd)
+				So(err, ShouldEqual, models.ErrUserAlreadyExists)
 			})
 		})
 	})
--- a/pkg/services/sqlstore/user_test.go
+++ b/pkg/services/sqlstore/user_test.go
@@ -570,18 +570,17 @@ func TestUserDataAccess(t *testing.T) {
 		})
 
 		Convey("Given one user", func() {
-			var err error
 			createUserCmd := &models.CreateUserCommand{
-				Email: fmt.Sprint("user", "@test.com"),
+				Email: "user@test.com",
 				Name:  "user",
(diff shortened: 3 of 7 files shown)


pkg/api/admin_users.go
```diff
@@ -33,6 +35,10 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 			return Error(400, models.ErrOrgNotFound.Error(), nil)
 		}
 
+		if err == models.ErrUserAlreadyExists {
+			return Error(412, fmt.Sprintf("User with email %s and/or login %s already exists", form.Email, form.Login), err)
```
sakjur, line 39: We call the login param username in the form. I also changed `and/or` to just `or`, since there's nothing that implies that the or is exclusive and it's a slightly more natural sentence (:joy: not sure any sentence with an email in the middle is natural).

And consider adding quotes around the email and username if the user uses whitespace in them to make the error clearer (not sure we allow that, but in case)

```suggestion
			return Error(412, fmt.Sprintf("User with email '%s' or username '%s' already exists", form.Email, form.Login), err)
```

sakjur (commented):
:+1: This looks really good to me. I have some minor feedback on the user facing error message, but other than that this is a really good patch.

I like that this logic is kept to a few lines, like we discussed earlier the additional benefits of having a longer logic to be able to tell the admin that it was the email in particular or the username specifically that didn't match.

I also appreciate how well these tests integrates with the rest, I'm working on converting these tests to testify in another PR, but it's more work than I'd anticipated so it's taking some time.

You don't have to mention the bug you're fixing in the title, when we squash-merge this that'll be confusing as you'll have a commit message of `API: Improve error handling (#26934) (#26958)` 🙂

pkg/api/admin_users.go
joanlopez: Thanks for the suggestion! Indeed, your sentence really looks more natural 👌

sakjur (approved):
LGTM. Let's see if someone from the backend team has some feedback before merging 🙂

pkg/services/sqlstore/user_test.go
```diff
@@ -541,6 +541,28 @@ func TestUserDataAccess(t *testing.T) {
 					}
 				})
 			})
+
+			Convey("When creating a new user with an already existing email returns error", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        "user2@test.com",
```
aknuds1, line 547: Where is user2@test.com already inserted into the database here? Couldn't find it.

pkg/services/sqlstore/user_test.go
```diff
@@ -541,6 +541,28 @@ func TestUserDataAccess(t *testing.T) {
...
+					SkipOrgSetup: true,
+				}
+				err := CreateUser(context.Background(), createUserCmd)
+				So(err, ShouldEqual, models.ErrUserAlreadyExists)
+			})
+
+			Convey("When creating a new user with an already existing login returns error", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        "user-2@test.com",
+					Name:         "user2",
```
aknuds1, line 559: Where is user2 already inserted into the database here? Couldn't find it.

aknuds1 (changes requested):
Generally looks good, but left some suggestions for improvement.

aknuds1:
> LGTM. Let's see if someone from the backend team has some feedback before merging

Thank you @sakjur, just finished my review :)

pkg/services/sqlstore/user_test.go
```diff
@@ -541,6 +541,28 @@ func TestUserDataAccess(t *testing.T) {
 					}
 				})
 			})
+
+			Convey("When creating a new user with an already existing email returns error", func() {
```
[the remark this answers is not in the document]
joanlopez: Thanks for the suggestion, I already applied the suggested changes 👌

pkg/services/sqlstore/user_test.go
joanlopez: Sorry, on the previous version I included these tests under the `Given 5 users` group, which, in fact, is not really clear. That's the reason it wasn't clear there's a user with `Email` equals to `user", i, "@test.com` where `i` is `0...4` and the same for the `Login` attribute equals to `"loginuser", i`.

Now, I refactored the code to create an specific Convey section for these two tests.

pkg/services/sqlstore/user_test.go
joanlopez: Ditto 👉 https://github.com/grafana/grafana/pull/26958#discussion_r469511437

joanlopez:
> > LGTM. Let's see if someone from the backend team has some feedback before merging
> 
> Thank you @sakjur, just finished my review :)

Feedback applied! Can you @aknuds1  PTAL? Thanks! 🙏

pkg/services/sqlstore/user_test.go
```diff
@@ -590,6 +568,39 @@ func TestUserDataAccess(t *testing.T) {
...
+
+		Convey("Given one user", func() {
+			var err error
+			createUserCmd := &models.CreateUserCommand{
+				Email: fmt.Sprint("user", "@test.com"),
+				Name:  "user",
+				Login: "user",
+			}
+			err = CreateUser(context.Background(), createUserCmd)
+			So(err, ShouldBeNil)
```
aknuds1, line 580:
```suggestion
			createUserCmd := &models.CreateUserCommand{
				Email: "user@test.com",
				Name:  "user",
				Login: "user",
			}
			err := CreateUser(context.Background(), createUserCmd)
			So(err, ShouldBeNil)
```

pkg/services/sqlstore/user_test.go
```diff
@@ -590,6 +568,39 @@ func TestUserDataAccess(t *testing.T) {
...
+
+			Convey("When trying to create a new user with the same email, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        fmt.Sprint("user", "@test.com"),
+					Name:         "user2",
+					Login:        "user2",
+					SkipOrgSetup: true,
+				}
```
aknuds1, line 588:
```suggestion
				createUserCmd := &models.CreateUserCommand{
					Email:        "user@test.com",
					Name:         "user2",
					Login:        "user2",
					SkipOrgSetup: true,
				}
```

pkg/services/sqlstore/user_test.go
```diff
@@ -590,6 +568,39 @@ func TestUserDataAccess(t *testing.T) {
...
+				So(err, ShouldEqual, models.ErrUserAlreadyExists)
+			})
+
+			Convey("When trying to create a new user with the same login, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        fmt.Sprint("user2", "@test.com"),
+					Name:         "user2",
+					Login:        "user",
+					SkipOrgSetup: true,
+				}
```
aknuds1, line 599:
```suggestion
				createUserCmd := &models.CreateUserCommand{
					Email:        "user2@test.com",
					Name:         "user2",
					Login:        "user",
					SkipOrgSetup: true,
				}
```

aknuds1 (changes requested):

pkg/api/admin_users.go
```diff
@@ -29,10 +32,14 @@ func AdminCreateUser(c *models.ReqContext, form dtos.AdminCreateUserForm) Respon
 	}
 
 	if err := bus.Dispatch(&cmd); err != nil {
-		if err == models.ErrOrgNotFound {
+		if errors.Is(err, models.ErrOrgNotFound) {
 			return Error(400, models.ErrOrgNotFound.Error(), nil)
```
aknuds1, line 36: Do you mind updating this as well? I don't see any point in fixating the error message, when we can get it from `err`:
```suggestion
			return Error(400, err.Error(), nil)
```

pkg/api/admin_users_test.go
```diff
@@ -259,6 +259,26 @@ func TestAdminApiEndpoint(t *testing.T) {
 			})
 		})
 	})
+
+	Convey("When a server admin attempts to create a user with an already existing email / login", t, func() {
```
aknuds1, line 263:
```suggestion
	Convey("When a server admin attempts to create a user with an already existing email/login", t, func() {
```

pkg/services/sqlstore/user_test.go
```diff
@@ -568,6 +568,38 @@ func TestUserDataAccess(t *testing.T) {
 				So(query.Result.IsAdmin, ShouldEqual, true)
 			})
 		})
+
+		Convey("Given one user", func() {
+			createUserCmd := &models.CreateUserCommand{
+				Email: "user@test.com",
+				Name:  "user",
+				Login: "user",
+			}
```
aknuds1, line 577:
```suggestion
			const email = "user@test.com"
			const username = "user"
			createUserCmd := &models.CreateUserCommand{
				Email: email,
				Name:  "user",
				Login: username,
			}
```

pkg/services/sqlstore/user_test.go
```diff
@@ -568,6 +568,38 @@ func TestUserDataAccess(t *testing.T) {
...
+			err := CreateUser(context.Background(), createUserCmd)
+			So(err, ShouldBeNil)
+
+			Convey("When trying to create a new user with the same email, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        "user@test.com",
+					Name:         "user2",
+					Login:        "user2",
+					SkipOrgSetup: true,
+				}
```
aknuds1, line 587:
```suggestion
				createUserCmd := &models.CreateUserCommand{
					Email:        email,
					Name:         "user2",
					Login:        "user2",
					SkipOrgSetup: true,
				}
```

pkg/services/sqlstore/user_test.go
```diff
@@ -568,6 +568,38 @@ func TestUserDataAccess(t *testing.T) {
...
+				err := CreateUser(context.Background(), createUserCmd)
+				So(err, ShouldEqual, models.ErrUserAlreadyExists)
+			})
+
+			Convey("When trying to create a new user with the same login, an error is returned", func() {
+				createUserCmd := &models.CreateUserCommand{
+					Email:        "user2@test.com",
+					Name:         "user2",
+					Login:        "user",
```
aknuds1, line 596:
```suggestion
					Login:        username,
```

aknuds1 (changes requested):
Almost there, but I left suggestions for a couple of adjustments I'd like to see.

pkg/api/admin_users.go
joanlopez: Well, I'd say it depends. With the current implementation, it's true that both statements are equivalent. However, the usage of `errors.Is` is more recommended than the equality comparator precisely to get it working despite of eventual error wraps. So, with the same reasoning, I'm not sure if returning the message present on given error is what we really want.

aknuds1: Personally I think it makes sense to return any wrapping error message, since it should wrap the error for a reason. The wrapped error message should also include `models.ErrOrgNotFound.Error()`. @sakjur @papagian WDYT?

aknuds1 (approved):
Looks good now, thanks! Still wondering about my remaining comment though, if that's a change we ought to make.

joanlopez:
> Looks good now, thanks! Still wondering about my remaining comment though, if that's a change we ought to make.

Sure, let's wait for the comments. I'm completely open to apply what we decide on consensus, I raised the question in order to clarify it all of us together.

Thanks for you feedback & suggestions!! 😇 🙏

pkg/api/admin_users.go
sakjur: I'm pro changing to `err.Error()`.

The only reason I can see why we wouldn't want to use `err.Error()` is to prevent exposure of secrets or over-technical error messages, and since this is an operation only available to admins I don't think that should be a concern.

Alternatively info log `err.Error()` and then return `models.ErrOrgNotFound.Error()`, that way we're not throwing away possibly useful debug information (Does the org not exist? Did the database crash during querying? Timeout while looking?)

papagian: I agree with @aknuds1; I think we don't want to loose this information.

aknuds1: @sakjur I guess that is also my one uncertainty, divulging secrets. My current thinking is that an error wrapping ErrOrgNotFound shouldn't divulge secrets, since it's meant for translating into an API error. That's my thinking around API error design at least, to have certain error types which are meant to be passed on via the API. I'm open to opposing points of view, though. I'm not super used to Go wrapped errors yet, coming from an exception background.

joanlopez: Okay, I will change it to `err.Error()`.

Thanks for your point of view!

sakjur: @aknuds1 I agree! If we want to send sensitive/debugging information with the error we should probably extend the error type to have something else than `.Error()` to access that information.

aknuds1: That's a good point @sakjur! It's definitely worth thinking about.

papagian:
I think this should change too:
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/org_invite.go#L184

Also, this query has become redundant; it can be treated by the [subsequent](https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L88
) `CreateUserCommand` query
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L82

papagian:
I have also noticed some inconsistency with the return status codes in similar cases:
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L29
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L84

I'm a bit reluctant change them but I think that we can decide if we want actually `412` to be returned for the cases where currently `500` is returned.
@aknuds1 @sakjur any thoughts?

sakjur:
@papagian Let's keep the scope down by just changing 500->412 for now?

For the end-user registration forms I think ideally we wouldn't tell the user if a user with that email address is already signed up, but that's a wider discussion 🙂

aknuds1:
@papagian Well, [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) sounds like the wrong status code at least. [401 unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) I'm unsure about.

aknuds1:
But I guess I agree with @sakjur about keeping the scope down.

joanlopez:
Is there already any doc about HTTP status codes conventions?

IMHO, these incoherences (up to 4-5 different status codes for -almost- the same) are common when dealing with HTTP APIs, so I guess the best to deal with it is at least formalize some conventions. Otherwise, we are at the risk of having new incoherences on the future.

aknuds1:
@joanlopez I'm not aware of any doc, but it sounds like we might need to ensure some consistency. Good news is that we are working on improving the API specifications, including functional tests :)

joanlopez:
So, to avoid converting this PR in an endless task, what do you think about focusing on:

1. Adding the `models.ErrUserAlreadyExists` -> `412` error handling at:
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/org_invite.go#L184

2. Removing the pre-check query and handling errors properly (without modifying the current returned HTTP status codes) at:
https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L84

@aknuds1 @papagian @sakjur

papagian:
> So, to avoid converting this PR in an endless task, what do you think about focusing on:
> 
> 1. Adding the `models.ErrUserAlreadyExists` -> `412` error handling at:
>    https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/org_invite.go#L184
> 2. Removing the pre-check query and handling errors properly (without modifying the current returned HTTP status codes) at:
>    https://github.com/grafana/grafana/blob/3798ac903d04a903b1cf09656af56d24864bbe5f/pkg/api/signup.go#L84
> 
> @aknuds1 @papagian @sakjur

sounds. good to me!

sakjur (approved):

papagian (approved):
LGTM!

aknuds1 (approved):

PR #26964, opened 2020-08-12
docs: mention httpMode in the docs

martinrw:
Simply just including the httpMode option in the provisioning docs


oddlittlebird (approved):
Thanks for contributing to the docs!

PR #26969, opened 2020-08-12
OAuth: Handle DEFLATE compressed payloads in JWT for Generic OAuth

### LINKED ISSUE #26967: OAuth2: GenericOAuth should handle GZIP/DEFLATE compressed payloads in JWT
If the JWT header indicates zip=GZIP or zip=DEF, then the JWT payload needs to be uncompressed after base64 decode and before unmarshalling to the UserInfoJson struct.

OAuth2 token providers can choose to compress the payload portion of the JWT and this change would improve the generic OAuth2 implementation's compliance with the standard.

billoley:
Fixes #26967

If the JWT header indicates zip=DEF, then the JWT payload needs to be uncompressed after base64 decode and before unmarshalling to the UserInfoJson struct.

OAuth2 token providers can choose to compress the payload portion of the JWT and this change would improve the generic OAuth2 implementation's compliance with the standard.

Added cases to an existing unit test to verify that DEF code path works.

### DIFF
--- a/pkg/login/social/generic_oauth.go
+++ b/pkg/login/social/generic_oauth.go
@@ -1,13 +1,18 @@
 package social
 
 import (
+	"bytes"
+	"compress/flate"
+	"compress/gzip"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/mail"
 	"regexp"
+	"strings"
 
 	"github.com/grafana/grafana/pkg/util/errutil"
 
@@ -207,6 +212,42 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
 	if err != nil {
 		s.log.Error("Error base64 decoding id_token", "raw_payload", matched[2], "error", err)
 		return nil
+	}
+
+	var headerBytes []byte
+	headerBytes, err = base64.RawURLEncoding.DecodeString(matched[1])
+	if err != nil {
+		s.log.Error("Error base64 decoding header", "header", matched[1], "error", err)
+		return nil
+	}
+
+	var header map[string]string
+	err = json.Unmarshal(headerBytes, &header)
+	if compression, ok := header["zip"]; ok {
+		if strings.EqualFold(compression, "GZIP") {
+			gr, err := gzip.NewReader(bytes.NewBuffer(rawJSON))
+			if err != nil {
+				s.log.Error("Error creating decompressor", "decoded_payload", string(rawJSON), "error", err)
+				return nil
+			}
+			defer gr.Close()
+			rawJSON, err = ioutil.ReadAll(gr)
+			if err != nil {
+				s.log.Error("Error decompressing payload", "decoded_payload", string(rawJSON), "error", err)
+				return nil
+			}
+		} else if strings.EqualFold(compression, "DEF") {
+			fr := flate.NewReader(bytes.NewReader(rawJSON))
+			defer fr.Close()
+			rawJSON, err = ioutil.ReadAll(fr)
+			if err != nil {
+				s.log.Error("Error decompressing raw_payload", "decoded_payload", string(rawJSON), "error", err)
+				return nil
+			}
+		} else {
+			s.log.Error("Invalid compression type: "+compression, "decoded_payload", string(rawJSON), "error", err)
+			return nil
+		}
 	}
 
 	var data UserInfoJson
--- a/pkg/login/social/generic_oauth.go
+++ b/pkg/login/social/generic_oauth.go
@@ -218,7 +218,11 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
 	}
 
 	var header map[string]string
-	err = json.Unmarshal(headerBytes, &header)
+	if err := json.Unmarshal(headerBytes, &header); err != nil {
+		s.log.Error("Error deserializing header", "error", err)
+		return nil
+	}
+
 	if compression, ok := header["zip"]; ok {
 		if compression == "DEF" {
 			fr, err := zlib.NewReader(bytes.NewReader(rawJSON))
@@ -233,7 +237,7 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
 				return nil
 			}
 		} else {
-			s.log.Error("Unknown compression algorithm", "algorithm", compression)
+			s.log.Warn("Unknown compression algorithm", "algorithm", compression)
 			return nil
 		}
 	}
--- a/pkg/login/social/generic_oauth.go
+++ b/pkg/login/social/generic_oauth.go
@@ -2,8 +2,7 @@ package social
 
 import (
 	"bytes"
-	"compress/flate"
-	"compress/gzip"
+	"compress/zlib"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -12,7 +11,6 @@ import (
 	"net/http"
 	"net/mail"
 	"regexp"
-	"strings"
 
 	"github.com/grafana/grafana/pkg/models"
 	"github.com/grafana/grafana/pkg/util/errutil"
@@ -213,8 +211,7 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
 		return nil
 	}
 
-	var headerBytes []byte
-	headerBytes, err = base64.RawURLEncoding.DecodeString(matched[1])
+	headerBytes, err := base64.RawURLEncoding.DecodeString(matched[1])
 	if err != nil {
 		s.log.Error("Error base64 decoding header", "header", matched[1], "error", err)
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/login/social/generic_oauth.go
```diff
@@ -209,6 +213,43 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
...
+	headerBytes, err = base64.RawURLEncoding.DecodeString(matched[1])
+	if err != nil {
+		s.log.Error("Error base64 decoding header", "header", matched[1], "error", err)
+		return nil
+	}
+
+	var header map[string]string
+	err = json.Unmarshal(headerBytes, &header)
+	if compression, ok := header["zip"]; ok {
+		switch strings.ToLower(compression) {
```
sakjur, line 226:
```suggestion
		switch compression {
```

> The "zip" value is a case-sensitive string.

from https://tools.ietf.org/html/rfc7516#section-4.1.3

(and adjust the cases accordingly)

pkg/login/social/generic_oauth.go
```diff
@@ -209,6 +213,43 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
...
+			if err != nil {
+				s.log.Error("Error decompressing payload", "decoded_payload", string(rawJSON), "error", err)
+				return nil
+			}
+		case "def":
+			fr := flate.NewReader(bytes.NewReader(rawJSON))
+			defer fr.Close()
+			rawJSON, err = ioutil.ReadAll(fr)
+			if err != nil {
+				s.log.Error("Error decompressing raw_payload", "decoded_payload", string(rawJSON), "error", err)
```
sakjur, line 244: Why is this different from :235?

```suggestion
				s.log.Error("Error decompressing payload", "decoded_payload", string(rawJSON), "error", err)
```

pkg/login/social/generic_oauth.go
```diff
@@ -209,6 +213,43 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
...
+				return nil
+			}
+		default:
+			s.log.Error("Invalid compression type: "+compression, "decoded_payload", string(rawJSON), "error", err)
```
sakjur, line 248:
```suggestion
			s.log.Error("Unknown compression algorithm", "algorithm", compression, "decoded_payload", string(rawJSON), "error", err)
```

sakjur (commented):
Thank you for sending a PR to improve Grafana :pray: I've left a few comments and have some general thoughts that I'd like to bounce with you before approving this.

I left a comment on the feature request re: where the `GZIP` algorithm is defined. We don't need to strictly implement the standard (in fairness, we're not even really implementing JWT), but it'd be nice to see some kind of proof that we're not the first to add the `GZIP` decompressor. https://github.com/grafana/grafana/issues/26967#issuecomment-673327799

You're testing the happy path, would it be possible to test a couple of error cases as well. One case each for "garbage payload" and "unknown algorithm" would do fine.

NB: I'm not sure how I feel about adding more complexity to the JWT parsing _inside_ the OAuth package, it seems like it's its own domain, but I also don't expect you to do that refactoring for us.

pkg/login/social/generic_oauth.go
billoley: copy/paste error, fixing

pkg/login/social/generic_oauth.go
billoley: Thanks.  Good catch - fixing

pkg/login/social/generic_oauth_test.go
```diff
@@ -427,3 +427,68 @@ func TestUserInfoSearchesForLogin(t *testing.T) {
 		}
 	})
 }
+
+func TestPayloadCompression(t *testing.T) {
+	t.Run("Given a generic OAuth provider", func(t *testing.T) {
```
sakjur, line 432: Thank you for keeping the style of the surrounding tests. We're moving over to using testify from having used GoConvey and these tests are somewhere in between 😅

You can remove the outermost layer of subtests if you'd like, but there's no harm done in keeping it.

```suggestion
```

sakjur (approved):
This looks good to me :+1:

I'm not merging this quite yet in case someone else has feedback I've overlooked or if you want to make some last minute modifications.

pkg/login/social/generic_oauth.go
```diff
@@ -209,6 +211,33 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
...
+			if err != nil {
+				s.log.Error("Error decompressing payload", "error", err)
+				return nil
+			}
+		} else {
+			s.log.Error("Unknown compression algorithm", "algorithm", compression)
+			return nil
+		}
+	}
+
```
aknuds1, line 240: In Go it's idiomatic to return early, and I think we should warn instead of error when an unrecognized compression algorithm is encountered, since it's not really an error on our part:
```suggestion
	if compression, ok := header["zip"]; ok {
		if compression != "DEF" {
			s.log.Warn("Unknown compression algorithm", "algorithm", compression)
			return nil
		}
		
		fr, err := zlib.NewReader(bytes.NewReader(rawJSON))
		if err != nil {
			s.log.Error("Error creating zlib reader", "error", err)
			return nil
		}
		defer fr.Close()
		rawJSON, err = ioutil.ReadAll(fr)
		if err != nil {
			s.log.Error("Error decompressing payload", "error", err)
			return nil
		}
	}

```

pkg/login/social/generic_oauth_test.go
```diff
@@ -427,3 +427,68 @@ func TestUserInfoSearchesForLogin(t *testing.T) {
...
+					AccessToken:  "",
+					TokenType:    "",
+					RefreshToken: "",
+					Expiry:       time.Now(),
+				}
+
+				token := staticToken.WithExtra(test.OAuth2Extra)
+				userInfo := provider.extractFromToken(token)
+
+				if test.ExpectResult == true {
```
aknuds1, line 485: I think it's less error prone, and less tedious to just tell if a result is expected by checking if test.ExpectedEmail is non-empty:
```suggestion
				if test.ExpectedEmail != "" {
```

aknuds1 (changes requested):
I need to discuss with my colleagues if we should include this functionality or not, but I left some suggestions for improvement in the meantime.

pkg/login/social/generic_oauth_test.go
```diff
@@ -427,3 +427,68 @@ func TestUserInfoSearchesForLogin(t *testing.T) {
...
+			SocialBase: &SocialBase{
+				log: log.NewWithLevel("generic_oauth_test", log15.LvlDebug),
+			},
+			emailAttributePath: "email",
+		}
+
+		tests := []struct {
+			Name          string
+			OAuth2Extra   interface{}
+			ExpectResult  bool
```
aknuds1, line 443: I think this attribute is superfluous and it's enough to use the default value for `ExpectedEmail` to signal that there should be no result.

aknuds1:
Why is only the DEF algorithm handled and not also GZIP? The PR description indicates that GZIP is also handled, doesn't it?

sakjur:
@aknuds1 Yes, @billoley removed `GZIP` after I asked where it was defined as it isn't mentioned anywhere in the standard/by the appointed standardization org.

I'd say it's highly optional since it strictly speaking doesn't exist as part of JOSE, but also there are popular implementations which implements it so it wouldn't necessarily hurt to have it.

sakjur:
@aknuds1 See discussion on #26967 for ref

aknuds1:
> @aknuds1 Yes, @billoley removed `GZIP` after I asked where it was defined as it isn't mentioned anywhere in the standard/by the appointed standardization org.
> 
> I'd say it's highly optional since it strictly speaking doesn't exist as part of JOSE, but also there are popular implementations which implements it so it wouldn't necessarily hurt to have it.

Ah thanks for clarifying that @sakjur. I suggest we edit the PR description title, to reflect that GZIP is out.

aknuds1:
I edited the PR title/description to reflect that only DEF support is added.

pkg/login/social/generic_oauth_test.go
billoley: This test is calling provider.extractFromToken(token), which returns nil on error.... so I can't check the userInfo.Email value.  I tried switching the test to make an ExpectedEmail == "" imply that userInfo should be nil, but that didn't really seem clearer that the way it is.

pkg/login/social/generic_oauth.go
```diff
@@ -209,6 +211,37 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
...
+			defer fr.Close()
+			rawJSON, err = ioutil.ReadAll(fr)
+			if err != nil {
+				s.log.Error("Error decompressing payload", "error", err)
+				return nil
+			}
+		} else {
+			s.log.Warn("Unknown compression algorithm", "algorithm", compression)
+			return nil
+		}
```
aknuds1, line 242: I don't know why this suggestion of mine from before disappeared?
```suggestion
		if compression != "DEF" {
			s.log.Warn("Unknown compression algorithm", "algorithm", compression)
			return nil
		}
		
		fr, err := zlib.NewReader(bytes.NewReader(rawJSON))
		if err != nil {
			s.log.Error("Error creating zlib reader", "error", err)
			return nil
		}
		defer fr.Close()
		rawJSON, err = ioutil.ReadAll(fr)
		if err != nil {
			s.log.Error("Error decompressing payload", "error", err)
			return nil
		}
```

pkg/login/social/generic_oauth_test.go
```diff
@@ -427,3 +427,66 @@ func TestUserInfoSearchesForLogin(t *testing.T) {
...
+			log: log.NewWithLevel("generic_oauth_test", log15.LvlDebug),
+		},
+		emailAttributePath: "email",
+	}
+
+	tests := []struct {
+		Name          string
+		OAuth2Extra   interface{}
+		ExpectResult  bool
+		ExpectedEmail string
```
aknuds1, line 443: I still think it's better to remove `ExpectResult`, since `ExpectResult: true` and `ExpectedEmail: ""` would be an invalid combination, right? As such, you might as well avoid the tedium of the extra boolean attribute.

aknuds1 (changes requested):

pkg/login/social/generic_oauth_test.go
```diff
@@ -427,3 +427,66 @@ func TestUserInfoSearchesForLogin(t *testing.T) {
...
+				AccessToken:  "",
+				TokenType:    "",
+				RefreshToken: "",
+				Expiry:       time.Now(),
+			}
+
+			token := staticToken.WithExtra(test.OAuth2Extra)
+			userInfo := provider.extractFromToken(token)
+
+			if test.ExpectResult == true {
```
aknuds1, line 484:
```suggestion
			if test.ExpectedEmail != "" {
```

billoley:
> I need to discuss with my colleagues if we should include this functionality or not, but I left some suggestions for improvement in the meantime.

Background: I'm switching a frontend datasource to a backend datasource where the ultimate datastore uses authentication (previously PKI, now OAuth token).  Our JWT token uses compression because it contains serialized user roles/authorizations/info.  I am switching our OAuth code from using GZIP to using DEF.  This PR and another coming soon (passing the OAuth token to the backend plugin) are fairly minor changes, but necessary to our integration.

pkg/login/social/generic_oauth_test.go
billoley: I'm fine either way.  I just updated the test and pushed to the PR branch.

aknuds1 (approved):
Looks good AFAICT. Will just check with the team if there are any opposing views.

billoley:
Should this PR be rebased on branch v7.1.x instead of master.... which will be 7.2.0

marefr:
@billoley no, PR's are always targeting master. We use squash and merge when merging the PR so no need to rebase. If you want to sync with master, please do a merge. If not, the reviews can be really hard to follow.

PR #26970, opened 2020-08-12
Chore: Use theme-specific color gradient for active tab header

twei55:
This is a change for the Grafana UI.

## Change

The active tab header color is hardcoded in the Tab component. This PR makes the tab header color theme-specific by adding two additional variables to the theme and using their values in the Tab component.

### DIFF
--- a/packages/grafana-data/src/types/theme.ts
+++ b/packages/grafana-data/src/types/theme.ts
@@ -240,6 +240,10 @@ export interface GrafanaTheme extends GrafanaThemeCommons {
     formCheckboxBgChecked: string;
     formCheckboxBgCheckedHover: string;
     formCheckboxCheckmark: string;
+
+    // Active Tab Header Color
+    tabHeaderGradientStart: string;
+    tabHeaderGradientEnd: string;
   };
   shadows: {
     listItem: string;
--- a/packages/grafana-ui/src/components/Tabs/Tab.tsx
+++ b/packages/grafana-ui/src/components/Tabs/Tab.tsx
@@ -97,7 +97,7 @@ const getTabStyles = stylesFactory((theme: GrafanaTheme) => {
         right: 0;
         height: 2px;
         top: 0;
-        background-image: linear-gradient(to right, #f05a28 30%, #fbca0a 99%);
+        background-image: linear-gradient(to right, ${colors.tabHeaderGradientStart}, ${colors.tabHeaderGradientEnd});
       }
     `,
   };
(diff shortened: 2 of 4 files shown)


kaydelaney (approved):
LGTM

packages/grafana-data/src/types/theme.ts
```diff
@@ -240,6 +240,10 @@ export interface GrafanaTheme extends GrafanaThemeCommons {
     formCheckboxBgChecked: string;
     formCheckboxBgCheckedHover: string;
     formCheckboxCheckmark: string;
+
+    // Active Tab Header Color
+    tabHeaderGradientStart: string;
+    tabHeaderGradientEnd: string;
```
dprokop, line 246: Can you please remove those variables from the theme and just use correct theme colors directly in the Tab component?

Also pinging @jessover9000 - what would be the theme colors to use for dark theme's Tab instead of currently used  `#f05a28 30%, #fbca0a 99%`?

dprokop (changes requested):

packages/grafana-data/src/types/theme.ts
jessover9000: Thank you for asking! Since red and orange are hard to see on black for certain types of color blindness and consistent high contrast can't hurt, for the dark theme I would suggest choosing the lightest orange from the accessible color palette that @Ijin08 defined (#ffb375) in stead of the brand orange. In the light theme, we can keep using our currently used brand orange (#f05a28). The yellow can stay the same in both light and dark (#ffee52). I haven't really played around with the position of the colors in the gradient, I guess we can keep 30% and 99% for minimal change.

twei55: @dprokop @jessover9000 Thanks for your feedback. I am not yet completely familiar with the codebase but currently I can't find none of the mentioned colors in the theme palettes.

Colors used in dark theme:
```
yellow: '#ecbb13',
orange: '#eb7b18',
orangeDark: '#ff780a',
```

Colors used in light theme:
```
yellow: '#ff851b',
orange: '#ff7941',
orangeDark: '#ed5700',
```

I would not like to change these values because it might break other component styles that I am not aware of.

dprokop: @jessover9000 sorry for not being clear in my question. It was more about what colors from our palette should be used instead of the ones used currently? We probably should not introduce new colors but either use the ones we have or adjust them to the needs of the Tab component. I would opt for reusing existing colors TBH.

twei55: @dprokop @jessover9000 Could you clarify the colors?

jessover9000: ah, yes, sorry for the delay. this issue keeps recurring, we need to fix up the grafana-ui color palette so that we're all using the same colors. 🤪 In the meantime, I guess we could use a gradient from the values @twei55 mentioned for dark theme yellow and orange and exclude orange-dark, to give more color contrast.

PR #26972, opened 2020-08-12
Alerting: Increase PagerDuty dedup_key complexity

### LINKED ISSUE #26863: Alerting: PagerDuty de-duplication key too simple, may cause incorrect incident de-duplication
**What happened**:
I have two Grafana instances in different clusters, both Grafanas have the same PG account and integration key configured. Both instances have the same dashboards configured. (through import-export).

If alerts with the same alert ID but from different Grafana instances are triggered, only one incident shows up in Pager Duty. It is caused by de-duplication on the PG side (note that the same ID can have different alerts).

All alerts should be shown in PG. The dedup key should be unique or more complex.

The current dedup key is created by the following scheme: `bodyJSON.Set("dedup_key", "alertId-"+strconv.FormatInt(evalContext.Rule.ID, 10))`.

https://github.com/grafana/grafana/blob/9c81c7aa4462057abec2cc2426d575843cb6b2b4/pkg/services/alerting/notifiers/pagerduty.go#L223

We can probably create a more complex key, something like this: `alerId+dashboardId+panelId`.

1. Create fresh identical Grafana instances in two similar environments.
2. Configure identical PagerDuty notification channels in the Grafana instances. 
3. Create or import dashboards with alerts.
4. Trigger alerts with the same alert ID on the Grafana instances. Check the alert ID through the Alerting API.
5. Check Pager Duty incidents, only one should present. 

**Environment**:

- Grafana version: grafana:6.5.1
- Data source type & version: Prometheus 2.17.2
- OS Grafana is installed on: docker container,
- User OS & Browser: Ubuntu 20.4, firefox
- Grafana plugins: no
- Others: -

alavrovinfb:
- Updated dedup_key generation scheme

 - Updated UT's

Changes from this PR  should avoid incorrect PagerDuty incidents de-duplication. 

This has been raised to fix Grafana issue [#26863 ](https://github.com/grafana/grafana/issues/26863)

Fixes #26863 

Currently, alerts haven't got generated UID, for this reason, I've used the corresponding dashboard UID. Since dashboard UID could be overridden by custom UID, the mentioned issue could be faced.

### DIFF
--- a/pkg/services/alerting/eval_context.go
+++ b/pkg/services/alerting/eval_context.go
@@ -99,6 +99,10 @@ func (c *EvalContext) GetNotificationTitle() string {
 
 // GetDashboardUID returns the dashboard uid for the alert rule.
 func (c *EvalContext) GetDashboardUID() (*models.DashboardRef, error) {
+	if c.IsTestRun {
+		return &models.DashboardRef{Uid: "testUID"}, nil
+	}
+
 	if c.dashboardRef != nil {
 		return c.dashboardRef, nil
 	}
--- a/pkg/services/alerting/notifiers/pagerduty.go
+++ b/pkg/services/alerting/notifiers/pagerduty.go
@@ -220,7 +220,18 @@ func (pn *PagerdutyNotifier) buildEventPayload(evalContext *alerting.EvalContext
 	bodyJSON := simplejson.New()
 	bodyJSON.Set("routing_key", pn.Key)
 	bodyJSON.Set("event_action", eventType)
-	bodyJSON.Set("dedup_key", "alertId-"+strconv.FormatInt(evalContext.Rule.ID, 10))
+	dedupKey := "alertId-" + strconv.FormatInt(evalContext.Rule.ID, 10) +
+		"-" + strconv.FormatInt(evalContext.Rule.PanelID, 10)
+	dashboardRef, err := evalContext.GetDashboardUID()
+	if err != nil {
+		pn.log.Error("Failed get dashboard reference", "error", err)
+		return []byte{}, err
+	}
+	dedupKey += "-" + dashboardRef.Uid
+	if len(dedupKey) > 255 {
+		dedupKey = dedupKey[0:254]
+	}
+	bodyJSON.Set("dedup_key", dedupKey)
 	bodyJSON.Set("payload", payloadJSON)
 
 	ruleURL, err := evalContext.GetRuleURL()
--- a/pkg/services/alerting/notifiers/pagerduty.go
+++ b/pkg/services/alerting/notifiers/pagerduty.go
@@ -1,8 +1,8 @@
 package notifiers
 
 import (
+	"fmt"
 	"os"
-	"strconv"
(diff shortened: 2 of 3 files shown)


pkg/services/alerting/notifiers/pagerduty.go
```diff
@@ -220,7 +220,18 @@ func (pn *PagerdutyNotifier) buildEventPayload(evalContext *alerting.EvalContext
...
+		"-" + strconv.FormatInt(evalContext.Rule.PanelID, 10)
+	dashboardRef, err := evalContext.GetDashboardUID()
+	if err != nil {
+		pn.log.Error("Failed get dashboard reference", "error", err)
+		return []byte{}, err
+	}
+	dedupKey += "-" + dashboardRef.Uid
+	if len(dedupKey) > 255 {
+		dedupKey = dedupKey[0:254]
+	}
```
aknuds1, line 233:
```suggestion
	dashboardRef, err := evalContext.GetDashboardUID()
	if err != nil {
		pn.log.Error("Failed get dashboard reference", "error", err)
		return []byte{}, err
	}
	
	dedupKey := fmt.Sprintf("alertId-%d-%d-%d", evalContext.Rule.ID, evalContext.Rule.PanelID, dashboardRef.Uid)
	if len(dedupKey) > 255 {
		dedupKey = dedupKey[0:254]
	}
```

aknuds1 (changes requested):
Haven't reviewed in depth yet, but left a suggestion for cleaner code.

aknuds1 (commented):
Thanks for applying my suggestion, will review more.

aknuds1:
The build failure is because "fmt" has to be imported in pagerduty.go.

alavrovinfb:
> The build failure is because "fmt" has to be imported in pagerduty.go.

Missed import has been added.

pkg/services/alerting/eval_context.go
```diff
@@ -99,6 +99,10 @@ func (c *EvalContext) GetNotificationTitle() string {
 
 // GetDashboardUID returns the dashboard uid for the alert rule.
 func (c *EvalContext) GetDashboardUID() (*models.DashboardRef, error) {
+	if c.IsTestRun {
```
kylebrandt, line 102: Need to look closer but I think "IsTestRun" is not for go tests, but rather when someone is using the "Test rule".

So I think when we do this, there should be a dashboard UID.

marefr:
@alavrovinfb thank you for your contribution. However, reading the issue and looking at these changes feels to me as you would end up with the same problem given you use the same dashboards in both Grafana instances which yields the same UID and panel id. 

The issue haven't received any other feedback/upvote why I would like to wait a bit and see if others having the same problem. Until then I would suggest you to investigate whether using multiple integrations or something else in Pagerduty allows you to solve your problem. 

I'm closing this for now since we're trying to only have active PR's open , but we can reopen it later should there be more activity around it or if I misunderstood something. Suggests continuing discussion in the issue.

PR #26974, opened 2020-08-12
Dashboard: Panel menu no longer lingers

### LINKED ISSUE #26306: Panel Menu Dropdown Lingers
**What happened**:
When clicking the Panel Menu Toggle, the menu that comes up is not dismissed when clicking another panel menu edit button while the other dropdown is still open. 

These drop downs then linger when you go into editing one of the panels:

Azure Multi-Dimensions (New in 7 1) - Grafana (3)

**Additionally**, This can also be triggered by setting a refresh interval on the dashboard, opening a panel menu, waiting for a refresh, and then the panel menu also lingers (in a worse state that is not dismissable). 

**Environment**:
- Grafana version: 9b6b0cad21
- User OS & Browser: Chrome/Linux

kaydelaney:
Changes `ClickOutsideWrapper` so we listen on `document` rather than `window`, fixing an issue where menu panels would linger as detailed in #26306

Closes #26306

### DIFF
--- a/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
+++ b/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
@@ -25,17 +25,17 @@ export class ClickOutsideWrapper extends PureComponent<Props, State> {
   };
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


dprokop (approved):
LGTM!

PR #26978, opened 2020-08-13
Chore: Use `--no-cache-dir` flag to `pip` in Dockerfiles, to save space

rajpratik71:
Using "--no-cache-dir" flag in pip install ,make sure downloaded packages
by pip don't cached on system . This is a best practice which make sure
to fetch from repo instead of using local cached one . Further , in case
of Docker Containers , by restricting caching , we can reduce image size.
In term of stats , it depends upon the number of python packages
multiplied by their respective size . e.g for heavy packages with a lot
of dependencies it reduce a lot by don't caching pip packages.

Further , more detail information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/devenv/docker/blocks/collectd/Dockerfile
+++ b/devenv/docker/blocks/collectd/Dockerfile
@@ -10,7 +10,7 @@ ADD     etc_mtab /etc/mtab
 
 ADD     collectd.conf.tpl /etc/collectd/collectd.conf.tpl
 
-RUN	pip install envtpl
+RUN	pip install --no-cache-dir envtpl
 ADD     start_container /usr/bin/start_container
 RUN     chmod +x /usr/bin/start_container
 CMD     start_container
--- a/devenv/docker/blocks/graphite1/Dockerfile
+++ b/devenv/docker/blocks/graphite1/Dockerfile
@@ -35,7 +35,7 @@ RUN if [ ! -z "${CONTAINER_TIMEZONE}" ]; \
 
 # fix python dependencies (LTS Django and newer memcached/txAMQP)
 RUN pip install --upgrade pip && \
-  pip install django==1.8.18 \
+  pip install --no-cache-dir django==1.8.18 \
   python-memcached==1.53 \
   txAMQP==0.6.2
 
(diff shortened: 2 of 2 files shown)


docs/sources/tutorials/stack_guide_graphite.md
```diff
@@ -169,7 +169,7 @@ Graphite api is a light weight version of graphite-web with only the API compone
 to install.
 
 ```
-pip install gunicorn graphite-api
+pip install --no-cache-dir gunicorn graphite-api
```
sakjur, line 172:
```suggestion
pip install gunicorn graphite-api
```

This is docs meant for end users. I'd rather avoid adding extra arguments and increasing the complexity, especially since the cache is presumably there for a reason when using pip on a regular operating system and not within the context of a Docker container.

sakjur (changes requested):
Thank you for this PR.

This makes a lot of sense to me except for in documentation when we shouldn't increase the complexity of the instructions for the user. Can we remove that change?

scripts/build/ci-msi-build/msigenerator/Makefile
```diff
@@ -1,5 +1,5 @@
 all: build
 
 build:
-	pip3 install -r requirements.txt
+	pip3 install --no-cache-dir -r requirements.txt
```
aknuds1, line 4: Please remove this, eventual benefits aren't clear in this case.

aknuds1 (changes requested):

docs/sources/tutorials/stack_guide_graphite.md
aknuds1: Agree with @sakjur.

aknuds1:
This branch needs to be merged with master, otherwise CI won't work.

scripts/build/ci-msi-build/msigenerator/Makefile
Rajpratik71: done changes as requested

sakjur (approved):
:+1: LGTM!

aknuds1 (approved):

aknuds1:

PR #26981, opened 2020-08-13
Templating: Fixed recursive url update issue for textbox variable

### LINKED ISSUE #26980: Variables: Cannot update textbox variable from data link on same dashboard
* Create a textbox variable. 
* Create a table with a datalink that sets the textbox variable (on current dashboard)
* Leads to infinte recursion (bridge_srv issue) 
* Fixing infinite recursion issue does not fix textbox update issue, only query updates not the current value. So a problem in updating current value in textbox.

torkelo:
Fixes #26980 

* Another recursive bridgeSrv issue fixed 
* Adds a gdev test dashboard for textbox and datalink that updates it on the same dashboard
* Removes our Tooltip component from Table cell links as it already had a native tooltip (so showed double tooltips, think the native one is better/enough)


hugohaggmark (approved):
Looks great as usual bridgeSrv slayer :) ⚔️

PR #18112, opened 2019-07-15
Build: allow dynamically change docker image

markelog:
Right now, only for couple DBs.
With the 2.1, you can do `${TAG:-5.6}` without needing the `.env` file, but
it seems it would be better for back-compat to support `.env` right now.

Fixes #17891

PR #26983, opened 2020-08-13
InfluxDB: Upgrade InfluxDB in devenv

aknuds1:
Upgrade from InfluxDB 1.7 to 1.8 in devenv

I'm also doing some serious cleanup of the Flux code, mostly de-exporting symbols not used outside of the package (the less exports the better).


papagian:
For `mysql` and `postgres` we support [dynamically changing](https://github.com/grafana/grafana/pull/18112) the docker image. We could do the same for InfluxDB.

Also I [suspect](https://github.com/grafana/grafana/blob/master/devenv/datasources.yaml#L30) that we shouldn't have had `/api/v2` suffix even before.

aknuds1:
I can look into making the image configurable @papagian. About the suffix, it is actually necessary before the upgrade, I tested.

aknuds1:
@papagian Could you show me exactly how to make the InfluxDB tag configurable in devenv? I found that MySQL uses the variable mysql_version, but when I grep for it, I can only find where it's used.

papagian:
> @papagian Could you show me exactly how to make the InfluxDB tag configurable in devenv? I found that MySQL uses the variable mysql_version, but when I grep for it, I can only find where it's used.

You only need to create an `.env` file under influxdb block with an `influxdb_version` variable and change the image in the docker-compose file to use this variable. The you can use the command:
`make devenv sources=influxdb influxdb_version=1.8.1-alpine` because values in the shell take precedence over those specified in the .env file.

devenv/docker/blocks/influxdb/docker-compose.yaml
```diff
@@ -1,5 +1,5 @@
   influxdb:
-    image: influxdb:1.7.6
+    image: influxdb:1.8.1-alpine
```
ryantxu, line 2: Am i right that this is the only thing related to the issue title?  The rest is making things private

ryantxu (approved):
seems fine -- the exported stuff is like that because it came from a package where it is appropriate to be exported... now in core it should not be.

However I do think we should put all the `XxxToOptionalYyy` helpers in the SDK -- they are currently copied in most datasources

devenv/docker/blocks/influxdb/docker-compose.yaml
aknuds1: You're right @ryantxu. Originally, I was also upgrading the InfluxDB Go client, but I ended up doing that in another PR to fix a serious bug of leaking TCP connections.

While I was working on the code, I figured I could also tighten up the code.

Issue #26920, opened 2020-08-11
Table - make it possible to pass in a custom renderer 


leventebalogh:
### What would you like to be added?
Make it possible to use a custom Cell renderer for the `<Table />` component in `@grafana/ui`.

**Example:**
```javascript
const data = new MutableDataFrame({
    fields: [
      {
        name: 'Status',
        type: FieldType.string,
        values: [
            'active',
            'active',
            'unavailable',
        ],
        // An idea for how we could set it is to pass an optional "render" property to a field
        render: (status) => <StatusComponent status={status} /> 
      },
    ],
  });
```

---

### Why is this needed?
Currently if we would like to use the `<Table />` component from our UI library we can only choose from the following options to display a value for a given cell:
- `ColorText`
- `ColorBackground`
- `LcdGauge`
- `BasicGauge`
- `GradientGauge`
- `JSONView` 

**It can easily happen though that we would like to have a custom way for visualising values in a certain field**, for example displaying a colorised dot for certain status values.

peterholmberg:
@leventebalogh what's the latest on this? Are you working on it?

PR #26985, opened 2020-08-13
@grafana-ui/Table: Add ability to use a custom renderer for table cells

leventebalogh:
Adds the ability to specify custom Cell renderers for Table fields.

**Example.:**
```javascript
const CustomCell = ({ field, cell }: TableCellProps) => {
  const value = field.display ? formattedValueToString(field.display(cell.value)) : cell.value;

  return <div className="my-fancy-cell">{value}</div>;
};

// Defining the table data
const data = new MutableDataFrame({
    fields: [
     // ...
      {
        name: 'Progress',
        type: FieldType.number,
        values: [],
        config: {
          unit: 'percent',
          min: 0,
          max: 100,
          custom: {
            width: 150,
            render: CustomCell,
          },
        },
      },
    ],
  });
```

https://github.com/grafana/grafana/issues/26920

**Why?**
I am actually not sure what plans we had with this component and if this kind of direction is aligned with those plans, however I thought that if it is going to be the "generic" Table component that people could use in plugins as well, then we probably want to give the ability to use custom cell renderers. 
[More info in the issue &rarr;](https://github.com/grafana/grafana/issues/26920)

### DIFF
--- a/packages/grafana-ui/src/components/Table/types.ts
+++ b/packages/grafana-ui/src/components/Table/types.ts
@@ -1,3 +1,4 @@
+import React from 'react';
 import { CellProps } from 'react-table';
 import { Field } from '@grafana/data';
 import { TableStyles } from './styles';
@@ -8,6 +9,7 @@ export interface TableFieldOptions {
   align: FieldTextAlignment;
   displayMode: TableCellDisplayMode;
   hidden?: boolean;
+  render?: (props: TableCellProps) => React.ReactNode | null | number | string;
 }
 
 export enum TableCellDisplayMode {
--- a/packages/grafana-ui/src/components/Table/Table.story.tsx
+++ b/packages/grafana-ui/src/components/Table/Table.story.tsx
@@ -14,7 +14,9 @@ import {
   ThresholdsConfig,
   ThresholdsMode,
   FieldConfig,
+  formattedValueToString,
 } from '@grafana/data';
+import { TableCellProps } from './types';
 
 export default {
   title: 'Visualizations/Table',
@@ -148,6 +150,33 @@ export const ColoredCells = () => {
       custom: {
         width: 80,
         displayMode: 'color-background',
+      },
+      thresholds: defaultThresholds,
+    },
+  });
+
+  return (
+    <div className="panel-container" style={{ width: 'auto' }}>
+      <Table data={data} height={500} width={width} />
+    </div>
+  );
+};
+
+const CustomCellRenderer = ({ field, cell }: TableCellProps) => {
+  const value = field.display ? formattedValueToString(field.display(cell.value)) : cell.value;
+
+  return <div style={{ backgroundColor: 'red', color: 'white', fontWeight: 'bold' }}>{value}</div>;
+};
+
+export const CustomCells = () => {
+  const theme = useTheme();
+  const width = number('width', 750, {}, 'Props');
+  const data = buildData(theme, {
+    Progress: {
+      custom: {
+        width: 80,
+        // Use a custom component to render the value in the cell
+        render: CustomCellRenderer,
       },
       thresholds: defaultThresholds,
     },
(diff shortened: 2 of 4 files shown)


hugohaggmark (commented):
I think this looks great, not sure how well it fits with the "field override" architecture, so I added @dprokop as reviewer too.

leventebalogh:
Thanks 🙏 

Let's see what @dprokop things 👍

packages/grafana-ui/src/components/Table/types.ts
```diff
@@ -8,6 +9,7 @@ export interface TableFieldOptions {
   align: FieldTextAlignment;
   displayMode: TableCellDisplayMode;
   hidden?: boolean;
+  render?: (props: TableCellProps) => React.ReactNode | null | number | string;
```
dprokop, line 12: Don' think we need this as this type is not used anywhere in the end.

dprokop (commented):
@leventebalogh think this is great. Code wise it look 👍 . Is this driven by some requirement you need to address btw? 

My concern is a little bit about keeping the render in the field config. Since this won't be persisted it might be fine. But on the other hand, I wonder if we shouldn't have `renderCell` prop on the table component instead that would accept the field and render a custom component. This way we wouldn't force consumer to actually modify the incoming data frame to attach renderers to the fields. WDYT?

Will also ping @torkelo and @ryantxu as they might have had some thoughts about this.

torkelo (commented):
Not sure the render / component should be on the DataFrame, but more a property passed to Table, like a fieldComponents={}  that is a dictionary of the field name to cell component. 

I guess having a renderer defined on the FieldConfig would allow a data source to control the rendering 

thoughts @ryantxu & @dprokop  ?

leventebalogh:
That sounds like a good point, didn't think about that, thanks guys!

I think passing it in as an _optional_ prop to the table can work just as well, I will update it if you are all ok with doing it in the following way:

```javascript
const CustomFieldComponent = ({ field, cell }: TableCellProps) => {
  const value = field.display ? formattedValueToString(field.display(cell.value)) : cell.value;

  return <div style={{ backgroundColor: 'red', color: 'white', fontWeight: 'bold' }}>{value}</div>;
};

// ...

return (
    <Table
        data={data}
        height={500}
        width={width}
        fieldComponents={{
            Progress: CustomFieldComponent
        }}
    />
);

```

packages/grafana-ui/src/components/Table/types.ts
leventebalogh: Thanks mate, good catch 👍

ryantxu:
@leventebalogh what are you trying to solve?  If the goal is full explicit control of cell rendering, maybe something like this:
```
<Table
        data={data}
        height={500}
        width={width}
        fieldCellRenderrer={ (field:Field) => {
            return CustomFieldComponent
        }}
    />
```

If we want to use FieldConfig -- i think that should be adding additional cell types, and then looking them up by key.

mckn (commented):
Great work @leventebalogh!

I can see the need of having the possibility to extend the Table with this kind of functionality. Just to make sure that it is easy to extend for other scenarios. With that said I would lean towards the solution that @ryantxu suggested but with a small adjustment.

```ts
<Table
        data={data}
        height={500}
        width={width}
        getCellComponent={ (context: CellContext) => {
            const { field } = context;
            if(field.name === 'Progress' ) {
               return CustomFieldComponent
            }
        }}
    />
```

My thinking is that this gives us a simple way of extending the `CellContext` with more information as new scenarios for extending the cells occurs. For example: you would like to change all cells for a specific field. But next time someone wants to change it for every second cell in field X etc.

PhML:
Would this change allow to have row selection with checkboxes for example? Or action buttons based on the content of the row?

leventebalogh:
Thanks for the ideas @ryantxu and @mckn!
I guess those are both good ideas, will change the code based on them 👍 

> Would this change allow to have row selection with checkboxes for example? Or action buttons based on the content of the row?
@PhML Yes, it would let you render the cell values in a custom way, for example rendering action buttons or checkboxes in certain columns.

mudumat:
Any plans to make ist possible to make getCellComponent as a property overwritable or extendable? It would make  the use of the table component more powerful. I need a lot of custum cells in my panelplug in. now i have to rewrite a kind of table and loose a lot of the given functionality.

PR #26990, opened 2020-08-13
AzureMonitor: map more units

kylebrandt:
More Azure monitor units need to be mapped to their Grafana values. In particular BitsPerSecond, but should take care of everything listed in https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list#unit


aknuds1:
Is this still WIP?

aknuds1 (approved):
LGTM

PR #26993, opened 2020-08-13
Components: Add FieldArray component

rdubrock:
The `Form` component wraps `react-hook-form` and exposes the `useForm` hook functionality. We are currently not exporting the functionality in `useFieldArray`. This PR exposes that via a component with render props

This is my first PR to Grafana, apologies ahead of time if I missed the mark!

### DIFF
--- a/packages/grafana-ui/src/components/Forms/FieldArray.story.tsx
+++ b/packages/grafana-ui/src/components/Forms/FieldArray.story.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import { withCenteredStory } from '../../utils/storybook/withCenteredStory';
+import { withStoryContainer } from '../../utils/storybook/withStoryContainer';
+import { Form, Input, Button, HorizontalGroup } from '@grafana/ui';
+import { FieldArray } from './FieldArray';
+import mdx from './FieldArray.mdx';
+
+export default {
+  title: 'Forms/FieldArray',
+  component: FieldArray,
+  decorators: [withStoryContainer, withCenteredStory],
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const simple = () => {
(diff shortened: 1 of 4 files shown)


packages/grafana-ui/src/components/Forms/FieldArray.story.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const simple = () => {
+  const defaultValues = {
+    people: [{ firstName: 'Janis', lastName: 'Joplin' }],
+  };
+  return (
+    <Form onSubmit={values => console.log(values)} defaultValues={defaultValues}>
+      {({ control, register }) => (
+        <div>
+          <FieldArray control={control} name="people">
+            {({ fields, append }) => (
```
Clarity-89, line 28: Not 100% sure about the double nested `children` when using `FieldArray` component, maybe instead of making it a separate component we could pass the result of `useFieldArray` into the Form's children function: 
```
<Form onSubmit={values => console.log(values)} defaultValues={defaultValues}>
      {({ control, register, fieldArrayProps }) => (
```
The one downside of that is that `useFieldArray` would have to be called every time the form is used, even tho its result might not be necessary there. Plus the Form would need to take an extra `name` prop.
@dprokop @peterholmberg any ideas?

rdubrock: I veered away from that approach just because it would muddy the api of `Form` and limit the user to one `FieldArray` per form, but definitely open to it. Another option would be to just export the `useFieldArray` hook directly. That would require the user to wrap field arrays in a separate component, but it would get around the render prop pyramid

Clarity-89 (approved):
@rdubrock let's go with a separate `FieldArray` component. It doesn't change the existing Form api and I guess nested children functions is not such a big deal.

PR #26999, opened 2020-08-14
Server: Add health check route

aknuds1:
Add health check route to Grafana server, since we need it when writing API tests (this was derived from a [bigger PR](https://github.com/grafana/grafana/pull/26517)). @bergquist suggested we use HTTPServer.healthHandler.

I'm also including a bug fix (closing SqlStore sessions) and some minor cleanup from the bigger PR.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -435,5 +435,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/api/snapshots-delete/:deleteKey", reqSnapshotPublicModeOrSignedIn, Wrap(DeleteDashboardSnapshotByDeleteKey))
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
 
+	// Health check
+	r.Get("/api/health", hs.healthHandler)
+
 	r.Get("/*", reqSignedIn, hs.Index)
 }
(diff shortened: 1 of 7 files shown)


pkg/api/api.go
```diff
@@ -435,5 +435,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/api/snapshots-delete/:deleteKey", reqSnapshotPublicModeOrSignedIn, Wrap(DeleteDashboardSnapshotByDeleteKey))
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
 
+	// Health check
+	r.Get("/api/health", hs.healthHandler)
```
bergquist, line 439: If we add a route for this we can remove the handler as a middleware. https://github.com/grafana/grafana/blob/master/pkg/api/http_server.go#L327 Currently the middleware will catch the requests.

aknuds1: @bergquist You mean I can just remove https://github.com/grafana/grafana/blob/master/pkg/api/http_server.go#L327?

aknuds1: @bergquist Making the assumption that's what you mean, and pushed a commit to remove it :) Please let me know in case I misunderstood.

bergquist: :+1:

bergquist (approved):
LGTM :)

PR #27000, opened 2020-08-14
Data sources: Use SQL store directly, not via bus

aknuds1:
Another PR to enable [API tests](https://github.com/grafana/grafana/pull/26517), by making the data source cache use the SQL store directly (dependency injected), instead of via the bus. The PR doesn't remove getting of the store from the bus, but I'd like a discussion as to whether it's possible (although probably out of scope for this PR).

The PR also makes HTTPServer.RouteRegister _not_ dependency injected, since that's also required by the tests, and I don't see any reason why it should be (just complicates things AFAICT).

@bergquist has, as I understand him, confirmed we should drop this usage of the bus. Just need approval of the details :)


aknuds1:
@bergquist Would you mind having a look now? After a rethink, I went with dependency injecting the SQL store into `datasources.CacheServiceImpl`, so it can avoid the bus entirely when querying.

pkg/services/sqlstore/datasource.go
```diff
@@ -27,6 +27,23 @@ func init() {
 	bus.AddHandler("sql", GetDataSourceByName)
 }
 
+func (ss *SqlStore) GetDataSourceByID(id, orgID int64) (*models.DataSource, error) {
```
xlson, line 30: Shouldn't we also make it so that the query-based version of this method uses this method? Seems like a bad idea to have two separate implementations. From what I can tell, the difference is that one of them uses the global `x.Get` which I don't think is something we really want to do anyway.

pkg/api/http_server.go
```diff
@@ -54,7 +54,7 @@ type HTTPServer struct {
 	httpSrv     *http.Server
 	middlewares []macaron.Handler
 
-	RouteRegister        routing.RouteRegister            `inject:""`
+	RouteRegister        routing.RouteRegister
```
xlson, line 57: I don't think we can just stop injecting the RouteRegister without also fixing all the locations where it is modified. We use it in Enterprise to extend the API and it's injected in a few places in OSS too, I didn't look into what it was used for. But I presume having the http_server use a different instance than the rest of Grafana will cause problems.

xlson:
It's great that we take small steps towards not using the bus anymore.

pkg/services/sqlstore/datasource.go
aknuds1: @xlson Let me look into this.

pkg/api/http_server.go
aknuds1: @xlson Aha, thanks for making me aware of that.

pkg/services/sqlstore/datasource.go
aknuds1: @xlson Solved it by making a separate common function.

pkg/api/http_server.go
aknuds1: @xlson Reverted this change.

sakjur (approved):
Agree with @xlson, reducing our bus factor [pun intended] is a great thing 👍

Looks like you've addressed Leo's comments and otherwise LGTM.

pkg/services/datasources/cache.go
```diff
@@ -31,7 +31,8 @@ func (dc *CacheServiceImpl) Init() error {
 	return nil
 }
 
-func (dc *CacheServiceImpl) GetDatasource(datasourceID int64, user *models.SignedInUser, skipCache bool) (*models.DataSource, error) {
+func (dc *CacheServiceImpl) GetDatasource(datasourceID int64, user *models.SignedInUser, skipCache bool) (
+	*models.DataSource, error) {
```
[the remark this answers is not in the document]
aknuds1: @sakjur Oh yeah, the idea was to not make the line longer than 100 chars. Do you know if we have any rule about this? I think there's some linter can object to lines longer than a certain length, but don't recall exactly.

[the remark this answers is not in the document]
sakjur: :+1:

I don't think we have any strict rule about it. I remember seeing Go core devs discourage from having a particular line length limit enforced elsewhere (which is why it isn't in `go fmt`), I think with the rationale that it's better if that's something that's left to the judgment of the developer + reviewer since what is legible for a particular function might be different from function to function.

The most common pattern when this happens I've seen is to do this:

```suggestion
func (dc *CacheServiceImpl) GetDatasource(
	datasourceID int64,
	user *models.SignedInUser,
	skipCache bool,
) (*models.DataSource, error) {
```

aknuds1:
@sakjur I see what you mean, thanks for clarifying that. I was always wondering.

bergquist (approved):
Looks great! 

I think this is a good example of a small step towards avoid using the bus. Small iterative change that doesn't break anything but still improve things.

PR #27002, opened 2020-08-14
Templating: Fixes so texts show in picker not the values

### LINKED ISSUE #25711: Template variable from MySQL query shows value instead of key when selected from multi-value dropdown
**What happened**:
Template variable selection shows value instead of key for *multi-value* selections

Template variable to show key of selected entries

1. Create a [template variable using a MySQL source](https://grafana.com/docs/grafana/latest/features/datasources/mysql/#query-variable)
2. Make the template variable *multi-value*

- The `All` option does not need to be used
- Display is correct (key shown) when *single-value* mode is used for the template variable

**Environment**:
- Grafana version: v7.0.3
- Data source type & version: MySQL
- OS Grafana is installed on: GNU/Linux (Debian)
- User OS & Browser: Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0

### LINKED ISSUE #25959: Templating: Multi-value variable rendrered using __value instead of __text after selection changes
**What happened**:
On the variable __text is replaced by the __value.
tela

Keep the __text.

image
image
image

This only happens with the multi-value selected.

**Environment**:
- Grafana version: v7.0.5 (efbcbb838b)
- Data source type & version: PostgreSQL
- OS Grafana is installed on:  Debian GNU/Linux 9.12 (stretch)
- User OS & Browser: Windows Google Chrome
- Grafana plugins:
- Others:

hugohaggmark:
The templating system has gotten less complex when we migrated it to React but there is still at least one big complexity left that still causes us some issues. The complexity I'm talking about is how we for some VariableTypes have support for both `single value` and `multi value`. This causes a lot of if statement in our code that tries to handle this and `props` that can be both `string` and `string[]`.

This PR does not attempt to solve this complexity, but I think it pushes our code onto a better path. With this PR the `current` prop will always keep the same type as defined in `alignCurrentWithMulti` below: 
https://github.com/grafana/grafana/blob/1915d10980a1ac91fef6b3577432b47f7c744892/public/app/features/variables/shared/multiOptions.ts#L3 

This means that `OptionsPicker` is the only component/class/object/function that transforms the `current.text` to the correct value which seems a lot clearer to me.
https://github.com/grafana/grafana/blob/1915d10980a1ac91fef6b3577432b47f7c744892/public/app/features/variables/pickers/OptionsPicker/OptionsPicker.tsx#L107

I also did some refactoring of the tests because they weren't using "valid" data.

Fixes #25711
Fixes #25959

There are probably other cases that are messed up with this, but I couldn't find any :)

### DIFF
--- a/public/app/features/variables/pickers/OptionsPicker/actions.test.ts
+++ b/public/app/features/variables/pickers/OptionsPicker/actions.test.ts
@@ -4,12 +4,12 @@ import { TemplatingState } from '../../state/reducers';
 import { QueryVariableModel, VariableHide, VariableRefresh, VariableSort } from '../../types';
 import {
   hideOptions,
+  moveOptionsHighlight,
   showOptions,
   toggleOption,
   toggleTag,
   updateOptionsAndFilter,
   updateSearchQuery,
-  moveOptionsHighlight,
 } from './reducer';
 import {
   commitChangesToVariable,
@@ -45,7 +45,10 @@ describe('options picker actions', () => {
 
   describe('when navigateOptions is dispatched with navigation key cancel', () => {
     it('then correct actions are dispatched', async () => {
-      const variable = createVariable({ options: [createOption('A', 'A', true)] });
+      const variable = createMultiVariable({
+        options: [createOption('A', 'A', true)],
+        current: createOption(['A'], ['A'], true),
+      });
 
       const clearOthers = false;
       const key = NavigationKey.cancel;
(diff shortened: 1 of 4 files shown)


torkelo (approved):
Looks good, tested this with an already saved dashboard with many multi-variables and saved in specific multi selection state to make sure there was no problem loading an old dashboard

hugohaggmark:
> Looks good, tested this with an already saved dashboard with many multi-variables and saved in specific multi selection state to make sure there was no problem loading an old dashboard

Thank you @torkelo, before we merge I just want to discuss this change with @mckn because I might miss some of his insights on this.

mckn (approved):
This looks great! It becomes a lot simpler 🙌 

awesome

hugohaggmark:
> This looks great! It becomes a lot simpler 🙌
> 
> awesome

Thank you @mckn

PR #27008, opened 2020-08-14
Tests: Batch of GoConvey to Testify conversions

sakjur:

Follow-up to #26724, squashing some more trivial GoConvey tests.

Found some issues with some of the existing tests that I've tried to fix as well. I haven't removed any assertions, but I've made them more specific or made sure they're table-driven tests where the tests were already structured in such a way.

### DIFF
--- a/pkg/cmd/grafana-cli/commands/ls_command_test.go
+++ b/pkg/cmd/grafana-cli/commands/ls_command_test.go
@@ -4,67 +4,81 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/grafana/grafana/pkg/cmd/grafana-cli/commands/commandstest"
 	"github.com/grafana/grafana/pkg/cmd/grafana-cli/services"
-	. "github.com/smartystreets/goconvey/convey"
 )
 
 func TestMissingPath(t *testing.T) {
-	var org = validateLsCommand
+	tests := []struct {
+		description string
+		cliContext  map[string]string
+		ioHelper    *commandstest.FakeIoUtil
+		error       error
+	}{
+		{
+			description: "missing path flag",
+			cliContext:  make(map[string]string),
+			ioHelper:    &commandstest.FakeIoUtil{},
+			error:       errMissingPathFlag,
+		},
+		{
+			description: "not a directory",
+			cliContext:  map[string]string{"pluginsDir": "/var/lib/grafana/plugins/notadir.txt"},
+			ioHelper:    &commandstest.FakeIoUtil{FakeIsDirectory: false},
+			error:       errNotDirectory,
+		},
+	}
 
-	Convey("ls command", t, func() {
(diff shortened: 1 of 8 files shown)


pkg/cmd/grafana-cli/commands/ls_command_test.go
```diff
@@ -4,67 +4,81 @@ import (
...
+			c, err := commandstest.NewCliContext(tc.cliContext)
+			require.NoError(t, err)
 
-			Convey("should return error", func() {
-				err := cmd.lsCommand(c)
-				So(err, ShouldBeError, "missing path flag")
-			})
+			cmd := Command{}
+			err = cmd.lsCommand(c)
+			assert.Truef(t, errors.Is(err, tc.error), "expected '%s', got '%s'", tc.error.Error(), err.Error())
```
aknuds1, line 48: I wonder if it's better to be explicit in tests and not use `errors.Is`, but instead assert.EqualError? The test should have an exact expected behaviour after all. Writing such assertions is also much easier.

pkg/infra/log/file_test.go
```diff
@@ -17,30 +19,27 @@ func (w *FileLogWriter) WriteLine(line string) error {
...
+		require.NoError(t, err)
+		err = fileLogWrite.WriteLine("test3\n")
+		require.NoError(t, err)
+
+		assert.Equal(t, 3, fileLogWrite.maxlines_curlines)
 	})
+
+	fileLogWrite.Close()
+	err = os.Remove(fileLogWrite.Filename)
+	require.NoError(t, err)
```
aknuds1, line 44: Should these be put in a cleanup call, to be sure the file gets removed?

aknuds1 (approved):
LGTM, but see comments.

pkg/cmd/grafana-cli/commands/ls_command_test.go
sakjur: I can change that. I think I overapplied [Change-Detector Tests Considered Harmful](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) here

pkg/infra/log/file_test.go
sakjur: Nice catch, I didn't notice that `os.Remove` 🙈

pkg/cmd/grafana-cli/commands/ls_command_test.go
sakjur: I changed it to `assert.Equal` instead of `assert.EqualError`, since I have access to the precise version of the error and don't have to unpack `.Error()` for the compare.

aknuds1: Ah, I didn't know that pattern, but I also prefer that tests don't cause friction to refactoring. I just don't think it's the case here, but rather defining the output :)

aknuds1: @sakjur OK :+1: Thanks for taking care of things :)

PR #27012, opened 2020-08-14
InfluxDB: Upgrade Go client, use data source HTTP client

### LINKED ISSUE #26912: Flux datasource leaking TCP connections

aknuds1:
Upgrade InfluxDB Go client to v2 (have to use Git tag, since they haven't introduced a v2 sub-directory yet, so can't import v2.0.0) and make it use the data source HTTP client. The latter change fixes a TCP connection leakage issue, which was due to creating a new HTTP client for every query, where closed connections would linger for a while in the TIME_WAIT state (this is normal for closed connections). By using the data source HTTP client, we ensure connections are pooled/re-used.

I've tested the fix, and AFAICT no more than two connections to InfluxDB are created at a time.

Fixes #26912.


pkg/tsdb/influxdb/flux/executor.go
```diff
@@ -33,7 +33,6 @@ func executeQuery(ctx context.Context, query QueryModel, runner queryRunner, max
 		dr.Frames = append(dr.Frames, metaFrame)
 		return
 	}
-	defer tables.Close()
```
papagian, line 36: it doesn't hard to keep this, right?

aknuds1: @papagian Per se, but I found out in an issue thread in the Go client repo that it's actually called internally, so clients aren't supposed to call it after all. See this [comment](https://github.com/influxdata/influxdb-client-go/issues/183#issuecomment-674053677) and the [code](https://github.com/influxdata/influxdb-client-go/pull/184/files/7d3656ca553768bc7ee7d4a4b80993f4b6222271#diff-8b3ced92a99034e241e8d5c5818effa7R228-R239) in question.

go.sum
```diff
@@ -235,8 +238,8 @@ github.com/igm/sockjs-go/v3 v3.0.0 h1:4wLoB9WCnQ8RI87cmqUH778ACDFVmRpkKRCWBeuc+W
 github.com/igm/sockjs-go/v3 v3.0.0/go.mod h1:UqchsOjeagIBFHvd+RZpLaVRbCwGilEC08EDHsD1jYE=
 github.com/inconshreveable/log15 v0.0.0-20180818164646-67afb5ed74ec h1:CGkYB1Q7DSsH/ku+to+foV4agt2F2miquaLUgF6L178=
 github.com/inconshreveable/log15 v0.0.0-20180818164646-67afb5ed74ec/go.mod h1:cOaXtrgN4ScfRrD9Bre7U1thNq5RtJ8ZoP4iXVGRj6o=
-github.com/influxdata/influxdb-client-go v1.3.0 h1:R3hLAFtbXCCsPNjlRY8CD66B2ROYWT3miLYLuTIlbbw=
-github.com/influxdata/influxdb-client-go v1.3.0/go.mod h1:S+oZsPivqbcP1S9ur+T+QqXvrYS3NCZeMQtBoH4D1dw=
+github.com/influxdata/influxdb-client-go v0.0.0-20200814135139-d68836272a45 h1:UePjGzItJZeZ4a4375kxZ6pc7myUjuqlpegERMzVYmo=
```
wbrowne, line 241: >v0.0.0

paranoia

pkg/tsdb/influxdb/flux/executor.go
papagian: I see

papagian (approved):
LGTM

go.sum
wbrowne: Make sense but strange to see!

kylebrandt (approved):

PR #27022, opened 2020-08-16
Units: Binary-prefixed data rates

### LINKED ISSUE #14757: [Feature request] Axis unit - data rate (IEC)
When specifying the unit for the Y1/Y2 axis, we have these options for data amounts:
- data (IEC), 1K = 1024; 1M = 1024K, etc.
- data (Metric), 1K = 1000; 1M = 1000K, etc.

However, for data **rates**, we only have the option of "data rate" _(which appears to be 1K = 1000)_.

Change Request - rename "data rate" unit type to "data rate (Metric)".
Feature Request - add the option for IEC data rates.

Berbe:
# What this PR does / why we need it
Introduces binary-prefixed data rates

I took the opportunity to merge the 2 Data categories in a single one: prefixes & units allow them to set them apart.

I also took the opportunity to do a quick & dirty cleanup of the nomenclature being used:
- `decimalSI` & `binarySI` make no sense -> `SI` & `binary`. Could also used `IEC` instead of `binary`, but felt better
- `ps` units replaced with `/s` variants to homogeneise the way units look throughout the category

# Which issue(s) this PR fixes

Fixes #14757

# Special notes for your reviewer

Would be happy to rebase/split/merge commits as needed.
Would also be happy to submit different PR if you thing work shall be split further

If more tests are required, please give me detailed pointers about how to craft them & on what.


mckn (approved):
I think this looks great! What do you think @torkelo?

dprokop (approved):
@Berbe thanks for the contribution, this looks good!

tarko:
@Berbe I think the backwards compatibility was broken by this.

We have used the `bits/s (format: bps in JSON)` previously and this has always divided by 1000 but is now dividing by 1024 meaning all the graphs are now displaying incorrect values. We are graphing network utilization (hence the bps) and it needs to be divided by 1000 for correct results but this is not the case any more:

```
- { name: 'bits/sec', id: 'bps', fn: decimalSIPrefix('bps') },
+ { name: 'bits/sec(IEC)', id: 'bps', fn: binaryPrefix('b/s') },
```

We rollbacked from 7.2 for now as we need to update all (hundreds of) dashboards if this was intentional.

Berbe:
It is a problem of choosing the right IDs for each unit.

It might have been better to have kept the units which only existed in the decimal form before in that decimal form. Making them binary by default wreaked havoc.

I dropped #27838 to address the 2 broken defaults I identified.

tarko:
LGTM, thanks.

PR #27024, opened 2020-08-17
MixedDatasource: Shows retrieved data even if a data source fails

### LINKED ISSUE #26955: MixedDatasource: error on one data source prevents showing available data from other data sources
**What happened**:
I select multiple datasources from which one becomes non-responsive. I get a "Network Error: bad gateway (502)" error in the panel and no data is shown in the panel.

I expected the data from the available datasources to be shown, and the error to be more specific, e.g. mentioning the failing datasource.

* add multiple datasources, e.g. on different embedded devices
* add metric queries for all of these datasources in e.g. the graph panel
* disconnect/shutdown one of the (embedded) devices

I verified this behaviour on the flowcharting panel and the graph panel.

**Environment**:
- Grafana version: 7.0.3
- Data source type & version: influxdb bundled with grafana
- OS Grafana is installed on: armbian/debian buster
- User OS & Browser:
- Grafana plugins:
- Others:

hugohaggmark:
This PR makes sure that any successful data fetched by MixedDatasource is shown although other data sources are failing. The error will show the first error (enhanced with the data source name) but we could tweak this if we want to show all erroneous data sources in the message.

image

Fixes #26955

You can use this dashboard with **TestData** and **Postgres** as a starting point: https://gist.github.com/hugohaggmark/d6ba274ec21085fa34a358a6f8a2b4a6

### DIFF
--- a/public/app/plugins/datasource/mixed/MixedDataSource.test.ts
+++ b/public/app/plugins/datasource/mixed/MixedDataSource.test.ts
@@ -5,7 +5,6 @@ import { DataSourceInstanceSettings, LoadingState } from '@grafana/data';
 import { DatasourceSrvMock, MockDataSourceApi } from 'test/mocks/datasource_srv';
 import { getQueryOptions } from 'test/helpers/getQueryOptions';
 import { MixedDatasource } from './module';
-import { describe } from '../../../../test/lib/common';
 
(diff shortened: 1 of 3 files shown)


ryantxu (approved):
looks great!

hugohaggmark:
Thank you @ryantxu, anything you want to add @mckn?

PR #27029, opened 2020-08-17
plugin.json schema updates

yesoreyeram:
This PR covers three changes to the plugin.json schema.

* Allow **$schema** in plugin.json to improve the plugin.json authoring experience
* Allow **others"" as a fallback category for plugin category.
* Mandate minimum one keyword in the info section to improve the plugin discovery in plugin repository.

##  Allow **$schema** in plugin.json

Current `plugin.schema.json` doesn't have option to enter `$schema`. Allowing `$schema` will enhance IDEs to autocomplete and lint the `plugin.json` file. 

#### **With schema**: ( Autocomplete and Linting ) 

image
image
image
image

#### **Without schema**: ( No autocomplete, No Lint )

image
image

This was previously raised against [plugin-validator repo](https://github.com/grafana/plugin-validator/issues/1). Thanks @marcusolsson for guiding me here.

## Allow "others" as a fallback category for plugin category.

Currently the plugin schema allows only following 5 categories. Most of the plugins doesn't fall into one of the allowed categories. Examples : Clock panel, Theme panel, Newrelic datasource, Servicenow datasource, etc,. 

image

So, Allowing `others` as a category will help the plugin authors to properly categorize their plugin. Otherwise plugin schema validation will force them to provide any the irrelevant category. Even though the field `category` is not mandatory, it is good to guide the developers properly categorize their plugin.

On the other side, I haven't seen the use of `category` field in [plugins repository](https://grafana.com/grafana/plugins). 

## Mandate minimum one keyword

Mandate minimum one keyword in the info section to improve the plugin discovery in plugin repository.

### DIFF
--- a/docs/sources/developers/plugins/plugin.schema.json
+++ b/docs/sources/developers/plugins/plugin.schema.json
@@ -7,6 +7,11 @@
   "required": ["type", "name", "id", "info", "dependencies"],
   "additionalProperties": false,
   "properties": {
+    "$schema": {
+        "type": "string",
+        "description": "Schema definition for the plugin.json file",
+        "enum": ["https://raw.githubusercontent.com/grafana/grafana/master/docs/sources/developers/plugins/plugin.schema.json"]
+    },
     "id": {
       "type": "string",
       "description": "Unique name of the plugin. If the plugin is published on grafana.com, then the plugin id has to follow the naming conventions.",
--- a/docs/sources/developers/plugins/plugin.schema.json
+++ b/docs/sources/developers/plugins/plugin.schema.json
@@ -28,8 +28,8 @@
     },
     "category": {
       "type": "string",
-      "description": "Plugin category used on the Add data source page. Possible values are: `tsdb`, `logging`, `cloud`, `tracing`, and `sql`.",
-      "enum": ["tsdb", "logging", "cloud", "tracing", "sql"]
+      "description": "Plugin category used on the Add data source page. Possible values are: `tsdb`, `logging`, `cloud`, `tracing`, `sql` and `others`.",
+      "enum": ["tsdb", "logging", "cloud", "tracing", "sql", "others"]
     },
     "annotations": {
       "type": "boolean",
@@ -151,7 +151,7 @@
     "info": {
       "type": "object",
       "description": "Metadata for the plugin. Some fields are used on the plugins page in Grafana and others on grafana.com if the plugin is published.",
-      "required": ["logos", "version", "updated"],
+      "required": ["logos", "version", "updated", "keywords"],
       "additionalProperties": false,
       "properties": {
         "author": {
@@ -182,6 +182,7 @@
         "keywords": {
           "type": "array",
           "description": "Array of plugin keywords. Used for search on grafana.com.",
+          "minItems": 1,
           "items": {
             "type": "string"
           }
--- a/docs/sources/developers/plugins/plugin.schema.json
+++ b/docs/sources/developers/plugins/plugin.schema.json
@@ -9,8 +9,7 @@
   "properties": {
     "$schema": {
         "type": "string",
-        "description": "Schema definition for the plugin.json file",
-        "enum": ["https://raw.githubusercontent.com/grafana/grafana/master/docs/sources/developers/plugins/plugin.schema.json"]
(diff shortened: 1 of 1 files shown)


yesoreyeram:
Without `$schema` field, editors will complain that `$schema` field is not allowed.

image

docs/sources/developers/plugins/plugin.schema.json
```diff
@@ -23,8 +28,8 @@
     },
     "category": {
       "type": "string",
-      "description": "Plugin category used on the Add data source page. Possible values are: `tsdb`, `logging`, `cloud`, `tracing`, and `sql`.",
-      "enum": ["tsdb", "logging", "cloud", "tracing", "sql"]
+      "description": "Plugin category used on the Add data source page. Possible values are: `tsdb`, `logging`, `cloud`, `tracing`, `sql` and `others`.",
+      "enum": ["tsdb", "logging", "cloud", "tracing", "sql", "others"]
```
marcusolsson, line 32: Plugin will default to Others if `category` is not set. If we add `others`, should we make the category required? Since it then would be an exhaustive list?

docs/sources/developers/plugins/plugin.schema.json
```diff
@@ -7,6 +7,11 @@
   "required": ["type", "name", "id", "info", "dependencies"],
   "additionalProperties": false,
   "properties": {
+    "$schema": {
+        "type": "string",
+        "description": "Schema definition for the plugin.json file",
+        "enum": ["https://raw.githubusercontent.com/grafana/grafana/master/docs/sources/developers/plugins/plugin.schema.json"]
```
marcusolsson, line 13: I like the idea of being able to auto-complete the plugin.json, but I'd ideally like to have a more stable URL. Should we consider publishing the schema to https://www.schemastore.org/json/, i.e. `https://json.schemastore.org/grafana-plugin-metadata`?

docs/sources/developers/plugins/plugin.schema.json
yesoreyeram: I think this can be answered only if we know the purpose of the "category" field.  Otherwise, tags/keywords are more powerful than the category as the category will be more exhaustive as you said. So it should be either category/keywords should be mandatory.

docs/sources/developers/plugins/plugin.schema.json
yesoreyeram: Ideally this should be published to schema store where the IDEs will search for schema. But the source should be with grafana repo. Since this not a mandatory fields, the enum can be enhanced as we progress. 

Below example shows VS code search for schema list from schema store.

image

docs/sources/developers/plugins/plugin.schema.json
marcusolsson: `category` determines where the data source is displayed within the list of data sources.
Screen Shot 2020-08-18 at 07 43 06

docs/sources/developers/plugins/plugin.schema.json
marcusolsson: Maybe for now, we can remove the `enum` and allow any string until we have a stable URL? I'd like to open a PR to the schemastore once this has been merged.

ryantxu (approved):
LGTM

docs/sources/developers/plugins/plugin.schema.json
marcusolsson: Looking at the code, it looks like `enterprise` and `other` (not `others`) should be added to the schema.

https://github.com/grafana/grafana/blob/5a2efe1f62959f95a549d427bd692bc36dbd3ec3/public/app/features/datasources/state/buildCategories.ts#L5-L13

marcusolsson (changes requested):
I think we should drop the `enum` for `$schema`. We can tighten up the validation once we have a more stable URL.

Also, could you add `enterprise` and `other` to the valid categories? Otherwise I think it looks good!

docs/sources/developers/plugins/plugin.schema.json
yesoreyeram: Done marcus

marcusolsson (approved):

marcusolsson:
Thank you! 🎉

PR #27032, opened 2020-08-17
Add instructions on how to format multiple variables

marcusolsson:

Adds instructions on how to set default variable format for multiple variables.

Thanks @yesoreyeram for noticing this gap in our docs!

### DIFF
--- a/docs/sources/developers/plugins/add-support-for-variables.md
+++ b/docs/sources/developers/plugins/add-support-for-variables.md
@@ -51,13 +51,33 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
 
    ```ts
    async query(options: DataQueryRequest<MyQuery>): Promise<DataQueryResponse> {
-     const query = getTemplateSrv().replace('SELECT * FROM services WHERE id = "$service"'), options.scopedVars);
+     const query = getTemplateSrv().replace('SELECT * FROM services WHERE id = "$service"', options.scopedVars);
 
      const data = makeDbQuery(query);
 
      return { data };
    }
    ```
+
+## Format multiple variables
+
+If the user has selected multiple values, the value of the interpolated variable depends on the [variable format](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/).
+
+A data source can define the default format option when no format is specified, by adding a third argument to the interpolation function.
+
+Let's change the SQL query to use CSV format by default:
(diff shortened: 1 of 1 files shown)


docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -51,14 +51,34 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
 
      const data = makeDbQuery(query);
 
      return { data };
    }
    ```
 
+## Format multiple variables
+
+If the user has selected multiple values, the value of the interpolated variable depends on the [variable format](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/).
```
oddlittlebird, line 64:
```suggestion
If the user selects multiple values, then the value of the interpolated variable depends on the [variable format](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/).
```

oddlittlebird: Unclear what you mean with "multiple values." Do you mean multiple variable values? A variable with multiple values? Multi-value variables? Selected how? In the dropdown? Please clarify.

docs/sources/developers/plugins/add-support-for-variables.md
```diff
@@ -51,14 +51,34 @@ For data sources, you need to use the [getTemplateSrv]({{< relref "../../package
...
+
+A data source can define the default format option when no format is specified, by adding a third argument to the interpolation function.
```
oddlittlebird, line 66:
```suggestion
A data source can define the default format option when no format is specified by adding a third argument to the interpolation function.
```

oddlittlebird (changes requested):

docs/sources/developers/plugins/add-support-for-variables.md
marcusolsson: All of what you said is correct :) Maybe "When a user selects multiple values for variable"?

dprokop (approved):

docs/sources/developers/plugins/add-support-for-variables.md
oddlittlebird: Yes, that makes more sense.

oddlittlebird (approved):
Good job, Marcus!

PR #27051, opened 2020-08-18
Select: experiment with downshift.js as an alternative to react-select

dprokop:
Instead of using full blown react-select, I'm trying to implement selects with https://www.downshift-js.com/ that gives API for controling selects rather than taking care of everything.

TODO:
- [x] auto-positioning 
- [ ] Tests tests tests
- [x] Add support for options creation
- [ ] Implement multi selects(sync/async)
- [ ] Define final, limited(compared to current) API

and probably some more, but this is experimentation to make the Select work and be maintainable.

Notes:
1. popper.js and react-popper has been updated to the latest version, meaning that our current usage had to be slightly updated to reflect API changes. Nothing ground shaking though.
2. API is WIP - want to make it limited compared to current Select while keeping the same prop names whenever possible
3. Please run storybook to test different scenarios (positioning, states etc)
4. Added Fetch component, naive implementation for declarative async function execution, for async select to run options loading.


packages/grafana-ui/src/components/SelectNG/AsyncSelectNG.tsx
```diff
@@ -0,0 +1,135 @@
...
+            !!clearable,
+            !!filterable,
+            dropdownButton,
+            clearButton,
+            getInputProps,
+            {
+              openMenu,
+              closeMenu,
+            }
+          );
```
dprokop, line 80: Signatures of these functions need to be more aligned

packages/grafana-ui/src/components/SelectNG/Fetch.tsx
```diff
@@ -0,0 +1,30 @@
+import { useAsync } from 'react-use';
+
+type AsyncState<T> =
```
dprokop, line 3: copy of type from react-use, to be changed

packages/grafana-ui/src/components/SelectNG/SelectNG.tsx
```diff
@@ -0,0 +1,104 @@
...
+          highlightedIndex,
+          selectedItem,
+          inputValue,
+          openMenu,
+          closeMenu,
+          clearSelection,
+        }) => {
+          const dropdownButton = renderDropdownButton(triggerRef, isOpen, !!disabled, getToggleButtonProps);
+          const clearButton = clearable && selectedItem && renderClearButton(!!disabled, clearSelection);
+          const triggerElement = renderDefaultTrigger(
```
dprokop, line 56: As you can see Select and AsyncSelect have this code duplicated. Opinions on having something similar to SelectBase that would abstract away the Downshift?

packages/grafana-ui/src/components/SelectNG/hooks.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  useEffect(() => {
+    // I want the debounce to be built into AsyncSelect, yay or nay?
+    onLoadOptions.current = debounce(
+      async (inputValue: string | null) => {
+        return await loadOptions(inputValue);
+      },
+      300, // TODO: config via prop?
+      { leading: false }
+    );
```
dprokop, line 80: Happy to hear opinions on that. I personally get frustrated when I need to denounce functions individually. Thought It wood be good for the AsyncSelect to take care of that.

packages/grafana-ui/src/components/SelectNG/utils.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { DropdownButton } from './DropdownButton';
+import { GetToggleButtonPropsOptions, GetInputPropsOptions } from 'downshift';
+import { HorizontalGroup } from '../Layout/Layout';
+import { IconButton } from '../IconButton/IconButton';
+import { Icon } from '../Icon/Icon';
+import { Actions } from 'downshift';
+
+// Ref: https://codesandbox.io/s/bitter-sky-pe3z9?file=/src/index.js
+// https://popper.js.org/docs/v2/modifiers/community-modifiers
+export const minSameWidthModifier = {
```
dprokop, line 12: Makes sure popper is at least the width of the trigger

packages/grafana-ui/src/components/Tooltip/Popover.tsx
```diff
@@ -58,10 +58,9 @@ class Popover extends PureComponent<Props> {
                   placement={placement}
                   referenceElement={referenceElement}
                   eventsEnabled={eventsEnabled}
-                  // TODO: move modifiers config to popper controller
-                  modifiers={{ preventOverflow: { enabled: true, boundariesElement: 'window' } }}
+                  modifiers={[{ name: 'preventOverflow', enabled: true, options: { rootBoundary: 'viewport' } }]}
                 >
-                  {({ ref, style, placement, arrowProps, scheduleUpdate }) => {
+                  {({ ref, style, placement, arrowProps, update }) => {
```
dprokop, line 63: popper 1.x=>2.x migration

packages/grafana-ui/src/components/SelectNG/AsyncSelectNG.tsx
```diff
@@ -0,0 +1,135 @@
...
+    setPopperRef,
+    setPopperElement,
+    setCurrentValue,
+  } = useAsyncSelect<T>({
+    placement,
+    loadOptions,
+  });
+  return (
+    <>
+      <Downshift
```
dprokop, line 45: Will probably use Downshift hooks instead of the render prop

packages/grafana-ui/src/components/SelectNG/types.ts
```diff
@@ -0,0 +1,23 @@
+import { SelectableValue } from '@grafana/data';
+import { Placement } from '@popperjs/core';
+
+export interface SelectNGProps<T> {
```
Clarity-89, line 4: I think it'd be a good idea to extend these props from whatever HTML element is the main one for the Select and then enable passing those props without explicitly defining them. This would enable passing `aria-label` prop, etc.

Clarity-89 (commented):
I needed to do two things to make this PR run in storybook:
1. Import Input in `SelectNG/utils.tsx`
2. Remove `eventsEnabled` from `ReactPopper` in `Popover.tsx` since TS was complaining about it.  

Didn't look in depth at the code as this seems to be a heavy WIP, but a few preliminary thoughts: 
1. The code seems to be quite complex so my main concern is if we aren't replacing the complexity of `react-select` by writing components of similar complexity ourselves?
2. The select menu position for the basic component is storybook seems to be to the left instead below the select without any explicit configuration for it. 
3. Why do we need `popper.js` here?

dprokop:
>The code seems to be quite complex so my main concern is if we aren't replacing the complexity of react-select by writing components of similar complexity ourselves?

So react-select vs downshift is like angular vs react. Either you go all in and rely on library constraints or you use primitive to implement your own behaviours. You are right that a lot of complexity already implemented in react-select will have to be rewritten, which I'm not a huge fan of after implementing i.e. adding custom value or async behavior. But at the same time, thanks to downshift, we have the privilege of being of the entire control of how select behaves.

>The select menu position for the basic component is storybook seems to be to the left instead below the select without any explicit configuration for it.

Yes, that's because of the size of the storybook container in which the select is rendered. It has not enough height for dropdown to fit in, hence it's rendered on the right. That's my observation so far. Use the **positioning** story to see how it behaves in different scenarios, there's a knob for controlling the placement prop.

>Why do we need popper.js here?
For select dropdown to enable positioning.

packages/grafana-ui/src/components/TimePicker/TimeZonePicker.tsx
```diff
@@ -30,6 +30,7 @@ export const TimeZonePicker: React.FC<Props> = props => {
   const filterBySearchIndex = useFilterBySearchIndex();
   const TimeZoneOption = width && width <= 45 ? CompactTimeZoneOption : WideTimeZoneOption;
 
+  console.log(groupedTimeZones);
```
dprokop, line 33: not relevant

packages/grafana-ui/src/components/SelectNG/hooks.tsx
Clarity-89: How about using `useDebounce` [hook](https://github.com/streamich/react-use/blob/master/docs/useDebounce.md) here instead?

packages/grafana-ui/src/components/SelectNG/SelectNG.tsx
Clarity-89: Personally I'd prefer breaking down this code into components instead of using render functions and abstracting this logic into respective components and then composing those components as needed.

packages/grafana-ui/src/components/SelectNG/AsyncSelectNG.tsx
Clarity-89: The same comment about abstracting this to a separate component.

packages/grafana-ui/src/components/SelectNG/SelectMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+  renderOptionLabel?: (value: SelectableValue<T>) => JSX.Element;
+  item: SelectableValue<T>;
+}
+
+export const SelectMenuOption: React.FC<SelectMenuOptionProps<any>> = props => {
+  const theme = useTheme();
+  const styles = getSelectStyles(theme);
+  const { children, item, renderOptionLabel, isSelected, isFocused, ...otherProps } = props;
+  return (
+    <div {...otherProps} className={cx(styles.option, isFocused && styles.optionFocused)}>
```
Clarity-89, line 125: Not relevant here but I think this'd be a perfect use case for using `styled` as it'd abstract all this css login into a separate component.

packages/grafana-ui/src/components/SelectNG/SelectNG.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  {
+    value: 'opt3',
+    label: 'Option 3',
+  },
+];
+const fakeOnChangeHandler = () => {};
+describe('Select', () => {
+  describe('props', () => {
+    test('isOpen', async () => {
```
Clarity-89, line 25: Do we have use cases for `isOpen` to be controlled externally? Delegating this logic to Select would be better imo.

packages/grafana-ui/src/components/SelectNG/hooks.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const popperProps = {
+    style: popper.styles.popper,
+    ...popper.attributes.popper,
+  };
+
+  const defaultFilter = (o: SelectableValue, inputValue: string) => {
+    return o.label!.includes(inputValue);
+  };
+  // naive filter implementation, will probably need customization
+  const filterOptions = (filterable: SelectableValue[], inputValue?: string | null) => {
```
Clarity-89, line 51: Maybe we could use [match-sorter](https://github.com/kentcdodds/match-sorter) for this?

packages/grafana-ui/src/components/SelectNG/utils.tsx
```diff
@@ -0,0 +1,194 @@
...
+    state.elements.popper.style.width = `auto`;
+  },
+};
+
+// Converts SelectableValue to human-readable representation
+// Needs customization via props probably, similar to getOptionLabel
+export const itemToString = (value: SelectableValue | null) => value?.label || '';
+
+// Will need some customization most probably
+export const renderDropdownButton = (
```
Clarity-89, line 36: Not sure why we need these render functions for the select components.

Clarity-89 (commented):
Gave this a more through second pass and I like the general idea of using Downshift and the accessibility features we get from it. I think we'd get more use out of those, particularly have semantic `ul` and `li` elements for the menu components and use `getRootProps` prop to get all the proper aria attributes for the root element. 

I prefer the structure of `AsyncSelect` where the components are used directly without wrapper render functions, which add more clutter in my opinion.  

I guess there is a need to have `isOpen` controlled from outside of the component, otherwise I think it'd make it easier to have this logic inside the Select.

PR #27053, opened 2020-08-18
Alerting: copy queryType prop to tsdb.Query

### LINKED ISSUE #27048: Alerting: Difference in Query data structure compared to user query
I use `DataSourceWithBackend.query` default implementation on the frontend to do the user side querying. When logging the query structure on the backend side I see this difference between user query vs alerting query:

User query:
```
Queries:[map[
    Interval:5e+09
    JSON:map[
      columns:[
        ErrorStatistics.TotalCount
      ]
      datasource:x-ray-datasource
      datasourceId:14
      intervalMs:5000
      maxDataPoints:800
      orgId:1
      query:
      queryType:getTimeSeriesServiceStatistics
      refId:A
      resolution:60
    ]
    MaxDataPoints:800
    QueryType:getTimeSeriesServiceStatistics
    RefID:A
    TimeRange:map[From:2020-07-29T12:39:08.46+02:00 To:2020-07-29T13:39:08.46+02:00]
  ]]
```

Alerting query (using the same panel/query as the user query for this)

```
Queries:[map[
    Interval:0
    JSON:map[
      columns:[
        ErrorStatistics.TotalCount
      ]
      queryType:getTimeSeriesServiceStatistics
      refId:A
    ]
    MaxDataPoints:0
    QueryType:
    RefID:A
    TimeRange:map[
      From:2020-07-29T13:34:01.975+02:00
      To:2020-07-29T13:39:01.975+02:00
    ]
  ]]
```

The main issue is the QueryType which is used in `QueryTypeMux` to route the query to the handler.

This is the query that is sent during the user query from grafana to the backend:
Screenshot from 2020-07-29 16-42-32

Seems like when doing user query the QueryType is taken from the json data and added also to the Query itself which is not happening when doing alerting query. 

I would expect for the query to have the same structure (maybe expect the parts which are dashboard specific like `maxDataPoints` etc).

kylebrandt:
Copies queryType property in model for the case of alerting. This field was added to our metrics endpoint, but not in the code that fetches the query model from the dashboard and create the alerting query from it.

Fixes  #27048

@aocenas Can you try this and make sure it fixes what you are seeing? I'm not seeing a good place to put a test for this at the moment.

### DIFF
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -214,13 +214,15 @@ func (c *QueryCondition) executeQuery(context *alerting.EvalContext, timeRange *
 }
 
 func (c *QueryCondition) getRequestForAlertRule(datasource *models.DataSource, timeRange *tsdb.TimeRange, debug bool) *tsdb.TsdbQuery {
+	queryModel := c.Query.Model
 	req := &tsdb.TsdbQuery{
 		TimeRange: timeRange,
 		Queries: []*tsdb.Query{
 			{
 				RefId:      "A",
-				Model:      c.Query.Model,
+				Model:      queryModel,
 				DataSource: datasource,
+				QueryType:  queryModel.Get("queryType").MustString(""),
 			},
 		},
 		Headers: map[string]string{


aknuds1 (approved):

aocenas (approved):
Nice, seems to be working for me, thanks.

PR #27054, opened 2020-08-18
AzureAnalytics: Use fillmode Null with format as time series

### LINKED ISSUE #27013: Insights Analytics data (time series) are repeated during timeline
**What happened**:

Hi, I don't know if is it a bug or some configuration that I can't change.

Well, I'm trying to create a Graph using a Application Insights data from azure. When I use the service "Insights Analytics" the values on columns are repeated during timeline, as you can see bellow, when I transform it to Time Series.

image

If I use a service type "Application Insights" it works, but I have no options to improve my query as I wish.

image

- Grafana version: 7.1.3
- Data source type & version: Azure Analytics/Insights
- User OS & Browser: Windows/Edge
- Grafana plugins: Graph
- Others:

kylebrandt:
When formatting as time series, and having string columns that trigger the long to wide conversion, we were using fill mode previous. This is not what we want here, instead we make them nil.

Fixes #27013

previously was using fill mode previous, which is not what we want here (this was a mistake on my part).
Also we do not have fill missing with this data source anyways, so null makes more sense.

### DIFF
--- a/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
+++ b/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
@@ -184,7 +184,7 @@ func (e *AzureLogAnalyticsDatasource) executeQuery(ctx context.Context, query *A
 	if query.ResultFormat == "time_series" {
 		tsSchema := frame.TimeSeriesSchema()
 		if tsSchema.Type == data.TimeSeriesTypeLong {
-			wideFrame, err := data.LongToWide(frame, &data.FillMissing{Mode: data.FillModeNull})
+			wideFrame, err := data.LongToWide(frame, nil)
 			if err == nil {
 				frame = wideFrame
 			} else {
--- a/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
+++ b/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
@@ -184,7 +184,7 @@ func (e *AzureLogAnalyticsDatasource) executeQuery(ctx context.Context, query *A
 	if query.ResultFormat == "time_series" {
 		tsSchema := frame.TimeSeriesSchema()
 		if tsSchema.Type == data.TimeSeriesTypeLong {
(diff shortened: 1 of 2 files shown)


kylebrandt:
Converted to draft, want to think about changing the SDK...

In LongToWide, we call GetMissing.

```go
LongToWide(longFrame *Frame, fillMissing *FillMissing) (*Frame, error) {
```

```go
// GetMissing returns the value to be filled for a missing row field.
func GetMissing(fillMissing *FillMissing, field *Field, previousRowIdx int) (interface{}, error) {
	if fillMissing == nil {
		return nil, fmt.Errorf("fill missing is disabled")
	}
	var fillVal interface{}
	switch fillMissing.Mode {
	case FillModeNull:
	//	fillVal = nil
	case FillModeValue:
		convertedVal, err := float64ToType(fillMissing.Value, field.Type())
		if err != nil {
			return nil, err
		}
		fillVal = convertedVal
	case FillModePrevious:
		// if there is no previous value
		// the value will be null
		if previousRowIdx >= 0 {
			fillVal = field.At(previousRowIdx)
		}
	}
	return fillVal, nil
}
```

But we are ignoring the error when we call this. However, due to the way we call it, the default behavior will be to have FillModeNull.

For example:

```
field.Extend(1)
fillVal, err := GetMissing(fillMissing, field, wideFrameRowCounter-1)
if err == nil {
	wideFrame.Set(wideFrameIdx, wideFrameRowCounter, fillVal)
}
```

Extend will be zero-value allocated, so we do not need to call set. So that is why passing `nil` to functions like LongToWide is the same as FillModeNull. However, this results in us ignoring the possible error from the call to `float64ToType` within `GetMissing`.

kylebrandt:
Re my last comment, we discussed in meeting and do want to change the SDK, but the Grafana code here will end up being the same, so can go forward with this without updating the SDK before.

aknuds1 (approved):
Great.

marefr:
Great :+1:

PR #27056, opened 2020-08-18
Overrides: expose byType matcher UI

ryantxu:
This lets you define overrides byType

image

This exposes a selector for *, string, number, time -- maybe we should only include the fields in the picker that have results?


kylebrandt:
Works well from my testing with Azure. 

Main thing that confuses is me is that when selecting "Standard Options", you do not see there that an override is needed to change it, as they are blank even in the case where the datasource returns displayName or unit. Seems like we could use some sort of UI indicator there.

Although with overrides having matchers, I'm not sure I understand the point of "Standard Options", as I can not think of a scenario where it overrides would not also solve it.

dprokop:
> This exposes a selector for *, string, number, time -- maybe we should only include the fields in the picker that have results?

Agree we should display only suggestion for the fields with results. Apart from that, I'm not sure about * selector. Isn't it the same as just specifying field option rather than override?

packages/grafana-ui/src/components/MatchersUI/FieldTypeMatcherEditor.tsx
```diff
@@ -0,0 +1,89 @@
...
+    },
+    [counts, props.onChange]
+  );
+
+  const selectedOption = selectOptions.find(v => v.value === options);
+  return <Select value={selectedOption} options={selectOptions} onChange={onChange} />;
+});
+
+const allTypes: Array<SelectableValue<string>> = [
+  { value: FieldType.number, label: 'Numeric fields' },
```
dprokop, line 23: Nit: Numeric fields -> just Numeric ? Matcher name already suggests "fields"

dprokop (approved):
Think this looks good, some nits to consider:) @ryantxu Mind https://github.com/grafana/grafana/pull/27070 btw ;)

kylebrandt:
> I'm not sure about * selector. Isn't it the same as just specifying field option rather than override?

@dprokop Not presently. If a dataframe has Field Config information, then that currently takes precedence over the Field Options. So in this situation, the Field Options have no impact.

dprokop:
Ah yes, this makes sense then @kylebrandt

ryantxu:
Good point @dprokop -- updated with:
image

ryantxu:
Given that https://github.com/grafana/grafana/pull/27070 should also be in 7.1.4, I think we should remove the special case for "*" -- that makes more sense for something that is explicitly regex

PR #27058, opened 2020-08-18
WIP: use "server admin" role for grafana admin

### LINKED ISSUE #26790: Standardize the admin role names

The documentation and UI currently use a variety of terms to describe admin roles:
* Admin user
* Server admin
* Grafana admin  (this is the flag you set in the server config)
* Super admin 
* Organization admin

In reality we have two real admin roles:  grafana and org admin, but we do not really want to promote "Org admin" as a common term as we figure out what to do with orgs.

I suggest we use:
* Server admin -- this use has access to the "Server admin" page
* Admin -- this role has access to the "Configuration" page within an org

**Why is this needed**:

So we have consistent terms in the UI and documentation that are easy to explain

ryantxu:
Fixes #26790

Not sure this is a great idea... now that I see how much it changs!

The alternative is to make sure we clearly show "Grafana Admin" means you have access to the "Server Admin" section


torkelo (commented):
This looks good except the breaking HTTP API change, is there away to make that backward compatible, otherwise might have wait for Grafana v8? or maybe is small enough of a breaking change to be ok in a minor release. 

thoughts @marefr ?

public/app/features/admin/state/actions.ts
```diff
@@ -93,9 +93,9 @@ export function deleteUser(userId: number): ThunkResult<void> {
   };
 }
 
-export function updateUserPermissions(userId: number, isGrafanaAdmin: boolean): ThunkResult<void> {
+export function updateUserPermissions(userId: number, isServerAdmin: boolean): ThunkResult<void> {
   return async dispatch => {
-    const payload = { isGrafanaAdmin };
+    const payload = { isServerAdmin };
```
marefr, line 98: Suggest to keep this backward compatible until v8:

```suggestion
    const payload = { isGrafanaAdmin: isServerAdmin };
```

public/app/features/admin/state/actions.ts
```diff
@@ -212,10 +212,10 @@ export function loadUserMapping(username: string): ThunkResult<void> {
   return async dispatch => {
     try {
       const response = await getBackendSrv().get(`/api/admin/ldap/${encodeURIComponent(username)}`);
-      const { name, surname, email, login, isGrafanaAdmin, isDisabled, roles, teams } = response;
+      const { name, surname, email, login, isServerAdmin, isDisabled, roles, teams } = response;
```
marefr, line 215:
```suggestion
      const { name, surname, email, login, isGrafanaAdmin, isDisabled, roles, teams } = response;
```

public/app/features/admin/state/actions.ts
```diff
@@ -212,10 +212,10 @@ export function loadUserMapping(username: string): ThunkResult<void> {
...
       const userInfo: LdapUser = {
         info: { name, surname, email, login },
-        permissions: { isGrafanaAdmin, isDisabled },
+        permissions: { isServerAdmin, isDisabled },
```
marefr, line 218: Suggest to keep this backward compatible until v8:

```suggestion
        permissions: { isServerAdmin: isGrafanaAdmin, isDisabled },
```

public/app/features/admin/state/reducers.test.ts
```diff
@@ -61,7 +61,7 @@ const getTestUser = (): UserDTO => ({
   login: 'user',
   name: 'User',
   avatarUrl: '',
-  isGrafanaAdmin: false,
+  isServerAdmin: false,
```
marefr, line 64:
```suggestion
  isGrafanaAdmin: false,
```

public/app/types/user.ts
```diff
@@ -28,7 +28,7 @@ export interface UserDTO {
   login: string;
   email: string;
   name: string;
-  isGrafanaAdmin: boolean;
+  isServerAdmin: boolean;
```
marefr, line 31: If this is returned from API should be reverted to be backward compatible.

marefr (changes requested):
Great improvement :+1: 

I agree with Torkel that breaking changes in API we probably don't want to do until v8. So I've added some suggestions to keep the API backward compatible. Haven't check documentation updates though so you may have to adjust HTTP API documentation as well. 

It was quite a lot of changes to grasp so bear in mind I may have missed something :)

ryantxu:
I'll close this for now

PR #27070, opened 2020-08-19
Field overrides: Filter by field name using regex

### LINKED ISSUE #26831: Field Overrides: Multi-Field Matching

dprokop:
Closes https://github.com/grafana/grafana/issues/26831

Introduces a new matcher UI that enables matching fields for overrides using a regular expression.

@kylebrandt @mckn do you think this ultra-simple UI is fine for the needs?

----
**tldr;**
As an alternative, we could introduce a sort of combo UI that would either allow selecting from the list or providing a regex, similar as we do with Filter by name transformation. My initial idea was to have that, but I found it too complex:

Kapture 2020-08-19 at 13 34 36

### DIFF
--- a/docs/sources/developers/plugins/plugin.schema.json
+++ b/docs/sources/developers/plugins/plugin.schema.json
@@ -7,6 +7,10 @@
   "required": ["type", "name", "id", "info", "dependencies"],
   "additionalProperties": false,
   "properties": {
+    "$schema": {
+      "type": "string",
+      "description": "Schema definition for the plugin.json file"
+    },
     "id": {
       "type": "string",
       "description": "Unique name of the plugin. If the plugin is published on grafana.com, then the plugin id has to follow the naming conventions.",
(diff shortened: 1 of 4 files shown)


kylebrandt:
Also see https://github.com/grafana/grafana/pull/27056

dprokop:
> Also see #27056

@kylebrandt thanks for the heads up. I guess these two are complementary and might be useful :)

oddlittlebird:
Docs look good.

ryantxu:
> do you think this ultra-simple UI is fine for the needs?

Initially yes -- a small improvement would be to indicate how many fields (if any!) match the pattern.  may also be good to add autocomplete for regexp exact match?

ryantxu:
I removed * from the type matcher, since it feels better to do it explicitly in something that says regex :)

ryantxu (approved):

ryantxu:
@dprokop -- to be clear, I think this is perfect as is for 7.1.4, but we could give it more love for 7.2+

PR #27072, opened 2020-08-19
Testing: Use React Testing Library

### LINKED ISSUE #26989: Testing: Switch to React Testing Library
Recently, while writing some tests for a component using @grafana-ui's `Form` component with Enzyme, I thought that this way of testing is very inefficient and time-consuming. Then I wrote the same tests with [React testing Library](https://github.com/testing-library/react-testing-library) and had a much better experience. Since the talks of switching to RTL has been going on for quite a while now, I think now would be a good time to have a discussion about this. 

Enzyme vs RTL: 
I made a [gist ](https://gist.github.com/Clarity-89/0e484d76cb612f396c76173c9e55b383) where I wrote the same tests using both libraries and using RTL was a much better experience. 

Pros of using RTL:
- A more integration-minded testing approach where the components are being tested from the user interaction perspective vs focusing on the implementation details. This makes tests less brittle. 
- Better support for components with async logic, no need to use `act` in most of the cases or do smth like `wrapper.update()` to trigger state changes.
- Works really well with react-hook-form, which is the base for @grafana-ui `Form` component and is a [recommended way](https://react-hook-form.com/advanced-usage#TestingForm) to test RHF.
- Lots of queries to get DOM elements, particularly by text, which requires some extra code to do in Enzyme.
- Can be used together with Enzyme until the tests are refactored. It's a dev dependency, so shouldn't impact the production build.

Cons:
- May require custom queries to get elements by `aria-label`, the preferred way is to use `data-testid`.

This issue is intended to be a starting point for discussion so feel free to please some comments, I probably have missed some details.

Clarity-89:
This PR introduces react-testing-library as an alternative to Enzyme. The tests for `DashboardSearch` component have been re-written with RTL to make sure that it's setup properly. 

Fixes https://github.com/grafana/grafana/issues/26989

As mentioned in the comment to the issue, the `*ByLabelText` queries also match `aria-label` so it's possible to continue using them.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,6 +12,7 @@ module.exports = {
   testRegex: '(\\.|/)(test)\\.(jsx?|tsx?)$',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
   setupFiles: ['jest-canvas-mock', './public/test/jest-shim.ts', './public/test/jest-setup.ts'],
+  setupFilesAfterEnv: ['./public/test/setupTests.ts'],
   snapshotSerializers: ['enzyme-to-json/serializer'],
   globals: { 'ts-jest': { isolatedModules: true } },
   moduleNameMapper: {
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "@types/slate": "0.47.1",
     "@types/slate-plain-serializer": "0.6.1",
     "@types/slate-react": "0.22.5",
+    "@types/testing-library__jest-dom": "^5.9.2",
     "@types/testing-library__react-hooks": "^3.2.0",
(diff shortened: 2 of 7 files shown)


dprokop:
This is great, have pretty much the same setup on my select branch, but haven't pushed yet, you were faster :D <3

dprokop:
Can we get some numbers on how many tests we need to migrate?

Clarity-89:
> Can we get some numbers on how many tests we need to migrate? Would love to live with enzyme and react-testing-library side by side as short as possible.

This is what I get after a quick search in Goland, also includes enterprise files.
Screenshot 2020-08-19 at 15 37 18

ryantxu (approved):
LGTM

dprokop (approved):

PR #27073, opened 2020-08-19
Field overrides: Overrides UI improvements

### LINKED ISSUE #25433: Panel Edit: improve the readability/content of overrides
Currenttly, when overrides are collapsed it's hard to understand what is the override applying to:
image

There is info about what matcher has been used but based on community and customer feedback, we are missing information about what name was the matcher configured with and what properties were actually overriden.

TODO:
- [ ] Consult with @grafana/grafana-labs-ux-design what information and in what form should be displayed in the collapsed override
- [ ] Update panel edit/overrides documentation

dprokop:
Closes https://github.com/grafana/grafana/issues/25433
Based on #27070 

Introduces some more information about overrides when in collapsed state.
| Before  | After  |
|---|---|
| Screen Shot 2020-08-19 at 15 25 31 |Screen Shot 2020-08-19 at 15 25 03 |


public/app/features/dashboard/components/PanelEditor/OverrideEditor.tsx
```diff
@@ -118,8 +118,14 @@ export const OverrideEditor: React.FC<OverrideEditorProps> = ({
...
           <div className={styles.overrideDetails}>
-            Matcher <Icon name="angle-right" /> {matcherUi.name} <br />
-            {override.properties.length === 0 ? 'No' : override.properties.length} properties overriden
+            Matcher options <Icon name="angle-right" /> {matcherUi.optionsToLabel(override.matcher.options)}
+            <br />
+            Properties overridden <Icon name="angle-right" />
+            {override.properties
+              .map(p => {
+                return registry.get(p.id).name;
+              })
```
kaydelaney, line 127: Small nit:
```suggestion
              .map(p => registry.get(p.id).name)
```

kaydelaney (approved):
LGTM

ryantxu (approved):
LGTM

jessover9000:
All in all this looks great! I have a small padding nitpick: Due to some whitespace between the actual padding and the top of the text, the spacing between each expandable section and the lines separating them looks off.
image

And then I feel like there's a lot of repetition in the collapsed preview texts. I'm wondering text-wise whether it would make sense to move the matcher up into the title to shorten the entire thing, like so: 
image
This would help the users see quicker which filter to expand without looking at the small detail text. If you expand the content, it could still have the same text it has now so it would still be clear that it's filtering by the "matcher options"

oddlittlebird:
I assume doc updates are coming later?

thiagogcm:
@dprokop Will it be possible to use template variables as well?

dprokop:
Thanks for the review @jessover9000 ! Good catch with the spacing, I'm on it.

Regarding 
>And then I feel like there's a lot of repetition in the collapsed preview texts. I'm wondering text-wise whether it would make sense to move the matcher up into the title to shorten the entire thing, like so:
image
This would help the users see quicker which filter to expand without looking at the small detail text. If you expand the content, it could still have the same text it has now so it would still be clear that it's filtering by the "matcher options"

I see your point, but I have concerns about the readability in case when the options are long. The approach you are describing was the first one I took, but I think, with long options, the readability drastically decreases. Compare these two:

| image  |  image  |
|---|---|

One thing I consider is replacing the **Matcher options** with **Options** only and adding an ellipsis to long options and properties so that the every override has the same height:

image

WDYT @jessover9000

dprokop:
> I assume doc updates are coming later?

@oddlittlebird there isn't much to update as far as I see. Correct me if I'm wrong :)

jessover9000:
I see your point. But I'm still not convinced that the matcher name should be in the small options where it is so easy to be overlooked. Could we introduce a line break if it gets too long for one line, like so? 
image

dprokop:
@jessover9000 
> I see your point. But I'm still not convinced that the matcher name should be in the small options where it is so easy to be overlooked.

matcher replaces the `Override X` in this PR:
image

So I think this is the hardest thing to miss in the updated UI. 

I'd like to keep the complexity of the overrides display (collapsed) as low as possible. Introducing line breaks feels like adding a lot of visual clutter if some of the overrides take two lines and other single line.

oddlittlebird:
Sorry, @dprokop. You are correct.

Issue #26927, opened 2020-08-11
Flux queries (like pivot) contain metadata columns "result" and "table"


loganmc10:
**What happened**:

When using functions like |> pivot(), the resulting table contains 2 extra columns: "result", and "table", these columns cannot be removed using a function like drop(), the only way to remove them is using a transformation to hide them, but then I can't use alerts because they are incompatible with transformations.

**What you expected to happen**:

Either automatically remove/hide these columns, or have a way to do it in the query editor, rather than using a transformation.

**How to reproduce it (as minimally and precisely as possible)**:

Use the pivot function in a flux query, look at the resulting table

**Anything else we need to know?**:

I am using InfluxDB 2 (beta)

**Environment**:
- Grafana version: 7.1.3
- Data source type & version: InfluxDB 2 OSS, Flux query language
- OS Grafana is installed on: Docker
- User OS & Browser: Chrome, Windows
- Grafana plugins: None
- Others:

kchervonets:
You can remove them using
Transform -> Organize fields -> there is an eye icon, it will hide the fields

loganmc10:
Yes that it what I do now, but then I can't use any alarms

ryantxu:
Fixed by #27081

@loganmc10 can you check the latest nightly builds?  will be included in 7.2

loganmc10:
Unfortunately I don't really have a dev/test environment setup, but I trust you! I look forward to 7.2

PR #27081, opened 2020-08-19
InfluxDB: exclude result and table column from Flux table results

ryantxu:
Extracted from #26634

For non-timeseries results, we current include two useless columns: "results" and "table."  This PR removes them:

image


pkg/tsdb/influxdb/flux/executor.go
```diff
@@ -20,29 +20,25 @@ func executeQuery(ctx context.Context, query queryModel, runner queryRunner, max
...
+	// Make sure there is at least one frame
+	if len(dr.Frames) < 1 {
+		dr.Frames = append(dr.Frames, data.NewFrame(""))
 	}
-
+	firstFrame := dr.Frames[0]
+	if firstFrame.Meta == nil {
+		firstFrame.SetMeta(&data.FrameMeta{})
+	}
+	firstFrame.Meta.ExecutedQueryString = flux
```
ryantxu, line 41: These changes make sure we always return at least one frame, and the frame has the `ExecutedQueryString` set.

Previously, this was only attached when `runQuery` had an error, not when readDataFrames produces an error (or no results)

pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -270,3 +269,34 @@ func TestGoldenFiles(t *testing.T) {
 		}
 	})
 }
+
+func TestRealQuery(t *testing.T) {
+	t.Skip() // this is used for local testing
```
ryantxu, line 274: @aknuds1 -- thoughts on how/if we should include this?

I added this so I could easily start the debugger and step though execution process locally.

aknuds1: I will have to check,when time allows.

aknuds1 (changes requested):
Just a quick review for now, but left a suggestion for an error message.

pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -270,3 +269,34 @@ func TestGoldenFiles(t *testing.T) {
...
+			JsonData: json,
+			SecureJsonData: securejsondata.GetEncryptedJsonData(map[string]string{
+				"token": "PjSEcM5oWhqg2eI6IXcqYJFe5UbMM_xt-UNlAL0BRYJqLeVpcdMWidiPfWxGhu4Xrh6wioRR-CiadCg-ady68Q==",
+			}),
+		}
+
+		runner, err := runnerFromDataSource(dsInfo)
+		if err != nil {
+			t.Fatal(err)
+		}
```
aknuds1, line 291:
```suggestion
		require.NoError(t, err)
```

pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -270,3 +269,34 @@ func TestGoldenFiles(t *testing.T) {
...
+
+		dr := executeQuery(context.Background(), queryModel{
+			MaxDataPoints: 100,
+			RawQuery:      "buckets()",
+		}, runner, 50)
+		err = experimental.CheckGoldenDataResponse("./testdata/buckets-real.golden.txt", &dr, true)
+		if err != nil {
+			t.Fatal(err)
+		}
```
aknuds1, line 300:
```suggestion
		require.NoError(t, err)
```

pkg/tsdb/influxdb/flux/executor_test.go
aknuds1: I guess it's OK to keep it like this :shrug:

pkg/tsdb/influxdb/flux/executor.go
```diff
@@ -20,29 +20,25 @@ func executeQuery(ctx context.Context, query queryModel, runner queryRunner, max
...
 
-	dr = readDataFrames(tables, int(float64(query.MaxDataPoints)*1.5), maxSeries)
-
-	for _, frame := range dr.Frames {
-		if frame.Meta == nil {
-			frame.Meta = &data.FrameMeta{}
-		}
-		frame.Meta.ExecutedQueryString = flux
+	// Make sure there is at least one frame
+	if len(dr.Frames) < 1 {
```
aknuds1, line 34: I think this is more idiomatic and reads clearer:
```suggestion
	if len(dr.Frames) == 0 {
```

aknuds1 (approved):
LGTM, but see suggestions.

ryantxu:
Fixes #26927

PR #27086, opened 2020-08-20
InfluxDB: Update Flux placeholder URL with respect to latest Go client

aknuds1:
Remove /api/v2 suffix from Flux placeholder URL, since after upgrading the Go client, it's no longer required. Actually, if you use the suffix now, you get a 404 from InfluxDB. I assume the Go client now applies the suffix automatically, or something along those lines.


ryantxu (approved):
Should this be added to 7.1.4?

aknuds1:
@ryantxu I think so.

PR #27087, opened 2020-08-20
CI: Migrate to Drone for master builds

aknuds1:
Perform the migration to Drone for CI master builds.

- [x] Add enterprise pipeline back, in order to build and publish enterprise snapshot packages as well as Docker images, plus deploy enterprise image to Kubernetes

I will try to implement triggering of enterprise builds via the Drone [downstream](https://github.com/grafana/drone-downstream) plugin. The build number can be passed via an environment variable.


marefr (approved):
LGTM based on a quick lock. Easiest way to verify is probably to merge it and see what happens :) As discussed with Arve, better to wait with merge until he's back from vacation.

marefr:
Go go go

marefr:
Need to have a solution for enterprise master builds before merging this.

aknuds1:
@xlson @sakjur Having discussed with @marefr, I think Drone master builds also need to execute an enterprise pipeline, so we can upload enterprise snapshot packages and Docker images with the same version string as OSS, plus deploy the enterprise image to Kubernetes.

xlson:
@aknuds1 having enterprise master being trigger from the oss one makes sense to me. Are you planning on adding it to the oss flow or to trigger a flow within the enterprise with the commit-hash as an input?

aknuds1:
@xlson I haven't concluded yet. I see there's a Drone plugin to trigger a downstream build, where you can pass variables, so maybe that's actually a solution, instead of building enterprise from OSS.

sakjur (approved):

Issue #5475, opened 2016-06-30
[Feat Req]: Table panel, limit string length in by Column Style


3IWOH:
- **I'm submitting a ...**
- [x] Feature request

I am using Grafana to display a list of application alerts/errors in a table. The specific details of the error string can vary in length. It would be great to be able to limit the display of a table value by x number of characters to keep the view consistent. I imagine this could be added to Column Styles.

Perhaps there could also be the ability to see the entire string via a tootip when hovering over a truncated value?

Ideally, there would be a way to click on the truncated value to drill down into a detail dashboard...I know this has already been suggested for the table panel.

Mikanoshi:
Agreed, requests can be very long (when someone's fuzzing for example or you are making requests in grafana) and table widths glitch, horizontal scrolling appears and headers don't match columns.

![grafana](https://cloud.githubusercontent.com/assets/5113439/17137022/249778ca-5352-11e6-923e-3126f3f7d8be.jpg)

iamash:
Yes that'd be great. Anything updated on this request?

CantankerousBullMoose:
If this were implemented, I would like to see text wrap when it exceeds the specified column width for my usecase.  I could also see a need for setting a maximum row height and doing the tool-tip or drill down for text that is too long to show within a cell as the OP suggested.

CSSassess:
I'm out of the office returning July 30th

Your mail has been received and I'll get to it upon my return. For anything
urgent please contact IT Operations in my absence.

Robert Klickermann
Systems Architect
Hamilton Police Service
(905) 546-4791

***This email has been scanned for malicious content ***

hackeryard:
Hello, I face a problem that I use grafana v5.4.2, but the table palne doesn't support the limited length of row string. But I find this issue and also the changelog says this feature is included in v4.4. Is there anything I missed?

benjdl:
Any update on this issue ? No word on the roadmap about it

criskolkman:
So @torkelo is there any update on this?
I would like to shorten the output as well.

![image](https://user-images.githubusercontent.com/17678609/86252451-8699fc00-bbb3-11ea-87f8-2cb9ff9c67b3.png)

torkelo:
This should not be needed in new table panel

SincerelyUnique:
![image](https://user-images.githubusercontent.com/16249717/100960503-536be800-3518-11eb-98ef-e186b99556a9.png)

PR #27088, opened 2020-08-20
Panel: Table cell display mode - popup

yesoreyeram:

With this PR, table cells can be displayed as Popup / Modal window.

To Do:

- [ ] Ability to customize the popup content. Modal window should have the ability to render custom text/ custom html of the data provided. Including other cell data like data links.
- [ ] Ability to customize the **show** button. Show button text can be customized and background, text or color of the button/text should be reactive based on thresholds.

**Example:**

Look at the description field

image
image

Which issue(s) this PR fixes:

#5475


dprokop:
@yesoreyeram thanks for the contribution. This sounds like an interesting idea. @torkelo WDYT? Seems to me like a good way for showing long value cells. 

Another idea, instead of having a customizable "Show" button would be to truncate the value of the cell and convert it to a link opening a modal.

Conversation we had on Slack
```
yesoreyeram:grafana:  1:05 PM
Hi @dominik Hope you are doing good. If you have couple of mins, i need your early feedback about the PR https://github.com/grafana/grafana/pull/27088 .
Q1: Is it worth investing time on this or core team have other ideas?
Q2: If yes, does the current approach of using TableCellDisplayMode for this a good approach? Or any other approach should be followed to achieve this?
Q3: If answer for Q2 is yes, how to pass custom props to the TableCellDisplayMode (edited) 

dominik  1:12 PM
Hello @yesoreyeram!  I think you are on a good path here
1:12
I like this idea, but let me verify with Torkel who might have some ideas.
1:12
What custom props would you like to pass to the cell?

yesoreyeram:grafana:  1:13 PM
like customizing popup content.. like rendering detailed HTML report / links with content from sibling cells.
1:14
sometimes multile sibling cells content can be grouped into a sinlge popup.
1:14
so i want to pass some custom props like. popup content, show button text, etc..
1:15
This is something similar to customizing the inline bar chart

dominik  1:15 PM
yeah, I see

yesoreyeram:grafana:  1:15 PM
so i want my cell to be sibling aware and handle the content like data link

dominik  1:16 PM
currently there is no way for providing additional config for table cells display mode, but this also seems like an interesting idea.
1:16
so basically, the content of your cell may contain a data link templates?
1:16
do you need interpolation?

yesoreyeram:grafana:  1:16 PM
yes.
1:18
basically i will pass content like <a href="${__data.fields[url]}">details about ${__data.fields[number]}</a><img/>

dominik  1:18 PM
yeah, i get it
1:18
but will this be a conten of the popup, or the cell itself?

yesoreyeram:grafana:  1:19 PM
content of the popup. ( where i can refer cell content by ${__data.fields[number]}

dominik  1:19 PM
ok, I see
1:19
hm, let me think

yesoreyeram:grafana:  1:19 PM
content may/maynot contain references about cell. sometime plain text may go.

dominik  1:19 PM
also- can you provide the above details in the PR?
New

yesoreyeram:grafana:  1:20 PM
sure mate will do.

dominik  1:21 PM
a possible solution would be to to extend Table custom option and add a conditional option that shows up when cell dispaly mode is set to Modal
1:21
this custom option would be : Render as HTML | JSON | Plain text

yesoreyeram:grafana:  1:21 PM
or override property like datalink?

dominik  1:22 PM
well, not necessarily as then the link will be added to the entire content```

**Comment by dprokop** [comment]
@yesoreyeram mind that panels have accest to interpolate function (https://github.com/grafana/grafana/blob/cbe1d7b08c8d1a9338e86df7efe56f0b139b29fa/packages/grafana-data/src/types/panel.ts#L67) 

We could think about extending Table's interface to accept that function allowing table cells to access it. Then the data links and variables interpolation within the popup content should be possible.

**Comment by yesoreyeram** [comment]
Thanks @dprokop. Seems it will solve the problem for interpolation.  

Do you have any thoughts on how to pass the template for the popup to the cell? on the other words, how to attach additional properties to `TableCellDisplayMode`? ( also some times , we may need to customize the popup header along with the content )

Regarding your comment below comments,

> Another idea, instead of having a customizable "Show" button would be to truncate the value of the cell and convert it to a link opening a modal.

Since the popup can contain any details irrespective of cell content, it may be a good to consider an option on how to display the cell? ( truncated text / custom button / custom link ).

**Comment by dprokop** [comment]
>Do you have any thoughts on how to pass the template for the popup to the cell? on the other words, how to attach additional properties to TableCellDisplayMode? ( also some times , we may need to customize the popup header along with the content )

> Since the popup can contain any details irrespective of cell content, it may be a good to consider an option on how to display the cell? ( truncated text / custom button / custom link ).

All of this seems like an additional option of the display mode. There are a few ways of achieving this.
1. Extending the `displayMode` to accept options and implementing a custom option editor ([addCustomEditor](https://github.com/grafana/grafana/blob/53b9f325c00281576cf823cfdf333f5769dbb1ce/packages/grafana-data/src/types/OptionsUIRegistryBuilder.ts#L115)) that would show additional fields, when this particular display mode is chosen
2. Adding another field config option that's shown when the display mode is set to modal. To do so you can use https://github.com/grafana/grafana/blob/1716b706da536fe5b312541d6cdd972946fa3f89/packages/grafana-data/src/types/fieldOverrides.ts#L95

But to start with this I suggest to minimize the amount of possible options, and possibly iterate over it after initial work is done. I suggest:
1. Adding the modal display option
2. Adding option to display the content of the popup as HTML | JSON | Plain text
3. Rendering the cell as truncated text(link)

**Comment by ryantxu** [comment]
Interesting -- i have not looked closely... but this does not seem appropriate for the CellDisplay.  We could want this behavior with any TableCellDisplayMode right?

Maybe we should have an additional "onClick" mode.... or even better :) somehow having dataLinks open a modal.  Then it could work in any panel the same

**Comment by ryantxu** [comment]
yes, the more I think about it, tying the behavior to datalinks makes more sense to me

**Comment by yesoreyeram** [comment]
With the change, custom render changed to datalinks. So the datalink will have options to choose mode.  ( link - default mode & modal - new popup mode )

## In graph

image
image
image
image

## In table

image
image
image

**Comment by torkelo** [comment]
Think this is better handled as a cell display mode, similar to the original implementation.

**Comment by yesoreyeram** [comment]
As @torkelo mentioned, it is getting complicated with the datalinks implementation. To make it simple,  Reverted the pull request to original state.

* New Cell Display mode introduced  ( displayMode : popup )
* Additional property added to the popup cell display mode added ( popupDisplayMode : auto | json | code ). If required, This can be extended to HTML, Markdown etc.

Setting popupDisplayMode to code/trace will help to display the stack traces.

image

File: public/app/plugins/panel/table/module.tsx [hunk 1]
```diff
@@ -47,10 +47,26 @@ export const plugin = new PanelPlugin<Options, CustomFieldConfig>(TablePanel)
               { value: TableCellDisplayMode.LcdGauge, label: 'LCD gauge' },
               { value: TableCellDisplayMode.BasicGauge, label: 'Basic gauge' },
               { value: TableCellDisplayMode.JSONView, label: 'JSON View' },
+              { value: TableCellDisplayMode.Popup, label: 'Popup' },
               { value: TableCellDisplayMode.Image, label: 'Image' },
             ],
           },
         })
+        .addSelect({
+          path: 'popupDisplayMode',
```
torkelo, line 56: To support cell display modes with options I think the best think would be to change the display mode into an object so that each display mode can have a custom editor and sub options. 

Created an issue for that: https://github.com/grafana/grafana/issues/31315

torkelo:
Closing this for now, we can revisit this after #31315

Issue #26606, opened 2020-07-26
Table Panel converts string like "32E11" to number


negbie:
Table Panel converts string like "32E11" to number like you can see in the screenshot:

![table_num](https://user-images.githubusercontent.com/20154956/88477995-f1392000-cf44-11ea-8aa6-def850f30e7f.png)

**Environment**:
- Grafana version: 7.1.1
- Data source type & version: Clickhouse (latest)
- OS Grafana is installed on: Debian (buster)
- User OS & Browser: Chrome (latest)

negbie:
Could it be that the Clickhouse Datasource causes this?

negbie:
Ok it can be reproduced with Postgresql datasource aswell. See screenshot below
![table_num_pg](https://user-images.githubusercontent.com/20154956/88520699-38321e80-cff4-11ea-90e1-f1ff1960e4f7.png)

negbie:
Please notice that string +4912345 will be shown in table panel as 4912345. If I should open a seperate issue for that please tell me.

torkelo:
My guess is that something is wrong here https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/dataframe/processDataFrame.ts#L204 

If anyone wants to help investigate & fix please do :)

negbie:
As a temporary fix someone can create a Cell display mode override and use JSON view.

PR #27097, opened 2020-08-20
Table: Support showing numbers in strings with full original value

torkelo:
We have a few problems where we always try to parse values as numbers and if possible we do format them and show them as numbers without any way to show the raw string.

for example when we show a value as "10e2" and "1231232.12312312321" 
https://github.com/grafana/grafana/issues/26606

I tried to fix this by introducing a string unit that overrides the logic of always parsing things to numbers in the display processor. 

But units are not applied to string type fields so this did not work in the app:
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/utils/standardEditors.tsx#L74

### DIFF
--- a/docs/sources/panels/field-configuration-options.md
+++ b/docs/sources/panels/field-configuration-options.md
@@ -203,7 +203,7 @@ You can also paste a native emoji in the unit picker and pick it as a custom uni
 
 #### String unit
 
-Grafana can sometime be too agressive in parsing strings and displaying them as numbers. To make Grafana show the original
+Grafana can sometime be too aggressive in parsing strings and displaying them as numbers. To make Grafana show the original
 string create a field override and add a unit property with the `string` unit.
 
 ### Thresholds
--- a/docs/sources/panels/field-configuration-options.md
+++ b/docs/sources/panels/field-configuration-options.md
@@ -99,7 +99,6 @@ Allows you to type in a regular expression against which fields to be overridden
 
 Allows you to select fields by their type (string, numeric, etc).
 
-
 ## Field options
 
 This section explains all available field options. They are listed in alphabetical order.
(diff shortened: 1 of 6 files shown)


ryantxu:
When applied as an override it *does* work:
image

vs
image

image


I'm not sure how common or weird this is -- feels explicitly like a workaround

packages/grafana-data/src/valueFormats/valueFormats.ts
```diff
@@ -156,6 +156,10 @@ export function simpleCountUnit(symbol: string): ValueFormatter {
   };
 }
 
+export function stringFormater(value: number): FormattedValue {
+  return { text: value as any };
```
ryantxu, line 160: or maybe expicitly a string?
```suggestion
  return { text: `${value}` };
```

or:
```suggestion
  return { text: ''+value };
```

dprokop:
@torkelo using `unit` property feels very weird to me. Was thinking, not sure how common problem this will be, but maybe an alternative solution would be to introduce new standard property, something like `Value as = number | string | defined by field`? Then the config would be also done via overrides. Just an idea.

torkelo:
Using unit for this is definitely a bit strange (also that you have to use an override), but so far this seems to be mostly an edge case. And not sure I want to add a new field config setting for something that is very rarely used.

docs/sources/panels/field-configuration-options.md
```diff
@@ -202,6 +201,11 @@ You can also paste a native emoji in the unit picker and pick it as a custom uni
 
 {{< docs-imagebox img="/img/docs/v66/custom_unit_burger2.png" max-width="600px" caption="Custom unit emoji" >}}
 
+#### String unit
+
+Grafana can sometime be too agressive in parsing strings and displaying them as numbers. To make Grafana show the original
+string create a field override and add a unit property with the `string` unit.
```
oddlittlebird, line 207:
```suggestion
string, create a field override and then add a unit property with the `string` unit.
```

docs/sources/panels/visualizations/table-panel.md
```diff
@@ -26,3 +26,10 @@ Table visualizations allow you to apply:
 ## Display options
 
 - **Show header -** Show or hide column names imported from your data source..
+
+## Tips
+
+### Display original string value
+
+Grafana can sometime be too agressive in parsing strings and displaying them as numbers. To make Grafana show the original
+string create a field override and add a unit property with the `string` unit.
```
oddlittlebird, line 35:
```suggestion
string, create a field override and then add a unit property with the `string` unit.
```

oddlittlebird: If possible, add a second tip and an intro paragraph under the Tips heading.

PR #27113, opened 2020-08-20
AzureMonitor: fix panic introduced in 7.1.4 when unit was unspecified and alias was used 

kylebrandt:
happens when unit was "Unspecified" and alias is being used


ryantxu (approved):

torkelo:
Can we cherry pick this into 7.1.5 ?

PR #27124, opened 2020-08-21
Core: Add loki/logproto dependency

joanlopez:

Add `github.com/grafana/loki/pkg/logproto` package to push logs to the Loki HTTP API (used in auditing extension).

- Note that a new `replace` statement has been added [inherited from Loki](https://github.com/grafana/loki/blob/master/go.mod#L72).
- Since some shared dependencies has been updated, I also updated the `logWrapper` implementation to keep it compatible with the latest version of the `hclog.Logger` interface.


xlson:
I agree, lets remove the catch and test since that bug has been fixed.

xlson (approved):
Looks good to me, but I'd like to see a review from @marefr or someone else on the backend squad since this touches logging for the backend plugins.

joanlopez:
Pipeline fixed by applying the discussed changes.

Thanks!

marefr (approved):
LGTM

Issue #15279, opened 2019-02-06
Enhancement Request: Support grouping values into columns on tables by matching tags in graphite


shanson7:
See https://github.com/grafana/grafana/pull/10050

That seems like support is almost there, but not for the graphite datasource. 

### What Grafana version are you using?
v5.3.4 and v6.0.0-beta1

### What datasource are you using?
Graphite (via MetricTank), with tags

### What OS are you running grafana on?
RHEL 7.3

shanson7:
Trying to use the grafana 7 beta "Transforms" to do this. 

Something like:
|User  | Num Requests | Request Duration|
|---|---|---|
|user1 | 35      | 14.15 s |
|user2 | 334        | 2.99s |

I have MDP set to 1, so that I get a single datapoint per series already. The request/response looks like this:
https://gist.github.com/shanson7/ec5245bc8e4f705737f3330a5142a5ec

I've tried using labels to fields and outer join on `userName` but it seems that the labels/tags with the same names (`name` and `Value`) are conflicting.

torkelo:
that example does not include a group by, on a call you mentioned having data center and server (I think), and wanting to group by multiple columns and then get the max value. Can you provide an example for that scenario as well?

shanson7:
> that example does not include a group by

It is grouped by `User`. This request isn't for grafana to do a SQL like "Group By" calculation, but to allow multiple *independent* metrics be represented in a column joined on 1 or more "labels" to form a row key.

torkelo:
So this is what the current labels to fields transform results in 

![Screenshot from 2020-08-20 12-17-38](https://user-images.githubusercontent.com/10999/90758355-2e43c880-e2df-11ea-97d6-85ef1f71be18.png)

Not sure how a group by get's what your after. Looks like you want label values to be fields (column headers), which I not sure I understand how that can work. 

@ryantxu any ideas for a transform that can do this? 

The data looks like this
DataFrame:1
Time
Value (Name="Num Requests", userName="user1")

DataFrame:2
Time
Value (Name="Num Requests", userName="user2")

DataFrame:3
Time
Value (Name="Request Duration", userName="user1")

DataFrame:4
Time
Value (Name="Request Duration", userName="user2")

Tricky to get this to: 

|User  | Num Requests | Request Duration|
|---|---|---|
|user1 | 35      | 14.15 s |
|user2 | 334        | 2.99s |

torkelo:
Think some kind of pivot operation is what we need?

[...]

PR #27125, opened 2020-08-21
Transforms: Labels to fields rewrite that uses merge transform inside it 

torkelo:
Looking at solving a graphite labels grouping issue  (https://github.com/grafana/grafana/issues/15279)  and wanted to try to add an option to labels to fields transform. 

Discovered a big possible simplification to the labels to fields transform. If all this transform did was operate on single data frames and extract the labels to new fields, and then use the merge transform (that did not exist when we wrote labels to fields). Then labels to fields can be made much simpler. 

thoughts @hugohaggmark @mckn ?


hugohaggmark:
Taking it for a spin now

packages/grafana-data/src/transformations/transformers/labelsToFields.test.ts
```diff
@@ -1,232 +1,104 @@
...
-      { name: 'feelsLike', type: FieldType.string, values: new ArrayVector(['ok', null, 'ok', null]), config: {} },
-      { name: 'sky', type: FieldType.string, values: new ArrayVector([null, 'cloudy', null, 'cloudy']), config: {} },
-      { name: 'temp', type: FieldType.number, values: new ArrayVector([1, -1, 2, -2]), config: {} },
+    const expected: FieldDTO[] = [
+      { name: 'time', type: FieldType.time, values: [1000, 2000], config: {} },
+      { name: 'location', type: FieldType.string, values: ['inside', 'outside'], config: {} },
+      { name: 'temp', type: FieldType.number, values: [1, -1], config: {} },
     ];
 
-    expect(result[0].fields).toEqual(expected);
```
hugohaggmark, line 230: Are these tests covered by the `merge` transformer? Those tests tested data frames with different labels.

packages/grafana-data/src/transformations/transformers/labelsToFields.ts
```diff
@@ -1,173 +1,61 @@
...
+    for (const frame of data) {
+      const newFields: Field[] = [];
 
-    const timeColumnItem = columnsMap[fields[0].name];
-    const seriesIndexStrings = Object.keys(timeColumnItem);
-    for (const seriesIndexString of seriesIndexStrings) {
-      const seriesItem = timeColumnItem[seriesIndexString];
-      const timeValueStrings = Object.keys(seriesItem.values);
+      for (const field of frame.fields) {
+        if (field.labels) {
```
hugohaggmark, line 25: nitpick: maybe reverse `if` statement and continue to minimize nesting?

packages/grafana-data/src/transformations/transformers/labelsToFields.ts
```diff
@@ -1,173 +1,61 @@
...
+              config: {},
+            });
+          }
 
-      for (const timeValueString of timeValueStrings) {
-        if (!values[timeValueString]) {
-          values[timeValueString] = [];
+          const newField = { ...field, name };
+          delete newField.labels;
+          delete newField.config.displayName;
```
hugohaggmark, line 46: This looks weird, maybe add a comment.

hugohaggmark (commented):
Looks good ⭐ , love that you could reduce the code so much. But some tests were removed that need some explanation

packages/grafana-data/src/transformations/transformers/labelsToFields.test.ts
torkelo: yes, it has tests with data frames with different fields

packages/grafana-data/src/transformations/transformers/labelsToFields.ts
torkelo: yea, do no love this either. But want to remove these on the value field after the transform. But setting them to undefined is also ugly as it leaves the props there in the object. Maybe that is ok

public/app/core/components/TransformersUI/LabelsToFieldsTransformerEditor.tsx
```diff
@@ -8,14 +15,52 @@ export const LabelsAsFieldsTransformerEditor: React.FC<TransformerUIProps<Labels
...
+        <Select
+          isClearable={true}
+          allowCustomValue={false}
+          placeholder="(Optional) Select label"
+          options={labelNames}
+          className="min-width-18 gf-form-spacing"
+          value={options?.valueLabel}
+          onChange={onValueLabelChange}
+          menuPlacement="bottom"
+        />
```
dprokop, line 53: Would be great to add some information about value field name behavior in the UI

dprokop (commented):
Code-wise looks good, great that you have removed so much complexity there! Think docs need some update as well after introducing new option: https://grafana.com/docs/grafana/latest/panels/transformations/#labels-to-fields

packages/grafana-data/src/transformations/transformers/labelsToFields.ts
torkelo: fixed

torkelo:
Dam merged the wrong PR, need to create a follow up PR with docs and refining this

PR #27128, opened 2020-08-21
Grafana-UI: Alert components tweaks

Clarity-89:
- Add Alert story and .mdx file
- Mark `onButtonClick` and `buttonText` props as deprecated as they conflict with `onRemove`
- Move styles to emotion


Clarity-89:
@torkelo I removed the outline from default close icon, but if any custom button content is provided it will have the outline. 

Screenshot 2020-08-24 at 9 37 24

packages/grafana-ui/src/components/Alert/Alert.tsx
```diff
@@ -60,3 +70,66 @@ export const Alert: FC<AlertProps> = ({ title, buttonText, onButtonClick, onRemo
...
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      width: 35px;
+    `,
+    title: css`
+      font-weight: ${theme.typography.weight.semibold};
+    `,
+    body: css`
+      .alert-body {
```
torkelo, line 118: this nested rule is a bit strange, where is this class defined / used?

Clarity-89: Good catch, that'd be removed completely, missed it while refactoring the old styles.

packages/grafana-ui/src/components/Alert/Alert.tsx
```diff
@@ -1,54 +1,64 @@
...
+}) => {
+  const theme = useTheme();
+  const styles = getStyles(theme, severity, !!removeButtonContent);
+
   return (
-    <div className="alert-container">
-      <div className={alertClass}>
-        <div className="alert-icon">
+    <div className={styles.container}>
+      <div className={styles.alert} aria-label={`alert-${severity}`}>
```
dprokop, line 50: Think we can define this label in the e2e-selectors package and use it both here and in tests.

packages/grafana-ui/src/components/Alert/_Alert.scss
```diff
@@ -8,7 +8,6 @@
 .alert {
```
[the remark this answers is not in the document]
dprokop: I think this should be safe to remove, apart from the page-alert-list OFC.

packages/grafana-ui/src/components/Alert/Alert.tsx
```diff
@@ -1,54 +1,64 @@
...
-interface AlertProps {
+export interface Props {
   title: string;
-  buttonText?: string;
-  onButtonClick?: (event: React.MouseEvent) => void;
   onRemove?: (event: React.MouseEvent) => void;
   severity?: AlertVariant;
   children?: ReactNode;
+  removeButtonContent?: ReactNode | string;
+  /** Deprecated. Use onRemove instead */
```
dprokop, line 16: I think adding `@deprecate` annotation might be helpful here from docs generation perspective

packages/grafana-ui/src/components/Alert/Alert.tsx
```diff
@@ -1,54 +1,64 @@
...
 
-interface AlertProps {
+export interface Props {
   title: string;
-  buttonText?: string;
-  onButtonClick?: (event: React.MouseEvent) => void;
   onRemove?: (event: React.MouseEvent) => void;
   severity?: AlertVariant;
   children?: ReactNode;
+  removeButtonContent?: ReactNode | string;
```
dprokop, line 15: I can't find any usages of this property, maybe we can just remove it? Maybe it's used in enterprise? TBH I don't think we should allow customization of the close/remove button.

dprokop (approved):
I think this looks good, added some minor suggestions 👍

packages/grafana-ui/src/components/Alert/_Alert.scss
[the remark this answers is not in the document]
Clarity-89: Any idea where I could move `page-alert-list` styles to, so this whole file could be deleted then?

packages/grafana-ui/src/components/Alert/Alert.tsx
Clarity-89: Ah, great, didn't know it worked here too.

packages/grafana-ui/src/components/Alert/_Alert.scss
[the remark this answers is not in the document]
dprokop: @Clarity-89 doing quick look at the code. Seems like the class name is passed to the angular wrapper for notifications list: https://github.com/grafana/grafana/blob/d04dce6a37baa21ceb35e3caf7a3644cf33cd4c7/public/views/index-template.html#L214

This wrapper renders https://github.com/grafana/grafana/blob/99411bf37a66f8a8d6720ee06c5849855533d424/public/app/core/components/AppNotifications/AppNotificationList.tsx component, which does not use className at all. Maybe this class is not used in the app at all?

packages/grafana-ui/src/components/Alert/Alert.tsx
Clarity-89: I added `removeButtonContent` as an alternative to `buttonText`, to unify both button and remove properties into one. Apparently there are some cases where a button is needed instead of the close icon.

packages/grafana-ui/src/components/Alert/_Alert.scss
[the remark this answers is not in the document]
Clarity-89: Thanks for the pointer! Seems like this file is a copy of https://github.com/grafana/grafana/blob/master/public/sass/components/_alerts.scss, so it can be safely deleted.

PR #27135, opened 2020-08-21
InfluxDB: Clean up tests

aknuds1:
Clean up tests in InfluxDB to use testify assertions, as we usually do.


kylebrandt (approved):
Looks better to me

wbrowne (approved):
Nice and clean ✨ 

👍

PR #26903, opened 2020-08-10
Dataframe/Alerting: to tsdb.TimeSeriesSlice - accept "empty" time series

kylebrandt:
**What this PR does / why we need it**:
Make it so at least azure insights "empty response" will become no data in alerting, instead of getting an error.

**Which issue(s) this PR fixes**:
I currently assume this is cause of issue #26897.

**Special notes for your reviewer**:
This is perhaps better solved in the SDK but that needs more thought and is tracked in grafana/grafana-plugin-sdk-go#214

PR #27140, opened 2020-08-21
Alerting: remove LongToWide call in alerting

### LINKED ISSUE #27132: Alerting: tsdb.HandleRequest() failed to convert dataframe (LongToWide, no Rows)
**What happened**:
If Frame with no rows is returned, *and* that frame is has time, numeric, and string columns - then long to wide conversion is triggered but fails.

> "tsdb.HandleRequest() failed to convert dataframe "" to tsdb.TimeSeriesSlice: failed to convert long to wide series when converting from dataframe: can not convert to wide series, input fields have no rows"

This should result in no data.

This is a variation of the problem same problem https://github.com/grafana/grafana/pull/26903 solved, but this varation was missed.

I don't think tsdb.FrameToSeriesSlice, which seems to only be called from conditions/query, should even be doing a long to wide conversion - but need to think on that a little more. 

**Environment**:
- Grafana version: 7.1.4

kylebrandt:
Makes it so when alerting receives a "Long" formatted time series with no rows, there is "no_data" instead of an error.

Fixes #27132

This removes the LongToWide call from the tsdb.FrameToSeriesSlice(). Alerting is the only place that calls this function, and this issue is a side effect of this call. The call does not make much sense here, as that call should be done when using formatAs timeseries on a table-model backed data source. (Like Azure's Kusto services, and in the future SQL - when switched to data frames).

This is a variation of the problem fixed in https://github.com/grafana/grafana/pull/26903 that was missed.

### DIFF
--- a/pkg/tsdb/frame_util.go
+++ b/pkg/tsdb/frame_util.go
@@ -54,15 +54,6 @@ func FrameToSeriesSlice(frame *data.Frame) (TimeSeriesSlice, error) {
 		}
 		return nil, fmt.Errorf("input frame is not recognized as a time series")
 	}
-	// If Long, make wide
-	if tsSchema.Type == data.TimeSeriesTypeLong {
-		var err error
-		frame, err = data.LongToWide(frame, nil)
-		if err != nil {
-			return nil, errutil.Wrap("failed to convert long to wide series when converting from dataframe", err)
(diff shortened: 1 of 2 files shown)


marefr (approved):
LGTM

PR #27141, opened 2020-08-21
Transformers: Fix reduce transformer so that you can use display name config without using an override

torkelo:
The reduce transfomer fields did set the displayName, this made Field config option for Display name not have any effect (as having the reducer set this makes it a data driven config). 

Not sure the reduce transformer needs to set displayName here, is there an argument for this?


ryantxu (approved):
looks good -- agree we should not set the name... but maybe add a label?  or maybe optionally add a label?

torkelo:
does not make sense as a label (as it would be part field identity and duplicate info found in name). But maybe we can add it to some future field meta?

PR #27148, opened 2020-08-23
Explore/Logs: Allow toggling of derived fields

fredr:
This fixes a regression from #26178

The log details shows both parsed fields and derived fields. In the current implementation, when toggling a field like `ts` or `tsNs`, nothing is displayed, since those fields are not part of the parsed message.


ivanahuckova (approved):
LGTM! This fix works especially well with Elastic logs where we have more derived fields.

before: 
image

after: 
image

Thanks so much @fredr for your work! 🚀🎉💙

fredr:
> Thanks so much @fredr for your work! rockettadablue_heart

Thanks for testing and reviewing!

judemorrissey:
@fredr my and my team's sincere thanks to you and the grafana team for this contribution, we were in the same boat as you with pretty dense JSON blobs, and it was also a blocker for us to not be able to just view certain parsed fields, cheers!

PR #27153, opened 2020-08-24
API: Make default credentials in XHR request for all browsers

### LINKED ISSUE #27007: Can not login in 7.1.3 with Safari 11.1.2
**What happened**:
When upgrade our grafana server from 6.6.2 to 7.1.3, can not login with safari 11.1.2. Safari works with 6.6.2 and chrome works with both version.

Can login grafana 7.1.3 with safari 11.1.2

Launch grafana server with docker
```
docker run -p 3000:3000 grafana/grafana:7.1.3
```
and open safari with http://localhost:3000 input username and password with both 'admin' then click "Log in", then click Skip, the browser redirect to login page.
grafana713

logs:
```
t=2020-08-14T10:24:32+0000 lvl=info msg="Successful Login" logger=http.server User=admin@localhost
t=2020-08-14T10:24:37+0000 lvl=info msg="Request Completed" logger=context userId=0 orgId=0 uname= method=GET path=/ status=302 remote_addr=172.17.0.1 time_ms=0 size=29 referer=http://localhost:3000/login
```
Screen Shot 2020-08-14 at 6 25 58 PM

safari login successfully in grafana 6.6.2
Screen Shot 2020-08-14 at 6 30 53 PM
grafana662

We also use tv browser to open grafana pages, these browser worked well with grafana 6.6.2 and they can not be upgraded.

**Environment**:
- Grafana version: 7.1.3
- User OS & Browser: Mac 10.13.6 and Safari 11.1.2
- Grafana plugins: N/A
- Others:

yangliuyu:
The API will not send and set cookie correctly for some old version browsers which not provide default credentials as [same-origin](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials). We can set the defaulty value explicitly. 

Fixes #27007 

### DIFF
--- a/public/app/core/utils/fetch.ts
+++ b/public/app/core/utils/fetch.ts
@@ -7,12 +7,12 @@ export const parseInitFromOptions = (options: BackendSrvRequest): RequestInit =>
   const isAppJson = isContentTypeApplicationJson(headers);
   const body = parseBody(options, isAppJson);
 
-  if (options?.withCredentials) {
+  if (options?.credentials) {
     return {
       method,
       headers,
       body,
-      credentials: 'include',
+      credentials: options.credentials,
     };
   }
 
@@ -20,6 +20,7 @@ export const parseInitFromOptions = (options: BackendSrvRequest): RequestInit =>
     method,
     headers,
     body,
+    credentials: 'same-origin',
   };
 };
 


public/app/core/utils/fetch.ts
```diff
@@ -7,19 +7,20 @@ export const parseInitFromOptions = (options: BackendSrvRequest): RequestInit =>
   const isAppJson = isContentTypeApplicationJson(headers);
   const body = parseBody(options, isAppJson);
 
-  if (options?.withCredentials) {
```
torkelo, line 10: this is a big breaking change, withCredentials is a boolean in the our BackendSrv request api

public/app/core/utils/fetch.ts
```diff
@@ -7,19 +7,20 @@ export const parseInitFromOptions = (options: BackendSrvRequest): RequestInit =>
...
-      credentials: 'include',
+      credentials: options.credentials,
     };
   }
 
   return {
     method,
     headers,
     body,
+    credentials: 'same-origin',
```
torkelo, line 23: can you describe how this impacts https://github.com/grafana/grafana/issues/27007 ? Are you using multiple domains?

hugohaggmark (changes requested):
I think you need to change the merge target to `master`.

public/app/core/utils/fetch.ts
yangliuyu: I think the name convention of `withCredentials` comes from [XMLHttpRequest.withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) property which has true or false value and `credentials` comes from fetch API [Request.credentials](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) property and has `omit`, `include` and `same-origin`. In order to keep the parameter(like the options variable) consistency, it is better to use credentials instead of withCredentials. Since modern browsers with latest version use `same-origin` as the default value, we may explicitly set it(make browser with old version works in the same way). 
The `options.credentials` used in many places in grafana frontend source code, so change the request api directly may not be a good idea. But I recommend use `same-origin` as the default value and use can override by passing [`return await this.request({ method: 'GET', credentials: 'include', url, params, requestId });`](https://github.com/grafana/grafana/blob/1e8b10ef057863b009299f70d0634a65090dcc75/public/app/core/services/backend_srv.ts#L358); @torkelo

public/app/core/utils/fetch.ts
yangliuyu: We use some android browsers running in TV device for visualizing grafana dashboards. These old version browsers do not use `same-origin` as the default value. So after input username and password in the login page, the cookie `grafana_session` is not set correctly and then the browser will redirect to the login page again.

yangliuyu:
> I think you need to change the merge target to `master`.

Hope it can be merged in `v7.1.6` and then merged in upstream master. I think `v7.1.6` will be released based on `v7.1.x` branch. If `v7.1.6` is released based on master, this PR need to be change the merget target to `master` 
@hugohaggmark

yangliuyu:
Found another problem after changing credentials to 'same-origin' https://github.com/grafana/grafana/issues/27007#issuecomment-681692721

hugohaggmark:
@yangliuyu hi again, so we've looked into this PR more and we've come to the conclusion that we need support both `credentials` and `withCredentials` (for a while) and that `credentials` should take precedence. Also this PR needs to point to `master` so we can add cherry-pick needed label and a proper milestone.

Let me know if you need any help to introduce these changes.

public/app/core/utils/fetch.ts
```diff
@@ -7,19 +7,20 @@ export const parseInitFromOptions = (options: BackendSrvRequest): RequestInit =>
...
 
-  if (options?.withCredentials) {
+  if (options?.credentials) {
     return {
       method,
       headers,
       body,
-      credentials: 'include',
+      credentials: options.credentials,
     };
```
hugohaggmark, line 16: This needs to support both options.credentials (a string) and options.withCredentials and withCredentials should have a deprecated notice in our api.

hugohaggmark:
@yangliuyu I'm going to help out getting this PR in a mergable state

hugohaggmark:
Closed in favor of #27385

yangliuyu:
@hugohaggmark thanks a lot

PR #27168, opened 2020-08-24
Prometheus: Fix rate_interval to use min step properly

### LINKED ISSUE #21417: Prometheus data source: Add `$__rate_interval`

zoltanbedi:
## In this pr

- Added a note to Prometheus's rate_interval section that it is available from 7.2
- Fixed getRateIntervalScopedVariable method to use min step properly and also the adjusted interval (for resolution changes)

Fixes #21417

### DIFF
--- a/docs/sources/features/datasources/prometheus.md
+++ b/docs/sources/features/datasources/prometheus.md
@@ -112,7 +112,7 @@ Regex:
 ```
 ### Using `$__rate_interval` variable
 
-> Available from Grafana 7.2 and above
+> **Note:** Available in Grafana 7.2 and above
 
 The `$__rate_interval` variable is meant to be used in the rate function. It is defined as max( `$__interval` + _Scrape interval_, 4 * _Scrape interval_), where _Scrape interval_ is the Min step setting (AKA query_interval, a setting per PromQL query), if any is set, and otherwise the _Scrape interval_ as set in the Prometheus data source (but ignoring any Min interval setting in the panel, because the latter is modified by the resolution setting).
 
--- a/docs/sources/features/datasources/prometheus.md
+++ b/docs/sources/features/datasources/prometheus.md
@@ -112,6 +112,8 @@ Regex:
 ```
 ### Using `$__rate_interval` variable
 
+> Available from Grafana 7.2 and above
+
 The `$__rate_interval` variable is meant to be used in the rate function. It is defined as max( `$__interval` + _Scrape interval_, 4 * _Scrape interval_), where _Scrape interval_ is the Min step setting (AKA query_interval, a setting per PromQL query), if any is set, and otherwise the _Scrape interval_ as set in the Prometheus data source (but ignoring any Min interval setting in the panel, because the latter is modified by the resolution setting).
 
 ### Using variables in queries
--- a/public/app/plugins/datasource/prometheus/datasource.test.ts
+++ b/public/app/plugins/datasource/prometheus/datasource.test.ts
@@ -1620,16 +1620,36 @@ describe('PrometheusDatasource', () => {
 
   describe('The __rate_interval variable', () => {
     it('should be 4 times the scrape interval if interval + scrape interval is lower', () => {
-      const { __rate_interval } = ds.getRateIntervalScopedVariable(23, 23);
+      const { __rate_interval } = ds.getRateIntervalScopedVariable({
+        isMinIntervalSet: false,
+        interval: 23,
+        minInterval: 23,
+      });
       expect(__rate_interval.value).toBe('60s');
     });
     it('should be interval + scrape interval if 4 times the scrape interval is lower', () => {
-      const { __rate_interval } = ds.getRateIntervalScopedVariable(56, 56);
+      const { __rate_interval } = ds.getRateIntervalScopedVariable({
+        isMinIntervalSet: false,
+        interval: 56,
+        minInterval: 56,
+      });
       expect(__rate_interval.value).toBe('71s');
     });
     it('should fall back to 60s if interval is 0', () => {
-      const { __rate_interval } = ds.getRateIntervalScopedVariable(0, 0);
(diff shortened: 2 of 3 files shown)


beorn7:
Will have a look today (or "tonight"… :o).

docs/sources/features/datasources/prometheus.md
```diff
@@ -112,6 +112,8 @@ Regex:
 ```
 ### Using `$__rate_interval` variable
 
+> Available from Grafana 7.2 and above
```
oddlittlebird, line 115:
```suggestion
> **Note:** Available in Grafana 7.2 and above
```

beorn7:
I had an idea how to simplify this. Will create a WIP PR on top of this one.

docs/sources/features/datasources/prometheus.md
zoltanbedi: Thank you :)

beorn7:
#27174 is merged now. @zoltanbedi will add some more sophisticated tests.

public/app/plugins/datasource/prometheus/datasource.test.ts
```diff
@@ -1619,17 +1619,43 @@ describe('PrometheusDatasource', () => {
...
-      const { __rate_interval } = ds.getRateIntervalScopedVariable(56, 56);
-      expect(__rate_interval.value).toBe('71s');
-    });
-    it('should fall back to 60s if interval is 0', () => {
-      const { __rate_interval } = ds.getRateIntervalScopedVariable(0, 0);
-      expect(__rate_interval.value).toBe('60s');
+      ds.createQuery(target, { interval: '5m' } as any, 0, 10080);
+      expect((templateSrv.replace as any).mock.calls[1][1]['__rate_interval'].value).toBe('315s');
+    });
+    it('should fall back to 60s if interval and scrapeInterval is 0', () => {
```
beorn7, line 1636: This is actually not quite accurate now because the interval is not 0. 

I guess it should be: 'should fall back to a scrape interval of 15s if min step is set to 0, resulting in 4*15s = 60s'

beorn7 (changes requested):
Great. I think we are almost there. Now we only need the right `intervalFactor` in the test (see below), and then it all matches my expectations.

public/app/plugins/datasource/prometheus/datasource.test.ts
```diff
@@ -1619,17 +1619,43 @@ describe('PrometheusDatasource', () => {
...
+    });
+    it('should be interval + scrape interval if min step set to 1m and interval is 5m', () => {
+      // For a 7d graph, $__interval is 5m
+      ds.createQuery({ ...target, interval: '1m' }, { interval: '5m' } as any, 0, 10080);
+      expect((templateSrv.replace as any).mock.calls[1][1]['__rate_interval'].value).toBe('360s');
+    });
+    it('should be interval + scrape interval if resolution is set to 1/2 and interval is 10m', () => {
+      // For a 7d graph, $__interval is 10m
+      ds.createQuery({ ...target, intervalFactor: 0.5 }, { interval: '10m' } as any, 0, 10080);
+      expect((templateSrv.replace as any).mock.calls[1][1]['__rate_interval'].value).toBe('615s');
```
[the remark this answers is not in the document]
beorn7: WRT the quote from #21417: "For a 7d graph, $__interval is 10m" → $__interval is actually the final interval, not the interval we inject here in the test (which is pre-adjustment by the factor). Thus, you could alternatively change `{ interval: '10m' }` above to `{ interval: '5m' }`, which should then result in the expected 615s.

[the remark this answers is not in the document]
zoltanbedi: Crap, I've used 2 for the first time but then I check your https://github.com/grafana/grafana/issues/21417#issuecomment-674149117 and it puzzled me 😄

beorn7 (approved):

PR #27174, opened 2020-08-24
Simplify `getRateIntervalScopedVariable`

beorn7:
I don't know yet if this works as intended, but if it does, I think it's much clearer.


beorn7:
It would be nice to have tests that are testing more than just the `getRateIntervalScopedVariable` function. (Which is also true for the orginial PR, but in the case of the latter, `getRateIntervalScopedVariable` had more logic than here.)

It would be cool to have a test similar to those for `$__interval`, see line 1283, `describe('The __interval and __interval_ms template variables', () => {`

PR #27176, opened 2020-08-25
BackendSrv: Use uuid v4 as an identifier instead of Date.now()

### LINKED ISSUE #27173: Grafana sometimes does not emit request for all panels on reload
_Opening the issue for the behaviour when Grafana does not emit http request for all my panels but for some. Only sometimes. It's hard to isolate the issue but I'm working on it._

Let's say there are 8 panels on dashboard. Datasource emits `backendSrv().datasourceRequest(options)` for each of them (8 emits)

7 http requests are emitted.

Missing panel with missing request is populated with data from the next one.

So when http request for panel 4 is missing, it gets data from panel 5.

Currently I'm investigating the issue further. Any ideas appreciated.

**Environment**:
- Grafana version: cd9a773427ff04ef92ef6d896d8a87d151eb2e5d

simPod:

Use UUID instead of `Date.now()` for identifiers. `Date.now()` does not guarantee to return a unique value and on machines with enough power it leads to generating duplicate values because millisecond precision is not significant.

Fixes #27173

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,7 @@
     "@types/react-loadable": "5.5.2",
     "@types/react-virtualized-auto-sizer": "1.0.0",
     "@types/sockjs-client": "^1.1.1",
+    "@types/uuid": "8.3.0",
     "@welldone-software/why-did-you-render": "4.0.6",
     "abortcontroller-polyfill": "1.4.0",
     "angular": "1.6.9",
@@ -286,6 +287,7 @@
     "tether-drop": "https://github.com/torkelo/drop/tarball/master",
     "tinycolor2": "1.4.1",
     "tti-polyfill": "0.2.2",
+    "uuid": "8.3.0",
     "whatwg-fetch": "3.1.0"
   },
   "resolutions": {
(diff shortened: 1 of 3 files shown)


hugohaggmark:
Thanks for contributing @simPod, I'm hesitant to change all those files in this PR. Could we keep the change to `backend_srv` only if that solves your issue? And out of curiosity, does it solve it?

torkelo:
Nice!

simPod:
@hugohaggmark I have split the PR

Seems like it fixes it, yes. Hard to say as it was occurring very rarely but duplicate timestamps were being generated.

package.json
```diff
@@ -215,6 +215,7 @@
     "@types/react-loadable": "5.5.2",
     "@types/react-virtualized-auto-sizer": "1.0.0",
     "@types/sockjs-client": "^1.1.1",
+    "@types/uuid": "^8.3.0",
```
hugohaggmark, line 218: Please use specific versions
```suggestion
    "@types/uuid": "8.3.0",
```

package.json
```diff
@@ -286,6 +287,7 @@
     "tether-drop": "https://github.com/torkelo/drop/tarball/master",
     "tinycolor2": "1.4.1",
     "tti-polyfill": "0.2.2",
+    "uuid": "^8.3.0",
```
hugohaggmark, line 290: Please use specific versions
```suggestion
    "uuid": "8.3.0",
```

hugohaggmark (approved):
Great work with this @simPod ⭐

PR #27178, opened 2020-08-25
Use uuid v4 as an identifier instead of Date.now()

simPod:
Not sure if desired but this is followup of https://github.com/grafana/grafana/pull/27176 

It replaces `Date.now()` ids in full codebase.


torkelo (approved):
LGTM

PR #27181, opened 2020-08-25
Stats: fix admin stats query parameters

AgnesToulet:
Fix number of parameters for AdminStats query

This is a quick fix to fix the failing tests but I think it would be good to refactor this function to use the `SqlBuilder` and avoid this kind of errors in the future.

### DIFF
--- a/pkg/services/sqlstore/stats.go
+++ b/pkg/services/sqlstore/stats.go
@@ -158,7 +158,7 @@ func GetAdminStats(query *models.GetAdminStatsQuery) error {
 		) AS active_sessions`
 
 	var stats models.AdminStats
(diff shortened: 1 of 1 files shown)


sakjur:
I have this in #27182 which adds the test as an integration test so that it'll run on MySQL/Postgres for OSS as well.

sakjur:
> This is a quick fix to fix the failing tests but I think it would be good to refactor this function to use the SqlBuilder and avoid this kind of errors in the future.

I'm not sure I agree. I don't like anything that ties us deeper into a particular SQL framework and I think having a DSL for a DSL makes for confusion more than anything.

The way we should've detected this is that all the SQL tests needs to run for all the databases. We don't do that currently because our SQL testing story is too expensive on MySQL/Postgres. Enterprise happens to do that in the release pipeline because it hasn't been upgraded, which turned out to be fortunate here.

PR #27186, opened 2020-08-25
DataFrame: Data source field naming control that does not use field.config.displayName

### LINKED ISSUE #25212: v7 Stat & Gauge panel not respecting "Display Name" for graphite queries

torkelo:
In v7.0 we added label display name logic where Grafana uses the labels to render the field name as well as some initial smart behavior of removing duplicate info (duplicate metric names or field labels). This smart logic would be a big breaking change for Prometheus and graphite and with no time to do a UI option to control or disable this smart naming we opted to disable the smart behavior by setting field.config.displayName from Prometheus and graphite data sources as setting displayName from data source overrides the smart naming.

However, setting field.config.displayName from data source makes the panel Field tab (field config defaults) to have no effect.

Creating bugs like #25212 for both Graphite and Prometheus

First idea was to have a FieldNaming scheme on the frame, but that was tricky. Think it makes more sense to have on field config but also need some property to override field naming for data sources that have a custom legend naming schemes (like prometheus and graphite). So in order for data sources to be able to return a pre-backed series name I added a new field config option useFrameName bool that is currently always true for graphite and prometheus. With this new bool users can still override the naming using display name (and Gauge visualization won't show name for single series any). 

Need to pursue the short and complete naming schemes in future PR. 

There is a strange old time series title -> useFrameName bool translation that feels off. Might have to change TimeSeries.title to bool but do not know what to name it, as useFrameName is strange in that context. 

Updated to use new displayNameDS property

Fixes #25212

### DIFF
--- a/packages/grafana-data/src/types/dataFrame.ts
+++ b/packages/grafana-data/src/types/dataFrame.ts
@@ -29,11 +29,10 @@ export interface FieldConfig<TOptions extends object = any> {
   displayName?: string;
 
   /**
-   * This can be used by data sources that return time series with labels to indicate that frame name contains the
-   * full series name.  Useful for data sources that have custom legend formats and naming schemes and want to
-   * control naming, but still return labels and allow for user override of name via default displayName option.
+   * This can be used by data sources that return time series with labels that want to control
+   * series naming and disable the default naming scheme.
    */
-  useFrameName?: boolean;
+  displayNameDS?: string;
 
   /**
    * True if data source field supports ad-hoc filters
--- a/packages/grafana-data/src/types/dataFrame.ts
+++ b/packages/grafana-data/src/types/dataFrame.ts
@@ -29,8 +29,8 @@ export interface FieldConfig<TOptions extends object = any> {
   displayName?: string;
 
   /**
-   * This can be used by data sources that return time series with labels that want to control
-   * series naming and disable the default naming scheme.
+   * This can be used by data sources that return and explicit naming structure for values and labels 
+   * When this property is configured, this value is used rather than the default naming strategy.  
    */
   displayNameDS?: string;
 
--- a/packages/grafana-data/src/types/dataFrame.ts
+++ b/packages/grafana-data/src/types/dataFrame.ts
@@ -29,10 +29,10 @@ export interface FieldConfig<TOptions extends object = any> {
   displayName?: string;
 
   /**
-   * This can be used by data sources that return and explicit naming structure for values and labels 
-   * When this property is configured, this value is used rather than the default naming strategy.  
+   * This can be used by data sources that return and explicit naming structure for values and labels
+   * When this property is configured, this value is used rather than the default naming strategy.
    */
-  displayNameDS?: string;
+  displayNameFromDS?: string;
 
   /**
    * True if data source field supports ad-hoc filters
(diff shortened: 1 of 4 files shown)


ryantxu (commented):
~~I would expect to see prometheus and graphite set `useFrameName` -- or am i missing something?~~

Nevermind... they set .title, and that changes the value

ryantxu:
what should happen with transformations?  IIUC, a merge would add multiple columns, but everything would get the same frame name

torkelo:
> what should happen with transformations? IIUC, a merge would add multiple columns, but everything would get the same frame name'

Transformations that merge / combine frames will always use getFieldDisplayTitle so the new frame the fields will have the correct name

mckn (approved):
I think it looks good! Seems to work as described. I am not suuper happy with the name of the `displayNameDS`. 

It feels a bit cryptic for the consumer (and in this case I guess it is the datasource developers). Would `displayNameSuggestion` or `displayNameFromDS` work? hmm, this is a tricky one. Not sure I can suggest any better 🙈

mckn:
> I think it looks good! Seems to work as described. I am not suuper happy with the name of the `displayNameDS`.
> 
> It feels a bit cryptic for the consumer (and in this case I guess it is the datasource developers). Would `displayNameSuggestion` or `displayNameFromDS` work? hmm, this is a tricky one. Not sure I can suggest any better 🙈

My thinking regarding the consumer of the property is that the developers of the datasource will be the ones setting the value all other will consume it via `getFieldDisplayName`.

torkelo:
yea, this will only be used by data source developers. 

Naming is hard

* displayNameDS
* displayNameFromDS
* displayNameSuggestion
* dataDisplayName

packages/grafana-data/src/types/dataFrame.ts
```diff
@@ -23,7 +23,20 @@ export enum FieldType {
...
 export interface FieldConfig<TOptions extends object = any> {
-  displayName?: string; // The display value for this field.  This supports template variables blank is auto
+  /**
+   * The display value for this field.  This supports template variables blank is auto
+   */
+  displayName?: string;
+
+  /**
+   * This can be used by data sources that return time series with labels that want to control
+   * series naming and disable the default naming scheme.
```
ryantxu, line 33:
```suggestion
   * This can be used by data sources that return and explicit naming structure for values and labels 
   * When this property is configured, this value is used rather than the default naming strategy.  
```

ryantxu (approved):
naming is hard!   The only possibly better thing is maybe `displayNameFromDS`

Links2004:
is there a time frame for a release of this fix?
currently most of my dashboards are looking not very good (#27189) ;)

torkelo:
This is currently scheduled for 7.2 (beta next week)

PR #27188, opened 2020-08-25
Grafana-UI: BarGauge docs

Clarity-89:
- Add docs for `BarGauge` component
- Remove `_BarGauge.scss` and replace class matchers with aria-label selector

### DIFF
--- a/e2e/suite1/specs/bar-gauge.spec.ts
+++ b/e2e/suite1/specs/bar-gauge.spec.ts
@@ -11,7 +11,7 @@ e2e.scenario({
     e2e.flows.openDashboard({ uid: 'O6f11TZWk' });
 
     e2e()
-      .get('#panel-6 .bar-gauge__value')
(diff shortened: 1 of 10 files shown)


e2e/suite1/specs/bar-gauge.spec.ts
```diff
@@ -11,7 +11,7 @@ e2e.scenario({
     e2e.flows.openDashboard({ uid: 'O6f11TZWk' });
 
     e2e()
-      .get('#panel-6 .bar-gauge__value')
+      .get('#panel-6 [aria-label^="Bar gauge value"]')
```
dprokop, line 14: Can you use selector instead?

dprokop (approved):

e2e/suite1/specs/bar-gauge.spec.ts
Clarity-89: Hmm, sure, how did I miss that 🤔

PR #27197, opened 2020-08-25
Transformations: enable transformations reorder 

### LINKED ISSUE #25482: Transformations: Enable reordering transformations
* First MVP implementation is with arrow up and down same as Queries (Try to reuse logic). 
* Drag and drop using same library as organize fields transform  (if this is done query rows should use same UX and try share logic)

dprokop:
Closes #25482

todo: 
- [x] Update tests


torkelo:
There is a new draggable icon in unicons! draggabledots

dprokop:
> There is a new draggable icon in unicons! draggabledots

Yes, updated it now. 

@ivanahuckova I've encountered an issue where `xl` icon is rendered smaller than `lg` one, and I cannot figure it out:

Screen Shot 2020-08-26 at 10 59 31

The icon marked red has `lg` size, while the one marked with yellow is `xl`. Any ideas?

ivanahuckova:
> The icon marked red has `lg` size, while the one marked with yellow is `xl`. Any ideas?

Hmmm, that is weird. I am not sure, but let me look at it for a sec if I'll be able to figure out what is going on.

ivanahuckova:
@dprokop I figured it out! We are passing a `styles.draggable` className to Icon in [here](https://github.com/grafana/grafana/blob/master/public/app/core/components/TransformersUI/OrganizeFieldsTransformerEditor.tsx#L171) and that className has left-right padding, which is implemented within the svg and therefore makes svg smaller. 

**with padding:**
image

**without padding: **
image

dprokop:
@ivanahuckova thank you so much, you're the best!

public/app/core/components/QueryOperationRow/QueryOperationRow.tsx
```diff
@@ -65,6 +72,7 @@ export const QueryOperationRow: React.FC<QueryOperationRowProps> = ({
             }}
             aria-label="Query operation row title"
           >
+            {draggable && <Icon name="draggabledots" size="lg" className={styles.dragIcon} />}
```
ivanahuckova, line 75: Super tiny nit pick - I would add a title to the draggable icon to let users know that you are able to reorder transformations by dragging the row.

public/app/features/explore/TimeSyncButton.test.tsx
```diff
@@ -8,10 +8,6 @@ const setup = (isSynced: boolean) => {
 };
```
ivanahuckova, line 8: This doesn't look like it is related, but I like the change and removing of snapshot test. 💙

public/app/core/components/TransformersUI/OrganizeFieldsTransformerEditor.tsx
```diff
@@ -135,7 +135,7 @@ const DraggableFieldName: React.FC<DraggableFieldProps> = ({
         >
           <div className="gf-form gf-form--grow">
             <div className="gf-form-label gf-form-label--justify-left width-30">
-              <i className={cx('fa fa-ellipsis-v', styles.draggable)} />
+              <Icon name="draggabledots" size="xl" className={styles.draggable} />
               <IconButton
```
ivanahuckova, line 139: Based on discussion, I would remove the padding and change to `size="lg"` for consistent sizing.

public/app/core/components/QueryOperationRow/QueryOperationRow.tsx
```diff
@@ -92,6 +118,10 @@ const getQueryOperationRowStyles = stylesFactory((theme: GrafanaTheme) => {
       align-items: center;
       justify-content: space-between;
     `,
+    dragIcon: css`
+      opacity: 0.4;
+      cursor: hand;
+    `,
```
ivanahuckova, line 124: Shouldn't it be `drag` instead of hand?  `cursor: hand;` is probably invalid (couldn't find it in [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor))

public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -63,27 +90,45 @@ export class TransformationsEditor extends React.PureComponent<Props, State> {
...
   }
 
-  onChange(transformations: DataTransformerConfig[]) {
-    this.props.panel.setTransformations(transformations);
+  onChange(transformations: TransformationsEditorTransformation[]) {
     this.setState({ transformations });
+    this.props.panel.setTransformations(transformations.map(t => t.transformation));
   }
 
+  // Transformation uid are stored in a name-X form. name is NOT unique hence we need to parse the ids and increase X
```
ivanahuckova, line 98:
```suggestion
  // Transformation uid are stored in a name-X form. Name is NOT unique hence we need to parse the ids and increase X
```

ivanahuckova (approved):
LGTM @dprokop. 🚀  I have left some comments for minor changes, but other than that looks and works great! This is an awesome update! Great job. 💪

public/app/features/explore/TimeSyncButton.test.tsx
dprokop: Removed, as it failed in a weird way after bumping unicons. So the snapshot didn't make any sense

public/app/core/components/QueryOperationRow/QueryOperationRow.tsx
dprokop: you are right! great catch

public/app/core/components/QueryOperationRow/QueryOperationRow.tsx
```diff
@@ -54,24 +61,43 @@ export const QueryOperationRow: React.FC<QueryOperationRowProps> = ({
...
     });
 
-  return (
+  const rowHeader = (
+    <div className={styles.header}>
+      <HorizontalGroup justify="space-between">
+        <div
+          className={styles.titleWrapper}
+          onClick={() => {
+            setIsContentVisible(!isContentVisible);
```
mckn, line 70: nit: `useCallback` to prevent of creating a new function on each render.

mckn (approved):
me after seeing how good this works

Me after seeing how good this works. Awesome work 🙌

PR #27201, opened 2020-08-25
Explore: Keep typeahead suggestions for exact label matches in Loki q…

### LINKED ISSUE #26661: Loki: Show query autocomplete suggestions for exact matches
When writing a Loki query, the autocomplete/typeahead removes exact matches:

The suggestion is given for incomplete text here:
Screenshot from 2020-07-28 12-29-21

The suggestion is removed for an exact match here: 
Screenshot from 2020-07-28 12-29-38

(Maybe there's a duplicate issue, but I couldn't find it)

Perhaps this was intentional, because if there's an exact match then the user doesn't need help completing the word anymore. However, it would be nice to leave the exact match suggestion because then the user doesn't have to go back to the beginning of the word to place a quotation mark. Plus the suggestions are nice to have as a sanity check that the user has typed in the correct word.

rdubrock:
Exact matches in the Loki query editor typeahead weren't autosuggesting. The root cause is that they were getting filtered out by the slate suggestions plugin [here](https://github.com/grafana/grafana/blob/d1ee1d93c8817d6f09d519131c87511abaadbd9e/packages/grafana-ui/src/slate-plugins/suggestions.tsx#L310)

I didn't want to change the plugin behavior, so I added an extra property to the suggestion items in the loki query editor to get around that filter case. 

Fixes https://github.com/grafana/grafana/issues/26661

### DIFF
--- a/public/app/plugins/datasource/loki/language_provider.ts
+++ b/public/app/plugins/datasource/loki/language_provider.ts
@@ -28,7 +28,7 @@ const HISTORY_COUNT_CUTOFF = 1000 * 60 * 60 * 24; // 24h
 const NS_IN_MS = 1000000;
 export const LABEL_REFRESH_INTERVAL = 1000 * 30; // 30sec
 
-const wrapLabel = (label: string) => ({ label });
+const wrapLabel = (label: string) => ({ label, filterText: `\"${label}\"` });
 export const rangeToParams = (range: AbsoluteTimeRange) => ({ start: range.from * NS_IN_MS, end: range.to * NS_IN_MS });
 
 export type LokiHistoryItem = HistoryItem<LokiQuery>;
@@ -196,13 +196,13 @@ export default class LokiLanguageProvider extends LanguageProvider {
         .take(HISTORY_ITEM_COUNT)
         .map(wrapLabel)
         .map((item: CompletionItem) => addHistoryMetadata(item, history))
-        .value();
+        .value() as unknown;
 
       suggestions.push({
         prefixMatch: true,
         skipSort: true,
         label: 'History',
-        items: historyItems,
+        items: historyItems as CompletionItem[],
       });
     }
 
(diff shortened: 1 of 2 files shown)


kaydelaney:
Made a commit that simplifies some of the typing, but otherwise looks good!

kaydelaney (approved):

Elfo404:
FYI, I was just tinkering with the PR and I think it introduces some confusing behaviour when selecting an item from the autocomplete list. Basically it keeps opening the suggestions popover after clicking on an item (without highlighting it):

recording

rdubrock:
@aocenas My apologies for letting this go stale. Looking at your comment [here](https://github.com/grafana/grafana/issues/26661#issuecomment-684812929) on the issue, I'm thinking I should close this out. Is that correct, or should I work on a fix for the behavior @Elfo404 mentioned?

aocenas:
So my feeling is that this makes sense but the issue mentioned by @Elfo404 should be fixed.
To explain:
1. In case the selecting of the autocomplete option will alter the text in the input the autocomplete item should be show. So if I have `grafana` it is ok to suggest `grafana` if clicking on it will change it to `"grafana"`.
2. We should stop suggesting an item if clicking on it is noop. So if I have `"grafana"` there is no need to suggest `grafana` anymore cause it won't do anything and it will keep suggesting that forever.

So not sure if this is easily fixable or not but this PR adds 1. so if we can just fix 2. I think we can merge this.

rdubrock:
Thanks for bearing with me on this, folks!

public/app/plugins/datasource/loki/language_provider.ts
```diff
@@ -28,7 +28,10 @@ const HISTORY_COUNT_CUTOFF = 1000 * 60 * 60 * 24; // 24h
 const NS_IN_MS = 1000000;
 export const LABEL_REFRESH_INTERVAL = 1000 * 30; // 30sec
 
-const wrapLabel = (label: string) => ({ label });
+// const isQuoted = (label: string) => label;
```
Elfo404, line 31: is this a leftover? :)

Elfo404 (approved):
Awesome! Just a nit on (I guess) a leftover comment, but looks nice!

public/app/plugins/datasource/loki/language_provider.ts
rdubrock: Yup! Whoops.

PR #27203, opened 2020-08-25
Added aria-label to dashboard timezone picker

stevenvachon:
... for E2E tests

### DIFF
--- a/packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerFooter.tsx
+++ b/packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerFooter.tsx
@@ -45,7 +45,7 @@ export const TimePickerFooter: FC<Props> = props => {
   if (isEditing) {
     return (
       <div className={cx(style.container, style.editContainer)}>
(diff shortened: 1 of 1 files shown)


ryantxu (approved):

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerFooter.tsx
```diff
@@ -45,7 +45,7 @@ export const TimePickerFooter: FC<Props> = props => {
   if (isEditing) {
     return (
       <div className={cx(style.container, style.editContainer)}>
-        <div className={style.timeZoneContainer}>
+        <div aria-label="Time zone picker select container" className={style.timeZoneContainer}>
```
Clarity-89, line 48: Shouldn't we keep these in e2e selectors?

stevenvachon: You're right. I wasn't aware that we were importing them into the app. It does make sense to.

PR #27204, opened 2020-08-25
Proposal to include a `tidy check`

joanlopez:

I realized that adding or updating a dependency can result into inconsistent `go.mod` and `go.sum` files. Especially if the developer forgets to run `go mod tidy` when needed.

So, this PR is a proposal to include a `tidy check` in terms of ensuring the uploaded `go.mod` and `go.sum` files are exactly the same as the ones resulting after running the `go mod tidy` command.

Notice that I didn't add it yet into CI pipeline (Drone?), waiting for feedback and for the proper way to do it (build-container?).
It eventually can be added into some kind of pre-commit hook or at least we might let the developer use it, that's why I added it as a new Makefile task.

Thanks!


sakjur (approved):
Looked at this yesterday but forgot to approve 🙈

LGTM!

Let's wait with merging this, would like @aknuds1 to opine on this, and this can wait until he's back 🙂

aknuds1:
@kylebrandt Do you think it makes sense to require that go.mod is tidied?

marefr (approved):
Yes great idea. LGTM (given it works as expected, my bash skills are limited)

kylebrandt:
> Do you think it makes sense to require that go.mod is tidied? I'm not sure if there might be any drawbacks to this.

I like the idea. I generally need to run it anyways if I have used go-spew for some inspection and then removed it. Making sure it is done with each PR means that running go tidy should be more constrained to the PR at hand, and not other past stuff that one is maybe less sure about.

Also:

> It is always good practice to run go mod tidy before committing a go.mod file to version control.

-- https://blog.golang.org/migrating-to-go-modules

I'm not sure of any drawbacks, but I assume we will find out if there are in practice, and can disable it or revert changes if needed.

aknuds1 (approved):
LGTM

aknuds1:
You can add the tidy command to the Drone commands [here](https://github.com/grafana/grafana/blob/master/scripts/lib.star#L248) and for CircleCI (we're still using it for master and release) [here](https://github.com/grafana/grafana/blob/master/.circleci/config.yml#L816).

joanlopez:
Just added the script run to Drone & CircleCI definitions.
Let me know if there's anything else I should add or change, please.

Thanks!! 🙏

marefr:
Related to https://github.com/golang/go/issues/27005

scripts/tidy-check.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -eo pipefail
+
+# Check Go is properly installed and available
+command -v go >/dev/null 2>&1 || { echo 'please install go or use image that has it'; exit 1; }
```
aknuds1, line 5:
```suggestion
# Verify that Go is properly installed and available
command -v go >/dev/null 2>&1 || { echo 'please install Go or use an image that has it'; exit 1; }
```

Makefile
```diff
@@ -95,7 +95,11 @@ golangci-lint: scripts/go/bin/golangci-lint
 		--config ./scripts/go/configs/.golangci.toml \
 		$(GO_FILES)
 
-lint-go: golangci-lint revive revive-strict # Run all code checks for backend.
+tidy-check:
+	@echo "check whether the go.mod and go.sum are consistent"
```
aknuds1, line 99:
```suggestion
	@echo "check whether go.mod and go.sum are consistent"
```

aknuds1 (approved):
LGTM, with some nits.

aknuds1:
Committed my nits, will merge once Drone finishes.

aknuds1:

PR #27205, opened 2020-08-26
CloudWatch: Fix error message returned from tag:GetResources

kichik:

This fixes the log for AWS permission related error that can confuse the user.

The action name is `tag:GetResources`, not `tags:GetResources`. This creates a confusing error message that can have the user add the wrong permission to the IAM role.


wbrowne (approved):

wbrowne:
@kichik Thanks for your contribution to Grafana! 😄

PR #27211, opened 2020-08-26
e2e: Adds error catching to prevent flakiness

hugohaggmark:
On, occasion the Monaco editor will not have the time to be properly unloaded (I think) when we change the tab during the e2e test and then the e2e test fails with the `uncaught:exception`:
        
```log
TypeError: Cannot read property 'getText' of null
  at Object.ai [as getFoldingRanges] (http://localhost:3001/public/build/monaco-json.worker.js:2:215257)  
  at e.getFoldingRanges (http://localhost:3001/public/build/monaco-json.worker.js:2:221188)
  at e.fmr (http://localhost:3001/public/build/monaco-json.worker.js:2:116605)
  at e._handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7414)
  at Object.handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7018)
  at e._handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:5038)
  at e.handleMessage (http://localhost:3001/public/build/monaco-json.worker.js:2:4606)
  at e.onmessage (http://localhost:3001/public/build/monaco-json.worker.js:2:7097)
  at Tt.self.onmessage (http://localhost:3001/public/build/monaco-json.worker.js:2:117109)
```
I'm not able to repro this manually (I'm not fast enough) so I think it should be safe to ignore this error in e2e tests so this is what I've done.

Fixes #

### DIFF
--- a/e2e/suite1/specs/inspect-drawer.spec.ts
+++ b/e2e/suite1/specs/inspect-drawer.spec.ts
@@ -9,8 +9,9 @@ e2e.scenario({
   addScenarioDashBoard: false,
   skipScenario: false,
   scenario: () => {
-    Cypress.on('uncaught:exception', err => {
-      if (err.stack?.indexOf("TypeError: Cannot read property 'getText' of null")) {
+    // @ts-ignore some typing issue
+    e2e().on('uncaught:exception', err => {
+      if (err.stack?.indexOf("TypeError: Cannot read property 'getText' of null") !== -1) {
         // On occasion monaco editor will not have the time to be properly unloaded when we change the tab
         // and then the e2e test fails with the uncaught:exception:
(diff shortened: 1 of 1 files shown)


dprokop:
> I'm not able to repro this manually (I'm not fast enough)

hugohaggmark:
> > I'm not able to repro this manually (I'm not fast enough)
> 
> Not sure this would help, but have you tried Chrome's CPU throttling?

Brilliant suggestion ⭐  and it worked, but I can't see that the Monaco Editor stops working after that error at all so I think we're safe ignoring it in the e2e tests.

dprokop (approved):

PR #27216, opened 2020-08-26
System: Date formating options

### LINKED ISSUE #1459: Localize time_format in graphs

torkelo:
Builds on https://github.com/grafana/grafana/pull/25602 and adds system wide defaults that can be modified by server side settings 

Closes #1459

Moves the system defaults to an object that is mutated on the application start. 

Todo

* [x] Add settings that can be set via ini config
* [x] Changing default date formats impact the time picker input box (so dates are not valid)
* [x] Angular graph tooltip 
* [x] Solve millisecond format automatically
* [x] Add experimental local browser format support 
* [x] Clicking on past ranges results in invalid date
* [x] Fixed parsing of date inputs in time picker in custom formats 
* [x] Update config docs


marefr:
You can convert to draft below reviewers if this not ready for review yet.

packages/grafana-data/src/datetime/formats.ts
```diff
@@ -1,2 +1,68 @@
...
+    weekday: 'ddd',
+    era: 'N',
+    dayPeriod: 'A',
+    timeZoneName: 'Z',
+  };
+
+  return parts.map(part => mapping[part.type] || part.value).join('');
+};
+
+export const defaultDateFormats = new SystemDateFormatsState();
```
torkelo, line 68: This is what other parts of grafana will import and use for default (system) date formats. What's the best name of this variable? defaultDateFormats or systemDateFormats? skip the plural?

torkelo:
Locale browser date formatting is strange, (using Intl.DateTimeFormat) asking for month and day, result in DD/MM,  asking year and month and day is resulting in ISO YYYY-MM-DD (edited) 

Screen Shot 2020-08-26 at 15 25 36

mckn:
Awesome work! Will dig into this one a bit more in detail but before that some reflections when I worked with the TimePicker and time zones. My plan is to open an issue regarding these thoughts but feels like a good time to start sharing it when you have start to look into these parts as well.

I would love to try to move away from using momentjs because it is really hard to get the context of a momentjs value. It could be local time, adjusted to a time zone or UTC. It is not suuper clear to know in what state the momentjs date currently is in.

I think things would be a lot simpler if we start parsing dateTime values from the given time zone (same as today) but we always parse it to an UTC based type (unix timestamp or Date type). Use those formats internally, when working with the dates, until you want to present it. Then we simply could apply formatting with the correct time zone.

There are multiple libraries to help out working with UTC based dates, since that could be less then optimal with the native Date type. One example is the: https://date-fns.org/.

conf/defaults.ini
```diff
@@ -773,3 +773,17 @@ license_path =
...
+# Default date format
+date_format = YYYY-MM-DD HH:mm:ss
+# Used by graph and other places where we only show small intervals
+date_format_interval_PT1S = HH:mm:ss
+date_format_interval_PT1M = HH:mm
+date_format_interval_PT1H = MM-DD HH:mm
+date_format_interval_PT1D = YYYY-MM-DD
+date_format_interval_P1YT = YYYY-MM
+# Experimental feature
+date_format_use_browser_locale = false
```
marefr, line 787: Should this be in `feature_toggles` instead?

conf/defaults.ini
```diff
@@ -773,3 +773,17 @@ license_path =
 [feature_toggles]
 # enable features, separated by spaces
 enable =
+
+[preferences]
+# Default date format
+date_format = YYYY-MM-DD HH:mm:ss
+# Used by graph and other places where we only show small intervals
+date_format_interval_PT1S = HH:mm:ss
```
marefr, line 781: What does `PT` stand for? I think we're trying to only use lower case characters for settings.

conf/defaults.ini
```diff
@@ -773,3 +773,17 @@ license_path =
 [feature_toggles]
 # enable features, separated by spaces
 enable =
+
+[preferences]
```
marefr, line 777: Agree this makes most sense, but if we're going this way we should add `default_theme` here as well with a fallback to legacy setting to make it consistent.

marefr (commented):
Left some comments on backend changes. Looks good in general

conf/defaults.ini
torkelo: They are standard ISO time intervals, and are uppercase in the standard: https://en.wikipedia.org/wiki/ISO_8601#Time_intervals

conf/defaults.ini
torkelo: this is not a feature toggle but an option, not sure we every turn this on to true by default, depends on how well it works and what users prefer

conf/defaults.ini
torkelo: yes, agree.

conf/defaults.ini
marefr: Oh right :+1:

conf/defaults.ini
marefr: Got it. But all other settings is lowercase so think consistency is more important. You can still refer to standard ISO time intervals in descriptions/docs. HTTP and OAuth is http and oauth in our settings as an example. 

The alternative I had in mind was to postfix `date_format_interval_` with second, minute, hour, day, month, year etc but no biggie.

torkelo: yea, that sounds better

packages/grafana-data/src/datetime/formats.ts
mckn: I think both works but we started to use the `system` naming when we migrated the global template variables to react. So maybe it would be a good idea to start using `system` as these kind of things?

packages/grafana-ui/src/components/Logs/LogRow.tsx
```diff
@@ -127,6 +128,13 @@ class UnThemedLogRow extends PureComponent<Props, State> {
     });
   };
 
+  renderTimeStamp(epochMs: number) {
+    return dateTimeFormat(epochMs, {
+      timeZone: this.props.timeZone,
+      format: defaultDateFormats.fullDate,
```
mckn, line 134: We don't need to specify this here. It will use the `fullDate` as default - right? But maybe it is good to be explicit.

mckn:
@torkelo how do you trigger the "Clicking on past ranges results in invalid date" to happen? It seems to work for me when doing some quick testing.

mckn:
Overall I think this looks great! Well done 🙌

conf/defaults.ini
torkelo: changed this section to be only about date formating

conf/defaults.ini
torkelo: changed the interval names to be just seconds, minutes, days, months, years

conf/sample.ini
```diff
@@ -794,3 +794,20 @@
...
+# Used by graph and other places where we only show small intervals
+interval_second = HH:mm:ss
+interval_minute = HH:mm
+interval_hour = MM/DD HH:mm
+interval_day = MM/DD
+interval_month = YYYY-MM
+interval_year = YYYY
+
+# Experimental feature
+use_browser_locale = false
```
marefr, line 813: Should be commented in sample.ini

```suggestion
;full_date = YYYY-MM-DD HH:mm:ss

# Used by graph and other places where we only show small intervals
;interval_second = HH:mm:ss
;interval_minute = HH:mm
;interval_hour = MM/DD HH:mm
;interval_day = MM/DD
;interval_month = YYYY-MM
;interval_year = YYYY

# Experimental feature
;use_browser_locale = false
```

pkg/setting/date_formats.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (cfg *Cfg) readDateFormats() {
+	preferences := cfg.Raw.Section("date_formats")
+	cfg.DateFormats.FullDate, _ = valueAsString(preferences, "full_date", "YYYY-MM-DD HH:mm:ss")
+	cfg.DateFormats.Interval.Second, _ = valueAsString(preferences, "interval_second", "HH:mm:ss")
+	cfg.DateFormats.Interval.Minute, _ = valueAsString(preferences, "interval_minute", "HH:mm")
+	cfg.DateFormats.Interval.Hour, _ = valueAsString(preferences, "interval_hour", "MM-DD HH:mm")
+	cfg.DateFormats.Interval.Day, _ = valueAsString(preferences, "interval_day", "YYYY-MM-DD")
+	cfg.DateFormats.Interval.Month, _ = valueAsString(preferences, "interval_month", "YYYY-MM")
```
marefr, line 25: Need to handle possible returned error values

pkg/setting/date_formats.go
```diff
@@ -0,0 +1,28 @@
...
+	Second string `json:"second"`
+	Minute string `json:"minute"`
+	Hour   string `json:"hour"`
+	Day    string `json:"day"`
+	Month  string `json:"month"`
+	Year   string `json:"year"`
+}
+
+func (cfg *Cfg) readDateFormats() {
+	preferences := cfg.Raw.Section("date_formats")
```
marefr, line 19: Nit. Could rename preferences to something "date formats section".

marefr (changes requested):
Great! Just a couple minor things left, see suggestions and comments.

pkg/setting/date_formats.go
torkelo: This cannot error, we do not do that in many places

torkelo: valueAsString function looks pretty bad, and the panic recover? why not just check if key exists? settings.go needs a lot of refactoring, cannot have error handling duplicated for every key that seems really  bad

conf/sample.ini
torkelo: thanks for spotting this!

pkg/setting/date_formats.go
torkelo: fixed

PR #27221, opened 2020-08-26
UI/ClickOutsideWrapper: Adds eventType prop for situations where event should be on document

kaydelaney:
This PR adds a `parent` prop to `ClickOutsideWrapper`, with a default of `window`, that can be overridden if the event should be on `document`, as in the case with `PanelHeader`.
This PR should also resolve issues with the dashboard variable dropdown not behaving properly.

### DIFF
--- a/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
+++ b/packages/grafana-ui/src/components/ClickOutsideWrapper/ClickOutsideWrapper.tsx
@@ -10,6 +10,7 @@ export interface Props {
    *  Runs the 'onClick' function when pressing a key outside of the current element. Defaults to true.
    */
   includeButtonPress: boolean;
+  parent: Window | Document;
 }
 
 interface State {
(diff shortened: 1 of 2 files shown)


hugohaggmark (changes requested):
Now when you click on a value in the OptionPicker it closes the options. It should remain opened for multi valued variables.

hugohaggmark:
> Now when you click on a value in the OptionPicker it closes the options. It should remain opened for multi valued variables.

This error is because we've switched from `window` to `document` in ClickOutsideWrapper not because of the latest changes in this PR :(

kaydelaney:
@hugohaggmark Ack. 😦 It's actually hard to find a straight answer about what the difference between events attached on `document` or `window` is - lots of conflicting info. I suppose it's better to revert to using `window` then and to figure another solution out for the lingering panel menus...

hugohaggmark:
> @hugohaggmark Ack. 😦 It's actually hard to find a straight answer about what the difference between events attached on `document` or `window` is - lots of conflicting info. I suppose it's better to revert to using `window` then and to figure another solution out for the lingering panel menus...

yeah, it could probably differ between browsers too? Maybe you could add a `parent` prop that defaults to `window` and pass `document` to solve the PanelHeader issue?

hugohaggmark (approved):
Looks good and works as expected! ⭐ 

Tested the following:
- LogRowContext
- FuncEditor in Graphite
- OptionsPicker
- TimePickerCalendar
- TimeRangePicker
- PanelHeader

PR #27232, opened 2020-08-26
Dashboard: Ignore cancelled api request error when loading dashboard

torkelo:
After backendSrv update cancellations now cause promise rejections


hugohaggmark (approved):
Good catch! This looks good to me.

PR #27249, opened 2020-08-27
API: send Login actions

AgnesToulet:
Send Login actions to be handled in Enterprise.

**Special notes for your reviewer:**
Some updates have been made to the `AuthenticateUser` function to have more information in the logs. I updated the authentication middleware function and the login function to keep the same behavior as before.


pkg/api/login_oauth.go
```diff
@@ -256,3 +274,31 @@ func hashStatecode(code, seed string) string {
 	hashBytes := sha256.Sum256([]byte(code + setting.SecretKey + seed))
 	return hex.EncodeToString(hashBytes[:])
 }
+
+func (hs *HTTPServer) handleOAuthLoginError(ctx *models.ReqContext, action string, status int, errMsg string, err error) {
```
sakjur, line 278: I find it hard to follow when there are more than ~3 arguments to a function, this structure could be replaced by something like:

```suggestion
type loginError struct {
	action string
	httpStatus int
	publicMessage string
	err error
}

func (hs *HTTPServer) handleOAuthLoginError(ctx *models.ReqContext, loginError loginError) {
```

pkg/login/auth.go
```diff
@@ -63,10 +65,6 @@ func AuthenticateUser(query *models.LoginUserQuery) error {
 		return ErrInvalidCredentials
 	}
 
-	if err == models.ErrUserNotFound {
-		return ErrInvalidCredentials
```
sakjur, line 67: Random thought: If we'd have a custom `AuthenticationError` type we would be able to send metadata about what's wrong that isn't part of the `.Error()` 🤔 That would allow for making sure that any and all authentication errors give a vague message outwards while retaining full legibility internally.

pkg/api/login.go
```diff
@@ -283,3 +323,11 @@ func (hs *HTTPServer) RedirectResponseWithError(ctx *models.ReqContext, err erro
 
 	return Redirect(setting.AppSubUrl + "/login")
 }
+
+func (hs *HTTPServer) SendLoginLog(cmd *models.SendLoginLogCommand) {
+	if err := bus.Dispatch(cmd); err != nil {
+		if err != bus.ErrHandlerNotFound {
+			hs.log.Warn("Error while sending login log", "err", err)
+		}
```
sakjur, line 331: Won't this complain for everything all the time when you're running OSS?

AgnesToulet: No, the error will be `bus.ErrHandlerNotFound` so we won't log it.

pkg/api/login_oauth.go
```diff
@@ -306,42 +306,42 @@ func hashStatecode(code, seed string) string {
...
+	HttpStatus    int
+	PublicMessage string
+	Err           error
 }
 
-type loginInformation struct {
-	action      string
-	user        *models.User
-	extUserInfo *models.ExternalUserInfo
+type LoginInformation struct {
```
sakjur, line 315:
```suggestion
type loginInformation struct {
```

Only the inner fields have to be public 🙂

AgnesToulet: I need to access it from the `saml` package as I do pretty much the same there.

marefr:
Would be nice to plug this in as some kind of middleware instead, but guessing that require a lot of additional work? Don't know how much of the codebase you're planning to cover, but if it's quite a lot more I would definitely consider a pluggable solution to cover this cross-cutting concern.

sakjur:
@marefr The other part of the auditing is added as a global middleware, for login specifically we need some extra information that isn't accessible for a middleware with the current architecture.

marefr:
@sakjur great :+1:

pkg/api/login.go
```diff
@@ -159,8 +161,16 @@ func (hs *HTTPServer) LoginAPIPing(c *models.ReqContext) Response {
...
+	action := "login"
 	if setting.DisableLoginForm {
-		return Error(401, "Login is disabled", nil)
+		errMsg := "Login is disabled"
+		hs.SendLoginLog(&models.SendLoginLogCommand{
+			ReqContext: c,
+			LogAction:  action,
+			HTTPStatus: http.StatusUnauthorized,
+			Error:      errors.New(errMsg),
+		})
```
sakjur, line 172: Would it make sense to use defer to send this rather than defining it at all the exit points? 🤔

If we could pick up the status and error from the response object being returned instead of having to duplicate it everywhere, that'd reduce the complexity of the rest of the logic in this function quite a bit.

pkg/api/login_oauth.go
```diff
@@ -232,6 +274,13 @@ func syncUser(ctx *models.ReqContext, token *oauth2.Token, userInfo *social.Basi
 		}
 	}
 
+	return extUser
+}
+
+// syncUser syncs a Grafana user profile with the corresponding OAuth profile.
+func syncUser(ctx *models.ReqContext, extUser *models.ExternalUserInfo,
+	connect social.SocialConnector) (*models.User, error) {
```
sakjur, line 282: Preference: When you need multiple lines for function declarations and calls in Go, it's quite common to put each variable on it's own line.

```suggestion
func syncUser(
	ctx *models.ReqContext,
	extUser *models.ExternalUserInfo,
	connect social.SocialConnector,
) (*models.User, error) {
```

pkg/login/auth_test.go
```diff
@@ -27,6 +27,7 @@ func TestAuthenticateUser(t *testing.T) {
 				So(sc.grafanaLoginWasCalled, ShouldBeFalse)
 				So(sc.ldapLoginWasCalled, ShouldBeFalse)
 				So(err, ShouldEqual, ErrPasswordEmpty)
+				So(sc.loginUserQuery.AuthModule, ShouldEqual, "")
```
sakjur, line 30: Is it important for this particular test to test for the value of the auth module?

sakjur (commented):
:+1: I think this is starting to look quite finished, I'm a little unsure about the LoginPost function, it feels like that could benefit from being slightly refactored.

I'd also like to test running this before we merge 🙂

pkg/api/login.go
AgnesToulet: Good point! I searched for a better way of doing this but didn't think about defer.

pkg/login/auth_test.go
AgnesToulet: I think it helps to understand the behavior of the function for this field.

xlson (approved):
Great work Agnes! The only thing that comes to find is that I think we should be asynchronous on the service side rather than in the calling code.

pkg/api/login.go
```diff
@@ -159,8 +161,27 @@ func (hs *HTTPServer) LoginAPIPing(c *models.ReqContext) Response {
 }
 
 func (hs *HTTPServer) LoginPost(c *models.ReqContext, cmd dtos.LoginCommand) Response {
+	action := "login"
+	var user *models.User
+	var response *NormalResponse
+
+	defer func() {
```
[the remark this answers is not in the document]
AgnesToulet: I agree with you that we want the audit logging to be asynchronous but it's not really what `defer` is doing and I used it for clarity mostly. (Thanks to [Emil's comment](https://github.com/grafana/grafana/pull/27249#discussion_r482098163))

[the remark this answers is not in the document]
xlson: Sorry, I misread the code. Then it makes perfect sense.

Issue #27222, opened 2020-08-26
Migrate core datasources to fetch API


dprokop:
Core data sources to migrate to observables:
- [x] Google Cloud Monitoring (#28847) 
- [x] Cloudwatch (#28865) 
- [x] Elastic (#30009)
- [x] Graphite (#31928)
- [x] InfluxDB (https://github.com/grafana/grafana/pull/28555)
- [x] Jaeger (#31969)
- [x] Loki (#27265)
- [x] MSSQL (#28809)
- [x] MySQL (#27649)
- [x] OpenTSDB (#29992)
- [x] Postgres (#27478)
- [x] Zipkin (#30829)
- [x] Prometheus (#27090)
- [x] Tempo (#31618)

**Why is this needed**:

[datasourceRequest](https://grafana.com/docs/grafana/latest/packages_api/runtime/backendsrv/#datasourcerequest-method) method is now deprecated in favour of [fetch](https://github.com/grafana/grafana/blob/master/public/app/core/services/backend_srv.ts#L67) method. This was introduced in https://github.com/grafana/grafana/pull/26043

For some inspiration please take a look at [Test Datasource](https://github.com/grafana/grafana/blob/64bc85963b2b2afe86ad3c61510b9b1e41e9fb7a/public/app/plugins/datasource/testdata/datasource.ts) or @hugohaggmark [PR migrating prometheus data source](https://github.com/grafana/grafana/pull/27090)

mckn:
@zoltanbedi are you working on migrating the Jaeger datasource to the fetch API?

zoltanbedi:
@mckn not right now but can do that next week.

mckn:
> @mckn not right now but can do that next week.

Ok ok, it is up to you :) If you don't have the time I can give it a shot at the end of next week. Just wanted to check so we don't work on the same thing independently.

mckn:
Done and done

PR #27265, opened 2020-08-28
Loki: Support request cancellation properly (Uses new backendSrv.fetch Observable request API) 

hugohaggmark:
Makes the Loki data source use the new fetch api. Also did a lot of test refactoring and added a small `observableTester` utility function for simpler observable subscribe and expect. Would love to look into using `rxjs/testing` and `TestScheduler` in particular but leaving that for another day :)

Relates #27222

More work will have to be done when we move the metricFindQuery etc to Observables.

### DIFF
--- a/public/app/plugins/datasource/loki/datasource.ts
+++ b/public/app/plugins/datasource/loki/datasource.ts
@@ -1,41 +1,44 @@
 // Libraries
-import { isEmpty, map as lodashMap, cloneDeep } from 'lodash';
-import { Observable, from, merge, of } from 'rxjs';
-import { map, catchError, switchMap } from 'rxjs/operators';
+import { cloneDeep, isEmpty, map as lodashMap } from 'lodash';
+import { merge, Observable, of } from 'rxjs';
+import { catchError, map, switchMap } from 'rxjs/operators';
 
 // Services & Utils
-import { DataFrame, dateMath, FieldCache, QueryResultMeta, TimeRange } from '@grafana/data';
-import { getBackendSrv, BackendSrvRequest, FetchError } from '@grafana/runtime';
-import { addLabelToQuery } from 'app/plugins/datasource/prometheus/add_label_to_query';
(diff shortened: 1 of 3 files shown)


public/test/helpers/observableTester.ts
```diff
@@ -0,0 +1,68 @@
...
+
+interface SubscribeAndExpectOnComplete<T> extends ObservableTester<T> {
+  expect: () => void;
+}
+
+interface SubscribeAndExpectOnError<T> extends ObservableTester<T> {
+  expect: (err: any) => void;
+}
+
+export const observableTester = () => {
```
dprokop, line 20: Why not making `observableTester` an object instead of a function? Are you expecting it to be configurable at some point?

dprokop (approved):
Goodie! One question, other than that it's 👍

public/test/helpers/observableTester.ts
hugohaggmark: No specific reason, just a pattern I use, I suppose. Makes it immutable, too I guess.

zoltanbedi (approved):
lgtm. Love the import order fixes 😍

hugohaggmark:
> lgtm. Love the import order fixes 😍

thank you webstorm & saveactions 🙏

public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -1,41 +1,44 @@
 // Libraries
-import { isEmpty, map as lodashMap, cloneDeep } from 'lodash';
-import { Observable, from, merge, of } from 'rxjs';
-import { map, catchError, switchMap } from 'rxjs/operators';
+import { cloneDeep, isEmpty, map as lodashMap } from 'lodash';
+import { merge, Observable, of } from 'rxjs';
+import { catchError, map, switchMap } from 'rxjs/operators';
 
 // Services & Utils
```
zoltanbedi, line 6:
```suggestion
```
Maybe no need for this comment after the import order fix.

Issue #25513, opened 2020-06-10
Storybook: Documenting @grafana/ui components


tskarhed:
**What would you like to be added**:
Tons of documentation.

**Why is this needed**:
So people understand how to use @grafana/ui.

**How**
Please see [the guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/storybook.md#writing-mdx-documentation) on writing MDX docs.

## Undocumented stories/components

- [x] Alert
- [x] AlphaNotice
- [x] Badge
- [x] BarGauge
- [x] BigValue
- [x] Button
- [ ] ButtonCascader – details in [#27308](https://github.com/grafana/grafana/issues/27308)
- [x] Cascader
- [x] ClickOutsideWrapper
- [x] ClipboardButton https://github.com/grafana/grafana/pull/27451
- [x] Collapse - https://github.com/grafana/grafana/pull/27291
- [x] ColorPicker
- [x] ConfirmButton https://github.com/grafana/grafana/pull/27477
- [x] ConfirmModal
- [x] DataSourceSettings
- [x] Drawer #27267
- [x] FileUpload #27393
- [x] FormField https://github.com/grafana/grafana/issues/27596
- [x] FormLabel https://github.com/grafana/grafana/issues/27596
- [x] FormattedValueDisplay https://github.com/grafana/grafana/pull/27759
- [x] Forms
- [ ] Gauge
- [ ] Graph
- [ ] Icon – see details in [#27183](https://github.com/grafana/grafana/issues/27183)
- [x] IconButton
- [ ] InfoBox
- [ ] InfoTooltip
- [ ] Input
- [ ] JSONFormatter
- [x] Label [#27350](https://github.com/grafana/grafana/pull/27350)
- [ ] Layout
- [ ] Legend
- [ ] List
- [ ] LoadingPlaceholder
- [ ] Logs
- [ ] MatchersUI
- [ ] Modal
- [ ] Monaco
- [ ] OptionsUI
- [ ] Pagination
- [ ] PanelOptionsGrid
- [ ] PanelOptionsGroup
- [ ] PieChart
- [ ] Portal
- [ ] QueryField
- [ ] RefreshPicker
- [ ] SecretFormField
- [ ] Segment
- [ ] Select
- [ ] SetInterval
- [ ] SingleStatShared
- [x] Slider
- [ ] Spinner
- [ ] StatsPicker
- [ ] Switch
- [ ] TabbedContainer
- [ ] Table
[...]

PR #27267, opened 2020-08-28
Docs: Drawer component

peterholmberg:
Adds documentation and example for the Drawer component.

Part of #25513

Would love feedback on how to expand the docs, struggled to come up with more information.

### DIFF
--- a/packages/grafana-ui/src/components/Drawer/Drawer.mdx
+++ b/packages/grafana-ui/src/components/Drawer/Drawer.mdx
@@ -8,7 +8,7 @@ import { Drawer } from './Drawer';
 Drawer is a slide in overlay that can be used to display additional information without hiding the main page content. It is usually anchored to the left or right edge of the screen.
 
 
-One example of the Drawer in use is in Grafanas [panel inspector](https://grafana.com/docs/grafana/latest/panels/inspect-panel/).
+One example of the Drawer in use is in Grafana's [panel inspector](https://grafana.com/docs/grafana/latest/panels/inspect-panel/).
 
 ### Usage
 ```tsx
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/Drawer/Drawer.mdx
```diff
@@ -1,4 +1,35 @@
-import { Props } from '@storybook/addon-docs/blocks';
+import { Meta, Props } from '@storybook/addon-docs/blocks';
 import { Drawer } from './Drawer';
 
+<Meta title="MDX|Drawer" component={Drawer} />
+
+# Drawer
+### When to use
+The Drawer component is a slide in overlay that can be used to display information without
+hiding content.
```
Clarity-89, line 9:
```suggestion
Drawer is a slide in overlay that can be used to display additional information without hiding the main page content. It is usually anchored to the left or right edge of the screen.

```

packages/grafana-ui/src/components/Drawer/Drawer.mdx
```diff
@@ -1,4 +1,35 @@
...
+
+One example of the Drawer in use is in Grafanas [panel inspector](https://grafana.com/docs/grafana/latest/panels/inspect-panel/).
```
Clarity-89, line 11:
```suggestion
One example of the Drawer in use is in Grafana's [panel inspector](https://grafana.com/docs/grafana/latest/panels/inspect-panel/).
```

Clarity-89 (approved):
Looks good! Added some suggestions for the wording.

jessover9000 (commented):
Cool! About the "When to use" part: Saying "It is usually anchored to the left or right edge of the screen" makes me wonder whether there's any other alternatives. Can the drawer component do top or bottom drawers/are there examples for that? (Like query history for example?) If not, we shouldn't say "is usually anchored" but should make the phrasing exclusive: "It can be anchored either to the left or right edge …". I would also add that the drawer cannot only display information but also contain interactive elements, or rather, it can contain pretty much anything you want right? 🤓 
And one thing not directly related to the docs: looking at the storybook canvas makes me wonder why that "open drawer" button is styled so weirdly. couldn't we put a proper Grafana button style there?

PR #27273, opened 2020-08-28
AzureMonitor: select plugin route from cloudname

### LINKED ISSUE #27243: Can not get resource metics data from Azure China
**What happened**: All resource metric data are NULL.

**What you expected to happen**: show the graph of resource metric

**How to reproduce it (as minimally and precisely as possible)**: configurate azure monitor for Azure china 

**Environment**:
- Grafana version: 7.0+
- Data source type & version: Azure monitor for Azure China
- User OS & Browser: MACos & Chrome
- Grafana plugins: N/A
- Others:
image

kylebrandt:
For the Azure Monitor service the Azure Monitor data source, selects the plugin route.

maybe fixes #27243

**Notes for Reviewer**
https://github.com/grafana/grafana/issues/27243#issuecomment-682910259 explains this

### DIFF
--- a/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
+++ b/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
@@ -226,7 +226,7 @@ func (e *AzureMonitorDatasource) createRequest(ctx context.Context, dsInfo *mode
 	cloudName := dsInfo.JsonData.Get("cloudName").MustString("azuremonitor")
 	var azureMonitorRoute *plugins.AppPluginRoute
 	for _, route := range plugin.Routes {
-		if strings.Contains(route.Path, cloudName) {
+		if route.Path == cloudName {
 			azureMonitorRoute = route
 			break
 		}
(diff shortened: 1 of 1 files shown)


pkg/tsdb/azuremonitor/azuremonitor-datasource.go
```diff
@@ -223,15 +223,15 @@ func (e *AzureMonitorDatasource) createRequest(ctx context.Context, dsInfo *mode
 		return nil, errors.New("Unable to find datasource plugin Azure Monitor")
 	}
 
+	cloudName := dsInfo.JsonData.Get("cloudName").MustString("azuremonitor")
 	var azureMonitorRoute *plugins.AppPluginRoute
 	for _, route := range plugin.Routes {
-		if route.Path == "azuremonitor" {
+		if strings.Contains(route.Path, cloudName) {
```
marefr, line 229: Shouldn't it still be equals here?

marefr (commented):
Left a comment

pkg/tsdb/azuremonitor/azuremonitor-datasource.go
kylebrandt: Yes, I think somehow I confused myself when looking at the plugin json and where cloudName comes from in the context of the different services.

marefr (approved):
LGTM

Given the referenced issue is making a request to the backend and not going through the data proxy this should make it work better :+1:

PR #27235, opened 2020-08-26
@grafana/e2e: selector improvements

stevenvachon:
See commit message(s).

PR #27274, opened 2020-08-28
@grafana/e2e-selectors: fix to qualify as not aria-label

stevenvachon:
I wrote it wrong in #27235. The magic syntax for avoiding the default `aria-label` was not obvious to me. The mapped value must be a function with no argument.

Related: https://github.com/grafana/grafana/issues/25166


ryantxu (approved):
can you add more context for why next time?  i assume this helps a test in a different repo?

stevenvachon:
Added description

PR #27276, opened 2020-08-29
Do not calculate min and max globally

simPod:
_According to the code the behaviour seemed intentional but I decided to open this up anyway._

Gauges are currently computed based on all values in the table. I thought that Gauge should rather be computed from values in a column where it is placed.

That said having

```
+---+---+
| A | B |
+---+---+
| 1 | 4 |
+---+---+
| 2 | 5 |
+---+---+
| 3 | 6 |
+---+---+
```
Gauge in column `A` should have min 1 and max 3.  
Gauge in column `B` should have min 4 and max 6.

Currently, all Gauges have min 1 and max 6.

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.test.ts
+++ b/packages/grafana-data/src/field/displayProcessor.test.ts
@@ -44,7 +44,7 @@ describe('Process simple display values', () => {
     getDisplayProcessor(),
 
     // Add a simple option that is not used (uses a different base class)
-    getDisplayProcessorFromConfig({ min: 0, max: 100 }),
+    getDisplayProcessorFromConfig({ globalMin: 0, globalMax: 100 }),
 
     // Add a simple option that is not used (uses a different base class)
     getDisplayProcessorFromConfig({ unit: 'locale' }),
--- a/packages/grafana-data/src/field/fieldDisplay.test.ts
+++ b/packages/grafana-data/src/field/fieldDisplay.test.ts
@@ -146,8 +146,8 @@ describe('FieldDisplay', () => {
     });
 
     const display = getFieldDisplayValues(options);
-    expect(display[0].field.min).toEqual(0);
-    expect(display[0].field.max).toEqual(0);
+    expect(display[0].field.globalMin).toEqual(0);
+    expect(display[0].field.globalMax).toEqual(0);
   });
 
   describe('Value mapping', () => {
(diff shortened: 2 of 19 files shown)


hugohaggmark:
Not very familiar with this domain so maybe @ryantxu or @dprokop could have a look?

dprokop:
@simPod Can't recall precisely what the reasoning for global min/max calculation was(probably @ryantxu or @torkelo can shed some light on this), but my gut feeling is to provide consistent range context for visualizations.

Changing this feels like a potential breaking change to me. Maybe instead of altering this behavior, we could introduce `fieldMin/fieldMax` config properties? Let's wait until @ryantxu or @torkelo will provide more context behind the current implementation decision.

ryantxu:
The purpose of `findNumericFieldMinMax` is to find the global min/max -- if you want the min/max for a colum... just use the reducer no?

In early versions of gauges panels we just used min/max for each column,  BUT all feedback was that people expected "auto" min/max should be across the whole result, not a single column.  Obviously it depends :) but I'm not sure how to best toggle the behavior

simPod:
Ah 🤔  That sounds really weird TBH, but if that's expected among other users, well...

I though the use case for gauge was more general. Eg. I have 3 columns with different values, have gauge in each of them in order to visualise value ratio per column, not globally.

With current behaviour for table bellow we get gauge between 15 and 2020 for temperature column, seems weird.

```
+-------------+------+
| Temperature | Year |
+-------------+------+
| 15          | 2018 |
+-------------+------+
| 18          | 2019 |
+-------------+------+
| 20          | 2020 |
+-------------+------+
```

torkelo:
I think the most intuitive behavior here is very different for different visualizations and data sets. For time series having the min / max be based on all series makes a lot more sense as it would be impossible to compare different gauges otherwise if they did not use the same min / max. 

But for table where each column is a isolated thing it makes less sense

simPod:
What about renaming it to globalMin and globalMax so it's more clear and for table gauge use values from FieldCalc?

torkelo:
Can you create an issue to document the problem with global min max, and the different solution proposals. 

Currently, we do not have any sense of how much a problem this is to warrant new standard field options. 

Closing this for now as this solution does not make sense. the min max are not just globalMin globalMax as these can be defined manually per field as well

PR #8307, opened 2017-05-05
Updated cloudwatch plugin to allow specific tag selection

paco3346:
Tags come back from AWS as a key value pair inside an array
This array is now converted to an object
Tags can be selected when using the 'ec2_instance_attribute' query
Example: `ec2_instance_attribute(us-east-1, Tags.Name, { "tag:Grafana": [ "true" ] })`

Fixes #7294

Issue #7294, opened 2017-01-18
[Bug] [AWS Cloudwatch] It is not possible to get a single tag value by its key


chrootLogin:
It seems that it is not possible that you can get single tag values from AWS cloudwatch.

Please include this information:
- What Grafana version are you using?
4.2
- What datasource are you using?
CloudWatch and Graphite
- What OS are you running grafana on?
Docker on CentOS
- What did you do?
I tried to create a template variable with the content of the EC2 instance name tag. I tried different queries but it didn't work. The only thing I found out was that I can get the indexes by their id, as example:
```
ec2_instance_attribute(eu-central-1, Tags[0].Value, {"instance-id": ["$instanceid"]})
ec2_instance_attribute(eu-central-1, Tags[1], {"instance-id": ["$instanceid"]})
```
But the order isn't the same all the time, so this is no solution. 

- What was the expected result?
I can get the value from an single ec2 instance tag by the name of the key. I want something like this:
```
ec2_instance_attribute(eu-central-1, Tags[Name].Value, {"instance-id": ["$instanceid"]})
```

- What happened instead?
I can only get all tags of an EC2 instance as JSON object, but I'm not able to get only one tag.
In the end I want the value of the tag with the key "Name", in the following example "models-int-p".

## Some screenshots
![bildschirmfoto_2017-01-18_17-40-46](https://cloud.githubusercontent.com/assets/4862947/22075245/83bc6aca-ddab-11e6-925f-f621a3e2bf61.png)
The identifier of our systems is the AWS EC2 Instance Id which I can get from Graphite. The JSON with the tags from AWS is displayed weird.

![bildschirmfoto_2017-01-18_17-41-52](https://cloud.githubusercontent.com/assets/4862947/22075244/836b356a-ddab-11e6-8e40-3ca4672bf063.png)
Here you can see the query. It seems that Grafana gets some JSON object from AWS.

Thanks for fixing that in advance.

torkelo:
The syntax is like this: 

```
ec2_instance_attribute(us-east-1, InstanceId, { "tag:Environment": [ "production" ] })
```

chrootLogin:
@torkelo But I have the Instance Ids already. I want the according name tag (or another one) from an Instance Id. 
As you can see in the first thread I want to do something like this:

> `ec2_instance_attribute(eu-central-1, Tags[Name].Value, {"instance-id": ["$instanceid"]})`

But this seems not supported at the moment.

torkelo:
[...]

Issue #3478, opened 2015-12-09
Feature Request: Generic key/value mapping


joeshaw:
First, some background: I have a template variable that lets you select the environment you want to look at: prod, stage, dev.  These map onto our Graphite data source, so we do queries like `$env.counts.etc`.  We also recently started using the CloudWatch data source.  We have ELB instances with names roughly of the form "foo-develb-abcdef."  I would like to be able to take `$env` and map it onto an ELB instance, so that I can keep a single template variable.

So, it'd be nice if I could set up a generic key-value mapping that's something like:

env-to-ELB-name:
dev -> foo-develb-abcdef
stage -> foo-stagee-123456
prod -> foo-prodel-xyz123

and then I could key into that somehow in queries, like `kv_map(env-to-ELB-name, $env)`, and get the proper ELB name.

We would then do the same thing for RDS, SQS, and other AWS services.  Unfortunately while we set AWS tags for env on all the services that support it, CloudWatch doesn't support them as a dimension that can be queried.

For my purposes, I don't need the mapping to be dynamically generated in any way.  I could just set the ELB names by hand, but I could somehow see populating it automatically from elsewhere.

mtanda:
I have similar problem in AWS environment.
And I want to generate mapping automatically.

I think if Grafana has feature "external templating query source" (https://github.com/grafana/grafana/issues/2188), we can handle this well.
Generate mapping in EC2 instance which has enough IAM, and put the mapping result somewhere accessible, then set templating to the mapping.

eheydrick:
This would be useful for mapping instance name -> instanceId.

spinus:
+1
I think this could be implemented as another regexp field, right?
There is one regexp field to extract value from query, probably can be another regexp to extract name.

Than, would be nice if grafana exports those two values to templates, let's say "$var" and "$var.name" (or "$var_name")

mtanda:
I create datasource plugin for this.
https://github.com/mtanda/datasource-plugin-templating_json

It can loading external json file, and query by jmespath.

joeshaw:
@mtanda could you add a README to explain how to install and use it?

mtanda:
@joeshaw The plugin is seems to be broken due to recent Grafana change. I fixed it.
And add README. To use the plugin, you will need to install latest Grafana build (3.0 pre1).

hugohaggmark:
[...]

PR #27281, opened 2020-08-29
Cloudwatch: Add support for resolution of resource tags in alias

### LINKED ISSUE #27280: Cloudwatch: Add support for resolution of resource tags in alias
Add support for using resource tags in the CloudWatch query alias.

See attached PR for details.

Example:

Alex-Richman:
This PR adds support for using resource tags in the CloudWatch query alias.  This is quite useful since the CloudWatch metric label for queries with a resource dimension is the resource ID, which is not particularly human-readable.

Seems to be a fairly desired feature [0][1][2][3][4][5], and personally I couldn't see myself using the CloudWatch datasource without it (who remembers their CloudFront distribution IDs?).

Existing "solutions" (read: bodges) involve using dashboard variable queries (ec2_instance_attributes) and value groups[3], which prevents use of Alerts and only supports EC2 instances.

Resources are fetched using the `resourcegroupstaggingapi` as appropriate for the Namespace in the queries.  For example, a query in the `AWS/EC2` namespace/dimension fetches all resources of type `ec2:instance` and populates a map with their tags.  If all queries are in the same region then this is a single RGTA request for all the queries, so the performance impact is negligible.

The fetched resource tag map is then injected into the `formatAlias` data map based on the CloudWatch metric label, so that tags can be used in the query alias string.

Currently supports fetching resources for EC2 (InstanceID), CloudFront (DistributionId), ApplicationELB (LoadBalancer, TargetGroup), ECS (ClusterName, ServiceName), RDS (DBInstanceIdentifier), and S3 (BucketName).  More resources can be easily added by updating the `resourceTypesMap`, but the current set is all I'm using right now so I wouldn't have any resources to test against for the other types.

Note also that CloudWatch is somewhat inconsistent in how it constructs the metric labels, so we have to do a little feathering to make the label components consistent with the resource IDs from RGTA.  For example ELBv2 TargetGroups have an extra `targetgroup/` prefix in the CloudWatch label compared to the RGTA resource ID.

Closes #27280

[0] https://community.grafana.com/t/cloudwatch-show-servername-instead-of-instance-id/10962
[1] https://community.grafana.com/t/multiple-ec2-instances-in-grafana-dashboard/22032
[2] https://github.com/monitoringartist/grafana-aws-cloudwatch-dashboards/issues/19
[3] Comments in https://github.com/grafana/grafana/pull/8307
[4] https://github.com/grafana/grafana/issues/3478
[5] https://github.com/grafana/grafana/issues/7294

### DIFF
--- a/pkg/tsdb/cloudwatch/response_parser.go
+++ b/pkg/tsdb/cloudwatch/response_parser.go
@@ -7,7 +7,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+	"github.com/gigawattio/awsarn"
 	"github.com/grafana/grafana/pkg/components/null"
 	"github.com/grafana/grafana/pkg/tsdb"
 )
@@ -49,7 +52,7 @@ func (e *cloudWatchExecutor) parseResponse(metricDataOutputs []*cloudwatch.GetMe
 	cloudWatchResponses := make([]*cloudwatchResponse, 0)
 	for id, lr := range mdrs {
 		query := queries[id]
-		series, partialData, err := parseGetMetricDataTimeSeries(lr, labels[id], query)
+		series, partialData, err := e.parseGetMetricDataTimeSeries(lr, labels[id], query)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -7,11 +7,91 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+	"github.com/gigawattio/awsarn"
```
aknuds1, line 13: This adds a dependency that should be reflected in go.mod/go.sum.

aknuds1 (changes requested):
Please see comments.

Alex-Richman:
> Please see comments.

Thanks for your review - I've applied your suggestions and made a further fixup commit to take care of the rest.  Let me know if there's anything else.

aknuds1:
Thanks @Alex-Richman, looks better now, but I'll have to read more to get a grip on the functionality.

pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -190,6 +275,19 @@ func formatAlias(query *cloudWatchQuery, stat string, dimensions map[string]stri
...
+	 */
+	for i, part := range strings.Split(label, " ") {
+		if _, ok := resourceTags[part]; !ok {
+			continue
+		}
+
+		for key, value := range resourceTags[part] {
+			data[fmt.Sprintf("tags.%d.%s", i, key)] = value
+		}
+	}
```
aknuds1, line 289: It's more understandable to me if the variable is named `resource`, since it represents a resource ID. It's also more robust to use the default value in `resourceTags[resource]`, since it will correspond to an empty map:
```suggestion
	for i, resource := range strings.Split(label, " ") {
		tags := resourceTags[resource]
		for key, value := range tags {
			data[fmt.Sprintf("tags.%d.%s", i, key)] = value
		}
	}
```

aknuds1 (dismissed):
Left more suggestions.

I haven't 100% got my head around the functionality yet, but I think we might need tests for `parseResponse` that exercise the new functionality.

Alex-Richman:
@aknuds1 Applied and resolved round 2, thanks again

Alex-Richman:
@aknuds1 Added a few tests related to the functionality

aknuds1:
Thanks - I will look more at this when time allows.

aknuds1 (commented):
Just a nit I spotted.

Alex-Richman:
@aknuds1 Thanks - Applied.

Easiest way to test would be to spin up a couple of EC2 instances, add some tags to them, configure a panel with a CloudWatch datasource query for an EC2 metric (for example CPUUtilization) and play around with the Alias field to render tags from the instances.

Here's a sample panel before & after using a tag based alias (excuse the redactions).

pkg/tsdb/cloudwatch/response_parser.go
```diff
@@ -7,11 +7,85 @@ import (
...
+		typeFilters := []*string{}
+		typeFiltersAdded := map[string]bool{}
+		for _, tp := range resourceTypes {
+			if _, ok := typeFiltersAdded[tp]; !ok {
+				typeFilters = append(typeFilters, aws.String(tp))
+				typeFiltersAdded[tp] = true
+			}
+		}
+
+		resources, err := e.resourceGroupsGetResources(region, nil, typeFilters)
```
marefr, line 61: What's the performance/cost of doing this call per each query/region? At a first glance seems like it could be problematic?

Alex-Richman: Queries in the same region are consolidated into the same RGTA query, so for most panels (which are single region) this will just be a single RGTA call.

RGTA is a very lightweight API (basically just key/value fetching), so the performance impact is negligible.

Additionally, the existing "solution" for this sort of lookup (at least for EC2 instances) is via ec2_instance_attribute at the dashboard level with query variables, which just boils down to a bunch of RGTA queries in the background anyway (one per instance, rather than one per panel).

marefr:
@Alex-Richman thanks for the contribution. However, the referenced feature request #27280 haven't seen any activity/comments/upvotes, that together with the extra complexity, maintenance burden and possibly cost (performance and calls) makes it hard for us to accept this currently. We expect CloudWatch API to have this support built-in when querying metrics. Obviously it doesn't, but currently don't think Grafana should try to solve this limitation. We're very sorry we've given you false expectations accepting this change by requesting changes and so on.  

I'll close this now since we're trying to only have relevant/acceptable PR's open, but please note that a closed PR can always be reopened should a feature request see more activity/discussions or if I misunderstood something. Would suggest to continue the discussions, if any, in the feature request issue #27280.

papagian:
Adding a comment for future reference:
this feature only makes sense if all the timeseries have the specific tag defined
otherwise the other timeseries are not interpolated:
Screenshot 2020-09-16 at 17 06 55

waqark3389:
@papagian Well obviously... 

@marefr Its very annoying that you have closed this. This has driven me to not buy grafana enterprise after much consideration and move to Datadog.
Especially since @Alex-Richman took the time to fix and make improvments on suggestions, you go ahead and close it.

marefr:
@waqark3389 I think I was quite detailed in my response to why we decided to close it and it doesn't mean we can't reopen it. Please provide more information in #27280 to help us understand the problem better and to give other users higher chances of finding this issue by linking all related information you mention together:

> you state you havent seen any activity etc. This question on tags (specifically instance name) has come up countless times on the forum, in IRC, in Slack and in previous github issues. I hadnt upvoted the feature request because I hadnt seen it, as I'm sure is the case with many others.

Thanks

wallet72:
Is there a way to make this an option or a module? I see the cost and workload argument but to be honest if I have to display instance IDs and not tag.Name then I'm going to have a lot of very confused managers.

I mean if you are not familiar with AWS, what is easier to interpret - a name or an instance ID?

Considering these guys pay the bills, they won't like paying for Grafana if it just confuses them.

macobo:
Since more voices were asked for - this is a must-have feature. EC2 is the backbone of aws. Cloudwatch is a great way to get metrics out for EC2 into grafana. 

However without usable labels what use is there for metrics?

chrisolido:
I believe we really need to push this tags feature as all resources are being tag. Its a must for diplaying human readable Name than label we see now.
image

mean not working 
image

{{tags.0.Name}} is not working
image

kmoberg:
Agreed with previous statements here: this should be a no-brainer issue to fix since the fix has already been created and is almost ready to implement. AWS clearly isn't going to change, and we have the opportunity to solve the entire issue by implementing this PR. 

@marefr  Please reconsider this, as others have stated - managers that pay the bills can't have instance IDs shown but need instance names, and us not being able to show this, most likely will force us to move to Datadog or another service other than Grafana Enterprise.

Seven2Nine:
image
why latest grafana version don't have alias field.

marefr:
@grafana/aws-plugins can you please have a look at comment above? Seems like a potential bug

fridgepoet:
Hi @Seven2Nine, Alias has been replaced by the `Label` field which supports [Dynamic Labels].(https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/graph-dynamic-labels.html)
See more about that change here -- https://github.com/grafana/grafana/issues/48434

sunker:
Looks like the Monaco editor in the `Label` field is not being rendered properly here. @Seven2Nine would you mind opening the javascript console and checking if there are any errors? Also, what browser are you using?

PR #27284, opened 2020-08-30
Panel Inspect: Allow CSV download for Excel

### LINKED ISSUE #27268: Unable to Import csv export  in Excel (German Locale) with new lines inside Values
**What happened**:
Download the csv file from Table panel with some of the columns having  **new line character** in them.
Unable to import it  in Excel inside windows 10 machine with German locale

Excel should open the file successfully.
At least after doing Text to Column wizard it should open correctly.(required in german locale where it expects ; and NOT ,)
Instead , it got jumbled up even after that wherever we have newline inside the column/cell value

Have a Table to populate with data containing new line characters
export as csv and open in windows 10 using German locale in Excel and use Text to Column COnversion.
You see data jumbled up and not formatted properly

**Also Users are extremely unhappy with the upgrade from 6 to 7 because they cannot open file diretly anymore as it want semicolon as seperator. Can we make it global configuration  or give choice to user while download or detect user locale and accordingly use appropriate delimiter**

This Occurred only in German locale.

**Environment**:
- Grafana version:v7.1.5  hosted on SUSE sp4
- Data source type & version: Postgres V10
- OS Grafana is installed on:
- Grafana plugins: Table Plugin
- Others:

tomdaly:
Checks user locale to detect list separator for use in CSV export

Fixes #27268 

The issue mentions a problem with newline characters - I think this is a red herring, and the newlines are only an issue because of the incorrect delimiter which causes the newlines to format as part of cells when opened in Excel, instead of representing a new row.


dprokop:
@tomdaly thanks for your contribution! I've tested it and I don't feel like it's working as expected. As described in #27268 the expected delimited with German locale is `;`.

Try doing the following in Chrome's console 

```
var a = ['x', 'y'];
a.toLocaleString('de-DE').charAt(1)
```

It returns `,` as a delimiter, which still seems to be wrong. 

Did you have a chance to test it against Excel?

tomdaly:
@dprokop thanks for the response, I had it working at some point yesterday but cannot reproduce it today, getting the same comma delimiter as you when trying `console.log(['x', 'y'].toLocaleString('de-DE'))`. An alternative fix for this is to just add `sep=,` at the start of the CSV file so that Excel reads it properly with the separator, but I'm not sure how other spreadsheet applications interpret this. The more user-defined option is to ask on download whether the delimiter should be a comma or semicolon, but that will require a few more commits.

torkelo:
Another option is to add back the "excel" option to the csv export. We would have to make the download csv into modal to show csv options (like before). 

With the excel option we before added a excel header to the csv that informs excel about what delimiter is used https://github.com/grafana/grafana/blob/v6.4.3/public/app/core/utils/file_export.ts#L59

dprokop:
> Another option is to add back the "excel" option to the csv export. We would have to make the download csv into modal to show csv options (like before).

Yeah, I think this is the way to go. We probably won't cover 100% use cases, but at least we will have feature parity with previous CSV export. I did some read up on cross-platform CSV export and adding `sep=` at the top is a solution that shows up regularly for Excel CSV issues

dprokop (changes requested):
Another option is to add back the "excel" option to the csv export. We would have to make the download csv into modal to show csv options (like before).

Yeah, I think this is the way to go. We probably won't cover 100% use cases, but at least we will have feature parity with previous CSV export. I did some read up on cross-platform CSV export and adding sep= at the top is a solution that shows up regularly for Excel CSV issues

tomdaly:
A modal with an Excel option sounds like a good idea - I'd be happy to make a start on this from Monday as I'm away from my laptop at the moment. Shall I change the PR to a draft until its ready?

dprokop:
Thanks @tomdaly. I think it's ok to keep it open, I'll rename it to WIP. Let me know if you need any support with this one.

tomdaly:
Added commits to include an Excel header after a modal is shown, currently looks like this:
Screenshot 2020-09-13 at 11 18 38

Are you happy with this or could it be improved (e.g. different text/slider to look like v6.4.3)?

dprokop:
@tomdaly I think the modal flow that was present in v6 was in some way better:

image

What do you think @torkelo?

Long term I think we should aim for a more "inline" solution, I think Dropdown like this would work best:
image

tomdaly:
Sorry for the delay for changes, was waiting for more comments on the UI. I tried doing it Inline but it didn't look that great with a Cascader, and felt like it was creating too many code changes to try and create a custom/new inline cascade component for the button.

Moved the Excel header option to a new toggle within 'Data options', which fits well as it is fundamentally just another option to format the data before you download.

_Screenshot_
Screenshot 2020-09-26 at 11 06 31

What do you think @dprokop ?

dprokop:
@tomdaly I think this is a really good idea! Let me check this with the rest of the team and test how it works!

dprokop:
@tomdaly I've just tested this and I think it works great. Could you please update the documentation with the information about the newly added option? The documentation I'm thinking about is this one: https://grafana.com/docs/grafana/latest/panels/inspect-panel/#download-raw-query-results-as-csv and the and the source code for it you can find here https://github.com/grafana/grafana/blob/eb985ffbeec91cb9c2e9f2fc7f422a20c06b0cc8/docs/sources/panels/inspect-panel.md#L58

Adding @oddlittlebird for the docs review

dprokop (approved):

oddlittlebird:
I don't see any docs yet. Please tag me when they are ready for me to review.

tomdaly:
Thanks for the approval @dprokop 👍 Docs have been updated and are ready to review now @oddlittlebird

coussej:
@tomdaly does this include support for the locale's decimal separator as well?

tomdaly:
> @tomdaly does this include support for the locale's decimal separator as well?

@coussej this specifically solves #27268 to fix the locale's incorrect newline and CSV cell delimiter to allow CSVs to be used in Excel internationally. This shouldn't affect the decimal separator, and I'm not sure what the best way would be to figure this out because it's hard to detect a user's locale in browser (see first few comments/commits before I refactored)

PR #27291, opened 2020-08-31
Grafana-UI: Add Collapse docs

Clarity-89:
Add story and mdx file for Collapse.

### DIFF
--- a/packages/grafana-ui/src/components/Collapse/Collapse.mdx
+++ b/packages/grafana-ui/src/components/Collapse/Collapse.mdx
@@ -1,0 +1,18 @@
+import { Meta, Preview, Props } from '@storybook/addon-docs/blocks';
+import {Collapse} from "./Collapse";
+
+# Collapse
+
+A component that allows users to toggle the display of sections of content.
+
+## Usage
+
+```jsx
+const [isOpen, setIsOpen] = useState(false);
+
+<Collapse label="Collapse panel" isOpen={isOpen} onToggle={() => setIsOpen(!isOpen)}>
+  <p>Panel data</p>
+</Collapse>
+```
+
+<Props of={Collapse}/>
--- a/packages/grafana-ui/src/components/Collapse/Collapse.mdx
+++ b/packages/grafana-ui/src/components/Collapse/Collapse.mdx
@@ -3,7 +3,7 @@ import {Collapse} from "./Collapse";
 
 # Collapse
 
-A component that allows users to toggle the display of sections of content.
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/Collapse/Collapse.story.tsx
```diff
@@ -0,0 +1,43 @@
...
+          </Collapse>
+        );
+      }}
+    </UseState>
+  );
+};
+
+export const controlled = () => {
+  return (
+    <ControlledCollapse label="Collapse panel">
```
Clarity-89, line 39: I think this should actually be called `Uncontrolled` or smth like that but not sure if it's safe to rename.

packages/grafana-ui/src/components/Collapse/Collapse.tsx
```diff
@@ -82,10 +82,11 @@ const getStyles = stylesFactory((theme: GrafanaTheme) => ({
...
 }));
 
-interface Props {
+export interface Props {
   isOpen?: boolean;
   label: string;
   loading?: boolean;
+  /** Toggle collapsed header icon */
   collapsible?: boolean;
   onToggle?: (isOpen: boolean) => void;
```
peterholmberg, line 91: Per the definition of done, all props should have a comment. See component request issue template

peterholmberg (changes requested):

peterholmberg (approved):
:+1: :+1: :+1: :+1: :+1:

packages/grafana-ui/src/components/Collapse/Collapse.mdx
```diff
@@ -0,0 +1,18 @@
+import { Meta, Preview, Props } from '@storybook/addon-docs/blocks';
+import {Collapse} from "./Collapse";
+
+# Collapse
+
+A component that allows users to toggle the display of sections of content.
```
jessover9000, line 6: I don't know for sure how it works because it's not in the storybook yet, but does it work like an accordion?
I would probably say "collapse and expand sections of content" instead of "toggle the display", which is a bit unclear what exactly it means. If it's an accordion we could explain in even more detail how it works – which part is clickable/where does the toggling happen? Is there a part that doesn't get collapsed? How are the sections separated visually? Describing in more detail what exactly the component does and how it looks might go a long way :)

Clarity-89: Yeah, it's a basic accordion. Not sure we'd be that detailed with component descriptions especially when they are quite well known components. But I agree that the description could use some tweaking.

PR #27297, opened 2020-08-31
Thresholds: Fixed issue with thresholds in overrides not working after save and reload

### LINKED ISSUE #27290: Singlestat Panel Defaulting to Base Color After Dashboard Refresh
Hi there, this issue has been bugging me for a few weeks now.

**What happened**:
After the configuration of threshold overrides in the Singlestat Panel, the coloring shows correctly in the panel. However when I save and refresh the page, it defaults back to display the base color. Editing the panel and clicking back onto the threshold override settings the correct color displays again. (I'm using negative values for the threshold settings)

On dashboard refresh, the Singlestat panel should still show the correct color based on the threshold override settings.

image
Correctly configured and working

image
After save and refresh it defaults to the Base color. Clicking and unclicking the threshold value reverts it back to the correct color.

1. Read metric from prometheus into Singlestat Panel.
2. Set the Threshold Override in the Overrides tab - not the Field tab.
3. Use negative numbers for the threshold values.
4. Identify that the correct color is showing in the Singlestat panel.
5. Save and refresh.

**Environment**:
- Grafana version: 7.0.3
- User OS & Browser: Chrome, Windows 10

torkelo:
Fixes #27290

Currently, the fix of changing the first threshold value from null to -Infinity is done in applyFieldOverride but only after defaults are applied. 

Moved logic of fixing null to -Infinity to plugin load instead, feels like this part does not need to be executed on every field override apply.

### DIFF
--- a/public/app/features/dashboard/state/PanelModel.ts
+++ b/public/app/features/dashboard/state/PanelModel.ts
@@ -16,6 +16,8 @@ import {
   PanelEvents,
   PanelPlugin,
   ScopedVars,
+  ThresholdsConfig,
+  ThresholdsMode,
 } from '@grafana/data';
 import { EDIT_PANEL_ID } from 'app/core/constants';
 import config from 'app/core/config';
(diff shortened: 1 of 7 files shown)


public/app/features/dashboard/state/PanelModel.ts
```diff
@@ -501,6 +489,51 @@ export class PanelModel implements DataConfigSource {
...
+      {},
+      defaults.defaults,
+      fieldConfig ? fieldConfig.defaults : {},
+      (objValue: any, srcValue: any): any => {
+        if (_.isArray(srcValue)) {
+          return srcValue;
+        }
+      }
+    ),
+    overrides: [...defaults.overrides, ...(fieldConfig && fieldConfig.overrides ? fieldConfig.overrides : [])],
```
hugohaggmark, line 504: I think this code is a mouthful and deserves some tests to explain it :)

torkelo: also thought this looked strange (this was just copied from the existing code that I extracted to this new method). 

Can simplify this as we do not actually support setting default overrides

hugohaggmark (approved):
LGTM!

PR #27300, opened 2020-08-31
DateFormats: Default ISO & US formats never omit date part even if date is today (breaking change)

torkelo:
The logic of hiding the date for the ISO and US formats by default makes formatting too inconsistent and unpredictable. 

Think the default formatting should always include the full date. 

What to call these "smart" date formats that hide today's date I am not sure about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 7.2.0-beta1 (unreleased)
+
+### Breaking changes
+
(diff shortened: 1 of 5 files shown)


ryantxu (commented):
"smart" implies much more than dropping the data.  Maybe:
* Datetime ISO
* Datetime ISO (Only time if today)

torkelo:
Updated the naming, need to add breaking change notice to changelog

PR #27310, opened 2020-09-01
PanelEdit: Vertically align padding for panel edit option group body

torkelo:
The 0 top and 16px below padding looked a bit off, felt 8px above and below looks better

### DIFF
--- a/public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
+++ b/public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
@@ -162,7 +162,7 @@ const getStyles = stylesFactory((theme: GrafanaTheme, isExpanded: boolean, isNes
     ),
     body: cx(
       css`
-        padding: ${theme.spacing.sm} ${theme.spacing.md} ${theme.spacing.xl} ${theme.spacing.sm};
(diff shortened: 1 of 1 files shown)


public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
```diff
@@ -162,7 +162,7 @@ const getStyles = stylesFactory((theme: GrafanaTheme, isExpanded: boolean, isNes
     ),
     body: cx(
       css`
-        padding: 0 ${theme.spacing.md} ${theme.spacing.md} ${theme.spacing.xl};
+        padding: ${theme.spacing.sm} ${theme.spacing.md} ${theme.spacing.xl} ${theme.spacing.sm};
```
mckn, line 165:
```suggestion
        padding: ${theme.spacing.sm} ${theme.spacing.md} ${theme.spacing.xl} ${theme.spacing.md};
```

Shouldn't the last one be `${theme.spacing.md}` if we specify it? Or removed since it will be the same as the second value if we remove it.

mckn (changes requested):

public/app/features/dashboard/components/PanelEditor/OptionsGroup.tsx
torkelo: ah, thanks for catching this, I fixed it first in browser and made a mistake when implementing it code!

PR #27311, opened 2020-09-01
Chore: fixed text-alignment on hovering tableCells

### LINKED ISSUE #26496: Hovering over table cell with URL causes text to move to left
**What happened**:
Whenever I hover the mouse over the table cell, it causes the text content to move to the left. It resets itself when I move the mouse cursor outside of the cell. This doesn't happen to cells that doesn't have a URL link assigned to it.
Before:

After:

Text content in cell doesn't move when mouse cursor hovers it.

Create a table cell with random values and assign a URL to one column.

This doesn't happen in previous v7 updates.

**Environment**:
- Grafana version: v7.1.0
- Data source type & version: Prometheus v2.17.1
- OS Grafana is installed on: Ubuntu v20.04

michelengelen:
## FilterableTableCell

Hovering over a cell in an ad-hoc filtering table now does not push text to the left anymore (when it was previously aligned to center or right).

### added
- added `display: inline-flex` and `align-items: center` to
  `tableCellWrapper` class
- added `flex: 1` to `tableCell` class

### removed
- unused imports
- removed ternary if since it has no use anymore
- removed `tableCellWrapper` style definition from stylesFactory

Fixes #26496

### DIFF
--- a/packages/grafana-ui/src/components/Table/FilterableTableCell.tsx
+++ b/packages/grafana-ui/src/components/Table/FilterableTableCell.tsx
@@ -1,10 +1,9 @@
 import React, { FC, useCallback, useState } from 'react';
 import { TableCellProps } from 'react-table';
 import { GrafanaTheme } from '@grafana/data';
-import { css, cx } from 'emotion';
+import { css } from 'emotion';
 
 import { stylesFactory, useTheme } from '../../themes';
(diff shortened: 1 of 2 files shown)


dprokop (approved):
Thanks for the contribution @michelengelen, this looks good!

PR #27317, opened 2020-09-01
FrameView: add a more functional approach to iterate rows/columns (exploration)

ryantxu:
This is an experiment to see if we can have a more functional appraoch... feedback welcome.

This lets you iterate frame row/field values with:

```
const vals = view.rows.map((fields, index) => {
      return fields.map(f => {
        return f.value;
      });
    });
```

or

```
for (const fields of view.rows) {
      vals2.push(
        fields.map(f => {
          return f.value;
        })
      );
    }
```


marcusolsson:
Sorry for the delay. I'll play around with this first thing tomorrow! 👍

ryantxu:
@marcusolsson -- something we should consider for 7.3?

packages/grafana-data/src/dataframe/DataFrameView.ts
```diff
@@ -1,7 +1,12 @@
-import { DataFrame } from '../types/dataFrame';
+import { DataFrame, Field } from '../types/dataFrame';
 import { DisplayProcessor } from '../types';
 import { FunctionalVector } from '../vector/FunctionalVector';
 
+interface FieldAndValue {
+  field: Field;
+  value: any;
```
marcusolsson, line 7: This is handy, but makes it a bit weird to use:

```ts
view.rows.map((fields, index) => {
  return fields.map(field => field.field); // field.field?
});
```

What if `fields` was a `Field[]`, and you'd still use the index? Might not be reason enough to add it.

```ts
view.rows.map((fields, index) => {
  return fields.map(field => field.values.get(index));
});
```

Or—forgive my TypeScript skills here—could we return a union type that just adds add `value` property to the Field? Might be too much magic though.

PR #27320, opened 2020-09-01
InfluxDB: return a table rather than an error when timeseries is missing time

### LINKED ISSUE #27129: InfluxDB single value Flux queries get "no time column in timeSeries" error
I just tried to Grafana v7.1.4 and have gotten my time series Flux queries working, but stat panel results are having issues.
InfluxDB is version v1.8.2

Take the following query for example:
```
from(bucket: "maintainable_db/autogen")
  |> range(start: -30d)
  |> filter(fn: (r) => r._measurement == "summary" and r._field == "summary_serial_number" and r.fixture =~ /$fixture_id/ and r.part_number =~ /$part_number/)
  //Group results by serial number and get the first entry
  |> group(columns: ["_value"], mode: "by")
  |> first()
  |> filter(fn: (r) => r.result != "Passed")
  // Only keep entries with the first value in the desired time range
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> group(columns: ["_field"], mode: "by")
  // Use count to get a sum for the window
  |> count()
````
Gets the following result in the query inspector:
```
  {
  "response": {
    "results": {
      "A": {
        "error": "no time column in timeSeries",
        "refId": "A",
        "series": null,
        "tables": null,
        "dataframes": null
      }
    },
    "message": "no time column in timeSeries"
  }
}
```

The issue is the count operator drops the _start, _stop, and time column which is expected.

Running the same query in Chronograf results in:
```
#group,false,false,true,false
#datatype,string,long,string,long
#default,_result,,,
,result,table,_field,_value
,,0,summary_serial_number,6
```
which seems like a perfectly reasonable result to use in a panel that displays a single value.

### PR DESCRIPTION
**Pull request by ryantxu**
Fixes #27129

Rather than an error, this
```
#group,false,false,true,false
#datatype,string,long,string,long
#default,_result,,,
,result,table,_field,_value
,,0,summary_serial_number,6
```

will now return:
```
+-----------------------+----------------+
| Name: _field          | Name: _value   |
| Labels:               | Labels:        |
| Type: []*string       | Type: []*int64 |
+-----------------------+----------------+
| summary_serial_number | 6              |
+-----------------------+----------------+
```

### CODE REVIEW

File: pkg/tsdb/influxdb/flux/builder.go [hunk 1]
```diff
@@ -119,40 +119,42 @@ func (fb *frameBuilder) Init(metadata *query.FluxTableMetadata) error {
...
-		fb.timeColumn = col.Name()
-		fb.timeDisplay = "Time"
-		if "_time" != fb.timeColumn {
-			fb.timeDisplay = col.Name()
+	// reset any timeseries properties
+	fb.value = nil
+	fb.isTimeSeries = false
+	fb.labels = make([]string, 0)
+	for _, col := range columns {
+		// Skip the result column
```
srclosson, line 140: Why skip the result column? What data is in it?

srclosson (approved):

pkg/tsdb/influxdb/flux/builder.go
ryantxu: That is from:
https://github.com/grafana/grafana/pull/27081

when things are not a timeseries... they are just empty

ryantxu: see also https://github.com/grafana/grafana/issues/26927

ryantxu: (note the only change here is just indentation

PR #27326, opened 2020-09-02
Chore: Upgrade to Go 1.15.1

aknuds1:
Upgrade to Go 1.15.1.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ COPY emails emails
 ENV NODE_ENV production
 RUN ./node_modules/.bin/grunt build
 
-FROM golang:1.14.7-alpine3.12 as go-builder
+FROM golang:1.15.1-alpine3.12 as go-builder
 
 RUN apk add --no-cache gcc g++
 
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -16,7 +16,7 @@ COPY emails emails
 ENV NODE_ENV production
 RUN ./node_modules/.bin/grunt build
 
-FROM golang:1.14.7 AS go-builder
+FROM golang:1.15.1 AS go-builder
 
 WORKDIR /src/grafana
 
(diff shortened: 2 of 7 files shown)


sakjur (approved):

sakjur:
@marefr 1.15 isn't a very exciting release, and I've been running Grafana on master with 1.15 without any problems since the day of the 1.15 release without noticing any negative consequences.

It is possible that we might have users (for example orgs rolling internal certificates) that could run into the deprecation of x.509 certificate's common name happening in 1.15:  https://golang.org/doc/go1.15#commonname

I think, in a sense, that it makes sense to use the beta to verify that the behavior of 1.15 is not significantly changed in any way from the behavior of 1.14, where upgrading from 1.14->1.15 in a later patch release (or worse, 1.14->1.16 when that drops, since 1.14 won't be getting security patches at that point) wouldn't let us have the confidence of a tested beta behind us. Having a beta release ahead of us is better than a patch release.

aknuds1:
@marefr Yeah, the idea was to bring us up to date with latest Go, and to get the security fix in 1.15.1. I read the release notes and didn't find anything that would break.

aknuds1:
I agree with @sakjur that it would make sense to test Go 1.15 as part of the beta.

kylebrandt (approved):

kylebrandt:
Should go.mod go version be updated with this as well?

marefr (approved):
> It is possible that we might have users (for example orgs rolling internal certificates) that could run into the deprecation of x.509 certificate's common name happening in 1.15: https://golang.org/doc/go1.15#commonname

You can use a env variable to set this back so should be fine.

I approve

sakjur:
> > It is possible that we might have users (for example orgs rolling internal certificates) that could run into the deprecation of x.509 certificate's common name happening in 1.15: https://golang.org/doc/go1.15#commonname
> 
> You can use a env variable to set this back so should be fine.

Yes! We just need to be aware of it in case it appears in a bug report or something like that :)

PR #27342, opened 2020-09-02
Annotations: improve datasource annotation types and add basic query properties

ryantxu:
Extracted from #26634

The current annotation types are not quite right and make it difficult to implement react query editors with real types.

This also adds the standard parameters passed in any regular query:
```
  interval: string;
  intervalMs: number;
  maxDataPoints?: number;
  app: CoreApp | string;
```
to annotation queries so that we can simply call the default query endpoint and have it mostly work as expected


packages/grafana-data/src/types/datasource.ts
```diff
@@ -535,19 +542,19 @@ export interface DataSourceSelectItem {
 /**
  * Options passed to the datasource.annotationQuery method. See docs/plugins/developing/datasource.md
  */
-export interface AnnotationQueryRequest<MoreOptions = {}> {
+export interface AnnotationQueryRequest<TAnno = {}> {
```
ryantxu, line 545: would be nice if TAnno *extends* BaseAnnotationQuery... but that will break many existing types that use TQuery for the annotation type

packages/grafana-data/src/types/datasource.ts
```diff
@@ -535,19 +542,19 @@ export interface DataSourceSelectItem {
...
   range: TimeRange;
   rangeRaw: RawTimeRange;
+  interval: string;
+  intervalMs: number;
+  maxDataPoints?: number;
+  app: CoreApp | string;
```
dprokop, line 551: `app` is confusing here, first thing that came to my mind was - why is the app object passed to the request? Alternative would be `context` or `appContext`?

dprokop (commented):
I'm not familiar with annotations, but aren't new properties added to `AnnotationQueryRequest` a potential breaking change for plugins?

packages/grafana-data/src/types/datasource.ts
ryantxu: This just copies what exists in the query request object -- I think added so that an explore query could have different behavior.

I added it so I could set a flag when in the annotation editor

ryantxu:
Should not be a breaking change -- the property that people manipulate is 'annotation.annotation' these values sit next to that one.

The goal here is to have reasonable values to just use the standard query method

torkelo (approved):
Looks good!

PR #27347, opened 2020-09-02
RangeUtils: migrate logic from kbn to grafana/data

ryantxu:
Pulling this out of #21835 

`kbn.ts` still holds lots of interesting interval math logic... that makes it hard to use outside core ;(

### DIFF
--- a/public/app/core/utils/kbn.ts
+++ b/public/app/core/utils/kbn.ts
@@ -66,8 +66,8 @@ const kbn = {
   },
   /** deprecated since 7.2, use grafana/data */
   intervalToMs: (str: string) => {
-    deprecationWarning('kbn.ts', 'kbn.intervalToMs()', '@grafana/data');
-    return rangeUtil.intervalToMS(str);
+    deprecationWarning('kbn.ts', 'rangeUtil.intervalToMs()', '@grafana/data');
+    return rangeUtil.intervalToMs(str);
   },
   /** deprecated since 7.2, use grafana/data */
   calculateInterval: (range: TimeRange, resolution: number, lowLimitInterval?: string) => {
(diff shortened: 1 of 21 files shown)


public/app/features/variables/interval/actions.ts
```diff
@@ -29,22 +28,22 @@ export const updateIntervalVariableOptions = (
 };
 
 export interface UpdateAutoValueDependencies {
-  kbn: typeof kbn;
+  rangeUtil: any; // typeof rangeUtil?
```
ryantxu, line 31: @hugohaggmark any ideas here?  `rangeUtil` is an import namespace, not a real object like kbn

hugohaggmark, line 31: What about this?
```suggestion
  calculateInterval: typeof calculateInterval;
```

hugohaggmark: I think I only use it as a dependency to simplify testing :)

hugohaggmark: So you'll need to update some tests as well, but should be straightforward. Let me know if I can help out

ryantxu: that worked!  thank you

public/app/core/utils/kbn.ts
```diff
@@ -25,123 +25,16 @@ const kbn = {
...
-    if (numSeconds) {
-      return numSeconds + 's';
-    }
-    const numMilliseconds = Math.floor(seconds * 1000.0);
-    if (numMilliseconds) {
-      return numMilliseconds + 'ms';
-    }
-
-    return 'less than a millisecond'; //'just now' //or other string you like;
+  /** deprecated since 7.2, use grafana/data */
```
dprokop, line 34: I think adding `@depreceted` annotation will make the code editors highlight the method as deprecated. In Webstorm it looks like this:
image

dprokop (approved):
This is gold! We need to add this to the changelog though :)

PR #27349, opened 2020-09-03
RFC: show streaming indicator (and allow unsubscribe)

ryantxu:
We have a good loading indicator, but nothing that indicates data is still streaming.  I suggest a simple dot or something.  The best existing icon i see is circle: 

image


dprokop:
I think a green, pulsating dot would work well rather than a circle. A circle might be confusing IMO and be mistaken with a loading indicator that finished loading.

ryantxu:
yes -- I used the circle icon because it is the only one in the library :)

Ijin08:
The dot feels a bit dead and inactive. If blinking pulsating feels a bit too much maybe a green dot with a green circle around it. What happen if one unsubscribes? Can one click to resubscribe?

dprokop:
Uhm, is it @ryantxu? I can see it's available in unicons: https://iconscout.com/unicons/explore/solid (type circle). Maybe we just don't expose it in the IconType?

dprokop:
@ryantxu I may be wrong actually aboutsolid circle being available. I'm looking at Unicons package we have in our modules and it seems like there are no solid icons available. Maybe @ivanahuckova can help with this?

jessover9000:
I agree with @Ijin08 that the dot should look more interactive than just grey. I also do think that some sort of animation or glow to indicate data being transferred would be nice. In Loki live tailing we're using the brand orange/yellow glow, maybe it would make sense to adapt that "live" style for this use case as well. Also I think if the signifier we use is just an icon without text we need a hover tooltip to explain what it means. A circle is so abstract, it could be anything.

public/app/features/dashboard/dashgrid/PanelHeader/PanelHeader.tsx
```diff
@@ -91,14 +91,26 @@ export class PanelHeader extends Component<Props, State> {
...
+            <Icon className="panel-loading__spinner spin-clockwise" name="sync" />
+          </Tooltip>
+        </div>
+      );
+    }
+    if (state === LoadingState.Streaming) {
+      return (
+        <div className="panel-loading" onClick={this.onCancelQuery}>
+          <Tooltip content="Streaming (unsubscribe)">
+            <Icon name="circle" />
```
ivanahuckova, line 108: You can use already implemented [monochrome circle icon](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Icon/assets/Circle.tsx). 
image

So you just need to add `type="mono"` and use css (className/style) to update its color to green.

ryantxu:
Closing in favor of https://github.com/grafana/grafana/pull/28682

PR #27351, opened 2020-09-03
Elasticsearch: Fix localized dates in index pattern

### LINKED ISSUE #8403: [BUG] Elasticsearch datasource index pattern fails to fetch data with Arabic browser language
If my system language is set to Arabic all the date formats are in Arabic but when querying elasticsearch index pattern it sends the request in Arabic but the index exists only in english.

logstash-2017-05-17 index is in elasticsearch but the browser sends the request for 
logstash-٢.١٧-.‎٥-١٧‎
which does not exist.

domasx2:
Dates in elasticsearch index patterns are formatted using moment with whatever locale is currently set on it, depending on user's system. Some locales, like saudi-arabic (`ar-SA`) produce non-western-arabic numeric characters, for example `logstash-٢.١٧-.‎٥` instead of `logstash-2017-05-17`. This causes non existing index to be referenced. Changes in this PR force `en` locale for the index pattern date.

Fixes #8403

### DIFF
--- a/packages/grafana-data/src/datetime/moment_wrapper.ts
+++ b/packages/grafana-data/src/datetime/moment_wrapper.ts
@@ -79,6 +79,10 @@ export const setLocale = (language: string) => {
   moment.locale(language);
 };
 
+export const getLocale = () => {
+  return moment.locale();
+};
+
 export const getLocaleData = (): DateTimeLocale => {
   return moment.localeData();
 };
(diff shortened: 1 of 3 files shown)


packages/grafana-data/src/datetime/moment_wrapper.ts
```diff
@@ -79,6 +79,10 @@ export const setLocale = (language: string) => {
   moment.locale(language);
 };
 
+export const getLocale = () => {
+  return moment.locale();
+};
+
```
mckn, line 85: It would be better if we don't extend the moment_wrapper API with more functions but at the same time there is no better way of doing this.

mckn (approved):
I think this looks really good! Thanks for the contribution @domasx2!

marefr (approved):
LGTM

marefr:

PR #27352, opened 2020-09-03
Search: refactor tests to use react-testing-library

Clarity-89:
Replace Enzyme with RTL for components tests in search.


public/app/core/services/__mocks__/search_srv.ts
```diff
@@ -0,0 +1,22 @@
...
+  Promise.resolve({ sortOptions: [{ name: 'test', displayName: 'Test' }] })
+);
+
+export const SearchSrv = jest.fn().mockImplementation(() => {
+  return {
+    search: mockSearch,
+    getDashboardTags: mockGetDashboardTags,
+    getSortOptions: mockGetSortOptions,
+  };
+});
```
hugohaggmark, line 22: I think this ok for this PR but I find this convention using files under `__mocks__` very confusing when you try to debug/change test you haven't written. I remember one time me and @mckn spent like 30 minutes trying to understand a test that didn't behave like we understood it would because of a file under `__mocks__`

I'd like for us to have a discussion about this during a future FrontEnd weekly to hear what the consensus is amongst everyone.

Clarity-89: @hugohaggmark thanks for the feedback, wasn't aware that this causes some difficulties. Just to clarify, do you mean that we shouldn't use automatic mocks? I'm wondering what would be the alternative? In this case the mock is not applied unless `jest.mock('app/core/services/search_srv');` is called in the test file, so that in my opinion should be a clear indicator that the mock functionality is used.

hugohaggmark: Yes, I agree with you, that should be a sign. But most of our tests include the mock within the same test file, so before I didn't pay so much attention to that line. Now that I've been burned by this I pay more attention to this line.

My personal opinion is that it should be crystal clear in each test what mocks/dependencies you're using and we should define them in the same file. I'm probably the odd bird here so a proper discussion on the way forward would be great :)

Clarity-89: I see your point and I guess we do need to have some discussion as I was just thinking that actually having mocks in `__mock__` files is an improvement that allows reusing them without mocking the functionality in each test file :)

public/app/features/search/components/SearchResults.test.tsx
```diff
@@ -1,52 +1,56 @@
...
+    expect(screen.getAllByRole('checkbox')).toHaveLength(4);
+  });
+
+  it('should collapse folder item list on header click', () => {
+    const mockOnToggleSection = jest.fn();
+    setup({ onToggleSection: mockOnToggleSection });
+
+    fireEvent.click(screen.getByLabelText('Collapse folder 0'));
+    expect(mockOnToggleSection).toHaveBeenCalledTimes(1);
+    expect(mockOnToggleSection).toHaveBeenCalledWith(searchResults.find(result => result.id === 0));
```
hugohaggmark, line 54: nitpick: I think we should try to keep our expect statements raw and avoid using any logic/function to retrieve the expected value.

public/app/features/search/components/SearchResults.tsx
```diff
@@ -38,9 +38,13 @@ export const SearchResults: FC<Props> = ({
           return (
             <div aria-label="Search section" className={styles.section} key={section.id || section.title}>
               <SectionHeader onSectionClick={onToggleSection} {...{ onToggleChecked, editable, section }} />
-              <div aria-label="Search items" className={styles.sectionItems}>
-                {section.expanded && section.items.map(item => <SearchItem key={item.id} {...itemProps} item={item} />)}
-              </div>
+              {section.expanded && (
+                <div aria-label="Search items" className={styles.sectionItems}>
```
hugohaggmark, line 42: nit: These aria-labels could be added to e2e-selectors package

hugohaggmark (approved):
Great work on this ⭐, I just left some minor nits for consideration.

PR #27356, opened 2020-09-03
Alerting: Adds support for overriding 'dedup_key' via alert tags when using the Pagerduty notifier

### LINKED ISSUE #26863: Alerting: PagerDuty de-duplication key too simple, may cause incorrect incident de-duplication

alavrovinfb:
- [X] Added ability to override PagerDuty 'dedup_key' via Alert tags

- [X] Updated unit tests

- [X] Updated Documentation

Changes from this PR allows user to override weak Grafana generated `dedup_key` on their own.
It can avoid incorrect PagerDuty incidents de-duplication. A possible issue is described [here](https://github.com/grafana/grafana/issues/26863l)

This PR has been raised to fix Grafana issue [#26863 ](https://github.com/grafana/grafana/issues/26863)
 

Fixes https://github.com/grafana/grafana/issues/26863

### DIFF
--- a/docs/sources/alerting/notifications.md
+++ b/docs/sources/alerting/notifications.md
@@ -132,7 +132,7 @@ This might break custom event rules in your PagerDuty rules if you rely on the f
 Move any existing rules using `custom_details.myMetric` to `custom_details.queries.myMetric`.
 This behavior will become the default in a future version of Grafana.
 
->Using `dedup_key` tag will override grafana generated `dedup_key` with a custom key.
+> Using `dedup_key` tag will override grafana generated `dedup_key` with a custom key.
 ### Webhook
 
 The webhook notification is a simple way to send information about a state change over HTTP to a custom endpoint.
--- a/docs/sources/alerting/notifications.md
+++ b/docs/sources/alerting/notifications.md
@@ -125,7 +125,7 @@ Severity | Level for dynamic notifications, default is `critical` (1)
 Auto resolve incidents | Resolve incidents in PagerDuty once the alert goes back to ok
 Message in details | Removes the Alert message from the PD summary field and puts it into custom details instead (2)
 
->**Note:** The tags `Severity`, `Class`, `Group`, `dedup_key` and `Component` have special meaning in the [Pagerduty Common Event Format - PD-CEF](https://support.pagerduty.com/docs/pd-cef). If an alert panel defines these tag keys, then they are transposed to the root of the event sent to Pagerduty. This means they will be available within the Pagerduty UI and Filtering tools. A Severity tag set on an alert overrides the global Severity set on the notification channel if it's a valid level.
+>**Note:** The tags `Severity`, `Class`, `Group`, `dedup_key`, and `Component` have special meaning in the [Pagerduty Common Event Format - PD-CEF](https://support.pagerduty.com/docs/pd-cef). If an alert panel defines these tag keys, then they are transposed to the root of the event sent to Pagerduty. This means they will be available within the Pagerduty UI and Filtering tools. A Severity tag set on an alert overrides the global Severity set on the notification channel if it's a valid level.
 
 >Using Message In Details will change the structure of the `custom_details` field in the PagerDuty Event.
 This might break custom event rules in your PagerDuty rules if you rely on the fields in `payload.custom_details`.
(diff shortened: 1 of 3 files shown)


docs/sources/alerting/notifications.md
```diff
@@ -125,13 +125,14 @@ Severity | Level for dynamic notifications, default is `critical` (1)
 Auto resolve incidents | Resolve incidents in PagerDuty once the alert goes back to ok
 Message in details | Removes the Alert message from the PD summary field and puts it into custom details instead (2)
 
->**Note:** The tags `Severity`, `Class`, `Group`, and `Component` have special meaning in the [Pagerduty Common Event Format - PD-CEF](https://support.pagerduty.com/docs/pd-cef). If an alert panel defines these tag keys, then they are transposed to the root of the event sent to Pagerduty. This means they will be available within the Pagerduty UI and Filtering tools. A Severity tag set on an alert overrides the global Severity set on the notification channel if it's a valid level.
+>**Note:** The tags `Severity`, `Class`, `Group`, `dedup_key` and `Component` have special meaning in the [Pagerduty Common Event Format - PD-CEF](https://support.pagerduty.com/docs/pd-cef). If an alert panel defines these tag keys, then they are transposed to the root of the event sent to Pagerduty. This means they will be available within the Pagerduty UI and Filtering tools. A Severity tag set on an alert overrides the global Severity set on the notification channel if it's a valid level.
```
achatterjee-grafana, line 128: Doc change looks OK.

oddlittlebird, line 128:
```suggestion
>**Note:** The tags `Severity`, `Class`, `Group`, `dedup_key`, and `Component` have special meaning in the [Pagerduty Common Event Format - PD-CEF](https://support.pagerduty.com/docs/pd-cef). If an alert panel defines these tag keys, then they are transposed to the root of the event sent to Pagerduty. This means they will be available within the Pagerduty UI and Filtering tools. A Severity tag set on an alert overrides the global Severity set on the notification channel if it's a valid level.
```

docs/sources/alerting/notifications.md
```diff
@@ -125,13 +125,14 @@ Severity | Level for dynamic notifications, default is `critical` (1)
...
 
 >Using Message In Details will change the structure of the `custom_details` field in the PagerDuty Event.
 This might break custom event rules in your PagerDuty rules if you rely on the fields in `payload.custom_details`.
 Move any existing rules using `custom_details.myMetric` to `custom_details.queries.myMetric`.
 This behavior will become the default in a future version of Grafana.
 
+>Using `dedup_key` tag will override grafana generated `dedup_key` with a custom key.
```
oddlittlebird, line 135:
```suggestion
> Using `dedup_key` tag will override grafana generated `dedup_key` with a custom key.
```

oddlittlebird:
I made some suggestions on the docs.

marefr (approved):
LGTM

marefr:

PR #10947, opened 2018-02-16
login: uses epochs for login throtting.

xlson:

PR #10907, opened 2018-02-14
provisioning: uses unix epoch timestamps.

xlson:
DateTime was used previously but as the timezone info is not
persisted in the database that might cause issues with
up-to-date checking. Unix epoch timestamps sidesteps this issue
as they are always in UTC.

Fixes #10904

Issue #10904, opened 2018-02-14
Convert provisoning updated at field to epoch form datetime.


bergquist:
We had some problems with different timezones for the updated at field in the dashboard provisioning lookup table. To avoid dealing with timezones and the fact that sqlites doesn't include the timezone in datetimes we will be changing to using epoch instead.

PR #27361, opened 2020-09-03
Users: Expire old user invites

### LINKED ISSUE #20836: User invites should be expired after certain time and deleted from database
User invites should be expired after certain time and deleted from the database.

**Why is this needed**:
From a security perspective it's not good practice to let invites live forever and we don't want unneeded and old data to lie around in the database.

wbrowne:
Adds a cleanup step to expire user invites (by default those that have been created over 1 week ago).

Fixes #20836

Introduces a new status to represent expired invites. The cleanup services runs every 10 minutes and will mark any **pending** invites which are over 1 week old. I migrated the existed created+updated columns to UTC (as due to the time checking I found inconsistencies across the different supported DBs). Thanks @marefr for pointing this out (see #10907 and #10947) 

One thing I'm considering to change is maybe it is more flexible to have the expiry time reflected as hours instead of days? As it stands currently with this PR, pending invites can exist for 24 hours but maybe some people will want a shorter window (such as 12 hours or even just 1)


marefr (commented):
Code wise both ways looks good and is reasonable. Prefer the one using the cleanup service since it follows other kinds of cleanup tasks. Would be nice with an integration test of the cleanup invites job.

But I think we should rather expire rows than deleting them, why?

Looking at the current implementation, you can see a list of pending invites;
image

Clicking in the X-button will not delete the row in the database, but rather change the [status](https://github.com/grafana/grafana/blob/be8bc512cb97a9b82e00666b7a10774de856f51b/pkg/models/temp_user.go#L15-L20) to `Revoked`. 

Why I think we should introduce a new status `Expired` or `InviteExpired` and use that in cleanup job when updating rows.

The existing [guard](https://github.com/grafana/grafana/blob/be8bc512cb97a9b82e00666b7a10774de856f51b/pkg/api/org_invite.go#L150-L152) will make sure that only pending invites is possible to use for completing a sign up.

Questions/thoughts?

wbrowne:
@marefr that was my original thought when thinking about this problem (introducing another status), but I opted for just purging the data instead because I felt that they would be essentially dead rows that would never be updated (or at least shouldn't, IE if a temp_user had been expired, there was no way to revive that invite) and would essentially just be wasted space. It probably wouldn't be such a big deal in reality, so I'm not too pushed either way. To get a feel for the change, I've updated the code to introduce a status - tell me what you think and we can go from there

pkg/services/sqlstore/temp_user.go
```diff
@@ -132,3 +133,17 @@ func GetTempUserByCode(query *models.GetTempUserByCodeQuery) error {
 	query.Result = &tempUser
 	return err
 }
+
+func ExpireOldUserInvites(cmd *models.ExpireTempUsersCommand) error {
+	return inTransaction(func(sess *DBSession) error {
+		var rawSql = "UPDATE temp_user SET status = ? WHERE created <= ? AND status != ?"
+		expiredStatus := string(models.TmpUserExpired)
```
marefr, line 140: There's a potential timezone issue here when comparing date type (created) in database query. We had several bugs related to that throughout the years and the proper solution has been to add a new column holding an epoch value. I may be wrong, but think it should be possible to verify this with a test and mocking the time and then integration tests should catch it. For reference #10947 #10904 #10907 as examples where we fixed this earlier. 

It's tricky though since you normally have to drop the table and create a new one or add one additional column. Need to verify if this a problem first. Could be that our integration tests don't catch these kind of errors.

However, I suggest updating invites that are in pending state. WDYT?

```suggestion
		var rawSql = "UPDATE temp_user SET status = ? WHERE created <= ? AND status = ?"
		expiredStatus := string(models.TmpUserInvitePending)
```

marefr (commented):
Looks great :+1: But please see my comment. Some tricky situations to take into account.

pkg/services/sqlstore/temp_user.go
wbrowne: Ah good catch. Let me test it out and report back!

Regarding the state check, yep that makes sense since this will even change successfully completed invites for example which we don't want. With that in mind, we're okay with leaving invites in a state of `SignUpStarted` indefinitely?

wbrowne: Putting here for completeness:

Results show that there is indeed differing results between the 3 different SQL sources. Here's the test:

```
Convey("Should respect timezone differences when expiring invites ", func() {
    now := time.Date(2020, 1, 1, 12, 0, 0, 0, time.UTC) // 1 Jan 2020 12:00:00 UTC
    timeNow = func() time.Time { return now }
    cmd := models.CreateTempUserCommand{
        OrgId:  123,
        Name:   "test",
        Code:   "789",
        Email:  "test@grafana.com",
        Status: models.TmpUserInvitePending,
    }
    err := CreateTempUser(&cmd)
    So(err, ShouldBeNil)
    tmpUser := cmd.Result
    So(tmpUser.Created, ShouldEqual, now)
    utcPlus1 := time.FixedZone("UTC+1", 1*60*60)
    olderThan := time.Date(2020, 1, 1, 12, 0, 0, 0, utcPlus1) // 1 Jan 2020 12:00:00 UTC+1 (IE 1 Jan 2020 11:00:00 UTC)
    cmd4 := models.ExpireTempUsersCommand{OlderThan: olderThan}
    err = ExpireOldUserInvites(&cmd4)
    So(err, ShouldBeNil)
    So(cmd4.NumExpired, ShouldEqual, 0)
    olderThan = time.Date(2020, 1, 1, 13, 0, 0, 0, utcPlus1) // 1 Jan 2020 13:00:00 UTC+1 (IE 1 Jan 2020 12:00:00 UTC)
    cmd4 = models.ExpireTempUsersCommand{OlderThan: olderThan}
    err = ExpireOldUserInvites(&cmd4)
    So(err, ShouldBeNil)
    So(cmd4.NumExpired, ShouldEqual, 1)
})
```
Passes on MySQL, but fails on SQLite and PostgreSQL

The plan now is to look at a migration so that we deal exclusively with epoch similar to other migrations done in the past

marefr: Great :+1:

pkg/services/sqlstore/temp_user.go
```diff
@@ -132,3 +133,15 @@ func GetTempUserByCode(query *models.GetTempUserByCodeQuery) error {
...
 }
+
+func ExpireOldUserInvites(cmd *models.ExpireTempUsersCommand) error {
+	return inTransaction(func(sess *DBSession) error {
+		var rawSql = "UPDATE temp_user SET status = ?, updated = ? WHERE created <= ? AND status = ?"
+		if result, err := sess.Exec(rawSql, string(models.TmpUserExpired), time.Now().Unix(), cmd.OlderThan.Unix(), string(models.TmpUserInvitePending)); err != nil {
+			return err
+		} else if cmd.NumExpired, err = result.RowsAffected(); err != nil {
+			return err
+		}
```
aknuds1, line 144: The `else` is non-idiomatic I believe, with Go error handling.
```suggestion
		}
		if cmd.NumExpired, err := result.RowsAffected(); err != nil {
			return err
		}
```

pkg/services/sqlstore/temp_user_test.go
```diff
@@ -62,7 +62,21 @@ func TestTempUserCommandsAndQueries(t *testing.T) {
 
 				So(err, ShouldBeNil)
 				So(query.Result[0].EmailSent, ShouldBeTrue)
-				So(query.Result[0].EmailSentOn, ShouldHappenOnOrAfter, (query.Result[0].Created))
+				So(query.Result[0].EmailSentOn.UTC(), ShouldHappenOnOrAfter, (query.Result[0].Created.UTC()))
```
aknuds1, line 65: Removing what looks like an unnecessary parens pair:
```suggestion
				So(query.Result[0].EmailSentOn.UTC(), ShouldHappenOnOrAfter, query.Result[0].Created.UTC())
```

pkg/services/sqlstore/temp_user_test.go
```diff
@@ -62,7 +62,21 @@ func TestTempUserCommandsAndQueries(t *testing.T) {
...
+			})
+
+			Convey("Should be able expire temp user", func() {
+				cmd4 := models.ExpireTempUsersCommand{OlderThan: timeNow()}
```
aknuds1, line 69: Why is it called `cmd4`, as opposed to e.g. `cmd`?

pkg/models/temp_user.go
```diff
@@ -35,8 +36,8 @@ type TempUser struct {
 	Code        string
 	RemoteAddr  string
 
-	Created time.Time
-	Updated time.Time
+	Created int64
```
aknuds1, line 39: Why is it necessary to change the type?

marefr: Please see thread https://github.com/grafana/grafana/pull/27361#pullrequestreview-488743815

pkg/services/cleanup/cleanup.go
```diff
@@ -138,3 +138,16 @@ func (srv *CleanUpService) deleteOldLoginAttempts() {
...
 	}
 }
+
+func (srv *CleanUpService) expireOldUserInvites() {
+	maxInviteLifetime := time.Duration(srv.Cfg.UserInviteMaxLifetimeDays) * 24 * time.Hour
+
+	cmd := models.ExpireTempUsersCommand{
+		OlderThan: time.Now().Add(-maxInviteLifetime),
+	}
+	if err := bus.Dispatch(&cmd); err != nil {
```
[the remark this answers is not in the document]
marefr: We decided that using the bus for this is fine for now.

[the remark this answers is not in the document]
aknuds1: @marefr OK, thanks for the clarification.

pkg/services/sqlstore/temp_user.go
wbrowne: Omitting the `else` actually results in both `err` and `result` as unresolved references 🕵️  Since they no longer belong to the same block once splitting

aknuds1: Ah, I didn't take into account that `result` is defined on line 140 and used on line 142. With my change though, `err` should be defined at least.

pkg/services/sqlstore/temp_user_test.go
wbrowne: Ah yes I was confused about this originally too but seems to be a pattern followed in the preceding tests. There's an original `cmd` for setting up the TempUser and then subsequent nested tests could be changed to something more meaningful / less distracting. WDYT?

aknuds1: You probably have the right idea :) My main concern I guess is that the naming is confusing.

pkg/setting/setting.go
```diff
@@ -304,6 +304,9 @@ type Cfg struct {
 
 	DateFormats DateFormats
 
+	// User
+	UserInviteMaxLifetimeDays int
```
aknuds1, line 308: I'm wondering if this should be made into a duration type field, that is configured via a string such as "2d" or "2h"? We might want to bound the supported lower and upper bound. It sounds to me as if it would make for a more pleasant and granular user experience, although I'm interested in other people's POV.

marefr: Yes, we use that for other settings (I think).

aknuds1 (approved):
Looks good AFAICT. @marefr What do you think?

conf/sample.ini
```diff
@@ -274,15 +274,18 @@
...
+# The duration in time a user invite can be accepted before expiring. This setting should be expressed as a duration. Examples: 6h (hours), 2d (days), 1w (week). Default is 24h (24 hours). The minimum supported duration is 1h (1 hour), and the maximum is 1w (1 week).
+;user_invite_max_lifetime_duration = 24h
+
 [auth]
 # Login cookie name
 ;login_cookie_name = grafana_session
 
-# The maximum lifetime (duration) an authenticated user can be inactive before being required to login at next visit. Default is 7 days (7d). This setting should be expressed as a duration, e.g. 5m (minutes), 6h (hours), 10d (days), 2w (weeks), 1M (month). The lifetime resets at each successful token rotation 
-;login_maximum_inactive_lifetime_duration = 
+# The maximum lifetime (duration) an authenticated user can be inactive before being required to login at next visit. Default is 7 days (7d). This setting should be expressed as a duration, e.g. 5m (minutes), 6h (hours), 10d (days), 2w (weeks), 1M (month). The lifetime resets at each successful token rotation
```
oddlittlebird, line 284:
```suggestion
# The maximum lifetime (duration) an authenticated user can be inactive before being required to login at next visit. Default is 7 days (7d). This setting should be expressed as a duration, e.g. 5m (minutes), 6h (hours), 10d (days), 2w (weeks), 1M (month). The lifetime resets at each successful token rotation.
```

docs/sources/administration/configuration.md
```diff
@@ -601,6 +601,12 @@ Default is `false`.
 Editors can administrate dashboards, folders and teams they create.
 Default is `false`.
 
+### user_invite_max_lifetime_duration
+
+The duration in time a user invite can be accepted before expiring.
```
oddlittlebird, line 606:
```suggestion
The duration in time a user invitation remains valid before expiring. 
```

oddlittlebird: If you accept this change, please update it elsewhere as well.

oddlittlebird (commented):
A few editing suggestions.

pkg/services/sqlstore/temp_user.go
```diff
@@ -132,3 +133,15 @@ func GetTempUserByCode(query *models.GetTempUserByCodeQuery) error {
 	query.Result = &tempUser
 	return err
 }
+
+func ExpireOldUserInvites(cmd *models.ExpireTempUsersCommand) error {
```
kylebrandt, line 137: Do these get cleaned up (Deleted from db at some point?)

wbrowne: Conversation was had outside of GH but following up here for completeness:

Currently in its current state with this PR, the rows will only ever be marked as expired and never deleted. The original solution a few commits back was explicitly deleting the rows, but that was since changed to be more consistent with the existing pattern based on feedback. I'm happy with either solution, but had opted for purging initially because when the temp_user goes to status `expired`, it's essentially a dead row by that point and can't/shouldn't be revived.

It would be possible to include also deleting as part of this PR - maybe the job will first mark rows as expired, and then secondly delete expired rows that haven't been updated in 1 week (or whatever)

marefr: We can potentially add delete separately. One idea later could also be to show expired invites for tracking purpose. Would also allow to "re-send" the invite. But I'm not sure it makes sense.

pkg/setting/setting.go
```diff
@@ -1056,6 +1059,20 @@ func readUserSettings(iniFile *ini.File, cfg *Cfg) error {
...
+	userInviteMaxLifetimeDuration, err := gtime.ParseInterval(userInviteMaxLifetimeVal)
+	if err != nil {
+		return err
+	}
+
+	// Minimum supported lifetime is 1 hour, and the maximum is 1 week
+	cfg.UserInviteMaxLifetime = userInviteMaxLifetimeDuration
+	if cfg.UserInviteMaxLifetime < time.Hour*1 {
+		cfg.UserInviteMaxLifetime = time.Hour * 1
+	} else if cfg.UserInviteMaxLifetime > time.Hour*24*7 {
```
marefr, line 1072: Not sure if maximum allowed is needed. What's a good reason for keeping it?

marefr (changes requested):
Added a comment. I checked out the code and ran it. After the migration I have tmp_user table with rows having created and updated columns set to zero. Is this a potential problem? Basically, if you have outstanding invites they will be changed to expired by cleanup job when running this?

pkg/setting/setting.go
wbrowne: This came up before and it was mostly about having sane values as user invites are something that should expire within `<some reasonable amount of time>`. This varies depending on your use case of Grafana / threat model of course - 1 week may be too short, but 3 months seems like a long time. I'd say what's the harm in a maximum? 😏

pkg/services/sqlstore/temp_user.go
wbrowne: Sounds good to me. Let's revisit

pkg/setting/setting.go
marefr: We provide the default (24h) according to what we recommend is a good sane default. We shouldn't underestimate our users use cases and don't limit the maximum and let them change that to whatever they want. We've tried to limit certain things before with same reasoning as you provide and it came back and bit us eventually.

But it's no biggie really. I can live with this and instead wait and see if someone has a problem with it and then we may have to remove the limitation .

pkg/services/sqlstore/migrations/temp_user.go
```diff
@@ -44,4 +50,78 @@ func addTempUserMigrations(mg *Migrator) {
...
+	}
+
+	for _, user := range users {
+		created := time.Now().Unix()
+		if _, err := sess.Exec("UPDATE "+mg.Dialect.Quote("temp_user")+
+			" SET created = ?, updated = ? WHERE id = ?", created, created, user.Id); err != nil {
+			return err
+		}
+	}
+	return nil
```
marefr, line 126: I think a single update clause should work here:
```sql
UPDATE temp_user set created = ?, updated = ? WHERE created = 0 AND status in ('SignUpStarted', 'InvitePending')"
```

WDYT?

marefr (changes requested):
Great. This approach should work I think :+1: Added a suggestion though to simplify it.

pkg/setting/setting.go
aknuds1: It's a good point that we can just remove the limitation, shouldn't bother anyone if we end up doing that?

kylebrandt: If we are going to have a defined ranged. Then we should error if the config is outside of the range instead of changing it (since the config is not valid).

marefr (changes requested):
Added a couple suggestions. Not sure you agree? Besides those looks great

pkg/services/sqlstore/migrations/temp_user.go
```diff
@@ -44,4 +49,65 @@ func addTempUserMigrations(mg *Migrator) {
...
+type SetCreatedForOutstandingInvites struct {
+	MigrationBase
+}
+
+func (m *SetCreatedForOutstandingInvites) Sql(dialect Dialect) string {
+	return "code migration"
+}
+
+func (m *SetCreatedForOutstandingInvites) Exec(sess *xorm.Session, mg *Migrator) error {
+	created := time.Now().Unix()
```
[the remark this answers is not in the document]
wbrowne: Yeah that's a good point - might be strange to have a date in the future, but at least this will ensure everyone will some reasonable amount of time in the off chance any open invites are migrated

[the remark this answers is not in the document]
marefr: > might be strange to have a date in the future

That's very true. Didn't think about this. So when setting created to "now" it will expire after 24 hours, per default. I suggest to revert that then. Sorry for the confusion.

pkg/services/sqlstore/migrations/temp_user.go
```diff
@@ -44,4 +49,65 @@ func addTempUserMigrations(mg *Migrator) {
...
+type SetCreatedForOutstandingInvites struct {
+	MigrationBase
+}
+
+func (m *SetCreatedForOutstandingInvites) Sql(dialect Dialect) string {
+	return "code migration"
+}
+
+func (m *SetCreatedForOutstandingInvites) Exec(sess *xorm.Session, mg *Migrator) error {
+	created := time.Now().Add(24 * time.Hour).Unix()
```
marefr, line 107:
```suggestion
	created := time.Now().Unix()
```

marefr (changes requested):
Revert my earlier suggestion

pkg/services/sqlstore/temp_user.go
```diff
@@ -132,3 +133,15 @@ func GetTempUserByCode(query *models.GetTempUserByCodeQuery) error {
...
+	return inTransaction(func(sess *DBSession) error {
+		var rawSql = "UPDATE temp_user SET status = ?, updated = ? WHERE created <= ? AND status = ?"
+		if result, err := sess.Exec(rawSql, string(models.TmpUserExpired), time.Now().Unix(), cmd.OlderThan.Unix(), string(models.TmpUserInvitePending)); err != nil {
```
wbrowne, line 140: @marefr We also want to check for `models.TmpUserSignUpStarted` here too, right?

marefr: Yeah that's a good point :+1: Can you fix?

marefr (approved):
LGTM great :+1:

dwyer125:
How can we change the length of time the invite is valid for? 
24 hours is to short. 1 week would be suitable.

PR #27362, opened 2020-09-03
Transformations: merge will properly handle empty frames and frames with multiple rows where values are overlapping.

### LINKED ISSUE #26890: Merge transform not working if one of the query return empty response
I am trying to merge multiple Prometheus instant queries into into single table (like 6.x did by default) which is prerequisite to get "Organize Fields" to work (Organize fields only works with a single frame). Intent is to list all alerts in single table, multiple queries needed to filter the alerts properly using template variables, and then sort and filter some fields using Organize fields.

**Current behavior:**
When all queries return one or more entry then merge works in 7.1.0 and in 7.1.2.

But if one of the query return empty response then merge is not done at all and each query is listed separately, and only one can be shown in table.

**Expected behavior:**
Queries should be merged and empty response should be ignored. Otherwise merge will not work with dynamic data like alerts where some queries might be are empty.

**Screenshots from merge transform debug:**
Two instant query are merged: One instant query result 5 entries and another 2 entries resulting 7 entries (Not merged to single line because some label values are different, and in some queries also some labels are different)
image

**Empty response from query inspector:**
image

Three query when one is empty are not merged: One instant query result 5 entries, another 2 entries but third query is empty.
image

**Repro step:**
Create two or more query which are correctly merged. Then clone one query and add query parameter which result empty result like {noMatchLabel="noMatch"}

**Environment**:
- Grafana version 7.1.0 and 7.1.2:
- OS Grafana is installed on: Windows and Linux
- User OS & Browser: Chrome
- Grafana plugins:
- Others:

### LINKED ISSUE #26908: "Merge" transformer is overly simplistic, does not replicate "old-table" merge
**What happened**: Using the image below to explain. The first query aggregates on Country and Age, and generates a metric "Sum", the second query aggregates on just "Age", and generates a metric "Count".

image

As can be seen, the merge only works for the first instances of each "AgeGroup".

**What you expected to happen**: The desired functionality was present in the "old-table", where the metric generates form the second query would be merged into every matching row of the first query. "Count" should be present on every row of "AgeGroup", not just the first instances of each value.

**How to reproduce it (as minimally and precisely as possible)**: So far as I can tell, the test datasources used on the live demo site are all too limited to demonstrate this (I consider this limitation as a bit of a bug in and of itself). All that is needed to reproduce the problem is a datasource that can generate 2 different dimensions (with enough variety so the overlap between them is not 1:1), and 2 different metrics. Then try to merge the two queries as explained above.

**Anything else we need to know?**: This is a regression from "old-table", one that is preventing us from upgrading many of our panels.

- Grafana version: 7.1

mckn:
This PR resolves the issue reported in the following issues: #26908, #26890.

---------------------------------------------------------

Prior to this fix, incoming data with the following format:

**Query A:**
| Country       | AgeGroup   | Sum  |
|---------------|------------|------|
| United States | 50 or over | 998  |
| United States | 35 - 49    | 1193 |
| Mexico        | 0 - 17     | 1675 |
| Germany       | 35 - 49    | 146  |
| Canada        | 35 - 49    | 166  |
| Canada        | 25 - 34    | 219  |

**Query B:**
| AgeGroup   | Count |
|------------|-------|
| 0 - 17     | 1     |
| 18 - 24    | 3     |
| 25 - 34    | 2     |
| 35 - 49    | 4     |
| 50 or over | 2     |

Did not properly merged right hand side values in to all matching row and the result was the following:

| Country       | AgeGroup   | Sum  | Count |
|---------------|------------|------|-------|
| United States | 50 or over | 998  | 2     |
| United States | 35 - 49    | 1193 | 4     |
| United States | 25 - 34    | 1580 | 2     |
| Mexico        | 0 - 17     | 1675 | 1     |
| Germany       | 35 - 49    | 146  |       |
| Canada        | 35 - 49    | 166  |       |
| Canada        | 25 - 34    | 219  |       |
|               | 18 - 24    |      | 3     |

After applying this fix, the merge will return the following:

| Country       | AgeGroup   | Sum  | Count |
|---------------|------------|------|-------|
| United States | 50 or over | 998  | 2     |
| United States | 35 - 49    | 1193 | 4     |
| United States | 25 - 34    | 1580 | 2     |
| Mexico        | 0 - 17     | 1675 | 1     |
| Germany       | 35 - 49    | 146  | 4     |
| Canada        | 35 - 49    | 166  | 4     |
| Canada        | 25 - 34    | 219  | 2     |
|               | 18 - 24    |      | 3     |

---------------------------------------------------------

It will also handle the scenario where all incoming data frames are empty. In this scenario they will be merged into one single empty frame.

If only a couple of the incoming data frames are empty they will be removed prior to doing the merge. So all the frames containing data will be merged into one single frame.

---------------------------------------------------------

Fixes #26908
Fixes #26890


packages/grafana-data/src/transformations/transformers/merge.test.ts
```diff
@@ -196,6 +196,127 @@ describe('Merge multipe to single', () => {
     expect(unwrap(result[0].fields)).toEqual(expected);
   });
 
+  it('combine two tables, where first is partial overlapping, into one', () => {
+    const cfg: DataTransformerConfig<MergeTransformerOptions> = {
```
peterholmberg, line 200: Just a nit, maybe move this inside the describe scope to avoid duplication?

peterholmberg (approved):
Nice work, justa small nit :slightly_smiling_face:

packages/grafana-data/src/transformations/transformers/merge.test.ts
mckn: Awesome, feedback! I will update. Thanks 🙏

PR #27365, opened 2020-09-03
Loki: Show loading state longer when running query

zoltanbedi:

@slim-bean raised the issue that the spinner in the right corner is only showing for a short period of time but the query to show up takes longer.


public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -95,7 +95,12 @@ export class LokiDatasource extends DataSourceApi<LokiQuery, LokiOptions> {
 
     filteredTargets.forEach(target =>
       subQueries.push(
-        this.runInstantQuery(target, options, filteredTargets.length),
+        this.runInstantQuery(target, options, filteredTargets.length).pipe(
```
aocenas, line 98: Hmm I think this should be done in Explore itself. First there is no guaranties that instant query will finish first. Also the problem seems to be generic in the sense that the same issue may be in other datasources too and with other combinations of queries.

Explore should have something like `loading = queries.some(q => x.isLoading)`.

torkelo: in the dashboard to optimize performance and not render partial results as every query completes we wait to render until we get a Done.

if data source does not return a state we set state to Done automatically after first result:
https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/state/runRequest.ts#L44 

it's a bit tricky for Grafana to know what state should be as some queries do not return any results, and some queries trigger multiple requests and results. So not sure where this logic is best placed

torkelo: think in Prometheus we have different logic for queries from explore and dashboard, as in explore we return results as they complete and in dashboard Prometheus will wait for all queries to complete

public/app/plugins/datasource/loki/datasource.ts
```diff
@@ -98,7 +98,14 @@ export class LokiDatasource extends DataSourceApi<LokiQuery, LokiOptions> {
 
     for (const target of filteredTargets) {
       if (options.app === CoreApp.Explore) {
-        subQueries.push(this.runInstantQuery(target, options, filteredTargets.length));
+        subQueries.push(
+          this.runInstantQuery(target, options, filteredTargets.length).pipe(
+            map(response => {
+              // Let rangeQuery finish the loading
+              return { ...response, state: LoadingState.Loading };
```
aocenas, line 105: What if instant query returns after range query? Would explore be stuck in loading state?

zoltanbedi: You are right it would.

zoltanbedi:
Closing this as this is not a proper fix for #28718

PR #27367, opened 2020-09-03
InfluxDB/flux: support queries without a time

ryantxu:
Extracted from #26634

There are a few context where a query is sent without a valid time -- this happens for template variable queries that execute before the dashboard has loaded (and have no time context)

In normal plugins this is fine and the times just initialize to whatever time initializes to.  In the flux plugin, it throws an error, but should not

### DIFF
--- a/public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
+++ b/public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
@@ -60,7 +60,7 @@ v1.measurements(bucket: v.bucket)`,
     label: 'Schema Exploration: (fields)',
     description: 'Return every possible key in a single table',
     value: `from(bucket: v.bucket)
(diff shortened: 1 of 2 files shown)


public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
```diff
@@ -60,7 +60,7 @@ v1.measurements(bucket: v.bucket)`,
     label: 'Schema Exploration: (fields)',
     description: 'Return every possible key in a single table',
     value: `from(bucket: v.bucket)
-  |> range(start: v.timeRangeStart, stop:timeRangeStop)
+  |> range(start: v.timeRangeStart, stop:v.timeRangeStop)
```
ryantxu, line 63: this missing `v.` is a typo in our suggested queries. 

I added the change here just to reduce the size of #26634

ryantxu:
Closing in favor of #27370

PR #27369, opened 2020-09-03
Units: add new flow unit for cubic meters/hour (m³/h)

### LINKED ISSUE #16843: Add cubic meter per hour as a unit in flow category
**Flow rate unit - cubic meter per hour**:

**It's an industry standard unit for measuring flow**:

rlindner:
This PR adds a new flow unit for cubic meters/hour (m³/h).

Fixes https://github.com/grafana/grafana/issues/16843

I've considered using the specific unicode symbol for cubic meters (㎥), but then decided to use the better readable "m³". So it's also in line with the existing usage of "m²".

### DIFF
--- a/devenv/dev-dashboards-without-uid/panel_tests_polystat.json
+++ b/devenv/dev-dashboards-without-uid/panel_tests_polystat.json
@@ -959,6 +959,10 @@
               "value": "flowcms"
             },
             {
+              "text": "Cubic meters/hour (m³/h)",
+              "value": "flowcmh"
+            },
+            {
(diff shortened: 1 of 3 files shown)


hugohaggmark:
Hi @rlindner, since the reporter filed the original issue you can now use custom units instead: `suffix: m³/h` to solve this, so unfortunately I'm going to close this PR and the issue. Thank you for your efforts.

See [https://grafana.com/docs/grafana/latest/panels/field-configuration-options/#custom-units](https://grafana.com/docs/grafana/latest/panels/field-configuration-options/#custom-units)

PR #27375, opened 2020-09-04
Annotations: add standard annotations support (and use it for flux queries)

ryantxu:
Alternative approach from #26634

Rather than make a component that things can explicitly include, this will try to have a good behavior for everything that does not explicitly want the angular annotations UI.  The immediate goal is to get annotations working with flux.

In this iteration, when a datasource defines an `annotations` property, the new lifecycle will be execute.  Nothing needs to be defined unless it needs custom processing.  Eventually we should make this avaliable to any datasource, but for now it still needs the "annotations" meta property

image


packages/grafana-data/src/types/datasource.ts
```diff
@@ -268,10 +268,20 @@ export abstract class DataSourceApi<
...
+
+  /**
+   * Can be optionally implemented to allow datasource to be a source of annotations for dashboard.
+   * This function will only be called if an angular {@link AnnotationsQueryCtrl} is configured and
+   * the {@link annotationProcessor} is undefined
+   *
+   * @deprecated -- prefer using `annotationProcessor`
    */
-  annotationQuery?(options: AnnotationQueryRequest<TAnno>): Promise<AnnotationEvent[]>;
+  annotationQuery?(options: AnnotationQueryRequest<TQuery>): Promise<AnnotationEvent[]>;
```
ryantxu, line 284: This goes back to the old (pre 7.2) and incorrect types from before.  but that allows us to change the base class without conflicts

public/app/features/annotations/editor_ctrl.ts
```diff
@@ -69,6 +72,14 @@ export class AnnotationsEditorCtrl {
     });
   }
 
+  /**
+   * Called from the react editor
+   */
+  onAnnotationChange(annotation: any) {
+    console.log('SAVE THE annotation!', annotation, this);
```
ryantxu, line 79: @torkelo -- its too soon for a real review, but holler if this rings any bells for you:

image

any hints on what to check?

lots going on with angular > react > angular madness!

packages/grafana-data/src/types/annotations.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * @alpha -- any value other than `field` is experimental
+ */
+export enum AnnotationEventFieldSource {
+  Field = 'field', // Default -- find the value with a matching key
+  Text = 'text', // Write a constant string into the value
+  Skip = 'skip', // Do not include the field
+}
+
+export interface AnnotationEventFieldMapping {
```
ryantxu, line 51: currently, we are only using the `field` structure, but I want to make sure we can also support setting the text directly and with template variables.

packages/grafana-data/src/types/datasource.ts
```diff
@@ -267,13 +267,23 @@ export abstract class DataSourceApi<
...
+
   /**
-   * Can be optionally implemented to allow datasource to be a source of annotations for dashboard. To be visible
-   * in the annotation editor `annotations` capability also needs to be enabled in plugin.json.
+   * An annotation processor allows explict control for how annotations are managed.
+   *
+   * It is only necessary to configure an annotation processor if the default behavior is not desirable
    */
-  annotationQuery?(options: AnnotationQueryRequest<TAnno>): Promise<AnnotationEvent[]>;
+  annotations?: AnnotationSupport<TQuery>;
```
ryantxu, line 277: when an `annotations` property is defined it will trigger the new behavior.

Perhaps we should use a less common name?  I don't have a good sense if this is a real risk or not

packages/grafana-data/src/types/datasource.ts
```diff
@@ -541,20 +545,19 @@ export interface DataSourceSelectItem {
 
 /**
  * Options passed to the datasource.annotationQuery method. See docs/plugins/developing/datasource.md
+ *
+ * @deprecated -- use {@link AnnotationSupport}
  */
-export interface AnnotationQueryRequest<TAnno = {}> {
+export interface AnnotationQueryRequest<MoreOptions = {}> {
```
ryantxu, line 551: reverting the changes from #27342 -- this makes things the same as they were in 7.1

public/app/features/annotations/annotations_srv.ts
```diff
@@ -195,4 +210,64 @@ export class AnnotationsSrv {
   }
 }
 
+export function executeAnnotationQuery(
```
ryantxu, line 213: this is an internal function now -- but when we revisit the annotations query path (and expose it to react panels!), we may want to expose `AnnotationQueryResponse` in grafana/data

public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
```diff
@@ -12,12 +12,10 @@ import {
...
 } from '@grafana/ui';
 import { getTemplateSrv } from '@grafana/runtime';
+import InfluxDatasource from '../datasource';
 
-interface Props {
-  target: InfluxQuery;
-  change: (target: InfluxQuery) => void;
-  refresh: () => void;
-}
+// @ts-ignore -- complicated since the datasource is not really reactified yet!
```
ryantxu, line 17: This is not a real `QueryEditor` -- since influx is still an angular based datasource.

public/app/plugins/datasource/influxdb/datasource.ts
```diff
@@ -183,7 +200,7 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
   async annotationQuery(options: AnnotationQueryRequest<any>): Promise<AnnotationEvent[]> {
     if (this.isFlux) {
       return Promise.reject({
-        message: 'Annotations are not yet supported with flux queries',
+        message: 'Flux requires the standard annotation query',
```
ryantxu, line 203: should never get to this execution path

ryantxu:
@torkelo @dprokop -- this is ready for a more in depth review.

The UI is my MVP, it could be much better, but at this stage I am mostly worried about the JSON model and query pipeline

public/app/plugins/datasource/influxdb/datasource.ts
```diff
@@ -55,6 +57,21 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
...
+      // When flux, attach an annotation processor
+      this.annotations = {
+        QueryEditor: FluxQueryEditor,
+
+        // Do now send an empty query
+        prepareQuery: (anno: AnnotationQuery<InfluxQuery>) => {
+          if (!anno.target?.query) {
+            return undefined;
+          }
+          return anno.target;
```
torkelo, line 71: not sure I understand this part, the undefined for not sending empty qiuery, can't the main query function also do that? 
And what is the anno.target part and how does it relate to the flux query model, is anno.target a query string? or a query model?

torkelo (commented):
Think this looks good, will look in more detail tomorrow

public/app/plugins/datasource/influxdb/datasource.ts
ryantxu: just updated so that we use a filter callback :+1:

torkelo:
Tested this PR, with a Prometheus data source I get short browser freeze and a max recursion error. 

Screenshot from 2020-09-09 09-43-48

ryantxu:
@torkelo -- I updated it so that it only loads the react panel when it will use it.  This avoids angular->react->angular:

```html
    <h5 class="section-heading">Query</h5>
    <rebuild-on-change property="ctrl.currentDatasource">
      <!-- Legacy angular -->
      <plugin-component ng-if="!ctrl.currentDatasource.annotations" type="annotations-query-ctrl"> </plugin-component>

      <!-- React query editor -->
      <standard-annotation-editor
        ng-if="ctrl.currentDatasource.annotations"
        annotation="ctrl.currentAnnotation"
        datasource="ctrl.currentDatasource"
        change="ctrl.onAnnotationChange" />
    </rebuild-on-change>
```

packages/grafana-data/src/types/annotations.ts
```diff
@@ -0,0 +1,87 @@
...
+   */
+  prepareQuery?(anno: TAnno): TQuery | undefined;
+
+  /**
+   * When the standard frame > event processing is insufficient, this allows explicit control of the mappings
+   */
+  processEvents?(anno: TAnno, data: DataFrame): AnnotationEvent[] | undefined;
+
+  /**
+   * Specify a custom QueryEditor for the annotaiton page.  If not specified, the standard one will be used
```
dprokop, line 84:
```suggestion
   * Specify a custom QueryEditor for the annotation page.  If not specified, the standard one will be used
```

public/app/features/annotations/standardAnnotationSupport.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+  if (rsp.series.length === 1) {
+    return rsp.series[0];
+  }
+
+  return transformDataFrame(
+    [
+      {
+        id: 'seriesToColumns',
+        options: { byField: 'Time' },
```
dprokop, line 66: Ah, using the same assumption (name: Time) in graph panel work. Do you consider this safe? I don't think we gurantee the name of the time field, do we?

dprokop (commented):
This is my first more serious look into annotations and I find this a really good refactor. I like the direction and the API encapsulating datasource's annotaitions config under the `annotations` "namespace" in the DataSourceApi

public/app/features/annotations/standardAnnotationSupport.ts
ryantxu: oh right -- i copied this from you ;)  it would be better to have a function that does this for us in a consistent way.

public/app/features/annotations/standardAnnotationSupport.ts
```diff
@@ -0,0 +1,188 @@
...
+    placeholder: 'time, or the first time field',
+  },
+  { key: 'timeEnd', help: 'When this field is defined, the annotation will be treated as a range' },
+  {
+    key: 'title',
+  },
+  {
+    key: 'text',
+    field: (frame: DataFrame) => frame.fields.find(f => f.type === FieldType.string),
+    placeholder: 'text, or the first text field',
```
torkelo, line 103: currently, the preferred annotation event structure only has a text & tags field, the title field was deprecated many years ago (but I see that it still there in some annotation editors), Thought we deprecated it and only showed it if was configured but seems like that is not the case. Think Elasticsearch & InfluxDB only show the title field if it's already defined in the annotation (for backward compatibility).

torkelo: As for rendering text & title just get's appended in the annotation tooltip body

public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
torkelo: Is there an argument to split InfluxDB & Flux into two datasource plugins? (that share some code). Since it looks pretty binary with not being able to mix flux/InfluxQL in the same data source instance anyway.

public/app/features/annotations/standardAnnotationSupport.ts
ryantxu: So what should we expose?  Just text?


FYI -- plugins are currently blindly filling in as many of the fields as possible.  There is no indication anything is deprecated

public/app/plugins/datasource/influxdb/components/FluxQueryEditor.tsx
ryantxu: Yes -- there is a good argument, but let's discuss

public/app/features/annotations/standardAnnotationSupport.ts
torkelo: think it's better to use merge transform as it more automatically merges data (seriesToColumns also has a big limitation in that it does not handle duplicate rows for the same time value)

public/app/features/annotations/standardAnnotationSupport.ts
torkelo: yea, I know. This was removed a long time ago before we had types and deprecation warnings. 

I think we should only expose text for now as both title and text are just concatenated anyway when we render annotations.

torkelo (approved):
Looks like Loki would be a great next step to use this, as it is currently going through hoops to use the normal query editor as the annotation editor:

https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/loki/partials/annotations.editor.html

Question is should we include this in 7.2 or 7.3 (7.3 is a short release, and gives us some time to test and refine the UI).

public/app/features/annotations/standardAnnotationSupport.ts
dprokop: Not sure about merge TBH. Doesn't it work in a way that if you have two series:
A: [Time, Value]
B: [Time, Value, Value1]

it will result in :
C: [Time,Value,Value1]

@mckn correct my if I'm wrong ^^

mckn: @dprokop yes - but only if the values of Time, Value are exactly the same. Then the merge will consider those "lines" as being the same and hence merge them to one.

ryantxu:
Discussed with @torkelo and we will merge with 7.2 -- in 7.3 we will use this for more datasources and improve the UI/UX

PR #27376, opened 2020-09-04
Time format: 0 seconds is a number

kvc0:
When formatting a value, e.g., in the Stat panel, with a unit of Seconds,
Grafana chooses "NaN ns" when attempting to format 0 seconds.

0, however, is a number; and a quantity of 0 seconds is a perfectly
cromulent notion.  A quantity of 0 seconds should be presented as "0 s"
rather than "NaN ns"

```
>> Test Suites: 494 passed, 494 total
>> Tests:       6 skipped, 4061 passed, 4067 total
>> Snapshots:   179 passed, 179 total
>> Time:        167.998 s
>> Ran all test suites.

Running "no-only-tests" task

Running "no-focus-convey-tests" task

Done.
✨  Done in 227.62s.
```

Renders `select 0` as "0 s" when unit is Seconds  /  Because `select 0` does not return `NaN ns` when interpreted as "seconds."

A quick pass through every other time option shows them rendering 0 as "0 <unit>" as a reasonable person might otherwise naively assume.  S is special.

Before:
image

After:
image

### DIFF
--- a/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
+++ b/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
@@ -102,6 +102,9 @@ export function toSeconds(size: number, decimals?: DecimalCount, scaledDecimals?
   if (size === null) {
     return { text: '' };
   }
+  if (size === 0) {
+    return { text: '0 s' };
+  }
 
   // Less than 1 µs, divide in ns
   if (Math.abs(size) < 0.000001) {
--- a/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
+++ b/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
@@ -102,6 +102,8 @@ export function toSeconds(size: number, decimals?: DecimalCount, scaledDecimals?
   if (size === null) {
     return { text: '' };
   }
+
+  // If 0, use s unit instead of ns
   if (size === 0) {
     return { text: '0', suffix: ' s' };
   }
--- a/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
+++ b/packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
@@ -103,7 +103,7 @@ export function toSeconds(size: number, decimals?: DecimalCount, scaledDecimals?
     return { text: '' };
   }
   if (size === 0) {
-    return { text: '0 s' };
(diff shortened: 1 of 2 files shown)


packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
```diff
@@ -102,6 +102,9 @@ export function toSeconds(size: number, decimals?: DecimalCount, scaledDecimals?
   if (size === null) {
     return { text: '' };
   }
+  if (size === 0) {
+    return { text: '0 s' };
```
ryantxu, line 106:
```suggestion
    return { text: '0', suffix:' s', numeric:0 };
```

Maybe

WarriorOfWire: `FormattedValue` has no such member "numeric" but sure if you'd like the suffix separated then why not

hugohaggmark:
Thank you for contributing @WarriorOfWire, could you add a test for this as well?

All the best

hugohaggmark (commented):
Looks good, missing a test for this case.

packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
```diff
@@ -102,6 +102,9 @@ export function toSeconds(size: number, decimals?: DecimalCount, scaledDecimals?
   if (size === null) {
     return { text: '' };
   }
+  if (size === 0) {
```
hugohaggmark, line 105: nitpick on formatting, needs a new line.
```suggestion
  
  if (size === 0) {
```

WarriorOfWire:
@hugohaggmark thanks for the review.  Here's the test suite results with 1 more test:

```
>> Test Suites: 494 passed, 494 total
>> Tests:       6 skipped, 4062 passed, 4068 total
>> Snapshots:   179 passed, 179 total
>> Time:        155.706 s
>> Ran all test suites.
```

packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
WarriorOfWire: added newline and comment.

ryantxu (approved):
LGTM - thank you!

hugohaggmark:
Thank you for contributing @WarriorOfWire ⭐

PR #27383, opened 2020-09-04
Graph Plugin: Variable as threshold

### LINKED ISSUE #159: Dynamic Grid Threshholds
Could Grid thresholds be fed dynamically as well? We feed some of our thresholds into graphite at work, it'd be useful to set the thresholds based on those metrics.

Cheers!

josteinHellerud:
Hi! 
 
The Graph plugin does not support template variables as thresholds. This is something we miss in my team, and based on open issues it seems like we are not alone.

This PR changes the following:

- Input field allows "any"
- Threshold user input is handled by a new function called `handleValueChange(index)`
- `handleValueChange(index)` checks if the input is NaN. If this is the case it tries to replace it as a variable. 
- Changes to variables will now also trigger a check for changes to threshold variable values with the new function `updateThresholds(panel)`

Some cases might need looking at:
- What should happen if a user changes the variable threshold by dragging with their mouse? Can we lock it so this will not happen?
- Error messages when user inputs NaN which cannot be matched as a variable. Right now the threshold value is set to NaN in this case. 

Fixes https://github.com/grafana/grafana/issues/159

This is my first time contributing to Grafana, so my apologies if I have done any rookie mistakes. I'v tried to read all the contributors guides, however, if I have overlooked something I'm sorry.


torkelo:
Thanks for contributing! 

Sadly major updates to the angular graph is not something that is likely to be merged, we are working on new React based graph panels and components that is using our new react based panel architecture and data model

torkelo:
So if you want dynamic thresholds, you should look at ways to add this to the field config thresholds

PR #27390, opened 2020-09-04
Stats: Use more efficient SQL and add timeouts

### LINKED ISSUE #27339: grafana issue long queries to mysql
**What happened**:
After I upgrade grafana to 7.1.5, grafana began to issue queries to mysql that never seems to end. (our script that check for long queries killed them after awhile)
And those query were making mysql use all server CPU.

This is the query that was causing issue :
`SELECT role AS bitrole, active, COUNT(role) AS count FROM
    (SELECT active, SUM(role) AS role
     FROM (SELECT
               u.id,
               CASE org_user.role
                   WHEN 'Admin' THEN 4
                   WHEN 'Editor' THEN 2
                   ELSE 1
                   END AS role,
               u.last_seen_at>'2020-07-29 13:37:46.94542' AS active
           FROM `user` AS u LEFT JOIN org_user ON org_user.user_id = u.id
           GROUP BY u.id, u.last_seen_at, org_user.role) AS t2
     GROUP BY active, id) AS t1
GROUP BY active, role`

upgrade to 7.1.5

For information, I have ~19 000 rows in 'user' and ~950 000 rows in 'org_user'.

**Environment**:
- Grafana version: 7.1.5
- OS Grafana is installed on: centos7
- database : mysql

sakjur:

`LEFT JOIN` causes `O(n*m)` operations, whereas `INNER JOIN` allows us to stay within `O(n)`.

I created 19000 users, a hundred organizations and ~90000 (1/10th of reporters amount) org-user connections in a local MySQL and get responses nearly instantaneously locally, but when running `explain` and comparing the INNER vs LEFT join it is clear that the inner join is preferable regardless of this.

In case _that_ doesn't work, I am also adding a cap for when Grafana will give up counting users to ensure reasonable performance.

Fixes #27339

### DIFF
--- a/pkg/services/sqlstore/stats.go
+++ b/pkg/services/sqlstore/stats.go
@@ -14,7 +14,7 @@ func init() {
 	bus.AddHandler("sql", GetDataSourceStats)
 	bus.AddHandler("sql", GetDataSourceAccessStats)
 	bus.AddHandler("sql", GetAdminStats)
-	bus.AddHandler("sql", GetUserStats)
+	bus.AddHandlerCtx("sql", GetUserStats)
 	bus.AddHandlerCtx("sql", GetAlertNotifiersUsageStats)
 	bus.AddHandlerCtx("sql", GetSystemUserCountStats)
 }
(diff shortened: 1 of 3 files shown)


marefr (approved):
LGTM. Great work

PR #27394, opened 2020-09-04
Templating: global/system variables should be properly replaced in templated values.

### LINKED ISSUE #27335: PanelHeader: $__dashboard.name does not work in panel title
**What happened**:
I'm using the $__dashboard in a panel, but instead of the dashboard name I see undefined
image

That the dashboard name will appear.

Create a panel and add $__dashboard to the title.

**Environment**:
- Grafana version: 7.1.5 (but I don't think it's a regression)
- OS Grafana is installed on: docker
- User OS & Browser: fedora/chrome
- Grafana plugins: N/A
- Others:

mckn:
Looks like we have lost support of using system/global template variables when replacing labels. In the issue referenced in this PR the templating of panel title stopped working when entering `${__dashboard.name}` as the panel title.

Fixes #27335

The code is kind of weird. In this example we want to use a system/global template variable as the interpolation value but as far as I kan see we never use the `templateSrv.setGlobalVariable` function (it is deprecated) to set the `templateSrv.grafanaVariables`. In the `templateSrv.replace*` function we are still trying to use the `templateSrv.grafanaVariables` list as the source.

I would love to have a look at this together with someone else that knows more about the different scenarios.


hugohaggmark (changes requested):
I must have missed this function when I moved system variables to Redux state. I don't think we need this function so I think we should deprecate it and have it call the replace function internally or remove it and just use templateSrv.replace instead? That said, I don't know all the special cases here either :)

torkelo:
Tricky, this is slightly different from the normal replace as this replaces with the text instead if value property (So show say All if all are selected , and will show text representation if multiple selected, or text representation if text and value are different)

hugohaggmark:
Thank you for clarifying @torkelo.

hugohaggmark (commented):
Don't think using `this.grafanaVariables` is a way forward, but maybe that would become a bigger PR. Would be happy to figure something out together with you @mckn

mckn:
@torkelo me and @hugohaggmark had a quick sync regarding this and came up with a possible way forward.

1. Extend the `replace` with a new format (`:text`) that we register in the format registry.
2. Deprecate the old `replaceText` functionality by adding a deprecation warning. Change so `replaceText` calls `replace` with the `:text` format internally.
3. Implement a `TextFormatter` and add some more tests for the `replace` functionality.

Would this work or do you see any issues with this approach?

torkelo:
I was thinking  the same, that maybe a format could be used so we can share the same replace function. 

TextFormatter could ignore value and return variable.current.text

public/app/features/templating/formatRegistry.ts
```diff
@@ -2,8 +2,14 @@ import kbn from 'app/core/utils/kbn';
 import { Registry, RegistryItem, VariableModel, textUtil, dateTime } from '@grafana/data';
 import { map, isArray, replace } from 'lodash';
 
+export interface FormatOptions {
+  value: any;
+  args: string[];
+  text: string;
+}
+
```
hugohaggmark, line 10: Nice ⭐

public/app/features/templating/formatRegistry.ts
```diff
@@ -184,13 +190,30 @@ export const formatRegistry = new Registry<FormatRegistryItem>(() => {
...
+        }
+
+        const current = (variable as any)?.current;
+
+        if (typeof current?.value === 'string') {
+          return current.value;
+        }
+        return current.text;
+      },
+    },
```
hugohaggmark, line 216: nit: is this clearer?
```suggestion
       if(!current) {
         return '';
       }
       
       const {value, text} = current;

        if (typeof value === 'string') {
          return value;
        }
        return text;
      },
    },
```

public/app/features/templating/template_srv.ts
```diff
@@ -275,30 +301,31 @@ export class TemplateSrv implements BaseTemplateSrv {
...
+      if (format) {
+        return match.replace(`${variable}:${format}`, `${variable}:text`);
+      }
 
-      return typeof value === 'string' ? value : variable.current.text;
+      if (match.indexOf(`$${variable}`) > -1) {
+        return match.replace(`$${variable}`, `\${${variable}:text}`);
+      }
+      return match.replace(`${variable}`, `${variable}:text`);
     });
```
hugohaggmark, line 326: Something about this feels hacky

mckn: Yeah, it kind of is :)

hugohaggmark (changes requested):
Skilled work and I think we're almost there. When using multi value variable in the title the output should be exactly as the Picker, i.e. `B + C` and not `B, C`.

Current version:
image

Grafana 6.7.4:
image

In an follow-up PR we need to:

- [x] Add new format to docs
- [x] Update the gdev dashboard devenv/dev-dashboards/feature-templating/global-variables-and-interpolation.json
- [x] Update the e2e test that runs on that test dashboard
- [x] Picker and formatter should use same function

mckn:
> Skilled work and I think we're almost there. When using multi value variable in the title the output should be exactly as the Picker, i.e. `B + C` and not `B, C`.
> 
> Current version:
> image
> 
> Grafana 6.7.4:
> image
> 
> In an follow-up PR we need to:
> 
> * [ ]  Add new format to docs
> * [ ]  Update the gdev dashboard devenv/dev-dashboards/feature-templating/global-variables-and-interpolation.json
> * [ ]  Update the e2e test that runs on that test dashboard
> * [ ]  Picker and formatter should use same function

If you do a pull and try again it should be fixed. But we need to do all the things that you added on the list above.

public/app/features/templating/template_srv.test.ts
```diff
@@ -95,6 +95,20 @@ describe('templateSrv', () => {
...
+      });
+      expect(target).toBe('this.mupp.filters');
+    });
+
+    it('should replace ${test.name} with variable text', () => {
+      const target = _templateSrv.replaceWithText('this.${test.name}.filters', {
+        test: { value: { name: 1 }, text: 'asd' },
+      });
+      expect(target).toBe('this.asd.filters');
+    });
```
torkelo, line 110: Not sure if this use case exists

public/app/features/templating/formatRegistry.ts
```diff
@@ -184,13 +190,41 @@ export const formatRegistry = new Registry<FormatRegistryItem>(() => {
...
+      description: 'Format variables in their text representation. Example in multi variable scenario A + B + C.',
+      formatter: (options, variable) => {
+        if (typeof options.text === 'string') {
+          return options.text;
+        }
+
+        const current = (variable as any)?.current;
+
+        if (!current) {
+          return '';
```
torkelo, line 212: can we return value here?

mckn: yeah - why not :)

public/app/features/templating/formatRegistry.ts
```diff
@@ -184,13 +190,41 @@ export const formatRegistry = new Registry<FormatRegistryItem>(() => {
...
+        }
+
+        const { value, text } = current;
+
+        if (Array.isArray(value)) {
+          return value.join(' + ');
```
torkelo, line 218: Should this not just return text? and if text is array do this

torkelo: but use text

public/app/features/templating/template_srv.test.ts
torkelo: this looks wrong to me, value.name only has one representation here (name), the text is on a different level

hugohaggmark (approved):
That was super fast and works as expected 🌟  and I agree that we should probably fix the list of things I added.

public/app/features/templating/template_srv.test.ts
```diff
@@ -595,6 +609,45 @@ describe('templateSrv', () => {
     });
   });
 
+  describe('replaceWithText can pass all / multi value', () => {
+    beforeEach(() => {
+      initTemplateSrv([
+        {
+          type: 'query',
+          name: 'server',
+          current: { value: ['server1', 'server2'] },
```
torkelo, line 618: this is missing text representation

public/app/features/templating/template_srv.ts
```diff
@@ -233,8 +255,10 @@ export class TemplateSrv implements BaseTemplateSrv {
 
       if (scopedVars) {
         const value = this.getVariableValue(variableName, fieldPath, scopedVars);
+        const text = this.getVariableText(variableName, fieldPath, scopedVars);
```
torkelo, line 258: this is a bit expensive (if there is a field path), would be nice to only have to do this if format is text. 

Maybe we can have a special logic here for format == text, and then pass that as value to this.formatValue

public/app/features/templating/template_srv.ts
```diff
@@ -219,6 +220,27 @@ export class TemplateSrv implements BaseTemplateSrv {
...
+    if (!scopedVar) {
+      return null;
+    }
+
+    if (fieldPath) {
+      const value = this.getFieldAccessor(fieldPath)(scopedVar.value);
+
+      if (typeof value?.text === 'string') {
+        return value.text;
+      }
```
torkelo, line 234: I do not think we should support this (we do not currently), and not sure we have a use case for this

public/app/features/templating/template_srv.ts
torkelo: not sure I understand this, I though this function was just to call this.replace(format, scopedVars, 'text');

torkelo (changes requested):
Added some comments, can jump on a call later to discuss it but feels like we cannot merge in for beta1

public/app/features/templating/template_srv.ts
hugohaggmark: I agree with @torkelo, we should just call this.replace(format, scopedVars, 'text') here instead?

mckn: ahh, I didn't see the forest for all the trees. Thanks 🙏

public/app/features/templating/template_srv.ts
mckn: hmm, I can see that we have a cache. Is it expensive to do regardless of it being cached?

public/app/features/templating/template_srv.test.ts
```diff
@@ -95,6 +95,13 @@ describe('templateSrv', () => {
       expect(target).toBe('this.asd.filters');
     });
 
+    it('should replace ${test.name} with scoped text', () => {
```
torkelo, line 98: this description is no longer valid (it replaces with value)

public/app/features/templating/template_srv.ts
```diff
@@ -275,30 +295,23 @@ export class TemplateSrv implements BaseTemplateSrv {
...
+      const format = fmt2 ?? fmt3;
+      if (!format) {
         return match;
       }
-
-      const value = this.grafanaVariables[variable.current.value];
-
-      return typeof value === 'string' ? value : variable.current.text;
+      return match.replace(`:${format}`, '');
     });
```
torkelo, line 312: Why is this still needed? if a user has specified a format then that should be used always. So just returning `this.replace(target, scopedVars, 'text');` should be enough

torkelo (commented):
Looks good!, still some unnecessary parsing I think  replaceWithText

public/app/features/templating/template_srv.ts
mckn: Or should we change that logic?

torkelo: yea, think that one is not correct (or can be changed). 

The use case for replaceWithText is really only in panel title, and legend / display name inputs. And the reason for all the other formats are to work with query syntaxes. So if someone want's glob formatting then that's what they get.

public/app/features/templating/template_srv.ts
```diff
@@ -248,8 +265,11 @@ export class TemplateSrv implements BaseTemplateSrv {
       }
 
       let value = variable.current.value;
+      let text = variable.current.text;
+
       if (this.isAllValue(value)) {
         value = this.getAllValue(variable);
+        text = 'All';
```
hugohaggmark, line 272: nitpick: I think there is a `const` for this

hugohaggmark (commented):
Looks the best, great that you changed the gdev-dash, docs and e2e!! 🌟

torkelo (approved):
Looks great!

PR #27395, opened 2020-09-04
Drone: Report front-end test metrics

aknuds1:

* Report front-end test metrics on Drone, via `grabpl test-frontend`
* Change CircleCI to run front-end tests via `grabpl test-frontend`
* Change scripts/circle-frontend-metrics.sh to not publish metrics, but print them JSON encoded (for grabpl test-frontend to pick up)

See also https://github.com/grafana/build-pipeline/pull/122.


torkelo (commented):
Cannot see where ci-frontend-metrics.sh is called from.  

Would expect that to be in the drone.yml / circle.yml file, something like 

./scripts/ci-frontend-metrics.sh | ./bin/grabpl publish-metrics 

I do not think it should be test-frontend task that calls this script, that feels backwards, and we want the publish-metrics task to be reusable for other CI flows right? 

Also the publish metrics task should only be for master builds, is there a condition for that?

aknuds1:
> Cannot see where ci-frontend-metrics.sh is called from.
>
> Would expect that to be in the drone.yml / circle.yml file, something like
> 
> ./scripts/ci-frontend-metrics.sh | ./bin/grabpl publish-metrics
> 
> I do not think it should be test-frontend task that calls this script, that feels backwards, and we want the publish-metrics task to be reusable for other CI flows right?

@torkelo The current design is that `grabpl test-frontend --publish-metrics` calls ./scripts/ci-frontend-metrics.sh to get the metrics, which it then publishes. Do you still find that problematic? 

I think it's a good design, since it makes the CI configuration simple (one command instead of several) and I don't see any inherent reason why `grabpl test-frontend` shouldn't have the intelligence to publish metrics, if asked to (`--publish-metrics`). The current solution, ./scripts/circle-test-frontend.sh, also publishes metrics after all.

Regarding publishing metrics for other tasks. I haven't got to implementing that yet, but there's a shared function within the pipeline tool, so adding logic for publishing metrics for more CI tasks won't be a problem.

> Also the publish metrics task should only be for master builds, is there a condition for that?

Yes, if you see the Drone and CircleCI configs, metrics are only published for master builds.

aknuds1:
@torkelo BTW, I also implemented the change you requested from `--strictNullChecks` to `--strict`, and fixed what I think are bugs in the use of `grep`.

torkelo:
> **@torkelo The current (most recent revision) design is that grabpl test-frontend --publish-metrics calls ./scripts/ci-frontend-metrics.sh to get the metrics, which it then publishes. Do you still find that problematic?**

Well, it feels like a hidden dependency that will make it hard to trace & understand. Maybe if the path to the script path is also passed in it would be at least able to find this dependency if I search the main repo for usage of `ci-frontend-metrics.sh`

> if asked to (--publish-metrics). The current solution, ./scripts/circle-test-frontend.sh, also publishes metrics after all.

But there is also an independent publishing script that can be used independently from any frontend-test. Can be used from any CI step. Just not sure publish metrics logic should be integrated into the frontend-test task. 

https://github.com/grafana/grafana/blob/master/scripts/ci-metrics-publisher.sh

> I think it's a good design, since it makes the CI configuration simple (one command instead of several), plus independent of the CI platform, which is a core idea

Maybe it should be a separate CI job the; named `frontend-metrics` as it's not part of PR & release flows, so is not always part of test-frontend job. 

I sort of question the need for the test-frontend build-pipeline task. all it's doing is executing yarn task, which one is not visible from main repo. So troubleshooting or searching for what commands are executed from "test-frontend" needs to go to different repo and find this go task to just see that it executes 4 yarn commands. 

So my recommendation is to split the test-frontend into 2 CI jobs

test-frontend: 
  yarn ci:test-frontend 

Or 

test-frontend: 
  grabpl test-frontend --yarnTasks="ci:test-frontend" 

and:

frontend-metrics:
  grabpl publish-metrics ./scripts/ci-frontend-metrics.sh


What do you think? Just want to make this as maintainable as possible and make the dependencies clear & intuitive

aknuds1:
@torkelo I will look into it.

aknuds1:
@torkelo Can you please have a look at the current revision? Implemented what you suggested.

scripts/ci-frontend-metrics.sh
```diff
@@ -1,18 +1,18 @@
 #!/bin/bash
+set -eo pipefail
 
 echo -e "Collecting code stats (typescript errors & more)"
 
 ERROR_COUNT_LIMIT=398
```
torkelo, line 6: think this needs to be around 580 or something (as we changed the error count command from --strictNullChecks to just --strict)

torkelo (approved):
Looks good!

scripts/ci-frontend-metrics.sh
aknuds1: @torkelo OK, will make the change.

aknuds1: Took care of it.

PR #27396, opened 2020-09-04
RFC - Graph NG API

dprokop:
Based on https://github.com/grafana/grafana/pull/21835

This PR is a result of experimentation with [uPlot](https://github.com/leeoniya/uPlot) as a potential replacement for jquery.flot.

It aims to allow Graph visualization composition rather than implementing every interaction in a single component as we currently try to do in the [react graph panel](https://github.com/grafana/grafana/blob/d42a7e827a013905187307dd8f9be71d954a028e/public/app/plugins/panel/graph2/GraphPanel.tsx)

Major components (naming is a subject of a change, mos def):

#### UPlotChart

Given a data frame(already aligned on the x-axis, currently on `Time` field only, using series to columns transformation) prepare [uPlot data model](https://github.com/leeoniya/uPlot/tree/master/docs#data-format), configure series, axes and scales based on field config (this is done on a very small scale now, needs way more work)

#### Plugins 

`UPlotChart` holds a map of plugins that uPlot needs to be initialized with. Via `PlotPluginsContext` I'm exposing an API for plugin registration. This API is available via `usePlotPluginContext` hook that exposes `registerPlugin` method - it has the same API as uPlot plugin(https://github.com/leeoniya/uPlot/blob/fe5fac73eaf1caf753a5eb9fea6b41cbc0a9bea4/dist/uPlot.d.ts#L199). Since uPlot plugin system allows hooking into uPlot lifecycle, plugins can manage their state on their own, not depending on some global state exposed from for instance UPlotChar. _This is an important decision I've made as I think it will simplify and make the architecture more modular_. I've started with exposing i.e. cursor position via context defined in UPlotChart, but the component quickly became massive again. I think the benefit can be observed in the sample plugins I've implemented that basically are a composition of the more primitive plugins like i.e. ClickPlugin, SelectionPlugin etc. Those primitives and additional plugins will handle the state independently from the UPlotChart. See code for more details.

#### Declarative API

First and foremost - axes, scales and series config will be driven primarily by the field configuration. 
Graphs functionalities (i.e. context menu, zoom) will be configured in a declarative way. I mean, if graph needs zoom functionality, you render ZoomPlugin as it's child
```
  <UPlotChart data={alignedData} timeRange={timeRange} width={width} height={height}>
      <ZoomPlugin
        onZoom={range => {
          onChangeTimeRange(range);
        }}
      />
   </UPlotChart>
```

under the hood, `ZoomPlugin` uses `SelectionPlugin` that is a uPlot plugin to selection hook. Using the same idea the `AnnotationsEditorPlugin` is implemented, using exactly the same plugin under the hood, but processing the selection lazily.

I've implemented plugins using render prop pattern, maybe it's worth exploring hooks approach.

To be solved:
1. Annotations interaction - how to use `setSelect` uPlot plugin hook and get access to the underlying event, to allow annotations creation only when CMD+drag is performed? Currently, there is no access to the mouse event that initiated the selection.
2. Zoom - minimal drag distance? Currently, zooming is happening on minimal mouse move
3. Streaming performance - since I'm reinitialising the uPlot on data frames change, the streaming performance may become poor with large dashboards. I haven't tested it yet, but we probably would like to use `setData`, `setSeries` etc methods for uPlot instead. This becomes problematic when number of series changes but that's only in edit mode I guess. Needs more thought and investigation
4. x-axis time formatting, axis width
5. Legend plugin somehow?
6. And many many more things :)

TODO:
- [ ] Data retrieval in plugins (access to data frame and correct field within it)
- [ ] Annotations - editor + display
- [ ] Show/hide grid
- [ ] Axes customization - is there a way to dynamically calculate axes width?
- [ ] Time zone support on x-axis
- [ ] Bar chart


packages/grafana-data/src/dataframe/MutableDataFrame.ts
```diff
@@ -115,6 +115,8 @@ export class MutableDataFrame<T = any> extends FunctionalVector<T> implements Da
       config: f.config || {},
       values: this.creator(buffer),
       labels: f.labels,
+      display: f.display,
+      getLinks: f.getLinks,
```
dprokop, line 119: This is very ad-hoc, to be changed. After the transform when aligning data frames the display processor and get links need to be attached back to the fields

packages/grafana-ui/src/components/uPlot/MicroPlot.tsx
```diff
@@ -0,0 +1,376 @@
+import React, { PureComponent } from 'react';
```
dprokop, line 1: Ignore this file, look at UPlotChart.tsx instead. This is POC that @ryantxu did some time ago. Not sure why it shows here even though this PR is targeting the source branch

packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  top = 0,
+  right = 1,
+  bottom = 2,
+  left = 3,
+}
+
+export interface GraphCustomFieldConfig {
+  showLines: boolean;
+  lineWidth: number;
+  limeMode: 'connect' | 'staircase';
```
leeoniya, line 14: limeMode sounds tasty

leeoniya:
> 1. Annotations interaction - how to use setSelect uPlot plugin hook and get access to the underlying event, to allow annotations creation only when CMD+drag is performed? Currently, there is no access to the mouse event that initiated the selection.

i should probably add `self.cursor.evt` for read-back since it'll be useful for `setCursor`, `setSelect`, and possibly other hooks in the future.

> 2. Zoom - minimal drag distance? Currently, zooming is happening on minimal mouse move

`cursor.drag.dist`: https://github.com/leeoniya/uPlot/blob/9288f09582610e22b4538f13f3ec968e07456117/dist/uPlot.d.ts#L256-L257

though it looks like you're not using the native zoom?

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -0,0 +1,119 @@
...
+    config.plugins = [
+      ...config.plugins,
+      ...Object.entries(plugins).map(p => ({
+        hooks: p[1].hooks,
+      })),
+    ];
+
+    setConfig(config);
+    setPlotData(data);
+  }, [props.data, plugins]);
```
hugohaggmark, line 69: I think I found some more deps
```suggestion
  }, [plugins, setConfig, setPlotData, props]);
```

hugohaggmark:
Great work on this as usual @dprokop ⭐ I really like the declaritive api as a potential consumer.

leeoniya:
> 1. Annotations interaction - how to use setSelect uPlot plugin hook and get access to the underlying event, to allow annotations creation only when CMD+drag is performed? Currently, there is no access to the mouse event that initiated the selection.

i gave this some more thought over the weekend and came up with a mechanism that's more flexible than simply caching and exposing the last mouse event via `cursor.evt`. the idea is to expose event proxies which allow custom filtering of raw events. i pulled out the previous code which filtered `e.button == 0` into the default proxies, which can now be overridden to run additional logic, cache events and set temporary state as needed. one immediate benefit of this is that it's not possible to catch the start of an annotation drag (mousedown) and e.g. modify the selection color if `Ctrl + drag` is used.

take a look at the `cursor.bind` branch: https://github.com/leeoniya/uPlot/tree/cursor.bind, specifically the Ctrl+drag annotation demo: https://github.com/leeoniya/uPlot/blob/cursor.bind/demos/cursor-bind.html

the code is a bit ad-hoc, but i don't currently see a way to expose a "beautiful" API that's equally capable without adding a lot of fragility and extra complexity. this is due to how much delicate state has to be tracked internally, especially for things like cursor snapping, selection and focus sync, etc.

i'm not sure this can work neatly with a layered plugin architecture since there is only a single cursor state and it needs to account for all potentially-loaded additional plugins' behaviors.

dprokop:
@leeoniya thanks a ton for chiming in! I'm back on this later this today, will go through all your feedback.

dprokop:
@leeoniya thanks again for the efforts to expose those event handlers. I looked at it and it indeed is a simple and elegant solution. But I have another alternative, as I strongly believe the events that were the source of hook execution should be exposed via the hook handlers. Thanks to this approach each plugin could independently implement custom interactions and maintain it's state independently from uPlot state. I am also not convinced to keep the event object as `cursor.evt` property as it feels like the event belongs to the particular interaction handler rather than state of the visualization.

What I consider an alternative to your proposals is to expose the event via the hook handler, and by that, I mean changing the signatures of events being direct results of user interaction to:

```ts
setCursor?:  (self: uPlot, e:  MouseEvent) => void;
setSelect?:  (self: uPlot, e:  MouseEvent) => void;
```

Thanks to this the interaction could be altered independently by plugins.

I did a quick, minimal draft of this that you can take a look at here https://github.com/dprokop/uPlot/tree/mouse-events (https://github.com/dprokop/uPlot/blob/mouse-events/demos/hook-events.html)

torkelo (commented):
Where are x-axis ticks defined? (for example for different time intervals). 
is there anyway to use
graphTimeFormat and graphTickFormatter

 https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Graph/utils.ts#L114

dprokop:
> Where are x-axis ticks defined? (for example for different time intervals).
> is there anyway to use
> graphTimeFormat and graphTickFormatter
> 
> https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Graph/utils.ts#L114

still TBD

torkelo (commented):
a lot to dive into here, but WOW! a lot of progress, I like the plugin/composable approach

leeoniya:
@dprokop i've opened https://github.com/leeoniya/uPlot/issues/298 for discussion, with my reasoning behind preferring `cursor.bind` vs your proposed API (which i've considered and decided against previously when working on the `setCursor` api and hook).

while passing the events down is easy, it is also very limited, and you'll want `cursor.bind` to do anything even slightly more advanced, at which point you have to ask whether it's worth having more than one way to get the job done (uPlot's design is almost always "no").

i'm still open to caching and exposing a `cursor.evt` or even a namespaced `cursor.evt: {mousemove, mousedown,...}`. i want this rather than events passed into hooks because it makes this accessible to every function and API, and experience has shown that it's extremely common to want this information from numerous parts of the system, which is why a lot of state is expected to be read-back from the uPlot instance rather than passing all micro-parts of state everywhere as 20-long, ever-expanding function args.

packages/grafana-ui/src/components/uPlot/context.ts
```diff
@@ -0,0 +1,142 @@
...
+  getField: (idx: number) => Field;
+  /** Returns x-axis fields */
+  getXAxisFields: () => Field[];
+  /** Returns x-axis fields */
+  getYAxisFields: () => Field[];
+  /** Returns field value by field and value index */
+  getFieldValue: (fieldIdx: number, rowIdx: number) => any;
+  /** Returns field config by field index */
+  getFieldConfig: (fieldIdx: number) => FieldConfig;
+}
```
ryantxu, line 60: Why is this all necessary?  If you are passing in DataFrame, i would expect:

```
interface PlotDataAPI {
  /** Data frame passed to graph, x-axis aligned */
  data: DataFrame;
  getXAxisFields: () => number[];
  getYAxisFields: () => number[];
```

but even that makes me nervous... very soon you are going to want other types of dimenstions as well!
```
{
 x:[0],
 y:[2,3],
 size:[4],
}
```
Actually thinking about it... this is where the single frame will fall over :cry:

dprokop: @ryantxu this API is to expose original data frame rather than particular values. It's not used for the uPlot to build the model, but for plugins to get access to the fields within input data frame. See [TooltipPlugin](https://github.com/grafana/grafana/pull/27396/files#diff-5d849ca51f1c1f3d960849f4d14680ff) for an example. I think exposing the entire fields makes it more flexible to use the data in the plugins.

I'm OK with removing the X/Y specific functions for now and rely more on the `data` directly in the plugins.

dprokop:
Closing in favor of https://github.com/grafana/grafana/pull/27572

PR #27397, opened 2020-09-04
Dependencies: Adds @testing-library/user-event for best practices

kaydelaney:
Adds `@testing-library/user-event` so that [best practices](https://testing-library.com/docs/dom-testing-library/api-events) can be followed when writing tests.


Clarity-89 (approved):
Nice! Didn't have a use case for `userEvent` myself yet but I think it's a good idea to have it 👍

PR #27399, opened 2020-09-04
Slider: add step functionality

rdubrock:
The slider component is wrapping [this component](https://github.com/react-component/slider), which has the ability to lock values to a step. This PR exposes that prop in our wrapped component.


Clarity-89 (approved):
Looks good!

PR #27404, opened 2020-09-06
Dashboard: Support configuring default timezone via config file

### LINKED ISSUE #25654: Allow configuring default timezone via config file

It is possible to configure the default timezone for dashboards via the UI `Configuration > Preferences`, however, it would be nice to have this be configurable via files as well.

**Why is this needed**:

We run all our Grafana instances fully read only and 100% provisioned from files via the datasource and dashboard provisioning features. A lot of dashboards in the community are written using the "default" timezone, which due to lack of configurability in files ends up being "browser", but we'd like all dashboards to default to "UTC".

woutersmeenk:

Add a default timezone that the administrator can set in the settings. This setting is be used as default for the users timezone preference.

Can be used when creating Grafana instances without administrator intervention, in order to give user the correct default timezone.

Fixes #25654

### DIFF
--- a/conf/defaults.ini
+++ b/conf/defaults.ini
@@ -823,5 +823,5 @@ interval_year = YYYY
 # Experimental feature
 use_browser_locale = false
 
-# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database, e.g. 'UTC' or 'Europe/Amsterdam' etc.
 default_timezone = browser
--- a/conf/defaults.ini
+++ b/conf/defaults.ini
@@ -823,4 +823,5 @@ interval_year = YYYY
 # Experimental feature
 use_browser_locale = false
 
-
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
+default_timezone = browser
--- a/conf/sample.ini
+++ b/conf/sample.ini
@@ -811,3 +811,6 @@
 
 # Experimental feature
 ;use_browser_locale = false
+
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
+;default_timezone = browser
--- a/conf/sample.ini
+++ b/conf/sample.ini
@@ -812,5 +812,5 @@
 # Experimental feature
 ;use_browser_locale = false
 
-# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database, e.g. 'UTC' or 'Europe/Amsterdam' etc.
 ;default_timezone = browser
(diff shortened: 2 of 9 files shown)


conf/defaults.ini
```diff
@@ -264,6 +264,9 @@ password_hint = password
 # Default UI theme ("dark" or "light")
 default_theme = dark
 
+# Default timezone
```
woutersmeenk, line 267: Should we add a reference to what timezones are allowed? Should the timezone be checked and a warning logged? It defaults to browser if an invalid timezone is given, currently

conf/defaults.ini
```diff
@@ -264,6 +264,9 @@ password_hint = password
...
+default_timezone = browser
+
```
torkelo, line 269: can you move this to the new [date_formats]  section added in https://github.com/grafana/grafana/pull/27216

torkelo: yes, need to add what timezone format is supported and have validation

torkelo (changes requested):
Needs docs as well https://github.com/grafana/grafana/blob/master/docs/sources/administration/configuration.md

torkelo (approved):
looks good, @marefr can you take a quick glance?

conf/defaults.ini
woutersmeenk: Move the setting to date_formats settings and added validation. I am wondering if this is the right section, because timezone is not really a date format. I don't know of a better section thou.

conf/defaults.ini
```diff
@@ -823,4 +823,5 @@ interval_year = YYYY
 # Experimental feature
 use_browser_locale = false
 
-
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
```
marefr, line 826: Would be good with a couple examples here to understand what format is expected.

```suggestion
# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database, e.g. UTC or Europe/Amsterdam etc.
```

conf/sample.ini
```diff
@@ -811,3 +811,6 @@
 
 # Experimental feature
 ;use_browser_locale = false
+
+# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database
```
marefr, line 815: Would be good with a couple examples here to understand what format is expected.

```suggestion
# Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database, e.g. UTC or Europe/Amsterdam etc.
```

docs/sources/administration/configuration.md
```diff
@@ -1426,3 +1426,6 @@ interval_year = YYYY
 Set this to `true` to have date formats be automatically be derived from browser locale. Defaults to `false`. This
 is an experimental feature right now with a few problems that remain unsolved.
 
+### default_timezone
+
+Used as the default timezone for user preferences. Can be either `browser` for the browser local timezone or a timezone name from IANA Time Zone database
```
marefr, line 1431: Would be good with a couple examples here to understand what format is expected.

```suggestion
Used as the default timezone for user preferences. Can be either `browser` for the browser local timezone or a timezone name from IANA Time Zone database, e.g. UTC or Europe/Amsterdam etc.
```

marefr (changes requested):
Looks good, but added a couple of documentation suggestions.

conf/defaults.ini
achatterjee-grafana, line 826: Needs a "." at the end of the sentence.

marefr (approved):
LGTM

marefr:

PR #27408, opened 2020-09-06
PostgresSQL: editor support for materialized views

### LINKED ISSUE #27017: Postgres - Materialized Views are not handled like tables
**What happened**:
Postgres materialized views don't show up as selectable tables and neither does their columns when setting up a panel. 
Inputting the names of the tables and columns manually does work  

I expected that when setting up  a new panel, when in the query definition tab that the "From" field should be populated with "new_mview" and the the columns "timestamp" and "valueColoumn" should be populated as possible columns to select.

This is my sample definition for a materialized view that I expected Grafana to find:
```
CREATE MATERIALIZED VIEW public.new_mview
TABLESPACE pg_default
AS 
SELECT now() AS timestamp, 100 as valueColoumn
WITH DATA;
```

I looked at the discovering query that Grafana sends to the DB-server:

```
SELECT quote_ident(table_name) FROM information_schema.tables WHERE
table_schema IN (
  SELECT
    CASE WHEN trim(s[i]) = '"$user"' THEN user ELSE trim(s[i]) END
  FROM
    generate_series(
      array_lower(string_to_array(current_setting('search_path'),','),1),
      array_upper(string_to_array(current_setting('search_path'),','),1)
    ) as i,
    string_to_array(current_setting('search_path'),',') s
) ORDER BY table_name
```
This query seems to be built by: [meta_query.ts](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/postgres/meta_query.ts )

This returns normal tables and "normal views" but not materialized views. To query postgres for the materialized views you can use the query:
`select * from pg_matviews;`

I raised a forum post about the problem here: https://community.grafana.com/t/postgres-materialized-view-autocomplete/35081

**Environment**:
- Grafana version: v7.1.3
- Data source type & version: Postgres 10.12
- OS Grafana is installed on: Ubuntu 20 LTS
- User OS & Browser: OSX, Firefox
- Grafana plugins: Default only
- Others:

jdbranham:
MView information is stored in a different system table, so the current
implemention doesn't offer suggestions.
A union query is performed to get all tables plus mviews.
This PR adds table 'type' property to the ctrl target to identify MVIEW from TABLE.
This allows us to use the correct system tables to get column information.

Fixes: grafana/grafana#27017

### DIFF
--- a/public/app/plugins/datasource/postgres/meta_query.ts
+++ b/public/app/plugins/datasource/postgres/meta_query.ts
@@ -1,7 +1,10 @@
 import QueryModel from './postgres_query';
 
 export class PostgresMetaQuery {
-  constructor(private target: { table: string; timeColumn: string }, private queryModel: QueryModel) {}
+  constructor(
+    private target: { table: string; timeColumn: string; tableType?: 'MVIEW' | 'TABLE' },
+    private queryModel: QueryModel
+  ) {}
 
   getOperators(datatype: string) {
     switch (datatype) {
@@ -50,7 +53,7 @@ SELECT
   ) AS value_column
 FROM information_schema.tables t
 WHERE `;
-    query += this.buildSchemaConstraint();
+    query += this.buildTableSchemaConstraint();
     query += ` AND
   EXISTS
   ( SELECT 1
@@ -73,19 +76,12 @@ LIMIT 1
     return query;
   }
 
-  buildSchemaConstraint() {
-    const query = `
-table_schema IN (
-  SELECT
-    CASE WHEN trim(s[i]) = '"$user"' THEN user ELSE trim(s[i]) END
-  FROM
-    generate_series(
-      array_lower(string_to_array(current_setting('search_path'),','),1),
-      array_upper(string_to_array(current_setting('search_path'),','),1)
-    ) as i,
-    string_to_array(current_setting('search_path'),',') s
-)`;
-    return query;
+  buildTableSchemaConstraint() {
+    return `table_schema IN (${this.buildSearchPathQuery()})`;
+  }
+
+  buildMViewSchemaConstraint() {
+    return `schemaname IN (${this.buildSearchPathQuery()})`;
   }
 
   buildTableConstraint(table: string) {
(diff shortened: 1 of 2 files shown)


marefr:
@jdbranham thank you for your contribution. I'm very sorry for the slow response, but we been very busy lately and have been focusing on our most prioritized issues and pull requests. 

After consideration, we have discussed this proposal and have come to the conclusion that the referenced feature request don't have enough proof that it's useful for a lot of users for us to consider accepting it.In addition, the Postgres graphical query editor was built with simplicity in mind to help/guide users without any SQL knowledge create simple SQL time series queries. For the more advanced use cases we refer to using the raw query editor. We may change our mind in the future given more and more users expressing a need for a more advanced graphical query editor, for example in the referenced feature request.  

We try to only keep PRs open that we consider accepting why I'm going to close this now. Please note that closing a PR it doesn't mean that we cannot reopen it should we get proof from many users that this feature is useful. 

Thank you again for showing interest in contributing to Grafana.

johannes-777:
Hi guys,
I'd really love to see this feature for the sake of completeness.  Materialized Views are just right in between tables and views.  And yes, I know their problems, but sometimes we just have to use them.
Also I don't see why @torkelo tagged this issue with "feature request" and a someone from grafana decided to close this issue several months later, even though a pull request had been opened.
Thanks @jdbranham for your great work.

jdbranham:
Thanks @johannes-777 -
It would be nice to see this merged. 

I understand the earlier comment by @marefr 
> the Postgres graphical query editor was built with simplicity in mind to help/guide users without any SQL knowledge create simple SQL time series queries. For the more advanced use cases we refer to using the raw query editor. 

But I'd argue that this feature makes it more simple for the users with less SQL knowledge, as those users may not understand or care about the difference between a table and a materialized view.  
It also make it easier for me, as I don't need to use the advanced editor or remember column names between different data sources. 😄

marefr:
@grafana/grafana-bi-squad any ideas?

mdvictor:
@marefr We are working on improving the SQL query editor and are aware of the lack of views support. (see https://github.com/grafana/grafana/issues/53053)

PR #27411, opened 2020-09-06
Explore: table result should not override display property

zoltanbedi:

There could be cases when one would like to use the display field in the field config in explore for table view.


aocenas:
Is this connected to the issue of how the duration is displayed in the x-ray plugin?

zoltanbedi:
> Is this connected to the issue of how the duration is displayed in the x-ray plugin?

It is.

aocenas (approved):
Nice, looks good :+1:

PR #27413, opened 2020-09-06
Chore: minor storybook update for markdown-to-jsx security fix

ryantxu:
Alternative to #27377

dependabot wants to update markdown-to-jsx... it is only used in storybook.

Rather than bump the yarn.lock directly, it is maybe better to add the other minor updates to storybook


dprokop (approved):
Yes, think this is better

PR #27419, opened 2020-09-07
API: return resource ID for auth key creation, folder permissions update and user invite complete endpoints

AgnesToulet:
Enrich following endpoints to return the ID of the created / updated resource:
- `POST /api/auth/keys`
- `POST /api/folders/:uid/permissions`
- `POST /api/user/invite/complete`

It makes it easier to have access to the resource ID for auditing without any breaking change or important information leakage.


aknuds1 (approved):
Looks good AFAICT.

PR #27420, opened 2020-09-07
Settings: Remove unnessary error handling code

torkelo:
Convert `valueAsString` in the `setting` package to not handle panics from value conversion, as there shouldn't be any actual chance of this happening. I read the ini library code, and when you request a string it doesn't perform any conversion AFAICT.


pkg/setting/setting.go
```diff
@@ -838,6 +797,14 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 	return nil
 }
 
+func readSectionValueString(section *ini.Section, keyName string, defaultValue string) string {
+	if !section.HasKey(keyName) {
+		return defaultValue
+	}
```
aknuds1, line 803: Doesn't `section.Key(keyName).MustString(defaultValue)` have the same effect as this?

marefr (approved):
LGTM, but will let Arve review as well since he did a big refactoring of this file recently.

aknuds1 (commented):
I guess `valueAsString` will panic if the key doesn't exist, due to a nil pointer error, and convert the panic to an error? That does look weird indeed. I would expect `valueAsString` to instead convert the missing key to its default argument.

Why don't we instead of making a new function, fix `valueAsString`?

torkelo:
> Why don't we instead of making a new function, fix valueAsString?

yea, that sounds better!

aknuds1:
@torkelo Do you want to make the changes, or should I push them to this branch?

torkelo:
@aknuds1 if you can take over this PR that be great  this was just a quick POC to get the conversation started

aknuds1:
@torkelo Alright, will do!

aknuds1 (approved):

marefr (approved):
LGTM

PR #27423, opened 2020-09-07
Docs: added version note to the group-by transformation

### LINKED ISSUE #27386: Docs: Add version notice to Group by docs
**What you expected to happen**: I upgraded from v7.0.5 to v7.1.5, hoping to see the new ["Group By" transformation](https://grafana.com/docs/grafana/latest/panels/transformations/#group-by) advertised in the docs.

**What happened**: Sadly the transformation isn't listed in the Transformations tab. Has it been removed? Merged into another transformation? Was it ever there? :thinking: 

**Environment**:
- Grafana version: 7.1.5
- Data source type & version: InfluxDB, version 1.8.2
- OS Grafana is installed on: Linux 4.19
- User OS & Browser: Linux 4.19, Firefox
- Grafana plugins: N/A
- Others: N/A

mckn:
When approving the group-by transformation I missed to check that we had a version note (since that feature will be available starting from 7.2). So this PR adds that missing notice.

Fixes #27386

### DIFF
--- a/docs/sources/panels/transformations.md
+++ b/docs/sources/panels/transformations.md
@@ -98,7 +98,7 @@ After I apply the transformation, there is no time value and each column has bee
 
 ### Merge
 
-> **Note:** This documentation refers to a Grafana 7.1 feature.
+> **Note:** This transformation is only available in Grafana 7.1+.
 
 Use this transformation to combine the result from multiple queries into one single result. This is helpful when using the table panel visualization. Values that can be merged are combined into the same row. Values are mergeable if the shared fields contains the same data.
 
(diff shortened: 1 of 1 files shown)


docs/sources/panels/transformations.md
```diff
@@ -226,6 +226,8 @@ After I apply the transformation, my labels appear in the table as fields.
 
 ### Group By
 
+> **Note:** This documentation refers to a Grafana 7.2 feature.
```
torkelo, line 229: This transform is only available in Grafana 7.2+

mckn: Ah, I just used the same wording as other notes in this md file. But maybe update it on the other places as well?

dprokop (approved):

PR #27427, opened 2020-09-07
Instrumentation: Adds instrumentation for outgoing datasource requests

bergquist:
Adds instrumentation for datasource requests. This will include both data proxy requests and requests sent from alerting. 

response size will be especially helpful when tracking down why some hosted customers are OOMing.

### DIFF
--- a/pkg/models/datasource_cache.go
+++ b/pkg/models/datasource_cache.go
@@ -11,7 +11,52 @@ import (
 	"time"
 
 	"github.com/grafana/grafana/pkg/setting"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
+
+var datasourceRequestCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Namespace: "grafana",
+		Name:      "datasource_requests_total",
+		Help:      "A counter for requests from the wrapped client.",
+	},
+	[]string{"datasource", "code", "method"},
+)
+
+var datasourceRequestSummary = prometheus.NewSummaryVec(
+	prometheus.SummaryOpts{
+		Namespace:  "grafana",
+		Name:       "datasource_request_duration_seconds",
+		Help:       "summary of datasource requests sent from Grafana",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+	}, []string{"datasource", "code", "method"},
+)
+
+var datasourceResponseSummary = prometheus.NewSummaryVec(
+	prometheus.SummaryOpts{
+		Namespace:  "grafana",
+		Name:       "datasource_response_size_bytes",
+		Help:       "summary of datasource response sizes returned to Grafana",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+	}, []string{"datasource"},
+)
+
+var datasourceRequestsInFlight = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: "grafana",
+		Name:      "datasource_request_in_flight",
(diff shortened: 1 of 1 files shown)


pkg/models/datasource_cache.go
```diff
@@ -11,17 +11,87 @@ import (
 	"time"
 
 	"github.com/grafana/grafana/pkg/setting"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var datasourceRequestCounter = prometheus.NewCounterVec(
```
wbrowne, line 18: Just curious - are you hoping to measure success/failure rates with this counter (I'm assuming `code` refers to the response code)?

pkg/models/datasource_cache.go
```diff
@@ -86,8 +156,9 @@ func (ds *DataSource) GetHttpTransport() (*dataSourceTransport, error) {
 	}
 
 	dsTransport := &dataSourceTransport{
-		headers:   customHeaders,
-		transport: transport,
+		headers:        customHeaders,
+		transport:      transport,
```
marefr, line 160: Nit. Would construct the roundtrippers here and use that as transport, e.g. 

```go 
type RoundTripperFunc func(req *http.Request) (*http.Response, error)

func customHeadersRoundtrip(headers map[string]string, next http.RoundTripper) RoundTripperFunc { ... } 

roundTripper := customHeadersRoundtrip(customHeaders, instrumentRoundtrip(d.datasourceName, transport))

dsTransport := &dataSourceTransport{
  transport: roundTripper,
}
```

This way easier to add additional roundtrippers (auth middleware in the pipeline etc :) )

pkg/models/datasource_cache.go
```diff
@@ -11,17 +11,87 @@ import (
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+var datasourceRequestCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Namespace: "grafana",
+		Name:      "datasource_request_total",
+		Help:      "A counter for outgoing requests for a datasource",
+	},
+	[]string{"datasource", "code", "method"},
```
marefr, line 24: Datasource name as label may be subject for high cardinality. Possibility of same name in multiple organizations. Would using data source type make more sense and/or in addition to datasource name?

marefr (approved):
Nice! LGTM, but added a couple comments

pkg/models/datasource_cache.go
bergquist: I picked this solution to make it as simple as possible since I know you are gonna be working on it. Is it fine if you address the additional roundtrippers in your PR?

pkg/models/datasource_cache.go
bergquist: I think it would be very very very rare that the database count causes high cardinality problems but it could since its unbound in theory. How about we introduce a setting for that if it becomes an issue?

OrgId is a good point... Hm. how about a new orgId label? Cardinality wise is the same since it will always be the same per datasource.

pkg/models/datasource_cache.go
bergquist: Correct! :dart:

pkg/models/datasource_cache.go
marefr: Sure!

pkg/models/datasource_cache.go
marefr: > How about we introduce a setting for that if it becomes an issue?

Sounds good. 

> OrgId is a good point... Hm. how about a new orgId label? Cardinality wise is the same since it will always be the same per datasource.

I don't know. Maybe wait with this one as well until it becomes a problem given "How about we introduce a setting for that if it becomes an issue"

bergquist: :+1: I'll revert my commit and merge this.

PR #27430, opened 2020-09-07
Docs: Working with data frames guide

marcusolsson:
Adds a guide on how to work with data frames.

### DIFF
--- a/docs/sources/developers/plugins/working-with-data-frames.md
+++ b/docs/sources/developers/plugins/working-with-data-frames.md
@@ -1,0 +1,123 @@
++++
+title = "Working with data frames"
+type = "docs"
++++
+
+# Working with data frames
+
+The data frame is a columnar data structure which allows efficient querying of large amounts of data. Since data frames are a central concept when developing plugins for Grafana, in this guide we'll look at some ways you can use them.
+
+The [DataFrame]({{< relref "../../packages_api/data/dataframe.md" >}}>) interface contains a `name` and an array of `fields` where each field contains the name, type, and the values for the field.
+
+> **Note:** If you're looking to migrate an existing plugin to use the data frame format, refer to [Migrate to data frames]({{< relref "migration-guide.md#migrate-to-data-frames" >}}).
+
+## Create a data frame
+
+If you build a data source plugin, then you'll most likely want to convert a response from an external API to a data frame. Let's look at how to create a data frame.
+
+Let's start with creating a simple data frame that represents a time series. The easiest way to create a data frame is to use the [toDataFrame]({{< relref "../../packages_api/data/todataframe.md" >}}) function.
+
+```ts
+// Need to be of the same length.
+const timeValues = [1599471973065, 1599471975729];
+const numberValues = [12.3, 28.6];
+
+// Create data frame from values.
+const frame = toDataFrame({
+    name: "http_requests_total",
+    fields: [
+        { name: "Time", type: FieldType.time, values: timeValues },
+        { name: "Value", type: FieldType.number, values: numberValues }
+    ]
+});
+```
+
+> **Note:** Data frames representing time series contain at least a `time` field, and a `number` field. By convention, built-in plugins use `Time` and `Value` as field names for data frames containing time series data.
+
+As you can see from the example, creating data frames like this requires that your data is already stored as columnar data. If you already have the records in the form of an array of objects, then you can pass it to `toDataFrame` which tries to guess the schema based on the types and names of the objects in the array. If you're creating complex data frames this way, then be sure to verify that you get the schema you expect.
+
+```ts
+const series = [
+    { Time: 1599471973065, Value: 12.3 },
+    { Time: 1599471975729, Value: 28.6 }
+];
+
+const frame = toDataFrame(series);
+frame.name = 'http_requests_total'
+```
+
+## Read values from a data frame
+
+When you're building a panel plugin, the data frames returned by the data source are available from the `data` prop in your panel component.
+
+```ts
+const SimplePanel: React.FC<Props> = ({ data }) => {
+  const frame = data.series[0];
+
+  // ...
+}
+```
+
+Before you start reading the data, think about what data you expect. For example, to visualize a time series we'd need at least one time field, and one number field.
+
+```ts
+const timeField = frame.fields.find(field => field.type === FieldType.time);
+const valueField = frame.fields.find(field => field.type === FieldType.number);
+```
+
+Other types of visualizations might need multiple dimensions. For example, a bubble chart that uses three numeric fields: the X-axis, Y-axis, and one for the radius of each bubble. In this case, instead of hard coding the field names, we recommend that you let the user choose the field to use for each dimension.
+
+```ts
+const x = frame.fields.find(field => field.name === xField);
+const y = frame.fields.find(field => field.name === yField);
+const size = frame.fields.find(field => field.name === sizeField);
+
+for (let i = 0; i < frame.length; i++) {
+  const row = [x?.values.get(i), y?.values.get(i), size?.values.get(i)];
+
+  // ...
+}
+```
+
+Alternatively, you can use the [DataFrameView]({{< relref "../../packages_api/data/dataframeview.md" >}}), which gives you an array of objects that contain a property for each field in the frame.
+
+```ts
+const view = new DataFrameView(frame);
+
+view.forEach(row => {
+  console.log(row[options.xField], row[options.yField], row[options.sizeField]);
+})
+```
+
+## Display values from a data frame
+
+[Field options]({{< relref "../../panels/field-options.md" >}}) let the user control how Grafana displays the data in a data frame.
+
+To apply the field options to a value, use the `display` method on the corresponding field. The result contains information such as the color and suffix to use when display the value.
+
+```ts
+const valueField = frame.fields.find(field => field.type === FieldType.number);
+
+return (
+  <div>
+    {valueField
+      ? valueField.values.toArray().map(value => {
+          const displayValue = valueField.display!(value);
+          return (
+            <p style={{ color: displayValue.color }}>
+              {displayValue.text} {displayValue.suffix ? displayValue.suffix : ''}
+            </p>
+          );
+        })
+      : null}
+  </div>
+);
+```
+
+To apply field options to the name of a field, use [getFieldDisplayName]({{< relref "../../packages_api/data/getfielddisplayname.md" >}}).
+
+```ts
+const valueField = frame.fields.find(field => field.type === FieldType.number);
+const valueFieldName = getFieldDisplayName(valueField, frame);
+```
+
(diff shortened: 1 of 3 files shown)


docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,143 @@
...
+
+// Create data frame from values.
+const frame: DataFrame = {
+    name: "http_requests_total",
+    fields: [
+        { name: "Time", type: FieldType.time, values: timeValues },
+        { name: "Value", type: FieldType.number, values: numberValues }
+    ]
+    length: timeValues.length,
+}
```
hugohaggmark, line 29: There's also the option to use `toDataFrame`here: https://github.com/grafana/grafana/blob/10141c5e6caeacaf817cd06ed1b1b8964791c410/packages/grafana-data/src/dataframe/processDataFrame.ts#L283

Not sure where this fits in though.

marcusolsson: How would you use toDataFrame here? How's it different from using ArrayDataFrame?

hugohaggmark: I'm not an expert on this, but one difference is that it tries to discover the field types so you don't have to specify them yourself. Another difference is that it can handle legacy time series format conversions. And in your example there would be no need for the :DataFrame type. 

This is exactly what I find distracting/hard to reason about, which method to use when. So this doc is super important to align our api/thoughts :)
```suggestion
const frame = toDataFrame({
    name: "http_requests_total",
    fields: [
        { name: "Time", type: FieldType.time, values: timeValues },
        { name: "Value", type: FieldType.number, values: numberValues }
    ]
    length: timeValues.length,
})
```

marcusolsson: I believe that's what ArrayDataFrame does as well? In your suggestion, I don't see why the call to `toDataFrame` is needed, since I've already specified the fields and values?

Even though it's not needed, I added the `: DataFrame` annotation more as documentation to make it clear that it's fulfilling an interface. Too explicit?

marcusolsson: Oh I see now, `toDataFrame` returns either a MutableDataFrame or a ArrayDataFrame. In that case, I think we should prefer to use the actual classes rather than `toDataFrame` unless we're converting from legacy formats.

ryantxu: toDataFrame does its best to convert the input into a DataFrame -- if checks all variations of what the input may be and returns it as a DataFrame

https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/dataframe/processDataFrame.ts#L283

MutableDataFrame will convert the values DTO (a simple array) to a Vector -- with `get()` functions etc

ryantxu: > How's it different from using ArrayDataFrame?

ArrayDataFrame will take any array of objects and magic it into a DataFrame.

For ArrayDataFrame, you would start with:
```
const vals = [
{ time:123, value:1234, name:'abc' },
{ time:234, value:2345, name:'bcd' },
];
const frame = new ArrayDataFrame(vals);
```
That will look like a columnar data frame

marcusolsson: Yes, having looked at the code, `toDataFrame` looks like a smart factory that returns an implementation based on the input.

I think we should encourage people to understand and use MutableDataFrame and ArrayDataFrame directly rather than relying on the magic of `toDataFrame`. Using `toDataFrame` for legacy formats makes sense as you typically don't care. You just want to convert it.

marcusolsson:
@torkelo indicated that we might not want to promote the use of MutableDataFrame since the API is likely to change/break in coming releases.

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+
+```ts
+const SimplePanel: React.FC<Props> = ({ data }) => {
+  const frame = data.series[0];
+
+  // ...
+}
+```
+
+Before you start reading the data, think about what data you expect. For time series, we can expect a `Time` and a `Value` field, so it's safe to hard code them.
```
torkelo, line 59: this is not safe, visualizations should not expect any specific field naming

torkelo: Better to use the utility function getTimeField to find the first time field. and for value field use the first (or all) numeric fields.

torkelo (commented):
Looks good, added comment about hardcoding field names. 

Would be good to also mention that if a visualization wants to display a field name or time series name that they should use the utility function getFieldDisplayName, example: 

https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph/data_processor.ts#L46 

This function will generate a good field name taking into account all the other fields and labels and also take into account field configured displayName manually set by the user

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
++++
+title = "Working with data frames"
+type = "docs"
++++
+
+# Working with data frames
+
+The data frame is a columnar data structure, which allows efficient querying of large amounts of data. Data sources create data frames for panels to visualize. Since data frames are a central concept when developing plugins for Grafana, in this guide we'll look at some ways you can use them.
+
+The [DataFrame]({{< relref "../../packages_api/data/dataframe.md" >}}>) interface contains a `name` and an array of `fields` where each field contains the name, type, and the values for the field.
```
oddlittlebird, line 10: The title of the linked document is "Data frames." I'm not sure what "DataFrame" is referring to here.

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Create a data frame
+
+If you building a data source plugin, you'll most likely want to convert a response from an external API to a data frame. Let's look at how to create a data frame.
```
oddlittlebird, line 14:
```suggestion
If you build a data source plugin, then you'll most likely want to convert a response from an external API to a data frame. Let's look at how to create a data frame.
```

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+    fields: [
+        { name: "Time", type: FieldType.time, values: timeValues },
+        { name: "Value", type: FieldType.number, values: numberValues }
+    ]
+});
+```
+
+> **Note:** Data frames representing time series contain at least a `time` field, and a `number` field. By convention, built-in plugins use `Time` and `Value` as field names for data frames containing time series data.
+
+As you can see from the example, creating data frames like this requires that your data is already stored as columnar data. If you already have the records in the form of an array of objects, you can pass it to `toDataFrame` which tries to guess the schema based on the types and names of the objects in the array. If you're creating complex data frames this way, make sure to verify that you get the schema you expect.
```
oddlittlebird, line 35: I suggest putting this requirement about the data being stored as columnar data earlier. This is kind of important.

oddlittlebird, line 35:
```suggestion
As you can see from the example, creating data frames like this requires that your data is already stored as columnar data. If you already have the records in the form of an array of objects, then you can pass it to `toDataFrame` which tries to guess the schema based on the types and names of the objects in the array. If you're creating complex data frames this way, then be sure to verify that you get the schema you expect.
```

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+  throw new Error('Invalid type for Time field')
+}
+
+const valueField = frame.fields.find(field => field.name === 'Value');
+if (valueField.type !== FieldType.number) {
+  throw new Error('Invalid type for Value field')
+}
+```
+
+Consider other types of visualizations may need multiple dimensions. For example, a bubble chart that uses three numeric fields: The X-axis, Y-axis, and one for the radius of each bubble. In this case, instead of hard coding the field names, we recommend that you let the user choose the field to use for each dimension.
```
oddlittlebird, line 73:
```suggestion
Other types of visualizations might need multiple dimensions. For example, a bubble chart that uses three numeric fields: the X-axis, Y-axis, and one for the radius of each bubble. In this case, instead of hard coding the field names, we recommend that you let the user choose the field to use for each dimension.
```

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+const x = frame.fields.find(field => field.name === options.xField);
+const y = frame.fields.find(field => field.name === options.yField);
+const size = frame.fields.find(field => field.name === options.sizeField);
+
+for (let i = 0; i < frame.length; i++) {
+  const row = [x.get(i), y.get(i), size.get(i)];
+}
+```
+
+Alternatively, you can use the [DataFrameView]({{< relref "../../packages_api/data/dataframeview.md" >}}) which gives you an array of objects that contain a property for each field in the frame.
```
oddlittlebird, line 85:
```suggestion
Alternatively, you can use the [DataFrameView]({{< relref "../../packages_api/data/dataframeview.md" >}}), which gives you an array of objects that contain a property for each field in the frame.
```

docs/sources/developers/plugins/working-with-data-frames.md
```diff
@@ -0,0 +1,109 @@
...
+
+```ts
+const view = new DataFrameView(frame);
+
+view.forEach(row => {
+  console.log(row[options.xField], row[options.yField], row[options.sizeField]);
+})
+```
+
+## Convert legacy formats to data frames
```
oddlittlebird, line 95: This seems like an important task. I suggest either putting it first (since it converts old data into the required format for the example to work!) or putting it in a separate task topic by itself so that it has more visibility. If you choose the latter route, then you might expand the instructions a bit.

oddlittlebird (approved):
Minor grammar edits. Great stuff!

docs/sources/developers/plugins/working-with-data-frames.md
marcusolsson: DataFrame refers to the reference documentation for that interface. The reference seems right to me, it points to https://grafana.com/docs/grafana/latest/packages_api/data/dataframe/

docs/sources/developers/plugins/working-with-data-frames.md
marcusolsson: It's helpful if you're maintaining an older plugin, but not relevant to anyone starting out today I don't think this should be at the top of this document. I could see this living in the migration guide, but I believe this document is where users will look first.

marcusolsson:
@torkelo I added a new section "Display values from a data frame" to hold `getFieldDisplayName` and `DisplayProcessor`. Could you have a look?

docs/sources/developers/plugins/working-with-data-frames.md
oddlittlebird: Silly me, I thought this linked to the Data frames doc. Disregard!

docs/sources/developers/plugins/working-with-data-frames.md
oddlittlebird: I thought lots of people were converting from pre-7.0 to now. It certainly comes up a lot in #plugins. 

So...as a person who might need to update an old plugin, "working with data frames" does not necessarily lead me to "here's how to convert my old plugin data to the new scheme."

PR #27436, opened 2020-09-07
Live: check logged-in-user for grafana live

ryantxu:
Make sure the grafana live user is authenticated, and add info to the channel


public/app/core/core.ts
```diff
@@ -46,7 +45,6 @@ export {
   registerAngularDirectives,
   arrayJoin,
   coreModule,
-  liveSrv,
```
ryantxu, line 49: this was unused and out-of-date

torkelo (approved):
Looks ok

PR #27437, opened 2020-09-08
POC/Exemplars: possible query runner

ryantxu:
When thinking about exemplars, annotations, and in general data that represents alternative semantics to the generic panel data, there are a few ways to approach it.

Two possible options are:
* augment query results with where the data should go (property in DataFrame.meta?)
* have each "type" of data make its own query request, and have visualizations subscribe to the types they know how to display

This PR explores how the 2nd option may look.

Posting early as a way to start discussion


packages/grafana-data/src/types/datasource.ts
```diff
@@ -392,6 +405,11 @@ export interface DataQuery {
    */
   queryType?: string;
 
+  /**
+   * Specify the explict channel for where the query results should be sent
+   */
+  queryChannel?: DataQueryChannel;
```
ryantxu, line 411: Not sure a good term to use here -- any ideas?

I would like to reserve "stream" and "channel" for the "live" streaming feature.

Maybe:
* DataQuerySubject?
* DataQueryTopic?
* DataQuery????

dprokop: Think `topic` is a good comporomise!

aocenas:
I think it would be great if there was a way to unify this with what we did for Explore in https://github.com/grafana/grafana/pull/25890. Basically there is `preferredVisualisationType` metadata: 
https://github.com/grafana/grafana/blob/75e14aa12035691c4584ec446b22a8612ed2b7e9/packages/grafana-data/src/types/data.ts#L34

And then we use it to figure out how to present it:
https://github.com/grafana/grafana/blob/09c95714627d666fb73015f98945f04d5c85e70a/public/app/features/explore/utils/ResultProcessor.ts#L33

Seems like this could map to it in some sense. Like `annotation` could be `preferredVisualisationType` type that could apply as overlay to any other supported visualisations or still overridden to be shown in table or something else.

In any case I would assume exemplars should work also in Explore so seems like some unified concept of mapping data -> vis is needed for for both dashboards and explore.

ryantxu:
@aocenas -- the proposal here is an explicit stream for exemplar/annotation results.  Anything returned in that stream should be treated as exemplars/annotations.  This should help avoid the problems with `preferredVisualisationType` since many results would be happy in multiple visualizations  

> In any case I would assume exemplars should work also in Explore 

yes absolutely!  The proposed design here aims to support Prometheus as easily as possible. I don't know if explore should ever bother trying to support exemplars from other sources like SQL etc.

aocenas:
> This should help avoid the problems with preferredVisualisationType since many results would be happy in multiple visualizations

To me these 2 things just seem too similar to have 2 separate metadata for it. I have a data and I wan't to show them in certain way. I want this shown as table, this as graph and this as annotations. What the app does with that information is separate thing but I don't see a problem with this resulting in one graph one table and both showing some sort of annotation overlay.

> I don't know if explore should ever bother trying to support exemplars from other sources like SQL etc.
Why not? I mean would that be different from prometheus exemplars?

ryantxu:
After discussion, this route is too complicated for most exemplars since it would start parallel requests for each topic.  That has some advantages -- especially for annotations, BUT so much of our tooling expects the single request path that it will cause more headache.

For annotations that will typically share a global query we will need to do something else.

See #27646 for a first pass at an alternate approach

PR #27438, opened 2020-09-08
DashboardSettings: Fixes auto refresh crash with space in interval

### LINKED ISSUE #27075: Crash after save dashboard
**What happened**:
Crash after save dashboard without recovery dashboard or recovery from backup.

Simple save without crash

Dashboard Settings => Auto-Refresh =>
'10s,30s,1m... etc' change to
'10s,30s,1 m.... etc' ('SPACE' between '1' and 'm')

**Environment**:
- Grafana version: 7.1.0 and 7.1.3
- Data source type & version: InfluxDB 1.8.1
- OS Grafana is installed on: Windows Server 2016
- User OS & Browser: Windows Chrome
- Grafana plugins: Not
- Others:

hugohaggmark:
There were some things not working with auto refresh intervals:

1. Default intervals were not added for new dashboards
2. Spaces within intervals crashed the whole dashboard
3. A user could add duplicate intervals
4. A user could add intervals that were below minimum interval

I added `renderCount` to `SettingsCtrl` to make sure the React component is re-rendered whenever we update Angular. I also refactored out Auto Refresh into its own component because it has some additional complexity. I added invalid state to Auto Refresh when user changes values instead of on Blur. I struggled a bit which way to go. Please give feedback if we should change to invalidate on Blur instead.

I also automatically remove any values below minimal interval on Blur, not sure what the requirement is here. Should the user be able to enter an interval below the configured minimal interval in a dashboard? The refresh picker will only display/use the intervals above the configured minimal interval, so it doesn't matter what the dashboard settings say (I think).

Fixes #27075

Please test this PR for both new dashboards and existing.


dprokop:
I found some weird behavior:

Kapture 2020-09-08 at 9 08 47

Basically, the onChange validation and Angular forced re-render caues the input to loose focus. Maybe the validation should happen only onBlur to limit the amount of those glitches?

Another thing I have noticed is when you want to remove the entire intervals string. Select entire (cmd+A) or part of the intervals string and press Backspace - nothing happens.

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const onIntervalsBlur = useCallback(
+    (event: React.FormEvent<HTMLInputElement>) => {
+      if (!event.currentTarget.value || invalidIntervalsMessage) {
+        return;
+      }
+
+      onRefreshIntervalChange(getValidIntervals(intervals));
+    },
+    [intervals, onRefreshIntervalChange]
+  );
```
dprokop, line 56: Can't see much benefit of live(on change) validation - maybe onBlur would be sufficient?

public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
```diff
@@ -23,6 +23,8 @@ export class SettingsCtrl {
   sections: any[];
   hasUnsavedFolderChange: boolean;
   selectors: typeof selectors.pages.Dashboard.Settings.General;
+  renderCount: number; // hack to update React when Angular changes
```
dprokop, line 26: Interesting!

hugohaggmark:
> I found some weird behavior:
> 
> Kapture 2020-09-08 at 9 08 47
> 
> Basically, the onChange validation and Angular forced re-render make the input to lose focus. Maybe the validation should only happen onBlur to limit the amount of those glitches?
> 
> Another thing I have noticed is when you want to remove the entire intervals string. Select entire (cmd+A) or part of the intervals string and press Backspace - nothing happens.

Thanks for that great feedback, I'll change it to onBlur instead and look into the cmd+A thingy.

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.tsx
hugohaggmark: Thank you, I agree and will change it back.

public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
hugohaggmark: Not proud of that, `this.forceUpdate()` did not work for all cases unfortunately 😞

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
```diff
@@ -0,0 +1,179 @@
+import React from 'react';
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { defaultIntervals } from '@grafana/ui/src/components/RefreshPicker/RefreshPicker';
+
+import { AutoRefreshIntervals, getValidIntervals, Props, validateIntervals } from './AutoRefreshIntervals';
+import { TimeSrv } from '../../services/TimeSrv';
```
dprokop, line 6:
```suggestion
import { TimeSrv } from 'app/core/services/TimeSrv';
```

Nit. I believe we should have linter for that ;)

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  return { rerender, props };
+};
+
+describe('AutoRefreshIntervals', () => {
+  describe('when component is mounted with refreshIntervals', () => {
+    it('then supplied intervals should be shown', () => {
+      setupTestContext({ getIntervalsFunc: () => ['5s', '10s'] }); // remove 1s entry to validate we're calling getIntervalsFunc
+
+      const input = (screen.getByRole('textbox') as unknown) as HTMLInputElement;
```
Clarity-89, line 28: Wondering why `unknown` is necessary here? I remember writing similar tests and just casting as `HTMLInputElement` was enough. 

Also if the goal is to test input value after component has been rendered, i.e. there was no manual `onChange` trigger via `fireEvent`, `toHaveAttribute` can be used, which doesn't require type casting: ` expect(screen.getByRole('textbox')).toHaveAttribute('value', '5s,10s');`

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+    });
+  });
+
+  describe('when input loses focus and intervals are valid', () => {
+    it('then onRefreshIntervalChange should be called', () => {
+      const { props } = setupTestContext({ validateIntervalsFunc: () => null });
+
+      const input = (screen.getByRole('textbox') as unknown) as HTMLInputElement;
+
+      act(() => {
```
Clarity-89, line 63: `fireEvent` is internally wrapped in `act` so using it shouldn't be necessary here.

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      setInvalidIntervalsMessage(invalidMessage);
+    },
+    [intervals, onRefreshIntervalChange, setInvalidIntervalsMessage]
+  );
+
+  return (
+    <div className="gf-form">
+      <span className="gf-form-label width-7">Auto-refresh</span>
```
Clarity-89, line 63: Just a thought but I wonder if we could use `label` here instead of `span`? It'd make the field more accessible (combined with added `htmlFor` and `id` to the input element) and would make it easier to query for this input in tests.

public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
```diff
@@ -23,6 +23,8 @@ export class SettingsCtrl {
...
+  refreshIntervals: string[];
```
dprokop, line 27: I guess this is not used, isn't it?

dprokop (approved):
Looks and works good! I've left some final nits to consider :)

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
hugohaggmark: Great feedback @Clarity-89 , I actually read up a lot on RTL last evening so I already had a long list of things to change. This is one of those. I'll be right back, again thank you!

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
hugohaggmark: 👍

public/app/features/dashboard/components/DashboardSettings/SettingsCtrl.ts
hugohaggmark: Good catch 👍 !

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.tsx
hugohaggmark: Yepp this was on my list, but unfortunately `id` or `name` props are not honored by our Input component.

public/app/features/dashboard/components/DashboardSettings/AutoRefreshIntervals.test.tsx
hugohaggmark: Will fix!

PR #27441, opened 2020-09-08
Themes: Fix issue generating themes due to code in @grafana/data accessing window object during module load

torkelo:
Having code execute during module load is bad, changed how the locale format is constructed from eager (during module load) to lazy (ie function call to get it).,


hugohaggmark (approved):
LGTM!

dprokop (approved):

PR #27446, opened 2020-09-08
Transformations: make sure that we copy all the values from a field when using MutableDataFrame.

mckn:
We have some broken tests in master and this PR fixes that by making sure that we copy all properties from a field when creating a new DataFrame by using the MutableDataFarme.


torkelo (approved):
LGTM

hugohaggmark (approved):
LGTM!

PR #27453, opened 2020-09-08
Transformations: Fixes outer join transformation when frames are missing field to join by

### LINKED ISSUE #26186: Outer join table transform fails when one or more queries in a table returns no data
**What happened**:
When creating a table that uses multiple queries (all queries set format "Table", instant toggled on) and an outer join transform, I noticed that sometimes the outer join fails and the table reverts to showing a drop-down for queries depending on the time window I have selected.  This appears to happen because one or more queries in the set of queries returns **no data** over that time range.  This only reproduces with 3+ queries.

When doing an outer join, if a query returns NO DATA, either hide the affected table column or output no data for each row in the column (and allow user specified "no value" overrides).

Legacy tables (from < Grafana 7) do not have this problem.  They automatically do outer joins, and if a column is tied to a query which has no data, the column is hidden without affecting the rest of the table.

1. Create a **table** visualization
2. Using a **Prometheus** datasource, add 3+ queries that all have a common key (you can even use the same query 3 times).
3. Set the query format to **Table** and the enable the **Instant** option.
4. Add an **Outer join** transform and select the common key.
5. Observe that you see 3 value fields
6. Modify one of the queries so that it returns no data (ex: add a >0 to the end of a query that returns 0 for the time range)
7. Observe that the table now shows just 1 value column, and a drop-down to select between the 2 queries that are returning data

**Environment**:
- Grafana version: 7.0.5
- User OS & Browser: Windows / Chrome
- Grafana plugins: Kubernetes
- Others:

hugohaggmark:
See #26186 for a way to repro.

Fixes #26186

### DIFF
--- a/packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts
+++ b/packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts
@@ -288,4 +288,62 @@ describe('SeriesToColumns Transformer', () => {
       expect(filtered.fields).toEqual(expected);
     });
   });
+
+  it('joins if fields are missing', () => {
+    const cfg: DataTransformerConfig<SeriesToColumnsOptions> = {
+      id: DataTransformerID.seriesToColumns,
+      options: {
+        byField: 'time',
+      },
+    };
+
+    const frame1 = toDataFrame({
+      name: 'A',
+      fields: [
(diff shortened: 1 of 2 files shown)


dprokop (approved):
Nice!

PR #27454, opened 2020-09-08
CloudWatch/Logs: Adds metadata to grouped CloudWatch logs dataframe

kaydelaney:
Previously data frame metadata was stripped when results were being grouped, which could cause issues. This PR fixes that.


hugohaggmark (approved):
I have no idea what the impact is for this change, but it looks good 🌟

PR #27457, opened 2020-09-08
Drone: Make parallel step for publishing front-end metrics

aknuds1:
Make a parallel step for publishing front-end metrics on Drone, which should be faster.


papagian (approved):
AFAICT LGTM

PR #27459, opened 2020-09-08
AzureMonitor: Change filterDimensions property to match what is stored

kylebrandt:
Changes the `dimensionsFilters`  property to `dimensionsFilter` in the Azure Monitor Datasource's AzureMonitor service to make what is sent to match the saved model.

Before this, the property that the backend was expecting was not available in the case of alerting, where the stored model is fetched.


ryantxu (approved):
ahhh consistency!

Elfo404 (approved):
Awesome! I just tested it and this will also fix part of https://github.com/grafana/grafana/pull/27354 (dimension filters were interpolated correctly)

PR #27463, opened 2020-09-08
InfluxDB/Flux: Fix for Alerts on InfluxDB Flux datasources only use the first series

### LINKED ISSUE #27218: Alerts on InfluxDB Flux datasources only use the first series
**What happened**:
When adding an alert to a graph backed by a Flux query that has multiple series the alert will only trigger if the first series breaks the condition 

The alert should be triggered if any series breaks the condition

Create a multi-series graph against an InfluxDB Flux datasource. I used:

```
from(bucket: "servers")
  |> range(start: -10m)
  |> filter(fn: (r) =>
      r["_measurement"] == "cpu" and
      r["cpu"] == "cpu-total" and
      r["_field"] == "usage_idle"
  )
  |> group(columns: ["host"], mode: "by")
  |> aggregateWindow(every 1m, fn: mean)
```

Add an alert to the graph - e.g `WHEN avg() OF query(a, 5m, now) IS BELOW 10`

Save the panel

Click "Test Rule". In the returned object the `Condition[0]: Query Result` object will have a single element in the `data.series` array and there will be a single "Condition[0]: Eval:" object when there should be one per series.

I have reproduced this bug with 7.1.4, 7.1.5 and the nightly Grafana builds against Influx 2 and Influx 1.8.

I have tried more complex queries that change the series names to be unique and minimise the information returned and the result is consistent. 

**Environment**:
- Grafana version:
Most testing done on 7.1.4. Reproduced on 7.1.5 and nightly.
- Data source type & version:
InfluxQL Flux (built-in)
- OS Grafana is installed on:
Redhat
- User OS & Browser:
Windows 10, Chrome
- Grafana plugins:
None other than built-in
- Others:

ryantxu:
Fixes #27218

Alerting queries do not send interval or maxDataPoints, this picks reasonable defaults, but a better solution should send it for alert queries.

With this change, I now see:
image

### DIFF
--- a/pkg/tsdb/influxdb/flux/query_models.go
+++ b/pkg/tsdb/influxdb/flux/query_models.go
@@ -83,6 +83,12 @@ func getQueryModelTSDB(query *tsdb.Query, timeRange *tsdb.TimeRange, dsInfo *mod
 		To:   endTime,
 	}
 	model.MaxDataPoints = query.MaxDataPoints
+	if model.MaxDataPoints == 0 {
+		model.MaxDataPoints = 10000 // 10k/series should be a reasonable place to abort!
+	}
 	model.Interval = time.Millisecond * time.Duration(query.IntervalMs)
+	if model.Interval.Milliseconds() == 0 {
+		model.Interval = time.Millisecond // 1ms
+	}
 	return model, nil
 }


pkg/tsdb/influxdb/flux/query_models.go
```diff
@@ -83,6 +83,12 @@ func getQueryModelTSDB(query *tsdb.Query, timeRange *tsdb.TimeRange, dsInfo *mod
 		To:   endTime,
 	}
 	model.MaxDataPoints = query.MaxDataPoints
+	if model.MaxDataPoints == 0 {
+		model.MaxDataPoints = 10000 // 10k/series should be a reasonable place to abort!
+	}
 	model.Interval = time.Millisecond * time.Duration(query.IntervalMs)
+	if model.Interval.Milliseconds() == 0 {
+		model.Interval = time.Millisecond // 1ms
```
torkelo, line 91: his seems like a very low default interval, we should use the interval defined in the data source settings, and if no interval there default to something like 10s

torkelo: Created an issue for the general problem, https://github.com/grafana/grafana/issues/27469

torkelo (approved):

PR #27479, opened 2020-09-09
Notifications: UX tweak to redesign of form and sections to make it left aligned and cleaner

torkelo:
Was not super happy with the new design, felt the sections and indentation made it more messy / unaligned than it needed to be. 

Before:
Screenshot from 2020-09-09 15-48-45

After:
Screenshot from 2020-09-09 15-58-19

Changes
* Collapsable section headings: moved caret to the right so there is need for indentation, now all headings are left-aligned and aligned with form/content 
* Removed top collapsable section & heading as it felt unnecessary to have two headings back to back 
* Tweaked the reset action for saved secret fields from button to link and from addon to suffix (inset in input), and from reset -> clear 
* Increased width, I would like to reduce widths of some inputs so they are not all 100% width but that require component changes and model changes (InputControl does not support width). So postponing that to limit scope for this change 

A saved secret field looked like this before
Screenshot from 2020-09-09 15-53-31

### DIFF
--- a/packages/grafana-ui/src/components/Collapse/CollapsableSection.tsx
+++ b/packages/grafana-ui/src/components/Collapse/CollapsableSection.tsx
@@ -17,10 +17,10 @@ export const CollapsableSection: FC<Props> = ({ label, isOpen, children }) => {
   return (
     <div>
       <div onClick={() => toggleOpen(!open)} className={styles.header}>
-        <Icon name={open ? 'angle-down' : 'angle-right'} size="xl" />
         {label}
+        <Icon name={open ? 'angle-down' : 'angle-right'} size="xl" />
       </div>
-      <div className={styles.content}>{open && children}</div>
+      {open && <div className={styles.content}>{children}</div>}
     </div>
   );
 };
@@ -32,7 +32,7 @@ const collapsableSectionStyles = (theme: GrafanaTheme) => {
       cursor: pointer;
     `,
     content: css`
-      padding: ${theme.spacing.md} 0 ${theme.spacing.md} ${theme.spacing.md};
+      padding: ${theme.spacing.md} 0;
     `,
   };
 };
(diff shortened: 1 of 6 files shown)


hugohaggmark (approved):
I think this looks great, but UX is not my forté :)

oanamangiurea (commented):
1. For consistency with what we have so far, and because it's hard to follow forms that are too wide, I would revert the max-width of the form to 400px; If it's bigger it almost loses the impression that it was left-aligned, and looks that we missed to center it. 
2. I do like it to be aligned, one thing I'm thinking though is with moving the caret to the right I'm inclined to think that those are links to other pages rather than an accordion menu. This can for sure be just my mental model and users not see it that way. One way to resolve it though, without losing the alignment, is to have the caret icon right-aligned, but that will only work by introducing horizontal lines to mark the entire link. This is how I would see it https://www.figma.com/file/yIVK8u2Os1MlCeoKhG9Brl/Notifications-Channels?node-id=105%3A1242

dprokop (approved):
Implementation wise it looks great. I also like the UX update, the form is now way cleaner. And making the first section not collapsable makes a lot of sense!

Regarding @oanamangiurea proposal to have line separators and right-aligned icon - I'm not a fan of this solution as it creates quite a lot of visual noise IMO - maybe it's a matter of choosinga a quite dark gray in the figma? But after reading this: https://www.smashingmagazine.com/2017/06/designing-perfect-accordion-checklist/ (Choosing The Position Of The Icon section) placing the icon on the far right may make most sense from the usability perspective.

torkelo:
Think I have something that works well 

Right aligned caret, and line, but line is only rendered in collapsed state 
Screen Shot 2020-09-10 at 20 11 39


Screen Shot 2020-09-10 at 20 11 22

PR #25980, opened 2020-07-01
Migration: Edit notification channel

peterholmberg:
**What this PR does / why we need it**:
Migrates the second part of new/edit notification channel. Also removes the angular related code.

**Which issue(s) this PR fixes**:
Fixes #25898

**Special notes for your reviewer**:
I opted for an implementation with a "New" page and an "Edit" page instead of using one page for both states as it were in the angularcontroller.

PR #27484, opened 2020-09-09
Alerting: Fix integration key so it's stored encrypted for Pagerduty notifier

marefr:
Fixes an issue introduced by migration in #25980 which removed storing the Pagerduty integrating key encrypted.


papagian (approved):
LGTM, it works as expected!

PR #27501, opened 2020-09-10
Transforms: Fix for issue in labels to fields transform where the new option value field name did not work properly

### LINKED ISSUE #27185: Transforms: Labels to fields refine 7.2 update and add docs
In 7.2 we adding the value name label option and changed a bit how this transform works. 

This update needs updated docs.

torkelo:
The labels to fields option to use a label value as the value field name worked when it was developed but the PR that solved the naming problems with data sources that want to control naming (https://github.com/grafana/grafana/pull/27186) that added displayNameFromDS to field config made this transform not work 100% when using this new value name option. 

This also adds docs and examples to the labels to fields transform

Closes #27185


mckn (approved):
LGTM!

PR #27502, opened 2020-09-10
PanelEdit: Drag and drop query order & UI tweaks

### LINKED ISSUE #2253: Reorder graph metrics via drag an and drop
It is not possible to reorder metrics via drag and drop for now. It'd be nice if that were possible.

torkelo:
Fixes #2253 

* Adds drag and drop ability (and drag handle) to the query row header.  
* Removes the up & down action icons 
* Moves the drag handle to the far right so it's grouped with the other row actions/icons. This looks better and cleaner to me.

Screen Shot 2020-09-10 at 12 20 35

Not sure about milestone for this.

### DIFF
--- a/e2e/suite1/specs/panelEdit_queries.spec.ts
+++ b/e2e/suite1/specs/panelEdit_queries.spec.ts
@@ -75,44 +75,6 @@ e2e.scenario({
 
     e2e().wait('@apiPostQuery');
 
-    // Change order or query rows
-    // Check the order of the rows before
-    e2e.components.QueryEditorRows.rows()
-      .eq(0)
-      .within(() => {
-        e2e.components.QueryEditorRow.title('B').should('be.visible');
-      });
-
-    e2e.components.QueryEditorRows.rows()
-      .eq(1)
-      .within(() => {
-        e2e.components.QueryEditorRow.title('A').should('be.visible');
-      });
-
-    // Change so A is first
-    e2e.components.QueryEditorRow.actionButton('Move query up')
-      .eq(1)
-      .click();
-
-    e2e().wait('@apiPostQuery');
-
-    // Avoid flaky tests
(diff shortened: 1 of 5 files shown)


dprokop:
@torkelo I think this works great. 

I don't like the drag handle being pushed far to the right, both in query row and transformations. My mental model of the placement of the icons when I've added the dnd handle on the left was that the left-hand side icons are "behavioral" while the right-hand side ones are "functional". We use the same pattern in the Organize fields transformer, where the "functional" part is the text input. I also recall seeing the drag handle in other UIs(if present at all) to be positioned on the left, unless it's a big card that is being moved.

See Gmail for reference where they use the same pattern I have used :
image

One solution I see here is to bring back the icon to the left and make it smaller so that it does not influence the visual flow that much. 

Btw I've found this article https://www.nngroup.com/articles/drag-drop/ that highlights some small touches that we can add to make the dnd experience better.

torkelo:
not 100% what the difference between behavioral and functional, in the Gmail case it seems more like always present actions/icons on the left, hover actions on the right. 

The biggest problem for me with drag handle to the left is that the indentation is then no longer aligned, so the row header title text is no longer perfectly aligned with the content text like before. 

Before
Screen Shot 2020-09-10 at 20 16 50

Now
Screen Shot 2020-09-10 at 20 16 07

dprokop:
I see your point with the alignment. What about the following: 
1. Making the drag  2/3 of the current size
2. Making it hidden by default and show only on row hover?

torkelo:
Tried a couple of hover versions (replace caret hover, show only on hover and move caret and refId out of the way, smaller version) but the hover effects were mostly just distracting. 

Feel having this with the other icons make for a cleaner / simpler UI. And the whole header is draggable anyway. Maybe UX can come up with a better solution but think this will have to do for now.

PR #27505, opened 2020-09-10
Prometheus: Fix min step variable interpolation

### LINKED ISSUE #27500: Template interval variable breaks Prometheus min step
**What happened**:

I cannot use template interval variables in the min field on Grafana versions > 7.1.5. The moment I set min step to use the variable, the panel goes blank and in the console log I get the following error:

`PanelQueryRunner Error Error: Invalid interval string`

When I use the variable in other places, for example in the panel title or as the range vector selector everything works fine. Please check the recording.

gif4

1. docker run -d -p 3000:3000 grafana/grafana:master
2. docker run -d -p 9090:9090 prom/prometheus
3. Login to Grafana and configure the Prometheus datasource in browser mode (direct)
4. Create template interval variable
5. Create a dashboard and panel
6. Set min step to variable

image

**Environment**:
- Grafana version: tested on **grafana/grafana:master** and **grafana/grafana:latest**
- Data source type & version: Latest Prometheus **prom/prometheus:latest**
- User OS & Browser: Chrome
- Grafana plugins: Vanilla

It works in 7.1.5, just tested it again following the steps described above.

zoltanbedi:

Fixes #27500

### DIFF
--- a/public/app/plugins/datasource/prometheus/datasource.test.ts
+++ b/public/app/plugins/datasource/prometheus/datasource.test.ts
@@ -1640,12 +1640,12 @@ describe('PrometheusDatasource', () => {
     it('should be 4 times the scrape interval if min step set to 1m and interval is 15s', () => {
       // For a 5m graph, $__interval is 15s
       ds.createQuery({ ...target, interval: '1m' }, { interval: '15s' } as any, 0, 300);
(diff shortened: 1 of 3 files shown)


beorn7 (approved):
Thanks for fixing.

torkelo (approved):
Looks good

PR #27509, opened 2020-09-10
Fix: Hide long query expression on collapsed query row.

### LINKED ISSUE #26708: InfluxDB: when collapse query builder, buttons can get lost if query line is long
**What happened**:
When InfluxDB query is collapsed, you see a summary view of the query instead of the query builder. At the right side of this query line are a series of little grey buttons (pencil icon, etc.). If the query summary line is longer than the width of the query pane, the buttons become "lost" or "hidden" (i.e., they get pushed off he right side of the UI). 

Buttons showing:
buttons-showing

Buttons hidden:
buttons-hidden

 I.e. the desired behavior would be for the buttons to stay where they are, and the query to get cut off to the left of those buttons.

1. go into edit mode on a dashboard that uses InfluxDB datasource
2. collapse one of the queries, so you see the query summary rather than the query builder
3. make the browser window narrow enough so that the query goes off the right side of its portion of the UI

screen-recording

The issue is not critical, since you can simply re-expand the query (into query builder view) to work around it, but it does seem undesirable.

**Environment**:
- Grafana version: 7.1.0 Enterprise Grafana Cloud
- User OS & Browser: Windows/Chrome
- Grafana plugins: many Ent plugins, incl Splunk, AppDynamics, Azure, DataDog, etc... but I don't think these are relevant to this issue
- Others:

peterholmberg:

Fixes #26708

Screenshot from 2020-09-10 15-08-19

After 

Screenshot from 2020-09-10 15-07-30


dprokop (changes requested):
image


Can you please add an ellipsis to the cropped text?

public/app/core/components/QueryOperationRow/QueryOperationRow.tsx
```diff
@@ -68,13 +68,10 @@ export const QueryOperationRow: React.FC<QueryOperationRowProps> = ({
...
         <Icon name={isContentVisible ? 'angle-down' : 'angle-right'} className={styles.collapseIcon} />
-        {title && <span className={styles.title}>{titleElement}</span>}
+        {title && <div className={styles.title}>{titleElement}</div>}
         {headerElement}
       </div>
-      {actions && actionsElement}
-      {draggable && (
-        <Icon title="Drag and drop to reorder" name="draggabledots" size="lg" className={styles.dragIcon} />
-      )}
+      {actions && <div>{actionsElement}</div>}
```
dprokop, line 74: Think there's something wrong here as the drag icon is missing

dprokop:
Ok, now it's good with the ellipsis. Think there's something wrong with the merge tho, as the drag and drop of the query rows does not work anymor.

dprokop (approved):
Works as expected. Fix tests and we are good to go:)

hugohaggmark:
@peterholmberg seems like Frontend tests are failing: https://drone.grafana.net/grafana/grafana/2861/1/10

peterholmberg:
@hugohaggmark they should be fixed in the latest commit?

hugohaggmark:
@peterholmberg try to create a new PR, seems like Drone is behaving super weirdly in this case

PR #27510, opened 2020-09-10
fixed: First time dropdown box rendering showing only 1000 pieces

jonny-gg:
First render of the drop-down box, showing only 1000 pieces of data

Fixes #
When the drop-down box has a large amount of data, such as 20000, the first click on the drop-down menu will cause rendering delay


hugohaggmark:
closed in favor or #27525

PR #27511, opened 2020-09-10
@grafana/e2e: avoid query data flakiness

stevenvachon:


ryantxu (approved):

packages/grafana-e2e/src/flows/configurePanel.ts
```diff
@@ -127,9 +127,15 @@ export const configurePanel = (config: PartialAddPanelConfig | PartialEditPanelC
     // @todo instead wait for '@pluginModule'
     e2e().wait(2000);
 
-    e2e().wait('@chartData');
-
     if (!isExplore) {
+      // Avoid cache flakiness (where @chartData isn't requested)
+      e2e()
+        .get('.refresh-picker-buttons .btn')
```
hugohaggmark, line 133: nit: Shouldn't this use e2e-selectrors?

stevenvachon: Yeah. Thank you for pointing this out as I've now just updated one of my OKRs that I was thinking of cancelling. I had originally wanted to update all fields with aria labels, but now I'll only do so on elements that _@grafana/e2e_ selects.

PR #27520, opened 2020-09-10
Elasticsearch: Allow fields starting with underscore

### LINKED ISSUE #20692: Unable to use elasticsearch fields starting with underscore in Grafana
**What happened**: I tried to use the elasticsearch datasource in Grafana and it cannot detect my timestamp field as it starts with (two) underscores. 

**What you expected to happen**: Grafana should allow fields starting with one or more underscores

**How to reproduce it (as minimally and precisely as possible)**: Use `__timestamp` instead of `@timestamp` as the name of the date field in your mapping 

**Anything else we need to know?**: A large number of my fields start with an underscore. The system is large and old with multiple writers and consumers, which makes it very difficult to change all these fields to not start with an underscore.

**Environment**:
- Grafana version: v6.3.5 (67bad72)
- Data source type & version: elasticsearch 6.4
- User OS & Browser: macOS Mojave, Chrome
- Grafana plugins: elasticsearch
- Others:

Elfo404:

Currently in elasticsearch we filter out all the fields that starts with `_` even though they are completely valid as the idea was to exlude all the metadata fields: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-fields.html#mapping-fields
This fixes this behavior by only filtering against that list.

Fixes #20692

I'm not sure if there is a better way to determine if a filed is a metadata field as I couldn't find any information about it, but I'm not completely happy with the solution as in the future this list could change.

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/datasource.test.ts
+++ b/public/app/plugins/datasource/elasticsearch/datasource.test.ts
@@ -378,8 +378,12 @@ describe('ElasticDatasource', function(this: any) {
               mappings: {
                 metricsets: {
                   _all: {},
+                  _meta: {
+                    test: 'something',
+                  },
                   properties: {
                     '@timestamp': { type: 'date' },
+                    __timestamp: { type: 'date' },
                     beat: {
                       properties: {
                         name: {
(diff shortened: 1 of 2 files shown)


zoltanbedi (approved):
I think this looks good. 🥠

PR #27523, opened 2020-09-11
OAuth: Fix token refresh failure when custom SSL settings are configured for OAuth provider

### LINKED ISSUE #27514: OAuth token refresh fails when custom SSL settings are configured for oauth provider
When using generic oauth with SSL configured and tls_skip_verify_insecure=true, I was able to log in with my oauth provider, but when the token expired and grafana tried to fetch a new token (using the refresh token), I saw http ssl errors.

When the original token is fetched in login_oauth.go, the code gets an http client based on the appropriate oauth settings and setit into the context to be used.  This same functionality should be centralized and used from wherever oauth token http operations are initiated.

billoley:
Fixes #27514

OAuth token refresh fails when custom SSL settings are configured for oauth provider

When using generic oauth with SSL configured and tls_skip_verify_insecure=true, I was able to log in with my oauth provider, but when the token expired and grafana tried to fetch a new token (using the refresh token), I saw http ssl errors.

When the original token is fetched in login_oauth.go, the code gets an http client based on the appropriate oauth settings and setit into the context to be used. This same functionality should be centralized and used from wherever oauth token http operations are initiated.

Moved code for creating an http client using the OAuth SSL setting to a new package oauthtoken.
This allows code to be refactored from login_oauth.go so that it can be used from there, ds_proxy.go, and anywhere else that calls OAuth APIs and needs to configure a client.


pkg/api/pluginproxy/ds_proxy.go
```diff
@@ -311,8 +314,9 @@ func addOAuthPassThruAuth(c *models.ReqContext, req *http.Request) {
 		return
 	}
 
-	provider := authInfoQuery.Result.AuthModule
-	connect, ok := social.SocialMap[strings.TrimPrefix(provider, "oauth_")] // The socialMap keys don't have "oauth_" prefix, but everywhere else in the system does
+	// The socialMap keys don't have "oauth_" prefix, but everywhere else in the system does
+	provider := strings.TrimPrefix(authInfoQuery.Result.AuthModule, "oauth_")
+	connect, ok := social.SocialMap[provider]
```
marefr, line 319: Maybe can create a function for this in social package that encapsulates the handling of the "oauth_" prefix and getting a provider/connect from SocialMap?

marefr (changes requested):
Great. However, I have some minor suggestions regarding code structure.

pkg/services/oauthtoken/oauth_http_client.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/grafana/grafana/pkg/infra/log"
+
+	"github.com/grafana/grafana/pkg/setting"
+)
+
+var (
+	logger = log.New("oauthtoken")
+)
+
+func GetOAuthHttpClient(name string) (*http.Client, error) {
```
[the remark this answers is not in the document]
billoley: Of course.  Changes made.

pkg/api/pluginproxy/ds_proxy.go
billoley: Done

marefr (approved):
Great. LGTM

marefr:

PR #26395, opened 2020-07-17
Chore: Enable gocyclo linter

aknuds1:
**What this PR does / why we need it**:
Enable [gocyclo](https://github.com/fzipp/gocyclo) linter, which detects overly complex Go functions.

In my opinion, it's very useful since it prevents too long functions, which are really difficult to follow and understand (the old rule of thumb from Code Complete is that a function should generally not exceed the height of the screen more or less). Some functions become long for a reason (usually switch statements), and I've marked them with a `nolint` directive.

**NB:** The changes, although just refactoring, are a bit of a mouthful taken altogether, I know. I can break up into smaller PRs if it would help reviewers!

**Update:** Something breaks on login, so I will split this into several PRs and check more thoroughly to see what goes wrong (probably something in the settings parsing).

PR #27533, opened 2020-09-11
BootData: Fix nav tree sort regression

marefr:
#26395 introduced a regression regarding sort order of nav tree
items set in Grafana boot data and used for rendering the sidemenu.
This fixes so that sort happens after RunIndexDataHooks is called
in case the hook make changes to the nav tree.

This resolves correct placement of the Enterprise reporting nav tree item.

Ref #26395 


Clarity-89 (approved):
Checked the UI and the order is correct now.

papagian (approved):
It works as expected!

PR #27534, opened 2020-09-11
Docs: Update style guide for backend

sakjur:

Fixes #


contribute/style-guides/backend.md
```diff
@@ -30,6 +30,83 @@ To run all linters, use the `lint-go` Makefile target:
...
+
+func Function(str string, args Args) (string, error) {
+```
+
+or
+
+```go
+type FunctionOpts func(string) (string, error)
+
+func Function(str string, opts ...FunctionOpts)
```
sakjur, line 94: (function should still return (string, error) )

PR #27535, opened 2020-09-11
MySQL Datasource: Add timezone parameter

### LINKED ISSUE #18120: MySQL Timezone/Grafana Interface Issue
I'm unsure if the best place to place this is bug reports or feature enhancements.  There are numerous similar issues like my problem, but I do not think I've found one that exactly addresses my issue.

My Grafana dashboard is set to UTC.  I have data in MySQL database, with a column named 'time' that is datetime type--which implies UTC.  The server that hosts Grafana is linux and set to UTC.  The server that hosts MySQL is linux and set to local time (Central US).  I have found a variety of help posts say to set MySQL server to UTC--but I cannot do this, the MySQL DB server is used by many others and I've been told it needs to stay in local time.

**What:  A way to have grafana correctly display data when the MySQL server is set to local time and all other pieces are set to UTC.**

Why:  Otherwise current data is not displayed or incorrect data is not displayed.

More info:
My grafana query looks like this:
`SELECT
  time as time_sec,
 ...
WHERE $__timeFilter(time) AND ...
ORDER BY time ASC;`

When I run this the data displayed is shift back from current 5 hours (my UTC offset, I  know MySQL is returning it as such).  One solution I have found involves setting the MySQL session time zone, which if I go into MySQL and do this before I do my query that works, but grafana only allows 1 line of MySQL to be executed, so this is not option--perhaps adding the ability to execute multiple lines of MySQL would be a solution?

I've also found in some thread to wrap time with unix_timestamp().  When I do this, the data _looks_ to display correctly, but it is incorrect.  If I change the end time from now to now+6h, I see the data goes out 5 hours into the 'future' (my current UTC offset) and the data for the current time is actually data from 5 hours previous.

If this problem has been solved and I just have been unable to find the solution, I apologize, but I've have spent multiple work days trying to find a solution and still do not have a solution.  I appreciate any insights.

andipabst:
**What this PR does / why we need it**: If the database (or the host of the database) is in a timezone other than UTC, the returned times are wrong. This PR fixes this by introducing a timezone parameter for the datasource, which sets the time zone for a session. This can be used to open the session in a UTC timezone.

Fixes #18120

**Special notes for your reviewer**: I don't see a way to add tests for this scenario unless changing the docker container to have the 'mysql_tests' container/DB in a different timezone.


marefr:
Thanks. Need to prepare for some time reviewing this. Until then, you're referring to fixing #18120, but that issue involves Mysql, postgres and mssql. I'm not sure if mssql is applicable since it doesn't support setting a time zone per session as mysql and postgres. What's your idea here, only fixing this for mysql or postgres as well?

docs/sources/features/datasources/mysql.md
```diff
@@ -35,6 +35,7 @@ Name | Description
 *Database* | Name of your MySQL database.
 *User* | Database user's login/username
 *Password* | Database user's password
+*Session Timezone* | Specify the time zone used in the database session, e.g. `Europe/Berlin` or `+02:00`. The value will be set int the session with `SET time_zone='...'`. If you leave this field empty, the timezone will not be changed. You can find more information in the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html).
```
oddlittlebird, line 38:
```suggestion
*Session Timezone* | Specify the time zone used in the database session, such as `Europe/Berlin` or `+02:00`. Set the value in the session with `SET time_zone='...'`. If you leave this field empty, then the time zone will not be changed. You can find more information in the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html). 
```

andipabst:
Thank you for your response. Unfortunately I can not find any information in #18120 about the bug in MSSQL, except the tag?

Apart from this, I think the other two databases need a different kind of fix, so this PR would indeed only fix part of #18120, sorry!

I think for MSSQL and PostgreSQL `AT TIME ZONE` could be used (see https://docs.microsoft.com/en-us/sql/t-sql/queries/at-time-zone-transact-sql?view=sql-server-ver15 and https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-ZONECONVERT). A equivalent for MySQL would be `CONVERT_TZ` with a different syntax. This would involve changing the query instead of the database connection string, so I am not sure, if it would be a good solution. But I could manually verify that it would work. Do you think this would be worth pursuing and if so, could you please guide me in a direction on what would be the best way to implement this?

ying-jeanne (approved):
The code looks good, tested works fine. I would prefer doc team review the doc changes.

ying-jeanne:
@andipabst you need to sign the contributor license agreement, and merge origin/main into your branch. i can't push the changes into your fork, the merge i did for testing is here if you would like to check. https://github.com/grafana/grafana/tree/27535

docs/sources/features/datasources/mysql.md
```diff
@@ -35,6 +35,7 @@ Name | Description
 *Database* | Name of your MySQL database.
 *User* | Database user's login/username
 *Password* | Database user's password
+*Session Timezone* | Specify the time zone used in the database session, such as `Europe/Berlin` or `+02:00`. Set the value in the session with `SET time_zone='...'`. If you leave this field empty, then the time zone will not be changed. You can find more information in the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html).
```
achatterjee-grafana, line 38:
```suggestion
*Session Timezone* | Specify the time zone used in the database session, such as `Europe/Berlin` or `+02:00`. Set the value in the session with `SET time_zone='...'`. If you leave this field empty, then the time zone is not updated. For more information, refer to the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/time-zone-support.html). 
```

achatterjee-grafana (changes requested):
Requesting minor doc change.

docs/sources/features/datasources/mysql.md
andipabst: Updated in merge commit

andipabst: Updated after merge

achatterjee-grafana (approved):

andipabst:
@ying-jeanne and @achatterjee-grafana  Thanks for your review. I signed the contributor license agreement, merged the main-branch and updated the documentation.

achatterjee-grafana:
We will merge the changes as soon as the checks complete. @ying-jeanne, should this be backported to 7.5.x as well?

public/app/plugins/datasource/mysql/partials/config.html
```diff
@@ -27,6 +27,23 @@ <h3 class="page-heading">MySQL Connection</h3>
...
+    <input
+            type="text"
+            class="gf-form-input gf-form-input--has-help-icon"
+            ng-model="ctrl.current.jsonData.timezone"
+            spellcheck='false'
+            placeholder="(default)"
+    ></input>
+    <info-popover mode="right-absolute">
+      Specify the time zone used in the database session, e.g. <code>Europe/Berlin</code> or <code>+02:00</code>.
+      The value will be set int the session with <code>SET time_zone='...'</code>. If you leave this field empty,
```
marefr, line 41: Might want to add here and/or in the docs why and in what scenarios configuring this is a good idea.

```suggestion
      The value will be set in the session with <code>SET time_zone='...'</code>. If you leave this field empty,
```

marefr (changes requested):
Added a suggestion. Think some more documentation of why a user would configure this setting would be good, see comment.

public/app/plugins/datasource/mysql/partials/config.html
achatterjee-grafana, line 41:
```suggestion
     The value is set in the session with <code>SET time_zone='...'</code>. If you leave this field empty,
```

public/app/plugins/datasource/mysql/partials/config.html
```diff
@@ -27,6 +27,23 @@ <h3 class="page-heading">MySQL Connection</h3>
...
+      the timezone will not be changed. You can find more information in the
```
achatterjee-grafana, line 42:
```suggestion
      the timezone is not updated. You can find more information in the
```

achatterjee-grafana (changes requested):
Added a minor change request.

public/app/plugins/datasource/mysql/partials/config.html
andipabst: I added some information on why the configuration is necessary. Can you please have another look? Thanks!

achatterjee-grafana (approved):

marefr (approved):
LGTM

ying-jeanne:
> We will merge the changes as soon as the checks complete. @ying-jeanne, should this be backported to 7.5.x as well?

thanks for the remind :), i think we don't need the backport, main branch should be enough.

sushma-4:
This fix did not work for me since grafana ready mysql column as UTC automatically despite me setting the session-timezone

PR #27536, opened 2020-09-11
Api: Add /healthz endpoint for health checks

bergquist:
kuberentes (and I'm sure other orchestrators does as well) support two
kind of checks. readiness checks and liveness checks. Grafanas current
`/api/health` endpoint requires database access which might not
always be required for the instance to be considered active.

ref https://github.com/grafana/grafana/issues/3302

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -437,7 +437,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
 
 	// Health check
-	r.Get("/api/health", hs.healthHandler)
+	r.Get("/api/health", hs.apiHealthHandler)
+	r.Get("/healthz", hs.healthzHandler)
 
 	r.Get("/*", reqSignedIn, hs.Index)
 }
--- a/pkg/api/http_server.go
+++ b/pkg/api/http_server.go
@@ -375,12 +375,19 @@ func (hs *HTTPServer) metricsEndpoint(ctx *macaron.Context) {
 		ServeHTTP(ctx.Resp, ctx.Req.Request)
 }
 
-func (hs *HTTPServer) healthHandler(ctx *macaron.Context) {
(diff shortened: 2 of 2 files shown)


bergquist:
Totally fine to punt this one if you don't think it belongs in 7.2-beta2

pkg/api/api.go
```diff
@@ -437,7 +437,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
 
 	// Health check
-	r.Get("/api/health", hs.healthHandler)
+	r.Get("/api/health", hs.apiHealthHandler)
+	r.Get("/healthz", hs.healthzHandler)
```
torkelo, line 441: is this a common naming for this type of url endpoint? The other option I had was to just have this be a param for the /api/health endpoint `?ignore_db`  , feels messy to have two health endpoints

papagian (approved):
LGTM; it works as expected!

pkg/api/api.go
papagian: I was thinking also about having an extra parameter in the existing endpoint but I don't know since it's served under a different path (not under /api).

bergquist: `healthz` or `healthy` is kind the standard name for health checks in the Kubernetes world. I prefer `healthz` since its the likelihood of a collision is very low. But we could put them on another route. Prometheus has all these kinds of routes under `/-/` so `/-/healthz` for example. 

The `/api/health` endpoint could introduce more checks, ex checking that it can reach all configured datasources so I think it would be nice to have separate endpoints.

bergquist: `/healthz` is definitely a convention that many people follow. Ex

https://github.com/healthz
https://grafana.com/healthz

https://vimeo.com/173610242

@torkelo unless you object I would like to merge this for 7.3 so we can avoid using robots.txt in HG :)

torkelo: no objection! merge this baby!

torkelo (approved):

glensc:
kuber**en**tes :grin:

PR #27538, opened 2020-09-11
Elasticsearch: Add support for date_nanos type

### LINKED ISSUE #20963: Elasticsearch datasource: support date_nanos type
Support using an Elasticsearch field with type `date_nanos` (ES 7.x feature for nano-second precision timestamp) as the "Time field name" when configuring an ES datasource.

Note: I'm not asking for Grafana to handle the value with nanosecond precision (which I imagine would be difficult in Javascript since it only has double-precision types).

**Why is this needed**:
In Grafana 6.5.1, trying to use an ES index where `@timestamp` is of type date_nanos just reports `no date field named @timestamp found`.

Elfo404:
Adds `date_nanos` support to elasticsearch datasource. previously `date_nanos` fields were not considered as date fields.

Fixes #20963


torkelo (approved):
Thanks for contributing!

PR #27539, opened 2020-09-11
Elasticsearch: Increase maximum geohash aggregation precision to 12

### LINKED ISSUE #16188: Elasticsearch Geohash Aggregation arbitrarily limits to 7 units of precision

# Allow geohash precision as supported by elasticsearch
I believe this is simply because the Kibana Geo Visualizer defaults to the same behavior, limiting you to 7 units of precision in your geohash. While I certainly understand on large data sets this could represent an overwhelming amount of data, on sparse sets, you sometimes need the precision from higher levels of hashing.

Should easily support up to 9, perhaps more? While kibana does default to a limit of 7, there is an arbitrary json field that allows you to override to higher levels.

**Why is this needed**:
Allow useful display of precision location data in sparse sets.

Happy to provide a PR for this, just need to get my Grafana Dev Environment set back up.

Elfo404:
Since version 2 Elasticsearch supports precision for geohash grid aggregation up to 12 (was 7 in <2).
This PR just changes the maximum allowed precision from 7 to 12.

Fixes #16188

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/bucket_agg.ts
+++ b/public/app/plugins/datasource/elasticsearch/bucket_agg.ts
@@ -144,8 +144,8 @@ export class ElasticBucketAggCtrl {
           break;
         }
         case 'geohash_grid': {
-          // limit precision to 7
-          settings.precision = Math.max(Math.min(settings.precision, 7), 1);
+          // limit precision to 12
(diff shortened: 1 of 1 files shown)


zoltanbedi (approved):
lgtm

PR #27541, opened 2020-09-11
Grafana-UI: Expand ConfirmModal docs

Clarity-89:
Document `ConfirmModal` props and add usage example.


jessover9000 (commented):
Cool! I think the explanation part of what the component does could use some extra work but we can do that another time :) A dev should probably review the props and whether everything is in order here :D

peterholmberg (approved):
Super

PR #27551, opened 2020-09-11
API: Enrich add user to org endpoints with user ID in the response

AgnesToulet:
Enrich following endpoints to return the ID of the added user:
- `POST /api/org/invites` (if user exists)
- `POST /api/orgs/:orgId/users`
- `POST /api/org/users`

It makes it easier to have access to the resource ID for auditing without any breaking change or important information leakage.


marefr (approved):
LGTM

papagian (approved):
LGTM

PR #27552, opened 2020-09-11
Auth: Add SigV4 auth option to datasources

### LINKED ISSUE #7245: [Feature Request] Support AWS SigV4 authentication for Elasticsearch Datasource
I would like to use an [Amazon Elasticsearch Service](https://aws.amazon.com/elasticsearch-service/) cluster as an Elasticsearch datasource in Grafana. This works just fine if I allow public access to the cluster, but I'd prefer to lock it down to authenticated users only.

The only user-level access control Amazon ES supports is [through AWS SigV4 signing the HTTP requests to Elasticsearch with IAM credentials](http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-createupdatedomains.html#es-createdomain-configure-access-policies). For example, [here's how the Golang `elastic` client does it](https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service).

I had a look at how the `proxy` mode of the ElasticSearch datasource works, and it looks like it would be pretty straightforward to build this signing into `dataproxy.go`, similarly to [how it's handled for the CloudWatch data source.](https://github.com/grafana/grafana/blob/master/pkg/api/dataproxy.go#L81-L84)

wbrowne:
Adds Sigv4 authentication to the Grafana backend by introducing a RoundTripper middleware chain that will take care of the request signing.

The Sigv4 option is tied to HTTP-based datasources (namely Prometheus, Elasticsearch for now) and is accessible via the auth options in the datasource edit UI. The information it needs is identical to the Cloudwatch datasource (excluding the `Custom Metrics` field)

Fixes #7245 - but probably worth mentioning that this work is mostly done out of request by AWS as Sigv4 is their defacto request signing process. More info [here](https://docs.google.com/document/d/1QRNIW-pkQLOxiPDGsDalzypg25xwMfhbqsq7TeYbOUE/edit?ts=5f477860#)

Sigv4 isn't so much a standalone authentication mechanism but more a way to authenticate requests. It's an AWS specific thing and you can read more about it [here](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html)

Screenshot 2020-09-17 at 11 15 31
Screenshot 2020-09-17 at 11 15 51


packages/grafana-ui/src/components/DataSourceSettings/Sigv4AuthSettings.tsx
```diff
@@ -0,0 +1,242 @@
...
+export const Sigv4AuthSettings: React.FC<HttpSettingsProps> = props => {
+  const { dataSourceConfig } = props;
+
+  const authProviderOptions = [
+    { label: 'Access & secret key', value: 'keys' },
+    { label: 'Credentials file', value: 'credentials' },
+    { label: 'ARN', value: 'arn' },
+  ] as SelectableValue[];
+
+  const regions = [
```
wbrowne, line 16: These are hardcoded for simplicity for now, but can be changed to dynamically fetch if necessary

aocenas:
The FE code looks good to me, but probably somebody else should check the backend to give it approve.

One question should this be somehow aligned with existing AWS datasources? Meaning we have a few and they basically just copy paste auth code between them which is bad if we ever try to fix a bug or something and realize we have to do it on multiple places and will probably miss one of them.

This seems like it could be used by plugins to sort of say, just use generic grafanas aws auth so I do not have to handle it myself. That would make adding aws plugins so much easier. Not sure though about the backend part how that would work.

pkg/models/sigv4.go
```diff
@@ -0,0 +1,78 @@
...
+func (m *Sigv4Middleware) credentials() *credentials.Credentials {
+	authType := AuthType(m.Config.AuthType)
+
+	switch authType {
+	case Keys:
+		return credentials.NewStaticCredentials(m.Config.AccessKey, m.Config.SecretKey, "")
+	case Credentials:
+		return credentials.NewSharedCredentials("", m.Config.Profile)
+	case ARN:
+		s := session.Must(session.NewSession())
```
wbrowne, line 73: Not entirely sure if this is the most correct way to achieve what we need here. The session has a timeout which is defaulted to 15m - maybe we can be more aggressive. I'm waiting also for the rework of the current Cloudwatch auth work to be completed - see [here](https://github.com/grafana/grafana/pull/25548)

wbrowne:
@aocenas thanks for giving the frontend a look over :) 

One thing to point out is that Sigv4 is more about request signing than typical authentication and this was done in a way to make this request signing option generically available for any datasource that may operate within the scope of AWS. 

While I'm not super familiar with the extent of the range of AWS datasources yet, I do agree that even in this case there could be opportunities to unify some of the parts that are shared. As the information for Sigv4 is essentially the same as the info required by the existing CW datasource, I tried to re-use the existing Config Editor component but as that is strictly tied to a singular datasource plugin it felt like for now it was safer and simpler to split apart.

aocenas:
hmm so not sure I understand the context here but is this something that every aws datasource should use? Right now there are 2 plugins in development, x-ray and timestream so I wonder if there is any action needed on their part in this context.

wbrowne:
@aocenas So when using the AWS SDK, AWS will already take care of the signing for you. As per their docs:
>If you are using one of the AWS SDKs (including the SDK for C++, SDK for Go, SDK for Java, AWS SDK for JavaScript, AWS SDK for .NET, SDK for PHP, SDK for Python (Boto3), or SDK for Ruby), you do not have to manually perform the steps of deriving a signing key and adding authentication information to a request. The SDKs perform this work for you. You need to manually sign requests only if you are directly making HTTP or HTTPS requests.

In our particular case, we're just interested in providing the option for Sigv4 signing in cases where people will use datasources that are running within the AWS network (Prometheus, Elasticsearch, etc.)

conf/defaults.ini
```diff
@@ -307,6 +307,9 @@ oauth_state_cookie_max_age = 600
 # limit of api_key seconds to live before expiration
 api_key_max_seconds_to_live = -1
 
+# Set to true to enable Sig4 authentication option for HTTP-based datasources
```
aknuds1, line 310:
```suggestion
# Set to true to enable SigV4 authentication option for HTTP-based datasources
```

pkg/models/datasource_cache.go
```diff
@@ -155,10 +157,19 @@ func (ds *DataSource) GetHttpTransport() (*dataSourceTransport, error) {
 		IdleConnTimeout:       90 * time.Second,
 	}
 
+	// Set default next round tripper to the default transport
+	var next http.RoundTripper = transport
```
aknuds1, line 161: I think this is more idiomatic/simpler:
```suggestion
	next := transport
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,66 @@ func TestDataSourceDecryptionCache(t *testing.T) {
 	})
 }
 
+func TestDataSourceSigv4Auth(t *testing.T) {
+	Convey("When caching a datasource proxy with middleware", t, func() {
+		clearDSProxyCache()
+		setting.Sigv4AuthEnabled = true
```
aknuds1, line 297:
```suggestion
		origEnabled := setting.Sigv4AuthEnabled 
		setting.Sigv4AuthEnabled = true
		t.Cleanup(func() {
			setting.Sigv4AuthEnabled = origEnabled
		})
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,66 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+
+		ds := DataSource{
+			JsonData: json,
+		}
+
+		t, err := ds.GetHttpTransport()
+		So(err, ShouldBeNil)
+
+		Convey("Sigv4 is in middleware chain if configured in JsonData", func() {
+			m1, ok := interface{}(t.next).(*Sigv4Middleware)
```
aknuds1, line 310: Is the type conversion to `interface{}` really necessary here?

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,66 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+			So(ok, ShouldEqual, true)
+
+			_, ok = interface{}(m1.Next).(*http.Transport)
+			So(ok, ShouldEqual, true)
+		})
+	})
+
+	Convey("When caching a datasource proxy with middleware", t, func() {
+		clearDSProxyCache()
+		setting.Sigv4AuthEnabled = true
```
aknuds1, line 320:
```suggestion
		origEnabled := setting.Sigv4AuthEnabled 
		setting.Sigv4AuthEnabled = true
		t.Cleanup(func() {
			setting.Sigv4AuthEnabled = origEnabled
		})
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,66 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+
+		Convey("Should not include sigv4 middleware if not configured in JsonData", func() {
+			_, ok := interface{}(t.next).(*http.Transport)
+			So(ok, ShouldEqual, true)
+		})
+	})
+
+	Convey("When caching a datasource proxy with middleware", t, func() {
+		clearDSProxyCache()
+		setting.Sigv4AuthEnabled = false
```
aknuds1, line 335:
```suggestion
		origEnabled := setting.Sigv4AuthEnabled 
		setting.Sigv4AuthEnabled = true
		t.Cleanup(func() {
			setting.Sigv4AuthEnabled = origEnabled
		})
```

pkg/models/sigv4.go
```diff
@@ -0,0 +1,78 @@
...
+
+type AuthType string
+
+const (
+	Keys        AuthType = "keys"
+	Credentials AuthType = "credentials"
+	ARN         AuthType = "arn"
+)
+
+type Sigv4Middleware struct {
```
aknuds1, line 22: Isn't it better to name it "SigV4Middleware" to be consistent with how SigV4 is usually written?
```suggestion
type SigV4Middleware struct {
```

pkg/models/sigv4.go
```diff
@@ -0,0 +1,78 @@
...
+	case Keys:
+		return credentials.NewStaticCredentials(m.Config.AccessKey, m.Config.SecretKey, "")
+	case Credentials:
+		return credentials.NewSharedCredentials("", m.Config.Profile)
+	case ARN:
+		s := session.Must(session.NewSession())
+		return stscreds.NewCredentials(s, m.Config.AssumeRoleARN)
+	}
+
+	return nil
```
aknuds1, line 77: Why not return an error from here instead, including the invalid authType to ease debugging?

pkg/setting/setting.go
```diff
@@ -144,6 +144,7 @@ var (
 	AdminPassword    string
 	LoginCookieName  string
 	LoginMaxLifetime time.Duration
+	Sigv4AuthEnabled bool
```
aknuds1, line 147: See my comment above about casing of SigV4.
```suggestion
	SigV4AuthEnabled bool
```

pkg/setting/setting.go
```diff
@@ -991,6 +993,10 @@ func readAuthSettings(iniFile *ini.File, cfg *Cfg) (err error) {
 	cfg.OAuthCookieMaxAge = auth.Key("oauth_state_cookie_max_age").MustInt(600)
 	SignoutRedirectUrl = valueAsString(auth, "signout_redirect_url", "")
 
+	// Sigv4
```
aknuds1, line 996:
```suggestion
	// SigV4
```

aknuds1 (commented):
I left some comments for discussion, before digging too deep.

pkg/models/datasource_cache.go
wbrowne: Maybe it's just me getting used to the `http` package, but I found it helpful to explicitly specify `transport` as a round tripper, as that what allows us to form the middleware chain, with transport being the base. If that's obvious though, then I'm happy to simplify!

conf/defaults.ini
```diff
@@ -307,6 +307,9 @@ oauth_state_cookie_max_age = 600
...
+sigv4_auth_enabled = true
```
papagian, line 311: you have to add this configuration also in the [sample.ini](https://github.com/grafana/grafana/blob/master/conf/sample.ini#L277) so as users that copy this file for creating their custom configuration will be able to override it.

packages/grafana-ui/src/components/DataSourceSettings/Sigv4AuthSettings.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { Button, InlineFormLabel, Input } from '..';
+import Select from '../Forms/Legacy/Select/Select';
+
+export const Sigv4AuthSettings: React.FC<HttpSettingsProps> = props => {
+  const { dataSourceConfig } = props;
+
+  const authProviderOptions = [
+    { label: 'Access & secret key', value: 'keys' },
+    { label: 'Credentials file', value: 'credentials' },
+    { label: 'ARN', value: 'arn' },
```
papagian, line 13: There is a [PR](url) for refactoring the respective Cloudwatch authentication where ARN [will not be a separate authentication provider](https://github.com/grafana/grafana/pull/25548/files#diff-d554d037cd36865918be4c50e78ec9ddR19).
Maybe we want to have the same approach here as well
@aknuds1 what do you think?

pkg/models/sigv4.go
papagian: Following my above comment check how the role assumption is suggested to be implemented [here](https://github.com/grafana/grafana/blob/34fb224ec7e794d4b4b284ead20b857d5bc7b501/pkg/tsdb/cloudwatch/cloudwatch.go#L126)

pkg/models/datasource_cache.go
aknuds1: I still don't see the point to be honest. If it's useful to be explicit about variable types when declaring them, why don't we do it everywhere?

packages/grafana-ui/src/components/DataSourceSettings/Sigv4AuthSettings.tsx
aknuds1: @papagian I think you are right, assuming that the approach in the [PR](https://github.com/grafana/grafana/pull/25548) is correct. I'm not 100% on top of that one right now, but I think it's correct that assuming a role isn't an authentication type of itself, but something additional.

marefr:
@aocenas are you using AWS SDK for those new plugins? We've discussed making SigV4 available to use for backend plugins that are HTTP data sources that uses regular Go HTTP client directly or indirectly to communicate. However, something we could add later after this. WDYT?

pkg/models/sigv4.go
aknuds1: Yeah, I do think the session `s` should be from one of the authentication methods. I think how it's created here, it will use the default SDK method implicitly?

pkg/models/datasource_cache.go
wbrowne: >If it's useful to be explicit about variable types when declaring them, why don't we do it everywhere?
Ah well just like anything, not everything has to be black and white. In some cases it can be useful. Though, like I said this is likely just me getting used to `go`. Will fix

packages/grafana-ui/src/components/DataSourceSettings/Sigv4AuthSettings.tsx
wbrowne: Yeah I made a comment about this previously. Happy to wait merging this until that PR is figured out 😄

pkg/models/sigv4.go
wbrowne: Yep that's what I gathered from my testing of it too @aknuds1

marefr (commented):
Great work. Some nits and grits here and there, but not super important right now since those can be fixed and refactored later/before merge. Important now to get a feel for the implementation and important decision/details that needs to be discussed further. 

Some things that pops up in my head:
- Do we need to cache the session/credentials in the middleware or does AWS SDK do that automatically?

pkg/models/datasource_cache.go
aknuds1: Thanks. Yeah, I understand there's leeway depending on the situation, I just don't personally see how it differs from the general case.

aknuds1: @wbrowne @marefr Ah, now I see. `http.RoundTripper` is an interface. I'm going to play with the code locally, to get a better idea.

aknuds1: I think I would write `next := http.RoundTripper(transport)`.

docs/sources/packages_api/data/grafanaconfig.md
```diff
@@ -323,6 +324,14 @@ passwordHint: any;
 pluginsToPreload: string[];
 ```
 
+### sigV4AuthEnabled property
```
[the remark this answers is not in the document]
wbrowne: Ooh good to know 😄

pkg/models/datasource_cache.go
wbrowne: Hah I didn't even realize it was necessary in this case. Yeah `next := http.RoundTripper(transport)` seems reasonable I think if people are cool with that

packages/grafana-ui/src/components/DataSourceSettings/DataSourceHttpSettings.tsx
```diff
@@ -189,6 +190,21 @@ export const DataSourceHttpSettings: React.FC<HttpSettingsProps> = props => {
...
+                labelClass="width-13"
+                checked={dataSourceConfig.jsonData.sigV4Auth || false}
+                onChange={event => {
+                  onSettingsChange({
+                    jsonData: { ...dataSourceConfig.jsonData, sigV4Auth: event!.currentTarget.checked },
+                  });
+                }}
+              />
+            </div>
+          )}
```
dprokop, line 206: Wonder if instead of passing the entire config object which is used only for retrieving the `sigV4AuthEnabled` flag we should just pass the `sigV4AuthEnabled` property? I guess there might be other auth options in the future, but do we plan to add any short term?

dprokop (commented):
@wbrowne front-end wise this looks good. Left one question only.

packages/grafana-ui/src/components/DataSourceSettings/DataSourceHttpSettings.tsx
wbrowne: Great point @dprokop. It would be unlikely so that's a good change to make I reckon. Thanks for pointing that out 😄

packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  return (
+    <>
+      <h3 className="page-heading">Sigv4 Details</h3>
+      <div className="gf-form-group">
+        <div className="gf-form-inline">
+          <div className="gf-form">
+            <InlineFormLabel
+              className="width-14"
+              tooltip="Which AWS credentials chain to use. AWS SDK Default is the recommended option for EKS, ECS or if you've attached an IAM role to your EC2 instance."
```
achatterjee-grafana, line 93: Add a comma:  "..ECS,"

conf/sample.ini
```diff
@@ -306,6 +306,9 @@
 # limit of api_key seconds to live before expiration
 ;api_key_max_seconds_to_live = -1
 
+# Set to true to enable SigV4 authentication option for HTTP-based datasources
```
achatterjee-grafana, line 309: Period at the end.

achatterjee-grafana (commented):
I have added a few comments, with regards to tool-tip texts.

packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
```diff
@@ -0,0 +1,250 @@
...
+                value={dataSourceConfig.jsonData.assumeRoleArn || ''}
+                onChange={e => onJsonDataChange('assumeRoleArn', e.currentTarget.value)}
+              />
+            </div>
+          </div>
+          <div className="gf-form-inline">
+            <div className="gf-form">
+              <InlineFormLabel
+                className="width-14"
+                tooltip="If you are assuming a role in another account, that has been created with an external ID, specify the external ID here."
```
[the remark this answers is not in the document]
wbrowne: Thanks for the suggestion @achatterjee-grafana 😄  I think in this case, some of the meaning is lost when omitting the word `"assuming"`. As the role in which we're referring to is the one specified in the field above, then maybe the following is better:
`"If the role you are assuming is in another account, and it was created with an external ID, specify it here."`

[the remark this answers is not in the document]
aknuds1: Yeah, @achatterjee-grafana's suggestion drops important semantics. On re-reading this, it's probably better to write "that was created with an external ID".

pkg/models/sigv4.go
```diff
@@ -0,0 +1,100 @@
...
+	return m.Next.RoundTrip(req)
+}
+
+func (m *SigV4Middleware) signRequest(req *http.Request) (http.Header, error) {
+	signer, err := m.signer()
+	if err != nil {
+		return nil, err
+	}
+
+	return signer.Sign(req, nil, "grafana", m.Config.Region, time.Now())
```
marefr, line 64: @wbrowne I think we need to include the request body in signing, no?

wbrowne: Great catch! Went down a bit of a rabbit hole to figure out the right approach and I think it should be now in the right shape

pkg/models/sigv4.go
```diff
@@ -61,7 +63,19 @@ func (m *SigV4Middleware) signRequest(req *http.Request) (http.Header, error) {
...
 
-	return signer.Sign(req, nil, "grafana", m.Config.Region, time.Now())
+	if req.Body != nil {
+		// consume entire request body so that the signer can generate a hash from the contents
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			return nil, err
+		}
+		// re-wrap body with a io.ReadCloser for subsequent closure
+		req.Body = ioutil.NopCloser(bytes.NewReader(body))
```
aknuds1, line 73: What's the point of this? Doesn't `signer.Sign` already do this?

> // Sign will set the request's Body to be the `body` parameter passed in. If
> // the body is not already an io.ReadCloser, it will be wrapped within one

aknuds1 (commented):
See my question.

hugohaggmark (commented):
I agree with @dprokop and @aocenas, the FrontEnd code looks good!

pkg/models/sigv4.go
wbrowne: The point of this is before we were always calling `signer.Sign(req, nil <body>, ...)` so we would never pass the request body even if it was not nil (so we would never sign the body). With this change, in order to sign the body, we have to read the entire request body, and pass it in separately as an `io.ReadSeeker` so that in the SDK the following code will execute:

```go
var reader io.ReadCloser
if body != nil {
   var ok bool
   if reader, ok = body.(io.ReadCloser); !ok {
      reader = ioutil.NopCloser(body)
   }
}
r.Body = reader
```

aknuds1: That wasn't quite my question though. My question is if wrapping req.Body is necessary, considering the SDK method documentation says it does exactly the same.

wbrowne: If you're specifically referring to why wrapping it in a no-op closer, the only reason is that's the pattern that I've seen used (also by another sigV4 library which similarly wraps the AWS SDK) when you want to consume a body, but then wrap it up again

aknuds1: Yep that's what I'm referring to. What about the SDK documentation for the method saying it will do exactly that? 

> // Sign will set the request's Body to be the body parameter passed in. If
> // the body is not already an io.ReadCloser, it will be wrapped within one

Unless I misunderstand the above, you can drop line 73.

conf/defaults.ini
```diff
@@ -307,6 +307,9 @@ oauth_state_cookie_max_age = 600
 # limit of api_key seconds to live before expiration
 api_key_max_seconds_to_live = -1
 
+# Set to true to enable SigV4 authentication option for HTTP-based datasources
```
marefr, line 310: As discussed we may want to move this to another section, but lets consider that in follow up PR.

pkg/models/datasource_cache.go
```diff
@@ -177,6 +188,23 @@ func (ds *DataSource) GetHttpTransport() (*dataSourceTransport, error) {
 	return dsTransport, nil
 }
 
+func (ds *DataSource) sigV4Middleware(next http.RoundTripper) http.RoundTripper {
+	decrypted := ds.SecureJsonData.Decrypt()
```
marefr, line 192: This will use cached decrypted values if available, otherwise decrypt them and cache them.

```suggestion
	decrypted := ds.DecryptedValues()
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,78 @@ func TestDataSourceDecryptionCache(t *testing.T) {
 	})
 }
 
+func TestDataSourceSigV4Auth(t *testing.T) {
+	Convey("When caching a datasource proxy with middleware", t, func() {
```
marefr, line 295: We should convert these to go standard tests, but lets fix in separate PR.

pkg/models/sigv4.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return v4.NewSigner(stscreds.NewCredentials(s, m.Config.AssumeRoleARN)), nil
+	}
+
+	return v4.NewSigner(c), nil
+}
+
+func (m *SigV4Middleware) credentials() (*credentials.Credentials, error) {
```
marefr, line 97: We should try and extract this and share it between CloudWatch and SigV4, if possible. But lets consider in separate PR.

marefr (approved):
LGTM, but added a couple suggestions

docs/sources/administration/configuration.md
```diff
@@ -649,6 +649,10 @@ Administrators can increase this if they experience OAuth login state mismatch e
 
 Limit of API key seconds to live before expiration. Default is -1 (unlimited).
 
+### sigv4_auth_enabled
+
+Set to `true` to enable the AWS Signature Version 4 Authentication option for HTTP-based datasources. Default is `false`.
```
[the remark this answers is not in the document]
marefr: Hmm this suggested was added 2 times. Resolving this.

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,78 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+			setting.SigV4AuthEnabled = origEnabled
+		})
+
+		ds := DataSource{}
+
+		t, err := ds.GetHttpTransport()
+		So(err, ShouldBeNil)
+
+		Convey("Should not include sigV4 middleware if not configured in JsonData", func() {
+			_, ok := interface{}(t.next).(*http.Transport)
```
aknuds1, line 336:
```suggestion
			_, ok := t.next.(*http.Transport)
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,78 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+
+		ds := DataSource{
+			JsonData: json,
+		}
+
+		t, err := ds.GetHttpTransport()
+		So(err, ShouldBeNil)
+
+		Convey("Should not include sigV4 middleware if not configured in app config", func() {
+			_, ok := interface{}(t.next).(*http.Transport)
```
aknuds1, line 360:
```suggestion
			_, ok := t.next.(*http.Transport)
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,78 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+
+		ds := DataSource{
+			JsonData: json,
+		}
+
+		t, err := ds.GetHttpTransport()
+		So(err, ShouldBeNil)
+
+		Convey("SigV4 is in middleware chain if configured in JsonData", func() {
+			m1, ok := (t.next).(*SigV4Middleware)
```
aknuds1, line 314:
```suggestion
			m1, ok := t.next.(*SigV4Middleware)
```

pkg/models/datasource_cache_test.go
```diff
@@ -291,6 +291,78 @@ func TestDataSourceDecryptionCache(t *testing.T) {
...
+			So(ok, ShouldEqual, true)
+
+			_, ok = (m1.Next).(*http.Transport)
```
aknuds1, line 317:
```suggestion
			_, ok = m1.Next.(*http.Transport)
```

aknuds1 (approved):
LGTM!

PR #27554, opened 2020-09-11
Toolkit: clean node_modules/@grafana/data/node_modules in prepare

### LINKED ISSUE #26748: Plugin builds break whenever rxjs has a new release
When rxjs has a new release, plugins get unhappy ;(
https://github.com/ReactiveX/rxjs/releases

See also https://github.com/grafana/grafana/issues/26038

The workaround is to remove:
```
rm -rf node_modules/@grafana/data/node_modules/rxjs
```

but we should figure out something that will not happen with each new rxjs release

ryantxu:
Fixes #26748  (i think... without understanding the root cause)


torkelo (approved):
It feels like this will come back and bite us, but let's try it

PR #27555, opened 2020-09-11
add method to get usage report

DanCech:
This PR splits the usage report functionality into separate functions to generate the report and to submit it.


sakjur (approved):
LGTM.

To further improve from this point we might want to wrap and return the error to then let the error be handled one level up from `GetUsageReport` rather than both logging and returning it. I don't think that's necessary in this PR though, it's already an improvement in most aspects.

PR #27564, opened 2020-09-13
EventBus: Introduces new event bus with emitter backward compatible interface

torkelo:
Need a way for react panels to be able to:
* Emit events (think graph hover crosshair sync)
* Listen to events (think graph hover time) 
* Subscribe to annotations or exemplars

This PR: 
* New event bus/emitter interface and model that returns Unsubscribable / or Observable. 
* Event types are types that can be used for both subscribe and emit 
* Legacy emitter support (and the emitter is still used internally in the EventBusSrv) 

Things to figure out, we have 3 emitters 
* appEvents 
* dashboard.events
* panel.events

They are completely separate, so something emitted via appEvents is not available on the others and vice versa. 
So what to pass to react panels via PanelProps? panel.events are isolated from other panels so that can't be used for inter panel communication.

Continuation of #21010


public/app/features/dashboard/containers/__snapshots__/DashboardPage.test.tsx.snap
```diff
@@ -24,6 +24,26 @@ exports[`DashboardPage Dashboard init completed  Should render dashboard grid 1`
...
+          },
+        },
+=======
+        "eventBus": EventBusSrv {
+          "emitter": EventEmitter {
+            "_events": Object {},
+            "_eventsCount": 0,
+          },
+        },
+>>>>>>> EventBus & Emitter unification
```
torkelo, line 46: Need to fix this

public/app/plugins/panel/text/TextPanel.tsx
```diff
@@ -32,6 +35,16 @@ export class TextPanel extends PureComponent<Props, State> {
...
+  componentDidMount() {
+    this.eventSub = this.props.eventBus.$on(LegacyGraphHoverEvent, event => {
+      console.log('grap hover event', event);
+    });
+  }
+
+  componentWillUnmount() {
+    this.eventSub?.unsubscribe();
+  }
+
```
torkelo, line 47: Testing this from a react panel. Currently there is no automatic unsubscribe. For panels that is written as a functional component I think we can create a hook that automatically handled unsubscribe on unmount

packages/grafana-data/src/events/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Get observable of events
+   */
+  getStream<T extends BusEvent>(eventType: BusEventType<T>): Observable<T>;
+
+  /**
+   * Remove all event subscriptions
+   */
+  removeAllListeners(): void;
+}
```
torkelo, line 74: This is the main new interface, the $ sign is to not conflict with the old methods. The behavior of the handler is a bit different. The handler takes in the full event not just the payload. This will make it possible to subscribe to multiple events with the same handler (in the future).

torkelo: My thinking is that at some point in the future after a looong period of deprecation we can make emit/on functions equal to $emit/$on

packages/grafana-data/src/events/EventBus.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  constructor() {
+    this.emitter = new EventEmitter();
+  }
+
+  $emit<T extends BusEvent>(event: T): void {
+    this.emitter.emit(event.type, event);
+  }
+
+  $on<T extends BusEvent>(typeFilter: BusEventType<T>, handler: BusEventHandler<T>): Unsubscribable {
```
ryantxu, line 17: Can we avoid the $ conflict by just picking new names?  Perhaps:

$emit > broadcast, publish, push?
$on > listen, handle, when?

packages/grafana-data/src/types/legacyEvents.ts
```diff
@@ -0,0 +1,44 @@
...
+export type AlertPayload = [string, string?];
+export type AlertErrorPayload = [string, (string | Error)?];
+
+export const AppEvents = {
+  alertSuccess: eventFactory<AlertPayload>('alert-success'),
+  alertWarning: eventFactory<AlertPayload>('alert-warning'),
+  alertError: eventFactory<AlertErrorPayload>('alert-error'),
+};
+
+export const PanelEvents = {
```
ryantxu, line 16: Do we need to put this in `@grafana/data`?  Can the legacy things be internal packages?

torkelo: Some are used angular panel plugins

packages/grafana-data/src/events/EventBus.ts
torkelo: $emit -> publish
$on -> subscribe

@hugohaggmark thoughts?

hugohaggmark: @torkelo I like it a lot!! 🎉

packages/grafana-data/src/types/legacyEvents.ts
torkelo: and this is already in grafana/data, so would break plugins if we removed it

packages/grafana-data/src/events/EventBus.ts
```diff
@@ -0,0 +1,82 @@
...
+  off<T>(event: AppEvent<T> | string, handler: LegacyEventHandler<T>) {
+    if (typeof event === 'string') {
+      this.emitter.off(event, handler.wrapper);
+      return;
+    }
+
+    this.emitter.off(event.name, handler.wrapper);
+  }
+
+  removeAllListeners() {
```
mckn, line 79: Will we create a singleton of this class? Then this might be a bit dangerous to expose depending on how it is accessible.

packages/grafana-data/src/events/types.ts
mckn: Would it be possible to rename the $on and $emit to something else? I'm not a big fan of using the $ prefix since it is strongly connected with other libraries and in our case angularjs. When consuming this API it would be nice that it is clear that is has nothing to do with angularjs (if that makes sense?)

packages/grafana-data/src/events/EventBusGroup.ts
```diff
@@ -0,0 +1,67 @@
+// import { Unsubscribable, Subscription, Observable } from 'rxjs';
+// import { BusEvent, BusEventType, BusEventHandler, EventBus } from './types';
+
+// /**
+//  * Handles unsubscribing to all events subscribed through this group
+//  */
+// export class EventBusGroup implements EventBus {
```
mckn, line 7: hmm, what was the idea behind this one? Make it easier to unsubscribe to multiple subscriptions?

mckn (changes requested):
Overall I think it looks really good! I have added some small questions.

packages/grafana-data/src/events/EventBusGroup.ts
torkelo: yea, exactly. But think I will solve this some other way

packages/grafana-data/src/events/types.ts
torkelo: renamed to publish & subscribe

MGulraiz:
Hi @torkelo, I am trying to implement a shared tooltip functionality into my custom plugin. Could you please refer me to an example or documentation that uses this event bus?

mckn:
> Hi @torkelo, I am trying to implement a shared tooltip functionality into my custom plugin. Could you please refer me to an example or documentation that uses this event bus?

Hello, there will be an event bus passed to via props to your plugin. I believe the name is `appEvents`. This one can be used to both publish and subscribe to events. The same way as done in these tests: https://github.com/grafana/grafana/blob/dd3be4dc015d89eace541a9d388ab95d0752d0d1/packages/grafana-data/src/events/EventBus.test.ts#L38

torkelo:
on PanelProps we send an eventBus props  to add synced cursor features subscribe or publish DataHoverEvent events

https://github.com/search?q=repo%3Agrafana%2Fgrafana%20DataHoverEvent&type=code

PR #27565, opened 2020-09-13
Chore: Fixes strict typescript issues and counts remaining 

torkelo:
Even though strictNullChecks are done we have some remaining strict checks that are not on in core but in @grafan/data etc (they are all compiled with --strict true, where all strict checks are enabled). 

The remaining strict checks: 
- strictBindCallApply (8) 
- strictFunctionTypes (285 errors, can be tricky)
- strictPropertyInitialization (261 errors, relativly easy to fix)


hugohaggmark (approved):
LGTM!

PR #27572, opened 2020-09-14
RFC Graph NG

dprokop:
Replaces https://github.com/grafana/grafana/pull/27396

This PR is a result of experimentation with [uPlot](https://github.com/leeoniya/uPlot) as a potential replacement for jquery.flot.

It aims to allow Graph visualization composition rather than implementing every interaction in a single component as we currently try to do in the [react graph panel](https://github.com/grafana/grafana/blob/d42a7e827a013905187307dd8f9be71d954a028e/public/app/plugins/panel/graph2/GraphPanel.tsx)

Major components (naming is a subject of a change, mos def):

#### UPlotChart

Given a data frame(already aligned on the x-axis, currently on `Time` field only, using series to columns transformation) prepare [uPlot data model](https://github.com/leeoniya/uPlot/tree/master/docs#data-format), configure series, axes and scales based on field config (this is done on a very small scale now, needs way more work)

#### Plugins 

`UPlotChart` holds a map of plugins that uPlot needs to be initialized with. Via `PlotPluginsContext` I'm exposing an API for plugin registration. This API is available via `usePlotPluginContext` hook that exposes `registerPlugin` method - it has the same API as uPlot plugin(https://github.com/leeoniya/uPlot/blob/fe5fac73eaf1caf753a5eb9fea6b41cbc0a9bea4/dist/uPlot.d.ts#L199). Since uPlot plugin system allows hooking into uPlot lifecycle, plugins can manage their state on their own, not depending on some global state exposed from for instance UPlotChar. _This is an important decision I've made as I think it will simplify and make the architecture more modular_. I've started with exposing i.e. cursor position via context defined in UPlotChart, but the component quickly became massive again. I think the benefit can be observed in the sample plugins I've implemented that basically are a composition of the more primitive plugins like i.e. ClickPlugin, SelectionPlugin etc. Those primitives and additional plugins will handle the state independently from the UPlotChart. See code for more details.

#### Declarative API

First and foremost - axes, scales and series config will be driven primarily by the field configuration. 
Graphs functionalities (i.e. context menu, zoom) will be configured in a declarative way. I mean, if graph needs zoom functionality, you render ZoomPlugin as it's child
```
  <UPlotChart data={alignedData} timeRange={timeRange} width={width} height={height}>
      <ZoomPlugin
        onZoom={range => {
          onChangeTimeRange(range);
        }}
      />
   </UPlotChart>
```

under the hood, `ZoomPlugin` uses `SelectionPlugin` that is a uPlot plugin to selection hook. Using the same idea the `AnnotationsEditorPlugin` is implemented, using exactly the same plugin under the hood, but processing the selection lazily.

I've implemented plugins using render prop pattern, maybe it's worth exploring hooks approach.

To be solved:
1. Annotations interaction - how to use `setSelect` uPlot plugin hook and get access to the underlying event, to allow annotations creation only when CMD+drag is performed? Currently, there is no access to the mouse event that initiated the selection.
2. Zoom - minimal drag distance? Currently, zooming is happening on minimal mouse move
3. Streaming performance - since I'm reinitialising the uPlot on data frames change, the streaming performance may become poor with large dashboards. I haven't tested it yet, but we probably would like to use `setData`, `setSeries` etc methods for uPlot instead. This becomes problematic when number of series changes but that's only in edit mode I guess. Needs more thought and investigation
4. x-axis time formatting, axis width
5. Legend plugin somehow?
6. And many many more things :)

TODO:
- [x] Data retrieval in plugins (access to data frame and correct field within it)
- [ ] https://github.com/grafana/grafana/pull/27572#discussion_r488382540
- [x] Revert MutableDataFrame interface change https://github.com/grafana/grafana/pull/27572/files#diff-a50d3269cb61b09314ba1d22ba2fab39
- [ ] Legend plugin
  - [x] MVP(2e71401)
  - [x] Legend sizing
  - [ ] Toggling series
  - [ ] Table legend stats 
  - [ ] Series color control via legend icon (currently series color is configured via field settings)
- [ ] Data links support via context menu
- [ ] Bar chart
- [ ] Options and field config - organize options for better UX
- [ ] Annotations - editor + display
- [x] Show/hide grid
- [x] Time zone support on x-axis
- [x] Refactor contexts to a single PlotContext to simplify render path
- [x] Graph Panel layout API
- [ ] Color picker standard field config editor needs love
- [ ] Dark theme colors adjustments (text, axis grid etc)
- [ ] update uPlot to 1.2.x and update x-axis time formatting (ref https://github.com/leeoniya/uPlot/issues/303)
- [ ] Streaming - update graph data only when relative time range is applied


dprokop:
> Ref: Axes customization - is there a way to dynamically calculate axes width?

@ryantxu @torkelo since uPlot does not allow dynamic calculation of the axis width, I've added field config option to customize axis (`axis.width`)

torkelo:
Why do we need axis width option? We do not have that for the current graph panel/component

dprokop:
@torkelo 

>  since uPlot does not allow dynamic calculation of the axis width

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -0,0 +1,125 @@
...
+      plotInstance.setSize({
+        width: props.width,
+        height: props.height,
+      });
+    }
+  }, [props.width, props.height]);
+
+  return (
+    <PlotDataContext.Provider value={{ data: props.data }}>
+      <PlotPluginsContext.Provider value={{ registerPlugin }}>
```
torkelo, line 108: is it good to always create new objects for value here on every render, wont that trigger some update?

packages/grafana-ui/src/components/uPlot/plugins/CursorPlugin.tsx
```diff
@@ -0,0 +1,112 @@
...
+export const CursorPlugin: React.FC<CursorPluginProps> = ({ id, children, capture = 'mousemove', lock = false }) => {
+  const pluginId = `CursorPlugin:${id}`;
+  const plotCanvas = useRef<HTMLDivElement>(null);
+  const plotCanvasBBox = useRef<any>({ left: 0, top: 0, right: 0, bottom: 0, width: 0, height: 0 });
+  const pluginsApi = usePlotPluginContext();
+
+  // state exposed to the consumers, maybe better implement as CursorPlugin?
+  const [focusedSeriesIdx, setFocusedSeriesIdx] = useState<number | null>(null);
+  const [focusedPointIdx, setFocusedPointIdx] = useState<number | null>(null);
+  const [coords, setCoords] = useState<{ viewport: Coords; plotCanvas: Coords } | null>(null);
```
torkelo, line 43: wonder if it's not better to have a single state object here, react cannot batch state updates to useState updates unless they happen in a react synthetic event. So need to watch out for double/triple renders when using separate useState for every property.

public/app/plugins/panel/graph3/module.tsx
```diff
@@ -0,0 +1,177 @@
...
+      builder
+        .addBooleanSwitch({
+          path: 'line.show',
+          name: 'Show lines',
+          description: '',
+          defaultValue: true,
+        })
+        .addSelect({
+          path: 'line.width',
+          name: 'Line width',
```
torkelo, line 18: do we need the nesting of properties here? Flat properties are usually easier to work with

torkelo (commented):
Amazing work dominik! Took this for a spin, works really well. 

The default values for things like lines show seem off, they are disabled by default so enabling has no effect, have to enable and then disable for things to take effect.
 
Also time to create an issue for adding a slider field config editor! so we can use that for line width and area fill :)

packages/grafana-ui/src/components/uPlot/context.ts
```diff
@@ -0,0 +1,147 @@
...
+  /** Returns field by index */
+  getField: (idx: number) => Field;
+  /** Returns x-axis fields */
+  getXAxisFields: () => Field[];
+  /** Returns x-axis fields */
+  getYAxisFields: () => Field[];
+  /** Returns field value by field and value index */
+  getFieldValue: (fieldIdx: number, rowIdx: number) => any;
+  /** Returns field config by field index */
+  getFieldConfig: (fieldIdx: number) => FieldConfig;
```
ryantxu, line 59: I am confused by this API -- are the fields each thing returns different than what shows up in `data`?  

can `getField(2)` be different than `data.fields[2]`?

Should the X/Y getters return indexes?  or should we not have data and assert that everything is the same length?

public/app/plugins/panel/graph3/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { DataFrame, FieldType, getTimeField, sortDataFrame, transformDataFrame } from '@grafana/data';
+
+// very time oriented for now
+export const alignAndSortDataFramesByFieldName = (data: DataFrame[], fieldName: string) => {
+  // normalize time field names
+  // in each frame find first time field and rename it to unified name
```
ryantxu, line 6: :vomiting_face: mutates the potentially shared object

I think the better solution is to update `seriesToColumns` sol it can join by the first time field regardless of what it is named

public/app/plugins/panel/graph3/module.tsx
dprokop: yes, this is what I started with, but since there are plenty of "related properties I find it easier to have specific types in the interface

public/app/plugins/panel/graph3/utils.ts
dprokop: yes, I like this idea

public/app/plugins/panel/graph3/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  // Redraw as time passes
+  realTimeUpdates?: boolean;
+}
+
+export interface Options {
+  graph: GraphOptions;
+  legend: LegendOptions & GraphLegendEditorLegendOptions;
+  series: {
+    [alias: string]: SeriesOptions;
+  };
```
torkelo, line 22: what is this?

public/app/plugins/panel/graph3/module.tsx
```diff
@@ -0,0 +1,180 @@
...
+            { value: 'none', label: 'No tooltip' },
+          ],
+        },
+      })
+      .addBooleanSwitch({
+        path: 'graph.realTimeUpdates',
+        name: 'Real time updates',
+        description: 'continue to update the graph so the time axis matches the clock.',
+        defaultValue: false,
+      })
```
torkelo, line 152: do we need this? feels like we only need this for streaming data and if could detect that somehow

torkelo: and streaming data should have an update request time range

torkelo (commented):
Just some minor comments

packages/grafana-data/src/field/fieldOverrides.ts
```diff
@@ -459,3 +459,16 @@ export function applyRawFieldOverrides(data: DataFrame[]): DataFrame[] {
 
   return newData;
 }
+
+export const STANDARD_FIELD_OPTIONS = [
```
torkelo, line 463: nits: any reason for the upper case? Usually only do that for value constants.

torkelo (commented):
Looks really good! 

The only thing that currently blocks merging that I can find is that it adds the color standard option, which currently overrides thresholds for other panels and there is no way to reset/clear it. 

Screenshot from 2020-09-23 14-21-45

packages/grafana-data/src/field/fieldOverrides.ts
dprokop: well, isn't this a value constant? :)

dprokop:
> The only thing that currently blocks merging that I can find is that it adds the color standard option, which currently overrides thresholds for other panels and there is no way to reset/clear it. 

ah shit, this is problematic. Will try to figure this out

packages/grafana-data/src/field/fieldOverrides.ts
leeoniya: @torkelo probably means immutable values like numbers and strings.

dprokop: Removed this as it was hacking around standard color field config propert

dprokop:
@torkelo I have removed the color standard property editor for now. Graph has now series color config available via custom property - this will need to change when we properly support the color as standard field config.

public/app/plugins/panel/graph3/useMeasure.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const useMeasure = <E extends HTMLElement = HTMLElement>(): UseMeasureResult<E> => {
+  const [element, ref] = useState<E | null>(null);
+  const [rect, setRect] = useState<UseMeasureRect>(defaultState);
+
+  const observer = useMemo(
+    () =>
+      new (window as any).ResizeObserver((entries: any) => {
+        if (entries[0]) {
+          const { x, y, width, height, top, left, bottom, right } = entries[0].contentRect;
+          setRect({ x, y, width, height, top, left, bottom, right });
```
torkelo, line 31: Do we need this? the panel get's width & height from outside (that will change when window changes width).

torkelo (approved):

devenv/dev-dashboards/panel-graph/graph-ng.json
```diff
@@ -2248,5 +2248,6 @@
   "timezone": "Africa/Accra",
   "title": "Graph NG - tests",
```
torkelo, line 2249: with be good to follow same naming pattern "Panel Tests - Graph NG"

devenv/dev-dashboards/panel-graph/graph-ng.json
```diff
@@ -0,0 +1,2253 @@
...
+  },
+  "time": {
+    "from": "now-5m",
+    "to": "now"
+  },
+  "timepicker": {
+    "refresh_intervals": ["5s", "10s", "30s", "1m", "5m", "15m", "30m", "1h", "2h", "1d"]
+  },
+  "timezone": "Africa/Accra",
+  "title": "Graph NG - tests",
```
dprokop, line 2249:
```suggestion
  "title": "Panel Tests - Graph NG",
```

PR #27573, opened 2020-09-14
MutableDataFrame: Remove unique field name constraint and values field index and unused/seldom used stuff 

### LINKED ISSUE #24376: Transforms: MutableDataFrame and duplicate field name warnings
Both join (series to columns) transform and calculate field use MutableDataFrame that outputs a warning about duplicate fields. 

It's not super easy to remove this. We do build the fields in these transforms so need a MutableField, are we missing an abstraction or should we update MutableDataFrame?

torkelo:
There is much that feels a bit ad-hoc, hacky, edge case, or just undocumented about MutableDataFrame. And worry that many plugins are now starting to use it as the de facto way to build and interact with DataFrames 

This PR does not fix much but just tries to get a conversation going and highlight a few issues 

* MutableDataFrame.values is an object dictionary, used in from what I can see only from loki logs. This object dictionary causes duplicate field warnings in many transformations that use MutableDataFrame. Is there a better way to handle this use case of accessing fields by name? 
* Mixing DataFrame & DataFrameDTO, and Field & FieldDTO makes for subtle bugs. For example, we had a bug where addField does not actually add the but creates a new internal copy and copy non-DTO properties like display processor. Not addressed in this PR yet. 

What should our preferred way of building a DataFrame (from different structures) be? A DataFrame class (like MutabaleDataFrame), or toDateFrame? 

Fixes #24376


ryantxu (approved):
> What should our preferred way of building a DataFrame (from different structures) be? A DataFrame class (like MutabaleDataFrame), or toDateFrame?

MutabaleDataFrame should be used rarely, and only when you really need the semantics of appending rows.  In most cases it is pretty easy to just deal with a list of fields

PR #27577, opened 2020-09-14
Explore/Actions: Stop loadExploreDatasourcesAndSetDatasource from running twice

kaydelaney:
A change was [previously made](https://github.com/grafana/grafana/pull/26704) that would run queries if the `changeDatasource` action was dispatched with the `importQueries` option enabled. However, the `loadExploreDatasourcesAndSetDatasource` action ran queries after changing datasources with the above action and option, so queries were getting run twice on loading explore / browser refresh. This PR removes the dispatch of `runQueries` in `loadExploreDatasourcesAndSetDatasource`, so queries should only be run once now.


zoltanbedi (approved):
Yep, that's right. This looks good. 🍧

PR #27579, opened 2020-09-14
Chore: use visualization name in field config header

ryantxu:
In looking at the confusion around table docs, I think the section heading is not helping much.  We should use the panel name for the heading of "custom" config.  The "custom" name is useful for plugin developers, but not end-users

| Before | After |
| --- | --- |
|  |  |


torkelo (approved):

torkelo:
We used to use panel name here, can't remember when/why we did go back to custom

oddlittlebird:
> We used to use panel name here, can't remember when/why we did go back to custom

It was always "custom." I pointed out it was weird in pre-release, but was told that it needed to be that way because other custom configurations might go there.

PR #27580, opened 2020-09-14
Chore: eslint plugin react hooks fix in jeager

### LINKED ISSUE #27391: Chore: Enable and fix eslint-plugin-react-hooks warning and errors in the monorepo
After https://github.com/grafana/grafana/pull/26094 merged the we can start to work on enabling eslint-plugin-react-hooks for the packages and the main app.
As of today (2020.09.04.) there are ✖ 169 problems (86 errors, 83 warnings) in the whole monorepo after enabling the rules.
The plan is to fix these warnings continuously.

- [x] grafana-toolkit 26problems total -  React Hook "useSpinner" cannot be called at the top level. React Hooks must be called in a React function component or a custom React Hook function - This error should be silenced by leaving an override in the root eslintrc
- [x] grafana-ui has 58 problems total (I think this should be separated to storybook errors and component errors) Fixing the component errors we must make sure that we can not create any regressions.
  - [x] storybook errors
  - [x] component errors
  - [x] ~uPlot - 25 warnings.~ added to the eslintignore 
- [x] jaeger-ui-components 4 problems total
- [ ] public folder 81 problems total This is where things can get south real quick so we should focus not to create any regressions here (it is probably a good idea to divy this up to smaller chunks and not to do the whole in one pr)

To see the errors simply remove the overrides property in the eslintrc file and run yarn lint.

zoltanbedi:

This pr changes eslint to use eslintignore instead of gitignore so enterprise components can be checked as well.
Add override config to grafana-toolkit so the rule is turned off there.
Fix issues in jaeger-ui-components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "e2e:debug": "./e2e/start-and-run-suite debug",
     "e2e:dev": "./e2e/start-and-run-suite dev",
     "jest": "jest --notify --watch",
-    "lint": "eslint . --ext .js,.tsx,.ts --cache --ignore-path .gitignore --ignore-pattern devenv",
+    "lint": "eslint . --ext .js,.tsx,.ts --cache",
     "jest-ci": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit -w ${TEST_MAX_WORKERS:-100%}",
     "lint:fix": "yarn lint --fix",
     "packages:build": "lerna run clean && lerna run build --ignore @grafana-plugins/input-datasource",
--- a/packages/grafana-toolkit/.eslintrc
+++ b/packages/grafana-toolkit/.eslintrc
@@ -1,5 +1,14 @@
 {
   "rules": {
     "no-restricted-imports": ["error", { "patterns": ["@grafana/runtime"] }]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["./**/*.{ts,tsx}"],
+      "rules": {
+        "react-hooks/rules-of-hooks": "off",
+        "react-hooks/exhaustive-deps": "off"
+      }
+    }
+  ]
 }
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "e2e:debug": "./e2e/start-and-run-suite debug",
     "e2e:dev": "./e2e/start-and-run-suite dev",
     "jest": "jest --notify --watch",
-    "lint": "eslint . --ext .js,.tsx,.ts --cache --ignore-path .gitignore --ignore-pattern devenv",
```
mckn, line 16: Will the paths in the `.eslintignore` add anything more then whats already in the `.gitignore`?

packages/grafana-toolkit/.eslintrc
```diff
@@ -1,5 +1,14 @@
 {
   "rules": {
     "no-restricted-imports": ["error", { "patterns": ["@grafana/runtime"] }]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["./**/*.{ts,tsx}"],
+      "rules": {
+        "react-hooks/rules-of-hooks": "off",
```
mckn, line 9: Do we have any issue on fixing the code so we can turn these rules on?

package.json
zoltanbedi: Yeah it sounded good to use the gitignore file but @Clarity-89 raised a concern that because of this the enterprise code is not detected. Because it is added to the gitignore.

packages/grafana-toolkit/.eslintrc
zoltanbedi: In grafana toolkit we don't really have react components. Do we? Also it is needed to turn this off because of the useSpinner function that it found the rule.

mckn (approved):
A couple of questions but overall I think this works as expected! Great work! 🎉

packages/grafana-toolkit/.eslintrc
mckn: Ah, awesome! I read `grafana-ui`, mybad 🙈

package.json
mckn: Aha, good catch! 👍

PR #27583, opened 2020-09-14
Live: broadcast events when dashboard is saved

ryantxu:
When the "live" feature flag is enabled and a dashboard is saved, an event is broadcast to everyone looking at that dashboard.

If the viewer is in edit mode, or has local edits, we show a popup asking what to do.  Otherwise the dashboard is updated automatically.

dash


public/app/features/live/pageTracker.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.subject.next(evt);
+    }
+  }
+
+  watchLocationHref(interval = 500) {
+    window.setInterval(this.watchLocation, interval);
+  }
+
+  private watchLocation = () => {
+    this.check(document.location.href);
```
torkelo, line 40: this feels way too hacky :) needs to listen to either the angular router events, or redux path changes. Could be done in bridge_srv also .

packages/grafana-runtime/src/services/live.ts
```diff
@@ -37,13 +37,18 @@ export interface GrafanaLiveSrv {
...
 
   /**
    * Subscribe to activity on a given channel
    */
   getChannelStream<T>(channel: string): Observable<T>;
 
+  /**
+   * Unsubscribe all local references to a channel
+   */
+  closeChannelStream(channel: string): void;
```
torkelo, line 50: is this needed? can't unsubscribing to the initChannel Observable do this?

pkg/services/live/events.go
```diff
@@ -0,0 +1,18 @@
+package live
+
+// DashboardEventActionSaved action when dashbord is saved
+var DashboardEventActionSaved = "saved"
+
+// DashboardEventActionEditing action when dashbord is edited (sent from client)
+var DashboardEventActionEditing = "editing"
+
+// DashboardEventActionDeleted action when dashbord is deleted
+var DashboardEventActionDeleted = "deleted"
```
torkelo, line 10: for enums we do something like this usually https://github.com/grafana/grafana/blob/master/pkg/models/org_user.go#L18

torkelo (commented):
Left some comments, but before I read your description in detail. so missed your own comment on the router thing :) 

> this subscribes to a channel for each dashboard -- this would let use out-of-the-box centrifuge presence features for dashboards. Alternatively we could have a single channel and manage the notifications internally (not too hard and may be better) 

Any pros / cons? I assume a centrifuge channel is just an internal thing for routing events? (so not a new websocket connection). And don't we have to use that in order route events efficiently? (So we do not need implement our own event routing so only browsers who are interested in specific events get's them). For example we only want browsers who have a specific dashboard open to get events related to it, not get events for all open dashboards. So given that I think we need to use a channel per dashboard right?

bergquist:
:man_dancing:

packages/grafana-runtime/src/services/live.ts
ryantxu: removed

public/app/features/live/pageTracker.ts
ryantxu: yup  -- I replaced it with the `initDashboard` state

ryantxu:
> Any pros / cons? 

Yes -- i went in circles on this and yes one channel for each dashboard is the right approach.  


@torkelo -- I made a bunch of improvements since you looked, I would love feedback :)

bergquist:
I certainly find this an interesting feature. Hard to say exactly what the default behavior should be for the viewer who gets the new version pushed.

torkelo:
Sorry for no follow-up review, 7.2 release and fixes have been pressing. And after that, the annotation/exemplar architecture is higher prio

pkg/services/live/features/testdata.go
```diff
@@ -0,0 +1,124 @@
...
+	return centrifuge.ChannelOptions{}
+}
+
+// OnSubscribe for now allows anyone to subscribe to any dashboard
+func (g *testdataRunner) OnSubscribe(c *centrifuge.Client, e centrifuge.SubscribeEvent) error {
+	if !g.running {
+		g.running = true
+
+		// Run in the background
+		go g.runRandomCSV()
```
ryantxu, line 73: NOTE:  this now starts a random walk on the *first* subscribe.  The stuff in master starts it on startup.

We still need to add hooks to stop it when everyone has unsubscribed, but that will need to come in a follow on issue

packages/grafana-data/src/types/live.ts
```diff
@@ -85,28 +85,30 @@ export interface LiveChannelStatus {
...
 /**
  * @experimental
  */
-export interface LiveChannelPresense {
+export interface LiveChannelPresenseStatus {
   users: any;
 }
 
-export interface LiveChannelMessage<TMessage = any> {
-  type: 'status' | 'message' | 'join' | 'leave';
```
ryantxu, line 108: @hugohaggmark -- I switched from having a type flag to a limited set of properties that may exist.  This works better for typing, and the check for `if (event.status) {` seems more clear as `if (event.type === 'status') {`

pkg/models/live.go
```diff
@@ -0,0 +1,30 @@
+package models
+
+import "github.com/centrifugal/centrifuge"
+
+// ChannelPublisher writes data into a channel
+type ChannelPublisher func(channel string, data []byte) error
```
ryantxu, line 6: @kaydelaney -- this was my alternative to what you did with an async thread on HTTPServer

I don't think we should expose this on HTTPServer, but we should consider exposing this as as a go chan

packages/grafana-data/src/types/live.ts
hugohaggmark: Hmm, you can easily create TypeScript guard functions for that as in isStatusEvent, isLeaveEvent etc. WDYT?

ryantxu: But then you still have to cast message to the types value - or is there a typescript magic way to do that?

The semantic difference is this would now support multiple message types in one event (not great -- bet better than lots of casting)

hugohaggmark: No need for cast when you use a type guard as TypeScript then knows what type it is then, and yes it's magic :)
https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

pkg/services/live/features/dashboard.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/centrifugal/centrifuge"
+	"github.com/grafana/grafana/pkg/models"
+)
+
+// DashboardEvent events related to dashboards
+type dashboardEvent struct {
+	UID       string `json:"uid"`
+	Action    string `json:"action"` // saved, editing
+	UserID    int64  `json:"userId,omitempty"`
```
torkelo, line 14: Working with event systems before, it's always nice if the events are information-rich and more usable standalone.

It would mean including more dashboard & user info here like dashboard name, (maybe a user object, that includes id, login, email, avatarUrl) makes consuming and using these events much easier without requiring further API calls.

ryantxu: We could add the whole json model... But I think we would still need to do a full refresh in the current system regardless.  If only dashboard were an observable 🙂

torkelo: I was more thinking of other uses for this event (like an admin viewing a list of activity), and for the modal, we might want to show the user name & profile image of the user who updated the same dashboard you have. 

So ash much meta info about the entities in the event is usually good. Could also be used by a search or cache subsystem for example to update index state and for that we do not want to issue further db/api calls.

ryantxu: I'm happy to add more info now -- but since this is an internal event, I think we should wait and add more info when we have a real use for it.

ryantxu: > I was more thinking of other uses for this event 

That is info that should be attached on the 'presense' side -- that still needs to be explored some more, but I fear this PR is getting too big as is :)

public/app/features/live/dashboard/dashboardWatcher.ts
```diff
@@ -0,0 +1,166 @@
...
+    next: (event: LiveChannelEvent<DashboardEvent>) => {
+      // Send the editing state when connection starts
+      if (event.status && this.editing && event.status.state === LiveChannelConnectionState.Connected) {
+        this.sendEditingState();
+      }
+
+      if (event.message) {
+        if (event.message.sessionId === sessionId) {
+          return; // skip internal messages
+        }
```
torkelo, line 86: is there away to filter this out in liveSrv?

public/app/features/live/dashboard/dashboardWatcher.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+}
+
+export const dashboardWatcher = new DashboardWatcher();
+
+export function getDashboardChannelsFeature(): CoreGrafanaLiveFeature {
+  const dashboardConfig: LiveChannelConfig = {
+    path: '${uid}',
+    description: 'Dashboard change events',
+    variables: [{ value: 'uid', label: '${uid}', description: 'unique id for a dashboard' }],
```
torkelo, line 148: how are dashboardConfig.variables used?

public/app/features/live/dashboard/DashboardChangedModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  continueEditing: ActionInfo = {
+    label: 'Continue editing',
+    description:
+      'Keep your local changes and continue editing.  Note: when you save, this will overwrite the most recent chages',
+    action: () => {
+      this.onDismiss();
+    },
+  };
+
+  acceptDelte: ActionInfo = {
```
torkelo, line 43:
```suggestion
  acceptDelete: ActionInfo = {
```

public/app/features/live/dashboard/DashboardChangedModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }
+    `;
+
+    const isDelete = event?.action === DashboardEventAction.Deleted;
+
+    const options = isDelete ? [this.continueEditing, this.acceptDelte] : [this.continueEditing, this.discardAndReload];
+
+    return (
+      <Modal
+        isOpen={!!!dismiss}
```
torkelo, line 76: is this not the same as !dismiss ?

public/app/features/live/dashboard/DashboardChangedModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const radioClass = css`
+      cursor: pointer;
+      width: 100%;
+      padding: 20px;
+      background: ${config.theme.colors.dropdownBg};
+
+      &:hover {
+        background: ${config.theme.colors.formCheckboxBgCheckedHover};
+      }
+    `;
```
torkelo, line 68: this feels like a new design system component. Not sure what to do for this PR if there is no current component that fits. 

For a new component & we should use useTheme & useStyles

torkelo (commented):
Nice! left a few nits. But yea the typing looks a bit messy.

public/app/features/live/dashboard/dashboardWatcher.ts
ryantxu: idea here is that this is mostly documentation for what variables exist and what they need to be.  Looking now, we should likely remove this until it has something concrete, but I want to be careful that the path based routing has *some* way to be more constrained

public/app/features/live/dashboard/DashboardChangedModal.tsx
```diff
@@ -104,3 +96,17 @@ export class DashboardChangedModal extends PureComponent<Props, State> {
     );
   }
 }
+
+const getStyles = (theme: GrafanaTheme) => {
```
torkelo, line 100: use stylesFactory here so we only build the styles once

packages/grafana-data/src/types/live.ts
ryantxu: Updated with a new type structure I think you will like ;)  and OMG that is magic.  I feared you would have to cast after the check

public/app/features/live/dashboard/DashboardChangedModal.tsx
ryantxu: fixed -- looks like `!` and `!!!` are identical for a boolean value, but different if the value can be undefined of falsey

pkg/services/live/features/testdata.go
kaydelaney: Is this threadsafe? For example if nothing is subscribed to yet, but then two or more clients subscribe at around the same time is there any guarantee that multiple threads running `runRandomCSV` won't be started?

ryantxu: :crossed_fingers: I think so -- only a single handler will be initialized once, and should be blocked by this init logic:
https://github.com/grafana/grafana/pull/27583/files#diff-a2dd8b619e1ed03a64a4ad4cc0cc8e04R227

pkg/services/live/live.go
```diff
@@ -171,16 +219,71 @@ func InitalizeBroker() (*GrafanaLive, error) {
...
 	}
-	return b, nil
+	return glive, nil
 }
 
-// Publish sends the data to the channel
-func (b *GrafanaLive) Publish(channel string, data []byte) bool {
-	_, err := b.node.Publish(channel, data)
+// GetChannelHandler gives threadsafe access to the channel
+func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler, error) {
```
ryantxu, line 226: @marefr -- can you audit this function?  It is called frequently and potentially from multiple threads.

public/app/features/live/dashboard/DashboardChangedModal.tsx
hugohaggmark: You could use !Boolean(dismiss)

public/app/features/live/live.ts
```diff
@@ -12,6 +12,18 @@ import {
...
 import { registerLiveFeatures } from './features';
 
+export const sessionId =
+  (window as any)?.grafanaBootData?.user?.id +
+  '/' +
+  Date.now().toString(16) +
+  '/' +
+  // Math.random()
+  //   .toString(36)
+  //   .substring(2, 15) +
```
kaydelaney, line 22:
```suggestion
```

public/app/features/live/live.ts
```diff
@@ -94,7 +109,7 @@ export class CentrifugeSrv implements GrafanaLiveSrv {
     };
     this.open.set(id, channel);
 
-    // Initalize the channel in the bacground
+    // Initialize the channel in the bacground
```
kaydelaney, line 112:
```suggestion
    // Initialize the channel in the background
```

public/app/features/live/live.ts
```diff
@@ -113,13 +128,13 @@ export class CentrifugeSrv implements GrafanaLiveSrv {
     if (!config) {
       throw new Error('unknown path: ' + channel.path);
     }
+    if (config.canPublish && config.canPublish()) {
```
kaydelaney, line 131:
```suggestion
    if (config.canPublish?.()) {
```

kaydelaney:
Looking good! Just left some small comments.
One last thing: Is "Presen**s**e" intentional or is that a typo that should be "Presen**c**e"

ryantxu:
> Is "Presense" intentional or is that a typo that should be changed to "Presence"

dooh -- just dyslexia and poor spelling.  Thanks!

torkelo (approved):
Yay, so awesome to get to nice things like this 🚀 
tenor

PR #27584, opened 2020-09-15
Transformations: Use Observable

### LINKED ISSUE #27161: Transformations:  Refactor interface to use observable as return value
Current:

```ts
/**
 * Function that transform data frames (AKA transformer)
 */
export type DataTransformer = (data: DataFrame[]) => DataFrame[];

export interface DataTransformerInfo<TOptions = any> extends RegistryItemWithOptions {
  /**
   * Function that configures transformation and returns a transformer
   * @param options
   */
  transformer: (options: TOptions) => DataTransformer;
}
```

To something like 

```ts
/**
 * Function that transform data frames (AKA transformer)
 */
export interface DataTransformerInfo<TOptions = any> extends RegistryItemWithOptions {
  /**
   * Function that configures transformation and returns a transformer
   * @param options
   */
  transformer: (options: TOptions, data: DataFrame[]) => Observable<DataFrame[]>;
}
```

This would allow for async transformations that can do HTTP calls to do id lookups and translations.

hugohaggmark:
I think I got everything working as before, did a small change to the debug output. Now the output side will not show the resulting transformation from all transform but only the resulting transform from the debugged transform.

~~Also, I haven't started with any tests so those need a lot of work.~~ 

Fixes #27161

This also introduces a small!? change in Explore's data processing pipeline, please have a look @grafana/observability-squad


packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,62 @@
...
-        if (!series.meta) {
-          series.meta = {};
-        }
-        if (!series.meta.transformations) {
-          series.meta.transformations = [info.id];
-        } else {
-          series.meta.transformations = [...series.meta.transformations, info.id];
-        }
+export function transformDataFrame(options: DataTransformerConfig[], data: DataFrame[]): Observable<DataFrame[]> {
+  return of(of(data)).pipe(
```
ryantxu, line 49: What is `of(of(data))`?

public/app/features/dashboard/components/TransformationsEditor/TransformationEditor.tsx
```diff
@@ -1,29 +1,73 @@
...
+  onChange,
+}: TransformationEditorProps) => {
+  const styles = useStyles(getStyles);
+  const [input, setInput] = useState<DataFrame[]>([]);
+  const [output, setOutput] = useState<DataFrame[]>([]);
+  const config = useMemo(() => configs[index], [configs]);
+
+  useEffect(() => {
+    const inputTransforms = configs.slice(0, index).map(t => t.transformation);
+    const outputTransforms = configs.slice(index, index + 1).map(t => t.transformation);
```
hugohaggmark, line 40: Previously the output was defined as `configs.slice(index).map(t => t.transformation)`, ie include all transformations starting with the one being debugged. But I changed this because I think it makes more sense to just show the output from the debugged transformation. Easy to change back if I'm missunderstanding this :)

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: Good question, it's a stream of streams or Observable of Observable or a Higher Order Observable. Let me know if it's unclear.

hugohaggmark: An easier example here: https://stackblitz.com/edit/pg2o7z?devtoolsheight=33&file=index.ts

packages/grafana-data/src/types/transformations.ts
```diff
@@ -1,17 +1,17 @@
...
  */
-export type DataTransformer = (data: DataFrame[]) => DataFrame[];
-
 export interface DataTransformerInfo<TOptions = any> extends RegistryItemWithOptions {
   /**
    * Function that configures transformation and returns a transformer
    * @param options
    */
-  transformer: (options: TOptions) => DataTransformer;
+  transformer: (options: TOptions, data: DataFrame[]) => Observable<DataFrame[]>;
```
ryantxu, line 14: is there any value in the independent `DataTransformer` interface?  if not, this should be renamed so that it is the action `transform` rather than the thing that will transform

packages/grafana-data/src/transformations/transformDataFrame.ts
ryantxu: :exploding_head: I'll believe you :)  

just spent 10mins in docs and yes -- i think i get it!

hugohaggmark: I struggled to create an easier RxJs implementation of transformDataFrame, but it has a unique set of challenges:

- Each output from the stream uses the previous output as it's input
- We need to iterate over all supplied configs
- We're only interested in the last result

I'll have Another look and see if I can make it more readable

packages/grafana-data/src/types/transformations.ts
hugohaggmark: Sorry, not sure I follow

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: @ryantxu I made the `transformDataFrame` more readable, let me know if it's better, thank you.

packages/grafana-data/src/types/transformations.ts
ryantxu: Before `transformer(options)` returned a function (DataTransformer) that could process DataFrame[]'s

Now the registry item holds the function directly and it takes data as an input -- i think that is fine, but should be named a verb rather than a noun.

"transformer" is the name of a thing... a thing that transforms!
"transform" is an action on data

previously "transformer" returned a function, now it actually processes the data, so i think

packages/grafana-data/src/transformations/transformDataFrame.ts
ryantxu: yes better -- thanks

packages/grafana-data/src/types/transformations.ts
hugohaggmark: Gotcha and I agree that makes more sense.

torkelo (commented):
This looks really good. I do think we will have use of sync versions of merge/join transform but maybe we can add that when we have that clear need.

public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -194,53 +189,12 @@ export class TransformationsEditor extends React.PureComponent<TransformationsEd
...
-                      description={transformationUI.description}
-                    />
-                  );
-                })}
+                <TransformationOperationRows
+                  configs={transformations}
+                  data={data}
+                  onRemove={this.onTransformationRemove}
+                  onChange={this.onTransformationChange}
+                />
```
dprokop, line 197: Very nice!

public/app/types/explore.ts
```diff
@@ -233,3 +233,13 @@ export type RichHistoryQuery = {
...
+
+export interface DecoratedPanelData extends PanelData {
+  graphFrames: DataFrame[];
+  tableFrames: DataFrame[];
+  logsFrames: DataFrame[];
+  traceFrames: DataFrame[];
+  graphResult: GraphSeriesXY[] | null;
+  tableResult: DataFrame | null;
+  logsResult: LogsModel | null;
+}
```
dprokop, line 245: I'm not sure about the **name** of the interface. I would expect a type named `DecoratedPanelData` to be a generic one, something like this:

```
type DecoratedPanelData<T extends {}> = PanelData & T;
```

And in this case I guess the better name would just be `ExplotePanelData` (that  could be a `DecoratedPanelData<{...explore props}>`)

dprokop (commented):
I went through the code @hugohaggmark, amazing work! I've also tested it and it works great. I've left some minor nits

public/app/features/dashboard/components/TransformationsEditor/TransformationEditor.tsx
dprokop: Yeah, as long as the input data for the "output" transformation is the result of preceding transformations, which is super elegantly done here https://github.com/grafana/grafana/blob/4c2df34de8fd6f9bbcaa678e5b8ba5d4d0892c49/public/app/features/dashboard/components/TransformationsEditor/TransformationEditor.tsx#L42, this is fine. 

The output should basically show the result of **all preceding transformations & the debugged one**.

hugohaggmark:
@ryantxu I merged with master and got some new changes to make, can you check that 04727ca1bbecb876c16ad45a5299fb1c5798d9a5 doesn't destroy any annotations or point me to which datasource that uses this new interface so I can test it out myself. Thank you 🙏

public/app/types/explore.ts
hugohaggmark: Great finding, I like that name better!

public/app/features/dashboard/components/TransformationsEditor/TransformationEditor.tsx
hugohaggmark: Perfect, thank you

packages/grafana-data/src/transformations/transformers/calculateField.ts
```diff
@@ -68,56 +70,59 @@ export const calculateFieldTransformer: DataTransformerInfo<CalculateFieldTransf
...
-      if (!values) {
-        return frame;
-      }
+  transformer: (options, frames) => {
+    const source =
+      options && options.timeSeries !== false ? ensureColumnsTransformer.transformer(null, frames) : of(frames);
+
+    return source.pipe(
+      map(data => {
+        const mode = options.mode ?? CalculateFieldMode.ReduceRow;
```
Elfo404, line 79: just a question/nit about types:
isn't `options.mode` already of type `CalculateFieldMode` (non-optional)?

if it is then maybe we can omit all checks on `undefined` for `creator` since `options.mode` con be either (and only) `ReduceRow` or `BinaryOperation`:


```ts
let creator: ValuesCreator;

if (options.mode === CalculateFieldMode.ReduceRow) {
  creator = getReduceRowCreator(defaults(options.reduce, defaultReduceOptions), data);
} else {
  creator = getBinaryCreator(defaults(options.reduce, defaultBinaryOptions), data);
}

return data.map(frame => {
  // delegate field creation to the specific function
  const values = creator(frame);
// [...]
```

packages/grafana-data/src/transformations/transformers/filter.ts
```diff
@@ -72,30 +72,28 @@ export const filterFramesTransformer: DataTransformerInfo<FilterOptions> = {
...
-      for (const series of data) {
-        if (exclude) {
-          if (exclude(series)) {
-            continue;
-          }
-          if (!include) {
-            processed.push(series);
-          }
+    const processed: DataFrame[] = [];
+    for (const series of data) {
```
Elfo404, line 84: maybe this can be simplified?

```ts
for (const series of data) {
  if (exclude?.(series)) {
      continue;
  }

  if ((exclude && !include) || include?.(series)) {
    processed.push(series);
  }
}
```

Elfo404 (commented):
I just left a couple of nits as I don't feel comfortable really reviewing this, but for sure it gave me more context on transformations, so thank you 👍

packages/grafana-data/src/transformations/transformers/calculateField.ts
hugohaggmark: Thanks for feedback but I actually don't know because I didn't write this particular transformation but maybe @ryantxu can answer?

packages/grafana-data/src/transformations/transformers/filter.ts
hugohaggmark: Just a quick feedback, you can use the suggestion button (image) to make suggestions to code and then get credit as a co-author :)

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,71 @@
...
+        const curr = prev.pipe(
+          mergeMap(d => {
+            return getTransform(config, of(d));
+          })
+        );
+        observables.push(curr);
+      }
+
+      return observables[observables.length - 1];
+    })
```
mckn, line 69:
```suggestion
 const operators: Array<OperatorFunction<DataFrame[], DataFrame[]>> = mapConfigToOperators(options);
 const observable = of(data);

 return observable.pipe.apply(observable, operators); // we might need to run each of this befor applying them. It is just pseudo code.
```

hmm, I have not tested this code but what I am thinking is that each transform should just return an `OperatorFunction<DataFrame[], DataFrame[]>`. Then we create a stream out of the incoming data and pipe that stream through all the operators and return the result.

`mapConfigToOperators` should probably wrap each operator in another operator function that applies the meta data.

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,71 @@
...
+const getTransform = (config: DataTransformerConfig, data: Observable<DataFrame[]>): Observable<DataFrame[]> => {
+  const info = standardTransformersRegistry.get(config.id);
+
+  if (!info) {
+    return data;
+  }
+
+  return data.pipe(
+    mergeMap(before => {
+      const defaultOptions = info.transformation.defaultOptions ?? {};
```
mckn, line 16: This could probably be its own hoc-operator that takes in a configuration and an operator and returns an operator.

packages/grafana-data/src/types/transformations.ts
mckn, line 14: Instead of returning an `Observable<DataFrame[]>` I was thinking that we could return an `OperatorFunction<DataFrame[], DataFrame[]>`. This would convert each transformer into an operator (which basically is what they are) that we can apply on the Observable<DataFrame[]> stream.

I am happy to help out to see if we can do a small POC on this.

mckn (changes requested):
I think you have done a great job! It seems to work as expected. How ever I have added a couple of ideas on how to improve the code a bit and make it a little bit cleaner.

One of the things that I still having a hard time understanding is how we could turn the `DataFrame[]` into something that could be streamed in smaller chunks since it is basically a batch of data. But I guess that is another discussion not covered by this PR.

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: Not sure I follow, could you please evolve that?

packages/grafana-data/src/types/transformations.ts
hugohaggmark: This is a great suggestion, my first thought was to use `MonoTypeOperatorFunction` as it's the same input and output but then I thought that one of the optimizations with this PR was to get rid of the returning the a function in every transformer because of performance.

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: I really like this approach, very clever using the apply function. Never thought of using that before, great ⭐

hugohaggmark:
Don't think you need to hold the PR @mckn as no one has approved yet :)

packages/grafana-data/src/types/transformations.ts
hugohaggmark: What is `MonoTypeOperatorFunction` or `OperatorFunction`? (basically a function that returns a function with an Observable (source))
https://rxjs-dev.firebaseapp.com/api/index/interface/MonoTypeOperatorFunction
https://rxjs-dev.firebaseapp.com/api/index/interface/OperatorFunction

I guess that the benefits of using `MonoTypeOperatorFunction` or `OperatorFunction` as described by @mckn here compared to reintroducing returning a function in each transformer is something that we need to discuss with @ryantxu @dprokop @torkelo. WDYT? I think using `MonoTypeOperatorFunction` or `OperatorFunction` is much more the RxJs way.

mckn: Another benefit of doing it that way is that we can basically create TransformationEditorUI's for what ever RxJS operator we want to support, then our transformation would just need to be responsible for mapping the configuration from the editor to the RxJS operator.

Scenarios where I think this could be useful is in all kind of filtering, groupBy, merge and probably more scenarios.

A drawback is that we will have a big dependency on RxJS which will make it harder to replace in the future (if we would like to do that).

mckn:
> Don't think you need to hold the PR @mckn as no one has approved yet :)

Ah, sorry 🙈  I added the review to "mark" it as reviewed and removed from my list of things to review. If that makes sense? Maybe there is a better way?

hugohaggmark:
> > Don't think you need to hold the PR @mckn as no one has approved yet :)
> 
> Ah, sorry 🙈 I added the review to "mark" it as reviewed and removed from my list of things to review. If that makes sense? Maybe there is a better way?

No worries, can't you just leave a comment as a Review or does that not work in your flow?

mckn:
> > > Don't think you need to hold the PR @mckn as no one has approved yet :)
> > 
> > 
> > Ah, sorry 🙈 I added the review to "mark" it as reviewed and removed from my list of things to review. If that makes sense? Maybe there is a better way?
> 
> No worries, can't you just leave a comment as a Review or does that not work in your flow?

hmm, I need to try that :)

mckn:
> I think you have done a great job! It seems to work as expected. How ever I have added a couple of ideas on how to improve the code a bit and make it a little bit cleaner.
> 
> One of the things that I still having a hard time understanding is how we could turn the `DataFrame[]` into something that could be streamed in smaller chunks since it is basically a batch of data. But I guess that is another discussion not covered by this PR.

Have you thought anything on this @ryantxu @torkelo ?

ryantxu:
@hugohaggmark -- big picture this looks great.  My rxjs foo is so far from qualified to help much though


> One of the things that I still having a hard time understanding is how we could turn the DataFrame[] into something that could be streamed in smaller chunks since it is basically a batch of data. But I guess that is another discussion not covered by this PR.

Yes that is for sure a different discussion ;)    I don't think transformations need to think much about deltas, but DataSource may

ryantxu:
Once this can merge with master, I can test with a bunch of existing dashboards and make sure it still works.

hugohaggmark:
> Once this can merge with master, I can test with a bunch of existing dashboards and make sure it still works.

As I mentioned earlier in this PR I changed the annotations API so this https://github.com/grafana/grafana/blob/bd9758bb4a3e49cf84df721070b5dd6ae889b18f/packages/grafana-data/src/types/annotations.ts#L84 will break any plugin already using this new API

ryantxu:
That change will be ok

packages/grafana-data/src/types/transformations.ts
ryantxu: The OperatorFunciton seems like a great idea!  if we can make this fit smoothly into standard rxjs patterns that is a big help

ryantxu (approved):
My manual tests all still work :)

I like @mckn Idea of implementing OperatorFunciton, but perhaps in a new PR?  this seems to do most of the heavy lifting

hugohaggmark:
> My manual tests all still work :)
> 
> I like @mckn Idea of implementing OperatorFunciton, but perhaps in a new PR? this seems to do most of the heavy lifting

I agree, that could all be done in a later PR, WDYT @mckn ?

packages/grafana-data/src/transformations/transformers/calculateField.test.ts
```diff
@@ -4,9 +4,10 @@ import { FieldType } from '../../types/dataFrame';
 import { ReducerID } from '../fieldReducer';
 import { mockTransformationsRegistry } from '../../utils/tests/mockTransformationsRegistry';
 import { transformDataFrame } from '../transformDataFrame';
-import { calculateFieldTransformer, CalculateFieldMode, ReduceOptions } from './calculateField';
+import { CalculateFieldMode, calculateFieldTransformer, ReduceOptions } from './calculateField';
 import { DataFrameView } from '../../dataframe';
 import { BinaryOperationID } from '../../utils';
+import { observableTester } from '../../../../../public/test/helpers/observableTester';
```
dprokop, line 10: Usually, this kind of import in package gives me gooseflesh 😬 Wonder if we shouldn't move test helpers to separate package (grafana/test-utils ?) to make sure we don't break anything at some point. Just an idea, definitely not in the scope of this PR :)

packages/grafana-data/src/transformations/transformers/filter.ts
dprokop: Agree with @hugohaggmark , thinks this suggestion makes the code harder to reason about

packages/grafana-data/src/transformations/transformers/calculateField.test.ts
hugohaggmark: Hmm, you're right that isn't good at all. Strange that the linting doesn't prevent me from this?

packages/grafana-data/src/types/transformations.ts
dprokop: I agree with all of the above.

> A drawback is that we will have a big dependency on RxJS which will make it harder to replace in the future (if we would like to do that).

I'm not that concerned about this. Being in line with common RxJS patterns, when we already decided to use observables,  should be our goal. I think this will make it easier for the potential contributors to implement transformations as they will have a clear pattern they can refer to.

public/app/features/dashboard/state/PanelQueryRunner.ts
```diff
@@ -113,6 +102,25 @@ export class PanelQueryRunner {
     );
   }
 
+  private getTransformationsStream = (withTransforms: boolean): MonoTypeOperatorFunction<PanelData> => {
+    return inputStream =>
```
dprokop, line 106: Just out of curiosity: I remember some months ago you have used the standard RxJS naming convention with $ representing a stream (`input$`). Why are we no longer following this pattern?

dprokop (approved):
Such a massive work @hugohaggmark, amazing!

packages/grafana-data/src/transformations/transformers/calculateField.test.ts
mckn: Have you synced with latest master? Because with the updates med and Arve did last week the PR build should fail when this happens.

public/app/features/dashboard/state/PanelQueryRunner.ts
hugohaggmark: I removed that because I received comments about confusion with Angular :(

packages/grafana-data/src/transformations/transformers/calculateField.test.ts
hugohaggmark: I synced with master this morning and the build passed?

mckn:
> > My manual tests all still work :)
> > I like @mckn Idea of implementing OperatorFunciton, but perhaps in a new PR? this seems to do most of the heavy lifting
> 
> I agree, that could all be done in a later PR, WDYT @mckn ?

hmm, I guess that would work. The only drawback of that would be that we probably need to change the signature of the transforms again in that PR. But that should be ok, right?

hugohaggmark:
> > > My manual tests all still work :)
> > > I like @mckn Idea of implementing OperatorFunciton, but perhaps in a new PR? this seems to do most of the heavy lifting
> > 
> > 
> > I agree, that could all be done in a later PR, WDYT @mckn ?
> 
> hmm, I guess that would work. The only drawback of that would be that we probably need to change the signature of the transforms again in that PR. But that should be ok, right?

I would really appreciate merging this to master this week so others get a chance to find bugs before 7.3 but maybe we could chat about it during weekly today?

hugohaggmark:
@mckn can you re review the new changes and remove the request for changes if you agree with changes that is, thank you

packages/grafana-data/src/types/transformations.ts
hugohaggmark: Renamed it to operator as that is what is is now :)

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: Unfortunately TypeScript couldn't understand the typings as `apply` returns `any` but I left it as is with a `// @ts-ignore`

packages/grafana-data/src/types/transformations.ts
hugohaggmark: Done

packages/grafana-data/src/transformations/transformers/calculateField.test.ts
hugohaggmark: Moved observableTester to grafana/data instead until we add the testutils packages

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,68 @@
...
+ */
+export function transformDataFrame(options: DataTransformerConfig[], data: DataFrame[]): Observable<DataFrame[]> {
+  const stream = of<DataFrame[]>(data);
+
+  if (!options.length) {
+    return stream;
   }
-  return processed;
+
+  const operators: Array<OperatorFunction<DataFrame[], DataFrame[]>> = [];
```
mckn, line 59:
```suggestion
  const operators: Array<MonoTypeOperatorFunction<DataFrame[]>> = [];
```

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,68 @@
...
+    return source;
+  }
+
+  const defaultOptions = info.transformation.defaultOptions ?? {};
+  const options = { ...defaultOptions, ...config.options };
+
+  return source.pipe(
+    mergeMap(before => of(before).pipe(info.transformation.operator(options), postProcessTransform(before, info)))
+  );
+};
```
mckn, line 20:
```suggestion
const getOperator = (config: DataTransformerConfig): MonoTypeOperatorFunction<DataFrame[]> => {
  const info = standardTransformersRegistry.get(config.id);

  if (!info) {
    return pipe();
  }

  const defaultOptions = info.transformation.defaultOptions ?? {};
  const options = { ...defaultOptions, ...config.options };
  const operator = info.transformation.operator(options);
  
  return pipe(mergeMap(before => of(before).pipe(operator,  postProcessTransform(before, info))));
};
```

nit: I'm not 100% sure that you can return an empty pipe but might be worth testing to clean up the code a bit.

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,68 @@
...
+
+const postProcessTransform = (
+  before: DataFrame[],
+  info: TransformerRegistyItem<any>
+): MonoTypeOperatorFunction<DataFrame[]> => source =>
```
mckn, line 25: nit: same thing here. You could use the `pipe` utility function instead to clean up the code.

packages/grafana-data/src/transformations/transformers/ensureColumns.ts
```diff
@@ -1,25 +1,33 @@
...
+import { mergeMap } from 'rxjs/operators';
 
 export const ensureColumnsTransformer: DataTransformerInfo = {
   id: DataTransformerID.ensureColumns,
   name: 'Ensure Columns Transformer',
   description: 'Will check if current data frames is series or columns. If in series it will convert to columns.',
-  transformer: () => (data: DataFrame[]) => {
-    // Assume timeseries should first be joined by time
-    const timeFieldName = findConsistentTimeFieldName(data);
+  operator: (options = {}) => source =>
```
mckn, line 14: nit: I think I would prefer having the explicit return type here to make things super obvious.

mckn (approved):
I have added some suuper small nits but apart from that I think this looks great!! Really love the way we are going with this! Awesome work @hugohaggmark!

packages/grafana-data/src/transformations/transformDataFrame.ts
mckn: It is basically the same thing but for me it removes some noise.

packages/grafana-data/src/transformations/transformers/ensureColumns.ts
hugohaggmark: I've used the same construct in most transformer operators, I don't think I'll change them all.

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: Hmm I don't agree that it adds to readability and TypeScript complains about it. Maybe we can enhance this after this PR?
```ts
Type '(source: Observable<DataFrame[]>) => UnaryFunction<DataFrame[], unknown>' is not assignable to type 'MonoTypeOperatorFunction<DataFrame[]>'.
  Type 'UnaryFunction<DataFrame[], unknown>' is missing the following properties from type 'Observable<DataFrame[]>': _isScalar, source, operator, lift, and 6 more.
```

packages/grafana-data/src/transformations/transformDataFrame.ts
hugohaggmark: Same as above

public/app/features/dashboard/state/PanelQueryRunner.ts
```diff
@@ -113,6 +102,25 @@ export class PanelQueryRunner {
...
+            return of(data);
+          }
+
+          const transformations = this.dataConfigSource.getTransformations();
+
+          if (!transformations || transformations.length === 0) {
+            return of(data);
+          }
+
+          return transformDataFrame(transformations, data.series).pipe(map(series => ({ ...data, series })));
```
torkelo, line 119: why do we need this map?

hugohaggmark: I think this should be the eqvivalent of:
```ts
processedData = {	
              ...processedData,	
              series: transformDataFrame(transformations, data.series),	
            };
```
I'm not 100% sure why the original code did that, but I agree that it looks kind of odd.

torkelo: ah, now I get it

packages/grafana-data/src/transformations/transformDataFrame.ts
```diff
@@ -1,37 +1,68 @@
...
+
+  const operators: Array<MonoTypeOperatorFunction<DataFrame[]>> = [];
+
+  for (let index = 0; index < options.length; index++) {
+    const config = options[index];
+    operators.push(getOperator(config));
+  }
+
+  // @ts-ignore TypeScript has a hard time understanding this construct
+  return stream.pipe.apply(stream, operators);
```
dprokop, line 67: This is gold now @hugohaggmark  🥇

PR #27587, opened 2020-09-15
Annotation:  use DataFrame[] rather than a single DataFrame

ryantxu:
In 7.2 we are introducing a standard annotation support model.  I tried to apply this to loki, and discovered that we should not force merge all frames into a single frame. 

FYI: InfluxDB w/ flux is the only data source that uses this currently

### DIFF
--- a/packages/grafana-data/src/types/annotations.ts
+++ b/packages/grafana-data/src/types/annotations.ts
@@ -78,7 +78,7 @@ export interface AnnotationSupport<TQuery extends DataQuery = DataQuery, TAnno =
   /**
    * When the standard frame > event processing is insufficient, this allows explicit control of the mappings
    */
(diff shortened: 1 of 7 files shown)


packages/grafana-data/src/types/annotations.ts
```diff
@@ -78,7 +78,7 @@ export interface AnnotationSupport<TQuery extends DataQuery = DataQuery, TAnno =
   /**
    * When the standard frame > event processing is insufficient, this allows explicit control of the mappings
    */
-  processEvents?(anno: TAnno, data: DataFrame): AnnotationEvent[] | undefined;
+  processEvents?(anno: TAnno, data: DataFrame[]): AnnotationEvent[] | undefined;
```
ryantxu, line 81: This API is new in 7.2, so I think we should update before it gets released :grimacing:

torkelo (approved):
Looks good

PR #27591, opened 2020-09-15
Timezones: Include IANA timezone canonical name in TimeZoneInfo

dprokop:
Ref: https://www.iana.org/time-zones , https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
This PR also sets the global timezone for tests to UTC.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,5 @@
+process.env.TZ = 'UTC';
+
 module.exports = {
   verbose: false,
   transform: {
--- a/packages/grafana-data/src/datetime/timezones.ts
+++ b/packages/grafana-data/src/datetime/timezones.ts
@@ -30,6 +30,7 @@ export interface TimeZoneInfo {
   countries: TimeZoneCountry[];
   abbreviation: string;
   offsetInMins: number;
+  ianaName: string;
 }
 
 export interface GroupedTimeZones {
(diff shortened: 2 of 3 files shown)


packages/grafana-data/src/datetime/timezones.ts
```diff
@@ -148,13 +152,13 @@ const abbrevationWithoutOffset = (abbrevation: string): string => {
...
 const mapToInfo = (timeZone: TimeZone, timestamp: number): TimeZoneInfo | undefined => {
   const momentTz = moment.tz.zone(timeZone);
-
   if (!momentTz) {
     return undefined;
   }
 
   return {
     name: timeZone,
+    ianaName: momentTz.name,
```
mckn, line 161: I think the proper name is olsonId.

mckn (approved):

packages/grafana-data/src/datetime/timezones.ts
dprokop: Did a quick check on this and seems like there is no clear winner;) I think I'll stay with iana as [Wikipedia states Olson database is occasionally used in as a name of this DB](https://en.wikipedia.org/wiki/Tz_database)

jest.config.js
```diff
@@ -1,3 +1,5 @@
+process.env.TZ = 'UTC';
```
hugohaggmark, line 1: I guess this is deliberate, but feels very strange.

hugohaggmark (approved):
Looks great Dominik, I just left a minor nit.

jest.config.js
dprokop: Yeah, had the same feeling but apparently this is the way jest wants the time zone to be set. I have not found any other way to get the timezone "mocked".

PR #27594, opened 2020-09-15
Documentation: fixing args order for admin password reset command

Lenahilin:
Putting the optional argument "--homepath"  after the command. The previous one throws the following error: "Incorrect Usage. flag provided but not defined: -homepath". See also https://community.grafana.com/t/admin-password-reset/19455/9


achatterjee-grafana (approved):
Looks good to me.

PR #23566, opened 2020-04-14
CloudWatch: Adds support for Cloudwatch Logs

kaydelaney:
Initial PR for CloudWatch Logs support.

PR #27599, opened 2020-09-15
backend-plugins: point to request Headers in the wrapper

kylebrandt:
Points to the headers of a tsdb.Query from a QueryDataRequest so they are copied to protobuf and sent to a plugin.

In particular, this is so the plugin can see the `FromAlert` header:

https://github.com/grafana/grafana/blob/2e4191afca44c1eb303c096c25f6f98c21066614/pkg/services/alerting/conditions/query.go#L216-L234

@marefr, are there any headers they I should make sure are not relayed to the plugin (security maybe?). If that is the case I can make a copy and exclude those.

### DIFF
--- a/pkg/plugins/datasource/wrapper/datasource_plugin_wrapper_v2.go
+++ b/pkg/plugins/datasource/wrapper/datasource_plugin_wrapper_v2.go
@@ -59,6 +59,7 @@ func (tw *DatasourcePluginWrapperV2) Query(ctx context.Context, ds *models.DataS
 			DataSourceInstanceSettings: backend.ToProto().DataSourceInstanceSettings(instanceSettings),
 		},
 		Queries: []*pluginv2.DataQuery{},
+		Headers: query.Headers,
 	}
 
 	for _, q := range query.Queries {
--- a/pkg/plugins/backendplugin/coreplugin/query_endpoint_adapter.go
+++ b/pkg/plugins/backendplugin/coreplugin/query_endpoint_adapter.go
@@ -59,6 +59,7 @@ func (a *queryEndpointAdapter) Query(ctx context.Context, ds *models.DataSource,
 			DataSourceInstanceSettings: instanceSettings,
 		},
 		Queries: []backend.DataQuery{},
+		Headers: query.Headers,
 	}
 
 	for _, q := range query.Queries {


pkg/plugins/datasource/wrapper/datasource_plugin_wrapper_v2.go
```diff
@@ -59,6 +59,7 @@ func (tw *DatasourcePluginWrapperV2) Query(ctx context.Context, ds *models.DataS
 			DataSourceInstanceSettings: backend.ToProto().DataSourceInstanceSettings(instanceSettings),
 		},
 		Queries: []*pluginv2.DataQuery{},
+		Headers: query.Headers,
```
wbrowne, line 62: Are there headers other than `FromAlert` that you're interested in? If not, might be nice to package it up just as that to make it extra explicit

kylebrandt: It does make it a big ugly in the datasource:

```
var fromAlert bool
	if req.Headers != nil {
		if req.Headers["FromAlert"] == "true" {
			fromAlert = true
		}
	}
```

So perhaps should add a `FromAlert()` method to the request object in the sdk (which would be compatible with this, and basically just does this).

kylebrandt: And now that I see it that should be && :-P

kylebrandt: In terms of other headers, I think the idea is that the property is meant to be passed to plugins and was just skipped - but maybe I remember incorrectly and this was intentional - perhaps @marefr knows.

wbrowne: Yeah it does look a bit 💩  alright. Having it in the SDK could be an option too, but hopefully wouldn't unnecessarily complicate making changes by splitting the request composition.

Either way though probably a good idea to only pass what we need considering the cruft that gets thrown in headers most of the time - just my 2 cents!

marefr (changes requested):
Was added in #23566 5 months ago and CloudWatch seems to be the only data source using the header.

Should add it here to: https://github.com/grafana/grafana/blob/b9b96e4d25e5f63d2288580a61e14b6c9f2fab71/pkg/plugins/backendplugin/coreplugin/query_endpoint_adapter.go#L54-L62

Other than that, LGTM

marefr (approved):
LGTM

PR #27600, opened 2020-09-15
Grafana/Live: Adds function for retrieving promise-bound live srv

kaydelaney:
Adds a function, `getConnectedLiveSrv`, that returns a promise which resolves with the singleton live srv instance once it has connected. This should help simplify usage of the live srv in certain instances (as with LivePanel.tsx)

### DIFF
--- a/packages/grafana-runtime/src/services/live.ts
+++ b/packages/grafana-runtime/src/services/live.ts
@@ -80,10 +80,13 @@ export const getGrafanaLiveSrv = (): GrafanaLiveSrv => singletonInstance;
  * @experimental
  * @public
  */
-export const getConnectedLiveSrv = (): Promise<GrafanaLiveSrv> => {
-  return new Promise(resolve => {
+export const getConnectedLiveSrv = (): Promise<GrafanaLiveSrv> =>
+  new Promise(resolve => {
+    if (singletonInstance.isConnected()) {
+      return resolve(singletonInstance);
+    }
+
     singletonInstance.addListener('connect', () => {
       resolve(singletonInstance);
     });
   });
-};
--- a/packages/grafana-runtime/src/services/live.ts
+++ b/packages/grafana-runtime/src/services/live.ts
@@ -48,6 +48,8 @@ export interface GrafanaLiveSrv {
    * Send data to a channel.  This feature is disabled for most channels and will return an error
    */
   publish<T>(channel: string, data: any): Promise<T>;
+
+  addListener(event: string | symbol, listener: (...args: any[]) => void): void;
 }
 
 let singletonInstance: GrafanaLiveSrv;
@@ -70,3 +72,18 @@ export const setGrafanaLiveSrv = (instance: GrafanaLiveSrv) => {
  * @public
  */
 export const getGrafanaLiveSrv = (): GrafanaLiveSrv => singletonInstance;
+
+/**
+ * Returns a promise that resolves with the {@link GrafanaLiveSrv} once it has
+ * connected
+ *
+ * @experimental
+ * @public
+ */
+export const getConnectedLiveSrv = (): Promise<GrafanaLiveSrv> => {
+  return new Promise(resolve => {
+    singletonInstance.addListener('connect', () => {
+      resolve(singletonInstance);
+    });
+  });
+};
--- a/packages/grafana-runtime/src/services/live.ts
+++ b/packages/grafana-runtime/src/services/live.ts
@@ -30,6 +30,11 @@ export interface GrafanaLiveSrv {
   isConnected(): boolean;
 
   /**
+   * Resolves when the server is connected
+   */
+  waitUntilConnected(): Promise<void>;
+
+  /**
    * Listen for changes to the connection state
    */
   getConnectionState(): Observable<boolean>;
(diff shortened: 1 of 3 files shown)


packages/grafana-runtime/src/services/live.ts
```diff
@@ -48,6 +48,8 @@ export interface GrafanaLiveSrv {
    * Send data to a channel.  This feature is disabled for most channels and will return an error
    */
   publish<T>(channel: string, data: any): Promise<T>;
+
+  addListener(event: string | symbol, listener: (...args: any[]) => void): void;
```
ryantxu, line 52: what are possible event values?  for our API, should they be more specific than string|symbol?

packages/grafana-runtime/src/services/live.ts
```diff
@@ -70,3 +72,18 @@ export const setGrafanaLiveSrv = (instance: GrafanaLiveSrv) => {
...
+/**
+ * Returns a promise that resolves with the {@link GrafanaLiveSrv} once it has
+ * connected
+ *
+ * @experimental
+ * @public
+ */
+export const getConnectedLiveSrv = (): Promise<GrafanaLiveSrv> => {
+  return new Promise(resolve => {
+    singletonInstance.addListener('connect', () => {
```
ryantxu, line 85: what happens if it is already connected?

ryantxu: should we remove the listener after it is resolved?

packages/grafana-runtime/src/services/live.ts
kaydelaney: This is a good question, I just copied the call signature of the corresponding centrifuge method but you're right that there could be more specificity. I'll look into it

packages/grafana-runtime/src/services/live.ts
kaydelaney: Another good catch. Should add a check to see if we're already connected

packages/grafana-runtime/src/services/live.ts
kaydelaney: It seems like arbitrarily named events can be emitted with `centrifuge.emit`, but I suppose from the perspective of a consumer of the grafana live API that flexibility might not be needed or advisable. We could limit it just to the events that centrifuge-js itself makes use of (publish, join, leave, subscribe, unsubscribe, error, message, connect, disconnect)?

torkelo: Not sure what the purpose of this listener is we have the getChannelStream for this exact purpose no?

packages/grafana-runtime/src/services/live.ts
kaydelaney: I like this design! Pushed a change

packages/grafana-runtime/src/services/live.ts
```diff
@@ -70,3 +75,15 @@ export const setGrafanaLiveSrv = (instance: GrafanaLiveSrv) => {
...
+ * Returns a promise that resolves with the {@link GrafanaLiveSrv} once it has
+ * connected
+ *
+ * @experimental
+ * @public
+ */
+export const getConnectedLiveSrv = async () => {
+  await singletonInstance.waitUntilConnected();
+  return singletonInstance;
+};
```
torkelo, line 89: Looking at how we use the currently it does not look like we need two getLiveSrv functions. The connect happens in the constructor. So everyone who is interested in the LiveSrv should probably use getConnectedLiveSrv, so I would make this the default behavior. and only have a single getGrafanaLiveSrv 

But I also question a bit the need for this in the first place. Can't we make initChannel / getChannelStream / publish handle this connection wait internally?

torkelo: but having this as default could be a start

kaydelaney: This is fair actually, let me make some changes

public/app/features/live/live.ts
```diff
@@ -114,9 +115,19 @@ class CentrifugeSrv implements GrafanaLiveSrv {
   }
 
   initChannel<T>(path: string, handler: ChannelHandler<T>) {
+    const connected = from(
+      new Promise<void>(resolve => {
+        if (this.centrifuge.isConnected()) {
+          return resolve();
+        }
+
+        this.centrifuge.addListener('connect', resolve);
```
torkelo, line 124: you should be able to define this promise in the constructor and reuse it

torkelo: as a promise that has completed can be reused  (and will complete with same value)

torkelo (approved):
Looks good, but would move the connected promise to an instance property so it can be reused (like a cache of is connected state).

public/app/features/live/live.ts
```diff
@@ -133,15 +144,17 @@ class CentrifugeSrv implements GrafanaLiveSrv {
       },
     };
     c.subscription = this.centrifuge.subscribe(path, callbacks);
+    return connected;
   }
 
   getChannelStream<T>(path: string): Observable<T> {
```
ryantxu, line 150: The original idea is that a few things may connect to the same stream -- in an alternate PR, I tried:

https://github.com/grafana/grafana/pull/27583/files#diff-234f06c46c08d978307cf6c087712a55R134

torkelo: that should still be possible I think

packages/grafana-runtime/src/services/live.ts
hugohaggmark: I agree with Torkel, as a consumer I'm only interested in getting a stream to subscribe/publish too.

torkelo (approved):

PR #27603, opened 2020-09-15
Graph: only show range warning when all data is outside time range

### LINKED ISSUE #26279: data outside time range message
**What happened**: Get a message Data outside time range

**What you expected to happen**: To show no data or an option to change the message

**How to reproduce it (as minimally and precisely as possible)**: When grouping data by day, by default grafana will set the time to 00:00. But if the time now is for example 12PM and the timepicker is set to show the last 3 hours, the hour 00:00 will not show up on our timerange. 

This is related to my previous bug #24625  where i explained that this happens with Influx as well.
This happens to us when within the same dashboard we want to display both hourly summary data and daily summary data, simultaneously, using different group by methods.
For example, we want to show a daily summary of the data, but when use the variable $__interval, even when we choose a timespan of 7 days it many times will not default to a minimum interval of 1d. so we need to hard code the minimum value which causes this problem.

**Environment**:
- Grafana version: 6.7.2
- Data source type & version: Influx DB
- User OS & Browser: Chrome
- Grafana plugins:
- Others:
image

ryantxu:
Fixes #26279

We currently show a warning with an action when *any* series in the results is outside the time range.  This changes the logic to only show an error when *all* data is outside the time range.

I added this to 7.2 because it is minor and causes weird behavior with some non-timeseries data sources

### DIFF
--- a/public/app/plugins/panel/graph/module.ts
+++ b/public/app/plugins/panel/graph/module.ts
@@ -271,35 +271,35 @@ export class GraphCtrl extends MetricsPanelCtrl {
       };
     }
 
-    // Look for data points outside time range
+    // If any data is in range, do not return an error
     for (const series of this.seriesList) {
       if (!series.isOutsideRange) {
-        continue;
+        return undefined;
       }
-
-      const dataWarning: DataWarning = {
-        title: 'Data outside time range',
-        tip: 'Can be caused by timezone mismatch or missing time filter in query',
-      };
-
-      const range = getDataTimeRange(this.dataList);
-
-      if (range) {
-        dataWarning.actionText = 'Zoom to data';
-        dataWarning.action = () => {
-          getLocationSrv().update({
-            partial: true,
-            query: {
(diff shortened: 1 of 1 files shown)


public/app/plugins/panel/graph/module.ts
```diff
@@ -271,35 +271,35 @@ export class GraphCtrl extends MetricsPanelCtrl {
...
       if (!series.isOutsideRange) {
-        continue;
+        return undefined;
       }
+    }
 
-      const dataWarning: DataWarning = {
-        title: 'Data outside time range',
-        tip: 'Can be caused by timezone mismatch or missing time filter in query',
+    // All data is outside the time range
```
ryantxu, line 281: FYI, all changes after this are just indentation changes

torkelo (approved):
Yea, this is better logic :) :heart:

kylebrandt (approved):
Seems to work as I would expect.

PR #27614, opened 2020-09-16
Elasticsearch: Add query's refId to each series returned by a query

### LINKED ISSUE #8642: [Feature request] Add refId to each series returned by Elasticsearch query
- What Grafana version are you using?
  4.3.0

I am making a general data series visualization panel (ie plot the value of some elasticsearch tag vs another) and one feature I built in was to have the axes be from different query sources (A stupid example would be query A returning the average, query B returning the stdev). This works fine until groupings come into play, in which case the array index that is sent with data-receive no longer corresponds to the query issued. So in short, there is no way to extract what query generated a given source of data from js. Having an extra field for the query source (in the same object that has `datapoints`, `target`, etc.) would allow me to extract information based on the query type for visuals that can integrate across different datasources (ie pluck out a metric from a raw document from a ES data source and combine it with metrics in prometheus)

Elfo404:

This change adds the query's `refId` to each one of the series returned by that query so that it can be later used to perform transforms based on it.

Fixes #8642

I'm not completely happy with the verbosity of the tests, but I'm going to refactor them when performing the general plugin overhaul.

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/elastic_response.ts
+++ b/public/app/plugins/datasource/elasticsearch/elastic_response.ts
@@ -30,7 +30,7 @@ export class ElasticResponse {
 
       switch (metric.type) {
         case 'count': {
-          newSeries = { datapoints: [], metric: 'count', props: props };
+          newSeries = { datapoints: [], metric: 'count', props: props, refId: target.refId };
           for (i = 0; i < esAgg.buckets.length; i++) {
             bucket = esAgg.buckets[i];
             value = bucket.doc_count;
@@ -53,6 +53,7 @@ export class ElasticResponse {
               metric: 'p' + percentileName,
               props: props,
               field: metric.field,
+              refId: target.refId,
             };
 
             for (i = 0; i < esAgg.buckets.length; i++) {
@@ -76,6 +77,7 @@ export class ElasticResponse {
               metric: statName,
               props: props,
               field: metric.field,
+              refId: target.refId,
             };
 
             for (i = 0; i < esAgg.buckets.length; i++) {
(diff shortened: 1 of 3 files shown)


torkelo (commented):
Would be good to update tests 

https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/elasticsearch/specs/elastic_response.test.ts

marefr (commented):
We want refId for `processHits` as well. Think also for tables generated by `processAggregationDocs` - think that should be possible, but may be wrong in case always one table is returned even though you have multiple queries.

marefr (changes requested):
Tested this quickly for time series and with these changes I still can't get transformation filter data by query to render any series refid's:

image
image

Could it be something missing when legacy time series are converted to dataframes?

public/app/plugins/datasource/elasticsearch/elastic_response.ts
```diff
@@ -324,12 +327,13 @@ export class ElasticResponse {
...
 
-  processHits(hits: { total: { value: any }; hits: any[] }, seriesList: any[]) {
+  processHits(hits: { total: { value: any }; hits: any[] }, seriesList: any[], target: any) {
     const hitsTotal = typeof hits.total === 'number' ? hits.total : hits.total.value; // <- Works with Elasticsearch 7.0+
 
     const series: any = {
-      target: 'docs',
+      target: target.refId,
       type: 'docs',
+      refId: target.refId,
```
Elfo404, line 336: This was needed because otherwise the result would have been this: 
image

Elfo404, line 336: This was needed because otherwise the result would have been this: 
image

ivanahuckova:
LGTM! 🎈 Great work, awesome test coverage. Works as expected with all metrics that I have tested.

ivanahuckova (approved):
LGTM! 🎈 Great work, awesome test coverage. Works as expected with all metrics that I have tested.

marefr (approved):
LGTM and works as expected. Great work

PR #27615, opened 2020-09-16
InputControl: Fixed using InputControl in unit tests from plugins

### LINKED ISSUE #27488: InputControl is undefined in testing environment
**What happened**:

I upgraded the synthetic monitoring app plugin code from v7.0.6 to latest (v7.1.5). Tests that were working and testing components that made use of the `InputControl` component started failing because `InputControl` was being imported as undefined only in the test environment (code works fine in the browser). This is usually caused by a circular dependency, but I can't see one. If I install `react-hook-form` directly in my project and import the `Controller` component I don't have the same issue.

The `InputControl` component should be importable in a testing environment.

When testing [this component](https://github.com/grafana/synthetic-monitoring-app/blob/master/src/components/ProbeEditor.tsx), `InputControl` will be undefined. Tests are [here](https://github.com/grafana/synthetic-monitoring-app/blob/master/src/components/ProbeEditor.test.tsx)

[This](https://github.com/grafana/synthetic-monitoring-app/pull/97) is the PR that surfaced the issue.

**Environment**:
- Grafana version: 7.1.5
- User OS & Browser: Mac
- Grafana plugins: Synthetic monitoring

torkelo:
Fixes https://github.com/grafana/grafana/issues/27488

### DIFF
--- a/packages/grafana-ui/src/components/InputControl.tsx
+++ b/packages/grafana-ui/src/components/InputControl.tsx
@@ -1,1 +1,5 @@
-export { Controller as InputControl } from 'react-hook-form';
+/**
+ * Rollup does not support renamed exports so do not change this to export { Controller as InputControl } ...
+ */
+import { Controller } from 'react-hook-form';
+export const InputControl = Controller;


dprokop (approved):

peterholmberg (approved):
:+1:

PR #27622, opened 2020-09-16
DataLinks: Fixes issue with data links not interpolating values with correct field config 

torkelo:
Discovered bugs where `$[__value.text}`  and other data link variables did not actually use the field config unit. 

This was due the fact that the DataFrame instance passed to getLinkSupplier is a frame instance with field instances before field defaults & overrides being applied. 

Was a bit tricky to create a failing test but managed to replicate the error in a unit test.


kaydelaney (approved):
LGTM

robouden:
Is this fix also add to the nightly builds of 7.3.0?

PR #27628, opened 2020-09-16
ImageRendering: Fix rendering panel using shared query in png, PDF reports and embedded scenarios 

### LINKED ISSUE #27522: Add support for Image Rendering on Shared Panels
**What happened**:
Image renderer failing to display shared panels (panels that are using other panels as a datasource)

Screen Shot 2020-09-09 at 9 19 37 AM (1)

Panels that are using other dashboards as datasource should work and render properly (in the reports)

Screen Shot 2020-09-10 at 2 41 59 PM

**Environment**:
- Grafana version: Tested on both Grafana v7.0.1 and v7.1.5
- Data source type & version: "Dashboard"
- OS Grafana is installed on: CentOS Linux release 7.8.2003 | Also reproducible on Cloud and Enterprise

torkelo:
When rendering a solo route the isViewing or isEditing is both false so the logic of triggering the source panel query did not work. Setting isViewing to true did not work as it triggers special classes that affects height of panel. 

But using panel.isInView did fix this and it also fixes another bug where shared panel query did work if the source panel is out of view (scroll to see it). 

* Adds E2E test scenario for solo route & shared query 

Fixes #27522

### DIFF
--- a/e2e/suite1/specs/solo-route.spec.ts
+++ b/e2e/suite1/specs/solo-route.spec.ts
@@ -2,7 +2,7 @@ import { e2e } from '@grafana/e2e';
 
 e2e.scenario({
   describeName: 'Solo Route',
-  itName: 'Can view panes with shared queries in fullsceen',
+  itName: 'Can view panels with shared queries in fullsceen',
   addScenarioDataSource: false,
   addScenarioDashBoard: false,
   skipScenario: false,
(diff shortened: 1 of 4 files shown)


ryantxu (approved):

hugohaggmark (approved):
LGTM!

PR #27630, opened 2020-09-17
Live: scope channels to plugins

ryantxu:
The existing minimal implementation is a free-for-all where anything can control any path.  This PR lets plugins control the namespace under their pluginId, and delegates all channel handling to plugins.

TODO:
- [x] Error handling is not working
- [x] Define the channel handlers in the `grafana/*` namespace


public/app/features/live/live.ts
```diff
@@ -95,6 +93,77 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+    }
+
+    const liveSupport = await this.getChannelSupport(pluginId);
+    const handler = liveSupport.getChannelHandler(path);
+    if (!handler) {
+      return Promise.reject(`Invalid path: ${key}`);
+    }
+
+    // Make sure we are connected before trying to subscribe
+    if (!this.isConnected()) {
```
ryantxu, line 113: @kaydelaney -- I copied some stuff from #27600

This should have similar behavior now :crossed_fingers:

public/app/plugins/datasource/testdata/module.tsx
```diff
@@ -14,6 +14,17 @@ export const plugin = new DataSourcePlugin(TestDataDataSource)
   .setConfigEditor(ConfigEditor)
   .setQueryCtrl(TestDataQueryCtrl)
   .setAnnotationQueryCtrl(TestDataAnnotationsQueryCtrl)
+  .setLiveSupport({
+    getChannelHandler: (path: string) => {
```
ryantxu, line 18: note -- plugins now need to explicitly allow paths

public/app/features/admin/LiveAdmin.tsx
```diff
@@ -42,11 +42,19 @@ export class LiveAdmin extends PureComponent<Props, State> {
       const msg = {
         line: text,
       };
+      const idx = channel.indexOf('/');
+      const plugin = channel.substring(0, idx);
+      const path = channel.substring(idx + 1);
```
kaydelaney, line 47: If `plugin` and `path` can't contain forward slashes (is this validated anywhere?), then this can be replaced with
```suggestion
      const [plugin, path] = channel.split('/');
```

public/app/features/admin/LivePanel.tsx
```diff
@@ -39,14 +40,20 @@ export class LivePanel extends PureComponent<Props, State> {
...
 
     const srv = getGrafanaLiveSrv();
-    if (srv.isConnected()) {
-      const stream = srv.getChannelStream(this.props.channel);
+    const { channel } = this.props;
+
+    try {
+      const idx = channel.indexOf('/');
+      const plugin = channel.substring(0, idx);
+      const path = channel.substring(idx + 1);
```
kaydelaney, line 48: Same as above
```suggestion
      const [plugin, path] = channel.split('/');
```

public/app/features/live/live.ts
```diff
@@ -95,6 +93,77 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+      await new Promise<void>(resolve => {
```
kaydelaney, line 114: Think @torkelo made a good point in my PR that this could be an instance property

public/app/features/live/live.ts
```diff
@@ -95,6 +93,77 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+    return channel;
+  }
+
+  private async getChannelSupport(pluginId: string): Promise<LiveChannelSupport> {
+    return loadPlugin(pluginId).then(plugin => {
+      if (!plugin.liveSupport) {
+        return Promise.reject('Plugin does not have live support configured');
+      }
+      return plugin.liveSupport;
+    });
```
kaydelaney, line 164: Personal preference, but here's the function with async/await syntax

```suggestion
    const plugin = await loadPlugin(pluginId);
    if (!plugin.liveSupport) {
      throw new Error('Plugin does not have live support configured');
    }

    return plugin.liveSupport;
    });
```

public/app/features/live/live.ts
```diff
@@ -113,51 +182,34 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
    */
-  publish<T>(channel: string, data: any): Promise<T> {
-    return this.centrifuge.publish(channel, data);
+  async publish<T>(plugin: string, path: string, data: any): Promise<T> {
+    try {
+      const support = await this.getChannelSupport(plugin);
+      if (!support.onPublish) {
+        return Promise.reject({
+          text: `Channel ${path} does not allow publishing`,
+        });
```
kaydelaney, line 201: Don't think async functions need to wrap return values or errors in promises, so the following could just be
```suggestion
        throw {
          text: `Channel ${path} does not allow publishing`,
        };
```

public/app/features/live/live.ts
```diff
@@ -113,51 +182,34 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+      }
+      data = support.onPublish(path, data);
+
+      // Writes a message over the websocket to grafana server
+      return this.centrifuge.publish(`${plugin}/${path}`, data);
+    } catch (err) {
+      return Promise.reject(err);
```
kaydelaney, line 208: Don't think we need the try/catch blocks if the caught error is just going to be thrown again

kaydelaney (commented):
Left some comments, but it looks good!

public/app/features/admin/LiveAdmin.tsx
ryantxu: plugin can not contain slash, but path can and should be anything -- using a slash in the path should be normal.  For example in #27583 the channel should be `grafana/dashboard/${uid}`

public/app/features/live/live.ts
```diff
@@ -29,19 +29,21 @@ class CentrifugeSrv implements GrafanaLiveSrv {
     });
     this.centrifuge.connect(); // do connection
     this.connectionState = new BehaviorSubject<boolean>(this.centrifuge.isConnected());
+    this.connectionBlocker = new Promise<void>(resolve => {
```
ryantxu, line 32: @kaydelaney -- I made this a property rather than create a new one each time, is that what you mean?

public/app/features/live/live.ts
ryantxu: think changed that :)

public/app/features/live/live.ts
```diff
@@ -62,38 +65,86 @@ class CentrifugeSrv implements GrafanaLiveSrv {
     console.log('Publication from server-side channel', context);
   };
 
-  //----------------------------------------------------------
-  // Channel functions
-  //----------------------------------------------------------
+  /**
+   * Returns a channel that has been initalized and subscribed to the given path
+   */
+  private async getChannel<T>(pluginId: string, path: string): Promise<Channel<T>> {
```
hugohaggmark, line 71: I can see we're returning Promises internally. Do we need to do this? 

Instead of using `this.connectionBlocker` we could subscribe/react to connection states and react upon those. The consumer would start with an `empty` channel but that should be ok?

public/app/features/live/live.ts
```diff
@@ -29,19 +30,21 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+    this.connectionBlocker = new Promise<void>(resolve => {
+      if (this.centrifuge.isConnected()) {
+        return resolve();
+      }
+      const connectListener = () => {
+        resolve();
+        this.centrifuge.removeListener('connect', connectListener);
+      };
+      this.centrifuge.addListener('connect', connectListener);
+    });
```
torkelo, line 42: nice, yea, this was more how I envisioned it (a promise property to cache the connected async state)

public/app/features/live/live.ts
```diff
@@ -113,51 +164,32 @@ class CentrifugeSrv implements GrafanaLiveSrv {
...
+    if (!handler) {
+      throw new Error(`Invalid path: ${path}`);
+    }
+    if (!handler.onPublish) {
+      throw new Error(`Channel ${path} does not allow publishing`);
+    }
+    data = handler.onPublish(data);
+
+    // Writes a message over the websocket to grafana server
+    return this.centrifuge.publish(`${plugin}/${path}`, data);
```
torkelo, line 189: not sure I understand the handler.onPublish and onMessageReceived they are both event pre-processors?  what would the use case for handler.onPublish be? and handler. onMessageReceived

public/app/plugins/datasource/testdata/channels.ts
```diff
@@ -0,0 +1,52 @@
+import { ChannelSupport, ChannelHandler } from '@grafana/data';
+import {
+  SubscribeSuccessContext,
+  UnsubscribeContext,
+  JoinLeaveContext,
+  SubscribeErrorContext,
+} from 'centrifuge/dist/centrifuge.protobuf';
```
torkelo, line 7: I understand that this experimental, so ok. But this seems like a long term issue to expose the centrifuge types in the ChannelHandler interface that we expose to plugins

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  onLeave?: (ctx: any) => void;
+
+  /**
+   * This function must be defined for the channel to support publishing events
+   * into the websocket.  NOTE, the backend plugin must also allow publishing
+   * for this to work succesfully
+   */
+  onPublish?: (body: any) => any;
+}
```
torkelo, line 56: One other idea, is to use a single Observable for all events here. And define different event types for the different callbacks. But would require if/switch statement for the event types you care about.

should onPublish be (body: T) => any  , so like the inverse of onMessageReceived ?

torkelo: the Observable idea might not be great for onPublish and onMessageReceved as they need to return different data types.

torkelo (commented):
This is looking really promising! I like segregating channel namespaces with plugin!

public/app/features/live/live.ts
torkelo: > I can see we're returning Promises internally. Do we need to do this?

The plugin load returns a promise, or what do you mean?

Not sure what you mean, empty channel?

public/app/features/live/live.ts
ryantxu: `onMessageReceived` could get a single data point and will put it into a circular DataFrame -- the observable you subscribe to will give you a DataFrame -- or whatever makes sense for the channel.

The `onPublish` one is more awkward and I am not totally sure how it should work :)  We need some way for the channel handler to intercept and validate messages before we send them or even if we send them.  Currently the backend allows anyone to write messages to any channel -- that should not be the answer long term, but makes it easy to write cool features entirely in the frontend websocket.

public/app/plugins/datasource/testdata/channels.ts
ryantxu: you noticed!  yes the public ones are just <any> right now -- I had excluded them in the first round, but added them here so I can work with the join/leave callbacks (when other people join/leave a channel)

public/app/features/live/live.ts
hugohaggmark: I can see that my formulation is not the best. Sorry for that. `getChannel` is called from `getChannelStream` which is returning an Observable so I thought `getChannel` should also return an Observable and use more streams internally. 

In my mind, a live service or web socket handler is something that is inherently something that reacts to connection states and data. What happens when the connection between the client and server goes down? 

With that said, I think this is a cool and huge step forward for Grafana and I'm very excited about this, sorry for nitpicking at this stage.

torkelo: Don't apologize! your feedback on Observable patterns is super valuable and exactly what we need!  

I see what you mean now, getChannel, as it is used internally now is transformed to an Observable anyway via from

ryantxu: please nitpick!  and help figure out how to make it better :)  I am a bit out of my depth here, but feel good about the addition of a library that handles most of the details for us.

My use of promise here is really based on understanding them -- but I think it would be better as all Observable.  The thing i'm not sure is how to manage errors with observable -- i guess we could send an error stream for a plugin or channel that does not exist, but that feels weird

hugohaggmark: As a consumer of a live stream I would expect to subscribe to one stream and that stream to contain the entire state, with error or without. A lot like PanelData that contain both?

hugohaggmark: So you could handle that with two streams internally, merge them and expose just one externally?

ryantxu: Check #27672 -- hopefully it is a bit closer to what you would want.  The complication here is that the "channel" is not the stream -- the channel has other metadata and features attached that we need to query as well.  For example writing data and checking who else is on the channel (presence)

ryantxu:
Closing in favor of #27672 -- the single plugin namespace gets very complicated when you try to apply it to data sources!

PR #27643, opened 2020-09-17
Form: Add a form context consumer component

rdubrock:
Exposes `FormContext` functionality in `react-hook-form`. This allows folks to get access to form methods in deeply nested forms without prop drilling. I'm not 100% sure we _need_ this, it might be wedding our components too closely to `react-hook-form` if that's a concern, but I think it's a conversation worth having. Thanks for checking it out!

### DIFF
--- a/packages/grafana-ui/src/components/Forms/Form.tsx
+++ b/packages/grafana-ui/src/components/Forms/Form.tsx
@@ -1,5 +1,5 @@
 import React, { HTMLProps, useEffect } from 'react';
-import { useForm, Mode, OnSubmit, DeepPartial } from 'react-hook-form';
+import { useForm, Mode, OnSubmit, DeepPartial, FormContext } from 'react-hook-form';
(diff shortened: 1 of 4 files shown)


dprokop:
@rdubrock do you have any particular need for accessing the entire form context? 

I'm concerned about exposing the entire context as we loos the control over the API and this may make it hard to replace the react-hook-form if necessary in the future. But if you have some use case for this that you can describe then maybe we can find another way of addressing this.

rdubrock:
@dprokop In my own code I've only made use of `register`, `watch`, `control`, and `errors`. I imagine those are going to be the most important things to have access to, since you need at least `register`, `control` (for controlled components), and `errors` to set up a form field. If we're looking to limit the scope of methods we pass through, I'd advocate for those four. Or maybe just matching what the `Form` component exposes to `children` would be the simplest for maintainability?

PR #27646, opened 2020-09-18
RFC/POC/WIP: Exemplar and annotation support

ryantxu:
This is a super quick/dirty stab at how we may support exemplars and annotations.

This:
1. Adds the idea of a `data topic` that can be attached to the query or the result metadata
2. runRequest will set the dataTopic on results where the query should point to a channel
3. preProcessPanelData will move each topic to its own propety

3 is super debatable what the best path is.  With this route, we don't need special handling for "i want annotations/exemplars", but it may complicate (or improve!) other areas like applying transformations and series overrides.  We may want these configured by topic anyway

### DIFF
--- a/packages/grafana-data/src/types/data.ts
+++ b/packages/grafana-data/src/types/data.ts
@@ -15,6 +15,12 @@ export enum LoadingState {
   Error = 'Error',
 }
 
+export enum DataTopic {
+  Data = 'data', // or undefined
+  Annotations = 'annotations',
+  Exemplars = 'exemplars',
+}
+
 export type PreferredVisualisationType = 'graph' | 'table' | 'logs' | 'trace';
 
 export interface QueryResultMeta {
@@ -32,6 +38,12 @@ export interface QueryResultMeta {
 
   /** Currently used to show results in Explore only in preferred visualisation option */
   preferredVisualisationType?: PreferredVisualisationType;
+
+  /**
+   * Optionally identify which topic the frame should be assigned to.
+   * A value specified in the response will override what the request asked for.
+   */
+  dataTopic?: DataTopic;
 
   /**
    * This is the raw query sent to the underlying system.  All macros and templating
--- a/packages/grafana-data/src/types/data.ts
+++ b/packages/grafana-data/src/types/data.ts
@@ -16,7 +16,6 @@ export enum LoadingState {
 }
 
 export enum DataTopic {
-  Data = 'data', // or undefined
   Annotations = 'annotations',
   Exemplars = 'exemplars',
 }
--- a/packages/grafana-data/src/types/panel.ts
+++ b/packages/grafana-data/src/types/panel.ts
@@ -27,6 +27,12 @@ export interface PanelData {
   /** Contains data frames with field overrides applied */
   series: DataFrame[];
 
+  /** A list of annotation items */
+  annotations?: DataFrame[];
+
+  /** A list of exemplar data */
+  exemplars?: DataFrame[];
+
   /** Request contains the queries and properties sent to the datasource */
   request?: DataQueryRequest;
 
--- a/public/app/features/dashboard/state/runRequest.ts
+++ b/public/app/features/dashboard/state/runRequest.ts
@@ -17,6 +17,8 @@ import {
   toDataFrame,
   DataFrame,
   guessFieldTypes,
+  DataTopic,
+  KeyValue,
 } from '@grafana/data';
 import { toDataQueryError } from '@grafana/runtime';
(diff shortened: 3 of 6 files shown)


packages/grafana-data/src/types/data.ts
```diff
@@ -33,6 +39,12 @@ export interface QueryResultMeta {
   /** Currently used to show results in Explore only in preferred visualisation option */
   preferredVisualisationType?: PreferredVisualisationType;
 
+  /**
+   * Optionally identify which topic the frame should be assigned to.
+   * A value specified in the response will override what the request asked for.
+   */
+  dataTopic?: DataTopic;
```
ryantxu, line 46: @aocenas -- you will ask why not use the "preferredVisualisationType"  -- my understanding of this property when used in a dashboard context is that it would let the data source say which *panel* type would be a good choice.  Exemplars and annotations should exist in the same panel the series graph/table/logs etc

packages/grafana-data/src/types/panel.ts
```diff
@@ -27,6 +27,12 @@ export interface PanelData {
   /** Contains data frames with field overrides applied */
   series: DataFrame[];
 
+  /** A list of annotation items */
+  annotations?: DataFrame[];
+
+  /** A list of exemplar data */
+  exemplars?: DataFrame[];
```
torkelo, line 34: this makes me really want typed data frames :) 

like 
https://github.com/data-forge/data-forge-ts/blob/master/src/lib/dataframe.ts#L279

torkelo (commented):
Think this looks really promising!

packages/grafana-data/src/types/panel.ts
ryantxu: but why?  don't exemplars include the same variation in data?   I am assuming this is a method to say "the query is aggregating a bunch of data, but here are a few samples of real data" -- or am i thinking about it wrong?

Do exemplars have a strict structure?

torkelo: They need some assumed structure. But your right we can be quite flexible and use the first date field, first number for value, first string for traceId.

packages/grafana-data/src/types/data.ts
aocenas: well the meaning can evolve if needed I think. I mean prefVisType does not need to map 1 to 1 with panels. Like for example if somebody wants a line together with bars in one panel and create a plugin for that it may define it can show both bars and line and so could show 2 data frames with different prefVisType. I mean I am not saying this would be the best approach but this was sort of how I was thinking about it. In that case graph panel could handle multiple visTypes like Graph, Annotations (vertical lines) and Examplars (on hover click data).

aocenas: Again not sure if good idea I feel at that point in Explore it would be hard to route the data in all cases. But for example one thing we may be adding is some sort of group key to say what can be grouped how, so that if you have multiple graph results you could say if that should be 2 separate panels or rather single panel. That could be used to group exemplars and graph together so maybe it would still work.

packages/grafana-data/src/types/panel.ts
aocenas: Right now for example we have some runtime functions to check for example if dataframe is timeseries, which I think is implemented as has a time column and at least one numeric column. Having that typed would probably help quite a bit in general.

public/app/features/dashboard/state/runRequest.ts
```diff
@@ -178,10 +210,29 @@ export function getProcessedDataFrames(results?: DataQueryResponseData[]): DataF
       }
     }
 
-    dataFrames.push(dataFrame);
+    switch (dataFrame.meta?.dataTopic) {
```
aocenas, line 213: I feel like the dashboard actually should not know about what topics exists. I think datasource should return data with different topics and then that should be used by the panel in what ever way it deems fit. This way you can define new topic if you create a custom panel and a custom data source.

torkelo: this is not part of the dashboard, but just request processing. The question is where should the response segregation (into topics) be done, here or inside the panels? (single dataFrame array to panels or split into different topic arrays). 

Not sure I like these topic arrays.

aocenas: > I feel like the dashboard actually should not know about what topics exists.

I meant it more in the sense that Grafana itself should not dictate the topics imho. Ie panel and dashboards as plugins themselves should be able to negotiate what topics they provide and what topics they could consume, Grafana only passing it through without needing to understand what those topics mean.

torkelo:
This PR builds on this https://github.com/grafana/grafana/pull/27757

PR #27648, opened 2020-09-18
Fix queryoptionrow title

### LINKED ISSUE #26708: InfluxDB: when collapse query builder, buttons can get lost if query line is long

peterholmberg:
Recreate #27509 because of CI weirdness.

Fixes #26708


papagian:
@peterholmberg maybe you would like to update this PR's description with the issue that is fixing so as it will automatically be closed after this will be merged.

torkelo (approved):

PR #27654, opened 2020-09-18
Dashboard: Honour root_url for Explore link

### LINKED ISSUE #27588: "Explore" link in panel menu not honouring root_url
explore

**What happened**:
When opening "Explore" link from panel menu (with command + click to open in a new tab) the "graph" part of the URL is missing (root_url).

Expected "graph" to stay in the URL.

**Environment**:
- Grafana version: 7.1.3
- User OS & Browser: macOS & Chrome

tiagomotasantos:
Honour root_url when opening "Explore" in a new window from panel menu.

Fixes #27588

### DIFF
--- a/public/app/features/dashboard/utils/getPanelMenu.test.ts
+++ b/public/app/features/dashboard/utils/getPanelMenu.test.ts
@@ -1,6 +1,16 @@
+import { PanelMenuItem } from '@grafana/data';
 import { DashboardModel, PanelModel } from '../state';
 import { getPanelMenu } from './getPanelMenu';
 import { describe } from '../../../../test/lib/common';
+import { setStore } from 'app/store/store';
+import config from 'app/core/config';
+import * as actions from 'app/features/explore/state/actions';
+
+jest.mock('app/core/services/context_srv', () => ({
+  contextSrv: {
+    hasAccessToExplore: () => true,
+  },
+}));
 
 describe('getPanelMenu', () => {
   it('should return the correct panel menu items', () => {
@@ -27,6 +37,12 @@ describe('getPanelMenu', () => {
           "onClick": [Function],
           "shortcut": "p s",
           "text": "Share",
+        },
+        Object {
+          "iconClassName": "compass",
+          "onClick": [Function],
+          "shortcut": "x",
+          "text": "Explore",
         },
         Object {
           "iconClassName": "info-circle",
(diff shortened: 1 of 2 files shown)


hugohaggmark (commented):
Looks great @tiagomotasantos but would be nice to include a test. Let me know if you need help with that.

tiagomotasantos:
Thanks @hugohaggmark! Yes I could use some help with that. What type of test do you mean and how would you test it?

hugohaggmark:
Hi @tiagomotasantos if you could add a test for the new logic you added it would be great, let me know if you need any help.
https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/utils/getPanelMenu.test.ts

public/app/features/dashboard/utils/getPanelMenu.test.ts
```diff
@@ -143,4 +165,55 @@ describe('getPanelMenu', () => {
...
+        preventDefault: jest.fn(),
+      };
+
+      setStore({ dispatch: jest.fn() } as any);
+    });
+
+    it('should navigate to url without subUrl', () => {
+      if (explore?.onClick) {
+        explore.onClick(event);
+      }
```
hugohaggmark, line 196: No need for this if statement in tests I think?
```suggestion
        explore?.onClick(event);
```

public/app/features/dashboard/utils/getPanelMenu.test.ts
```diff
@@ -143,4 +165,55 @@ describe('getPanelMenu', () => {
...
+
+      expect(windowOpen).toHaveBeenLastCalledWith(testUrl);
+    });
+
+    it('should navigate to url with subUrl', () => {
+      config.appSubUrl = testSubUrl;
+
+      if (explore?.onClick) {
+        explore.onClick(event);
+      }
```
hugohaggmark, line 210: No need for this if statement in tests I think?
```suggestion
      explore?.onClick(event);      
```

hugohaggmark (commented):
Looking outstanding 🌟  just some small nits left!

public/app/features/dashboard/utils/getPanelMenu.test.ts
tiagomotasantos: We actually need because explore variable might be undefined and onClick method is optional. But I forced explore to be always of type PanelMenuItem and used non-null assertion operator for the onClick method to tell the compiler it will never be undefined. So it's cleaner now :)

hugohaggmark (approved):
Thank you for contributing @tiagomotasantos ⭐

PR #27655, opened 2020-09-18
Limit concurrent user sessions

joanlopez:

It introduces an extensible way to limit the maximum number of concurrent sessions per user.

- Still WIP since I'm looking for DELETE statement improvements, but completely open to feedback.
- Manually tested for user & SAML auth mechanisms (cc/ @mjseaman).


pkg/services/auth/auth_token.go
```diff
@@ -404,6 +410,35 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
 	return getTime().Add(-s.Cfg.LoginMaxInactiveLifetime).Unix()
 }
 
+const (
+	concurrentSessionLimit    = 3
+	concurrentSessionLimitKey = "concurrentSessionLimit"
+)
+
+func (s *UserAuthTokenService) isConcurrentSessionLimitEnabled() bool {
+	return s.Cfg.FeatureToggles[concurrentSessionLimitKey]
```
marefr, line 419: Why feature toggle and not a setting in [[auth]](https://grafana.com/docs/grafana/latest/administration/configuration/#auth)?

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +410,35 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
 	return getTime().Add(-s.Cfg.LoginMaxInactiveLifetime).Unix()
 }
 
+const (
+	concurrentSessionLimit    = 3
```
marefr, line 414: Would rather have this as a setting in [[auth]](https://grafana.com/docs/grafana/latest/administration/configuration/#auth). WDYT?

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +410,35 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
...
+	return s.Cfg.FeatureToggles[concurrentSessionLimitKey]
+}
+
+func (s *UserAuthTokenService) revokeOldUserTokens(ctx context.Context) error {
+	if !s.isConcurrentSessionLimitEnabled() {
+		return nil
+	}
+
+	return s.SQLStore.WithDbSession(ctx, func(dbSession *sqlstore.DBSession) error {
+		result, err := dbSession.Exec("DELETE FROM user_auth_token WHERE id NOT IN (SELECT id FROM user_auth_token WHERE seen_at <> 0 ORDER BY seen_at DESC LIMIT ?) AND seen_at <> 0;", concurrentSessionLimit-1)
```
marefr, line 428: Should filter by user id here. Couldn't this query be written with only delete clause/without `in(...)`, maybe not possible with order by then?

marefr (dismissed):
See comments

pkg/services/auth/auth_token.go
joanlopez: I'm not really sure about this. Maybe @mjseaman can give us with more details.
The issue definition specifies it as such, I suppose it's somehow related with AWS requirements.

pkg/services/auth/auth_token.go
joanlopez: Let's wait for Mitch to ask about [this](https://github.com/grafana/grafana/pull/27655#discussion_r491961771).

But yup, I agree that moving this constant into a configuration variable could be a good idea 👌

pkg/services/auth/auth_token.go
joanlopez: Completely agree, I already figured out the filter by user id missing but I was waiting for feedback to do it. Anyway, thanks for pointing it, it helped me to do not forget it.

Regarding the query, I cannot see a simpler way to do it. Probably, the subquery might be replaced by a JOIN clause but not really sure if in a simpler / clearer way. Feel free to write down an example if you have something in mind, I'd really appreciate it.

Thanks!

joanlopez: Related with the query ☝️ I already was exploring the approach using something like `DELETE ... LIMIT ... OFFSET` which might be enough for our case and also [seem to be supported by SQLite](https://www.sqlite.org/compile.html#enable_update_delete_limit) (also by PosgreSQL & MySQL I guess), but it [didn't seem to be supported by the Go client](https://github.com/mattn/go-sqlite3/issues/787).

pkg/services/auth/auth_token.go
marefr: Feature toggles are only used for temporary features that eventually will be enabled/included per default. Usually used when developing new experimental features. Adding it as a separate setting is basically a toggle.

pkg/services/auth/auth_token.go
alexanderzobnin: I think we shouldn't change default behavior when introducing new features. So I'd prefer to have two config options: one for enabling user session limit and second for setting limit, like:
```ini
[auth]
enable_user_session_limit = false
user_session_limit = 10
```
Or maybe even just one config option `user_session_limit` which is `0` by default and disables limit.

pkg/services/auth/auth_token.go
alexanderzobnin: Agreed with @marefr, we don't need feature toggle in this case, this is not an experimental feature.

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +410,49 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
...
+	concurrentSessionLimit    = 3
+	concurrentSessionLimitKey = "concurrentSessionLimit"
+)
+
+func (s *UserAuthTokenService) isConcurrentSessionLimitEnabled() bool {
+	return s.Cfg.FeatureToggles[concurrentSessionLimitKey]
+}
+
+func (s *UserAuthTokenService) revokeOldUserTokens(ctx context.Context, userId int64) error {
+	if !s.isConcurrentSessionLimitEnabled() {
```
alexanderzobnin, line 423: Isn't better to check this in `CreateToken()` function? In that case it looks more straightforward and clear I think.

alexanderzobnin (dismissed):

pkg/services/auth/auth_token.go
marefr: Based on earlier setting like `login_cookie_name`, `login_maximum_inactive_lifetime_duration` I would suggest  `login_maximum_concurrent_user_sessions` that default to 0 (disabled).

pkg/services/auth/auth_token.go
mjseaman: Hi @marefr and @alexanderzobnin, this is a licensing enforcement feature, which we are making available to AWS before rolling it out ourselves. 

The purpose of the feature is to discourage multiple users from signing in using the same credentials (and therefore paying for fewer users). I believe that if the concurrent session limit is a setting, then customers will always be able to disable the limit, right? That would make our license enforcement ineffective. 

The plan is to let AWS roll this out for their managed service in November, begin measuring the number of concurrent sessions for our own users, and then carefully roll it out as a default, non-configurable part of our own paid products.

joanlopez:
The initial approach has been changed. 

So, the concurrent user session limits are finally implemented as an extensible mechanism that has not limits by default but it can be overwritten when needed.

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +414,36 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
...
+		AND user_id = ?
+		AND seen_at != 0;`
+
+		result, err := dbSession.Exec(sql, userId, maxConcurrentSessions-1, userId)
+		if err != nil {
+			return err
+		}
+
+		rowsAffected, err := result.RowsAffected()
+		if err == nil && rowsAffected > 0 {
```
sakjur, line 439: Is it safe to ignore errors here?

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +414,36 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
...
+func (s *UserAuthTokenService) revokeOldUserTokens(ctx context.Context, userId int64, maxConcurrentSessions int) error {
+	return s.SQLStore.WithDbSession(ctx, func(dbSession *sqlstore.DBSession) error {
+		sql := `
+		DELETE
+		FROM user_auth_token
+		WHERE id NOT IN
+			(
+			  SELECT id
+			  FROM user_auth_token
+			  WHERE user_id = ?
```
sakjur, line 426: We don't have an index for `user_id` in https://github.com/grafana/grafana/blob/b0df7280be60be815078e572f5975a14521695bf/pkg/services/sqlstore/migrations/user_auth_token_mig.go#L8-L27

@marefr wdyt, should we add an index here or is there a better way to go about this?

sakjur (changes requested):
Logic looks good, but I think we're missing an index for **user_id** on the **user_auth_token** table.

pkg/services/auth/auth_token.go
```diff
@@ -404,6 +414,36 @@ func (s *UserAuthTokenService) rotatedAfterParam() int64 {
...
+			  ORDER BY seen_at DESC
+			  LIMIT ?
+			)
+		AND user_id = ?
+		AND seen_at != 0;`
```
marefr, line 431: Not sure about the logic for `seen_at`. Maybe moving this above inserting a new token would simplify things? Doing that could do a query to receive all auth tokens for user and then iterate and find the oldest (based on created_at or rotated_at) above `maxConcurrentSessions-1`  token and finally `delete from user_auth_token where id in(<list of id's>)`. 

Doing this you could potentially move all of this logic to enterprise by wrapping and overriding the `UserTokenService` in similar manner as done with the [data source cache service](https://github.com/grafana/grafana/blob/master/pkg/services/datasources/cache.go) using [registry.RegisterOverride](https://github.com/grafana/grafana/blob/master/pkg/registry/registry.go#L45) (in enterprise). @joanlopez @xlson thoughts?

pkg/services/auth/auth_token.go
joanlopez: It basically returns a `errors.New("AffectedArrows is not supported by this driver")` error or a similar one for DDL statements (create, alter, drop database, etc). So... yes? 😇 

If an error happens, the `Exec` method should return it, I guess.

pkg/services/auth/auth_token.go
joanlopez: Interesting one! Let's wait for @marefr point of view, but it might have sense since filtering by `user_id` will still be required, I guess 🤔  

Thanks!

pkg/services/auth/auth_token.go
joanlopez: Yup! I also thought about it, but I did it before just to avoid removing an existing session with the possibility of having issues while creating the new one. But this might be something insignificant for this case. So, if it helps us to simplify the code, I agree we can do it the other way round.

pkg/services/auth/auth_token.go
sakjur: :+1: Thanks!

pkg/services/auth/auth_token.go
joanlopez: > Doing this you could potentially move all of this logic to enterprise by wrapping and overriding the UserTokenService ª in similar manner as done with the data source cache service using registry.RegisterOverride (in enterprise). @joanlopez @xlson thoughts?

This approach produces `inject` errores because the `models.UserTokenService` interface is injected into some other services and once both implementations are present it does not know which one should be injected.

Any idea? 🤔  cc/ @marefr @xlson @sakjur 

Thank you so much!!! 🙏

joanlopez:
I proceed to close it since it's outdated and no longer valid.

The major outcome from this discussion was https://github.com/grafana/grafana/pull/27908.

Thanks guys!!! 👍

PR #27656, opened 2020-09-18
StatPanel: Fix problem where string values where always green

### LINKED ISSUE #27516: Stat panel font color always green for non numeric value
**What happened**:
Stat panel allows to change threshold color, and the default for the base is green. But when the result cannot convert to the number table, the result shows always green, even if I show the base color to another one. In the example, I had converted a timestamp to varchar so I could format it to show only hours. So this text cannot have the color changed, causing confusion on the dashboard.

image

When I change the base color to black, for example, and having no other threshold, I was expecting the content to be always black and not green.

Add a text field to Stat panel and change the base color to black. It will remain green.

**Environment**:
- Grafana version: Grafana v7.1.5 (9893b8c53d)
- Data source type & version: PostgreSQL 11 w/ Timescale
- OS Grafana is installed on: Ununtu 18.04
- User OS & Browser: Windows 10 / Chrome
- Grafana plugins: Stat
- Others:

peterholmberg:
As the issue describes, thresholds are only applied to numeric values right now (makes sense) but it also means that if you change the base color, string values will be green despite what you set the base color to.

Fixes #27516

This is a proposed solution and to be honest I'm not sure if we should use thresholds this way? And should we include all FieldTypes?

### DIFF
--- a/packages/grafana-ui/src/utils/standardEditors.tsx
+++ b/packages/grafana-ui/src/utils/standardEditors.tsx
@@ -146,7 +146,7 @@ export const getStandardFieldConfigs = () => {
         { value: 80, color: 'red' },
       ],
     },
-    shouldApply: field => field.type === FieldType.number || field.type === FieldType.string,
+    shouldApply: () => true,
     category: ['Thresholds'],
     getItemsCount: value => (value ? value.steps.length : 0),
   };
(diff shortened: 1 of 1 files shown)


packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -146,7 +146,7 @@ export const getStandardFieldConfigs = () => {
         { value: 80, color: 'red' },
       ],
     },
-    shouldApply: field => field.type === FieldType.number,
```
torkelo, line 149: Might as well return true here always

torkelo (approved):
Looks good! think it might be best to just always apply thresholds

ryantxu (approved):

Issue #26512, opened 2020-07-22
Grafana-UI: missing css prop warning for Form elements


Clarity-89:
**What happened**:
After upgrading @grafana/ui to 7.1.x in a plugin the following Typescript error appeared:
```
Property 'css' is missing in type '{ value: string; placeholder: string; onChange: (event: ChangeEvent<HTMLInputElement>) => void; }' but required in type 'Pick<Props, "type" | "key" | "label" | "value" | "cite" | "data" | "form" | "slot" | "span" | "style" | "summary" | "title" | "pattern" | "children" | "prefix" | "suffix" | ... 347 more ... | "addonAfter">'.
```
Seems like this happens for UI elements using `Pick` to extend interfaces. 

Could be related to the latest Typescript upgrade as it didn't happen before. Also might be relevant: https://github.com/emotion-js/emotion/issues/1249.

**What you expected to happen**:
The error shouldn't show.

**How to reproduce it (as minimally and precisely as possible)**:
Use @grafana/ui 7.1.x in a plugin. 

**Anything else we need to know?**:

**Environment**:
- Grafana version: 7.1.x
- Typescript: 3.9.3

Clarity-89:
Adding [this declaration](https://github.com/emotion-js/emotion/blob/master/packages/core/types/index.d.ts#L82) to index.d.ts seems to fix the issue.

wz2b:
<Input> and &lt;Switch&gt; have this problem.  For some reason, &lt;Select&gt; does not.  (&lt;Select&gt; also doesn't accept ref={register})

Clarity-89:
> Select> also doesn't accept ref={register}).


For Select you need to use `InputControl`, an example can be found in the [storybook](https://developers.grafana.com/ui/latest/index.html?path=/docs/forms-example-forms--async-validation) under `controls` heading. It is a wrapper for react-hook-form [Control](https://react-hook-form.com/api/#Controller) component.

daniellee:
Adding this import fixed it for me:

```ts
import {} from '@emotion/core';
```

Ref: https://github.com/emotion-js/emotion/issues/693#issuecomment-448088489

DislayPS:
Following on this issue please. I'm trying to use Checkbox on Grafana 7.1.x and Typescript tells me it's missing a bunch of properties

JorundMartinsen:
I'm getting this again, in 8.0.1


**How to reproduce it (as minimally and precisely as possible)**:

[...]

PR #27660, opened 2020-09-18
Grafana-UI: Refactor legacy inline form components

### LINKED ISSUE #27596: Grafana-UI: Refactor legacy inline form components
As discussed in the design group meeting it'd be a good idea to start refactoring the legacy inline form components, even though no redesigns have been done yet. As long as the api is kept consistent the future design changes shouldn't affect the components usage.

- [ ] Refactor `.gf-form` components into `InlineField` (or `Field` with `inline` toggle prop), similar to the current `Field` component.
- [ ] Move all the related styling from `.scss` to emotion.
- [ ] Make sure the components look and behave properly inside `Form`.

**Why is this needed**:
- Ensure consistent usage of components
- Simplify html structure by providing wrapper components 
- Improve accessibility by adding some attribute automatically (for ex. adding `htmFor` to the label based on input's `id`)

Clarity-89:
- Add `InlineField` component (refactor of legacy `FormField`)
- Add `InlineLabel` component (basically moved the current `InlineFieldLabel` to the `Forms`, so not to break legacy compatibility)
- Add `InlineFieldRow` to align inline fields in rows (works save way as current `gf-form-inline`).

Fixes https://github.com/grafana/grafana/issues/27596

Screenshots:
Screenshot 2020-09-21 at 15 33 17


peterholmberg:
The tooltip is a bit off :thinking: 
Screenshot from 2020-09-22 10-28-43

Clarity-89:
> The tooltip is a bit off 🤔
> Screenshot from 2020-09-22 10-28-43

Yeah, it seems like it's off for all the components in storybook, couldn't figure out what's wrong.

packages/grafana-ui/src/components/Forms/InlineField.tsx
```diff
@@ -0,0 +1,102 @@
...
+  className?: string;
+  /** Make the field's child to fill the width of the row. Equivalent to setting `flex-grow:1` on the field */
+  grow?: boolean;
+  /** A toggle to apply query keyword styling to the label */
+  isKeyword?: boolean;
+  /** Fill the remaining width of the row with the label's background */
+  fill?: boolean;
+}
+
+export const InlineField: FC<Props> = ({
```
dprokop, line 33: Shouldn't this extend [FieldProps](https://github.com/grafana/grafana/blob/416111c5f6afb32c4036724d7f87fa7a1102055e/packages/grafana-ui/src/components/Forms/Field.tsx#L8)? I think the HTMLProps extension should belong to the FieldProps too, then you could use that type as a base for InlineField properties

packages/grafana-ui/src/components/Forms/InlineFieldRow.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC, ReactNode, HTMLProps } from 'react';
+import { css, cx } from 'emotion';
+import { useStyles } from '../../themes';
+
+export interface Props extends Omit<HTMLProps<HTMLDivElement>, 'css'> {
```
dprokop, line 5: Hmm, why `css` is omitted here? How come is this a property on HTMLProps type?

packages/grafana-ui/src/components/Forms/InlineLabel.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { FunctionComponent, HTMLProps, ReactNode } from 'react';
+import { GrafanaTheme } from '@grafana/data';
+import { css, cx } from 'emotion';
+import { Tooltip, PopoverContent } from '../Tooltip/Tooltip';
+import { Icon } from '../Icon/Icon';
+import { useTheme } from '../../themes';
+
+export interface Props extends Omit<HTMLProps<HTMLLabelElement>, 'children' | 'className' | 'css'> {
```
dprokop, line 8: Think this should extend [Label props](https://github.com/grafana/grafana/blob/416111c5f6afb32c4036724d7f87fa7a1102055e/packages/grafana-ui/src/components/Forms/Label.tsx#L8) to keep the interfaces in sync. WDYT?

dprokop (commented):
Looks good, left few nits to consider :)

packages/grafana-ui/src/components/index.ts
```diff
@@ -145,6 +145,8 @@ export { Field } from './Forms/Field';
 export { Legend } from './Forms/Legend';
 export { FieldSet } from './Forms/FieldSet';
 export { FieldValidationMessage } from './Forms/FieldValidationMessage';
+export { InlineField } from './Forms/InlineField';
```
Elfo404, line 148: Shouldn't `InlineFieldRow` also be exported here?

```suggestion
export { InlineField } from './Forms/InlineField';
export { InlineFieldRow } from './Forms/InlineFieldRow';
```

Clarity-89: Good catch! Missed that one.

packages/grafana-ui/src/components/Forms/InlineFieldRow.tsx
Clarity-89: For some reason it ends up in the component's Prop table, which is confusing for the users. Could be related to https://github.com/grafana/grafana/issues/26512 maybe?

packages/grafana-ui/src/components/Forms/InlineLabel.tsx
Clarity-89: Good point! Haven' thought about that.

packages/grafana-ui/src/components/Forms/InlineField.tsx
```diff
@@ -0,0 +1,91 @@
...
+  /** Content for the label's tooltip */
+  tooltip?: PopoverContent;
+  /** Custom width for the label */
+  labelWidth?: number | 'auto';
+  /** Make the field's child to fill the width of the row. Equivalent to setting `flex-grow:1` on the field */
+  grow?: boolean;
+  /** A toggle to apply query keyword styling to the label */
+  isKeyword?: boolean;
+  /** Fill the remaining width of the row with the label's background */
+  fill?: boolean;
```
torkelo, line 19: what's the difference between grow and fill? Feels like we only need one of them?

Clarity-89: `grow` fills the rest of the row with a form element (input, etc):
Screenshot 2020-09-24 at 14 07 50
`fill` fills the rest of the row with dark background:
Screenshot 2020-09-24 at 14 07 57
But I also heard that we might be getting rid of the fill background, so I'm waiting for @jessover9000 feedback on this PR.

packages/grafana-ui/src/components/Forms/InlineField.mdx
```diff
@@ -0,0 +1,16 @@
+import { Props } from "@storybook/addon-docs/blocks";
+import { InlineField } from "./InlineField";
+
+# InlineField
+
+A basic component for rendering inline form elements together with their label. If the child element has `id` specified, the label's `htmlFor` attribute, pointing to the id, will be added.
```
jessover9000, line 6: Nice starting point. I think it's useful to list which inline form elements you could use to render with the label, and also give related component names here for 100% clarity (e.g. not just say "together with their label" but "together with `InlineLabel`", mention `InlineFieldRow`).

> If the child element has `id` specified, the label's `htmlFor` attribute, pointing to the id, will be added.
The phrasing is a bit complicated; maybe say "…specified, the component automatically gets a `htmlFor` attribute which points to the id". Also I don't yet understand why this information is important or why the label needs an attribute pointing to the id.

Maybe it's a good idea to explain how you can tweak the component to fit your needs, e.g. that you can modify the width of labels and/or fields using [list of all props that do this], what it looks like when the children prop is used, etc.

Can I have one label with multiple fields? If yes, how do I do that?

packages/grafana-ui/src/components/Forms/InlineFieldRow.mdx
```diff
@@ -0,0 +1,14 @@
+Used to align multiple inline form fields in one row. The row will wrap if the width of the children exceeds its own. Equivalent to the div with `gf-form-inline` class name.
```
jessover9000, line 1: > The row will wrap if the width of the children exceeds its own.
I kind of know what you mean but this is still a bit confusing to me. Maybe change to "If the width of the children exceeds the row width, the row will wrap." Also I guess a screenshot example of how wrapped rows look would be nice. 

You explain that a row stacks multiple fields, but you don't address using multiple rows and it's not in the code example either. I guess it's fine to just put multiple rows one after another? Also there is no props table for this component, is that on purpose?

Just like in the InlineField docs, I would list related components and how they can and should be used within this component.

packages/grafana-ui/src/components/Forms/InlineLabel.mdx
```diff
@@ -0,0 +1,15 @@
+import { Props } from "@storybook/addon-docs/blocks";
+import { InlineLabel } from "./InlineLabel";
+
+# InlineLabel
+
+A horizontal variant of Label, primarily used in query editors.
```
jessover9000, line 6: I feel like the limitations of the InlineLabel could be explained a bit more here: 
- There is no description option, if you need to add additional explanation use the tooltip variant. 
- For query editor readability, the label text should be as short as possible (4 words or less).
- Which components can the label be combined with and how does that work (`InlineField`)?

jessover9000 (commented):
I added some feedback for the docs text. I think this is a great start!

Regarding the fill and grow options, as mentioned in the last UX feedback session, I recommend getting rid of both. These options make add and remove buttons move to the very right of the query editor for many data sources, see this Elastic example:
image
On wide screens, the fill/grow principle would make the distance between the majority of query editor components and these buttons very big, and users are likely to miss elements that are too far away from their center of perceptions. The grow principle makes sense for inputs that we expect very long code input for, but should be used sparingly. (that's something to add to the docs ;) )

If we get rid of the fill principle, the only full-width elements would be the top rows separating queries. The whitespace could make it much easier to scan where your query ends and the next one starts. See screenshots for comparison:
image
image

Regarding font colors of the components, I think the white label text works better than the previous blue. I would make the labels Regular weight instead of Medium, to make the clickable parts stand out more. You are using form elements instead of the button-like field design in live Grafana, so the color tests I've done don't match 100% right now, but here's the team favourite from yesterday's UX feedback session: https://www.figma.com/file/ITCsaXtsANM17oRtU6GQS9/Component-Playground?node-id=672%3A6

packages/grafana-ui/src/components/Forms/InlineFieldRow.mdx
Clarity-89: Yeah, there's no props for this component except the native HTML ones and storybook doesn't display those for some reason.

jessover9000 (approved):
Added a small nit, otherwise looks good! 💪

PR #27669, opened 2020-09-20
Table: New cell hover behavior and image cell display mode 

### LINKED ISSUE #27761: Table Panel: Tooltip showing full content of cell on hover (if cell is too small)

A tooltip shows the complete content of a cell and appears while hovering above a cell in the Table panel (if the cell itself is too small to render the complete value and cuts it off instead). Additionally it should be possible to mark the content of the tooltip and copy it.

It should work just like the already existing Table tooltips:

image

**Why is this needed**:

I want to visualize the label values of certain metrics in a table. Some of these values are quite long. And so they are cut off. Most of the time I want this behavior, but sometimes I need the complete value for example to investigate logs. To get the complete value I have to manually increase the column width until the complete value is visible. This is a bit cumbersome. In addition this changes the dashboard by introducing an override. So to reset everything, I have to reload the dashboard without saving or manually remove the override again.

image

Having a tooltip show up would help with that

torkelo:
Just wanted to make a small weekend fun feature and one thing led to another and whoops, refactored a large part of the table component 

To make a new hover behavior that show full content on hover work consistently with different cell backgrounds and styles required big change to styles. 

Before:
Cell (absolute pos & width & border, 100% height)
  Cell inner (padding & background color, & text overflow elipsis, 100& height & width, text alignmenent with text-align)

After:
Cell (absolute pos, width & border & background color if any, 100% height, content alignment with justify-content ) 
  Cell text (only text-overflow ellipsis) 

This made it possible to have on hover cell css that extended width to auto adjust to content with the preserved background. To make this possible required a big rethink of cell style generation and sharing. There are also some optimizations here so we do not call displayProcessor twice per cell (was done before for colored & background cells). 

Consolidates more logic in the DefaultCell. 

This also will make the next table option I want to introduce easier (cell size = small | medium | large). 

table_cell_hover

Fixes #27761


packages/grafana-ui/src/components/Table/styles.ts
```diff
@@ -1,140 +1,163 @@
...
-    };
-  }
-);
+      'cell-filter-actions'
+    ),
+    filterItem: css`
+      label: filterItem;
+      cursor: pointer;
+      padding: 0 ${theme.spacing.xxs};
+    `,
```
mckn, line 159: I would move these to filterActions.

packages/grafana-ui/src/components/Table/styles.ts
```diff
@@ -1,140 +1,163 @@
...
+      right: -4px;
+      border-radius: 3px;
+      top: 0;
+      z-index: ${theme.zIndex.dropdown};
+      touch-action: none;
+
+      &:hover {
+        opacity: 1;
+      }
+    `,
```
mckn, line 143: I would move this to table.tsx

packages/grafana-ui/src/components/Table/styles.ts
```diff
@@ -1,140 +1,163 @@
...
+      label: row;
+      border-bottom: 1px solid ${borderColor};
+
+      &:hover {
+        background-color: ${rowHoverBg};
+      }
+    `,
+    imageCell: css`
+      height: 100%;
+    `,
```
mckn, line 123: I would move this to imageCell.tsx

packages/grafana-ui/src/components/Table/styles.ts
```diff
@@ -1,140 +1,163 @@
...
+      cursor: pointer;
+    `,
+    row: css`
+      label: row;
+      border-bottom: 1px solid ${borderColor};
+
+      &:hover {
+        background-color: ${rowHoverBg};
+      }
+    `,
```
mckn, line 120: I would move this to table.tsx.

packages/grafana-ui/src/components/Table/styles.ts
```diff
@@ -1,140 +1,163 @@
...
+      overflow: hidden;
+      text-overflow: ellipsis;
+      user-select: text;
+      white-space: nowrap;
+      text-decoration: underline;
+    `,
+    headerFilter: css`
+      label: headerFilter;
+      cursor: pointer;
+    `,
```
mckn, line 112: I would move this to filter.tsx

mckn (approved):
I think it looks really good and works as expected! Great work!

The only thing that I would do is to split up the `styles.ts` and move the styling to where it is used. I have not looked through all of the styling in `styles.ts` so not sure if it is possible. But I would basically just keep the styling that is truly shared between multiple components in that file and move the rest.

This might be a matter of taste but when keeping one styling file it tends to grow and be a bit harder to maintain in the long run.

packages/grafana-ui/src/components/Table/styles.ts
torkelo: there are a lot of variables that are shared in all the table styles, do not want to duplicate them in multiple places  (like rowHoverBg, is used in cell style)

mckn: Makes sense! Missed that one. So then I would categorized those into the shared category and keep them in the `styles.ts`.

mckn: Another approach that could be worth thinking about would be to move the shared colors etc. to the theme. Then it is also reusable for plugins if they want to extend the table cells in some way. But not sure it is a valid option.

torkelo:
I find having all the table styles in the same place is helpful as they are all part of the same big component and need to work together.

packages/grafana-ui/src/components/Table/styles.ts
torkelo: Yes, that could be a good idea. Most are high-level colors from the theme already, and trying to limit component variables in the theme. But if we had another use of these in other components or plugins then that would be good idea

MiguelAngelLV:
I try image cell. It works fine with value from query, but it not use the mapping value, so you need change the value directly in the query.

By other hand, if you use `data:image/svg+xml,SVG_URL_ENCONDED` works fine

Example

```
data:image/svg+xml,%3Csvg xmlns="http://www.w3.org/2000/svg" style="width:24px;height:24px" viewBox="0 0 24 24"%3E%3Cpath fill="green" d="M16.56,5.44L15.11,6.89C16.84,7.94 18,9.83 18,12A6,6 0 0,1 12,18A6,6 0 0,1 6,12C6,9.83 7.16,7.94 8.88,6.88L7.44,5.44C5.36,6.88 4,9.28 4,12A8,8 0 0,0 12,20A8,8 0 0,0 20,12C20,9.28 18.64,6.88 16.56,5.44M13,3H11V13H13" /%3E%3C/svg%3E
```

Maybe a option for SVG Image to set directly SVG as value can be very usefull. You can copy and paste the SVG icon from Material Design Icons directly and use instead upload to the server, incluse change the color with fill attribute.

PR #27672, opened 2020-09-21
Live: improved channel state

ryantxu:
This is an alternative to smaller steps taken in #27630

This adds more structure to the live channels and forces the paths to be:
${scope}/${namepsace}/${path}

scope = grafana | datasource | plugin

namespace: 
when 'grafana' = features
when 'datasource' = it is datasource name|id
when 'plugin' = it is the pluginId

This creates a channel object where you can request the status or stream as an observable.


packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,158 @@
...
+   * flag if the channel is activly connected to the channel.
+   * This may be false while the connections get established or if the network is lost
+   * As long as the `shutdown` flag is not set, the connection will try to reestablish
+   */
+  connected: boolean;
+
+  /**
+   * Indicate that the channel will not recived any updates and should be discarded
+   */
+  shutdown?: any;
```
hugohaggmark, line 75: Would it be simpler to use an enum here instead? enum ConnectionState: { NotConnected, Connected, Shutdown/Closed, Error }

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Get the channel status
+   */
+  getStatus: () => Observable<LiveChannelStatus>;
+
+  /**
+   * Get the stream of events and
+   */
+  getStream: () => Observable<TMessage>;
```
hugohaggmark, line 123: This looks ok, but I wonder if it would be even better from the consumers perspective if this was only one stream and on that stream you could have both ConnectionState and data/message/error.

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Indication of the presense indicator.
+   *
+   * NOTE: This feature is supported by a limited set of channels
+   */
+  getPresense?: () => Observable<LiveChannelPresense>;
```
hugohaggmark, line 130: Hmm, I would like to incorporate this in the stream I mentioned above to simplify for the consumer but that would mean we need different kinds of messages passing through.

some raw pseudo code:
interface LiveMessage<T =  TMessage | PresenseMessage | ErrorMessage> {
 state: ConnectionState,
 type: 'message' | 'presense' | 'error'
 message: T
)

hugohaggmark (commented):
Great work on this, left some points on the API.

ryantxu:
@hugohaggmark -- i incorporated your advice, I think it is getting better!

Rather than: having the stream be `TMessage | LiveChannelStatus | LiveChannelJoinLeave` I made it:

```
export interface LiveChannelMessage<TMessage = any> {
  type: 'status' | 'message' | 'join' | 'leave';
  message: TMessage | LiveChannelStatus | LiveChannelJoinLeave;
}
```
this way it has a well known field to switch on to know what type of message it is

public/app/features/admin/LiveAdmin.tsx
```diff
@@ -99,19 +178,36 @@ export class LiveAdmin extends PureComponent<Props, State> {
             <br />
           </Container>
 
-          <h2>Channels</h2>
-          <Select options={channels} value={current} onChange={this.onChannelChanged} allowCustomValue={true} />
-          <br />
-
-          <LivePanel channel={channel} />
-
+          <div
```
ryantxu, line 181: NOTE:  this page will be thrown out before it is actually released, so it is really just for debugging

hugohaggmark:
I'll have a look later today, super busy with ADX

hugohaggmark:
> @hugohaggmark -- i incorporated your advice, I think it is getting better!
> 
> Rather than: having the stream be `TMessage | LiveChannelStatus | LiveChannelJoinLeave` I made it:
> 
> ```
> export interface LiveChannelMessage<TMessage = any> {
>   type: 'status' | 'message' | 'join' | 'leave';
>   message: TMessage | LiveChannelStatus | LiveChannelJoinLeave;
> }
> ```
> 
> this way it has a well known field to switch on to know what type of message it is

Yepp looks great 🌟

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,172 @@
...
+  Pending = 'pending',
+  /** Connected to the channel */
+  Connected = 'connected',
+  /** Disconnected from the channel.  The channel will reconnect when possible */
+  Disconnected = 'disconnected',
+  /** Was at some point connected, and will not try to reconnect */
+  Shutdown = 'shutdown',
+  /** Channel configuraiton was invalid and will not connect */
+  Invalid = 'invalid',
+}
```
hugohaggmark, line 62: So great with the comments here, awesome Ryan!

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export interface LiveChannelMessage<TMessage = any> {
+  type: 'status' | 'message' | 'join' | 'leave';
+  message: TMessage | LiveChannelStatus | LiveChannelJoinLeave;
+}
+
+/**
+ * @experimental
+ */
+export interface LiveChannel<TMessage = any, TPublish = any> {
```
hugohaggmark, line 115: nit: I think <TMessage = any> is very generous but I understand this is still early stages. Maybe we should have a base interface you can extend?

packages/grafana-data/src/types/live.ts
```diff
@@ -0,0 +1,172 @@
...
+   * Watch all events in this channel
+   */
+  getStream: () => Observable<LiveChannelMessage<TMessage>>;
+
+  /**
+   * For channels that support presense, this will request the current state from the server.
+   *
+   * Join and leave messages will be sent to the open stream
+   */
+  getPresense?: () => Promise<LiveChannelPresense>;
```
hugohaggmark, line 144: getStream & getPresense should be merged I think. Or is there any good reason that I've to subscribe to both?

hugohaggmark (approved):
This is looking really great! I just left some small nits.

packages/grafana-data/src/types/live.ts
ryantxu: Let me try a few more implementations and see if there is a reasonable base class.  Given that each channel has its own scope, it may make sense to give it full control of the object

packages/grafana-data/src/types/live.ts
ryantxu: maybe -- in the centrifuge API, getPresense is an async call that gets the current state, and then the join/leave messages are deltas.  

We could potentially wrap that all and return a full managed list of users on join/leave... but again, I am not totally sure what pattern makes the most sense until we try it :)

PR #27680, opened 2020-09-21
Forms: Increase default form width from 400 to 600

torkelo:
Feel like many forms are to narrow, so increased the width to 600. 

The enterprise reporting form could need some tweaks so that not all inputs grow to full width. 

Before: 
Screenshot from 2020-09-21 15-31-52

After:
Screenshot from 2020-09-21 15-32-00


peterholmberg (approved):
Nice!

Issue #24256, opened 2020-05-04
Cloudwatch unable to create datasource using credential file


niraj8241:
**What happened**:
Trying to configure cloudwatch datasource on an on-prem system. Have installed AWS CLI and configure the required configs. I can access the services fine using multiple profiles that I have setup. But when I choose the Grafana's Cloudwatch Datasource and use the "Credential File" option I see that I am unable to get it to work and get the below error in logs:

`Failed to call cloudwatch:ListMetrics, NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"`

**What you expected to happen**:
I was expecting that I can configure cloudwatch on my system and create different datasources based on multiple profiles I have setup on my system.

**How to reproduce it (as minimally and precisely as possible)**:
1. Setup Grafana
2. Configure AWS CLI.
3. Configure multiple profiles
4. Choose Credential File Option in Cloudwatch Datasource
5. Specify the profile name to see if it works.

**Anything else we need to know?**:
I am using assume role when I configured AWS CLI. The default profile has a credential associated with it but the resulting profiles are assume roles in AWS. Something like below.

> [profile default]
> output = table
> region = us-east-1

> [profile prod]
> role_arn = arn:aws:iam::xxxxxxxxxx:role/Administrator
> output = json
> region = us-east-1
> source_profile = default

**Environment**: Production
- Grafana version: Grafana v6.7.2 (423a25fc32)
- Data source type & version: Cloudwatch
- OS Grafana is installed on: Ubuntu 18.04
- User OS & Browser: Windows 10 & Chrome
- Grafana plugins: NA
- Others: NA

dprokop:
@briangann can you please take a look at this one?

bergquist:
@niraj8241 We fixed a similar issue in 7.0.0. Could you please try upgrading and see if the problem remains?

daniellee:
No response from @niraj8241 so closing this issue. Can reopen if this is still an issue.

PR #27684, opened 2020-09-21
CloudWatch: Fix loading custom credentials profile

papagian:
I configure a cloudwatch datasource using a custom (non default) credential profile. I can save and test successfully the datasource but when I reopen it the custom credential profile is not loaded (although it is included in the respective API for getting the datasource details)

This information is not saved in datasource JSON data (as it is expected) but in the datasource [`database`](https://github.com/grafana/grafana/blob/43ef052d5711f9b4c0447b052f25642f965e8a2c/pkg/tsdb/cloudwatch/cloudwatch.go#L299) field.

This fix stores it in JSON settings and handles the migration between old/new field.

Screenshot 2020-09-21 at 18 45 55
Screenshot 2020-09-21 at 18 46 46

Fixes #

I have tested provisioning with the old and the new field as well like:
```
  - name: cloudwatch-custom-profile-legacy
    type: cloudwatch
    editable: true
    database: other
    jsonData:
      authType: credentials
      defaultRegion: us-east-2

  - name: cloudwatch-custom-profile
    type: cloudwatch
    editable: true
    jsonData:
      authType: credentials
      defaultRegion: us-east-2
      profile: other
```

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
+++ b/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
@@ -151,7 +151,7 @@ export class ConfigEditor extends PureComponent<Props, State> {
                   <Input
                     className="width-30"
                     placeholder="default"
-                    value={options.jsonData.database}
+                    value={options.database}
                     onChange={onUpdateDatasourceOption(this.props, 'database')}
                   />
                 </div>
--- a/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
+++ b/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
@@ -114,7 +114,7 @@ export class ConfigEditor extends PureComponent<Props, State> {
     const { regions } = this.state;
     const { options } = this.props;
     const secureJsonData = (options.secureJsonData || {}) as CloudWatchSecureJsonData;
-    var profile = options.jsonData.profile;
+    let profile = options.jsonData.profile;
     if (!profile) {
       profile = options.database;
     }
--- a/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
+++ b/public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
@@ -4,7 +4,6 @@ const { Select, Input } = LegacyForms;
 import {
   DataSourcePluginOptionsEditorProps,
   onUpdateDatasourceJsonDataOptionSelect,
-  onUpdateDatasourceOption,
   onUpdateDatasourceResetOption,
   onUpdateDatasourceJsonDataOption,
   onUpdateDatasourceSecureJsonDataOption,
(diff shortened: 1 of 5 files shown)


marefr (approved):
LGTM :+1:

aknuds1 (approved):
Strange that the field would be "database", but good job.

marefr:
Maybe related to #24256? Probably a good idea to add a fallback, but not optimal having to maintain that logic over long time.

papagian:
> Maybe related to #24256? Probably a good idea to add a fallback, but not optimal having to maintain that logic over long time.

I think that the root cause of #24256 is not this issue because this seems to be only in the frontend
(the backend [reads](https://github.com/grafana/grafana/blob/43ef052d5711f9b4c0447b052f25642f965e8a2c/pkg/tsdb/cloudwatch/cloudwatch.go#L299) this information from `DataSource.Database` and stores it in the `datasourceInfo.Profile` that is used by `getCredentials()`)

papagian:
> Strange that the field would be "database", but good job.

I suspect that initially `database` field was used because the [Datasource](https://github.com/grafana/grafana/blob/4bb3f66569b6650e02fe68adb802cf93581c6d2d/pkg/models/datasource.go#L44) struct does not have a relevant field for cloudwatch profiles and there is no database concept in cloudwatch to conflict with.
I agree that having this information under the JSON settings it would be better.

marefr:
Using the database field is probably a legacy thing

ivanahuckova (approved):

papagian:
@marefr @aknuds1 @ivanahuckova 
I have pushed the changes by storing the profile in the JSON therefore I've requested your reviews

public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
```diff
@@ -115,6 +114,10 @@ export class ConfigEditor extends PureComponent<Props, State> {
     const { regions } = this.state;
     const { options } = this.props;
     const secureJsonData = (options.secureJsonData || {}) as CloudWatchSecureJsonData;
+    var profile = options.jsonData.profile;
+    if (!profile) {
+      profile = options.database;
```
aknuds1, line 119: Should this be `options.jsonData.database`, since that's what the old code does?

aknuds1 (commented):
Have some questions about the TS code.

public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
papagian: that was actually the error: the old code used `options.jsonData.database` instead of `options.database`.

ivanahuckova (approved):
LGTM! Great job Sofia.

public/app/plugins/datasource/cloudwatch/components/ConfigEditor.tsx
aknuds1: @papagian Ah... thanks for clarifying that!

aknuds1 (approved):
LGTM!

marefr (approved):
LGTM

PR #27687, opened 2020-09-21
@grafana/e2e: flaky chart data workaround

stevenvachon:
See commit message(s).

### DIFF
--- a/packages/grafana-e2e/src/flows/configurePanel.ts
+++ b/packages/grafana-e2e/src/flows/configurePanel.ts
@@ -127,15 +127,20 @@ export const configurePanel = (config: PartialAddPanelConfig | PartialEditPanelC
     // @todo instead wait for '@pluginModule'
     e2e().wait(2000);
 
-    if (!isExplore) {
+    // There is no usable data when the query is empty,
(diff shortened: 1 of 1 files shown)


ryantxu (approved):
Is this after a query is added?  if so yes, the only way to assert that the query is requested is to hit the refresh icon

stevenvachon:
Before the query is added

packages/grafana-e2e/src/flows/configurePanel.ts
```diff
@@ -127,15 +127,20 @@ export const configurePanel = (config: PartialAddPanelConfig | PartialEditPanelC
...
     e2e().wait(2000);
 
-    if (!isExplore) {
+    // There is no usable data when the query is empty,
+    // and Cypress had an issue with some plugins where the request wasn't noticed, despite occuring in manual tests
+    if (isEdit) {
       // Avoid cache flakiness (where @chartData isn't requested)
+      // @todo this may not be necessary anymore
       e2e()
         .get('.refresh-picker-buttons .btn')
```
hugohaggmark, line 136: Friendly reminder to use selectors instead :)

stevenvachon: I have not forgotten :)

PR #27689, opened 2020-09-21
Docs: 7.2 filter options

oddlittlebird:
Updates field options documentation for 7.2 features. DO NOT MERGE UNTIL 9/23!

- Adds table filter feature
- Expands table display options
- Moves a bunch of content into smaller topics in the Field options folder
- Updates What's new in 7.2


docs/sources/panels/field-options/filter-table-columns.md
```diff
@@ -0,0 +1,39 @@
...
+## Turn on column filtering
+
+1. In Grafana, navigate to the dashboard with the table with the columns that you want to filter.
+2. On the table panel you want to filter, [open the panel editor]({{< relref "../panel-editor.md#open-the-panel-editor" >}}).
+3. Click the **Field** tab.
+4. In Table options, turn on the **Column filter** option.
+
+A filter icon appears next to each column title.
+
+{{< docs-imagebox img="/img/docs/tables/column-filter-with-icon.png" max-width="500px" caption="Column filtering turned on" class="docs-image--no-shadow" >}}
```
hugohaggmark, line 23: There is just one thing, somewhere out there there are users that were using the old table panel from Grafana <7. When they upgrade to 7 I think the Table panel is still the old one so they would need to actively choose the "new" Table panel to get this feature. Not sure this should be mentioned here but I can see that this can become confusing if your still using the "old" Table panel.

hugohaggmark (approved):
Looks great, outstandind work on this Diana 🌟

torkelo (changes requested):
Some great changes! But some I think are less great. 

Not sure we need to chop the docs up so much and spread out the information in small chunks everywhere. It has it pros & cons and can make it harder for a user to find the information they are looking for (or requiring traversing multiple pages). 

If I want to know how to setup display options for the table and I expect them to be listed and described on the table visualization reference page. Every visualization will in the future get some field-level options and so separating the field level display option into a separate page under a different section makes little sense to me. The coming graph panel will have around 10 field-level display options, should that get 10 pages? 

I would like to see table field options moved back to the table visualization page so it documents all the table display options. I do not think the table column filter option needs its own page.

docs/sources/panels/field-options/filter-table-columns.md
oddlittlebird: I have not seen anything regarding choosing a table type. @dprokop, does this ring a bell?

docs/sources/menu.yaml
```diff
@@ -167,6 +167,19 @@
...
+        - link: /panels/field-options/configure-all-fields/
+          name: Configure all fields
+        - link: /panels/field-options/configure-specific-fields/
+          name: Configure specific fields
+        - link: /panels/field-options/standard-field-options/
+          name: Standard field options
+        - link: /panels/field-options/table-field-options/
+          name: Table field options
+        - link: /panels/field-options/filter-table-columns/
+          name: Filter table columns
```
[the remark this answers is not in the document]
oddlittlebird: The reason it has its own page now is that according to @hugohaggmark, the current functionality is going to be expanded in the near future.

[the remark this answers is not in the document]
oddlittlebird: However, even as it stands, this standalone task is plenty long enough to justify a separate topic.

oddlittlebird:
> The coming graph panel will have around 10 field-level display options, should that get 10 pages?

If the field options result in tasks as powerful as filtering table columns, yes. That might ten pages worth of content.

When I think about how to document something, I try to break it down in terms of tasks that the user wants to perform, and part of why we break content up is to make it easier for the user to find it if, say, they search for the task they want to perform with a web search.

What we want to avoid is the monstrous everything-but-the-kitchen sink pages that contain absolutely everything. Those are not user-friendly. Ideally the documentation would be small, bite-sized pieces, that are a concept (in this case, what a table is and why I might want to use it), a reference (list of table options with definitions), or a task (instructions for filtering table columns).

docs/sources/menu.yaml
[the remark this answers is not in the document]
torkelo: Don't agree, column filters are part of task of configuring the table visualization. Not a seperate task or workflow. The table visualization page is the reference page for all table options, so do not understand why this option is not part of that page.

torkelo (changes requested):
Still think the table field options belong to the table visualization docs, and do not think column filter is a big enough feature to or task to deserve it's own page.

achatterjee-grafana:
Agree with @oddlittlebird about chunking information into task-based topics, but I not familiar with the conventions at Grafana yet. It will be useful if we come up with some guidelines with regard to documentation conventions.

torkelo:
I think this is breaking things up too much, breaking single tasks into mini sub-tasks that get their own page and spreads them out into different sections, instead of having them logically grouped (on the same page). Making information harder to find.

oddlittlebird:
> I think this is breaking things up too much, breaking single tasks into mini sub-tasks that get their own page and spreads them out into different sections, instead of having them logically grouped (on the same page). Making information harder to find.

That depends on how you are looking. If you are browsing, then you might be correct. However, if you are searching, then small chunks of information are preferable. This organization conforms to technical documentation best practices and task-based organization.

oddlittlebird:
This is ready for another review. I moved the table-specific topics into the table section, so that they are more closely associated with the table concept topic.

Eventually, all visualizations should have task and reference topics specific to them nested underneath, which means we'll want to move the Visualizations higher up in the documentation hierarchy, but that is out of scope for this PR.

docs/sources/menu.yaml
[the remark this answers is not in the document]
oddlittlebird: Because making huge topics that hold everything related to one subject is not good documentation. Long pages are tiring to read through, difficult to navigate, and they are not friendly to web searches.

The best practice is to have task-based documentation, and when possible, to chunk information into manageable concept, task, and reference pieces.

docs/sources/panels/visualizations/table/_index.md
```diff
@@ -0,0 +1,34 @@
...
+- [Field options and overrides]({{< relref "../../field-options/_index.md" >}})
+- [Thresholds]({{< relref "../../thresholds.md" >}})
+
+## Display options
+
+- **Show header -** Show or hide column names imported from your data source.
+- **Sort ascending/descending -** Click a column title to change the sort order from default to descending to ascending. Each time you click, the sort order changes to the next option in the cycle. You can only sort by one column at a time.
+  {{< docs-imagebox img="/img/docs/tables/sort-descending.png" max-width="500px" caption="Sort descending" class="docs-image--no-shadow" >}}
+- [Table field options]({{< relref "table-field-options.md" >}}) allow you to change [field options]({{< relref "../../field-options/_index.md" >}}) such as column width, alignment, and cell display mode.
+- [Filter table columns]({{< relref "filter-table-columns.md" >}}) - You can temporarily change how column data is displayed. For example, you can order values from highest to lowest or hide specific values.
```
torkelo, line 34: I would like to still make these options just part of this page, they are all part of the same task (configuring table display options). Not having them on the table documentation page feels like a missed opportunity.

oddlittlebird: Two documentarians out of two surveyed disagree. They are linked from the page, but they should be separate topics.

oddlittlebird: Between the two of us, @achatterjee-grafana and I have close to three decades of experience with technical documentation. There is a method to what we do and how we do it.

hugohaggmark: Is there some common ground or compromise that we could discuss as a way forward here?

torkelo: I respect your experience in technical documentation, I really do. But I think some domain knowledge is also required in order to logically group information. 

The current table documentation page: https://grafana.com/docs/grafana/latest/panels/visualizations/table-panel/ 

is far from a long page and in need of splitting up. The main purpose of this page is to list and document the different display options for the table panel. Not including the main display options (display mode) and hiding this on a sub-page is not very user friendly. Scrolling a tiny bit is much easier than finding a cryptic link "Table field options". Each page has a TOC on the side as well for easy navigation within larger documents. Each section heading will be indexed so they can be directly discovered via search. I see mostly downsides by splitting this page up. 

Many visualizations will have field display options, some will only have field display options. Splitting display options up in two pages feels ad-hoc and confusing. I know these options are split into two tabs in the UI, but that is to me an argument to document them together so it's clear that there are display options in the field/override tabs as well.

oddlittlebird: Any pattern we apply should be a pattern we can apply to all the visualizations. True, table is short. Many of the other visualizations are not.

A longer term goal is to expand all of the "main" visualization topics with more concept information. Things like why you would want to use this visualization, which data is best suited for it, examples, and so on. Also, the current field description is sadly lacking.

My goal when I rewrote the Panels section for 7.0 was:
- Each visualization has a page.
- Each field is defined.

That was a pretty low bar, but it was one that we were not meeting at that time. There is a lot more work to do, on all of them.

Your suggestion of putting all the display options into the main Table section might work in the short term, even if it would be contrary to what we know to be good documentation practices. However, it would require more rework in the future when we add the "deep dives" and expanded guidance that this section needs.

Regarding the purpose of the page, the panel display fields would ideally be in a different page and have expanded descriptions and explanations for all the visualizations, including table. We don't have the bandwidth for that right now, but it is on the To Do list.

docs/sources/panels/field-options/filter-table-columns.md
daniellee: @oddlittlebird you can't choose the old table panel from the visualization options but if you are using a dashboard that was created in a previous version of Grafana then it is not automatically upgraded to the new table panel. An example on play: https://play.grafana.org/d/000000059/influxdb-table?editPanel=6&orgId=1 If you go to the visualization options and choose Table then you will see that the panel changes. I agree with Hugo, it is confusing and non-obvious.

oddlittlebird: ...omg, I see what you mean. I clicked the Table visualization and I suddenly lost a bunch of options. I'll try to figure out how to explain that.

oddlittlebird: Updated. Please take a look.

docs/sources/panels/visualizations/table/_index.md
torkelo: Ok, fine. Just wished to have the display options listed together. If we can mention cell display mode and show examples on the main page later that would be great. I would like the main visualization page to be a source of inspiration and show different settings & examples. So documenting the settings on another page could be a good idea. Just splitting up display options and field options feel bad, they are both display options just different levels. 

Tested this PR but the side menu navigation did not work, all the links are wrong 
Screenshot from 2020-09-23 17-47-24

oddlittlebird: ARGH, thought I had fixed all the merge issues. Will check the side panel momentarily.

The longer term plan is to have the main page for each visualization have what you are talking about:
- What the visualization is
- How to use it, what data is best suited to it
- Use cases and examples of each visualization used well
- Links to tasks (such as Add and set up a table, which we currently don't have)
- Links to reference documentation. The list of fields and descriptions that are currently in the visualization topics is woefully incomplete. Most of them need to be expanded and possibly illustrated with images, explanations, and use cases. 

So...yeah. We have plans. Just trying to get pieces into place as we touch them right now, laying the groundwork for the future.

oddlittlebird: Heh. Whoops, that menu thing wasn't a merge problem, was a me problem. It looked right when I tested it before...oh, well. Fixing now.

docs/sources/menu.yaml
```diff
@@ -199,6 +208,13 @@
           name: Stat
         - link: /panels/visualizations/table-panel/
           name: Table
+          children:
+          - name: Overview
+            link: /panels/table/
+          - link: /panels/table/table-field-options/
```
torkelo, line 214: Missing word visualizations

torkelo (commented):
Added a comment on why the menu links don't work

docs/sources/menu.yaml
oddlittlebird: That should have thrown an error. WTH. Working on the fix now. Thank you for spotting these!

oddlittlebird: @robbymilo is hopefully going to make menu.yaml go away in the not-distant future, so probably isn't worth figuring out why it's not giving me errors.

oddlittlebird: Menu fix pushed.

torkelo (approved):

PR #27691, opened 2020-09-22
POC: load dashboard json from a plugin

ryantxu:
There have been many recent discussions about dynamic data-driven dashboards.  In particular, @sdboyer is looking at how to evolve mixins into something more.  

I think we need to expose some method for plugins to develop dynamic dashboards ASAP, and figure out how this will integrate into a unified search later.

This POC adds a new dashboard type 'plugin' that passes the rest of the URL path to a plugin that can then load dashboard json.  This is currently a frontend plugin with a hidden interface -- long term, I think this should be in the gRPC contract and managed in the backend plugin... BUT we should not block progress waiting for the more complicated path :)

In this PR, the URL:
http://localhost:3000/dashboard/p/dashgen/path/to/whatever?orgId=1

asks the `dashgen` plugin to load a dashboard for `path/to/whatever` -- in this case a super simple hardcoded dashboard:

image


public/app/plugins/panel/dashgen/loader.ts
```diff
@@ -0,0 +1,96 @@
+import { DashboardLoaderSupport } from 'app/features/dashboard/types';
+import { DashboardDTO } from 'app/types';
+import { of } from 'rxjs';
+
+export const loader: DashboardLoaderSupport = {
+  loadDashboard: (path: string) => {
+    console.log('LOAD', path);
+    return of(superSimple as DashboardDTO);
```
ryantxu, line 8: @sdboyer -- this example just loads a static definition, but if you have a simple HTTP server running we can load that dynamically

sdboyer:
Brilliant, thank you for taking first steps here! I definitely agree that standardizing on some approaches to programatically generated dashboards has enormous potential.

> and figure out how this will integrate into a unified search later.

For my edification - when you say "unified search," you're referring [dashboard search](https://grafana.com/docs/grafana/latest/dashboards/search/), correct?

ryantxu:
> when you say "unified search," you're referring dashboard search, correct?

yes.  If we have an extendable way to generate dashboards, I think we still need them in the search results

ryantxu:
Just a POC... will close for now

PR #12459, opened 2018-06-29
Avoid calling os.Exit outside main.go

bergquist:
This is an alternative fix to #12439

We merged #12439 and included it in 5.2.1 since it solved
the problem but I think this is a prefered solution. IMO main.go
should be responsible for start/exiting the server and should, at
some point be able to restart grafana server graceful. So calling
os.Exit within Grafana server will cause problems in the future.

PR #27700, opened 2020-09-22
Server: Don't clutter trace.Stop by exiting

aknuds1:
Fix a minor issue that was pointed out by the latest version of gocritic, i.e. that the server clutters a deferred call to `trace.Stop` by calling `os.Exit` in the same function. I've solved it by moving the deferred call into a utility function, and calling `os.Exit` in the caller instead.

### DIFF
--- a/pkg/cmd/grafana-server/main.go
+++ b/pkg/cmd/grafana-server/main.go
@@ -38,7 +38,6 @@ import (
 var version = "5.0.0"
 var commit = "NA"
 var buildBranch = "master"
-
 var buildstamp string
 
 type exitWithCode struct {
--- a/pkg/cmd/grafana-server/main.go
+++ b/pkg/cmd/grafana-server/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"net/http"
@@ -33,10 +34,20 @@ import (
 	_ "github.com/grafana/grafana/pkg/tsdb/testdatasource"
 )
 
-var version = "5.0.0"
-var commit = "NA"
-var buildBranch = "master"
+const version = "5.0.0"
+const commit = "NA"
+const buildBranch = "master"
+
 var buildstamp string
+
+type exitWithCode struct {
+	reason string
+	code   int
+}
+
+func (e exitWithCode) Error() string {
+	return e.reason
+}
 
 func main() {
 	var (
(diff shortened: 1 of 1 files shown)


pkg/cmd/grafana-server/main.go
```diff
@@ -82,6 +93,23 @@ func main() {
...
+		if code != 0 {
+			fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+		}
+
+		os.Exit(code)
+	}
+}
+
+func executeServer(configFile, homePath, pidFile, packaging string, traceDiagnostics *tracingDiagnostics) error {
+	defer log.Close()
```
papagian, line 111: I wonder if you have to call `trace.Stop()` in a defer statement as well. It is deleted from [here](https://github.com/grafana/grafana/pull/27700/files#diff-273afb154db0a7320bbd76f27cd95102L130) but it's not clear to me why [it was introduced](https://github.com/grafana/grafana/pull/12459/files#r199133065)

aknuds1: @papagian Yeah, true, it was also super unclear to me if `trace.Stop` should be called independently of `traceDiagnostics.enabled` or not.

aknuds1: AFAICT, `trace.Start` is only called if traceDiagnostics.enabled is true. @bergquist - what's your take?

marefr (changes requested):
See comment

pkg/cmd/grafana-server/main.go
```diff
@@ -33,11 +34,21 @@ import (
 	_ "github.com/grafana/grafana/pkg/tsdb/testdatasource"
 )
 
-var version = "5.0.0"
-var commit = "NA"
-var buildBranch = "master"
+const version = "5.0.0"
+const commit = "NA"
+const buildBranch = "master"
```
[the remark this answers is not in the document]
aknuds1: @marefr How does build.go override them? Can you point it out please? At least with this change, it still compiles.

pkg/cmd/grafana-server/main.go
marefr: It doesn't hurt calling Stop, but shouldn't be needed if not enabled. From docs `Stop stops the current tracing, if any. Stop only returns after all the writes for the trace have completed.`

pkg/cmd/grafana-server/main.go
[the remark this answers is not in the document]
aknuds1: Via `-X main.buildBranch=master` etc? I have to look into how that build flag works.

pkg/cmd/grafana-server/main.go
aknuds1: @marefr Yeah, I saw the same in the docs. It would be nice to have some clarity though, as to whether `trace.Stop` is only necessary when `traceDiagnostics.enabled` is true.

pkg/cmd/grafana-server/main.go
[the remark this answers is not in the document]
aknuds1: @marefr Thanks for the heads up! I've confirmed that the `-X` link flag requires variables and can't override consts.

pkg/cmd/grafana-server/main.go
marefr: My conclusion is it's not needed. Think that extra trace,Stop at the end before exit is maybe because the defer trace.Stop doesn't execute after exit?

aknuds1: @marefr Could you approve the PR then? :)

marefr (approved):
LGTM

PR #27704, opened 2020-09-22
Chore/fix lint issues

aknuds1:
Fix linting issues pointed out by latest version of golangci-lint:

* Space before comment (`// comment`)
* Use `strings.ReplaceAll(old, new)` instead of `strings.Replace(old, new, -1)`
* Some simplifications, such as removal of unnecessary type casts

Shouldn't have any functional effect at all, but allows upgrading of golangci-lint and cleans up the code a bit.

If you want, I can split into smaller PRs.

### DIFF
--- a/pkg/services/sqlstore/annotation.go
+++ b/pkg/services/sqlstore/annotation.go
@@ -23,9 +23,7 @@ func validateTimeRange(item *annotations.Item) error {
 		item.Epoch = item.EpochEnd
 	}
 	if item.EpochEnd < item.Epoch {
-		tmp := item.Epoch
-		item.Epoch = item.EpochEnd
-		item.EpochEnd = tmp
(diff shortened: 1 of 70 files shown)


pkg/services/sqlstore/annotation.go
```diff
@@ -23,9 +23,7 @@ func validateTimeRange(item *annotations.Item) error {
 		item.Epoch = item.EpochEnd
 	}
 	if item.EpochEnd < item.Epoch {
-		tmp := item.Epoch
-		item.Epoch = item.EpochEnd
-		item.EpochEnd = tmp
+		item.Epoch, item.EpochEnd = item.EpochEnd, item.Epoch
```
wbrowne, line 26: Wowza

wbrowne (approved):
👍

PR #27706, opened 2020-09-22
PanelEditor: Prevents adding transformations in panels with alerts

### LINKED ISSUE #25681: Transformations and Alerts Incompatibility
You should not be able to add transformations to a graph panel that already has an alert definition. 

Grafana stop users from adding an alert rule if you have transformations but not the other way around

hugohaggmark:
Before:
image

After:
image

When there is already transformations and alerts in a panel model
image

Fixes #25681


mckn (approved):
LGTM and works as expected!

PR #27708, opened 2020-09-22
Docs: Updated table docs, moved annotations notice down

torkelo:
Just noticed this notice at the top of this page, it felt out of place and distracting. So moved it down to a new section at the bottom. 

Screenshot from 2020-09-22 16-06-11

### DIFF
--- a/docs/sources/panels/visualizations/table-panel.md
+++ b/docs/sources/panels/visualizations/table-panel.md
@@ -37,8 +37,6 @@ In the **Field** tab you can set table specific display options that will affect
 
 ### Column alignment
 
-This custom field option applies only to table visualizations.
-
 Choose how Grafana should align cell contents:
 
 - Auto (default)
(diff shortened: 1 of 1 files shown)


achatterjee-grafana (approved):
LGTM

oddlittlebird (changes requested):
The note does not need its own h3. Please leave it at or near the top. This was added because several people asked why annotations weren't working in tables.

As for the unnecessary content, that's why we don't just copy and past stuff from other topics and drop them into different topics. The content was necessary when it was in its previous place.

torkelo:
Well then the top sections of most features would be littered with long list of notices for features that where removed or changed. 

I do not think a minor feature change to the table visualizations belongs in the top section. If you have a better idea other than moving this to its own section then please let me know.

oddlittlebird:
As I recall, you said that annotation support would be added to the new table in the near future. This is a temporary band-aid and can be removed once support is added. 

It's here and very visible because right now, users are confused and asking about it. I am attempting to answer their question. It's a short, very visible, one-line note, answering a question that is on the minds of our users.

torkelo:
When annotation support will be added it will get it's own section, so why have it at the top? This way if they miss annotations they can find this notice if they search annotations on this page, or search for table annotations in the docs search. 

We get hundreds of questions every week, not sure the top of each reference page is best for common Q&As

oddlittlebird:
@ryantxu just told me that annotations only work for graph and maybe heatmap panels. If that is the case, then what about moving this information out of the note and somewhere in the Annotations topic?

torkelo:
updated wording a bit, think it still makes sense to have a small section on this page for those missing the feature.

oddlittlebird (approved):
Works for me.

PR #27710, opened 2020-09-22
TraceView: Show full traceID and better discern multiple stackTraces in span details

aocenas:
Show full traceID in the header:
Screenshot from 2020-09-22 17-27-58

Add whitespace and header to each stack trace in text area:
Screenshot from 2020-09-22 17-27-50


zoltanbedi (approved):
lgtm 🍤

PR #27712, opened 2020-09-22
Annotations: check for null or undefined fields before conversion

ryantxu:
In 7.2, we are adding a method to automatically convert DataFrame to annotations -- if the frame returns a null field value, rather than undefined, it will break the tags extraction.


kminehart (approved):
Thanks for the quick turnaround on this! :)

PR #27716, opened 2020-09-22
Transformations changes

achatterjee-grafana:
This PR has the last few changes suggested by @oddlittlebird for Transformations changes.

I missed out on committing the changes before merge.


oddlittlebird (approved):
Love it. At some point in the future, we might want to further break down the transformation types, but that is not today.

You have some conflicts to resolve. Let me know if you need help with that.

docs/sources/manage-users/_index.md
```diff
@@ -9,7 +9,7 @@ weight = 3
 
 # Manage users
 
-Create users and teams and configure [Permissions]({{< relref "../permissions/overview.md" >}}) to make sure that users only have access to the resources they need.
+Create users and teams and configure [Permissions Overview]({{< relref "../permissions/overview.md" >}}) to make sure that users only have access to the resources they need.
```
oddlittlebird, line 12:
```suggestion
Create users and teams and configure [Permissions]({{< relref "../permissions/overview.md" >}}) to make sure that users only have access to the resources they need.
```

torkelo:
Great job! the Prerequisites file is still there: https://github.com/grafana/grafana/blob/master/docs/sources/panels/transformations/prereq.md 

On the Prerequisites section, it's not quite accurate (the old text, not related to this PR). 

It states

> You must have applied a visualization that supports queries. Examples are:

And then lists all visualizations. This line & list are not 100% accurate. What is true is that different transformations will result in a data structure that only some visualizations can show. For example Graph cannot show reduced data (that results from the reduce transform). And it's always good to switch to Table to check the result of a transform if what your current visualization does not show what you expect after applying a transform.

PR #27718, opened 2020-09-23
ValueMappings: Fix issue with value mappings in override applying to all columns

### LINKED ISSUE #25764: Value Mapping on specific Column in Table element is applied to all Columns
I have setup a new table element with a few columns and added for the column "CoT2" the following value mappings:
-1: Short
0: neutral
1: Long

This works for column "CoT2"  as designed, but the value mapping is also extended to the column "Sortierung".

**What happened:**
Value mapping of one column is applied to all columns.

**What is expected:**
Value mapping assigned to one specific column should only be applied to this specific column.

**Environment:**
Grafana v7.0.3 (00ee734baf), Docker hosted, Debian
MySQL 5.7 Data Source
Client: MacBook Pro 13 (2018), MacOS 10.15.5, Chrome Browser

image

**What happened**:

**Environment**:
- Grafana version:
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

torkelo:
Value mappings editor was not imutable in it's modification of the array, modifying the defaultValue array that was applied to the default override value and default, Fixes #25764

### DIFF
--- a/packages/grafana-ui/src/components/ValueMappingsEditor/MappingRow.tsx
+++ b/packages/grafana-ui/src/components/ValueMappingsEditor/MappingRow.tsx
@@ -7,8 +7,8 @@ import { MappingType, RangeMap, SelectableValue, ValueMap, ValueMapping } from '
 
 export interface Props {
   valueMapping: ValueMapping;
-  updateValueMapping: (valueMapping: ValueMapping) => void;
-  removeValueMapping: () => void;
+  onUpdate: (value: ValueMapping) => void;
+  onRemove: () => void;
 }
 
 const MAPPING_OPTIONS: Array<SelectableValue<MappingType>> = [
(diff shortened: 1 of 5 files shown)


packages/grafana-ui/src/components/ValueMappingsEditor/MappingRow.tsx
```diff
@@ -7,36 +7,36 @@ import { MappingType, RangeMap, SelectableValue, ValueMap, ValueMapping } from '
 
 export interface Props {
   valueMapping: ValueMapping;
```
hugohaggmark, line 9: Maybe change this for consistency?
```suggestion
  value: ValueMapping;
```

hugohaggmark (approved):
LGTM!

packages/grafana-ui/src/components/ValueMappingsEditor/MappingRow.tsx
torkelo: this component has a lot of callbacks with a value argument:

```ts
export const MappingRow: React.FC<Props> = ({ valueMapping, onUpdate, onRemove }) => {
  const { type } = valueMapping;

  const onMappingValueChange = (value: string) => {
    onUpdate({ ...valueMapping, value: value });
  };

  const onMappingFromChange = (value: string) => {
    onUpdate({ ...valueMapping, from: value });
  };

  const onMappingToChange = (value: string) => {
    onUpdate({ ...valueMapping, to: value });
  };

  const onMappingTextChange = (value: string) => {
    onUpdate({ ...valueMapping, text: value });
  };
```

So change the prop to value would have some ripple effects

PR #27720, opened 2020-09-23
CI: Upgrade build image

aknuds1:
Upgrade build image for Drone and CircleCI. Only upgrades golangci-lint to latest minor version, plus Node and Yarn to latest patch versions.


papagian (approved):
LGTM!

PR #27729, opened 2020-09-23
Fix: Update badges on PanelEditor tabs when adding a query/transformation/alert

### LINKED ISSUE #25196: Adding a query doesn't update the counter on the Query tab

1. Open the editor on any panel.
2. Click '+ Query'.
3. **The query counter doesn't update**.
4. Go to the Transformations or the Alerts tab.
5. **The query counter is incremented** only when you navigate away from the Query tab.

**Environment**:
- Grafana version: v7.0.1
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

peterholmberg:
Fixes an issue where the badges on panel editor tabs were not updated when adding a query/transformation/alert

Fixes #25196


public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
```diff
@@ -16,55 +16,79 @@ interface PanelEditorTabsProps {
...
-          const transformations = panel.getTransformations() ?? [];
-          return transformations.length;
-      }
+  componentWillUnmount() {
+    this.props.panel.off(CoreEvents.queryAdded, this.triggerForceUpdate);
+    this.props.panel.off(CoreEvents.queryRemoved, this.triggerForceUpdate);
+    this.props.panel.off(CoreEvents.alertAdded, this.triggerForceUpdate);
+    this.props.panel.off(CoreEvents.alertRemoved, this.triggerForceUpdate);
+    this.props.panel.off(CoreEvents.transformationAdded, this.triggerForceUpdate);
+    this.props.panel.off(CoreEvents.transformationRemoved, this.triggerForceUpdate);
```
hugohaggmark, line 35: @peterholmberg noticed that e2e tests were complaining with this:
```
(node:20782) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 preprocessor:close listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
(node:20782) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 preprocessor:close listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
```

I've checked and this is already on master @peterholmberg so this PR hasn't introduce it.

But when I debug this PR and reach `componentWillUnmount` the events array in the emitter is already empty. Could be because Panel has already been destroyed or the props.panel is a different reference? https://github.com/grafana/grafana/blob/17a1e78754a036b2671437e9b59b0734385c74e4/public/app/features/dashboard/state/PanelModel.ts#L454

Not sure what the best way is to register/unregister emitter this without memory leaks in this case, maybe @torkelo @dprokop has suggestions?

mckn: If this is the case maybe we should store the `unsubscribe` functions for each subscription within this panel and run those on unmount instead of the `off`.

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
```diff
@@ -16,55 +16,77 @@ interface PanelEditorTabsProps {
...
-  const styles = getPanelEditorTabsStyles();
-  const activeTab = tabs.find(item => item.active)!;
+export class PanelEditorTabs extends PureComponent<PanelEditorTabsProps> {
+  componentDidMount() {
+    const { panel } = this.props;
+    panel.on(CoreEvents.queryAdded, this.triggerForceUpdate);
+    panel.on(CoreEvents.queryRemoved, this.triggerForceUpdate);
+    panel.on(CoreEvents.transformationAdded, this.triggerForceUpdate);
+    panel.on(CoreEvents.transformationRemoved, this.triggerForceUpdate);
+  }
```
mckn, line 26: If this still would have been a functional component we could have used a `useEffect` hook to handle the unsubscribe of the events.

But in this case we might be able to do something like:

```suggestion
  componentDidMount() {
      const { panel } = this.props;
      this.subscriptions = [
          panel.on(CoreEvents.queryAdded, this.triggerForceUpdate),
          panel.on(CoreEvents.queryRemoved, this.triggerForceUpdate),
          panel.on(CoreEvents.transformationAdded, this.triggerForceUpdate),
          panel.on(CoreEvents.transformationRemoved, this.triggerForceUpdate),
      ];
  }
```

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
```diff
@@ -16,55 +16,77 @@ interface PanelEditorTabsProps {
...
-          const transformations = panel.getTransformations() ?? [];
-          return transformations.length;
-      }
+  componentWillUnmount() {
+    const { panel } = this.props;
+    panel.off(CoreEvents.queryAdded, this.triggerForceUpdate);
+    panel.off(CoreEvents.queryRemoved, this.triggerForceUpdate);
+    panel.off(CoreEvents.transformationAdded, this.triggerForceUpdate);
+    panel.off(CoreEvents.transformationRemoved, this.triggerForceUpdate);
+  }
```
mckn, line 34: And then run the following code to unsubscribe:

```suggestion
  componentWillUnmount() {
    this.subscriptions.forEach(unsubscribe => unsubscribe());
  }
```

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
mckn: I am assuming that the `panel.on` is returning a function to unsubscribe, but that might not be the case?

public/app/features/dashboard/state/PanelModel.ts
```diff
@@ -488,6 +505,14 @@ export class PanelModel implements DataConfigSource {
   getSavedId(): number {
     return this.editSourceId ?? this.id;
   }
+
+  on<T>(event: AppEvent<T>, callback: (payload?: T) => void) {
+    this.events.on(event, callback);
```
mckn, line 510: So maybe change so we return the `unsubscribe` function here.

mckn (changes requested):
I added some really small nits but over all I think the code looks great!

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
```diff
@@ -16,55 +16,77 @@ interface PanelEditorTabsProps {
...
-export const PanelEditorTabs: React.FC<PanelEditorTabsProps> = ({ panel, dashboard, tabs, onChangeTab }) => {
-  const styles = getPanelEditorTabsStyles();
-  const activeTab = tabs.find(item => item.active)!;
+export class PanelEditorTabs extends PureComponent<PanelEditorTabsProps> {
+  componentDidMount() {
+    const { panel } = this.props;
+    panel.on(CoreEvents.queryAdded, this.triggerForceUpdate);
+    panel.on(CoreEvents.queryRemoved, this.triggerForceUpdate);
+    panel.on(CoreEvents.transformationAdded, this.triggerForceUpdate);
+    panel.on(CoreEvents.transformationRemoved, this.triggerForceUpdate);
```
dprokop, line 25: Wonder if having `queriesChanged`/`transformationsChanged` events wouldn't be sufficient? I suggest this as we are using these events only in this single case, and limiting the number of events would simplify the logic responsible for actually sending those. You would avoid checking whether or not something was added/removed. I think the less code in this case the better:)

dprokop (changes requested):
I think this looks good, but I suggest limiting the number of introduced events as described in the comment. WDYT @mckn @hugohaggmark ?

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
peterholmberg: Hm, yes that's true. Updating

dprokop: thank you!

hugohaggmark: Great suggestion @dprokop

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
hugohaggmark: Didn't know it returned an `unsubscribe` function, that would be a better approach indeed!

public/app/features/dashboard/components/PanelEditor/PanelEditorTabs.tsx
peterholmberg: It seems like the CI agreed! :smile:

dprokop: hahah, yeah!

dprokop (approved):
🥳

mckn:
> I think this looks good, but I suggest limiting the number of introduced events as described in the comment. WDYT @mckn @hugohaggmark ?

My suggestion is that we merge this one as it is but open an issue at investigating that we unsubscribe properly after adding an event subscription. Seems like there is some missmatch in panel vs component lifecycle.

mckn:
> > I think this looks good, but I suggest limiting the number of introduced events as described in the comment. WDYT @mckn @hugohaggmark ?
> 
> My suggestion is that we merge this one as it is but open an issue at investigating that we unsubscribe properly after adding an event subscription. Seems like there is some missmatch in panel vs component lifecycle.

Opened this one: https://github.com/grafana/grafana/issues/27800

mckn (approved):

PR #27730, opened 2020-09-23
Transformations changes

achatterjee-grafana:
Two changes:

Reverted relref in Manage user topic as per Diana's suggestion.
Updated the "Prereq" section in the Transformations Overview topic based on Torkel's feedback.


achatterjee-grafana:
Will create a separate PR.

PR #27732, opened 2020-09-23
Transformtion changes 2

achatterjee-grafana:
Three changes:

1. Removed prereq.md file.
2. Updated description of the prereq section in the Transformations overview topic.
3. Reverted relref label to "Preview" in the Manage users Overview topic.


docs/sources/panels/transformations/_index.md
```diff
@@ -29,11 +29,11 @@ Before you can configure and apply transformations:
 
 - You must have entered a query and returned data from a data source. For more information on queries, refer to [Queries]({{< relref "../queries.md" >}}).
   
-- You must have applied a visualization that supports queries. Examples are:
+- You must have applied a visualization that supports your queries. Different transformations can result in a data structure that only some visualizations can render. For example, the visualization cannot show reduced data (that results from the Reduce transform). In case  your current visualization does not show you what you expect after applying a transform, switch to the Table visualization to check the result. For more information on the different types of visualizations, refer to:
```
torkelo, line 32:
```suggestion
- You must have applied a visualization that supports your query result and chosen transforms. Different transformations can result in a data structure that only some visualizations can render. For example, the visualization cannot show reduced data (that results from the Reduce transform). In case  your current visualization does not show you what you expect after applying a transform, switch to the Table visualization to check the result. For more information on the different types of visualizations, refer to:
```

torkelo (approved):
Looks good, added suggested edit

docs/sources/panels/transformations/_index.md
oddlittlebird: I suggest breaking up this paragraph. General rule is three sentences or fewer per paragraph, 25 words or less per sentence.

oddlittlebird: Also, you can probably replace "in case" with "if."

achatterjee-grafana: Keeping the content into one paragraph, as it is a bullet list.

PR #23048, opened 2020-03-25
Chore: replace Angular injected dependencies with dependencies pattern

hugohaggmark:
**What this PR does / why we need it**:
I've started a new side hobby project that relates to #21779. Before I dive in and do a lot of refactoring I'd like to get some early feedback if this suggested path is reasonable?

My intial thought was to go ahead and remove Angular dependencies from all constructurs but that would demand a lot of rewrite of tests that use these constructors. So instead I introduce this pattern:

**Before**
```typescript
/** ngInject **/
constructor(private templateSrv: TemplateSrv, private timeSrv: TimeSrv)
```

**After**
```typescript
interface Dependencies {
  templateSrv: TemplateSrv;
  timeSrv: () => TimeSrv;
}
constructor(private dependencies: Dependencies = { templateSrv: templateSrv, timeSrv: getTimeSrv })
```
Is there a better pattern to handle this? Is it worth it? All feedback is good feedback and because of that I've added a lot of reviewers so please don't be shy :)

**Which issue(s) this PR fixes**:
Fixes #21779

**Special notes for your reviewer**:

PR #27735, opened 2020-09-23
Chore: Remove angular dependency from data sources

kaydelaney:
Removes the use of `@ngInject` for dependency injection from data sources. Dependency injection is still used, but is done through default parameters. In some cases this can simplify testing since the ceremony of jest mocks is no longer required.
Builds upon some earlier work by @hugohaggmark (https://github.com/grafana/grafana/pull/23048)


public/app/features/dashboard/services/TimeSrv.ts
```diff
@@ -316,3 +316,5 @@ export function getTimeSrv(): TimeSrv {
 }
 
 coreModule.service('timeSrv', TimeSrv);
+
+export default getTimeSrv();
```
hugohaggmark, line 320: I think we try to avoid default exports.

hugohaggmark: Aha ok, I see what you're doing in the rest of the PR.

hugohaggmark (commented):
Looks great but I wasn't aware that we had exported TemplateSrv as default which is a pattern that we've previously tried to avoid.

Another thing that came up during the review is that datasource plugins should depend on @grafana/runtime and not files under app/features...

torkelo (approved):
Nice! 

yes, if possible would be better to get function and types from @grafana/runtime (but might be some cases where there are functions that are used that are not part of the public interface).

kaydelaney:
There are indeed some situations where data sources use `TemplateSrv` methods that aren't part of the public interface, for example the influx data source makes use of `templateSrv.variableExists` and `templateSrv.getAdhocFilters`.
Should these methods perhaps get added to the public interface? @torkelo

torkelo:
For those cases you can keep using the internal ones for now.

torkelo:
But I would still avoid export default

torkelo (approved):
Looks good. Not super happy with getTemplateSrv().replace in many places harder to read than templateSrv.replace . The idea behind the get functions was to help in jest mocks and with lazy eval of a singleston but this was before I knew you can have module variables that you can change and have that impact all places that already imported that variable. 

So we could technically stop with the get functions and just export/import the singleton variable instead. Maybe a good topic to discuss on a frontend weekly :)

hugohaggmark:
Shouldn't we try to use grafana/runtime in all the cases except the ones that use more then the exported functions?

dprokop:
> Shouldn't we try to use grafana/runtime in all the cases except the ones that use more then the exported functions?

I think we should and I believe this is what we did until now.

public/app/features/dashboard/components/DashboardRow/DashboardRow.tsx
```diff
@@ -69,7 +69,7 @@ export class DashboardRow extends React.Component<DashboardRowProps, any> {
       'dashboard-row--collapsed': this.state.collapsed,
     });
 
-    const title = templateSrv.replace(this.props.panel.title, this.props.panel.scopedVars, 'text');
+    const title = getTemplateSrv().replace(this.props.panel.title, this.props.panel.scopedVars, 'text');
```
hugohaggmark, line 72: This component uses only the replace function, should be ok to just import runtime?

public/app/features/dashboard/dashgrid/PanelHeader/PanelHeader.tsx
```diff
@@ -131,7 +131,7 @@ export class PanelHeader extends Component<Props, State> {
   render() {
     const { panel, scopedVars, error, isViewing, isEditing, data, alertState } = this.props;
     const { menuItems } = this.state;
-    const title = templateSrv.replace(panel.title, scopedVars, 'text');
+    const title = getTemplateSrv().replace(panel.title, scopedVars, 'text');
```
hugohaggmark, line 134: This component uses only the replace function, should be ok to just import `getTemplateSrv` from grafana/runtime?

hugohaggmark:
@kaydelaney maybe I'm just tired :) but I can still find components that only use the `replace` function which means that we could use the getTemplateSrv from `grafana/runtime`

public/app/features/variables/query/actions.ts
```diff
@@ -138,5 +138,5 @@ const getTemplatedRegex = (variable: QueryVariableModel): string => {
     return '';
   }
 
-  return templateSrv.replace(variable.regex, {}, 'regex');
+  return getTemplateSrv().replace(variable.regex, {}, 'regex');
```
hugohaggmark, line 141: Maybe runtime import for this one?

public/app/plugins/datasource/cloudwatch/query_parameter_ctrl.ts
```diff
@@ -1,12 +1,11 @@
 import angular from 'angular';
 import coreModule from 'app/core/core_module';
 import _ from 'lodash';
-import { TemplateSrv } from 'app/features/templating/template_srv';
-import DatasourceSrv from 'app/features/plugins/datasource_srv';
+import { TemplateSrv } from '@grafana/runtime';
 
 export class CloudWatchQueryParameterCtrl {
   /** @ngInject */
-  constructor($scope: any, templateSrv: TemplateSrv, uiSegmentSrv: any, datasourceSrv: DatasourceSrv) {
```
hugohaggmark, line 9: Is this an intentional change?

public/app/plugins/datasource/loki/result_transformer.ts
```diff
@@ -245,7 +245,7 @@ export function createMetricLabel(labelData: { [key: string]: string }, options?
   let label =
     options === undefined || _.isEmpty(options.legendFormat)
       ? getOriginalMetricName(labelData)
-      : renderTemplate(templateSrv.replace(options.legendFormat ?? '', options.scopedVars), labelData);
+      : renderTemplate(getTemplateSrv().replace(options.legendFormat ?? '', options.scopedVars), labelData);
```
hugohaggmark, line 248: This seems to be a runtime import too

hugohaggmark (approved):
Fantastic work Kay, found some more places that I think could use a runtime import

public/app/plugins/datasource/cloudwatch/query_parameter_ctrl.ts
kaydelaney: Yes, since the class doesn't actually make use of datasourceSrv I thought it could just be removed

hugohaggmark: Ok, sounds good!

PR #27737, opened 2020-09-23
Prometheus: refactor to DataFrame

zoltanbedi:
Adds typing to Prometheus data source.
Converts TableModel and TimeSeries to DataFrame.

## Todo

- [x] Heatmap format
- [x] Tests rephrasing


public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,47 +1,56 @@
...
-      ];
-    } else if (prometheusResult && options.format === 'heatmap') {
+      const tableData = this.transformMetricDataToTable(
+        prometheusResult.result,
+        options.responseListLength,
+        options.refId,
+        options.meta,
+        options.valueWithRefId
+      );
+      return [toDataFrame(tableData)];
```
aocenas, line 26: So adding `toDataFrame` seems like the main change right? I mean I assume this works, bot not ideal as we will do `proResponse -> tableModel -> dataframe` running two different transforms and keeping the code for both. Is this to save time?

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,47 +1,56 @@
...
-        } else if (response.data.data.resultType === 'vector') {
-          seriesList.push(this.transformInstantMetricData(metricData, options));
-        }
+      return seriesList.map(timeSeries => toDataFrame(timeSeries));
+    }
+
+    const seriesList: TimeSeries[] = [];
+
+    if (prometheusResult.resultType === 'matrix') {
+      for (const metricData of prometheusResult.result) {
```
aocenas, line 42: I assume if response has 1 `resultType` they cannot be mixed, like having both `vector` and `matrix` in the same response right?

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: Yeah, that's why I'm not sure if this is good enough or we don't want to do the tableModel/timeseries conversion but the dataframe. I can refactor this more so the prometheus request will be converted to dataframe.

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: no I don't think so.

ivanahuckova (commented):
Great job Zoltan! The added typings are refactors are 🔝🤩😍! Really like the changes.

Regarding the main refactor `toDataFrame`, I am not sure how much work would it be to create methods that transforms everything directly to DataFrames and how much of work with exemplars is still ahead. But if it would take some time that we don't have, maybe we can merge this as temp solution (@zoltanbedi  have you checked if it has any big and visible impact on performance) and continue to work on exemplars and then we can refactor it. Or someone else from squad could refactor it.

zoltanbedi:
I've removed the toDataFrame function from the result_formatter and the transform functions returns data frames now.

ialoh:
ср, 23 вер. 2020 о 21:17 Zoltán Bedi <notifications@github.com> пише:

public/app/plugins/datasource/prometheus/datasource.ts
```diff
@@ -493,8 +482,12 @@ export class PrometheusDatasource extends DataSourceApi<PromQuery, PromOptions>
       }
     }
 
-    return this._request(url, data, { requestId: query.requestId, headers: query.headers }).pipe(
-      catchError(err => {
+    // This can return a scalar data as well
```
ivanahuckova, line 485: I am not sure I understand the meaning of this comment. If this can return a scalar data as well, shouldn't the typing reflect that.

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,266 @@
...
+  PromMatrixData,
+  PromMetric,
+  PromValue,
+  PromVectorData,
+  TransformOptions,
+} from './types';
+
+type MatrixOrVectorResult = PromMatrixData['result'][0] | PromVectorData['result'][0];
+
+export function transform(response: FetchResponse<PromDataSuccessResponse>, options: TransformOptions) {
```
ivanahuckova, line 23: I really like how clean and organized this is.

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,266 @@
...
+  const prometheusResult = response.data.data;
```
ivanahuckova, line 24: I see that you have changed the `prometheusResult` from `response.data.data.result` to `response.data.data` and then you are using `prometheusResult.result` and `prometheusResult.resultType`. As a personal preference, I would maybe create `prometheusResult` and `prometheusResultType` and use just those.

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,266 @@
...
+    return [tableData];
+  }
 
-    if (metricData.values === undefined) {
-      throw new Error('Prometheus heatmap error: data should be a time series');
-    }
+  if (options.format === 'heatmap') {
+    const dataFrame: DataFrame[] = [];
+
+    prometheusResult.result.forEach((data: MatrixOrVectorResult) =>
```
ivanahuckova, line 49: Super tiny nit-pick, this function 


```
prometheusResult.result.forEach((data: MatrixOrVectorResult) =>
      dataFrame.push(transformToDataFrame(data, options))
    );
```

is the same for both - heatmap and non-heatmap format. So maybe we could reuse it and check for heatmap format after it.

ivanahuckova (approved):
LGTM! Really great job. Just curious - why did you decide to remove ResultTransformer class and export functions? Also, as this is such a big PR, I definitely recommend for more people to test it and review. Also added couple nit picks and questions.

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: This is because typescript can infer the type like this. so if resultType is scalar then it will know that the result is going to be PromScalarData.

public/app/plugins/datasource/prometheus/datasource.ts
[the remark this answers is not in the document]
zoltanbedi: You are right. Got me 😅

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: good catch!

zoltanbedi:
> why did you decide to remove ResultTransformer class and export functions?
Using a class here felt like unnecessary. Using just functions make the code more understandable/readable IMO.

public/app/plugins/datasource/prometheus/datasource.ts
```diff
@@ -193,13 +175,14 @@ export class PrometheusDatasource extends DataSourceApi<PromQuery, PromOptions>
...
-        /** Fix for showing of Prometheus results in Explore table.
+        /**
+         * Fix for showing of Prometheus results in Explore table.
          * We want to show result of instant query always in table and result of range query based on target.runAll;
          */
         preferredVisualisationType: target.instant ? 'table' : mixedQueries ? 'graph' : undefined,
       },
     };
-    const series = this.resultTransformer.transform(response, transformerOptions);
+    const series = transform(response, transformerOptions);
```
aocenas, line 185: So at this point this function is called only here and so i wonder if it makes sense to have the `processResult` function. It basically just maps some of the params and nothing else, maybe it would make sense to move that into the transform function and remove this additional step? What do you think?

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,262 @@
...
-
-    let i: number, j: number;
-    const metricLabels: { [key: string]: number } = {};
-
-    if (!md || md.length === 0) {
-      return table;
+  const metricLabels: string[] = [];
+  // Collect all labels across all metrics
+  md.forEach(series => {
+    for (const label in series.metric) {
```
aocenas, line 120: this should probably be `const label of series.metric`

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,262 @@
...
+    const tableData = transformMetricDataToTable(prometheusResult.result, options);
+    return [tableData];
+  }
 
-    if (metricData.values === undefined) {
-      throw new Error('Prometheus heatmap error: data should be a time series');
-    }
+  const dataFrame: DataFrame[] = [];
+
+  prometheusResult.result.forEach((data: MatrixOrVectorResult) => dataFrame.push(transformToDataFrame(data, options)));
```
aocenas, line 48: I feel like this is not super clear mainly in the sense that which branch is exclusive with what and what we are actually checking. Something like this, just adding a few comments and maybe add indentation to mark a block:
```
if (resultType === 'scalar') { transform and return }
if (format === 'table') {
  return table
} else {
  // format is timeseries or heatmap I assume

  // In case of heatmap we process timeseries anyway and heatmap is then createad from that
  processTimeseries

  if (format === 'heatmap') {
    return heatmap
  }
}
```

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,262 @@
...
+/**
+ * Transforms matrix and vector result from Prometheus to DataFrame
+ */
+function transformToDataFrame(data: MatrixOrVectorResult, options: TransformOptions): DataFrame {
+  const { name } = createLabelInfo(data.metric, options);
 
-    for (const value of metricData.values) {
+  const fields: Field[] = [];
+
+  if ('values' in data) {
```
aocenas, line 67: Could we change this to something like `if (isMatrixData(data))`? I think it would be clearer as I had to do 2 steps to check the typing to see that this is actually to check between metrics and vector type of result. If done as a type guard it should probably not be a problem for TS to keep working the same way.

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -1,234 +1,262 @@
...
-      title = options.query;
-    }
+  let title = __name__ || '';
 
-    title = `${__name__ ?? ''}${labelPart}`;
+  const labelPart = formatLabels(labelsWithoutName);
 
-    return { name: title, title, labels: labelsWithoutName };
+  if (!title && !labelPart) {
+    title = options.query;
```
aocenas, line 203: This is a bit weird here, we always overwrite this by the next line so this is a noop, not sure what it should have been doing seems like the same issue was there before.

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: series.metric is an object here so for of wouldn't work.

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: sure, good point.

aocenas:
There seems to be an issue with the table format:
Screenshot from 2020-09-30 14-34-39

On master all the rows are filled.

zoltanbedi:
> There seems to be an issue with the table format:
> Screenshot from 2020-09-30 14-34-39
> 
> On master all the rows are filled.

Nice catch!

public/app/plugins/datasource/prometheus/datasource.ts
zoltanbedi: I like the idea.

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: dah, that didn't stick out for me 🤦‍♂️

public/app/plugins/datasource/prometheus/result_transformer.ts
aocenas: hmm in that case maybe it should also have `hasOwnProperty` check? Or just use `Object.keys`? I assume it's not an issue at this moment but not sure if we should count on how the object is created.

Also in general I think you could just do something like:
```
Object.keys(md.reduce((acc, series) => ({...acc, ...series}) ,{})).sort().map(...)
// or
_.uniq(md.flatMap((series) => Object.keys(series))).sort().map()
```

To get the unique `metricLabels`

aocenas (approved):

ivanahuckova (approved):

public/app/plugins/datasource/prometheus/result_transformer.ts
zoltanbedi: nice, I like the Object.keys way more.

PR #27741, opened 2020-09-23
Minor correction for better readability

amalkurup89:
**What this PR does / why we need it**: Minor correction for better readability

**Which issue(s) this PR fixes**: Documentation fix


aknuds1:
The original language is correct.

achatterjee-grafana:
Agree

PR #27743, opened 2020-09-24
Toolkit: Add --coverage flag to plugin build command

### LINKED ISSUE #27623: Code Coverage to be supported with plugin build script
Support generating code coverage when building grafana plugin.

**Why is this needed**:
When building grafana plugin the tests are also running, but it is not possible to generate coverage. It is possible to generate coverage by running tests by using the flag --coverage, as "grafana-toolkit plugin:test --coverage". In this case in order to build and generate the coverage then we need to run both "grafana-toolkit plugin:test --coverage" and "grafana-toolkit plugin:build" which means the tests will run twice, a thing that can be avoided by supporting --coverage with the build script.

gassiss:

Adds an optional `--coverage` flag, which defaults to `false`. If provided, will run a coverage report after the test step of the build script.

Fixes #27623 

Please let me know if I missed anything! :)

### DIFF
--- a/packages/grafana-toolkit/README.md
+++ b/packages/grafana-toolkit/README.md
@@ -100,6 +100,10 @@ Available options:
 
 This command creates a production-ready build of your plugin.
 
+Available options:
+
+- `--coverage` - Reports code coverage after the test step of the build.
+
 ## FAQ
 
 ### Which version of grafana-toolkit should I use?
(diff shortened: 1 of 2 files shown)


hugohaggmark (approved):
Thank you for contributing ⭐ and works as expected!

PR #27752, opened 2020-09-24
Variables: Make variable queries cancellable with Observables

### LINKED ISSUE #27746: Variables: Make variable queries cancellable with Observables
We want to create a new addition to the datasource api like the addition for annotations added by @ryantxu in this PR:https://github.com/grafana/grafana/pull/27375
API proposal:
```ts
export interface VariableSupport<TQuery extends DataQuery = DataQuery> {
 query:(query: TQuery, options?: any): Observable<MetricFindValue[]>;
 tagKeys?(options?: any): Observable<MetricFindValue[]>;
 tagValues?(options?: any): Observable<MetricFindValue[]>;
}


export abstract class DataSourceApi<
  TQuery extends DataQuery = DataQuery,
  TOptions extends DataSourceJsonData = DataSourceJsonData
> {
...

/**
 ... ... ...
*/
  variables?: VariableSupport<TQuery>;
/**
   * Variable query action.
   *
   * @deprecated -- prefer using {@link VariableSupport}
   */
  metricFindQuery?(query: any, options?: any): Promise<MetricFindValue[]>;

  /**
   * Get tag keys for adhoc filters
   *
   * @deprecated -- prefer using {@link VariableSupport}
   */
  getTagKeys?(options?: any): Promise<MetricFindValue[]>;

  /**
   * Get tag values for adhoc filters
   *
   * @deprecated -- prefer using {@link VariableSupport}
   */
  getTagValues?(options: any): Promise<MetricFindValue[]>;
}
```

With this new api in place we'll need to upgrade the internals of the variable system to use this new api. This will be a bigger effort and I think we should look into Redux-Observable again because I think it's a perfect fit for the variables system. 

Only when we have a variables system that uses this new api can we start migrating our datasource plugins.
- [ ] Google Cloud Monitoring
- [ ] Cloudwatch
- [ ] Elastic
- [ ] Graphite
- [ ] InfluxDB
- [ ] Jaeger
- [ ] Loki (#27265)
- [ ] MSSQL 
- [ ] MySQL (#27649)
- [ ] OpenTSDB
- [ ] Postgres (#27478)
- [ ] Zipkin
- [ ] Prometheus (#27090) 

**Why is this needed**:
We have a challenge with long running variable queries because they are not cancellable and that makes a suboptimal user experience working with variables.

hugohaggmark:
One way to test this is PR is to:

1. Start a slow proxy (30s delay on every data source request) with `make devenv sources=slow_proxy_mac,prometheus`
2. Import this [dashboard](https://gist.github.com/hugohaggmark/067a470d697f91828c9b282836765d07)

Fixes #27746

So this PR introduces a new api for DataSourceAPI and you should be able to cancel individual variable queries if you click the loading indicator. For now if you cancel any other data source than TestData the Observable will be cancelled but the network request will still go on. 
Data sources need to migrate to this new API before the cancellation will work all the way to the network request.

Remaining work:

- [ ] Fix tests and write new
- [ ] Migrate Prometheus to get a feel for the API


public/app/plugins/datasource/testdata/datasource.ts
```diff
@@ -34,9 +38,57 @@ export class TestDataDataSource extends DataSourceApi<TestDataQuery> {
...
+        return request;
+      },
+    };
   }
 
   query(options: DataQueryRequest<TestDataQuery>): Observable<DataQueryResponse> {
+    if (options.app === CoreApp.Variables) {
+      return this.handleVariablesQuery(options);
+    }
+
```
hugohaggmark, line 91: I'm still not super convinced that this is a good API pattern. I think having a specific query function for variables is better. This is just the changes in TestData. I think I'll try to expand this PR to include prometheus as well and see how good/bad that would be.

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { DataQuery, DataQueryRequest, MetricFindValue } from './datasource';
+import { DataFrame } from './dataFrame';
+import { ScopedVars } from './ScopedVars';
+
+export interface VariableSupport<TQuery extends DataQuery = DataQuery> {
+  /**
+   * Defines how to transform the query provided from the variable into the DataQueryRequest used to query for MetricFindValues.
+   */
+  toDataQueryRequest: (query: string | any, scopedVars: ScopedVars) => DataQueryRequest;
```
kaydelaney, line 11: Think `any` just overrides `string` here, so it might as well just be `any`.

packages/grafana-data/src/types/app.ts
```diff
@@ -1,11 +1,12 @@
 import { ComponentClass } from 'react';
 import { KeyValue } from './data';
 import { NavModel } from './navModel';
-import { PluginMeta, GrafanaPlugin, PluginIncludeType } from './plugin';
+import { GrafanaPlugin, PluginIncludeType, PluginMeta } from './plugin';
 
 export enum CoreApp {
   Dashboard = 'dashboard',
   Explore = 'explore',
+  Variables = 'variables',
```
ryantxu, line 9: why is this really needed?

public/app/plugins/datasource/testdata/datasource.ts
ryantxu: If we can avoid the query ever checking `options.app` that would make me happier.  I see this is just in the testdata implementation, so I'm not sure how fundamental/required the app check is.

The use of options.app changing behavior causes a lot of inconsistency that is hard to track down.

ryantxu (commented):
Hugo this is looking great -- my biggest concern is with adding a new App type, I think we should only do that if it is absolutely required.

public/app/plugins/datasource/testdata/datasource.ts
hugohaggmark: Great catch! Maybe we don't need to use this anymore when I think about it. I introduced the toDataQueryRequest to the variable support API so that should be enough. Thank you Ryan!

packages/grafana-data/src/types/variables.ts
hugohaggmark: I know, I just wanted to communicate that for 99% of all current core plugins this is a string but then there is just one that passes an object.

packages/grafana-data/src/types/app.ts
hugohaggmark: Good catch, I will try to remove it :)

packages/grafana-data/src/types/variables.ts
hugohaggmark: I changed it to `any` makes it very clear that you should expect anything :) Thank you Kay!

hugohaggmark:
> Hugo this is looking great -- my biggest concern is with adding a new App type, I think we should only do that if it is absolutely required.

I'm now looking at migrating the Prometheus data source and I'm struggling to see how to solve that without adding a new App type or adding an extra props to the PromQuery interface. Any ideas @torkelo @ryantxu

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { DataQuery, DataQueryRequest, MetricFindValue } from './datasource';
+import { DataFrame } from './dataFrame';
+import { ScopedVars } from './ScopedVars';
+
+export interface VariableSupport<TQuery extends DataQuery = DataQuery> {
+  /**
+   * Defines how to transform the query provided from the variable into the DataQueryRequest used to query for MetricFindValues.
+   */
+  toDataQueryRequest: (query: any, scopedVars: ScopedVars) => DataQueryRequest<TQuery>;
```
torkelo, line 11: This looks a bit strange. The data source cannot define the full DataQueryRequest (it has time range, all sorts of information missing in this context). 

I was thinking more of something like 

toDataQuery(query: any) => TQuery

torkelo (commented):
Great progress! added a comment on the VariableSupport interface toDataQueryRequest function, that feels to me like a too low level of a hook.

packages/grafana-data/src/types/variables.ts
hugohaggmark: I see, thank you that makes a lot of sense. 
What is your thoughts about Ryans concern about adding `CoreApp.Variables`? I'm struggling to solve the Prometheus migration without adding that or extending the PromQuery interface with a new prop.

hugohaggmark:
So @ryantxu, @torkelo  I'm going to revert and add back `Variables` to `CoreApp` to solve the migration of Prometheus today, push it and then we can see if there is another way forward.

hugohaggmark:
closing this in favor of #28217

PR #27757, opened 2020-09-24
PanelData: Adds annotations and QueryResultMeta dataTopic 

torkelo:
This PR makes it possible for a data source to mark part of a query result as belonging to the annotation data topic. 

* Adds annotations to PanelData
* runRequest is responsible for segregating out this annotation data into a separate PanelData array'
* Adds panelData to MetricsPanelCtrl so that the old graph can access this new annotations array 
* Makes it possible for graph panel to use annotations returned via PanelData 

This should be enough to start experimenting with returning exemplar data via this annotation dataTopic and for the new graph component to start rendering annotation & exemplars.  

For later PRs

* Introduce a dashboard wide PanelQueryRunner (or equivalent) with a shared rxjs Subject to subscribe to dashboard wide annotation query results
* Make it possible to add query groups (with different data sources and query options, query option to mark query group to handle results as annotations) 
* Extend PanelQueryRunner or an extension / new level of runRequest to handle executing query groups and merging different PanelData results 
* Introduce a way for panels to add a query group that is just placeholder for a dashboard level annotation subscription (where we give panels an option to select which dashboard level annotation queries to subscribe to, defaults to all). 

To verify that this PR works there is an annotations TestData scenario:
Screenshot from 2020-09-24 12-49-10


public/app/features/annotations/standardAnnotationSupport.ts
```diff
@@ -54,19 +54,12 @@ export function singleFrameFromPanelData(data: DataFrame[]): DataFrame | undefin
...
-  return transformDataFrame(
-    [
-      {
-        id: 'seriesToColumns',
-        options: { byField: 'Time' },
-      },
-    ],
-    data
-  )[0];
+  return standardTransformers.mergeTransformer.transformer({})(data)[0];
```
dprokop, line 62: 👍

dprokop (approved):
As simple as it looks I think this is a good first step @torkelo

papagian (approved):
I've tested and AFAIKT LGTM

ryantxu (approved):

PR #27768, opened 2020-09-24
Chore: adding platform independent typings for e2e tests.

mckn:
Prior to this PR, when running the end-to-end tests locally, they would fail depending on what platform you run them on. If running on linux or windows they will succeed. But if you are running them on OSX they will fail.

The reason is that in one of the tests we wanted to do an undo action in a text field. The issue in the test was that it did a `{ctrl}z` which works on Windows and Linux. But on OSX it should be `{cmd}z`.

This PR resolves this by extracting that logic into a function that checks what platform we are running the tests for.

### DIFF
--- a/packages/grafana-e2e/src/typings/undo.ts
+++ b/packages/grafana-e2e/src/typings/undo.ts
@@ -1,16 +1,17 @@
-const platforms = {
-  osx: 'darwin',
-  windows: 'win32',
-  linux: 'linux',
-  aix: 'aix',
-  freebsd: 'freebsd',
-  openbsd: 'openbsd',
-  sunos: 'sunos',
-};
+// https://nodejs.org/api/os.html#os_os_platform
+enum Platform {
+  osx = 'darwin',
+  windows = 'win32',
+  linux = 'linux',
+  aix = 'aix',
+  freebsd = 'freebsd',
+  openbsd = 'openbsd',
+  sunos = 'sunos',
+}
 
 export const undo = () => {
   switch (Cypress.platform) {
-    case platforms.osx:
+    case Platform.osx:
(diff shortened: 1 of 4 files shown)


stevenvachon (approved):

packages/grafana-e2e/src/typings/undo.ts
```diff
@@ -0,0 +1,18 @@
...
+  linux: 'linux',
+  aix: 'aix',
+  freebsd: 'freebsd',
+  openbsd: 'openbsd',
+  sunos: 'sunos',
+};
+
+export const undo = () => {
+  switch (Cypress.platform) {
+    case platforms.osx:
```
kaydelaney, line 13: Nit: An enum might be more appropriate

```suggestion
enum Platform {
  osx = 'darwin',
  windows = 'win32',
  linux = 'linux',
  aix = 'aix',
  freebsd = 'freebsd',
  openbsd = 'openbsd',
  sunos = 'sunos',
}

export const undo = () => {
  switch (Cypress.platform) {
    case Platform.osx:
```

kaydelaney: Also might leave a comment indicating where the values for `platform` come from (https://nodejs.org/api/os.html#os_os_platform)

mckn: Great feedback! I will update, thanks!

kaydelaney (approved):
LGTM!

zoltanbedi:
A bit late to the party but awesome work here! 🕺

hugohaggmark (commented):
Outstanding work ⭐

PR #27784, opened 2020-09-25
BarGauge: Fixed scrollbar showing for bar gauge in Firefox

### LINKED ISSUE #27733: in firefox bar graphs always have a scroll bar
**What happened**:
I upgraded to 7.2.0 and in Firefox bar gauges now show a scroll bar (or at least space for a scroll bar), even when not required, and which is obscuring the values on the right. 
image
image

since there's no need to scroll, the bar shouldn't appear. If it does appear the panel content should not be obscured by it. 

Using firefox, create a panel, add a query, and choose Bar graph for the visualization. 

Seems normal in Chrome. 

**Environment**:
- Grafana version: 7.2.0
- Data source type & version: Any? I can reproduce with both graphite 1.x and ES 7
- OS Grafana is installed on: Ubuntu 18.04 x64
- User OS & Browser: Mac OS 10.15.6, FF 80.0.1, 81.0 
- Grafana plugins: bar graph
- Others:

torkelo:
Fixes #27733

### DIFF
--- a/packages/grafana-ui/src/components/VizRepeater/VizRepeater.tsx
+++ b/packages/grafana-ui/src/components/VizRepeater/VizRepeater.tsx
@@ -152,7 +152,7 @@ export class VizRepeater<V, D = {}> extends PureComponent<Props<V, D>, State<V>>
 
     const repeaterStyle: React.CSSProperties = {
       display: 'flex',
-      overflow: minVizHeight ? 'hidden scroll' : 'visible',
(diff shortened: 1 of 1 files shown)


dprokop (approved):

dprokop:
@torkelo I could not reproduce this, but I trust you have managed to :)

torkelo:
you need to use firefox with overlay scrollbars disabled

PR #27788, opened 2020-09-25
Orgs: Remove organisations deprecation notice from backend

### LINKED ISSUE #27769: Orgs: Remove org deprecation notice in backend
Deprecation message of orgs was partially removed in #26853. We missed that there's also a message in the backend being logged when Grafana starts that also needs to be removed.

Code: https://github.com/grafana/grafana/blob/f0a586a9143b40530a3025fbe88faeab6460dc82/pkg/services/sqlstore/sqlstore.go#L119-L141

wbrowne:
As we have decided to preserve multi-org support, we should remove the org deprecation warning from the backend logs

Fixes https://github.com/grafana/grafana/issues/27769


marefr (approved):
Great. LGTM

papagian (approved):
LGTM

PR #27789, opened 2020-09-25
Dashboard: Don't send analytics for explore or edit events

AgnesToulet:
This PR does two things:
- Fix the check to not send events in Explore because Explore events are not handled by the analytics engine for now.
- Add a check to not send events while editing a panel because it can lead to a lot of request errors that are only related to a temporary state.


public/app/features/dashboard/state/analyticsProcessor.ts
```diff
@@ -18,7 +18,12 @@ export function emitDataRequestEvent(datasource: DataSourceApi) {
       return;
     }
 
-    if (data.request.exploreMode) {
```
joanlopez, line 21: Just for curiosity: Is this field no longer user? 🤔  Thanks!

joanlopez (approved):
LGTM! 👍

jongyllen (commented):
Nice!
Maybe a few test cases would be nice to have around this behaviour?

public/app/features/dashboard/state/analyticsProcessor.ts
AgnesToulet: This field is used in Explore but it seems that it's not set all the time, so it's not reliable enough to use it here.

xlson:
Is the `edit` state set to true when a user changes time range? I'm thinking that we want to keep sending data when only the time range is modified.

AgnesToulet:
> 
> 
> Nice!
> Maybe a few test cases would be nice to have around this behaviour?

I will look at this tomorrow. I don't really know how to write relevant tests as if I set the `data.request.app` value or the query parameters, it will not catch errors if these fields are not used as expected or if they are replaced by other fields.  @Clarity-89 Do you have an idea on how to write good tests for this kind of feature?


> 
> 
> Is the `edit` state set to true when a user changes time range? I'm thinking that we want to keep sending data when only the time range is modified.

No, the `edit` parameter is only set when you are in edit view with one panel and its query editor displayed.

Clarity-89:
> > Nice!
> > Maybe a few test cases would be nice to have around this behaviour?
> 
> I will look at this tomorrow. I don't really know how to write relevant tests as if I set the `data.request.app` value or the query parameters, it will not catch errors if these fields are not used as expected or if they are replaced by other fields. @Clarity-89 Do you have an idea on how to write good tests for this kind of feature?

Maybe mocking the return function of `emitDataRequestEvent` and calling it with a custom data object could work here?

public/app/features/dashboard/state/analyticsProcessor.test.ts
```diff
@@ -0,0 +1,106 @@
...
+jest.mock('app/features/dashboard/services/DashboardSrv', () => ({
+  getDashboardSrv: () => {
+    return {
+      getCurrent: () => dashboardModel,
+    };
+  },
+}));
+
+jest.mock('@grafana/runtime', () => ({
+  ...((jest.requireActual('@grafana/runtime') as unknown) as object),
```
Clarity-89, line 29: `...(jest.requireActual('@grafana/runtime') as any)` should work a well (untested).

public/app/features/dashboard/state/analyticsProcessor.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  reportMetaAnalytics: jest.fn(),
+}));
+
+const mockGetUrlSearchParams = jest.fn(() => {
+  return {};
+});
+jest.mock('@grafana/data', () => ({
+  ...((jest.requireActual('@grafana/data') as unknown) as object),
```
Clarity-89, line 37: Same comment as above.

Clarity-89 (approved):
Great job with the tests! 🎉

public/app/features/dashboard/state/analyticsProcessor.test.ts
AgnesToulet: Yes, it works. Thanks!

xlson:
> No, the `edit` parameter is only set when you are in edit view with one panel and its query editor displayed.

Does the edit state stay set when you leave panel edit but haven't yet saved your Dashboard?

AgnesToulet:
No, it's used to display the edit view so it's only on when you are editing a panel.

xlson:
@AgnesToulet Interesting. I was originally thinking that we would not want to send analytics as long as a Dashboard had unsaved changes (unless those only related to the time range). But I guess it might make sense to only limit it to not sending data while excplicitly in Edit mode. What do you think @sakjur?

I still kind of lean towards wanting to disable analytics until save as large scale changes to a dashboard that includes template variables are likely to include broken panels even when you are not editing a specific panel.

PR #27790, opened 2020-09-25
Variables: Prevents unnecessary calls to update options

hugohaggmark:
I noticed how clicking back and forth between the Query field and RegEx field there were calls made to the datasource. This is particularly bad for long running variable queries. This PR should remedy this.

Fixes #

### DIFF
--- a/public/app/features/variables/query/QueryVariableEditor.test.tsx
+++ b/public/app/features/variables/query/QueryVariableEditor.test.tsx
@@ -1,0 +1,109 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { DataSourceApi } from '@grafana/data';
+
+import { Props, QueryVariableEditorUnConnected } from './QueryVariableEditor';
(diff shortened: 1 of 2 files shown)


dprokop (approved):
Tested and it works as described, great!

nachopv:
Hi @hugohaggmark 

Will this address the issue with the following setup?

- Root variable "On Dashboard Load" refresh policy
- All its chained variables with "Never" in the refresh policy

With this setup we have noticed that when sharing links with parametrized variables, only the root variable is populated and therefore no data is shown. Yo need to go back and forth in root var values in order to trigger the calculation of the chained variables
We have confirmed this is not working neither on grafana 6 and 7.1

hugohaggmark:
Hi @nachopv, previously when you tabbed between fields in the Editor for query variables, every onBlur event triggered a new variable query being fired. This PR makes sure that those calls are only fired when values in the Editor are actually changed. 

So reading your description I don't think this will fix your issue so please create a new issue if you can't find an existing one.

PR #27804, opened 2020-09-25
Move error log in social.go inside err != nil

billoley:
re #27514 

Error log added in https://github.com/grafana/grafana/pull/27523 should be inside err != nil conditional, not printing all the time


aknuds1 (approved):
LGTM!

aknuds1:

PR #27805, opened 2020-09-25
Document Update

NiteshKoushik:
Syntax correction

Fixes #


aknuds1:
This isn't a correction since JavaScript allows trailing commas like this.

NiteshKoushik:
Ok. Thanks for the info.

PR #27829, opened 2020-09-25
Variables: Adds support for key/value mapping in Custom variable

### LINKED ISSUE #1032: Allow custom mapping of template variable value -> display text
Use Case: You may store metrics based on an 'ID' property but wish to have the template variable selection UI use a more human friendly label. e.g. You track metrics by domain with an internal domain ID but wish to use the domain's URL in the template variable selector UI. 

@torkelo I can take a cut at implementing this, what are your thoughts on implementation? For my specific use case I would want to be able to provide an arbitrary JS function to perform the value -> text conversion as I need to hit an external service for the lookup. I was thinking an initial implementation could be adding a config value in the dashboard JSON that defines the mapping function. UI support could be added later to handle more trivial mappings with pre built mapping functions (e.g. regex substitutions). 

Also connected to this would be the ability to edit the full dashboard JSON via the UI, although export -> edit -> import would function as a workaround if this proves to be difficult.

### LINKED ISSUE #1988: Allow key/value in template variables
I would like to use a value list for a graphite query to group hosts. 

Example: 
Template Variable: $clusterhosts 
Value: cluster_a = {host1,host2}, cluster_b = {host[23]}

So i can use this variable in graphite query: 
sys.$clusterhosts.foo.bar

which could be resolved to:
sys.{host1,host2}.foo.bar

and via graphite api to metrics for:
sys.host1.foo.bar
sys.host2.foo.bar

sartaj10:

This PR adds a new variable type where the user can enter key-value pairs as comma separated values where the key-value pairs are separated by colon (:). This would be useful where the user wants the dropdown values to be user-friendly while the query expects them in a different format. Eg : The user-friendly dropdown value could be 100 MB while the query expects it to be in bits (800000000)

Fixes #1988
Fixes #1032 

@hugohaggmark 

Some screenshots are attached below

Screen Shot 2020-09-28 at 3 09 46 PM

Screen Shot 2020-09-28 at 3 09 39 PM


sartaj10:
@hugohaggmark

hugohaggmark:
@sartaj10 thanks for making this contribution, I don't think we need a new Variable Type for this. This behavior belongs to the custom variable type. WDTH?

hugohaggmark:
@sartaj10 not sure what the best separator char would be for this, maybe `|` is also a candidate as it might be more common to use `:` in the label?

torkelo:
yea, this feels like an extension of the custom variable

sartaj10:
@hugohaggmark @torkelo 
Yeah, it does make sense to include this as an extension to the custom variable. I'll work on this and update the PR.

Regarding the separator, the semicolon seems more intuitive (similar to a hashmap). Another option could be to split on "=>"? What do you think

sartaj10:
@hugohaggmark 

Updated the PR. It is now possible to include key/value pairs within the custom variable type and it takes care of the splitting whenever that is present.

public/app/features/variables/custom/reducer.ts
```diff
@@ -36,9 +36,16 @@ export const customVariableSlice = createSlice({
...
       const match = query.match(/(?:\\,|[^,])+/g) ?? [];
-
+      
       const options = match.map(text => {
         text = text.replace(/\\,/g, ',');
+        const textMatch = text.match(/(?:\\:|[^:])+/g) ?? [];
+        if (textMatch.length > 1) {
+          let [key, value] = textMatch;
+          key = key.replace(/\\:/g, ':');
+          value = value.replace(/\\:/g, ':');
```
hugohaggmark, line 46: Have you added any test cases for this use case? I would either add tests for this or remove these lines.

hugohaggmark (commented):
Fantastic work on this @sartaj10 and thank you for contributing ⭐   I left a small comment and you also need to update the docs here:
https://github.com/grafana/grafana/blob/master/docs/sources/variables/variable-types/add-custom-variable.md

public/app/features/variables/custom/reducer.ts
sartaj10: @hugohaggmark 
Do you think we need to cover for the case if someone has a semicolon within either the key or value? If not, i believe i can remove these lines.

hugohaggmark: @sartaj10 I think you should keep them but then you should add a test for them. Also prettiers seems to be complaining on your branch so the build is failing :)

sartaj10: Ok, i'll add that. Yeah, i need to figure out why that build is failing. Is there a way to check where that's happening before pushing the commit?

hugohaggmark: I think prettier is complaining about the formatting in `public/app/features/variables/custom/reducer.ts`. Let me know if I could help out

hugohaggmark:
@sartaj10 I'm going to change this Draft PR into Ready for Review, ok?

sartaj10:
@hugohaggmark 
Sounds good

hugohaggmark:
> @hugohaggmark
> Sounds good

@sartaj10 looks good, you removed the expansion of the colon here 0e52d2e but then you re added it in 318e08b. Have you decided if you're going to keep it or not :)
```ts
let [key, value] = textMatch;	
key = key.replace(/\\:/g, ':');	
value = value.replace(/\\:/g, ':');
```

sartaj10:
@hugohaggmark 

So, the reason for that was to see if that was failing the build or not. For some reason, it turns out, an empty line caused it and this commit fixed that build [318e08b](https://github.com/grafana/grafana/pull/27829/commits/318e08b3b42487df1190dbd51e8b76c5d7fbf40d)  (I'm still not sure why that's the case) 

Regarding keeping that expansion of the colon, I don't have any strong preference for it. It's achieving the same purpose as the comma when someone wants to escape it while adding values to the comma separated list.

hugohaggmark:
> @hugohaggmark
> 
> So, the reason for that was to see if that was failing the build or not. For some reason, it turns out, an empty line caused it and this commit fixed that build [318e08b](https://github.com/grafana/grafana/pull/27829/commits/318e08b3b42487df1190dbd51e8b76c5d7fbf40d) (I'm still not sure why that's the case)
> 
> Regarding keeping that expansion of the colon, I don't have any strong preference for it. It's achieving the same purpose as the comma when someone wants to escape it while adding values to the comma separated list.

Great, thanks for clarifying. You can remove those expansion lines then, if someone wants to add a colon in the label we can change it then so we keep change this lean and small. Thank you again for taking the time with this. As soon as those lines are removed I'll approve and merge this 🎉

sartaj10:
Awesome, thank you so much. I've removed those lines. Do we expect to see this in 7.3?

hugohaggmark:
> Awesome, thank you so much. I've removed those lines. Do we expect to see this in 7.3?

Thank you and yes this has the 7.3 milestone for now, great work!

hugohaggmark (approved):
Looks great! Before we merge this I'd like for the technical writers to approve as well.

sartaj10:
Perfect, thank you.

docs/sources/variables/variable-types/add-custom-variable.md
```diff
@@ -26,7 +26,7 @@ For example, if you have server names or region names that never change, then yo
 
 ## Enter Custom Options
 
-1. In the **Values separated by comma** list, enter the values for this variable in a comma-separated list. You can include numbers, strings, or other variables.
+1. In the **Values separated by comma** list, enter the values for this variable in a comma-separated list. You can include numbers, strings, or other variables. It is also possible to enter key/value pairs separated by a colon.
```
oddlittlebird, line 29:
```suggestion
1. In the **Values separated by comma** list, enter the values for this variable in a comma-separated list. You can include numbers, strings, other variables or key/value pairs separated by a colon.
```

oddlittlebird (approved):
One minor edit, then good to go!

sartaj10:
Thanks, Done 👍

mareksimunek:
would be possible to access also a key as value?
E.g I have
namespaceName : dev

and sometimes I want in query `namespaceName` value and sometimes I want `dev` value, but still with 1 custom variable

HigH-HawK:
Hi,

I have updated my Grafana to the latest version and tried the key/value mapping, which is working fine. I also tried working with the variable data in my panels but seem to do something wrong, as it doesn't do what i want it to do.

Example:

Constellation: **srv-a** is the grafana machine with Prometheus installed - **srv-b** and **srv-c** are the node_exporters, sending the data to **srv-a**

Prometheus has the IP addresses of **srv-b** and **srv-c** setup as targets

In my Dashboard on Grafana I setup a variable with the new mapping feature such as:

Variable name: host
Variable Label: Hosts
Variable Type: Custom
Variable Values: srv-b : IP, srv-c : IP

How can I query the IP (value) of the variable host in my panels? If I just put $host, it just returns the name (key)....I'm pretty sure I need to add something to $host but am just not sure what :(

torkelo:
This breaks all existing dashboards using custom variabes with values that include a colon `:` , think we have to revert or figure out another way to make this and not break existing dashboards

argais:
> would be possible to access also a key as value?
> E.g I have
> namespaceName : dev
> 
> and sometimes I want in query `namespaceName` value and sometimes I want `dev` value, but still with 1 custom variable

I came here looking for this exact same behaviour, did you manage to get this?

argais:
Well, answering myself and @mareksimunek , if you do `${namespace:text}` it'll return the key.

HigH-HawK:
How can I query the value of the variable in my panels?

If I put $variable_name, it just returns the name (key) but I can't seem to find a way, to get the value to use it in queries. Any help would be appreciated.

maard:
How about do not introduce a separator at all, and still use commas: '100MB, 800000000',
and include a new boolean flag: 'Dictionary' (perhaps, under 'Selection Ooptions')?

mingmamma:
> How can I query the value of the variable in my panels?
> 
> If I put $variable_name, it just returns the name (key) but I can't seem to find a way, to get the value to use it in queries. Any help would be appreciated.

I am facing the same problem and let me give a more practical example. My custom variables set up is as the following:
`Region={LAX:us-region,
SYD:australia-southeast1}`
Capture2
I can use `!{Region:text}` to reference the key and get the returned value as either `LAX` or `SYD` with no issue. 
Capture

Looking at the code change in the PR, I would assume that I can use `!{Region:value}` to get either `us-region` or `australia-southeast1`. However, that would throw an error. 
Capture1

offbyone:
@mingmamma did you find a solution for this? I just bumped into the same issue.

hugohaggmark:
@mingmamma & @offbyone we use GitHub for issues and feature requests. For questions we kindly request that you ask your question at the [community site](https://community.grafana.com/) or in our external [Slack   here](https://grafana.slack.com/)

Regarding your question  the `${Region:text}` doesn't format the "text" value but the "value" value so there is no  `${Region:value}`. All the formatting options are described in [https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/)

All the best

offbyone:
@hugohaggmark I understand not wanting to use github issues for support, that's fine. Given that, would someone from the Grafana dev crew mind having a look at [the support question for this topic](https://community.grafana.com/t/how-can-i-use-key-value-variables-in-formatting/41659/2)?

majorvin:
I just installed the latest version 7.4.2 and not seeing this feature.

hugohaggmark:
> I just installed the latest version 7.4.2 and not seeing this feature.
> 
> EDIT: Ok, it needs to have a space after the key and after colon. e.g. `A : value`

Thank you for noticing this @majorvin, I've opened a PR that tries to clarify this #31303

hilmialf:
Is it possible to make they selected one is based on key rather than value?
I recently use this kv pairs:
`key1 : value 1, key2 : value2, key3 : value2`
*notice I use value2 in both key2 and key3.
When I select key2 in my dashboard, key3 also got highlighted. Then I can no longer select key3 from there.
I can only select key3 if my previous selected key was key1.

PR #27838, opened 2020-09-26
Units: bps & Bps default scale remains decimal (backwards-compatibility)

Berbe:
Fixes broken backwards-compatibility by makings the ID of units not having binary equivalent to their decimal format stick to the decimal one.

Fixes #27022

**Special notes for your reviewer:**
Break of backwards-compatibility [reported](https://github.com/grafana/grafana/pull/27022#issuecomment-699084341) by @tarko.


Berbe:
@mckn You might wish to take a look at this for the next maintenance release.

torkelo:
Dam this is not good, I missed this breaking change in the https://github.com/grafana/grafana/pull/27022 PR, 

This is adding more breaking changes to everyone running 7.2 (decBps & decbps disappearing). 

We need to move the valueFormats registry to use https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/utils/Registry.ts  

That way we can easily add support for aliases to valueFormats to support legacy ids

Berbe:
Yup, I oversaw this one too... There were many things to keep in check.

I would think the breaking change towards 7.2 users is minimal, as the user base still using the old pre-7.2 IDs is the vast majority (and some of them probably not even noticing the change the scale).
Those having already updated to 7.2 will msot probably quickly update to any maintenance release.

Maybe do a quick fix & postpone deeper changes ensuring long-term avoidance of such problem until later?

torkelo (approved):

torkelo:
yea, agree, lets get this merged and released asap

Berbe:
I am not proficient _at all_ with the test suites your are using for this project.

However, wouldn't it be a good idea to cover these units with more tests in order to check to such a change of scale happens?
In our case, it could have been spotted by the change made to `public/app/plugins/panel/table-old/specs/renderer.test.ts` at the time, but since several eyes pairs did not catch it, might automation be key?

PR #27854, opened 2020-09-28
Docs: Add trace and exemplar to glossary

marcusolsson:

Adds Trace and Examplar to the glossary.

### DIFF
--- a/docs/sources/getting-started/glossary.md
+++ b/docs/sources/getting-started/glossary.md
@@ -29,6 +29,12 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
     </td>
   </tr>
   <tr>
+    <td style="vertical-align: top">Exemplar</td>
+    <td>
+      An exemplar is any data that serves as a detailed example of data used to create an aggregation. A common type of examplars is a trace that references an aggregated metric.
+    </td>
+  </tr>
+  <tr>
     <td style="vertical-align: top">Graph</td>
     <td>
       A commonly-used visualization that displays data as points, lines, or bars.
@@ -67,6 +73,12 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
     <td style="vertical-align: top">Time series</td>
     <td>
       A series of measurements, ordered by time. Time series are stored in data sources and returned as the result of a query.
+    </td>
+  </tr>
+  <tr>
+    <td style="vertical-align: top">Trace</td>
+    <td>
+      An observed execution path of a request through a distributed system.
     </td>
   </tr>
   <tr>
--- a/docs/sources/getting-started/glossary.md
+++ b/docs/sources/getting-started/glossary.md
@@ -31,7 +31,7 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
   <tr>
     <td style="vertical-align: top">Exemplar</td>
     <td>
-      An exemplar is any data that serves as a detailed example of data used to create an aggregation. A common type of examplar is a trace that references an aggregated metric.
+      An exemplar is any data that serves as a detailed example of one of the observations aggregated into a metric. An exemplar contains the observed value together with an optional timestamp and arbitrary labels, which are typically used to reference a trace.
     </td>
   </tr>
   <tr>
--- a/docs/sources/getting-started/glossary.md
+++ b/docs/sources/getting-started/glossary.md
@@ -25,7 +25,7 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
   <tr>
     <td style="vertical-align: top">Data source</td>
     <td>
-      A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional ones through plugins.
+      A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional data sources through plugins.
     </td>
   </tr>
   <tr>
--- a/docs/sources/getting-started/glossary.md
+++ b/docs/sources/getting-started/glossary.md
@@ -78,7 +78,7 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
   <tr>
     <td style="vertical-align: top">Trace</td>
     <td>
-      An observed execution path of a request through a distributed system.
+      An observed execution path of a request through a distributed system. For more information, refer to [What is Distributed Tracing?](https://opentracing.io/docs/overview/what-is-tracing/).
     </td>
   </tr>
   <tr>
--- a/docs/sources/getting-started/glossary.md
+++ b/docs/sources/getting-started/glossary.md
@@ -17,41 +17,60 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
 
 <table>
   <tr>
-    <td style = "vertical-align: top;">Dashboard</td>
-    <td>A set of one or more panels, organized and arranged into one or more rows, that provide an at-a-glance view of related information.</td>
-  </tr>
-  <tr>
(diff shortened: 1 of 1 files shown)


docs/sources/getting-started/glossary.md
```diff
@@ -28,6 +28,12 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
       A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional ones through plugins.
     </td>
   </tr>
+  <tr>
+    <td style="vertical-align: top">Exemplar</td>
+    <td>
+      An exemplar is any data that serves as a detailed example of data used to create an aggregation. A common type of examplars is a trace that references an aggregated metric.
```
joe-elliott: Some small suggestions.

"An exemplar is any data that " => "An exemplar is a piece of metadata that ... "

"type of examplars is a trace that references an aggregated metric" => "type of exemplars is a trace id that exemplifies a metric"

docs/sources/getting-started/glossary.md
```diff
@@ -69,6 +75,12 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
       A series of measurements, ordered by time. Time series are stored in data sources and returned as the result of a query.
     </td>
   </tr>
+  <tr>
+    <td style="vertical-align: top">Trace</td>
+    <td>
+      An observed execution path of a request through a distributed system.
```
joe-elliott, line 81: This is an extremely concise and good definition for a very complex idea.  Does it make sense to link to the OpenTracing "what is tracing" page for more details:

https://opentracing.io/docs/overview/what-is-tracing/

docs/sources/getting-started/glossary.md
marcusolsson: I think "piece of" is redundant, what about just "An examplar is metadata that ..."?

The latter suggestion, implies that an examplar is the _reference_ to the trace, rather the actual trace? Is this intended?

docs/sources/getting-started/glossary.md
marcusolsson: Makes sense 👍

docs/sources/getting-started/glossary.md
joe-elliott: > I think "piece of" is redundant, what about just "An examplar is metadata that ..."

Agree!

> The latter suggestion, implies that an examplar is the reference to the trace, rather the actual trace? Is this intended?

Yes.  The metadata attached is only a trace id.  The trace itself would need to be stored in a different system.

Also, heads up that you have misspelled "exemplars" as "examplars" in one spot.

marcusolsson: How about this then?:

> An exemplar is metadata that references a detailed example of data used to create an aggregation. A common type of exemplar is an ID that connects an aggregated metric with a trace for one of the request used in the aggregation.

joe-elliott: looks good to me!

docs/sources/getting-started/glossary.md
```diff
@@ -17,41 +17,72 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
...
-    <td>A set of one or more panels, organized and arranged into one or more rows, that provide an at-a-glance view of related information.</td>
+    <td style="vertical-align: top">Dashboard</td>
+    <td>
+      A set of one or more panels, organized and arranged into one or more rows, that provide an at-a-glance view of related information.
+    </td>
+  </tr>
+  <tr>
+    <td style="vertical-align: top">Data source</td>
+    <td>
+      A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional ones through plugins.
```
oddlittlebird, line 28: You might add a link to plugins using the last word as the link text.

docs/sources/getting-started/glossary.md
```diff
@@ -17,41 +17,72 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
...
+      Basic building block in Grafana, composed by a query and a visualization. Can be moved and resized within a dashboard.
+    </td>
   </tr>
-    <tr>
-    <td style = "vertical-align: top;">Panel</td>
-    <td>Basic building block in Grafana, composed by a query and a visualization. Can be moved and resized within a dashboard.</td>
+  <tr>
+    <td style="vertical-align: top">Plugin</td>
+    <td>
+      An extension of Grafana that allows users to provide additional functionality to enhance their experience. The types of plugins currently supported are:
```
oddlittlebird, line 52: Don't we have these defined somewhere else? It might be more efficient to link to the Plugins concept page. That way we don't have to maintain two lists. I'd say keep the first sentence, then say, "For more information, refer to LINK."

oddlittlebird (changes requested):
Had a couple questions.

docs/sources/getting-started/glossary.md
```diff
@@ -17,41 +17,72 @@ This topic lists words and abbreviations that are commonly used in the Grafana d
...
+    <td>
+      A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional ones through plugins.
+    </td>
   </tr>
   <tr>
-    <td style = "vertical-align: top;">Data source</td>
-    <td>A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional ones through plugins.</td>
+    <td style="vertical-align: top">Exemplar</td>
+    <td>
+      An exemplar is any data that serves as a detailed example of data used to create an aggregation. A common type of examplar is a trace that references an aggregated metric.
```
beorn7, line 34: Double "data" here. Also, an exemplar has a mandatory value, which is not mentioned here at all, and an optional timestamp. Proposal:

"An exemplar is any data that serves as a detailed example of one of the observations aggregated into a metric. An exemplar contains the observed value together with an optional timestamp and arbitrary labels, which are typically used to reference a trace."

beorn7 (commented):
Just talking about the exemplar part.

docs/sources/getting-started/glossary.md
marcusolsson: I'll fix this in a separate PR!

docs/sources/getting-started/glossary.md
marcusolsson: I'll fix this in a separate PR!

docs/sources/getting-started/glossary.md
marcusolsson: > one of the observations aggregated into a metric

This reads beautifully and clearly! ❤️ 

@joe-elliott suggested "metadata" rather than "data". Is it ok if I change it to:

> An exemplar is metadata that serves as a detailed example of one of the observations aggregated into a metric. An exemplar contains the observed value together with an optional timestamp and arbitrary labels, which are typically used to reference a trace.

beorn7: I would actually avoid the term "metadata". It appears to be more precise, but it might actually create misunderstandings because, in my experience, different groups of people in the metrics space have very different opinions about the line between "data" and "metadata". And exemplars are definitely in that grey area.

docs/sources/getting-started/glossary.md
oddlittlebird, line 28:
```suggestion
      A file, database, or service providing the data. Grafana supports a several data sources by default, and can be extended to support additional data sources through plugins.
```

oddlittlebird (approved):
One minor editing suggestion, then good to go!

beorn7 (approved):
Approved for exemplar definition (only).

Issue #24057, opened 2020-04-29
Expressions UI - UI improvements


kylebrandt:
- General:
  - Operation Drop down has form label
- Resample Node
  - "Rule" is really the "Interval". The units are currently strange. Change to number value and unit dropdown (will update backend units to take whatever frontend changes to).
  - "Series" is actually RefID - value has a $ in front of it, e.g. $B
  - Form for the whole Resample operation takes up too much vertical space
- Reduce Node
  - "Fields" is actually RefID, and just one of them.

Integrated Help:

- (?) or something next "+ Expression" brings up help modal, where can a lot of stuff explaining the general feature.
- Each Operation (Math, Resample, Reduce) has a short description text with each dropdown item
- Once operation is selected, (?) help for the operation. Most should not be that long, math will be long though.
- Some form fields, like "Downsample function" could use (?) to explain them.

Note: If there is a PR with the stubbing, or pointers to better form examples I can pretty much just copy, I should be able to take it from there. Currently using LegacyForms, so thinking maybe I should not iterate on this.

torkelo:
Please add labels to new issues otherwise it just more work for anyone doing issue triage

torkelo:
and if possible add to a project board

torkelo:
added to platform todo, but doubt we will get to in during beta

kylebrandt:
Closing this as we did some the things, and think it is better to make specific issues for whatever is most important to change in the UI for 7.4.

PR #27859, opened 2020-09-28
GEL: Editor UI updates

peterholmberg:
Using newly added Inline form styles instead of the "old" ones.

Part of #24057

image


kylebrandt (approved):
Thank you this is looking better :-)

PR #27862, opened 2020-09-28
Proposal: Declarative API for axes, scales and series configuration in Graph NG

dprokop:
This is a proposal of a different way of confugiring uPlot axes, series and scales. Instead of relying on the logic build in into Plot.tsx component, the config is now possible to be prepared by the consumer of the components. This means, that building a graph  will be possible in a declarative way:

```tsx
<UPlotGraph data={DataFrame}>
   <Scale time={true} scaleKey='scaleName' />
   <Axis scaleKey='scaleName' side={...} size={...} .../>

   // Single series display config:
    <SeriesGeometry>
      <Line stroke={...} width={} .../>
      <Points stroke={...} size={} .../>
      <Area fill={...} .../>
    </SeriesGeometry>

    //plugins. i.e. Tooltip, Context menu, etc
</UplotGraph>
```

For detailed usage see [graph3/GraphPanel.tsx](https://github.com/grafana/grafana/blob/b5af504839813bacce8422322da1b38a650e3835/public/app/plugins/panel/graph3/GraphPanel.tsx) component where the entire graph is built using field config.

Geometries introduced:
- `Line`
- `Point`
- `Area`
- `SeriesGeometry` wrapper: required when series needs to be rendered as "complex" geometry, i.e. line+points, points+area or similar

Todo: 
- [x] Bring back time zone support
  - [ ] Add test


torkelo:
Looks super nice. What is the difference between scale and axis and what is their relationship? See axis has a scaleKey

ryantxu:
@torkelo check: https://github.com/leeoniya/uPlot/tree/master/docs#series-scales-axes-grid

Scales:
> scales reflect the min/max ranges visible within the view. All view range adjustments such as zooming and pagination are done here. If not explicitly set via opts, scales are automatically initialized using the series config and auto-ranged using the provided data.


Axes
> axes render the ticks, values, labels and grid along their scale. Tick & grid spacing, value granularity & formatting, timezone & DST handling is done here.

torkelo:
Ah, thanks. That makes sense.

packages/grafana-ui/src/components/uPlot/geometries/configGetters.ts
```diff
@@ -0,0 +1,35 @@
+import { AreaProps, LineProps, PointProps } from './types';
+import tinycolor from 'tinycolor2';
+
+export const getAreaConfig = (props: AreaProps) => {
+  const fill = props.fill
+    ? tinycolor(props.color)
```
ryantxu, line 6: `getColorFromHexRgbOrName`?

ryantxu (approved):
looks good!

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -2,114 +2,73 @@ import 'uplot/dist/uPlot.min.css';
...
+    if (!plotInstance) {
       return;
     }
-    initPlot();
-  }, [currentPlotConfig]);
+    const data = preparePlotData(props.data);
+    pluginLog('uPlot core', false, 'updating plot data(throttled log!)');
+    // If config hasn't changed just update uPlot's data
+    plotInstance.setData(data);
+  }, [plotInstance, props.data]);
```
hugohaggmark, line 46: nit: best practice is to spread props that are used as dependencies in hooks. I mean `setData` & `props.data` in this case

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -123,7 +82,7 @@ export const UPlotChart: React.FC<PlotProps> = props => {
 
   // Memoize plot context
   const plotCtx = useMemo(() => {
-    return buildPlotContext(registerPlugin, canvasRef, props.data, plotInstance);
+    return buildPlotContext(registerPlugin, addSeries, addAxis, addScale, canvasRef, props.data, plotInstance);
   }, [registerPlugin, canvasRef, props.data, plotInstance]);
```
hugohaggmark, line 86: nit: same comment as above but should `addSeries, addAxis, addScale` be added to dependencies?

packages/grafana-ui/src/components/uPlot/geometries/Scale.tsx
```diff
@@ -0,0 +1,48 @@
...
+  useEffect(() => {
+    const config = getConfig();
+    getUpdateConfigRef()({ ...defaultScaleConfig, ...config });
+  }, [getConfig]);
+};
+
+export const Scale: React.FC<ScaleProps> = props => {
+  const { scaleKey, time } = props;
+
+  const getConfig = useCallback(() => {
```
hugohaggmark, line 36: nit: if this changes whenever props changes do we need to use `useCallback`

hugohaggmark (approved):
Such massive work, I really like this declarative API. Just small nits from me for you to consider

packages/grafana-ui/src/components/uPlot/Plot.tsx
dprokop: yes, great catch @hugohaggmark!

packages/grafana-ui/src/components/uPlot/geometries/Scale.tsx
dprokop: not really, great catch!

packages/grafana-ui/src/components/uPlot/Plot.tsx
dprokop: My thinking here - setData never changes independently from plotInstance, that's why I used the entire instance as a dependency, to keep reasoning about this hook simpler. But if you do not agree that this is easier to grasp I'm happy to update :)

hugohaggmark: That's ok for setData, but maybe props.data should be deconstructed and used instead? No biggie at all

PR #27868, opened 2020-09-28
@grafana-toolkit: promise improvements

stevenvachon:
Perhaps this will make it easier to debug [npm "next" package release errors](https://drone.grafana.net/grafana/grafana/3340/1/27), or even fix it outright if we're lucky.

See commit message(s).

### DIFF
--- a/packages/grafana-toolkit/src/cli/utils/useSpinner.ts
+++ b/packages/grafana-toolkit/src/cli/utils/useSpinner.ts
@@ -1,8 +1,8 @@
 import ora = require('ora');
 
-type FnToSpin<T> = (options: T) => Promise<void>;
+type FnToSpin<T> = (options: T) => Promise<any>;
 
-export const useSpinner = <T = any>(spinnerLabel: string, fn: FnToSpin<T>, killProcess = true) => {
+export function useSpinner<T = void>(spinnerLabel: string, fn: FnToSpin<T>, killProcess = true) {
   return async (options: T) => {
     const spinner = ora(spinnerLabel);
     spinner.start();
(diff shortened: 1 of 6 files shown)


packages/grafana-toolkit/src/cli/utils/useSpinner.ts
```diff
@@ -1,8 +1,8 @@
 import ora = require('ora');
 
-type FnToSpin<T> = (options: T) => Promise<void>;
+type FnToSpin<T> = (options: T) => Promise<any>;
 
-export const useSpinner = <T = any>(spinnerLabel: string, fn: FnToSpin<T>, killProcess = true) => {
+export function useSpinner<T = void>(spinnerLabel: string, fn: FnToSpin<T>, killProcess = true) {
```
ryantxu, line 5: IIUC, this is the only real change -- the rest just come from it?  In general, it seems better to export functions rather than constant implementations of a function!

:+1:

ryantxu (approved):

packages/grafana-toolkit/src/cli/utils/useSpinner.ts
stevenvachon: The timing change at https://github.com/grafana/grafana/pull/27868/files#diff-6e99f8c06e5db7f6d569cb8aec3bc74bR66 could be something more real as well

ryantxu: great -- for PRs like this where there are only a couple lines with meaningful changes, can you point them out in the PR with a comment?

stevenvachon: #27906

PR #27870, opened 2020-09-28
Chore: Convert internal grafana datasource from angular to react

ryantxu:
I am hoping to add some interesting out-of-the-box live streaming features.  It makes sense to add this to the builtin "-- Grafana --" 

This PR moves things to react so we can add new features in a later PR.

For now, this PR will match the existing behavior of ignoring multiple targets, but it should behave more like a normal datasource
image


public/app/plugins/datasource/grafana/types.ts
```diff
@@ -0,0 +1,34 @@
+import { AnnotationQuery, DataQuery, SelectableValue } from '@grafana/data';
+
+//----------------------------------------------
+// Query
+//----------------------------------------------
+
+export enum GrafanaQueryType {
+  RandomWalk = 'randomWalk',
+  RandomStream = 'randomStream',
+  HostMetrics = 'hostmetrics',
```
ryantxu, line 10: Not implemented yet... could remove now, but also not used or referenced :shrug:

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -60,7 +73,7 @@ class GrafanaDatasource extends DataSourceApi<any> {
       delete params.tags;
     } else {
       // require at least one tag
-      if (!_.isArray(options.annotation.tags) || options.annotation.tags.length === 0) {
+      if (!isArray(annotation.tags) || annotation.tags.length === 0) {
```
hugohaggmark, line 76: nit: maybe use builtin Array.isArray function instead
```suggestion
      if (!Array.isArray(annotation.tags) || annotation.tags.length === 0) {
```

hugohaggmark (approved):
Looks great!

PR #27875, opened 2020-09-29
Transformations: improve the reduce transformer

ryantxu:
The current `reduce` transformer was designed to have output that matches the old angular table format; this structure is hard to use when using it for other types of visualization and in particular it loses all label information.

This PR adds an additional mode to the reduce transformer that just operations on the field:

image

image

TODO:
- [x] testing
- [ ] docs


ryantxu:
cc: @mikempx

torkelo:
While your looking at this reducer, would be interesting to explore how we can solve the series to rows scenario but keep the time series values in a single field value (so we can have a sparkline cell mode in the table). So

kylebrandt:
The one result per frame is how GEL currently does things:

image

I think it is interesting to note that I think both the one dataframe per thing, and the one field per thing in a single frame, don't always visualize well (in the single frame, case with too many columns).

In terms of visualization, I think it is best to take the above and pass it through what I think "Labels to Fields" does:

image

(also might look nice as a 90 degree rotated bar graph that you can scroll down, sorted by the value.)

packages/grafana-data/src/transformations/transformers/reduce.ts
```diff
@@ -33,14 +37,55 @@ export const reduceTransformer: DataTransformerInfo<ReduceTransformerOptions> =
...
+              processed.push({
+                ...series,
+                fields,
+                length: 1, // always one row
+              });
+            }
+          }
+
+          return processed;
+        }
```
torkelo, line 86: This reducer function is 120 lines of pretty dense code. Would be nice to be if we could break it up a bit. And nitpick: I struggle reading large code blocks like this that has no breaks between control statement, it all mixes and together and I mistake statements to look like they form a single expression but are really completely separate expressions.

torkelo (changes requested):
Looks good! Think this reducer needs a minor refactoring (extract method), to make it a bit more readable. And this new logic needs some tests :)

ryantxu:
@torkelo, I split this into two smaller functions and added a simple test.  I will still add docs before I hope to merge, but check if the implementation looks better.

torkelo (approved):
Yes, a lot easier to navigate and read the code now!

PR #27879, opened 2020-09-29
Docs: generate packages docs before syncing to website

mckn:
To prevent having generated markdown docs in the repository we would like to generate the docs when syncing to the website repository. The benefits of doing this is that we will always keep the latest version of the packages docs in the website and when searching the repository for code you won't get matches in the documentation.


hugohaggmark (approved):
I think this looks good although I don't know all the intricate details, would be a great topic for a frontend/grafana weekly to present the docs flow :)

dprokop (approved):

robbymilo (approved):
@mckn Nice solution here.

PR #27880, opened 2020-09-29
Docs: remove checked-in generated package api docs.

mckn:
Since the following PR #27879 will generate the packages documentation from the packages gode when syncing to the website repository we do not need to keep this documentation in the grafana repository.


oddlittlebird:
Okay, so you're removing the docs, but you still need docs. Where will the documentation live?

mckn:
> Okay, so you're removing the docs, but you still need docs. Where will the documentation live?

They will live in `https://grafana.com/docs/` but instead of having the auto generated docs in this repository we will auto generate them when syncing this docs to the website.

oddlittlebird:
Will the page addresses be the same?

mckn:
> Will the page addresses be the same?

Yes, everything will be exactly the same. The only difference is that we will do this automatically. We will also start to track so we don't add new code without documentation. If anyone will do that we will fail the build so the PR can't be merged.

oddlittlebird (approved):
Sounds good to me.

DanCech (approved):

robbymilo (approved):

dprokop (approved):
great!

PR #27881, opened 2020-09-29
Add docs for TagsInput

jessover9000:


packages/grafana-ui/src/components/TagsInput/TagsInput.mdx
```diff
@@ -0,0 +1,8 @@
+import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
```
Clarity-89, line 1: Adding `mdx` files on their own won't enable docs, they also need to be added to the story file. Like so: https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Button/Button.story.tsx#L13-L17 (also note the import at the top).

Clarity-89 (changes requested):

packages/grafana-ui/src/components/TagsInput/TagsInput.mdx
jessover9000: ohhh that's good to know, probably Tobias mentioned that half a year ago and I forgot it again :D Thanks!

Clarity-89: Yeah, I was wondering earlier why do we have a few `mdx` files not connected to stories :D

packages/grafana-ui/src/components/TagsInput/TagsInput.mdx
```diff
@@ -0,0 +1,8 @@
...
+import { TagsInput } from './TagsInput';
+
+<Meta title="MDX|TagsInput" component={TagsInput} />
+
+# TagsInput
+
+A set of an input field and next to it that allows the user to add new tags. The added tags are previewed under the input and can be removed there by clicking the "X" icon. You can customize the width of the input.
```
Clarity-89, line 8:
```suggestion
A set of an input field and a button next to it that allows the user to add new tags. The added tags are previewed under the input and can be removed there by clicking the "X" icon. You can customize the width of the input.
```
Seems like a word is missing here, i suppose it's `button`?

Clarity-89 (approved):
Looks good, one minor comment.

Issue #19964, opened 2019-10-23
Make it possible to disable builtin authentication


mvanrompuy:
**What happened**: Our Grafana instance was using local authentication (auth.basic). I switched it over to use LDAP (and disable auth.basic). After the change I can successfully login using the LDAP password. I also notice that in the user list, my username is now showing an ldap badge. I then tried to login using my old password, which to my surprise was still working.

**What you expected to happen**: When activating ldap (and disabling basic authentication), I would expect the old password to be no longer usable, only the LDAP password should be usable.

Keeping these 'old' passwords active is a security risk as the user is not able to update them (the password change functionality is automatically disabled when ldap is enabled) and there are now two passwords which will both give access to the account

**How to reproduce it (as minimally and precisely as possible)**:

- Setup Grafana with auth.basic enabled and create a user with password 'abc123'.
- Disable auth.basic and enable auth.ldap
- Login with the ldap password (should work)
- Login with the password 'abc123' (will work, but shouldn't)

**Anything else we need to know?**:

**Environment**:
- Grafana version: v6.4.3 (3a2bfb7)
- Data source type & version: Prometheus
- OS Grafana is installed on: Centos 7 (docker image grafana/grafana:6.4.3)
- User OS & Browser: Chrome 78
- Grafana plugins: vonage-status-panel,grafana-piechart-panel,jdbranham-diagram-panel
- Others: Auth settings

auth |  
-- | --
api_key_max_seconds_to_live | -1
disable_login_form | false
disable_signout_menu | false
login_cookie_name | grafana_session
login_maximum_inactive_lifetime_days | 7
login_maximum_lifetime_days | 30
oauth_auto_login | false
signout_redirect_url |  
token_rotation_interval_minutes | 10
**auth.anonymous** |  
enabled | false
org_name | Main Org.
org_role | Viewer
**auth.basic** |  
enabled | false
**auth.generic_oauth** |  
allow_sign_up | true
allowed_domains |  
allowed_organizations |  
api_url |  
auth_url |  
client_id | some_id
client_secret | ************
email_attribute_name | email:primary
email_attribute_path |  
enabled | false
hosted_domain |  
name | OAuth
scopes | user:email
send_client_credentials_via_post | false
team_ids |  
tls_client_ca |  
tls_client_cert |  
tls_client_key |  
tls_skip_verify_insecure | false
token_url |  
**auth.github** |  
allow_sign_up | true
allowed_domains |  
allowed_organizations |  
[...]

PR #27886, opened 2020-09-29
Profile: Prevent OAuth users from changing user details or password 

### LINKED ISSUE #18819: Disable user from editing user profile
Possibility to disabling a user from edit his/her User Profile.

**Why is this needed**:
Otherwise a user can “give” the account to a user without knowledge to the server owner.

dupondje:

This PR adds an option to hide the Change Password / Profile links in Grafana for OAuth users.
As normally grafana isn't the place to change those users credentials.

Fixes #18819

### DIFF
--- a/public/app/features/profile/ChangePasswordForm.tsx
+++ b/public/app/features/profile/ChangePasswordForm.tsx
@@ -1,20 +1,27 @@
 import React, { FC } from 'react';
 import config from 'app/core/config';
+import { UserDTO } from 'app/types';
 import { Button, LinkButton, Form, Field, Input, HorizontalGroup } from '@grafana/ui';
 import { ChangePasswordFields } from 'app/core/utils/UserProvider';
 import { css } from 'emotion';
 
 export interface Props {
+  user: UserDTO;
   isSaving: boolean;
   onChangePassword: (payload: ChangePasswordFields) => void;
 }
 
-export const ChangePasswordForm: FC<Props> = ({ onChangePassword, isSaving }) => {
-  const { ldapEnabled, authProxyEnabled } = config;
+export const ChangePasswordForm: FC<Props> = ({ user, onChangePassword, isSaving }) => {
+  const { ldapEnabled, authProxyEnabled, disableLoginForm } = config;
+  const authSource = user.authLabels?.length && user.authLabels[0];
 
   if (ldapEnabled || authProxyEnabled) {
     return <p>You cannot change password when ldap or auth proxy authentication is enabled.</p>;
   }
+  if (authSource && disableLoginForm) {
+    return <p>Password cannot be changed here!</p>;
+  }
+
   return (
     <div
       className={css`
--- a/public/app/core/utils/UserProvider.tsx
+++ b/public/app/core/utils/UserProvider.tsx
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import { getBackendSrv } from '@grafana/runtime';
-import { User, Team, UserOrg, UserSession } from 'app/types';
+import { UserDTO, Team, UserOrg, UserSession } from 'app/types';
 import { config } from 'app/core/config';
 import { dateTimeFormat, dateTimeFormatTimeAgo } from '@grafana/data';
 
@@ -45,12 +45,12 @@ export interface Props {
     teams: Team[],
     orgs: UserOrg[],
     sessions: UserSession[],
-    user?: User
+    user?: UserDTO
   ) => JSX.Element;
 }
 
 export interface State {
-  user?: User;
+  user?: UserDTO;
   teams: Team[];
   orgs: UserOrg[];
   sessions: UserSession[];
--- a/public/app/features/profile/ChangePasswordPage.tsx
+++ b/public/app/features/profile/ChangePasswordPage.tsx
@@ -1,10 +1,12 @@
-import React, { PureComponent } from 'react';
+import React, { FC } from 'react';
 import { hot } from 'react-hot-loader';
 import { connect } from 'react-redux';
-import { StoreState } from 'app/types';
+import { config } from '@grafana/runtime';
+import { LoadingPlaceholder } from '@grafana/ui';
+import { UserDTO, Team, UserOrg, UserSession, StoreState } from 'app/types';
 import { NavModel } from '@grafana/data';
 import { getNavModel } from 'app/core/selectors/navModel';
-import { UserProvider } from 'app/core/utils/UserProvider';
+import { UserProvider, UserAPI, LoadingStates } from 'app/core/utils/UserProvider';
 import Page from 'app/core/components/Page/Page';
 import { ChangePasswordForm } from './ChangePasswordForm';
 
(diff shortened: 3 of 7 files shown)


pkg/api/api.go
```diff
@@ -136,7 +137,7 @@ func (hs *HTTPServer) registerRoutes() {
 
 			userRoute.Get("/auth-tokens", Wrap(hs.GetUserAuthTokens))
 			userRoute.Post("/revoke-auth-token", bind(models.RevokeAuthTokenCmd{}), Wrap(hs.RevokeUserAuthToken))
-		})
+		}, reqSignedInNonOAuth)
```
sakjur, line 140: This looks like it's applied too widely. There are a lot of endpoints in this group that shouldn't be deactivated when you want to prevent the user from setting a password.

pkg/api/api.go
```diff
@@ -32,9 +33,9 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/invite/:code", hs.Index)
 
 	// authed views
-	r.Get("/profile/", reqSignedIn, hs.Index)
-	r.Get("/profile/password", reqSignedIn, hs.Index)
-	r.Get("/profile/switch-org/:id", reqSignedIn, hs.ChangeActiveOrgAndRedirectToHome)
+	r.Get("/profile/", reqSignedInNonOAuth, hs.Index)
+	r.Get("/profile/password", reqSignedInNonOAuth, hs.Index)
+	r.Get("/profile/switch-org/:id", reqSignedInNonOAuth, hs.ChangeActiveOrgAndRedirectToHome)
```
sakjur, line 38:
```suggestion
	r.Get("/profile/", reqSignedIn, hs.Index)
	r.Get("/profile/password", reqSignedInNonOAuth, hs.Index)
	r.Get("/profile/switch-org/:id", reqSignedIn, hs.ChangeActiveOrgAndRedirectToHome)
```

I don't understand why the user shouldn't be allowed to switch organization or change their own theme, time zone, and home dashboard just because they're not supposed to set a password for themselves.

Changing this would imply re-adding the profile option to the sidebar as well.

sakjur (changes requested):
Thank you for your PR!

I haven't tested this out, so it's possible that it works in some way that I failed to see when reading the code, but I'm not sure how this option relates to OAuth specifically.
Or maybe it started out that way and pivoted into "prevent access to setting the user's display name/user name/email/password"? (if that's the case you should remove `OAuth` from the option and middleware names)

There are also a number of side-effects that you should either make explicit and explain why they're there or remove. See my comments for more on that.

oddlittlebird:
@sakjur, please let me know when you want me to look at docs. I'll hold off until you finish your review.

pkg/middleware/auth.go
```diff
@@ -137,3 +137,25 @@ func SnapshotPublicModeOrSignedIn() macaron.Handler {
...
+	return func(c *models.ReqContext) {
+		if !setting.OAuthHideProfile {
+			_, err := c.Invoke(ReqSignedIn)
+			if err != nil {
+				c.JsonApiErr(401, "Unauthorized", nil)
+			}
+			return
+		}
+
+		if c.UserToken != nil {
```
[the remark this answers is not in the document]
dupondje: Hmm you could be correct! The question is then how to properly find out (in this call) if the user is a oauth user or not?

roidelapluie:
Is this a fix for Is this a duplicate of https://github.com/grafana/grafana/issues/19964 ?

dupondje:
@roidelapluie : not really, Its not that builtin auth needs to be disabled. Its just that OAuth users should not get confused with a change-password tab in Grafana that is useless / not the right place to change their credentials.

roidelapluie:
> @roidelapluie : not really, Its not that builtin auth needs to be disabled. Its just that OAuth users should not get confused with a change-password tab in Grafana that is useless / not the right place to change their credentials.

You can login with oauth and still have a local account

roidelapluie:
OK, I see now. I think that  #19964  would have the same effect as this PR but for *all* users.

aknuds1:
We're discussing internally whether this is the right approach or not.

torkelo:
I feel like this is not the right approach, we should follow the same approach as the admin user page. There we just disable editing / changing some fields (for example for an ldap user).

I do not have a screenshot where the fields are disabled (but the Edit link is missing replaced with a lock and a notice on why). 
Screenshot from 2020-09-30 16-13-01

dupondje:
@torkelo : I can live with the idea indeed!

Hide some things from the profile page + hide the Change Password tab would be fine also.

Just need to find a proper way to find out a user is OAuth.
As c.UserToken seems to be incorrect. Any idea's?

Then I adjust my patch.

sakjur:
To add to @torkelo's example, here's how it looks on the admin page in Grafana Enterprise when your user is created via SAML:

Screen Shot 2020-09-30 at 16 32 02

dupondje:
Some other way to do it :)
But does not seem to work correctly yet

dupondje:
The latest review works correctly :)

dupondje:
@torkelo @sakjur @aknuds1 : Can somebody check this please?

torkelo:
Sorry for slow review & feedback, will try to test this out next week

public/app/features/profile/ChangePasswordForm.tsx
```diff
@@ -1,20 +1,27 @@
...
-export const ChangePasswordForm: FC<Props> = ({ onChangePassword, isSaving }) => {
+export const ChangePasswordForm: FC<Props> = ({ user, onChangePassword, isSaving }) => {
   const { ldapEnabled, authProxyEnabled } = config;
+  const authSource = user.authLabels?.length && user.authLabels[0];
 
   if (ldapEnabled || authProxyEnabled) {
     return <p>You cannot change password when ldap or auth proxy authentication is enabled.</p>;
   }
+  if (authSource) {
+    return <p>Password cannot be changed here!</p>;
```
torkelo, line 22: Can we remove this? We have no way currently to remove auth sources, so with this change there would be no way to go back to user name + password after disabling the oauth.

torkelo (commented):
Think this looks ok to merge, just not sure about the change to ChangePasswordForm

public/app/features/profile/ChangePasswordForm.tsx
dupondje: I guess we just need a better if then?
Cause the thing is we want to hide change password in Grafana for users logged in via OAuth. As it will just give confusion if they are able to change the password there, but it will not do anything..

torkelo: but the condition needs to be linked to disableLoginForm, not just the authSource

ms-novatti:
Why has this feature been tied in to disabling the login form? (Which disables the local login only, OAuth login option still appears on the login page).
For my use case I want to keep access to the local Admin user and also stop the OAuth users from changing their profile or password. This appears to be impossible in the release version.

sakjur:
@ms-novatti Torkel answered that question in https://github.com/grafana/grafana/pull/27886#discussion_r576629006

From taking a quick glance at this, I suspect you're correct and that your use case isn't supported by the current version of Grafana. In this case, it seems to be intentional (for better or worse), you should submit an enhancement request if you're interested in seeing this behavior changed. [Link to submitting a new feature request](https://github.com/grafana/grafana/issues/new?assignees=&labels=type%3A+feature+request&template=2-feature_request.md). I don't think we're very likely to change this behavior short-term, the amount of variables within the authentication scope is already massive and introducing more variables is not how we want to determine what a user can and can't do in the future.

PR #27887, opened 2020-09-29
Build: fixing failing release packaged build due to uPlot dependency not properly setup

mckn:
Currently master is failing due to a dependency on `app/public/plugins/panel/graph3` and uPlot styling not being configured properly. This PR is resolving those issues.


dprokop (approved):
Thanks, great catch!

mckn:
> Thanks, great catch!

I still have one issue to resolve

PR #27906, opened 2020-09-29
@grafana/toolkit: cleanup

stevenvachon:
See commit message(s).

Main edits:
* https://github.com/grafana/grafana/pull/27906/files#diff-ac6b234d1b9d086337b392292b962704R3

### DIFF
--- a/packages/grafana-toolkit/src/cli/utils/useSpinner.ts
+++ b/packages/grafana-toolkit/src/cli/utils/useSpinner.ts
@@ -1,20 +1,16 @@
-import ora = require('ora');
+import ora from 'ora';
 
-type FnToSpin<T> = (options: T) => Promise<any>;
-
-export function useSpinner<T = void>(spinnerLabel: string, fn: FnToSpin<T>, killProcess = true) {
-  return async (options: T) => {
-    const spinner = ora(spinnerLabel);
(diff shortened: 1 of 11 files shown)


packages/grafana-toolkit/src/cli/utils/useSpinner.ts
```diff
@@ -1,20 +1,16 @@
...
-    try {
-      await fn(options);
-      spinner.succeed();
-    } catch (e) {
-      console.trace(e); // eslint-disable-line no-console
-      spinner.fail(e.message || e);
-      if (killProcess) {
-        process.exit(1);
-      }
+export const useSpinner = async (label: string, fn: () => Promise<any>, killProcess = true) => {
```
ryantxu, line 3: when do you export const vs export function?  

I know the const stuff is useful for binding `this` but any use here?

ryantxu (approved):

packages/grafana-toolkit/src/cli/utils/useSpinner.ts
stevenvachon: TypeScript generics don't support arrow functions, so I use them then as well. I give preference to arrow functions mostly.

PR #27907, opened 2020-09-29
Toolkit: implement plugin signing in grafana-toolkit

DanCech:
This PR adds the ability to sign plugins via grafana-toolkit

Note: this is dependent on grafana.com endpoints which are not yet in production, but I wanted to get this PR opened for feedback.

@ryantxu @jhunthrop


packages/grafana-toolkit/src/cli/index.ts
```diff
@@ -176,6 +177,19 @@ export const run = (includeInternalScripts = false) => {
       });
     });
 
+  program
+    .command('plugin:sign')
+    .option('--signatureType <type>', 'Signature Type')
+    .option('--rootUrls <urls...>', 'Root URLs')
+    .description('Create a plugin signature')
```
ryantxu, line 184: the cli should maybe also accept `GRAFANA_API_KEY` and the help should say something about it

packages/grafana-toolkit/src/cli/index.ts
```diff
@@ -195,11 +209,14 @@ export const run = (includeInternalScripts = false) => {
 
   program
     .command('plugin:ci-package')
-    .option('--signing-admin', 'Use the admin API endpoint for signing the manifest.', false)
+    .option('--signatureType <type>', 'Signature Type')
+    .option('--rootUrls <urls...>', 'Root URLs')
```
ryantxu, line 213: It does not hurt to add here, but I don't think it can be used outside our internal CI now anyway

ryantxu:
looks good!   We will need some careful attention to the command line help, but other than that this looks close

packages/grafana-toolkit/src/plugins/manifest.ts
```diff
@@ -0,0 +1,87 @@
...
+    version: pluginJson.info.version,
+    files: {},
+  } as ManifestInfo;
+
+  for await (const p of walk(dir, dir)) {
+    if (p === MANIFEST_FILE) {
+      continue;
+    }
+
+    if (/\.(png|gif|svg)$/.test(p)) {
```
ryantxu, line 41: FYI: this filter was a bit add hoc.  I think we should avoid signing large files that do not add any real security.

packages/grafana-toolkit/src/cli/index.ts
DanCech: Not sure that we want to encourage users passing the api key on the command line, which is why I was following the environment variable pattern from ci-build.  We could support passing the location of a file to read it from maybe, do we have precedent for that anywhere else?

packages/grafana-toolkit/src/plugins/manifest.ts
DanCech: Hashing the files is pretty fast since it's just a single sha256sum (rather than a deliberately-expensive password hashing algorithm), so I'd like to err on the side of protecting as many files as possible.  This list of exclusions was just taken for the current code, but there is an argument to be made that we should just hash all files in the plugin.

packages/grafana-toolkit/src/cli/index.ts
DanCech: Yeah, this is more to keep things consistent than anything else.

packages/grafana-toolkit/src/plugins/manifest.ts
woodsaj: If you dont hash all of the files in the plugin folder, Grafana wont be able to verify that the plugin folder only contains files that should be present.  Untrusted files in the plugin folder are still a security issue, as a malicious user could link directly to them, allowing them to trick a user into running malicious code in the browser.

So, we should hash all files.

packages/grafana-toolkit/src/cli/index.ts
woodsaj: i agree with Dan, putting api keys in command line args isnt a great idea.
reading from an ENV var is going to be the most common approach used from CI/CD tooling.  For those building and signing locally, a config file would be good. we could allow the path to the file to be set, with a default of `~/.local/share/grafana-toolkit/config` or similar

A future enhancement could be to support a login command that uses an oauth flow, similar to how Google's `gcloud` or Azure's `az` commands work.
ie, users run
`grafana-toolkit --login`, which then opens grafana.com in a browser, authenticates the user and then passes back a token to grafana-toolkit that it can persist in the config file.

packages/grafana-toolkit/src/plugins/manifest.ts
ryantxu: +1 to change it to all files now.  We can have signatures with type 2.0+ require all files.

DanCech: works4me, will update

ryantxu: we may also want to look at what we do with grafana itself as it has all the same issues!

DanCech: yeah I think that can be a 2-stage approach, for v2+ signatures we can require that all files are signed, while leaving the v1 behavior alone, then at some point we remove support for v1 signatures.

DanCech:
@ryantxu @dprokop the grafana.com side of this is now live, if you could give it a final once-over that would be great.

ryantxu (approved):
looks good -- thanks @DanCech

PR #27909, opened 2020-09-29
Stackdriver: fix project name change regression

### LINKED ISSUE #27855: 7.2: stackdriver query render the wrong project
**What happened**:
The project selector inside Stackdriver query always render the default one, no matter of project is really use

To see the current project selected

1. Create a stackdriver datasource (probably with GCE authentification)
2. Create panel with the DS
3. Create a query and try to select another project
4. Query send to stackdriver use the project selected, but the UI display the default project

**Environment**:
- Grafana version: 7.2.0
- Data source type & version: stackdriver
- OS Grafana is installed on: kubernetes + official helm chart
- User OS & Browser: Chrome 85.0.4183.102 + MacOS 10.14.6

zoltanbedi:
This PR fixes a regression created in 7.2 that not allowed to change the project in google cloud monitoring data source. Also fixes an issue with legacy select component that it didn't show the noOptionsMessage. Also fixes the link in the config editor because it navigated to 404.

Fixes #27855

### DIFF
--- a/packages/grafana-ui/src/components/Forms/Legacy/Select/Select.tsx
+++ b/packages/grafana-ui/src/components/Forms/Legacy/Select/Select.tsx
@@ -154,7 +154,7 @@ export class Select<T> extends PureComponent<LegacySelectProps<T>> {
               onBlur={onBlur}
               openMenuOnFocus={openMenuOnFocus}
               maxMenuHeight={maxMenuHeight}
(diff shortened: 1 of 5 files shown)


papagian (approved):
I don't have context/opinion about the `noOptionsMessage` modification otherwise LGTM, I have tested it and it fixes the issue.

PR #27914, opened 2020-09-29
Docs: Update time-range-controls.md

oddlittlebird:
Adds a small but important bit of information based on a question from the public Slack.


papagian (approved):
LGTM

PR #27916, opened 2020-09-30
WIP: add real-time metrics to the internal grafana datasource

ryantxu:
Extends #27583

that PR is getting too big


packages/grafana-data/src/types/live.ts
```diff
@@ -85,28 +85,30 @@ export interface LiveChannelStatus {
...
+   * This will remain in the status until a new message is succesfully received from the channel
    */
   error?: any;
 }
 
 /**
  * @experimental
  */
-export interface LiveChannelJoinLeave {
+export interface LiveChannelPresenseEvent {
```
leeoniya, line 96: "Presence", presumably?

ryantxu: already fixed in master :)

ryantxu:
close in favor of the approach in #28022

PR #27917, opened 2020-09-30
Variables: Adds loading state and indicators

### LINKED ISSUE #17005: Provide feedback to user while updating template variables
Currently, there is no way to know if grafana is currently updating drop-down lists of template variables. This is especially a problem when a variable depends on the another: user selects var1, then var2 list gets updated but user can select outdated elements from var2

hugohaggmark:
new-variable-loading-state

This PR removed the old initLock logic and introduces loading state to all variable types. I also added some loading indicators in the UI. 
When the next PR that uses Observable for variable queries we can add cancellation of individual variable updates.

Fixes #17005

One way to test this is PR is to:

1. Start a slow proxy (30s delay on every data source request, the recording above is customized proxy using a 10s delay) with `make devenv sources=slow_proxy_mac,prometheus`
2. Import this [dashboard](https://gist.github.com/hugohaggmark/067a470d697f91828c9b282836765d07)


public/app/features/variables/types.ts
```diff
@@ -119,11 +119,32 @@ export interface SystemVariable<TProps extends { toString: () => string }> exten
   current: { value: TProps };
 }
 
+export enum VariableLoadingState {
+  NotStarted = 'Not started',
+  Fetching = 'Fetching',
+  Completed = 'Completed',
+  Failed = 'Failed',
```
torkelo, line 126: Can we use the LoadingState enum from grafana/data ? , only one that is not applicable is Streaming

hugohaggmark: Great suggestion, will lock into that

public/app/features/variables/state/actions.ts
```diff
@@ -222,13 +248,14 @@ export const processVariable = (
...
-        dispatch(resolveInitLock(toVariablePayload(variable)));
+        await dispatch(updateOptions(toVariableIdentifier(refreshableVariable)));
         return;
       }
     }
 
-    dispatch(resolveInitLock(toVariablePayload(variable)));
+    // for variables that aren't updated via url or refresh let's simulate the same state changes
+    dispatch(variableStateFetching(toVariablePayload(variable)));
+    dispatch(variableStateCompleted(toVariablePayload(variable)));
```
torkelo, line 258: Why do we need to go via fetching here can't we just emit completed action?

public/app/features/variables/state/actions.ts
```diff
@@ -582,3 +619,33 @@ export const cancelVariables = (
...
+  }
+
+  try {
+    dispatch(variableStateFetching(toVariablePayload(variableInState)));
+    await variableAdapters.get(variableInState.type).updateOptions(variableInState);
+    dispatch(variableStateCompleted(toVariablePayload(variableInState)));
+  } catch (error) {
+    console.error(error);
+    dispatch(variableStateFailed(toVariablePayload(variableInState, { error })));
+    dispatch(notifyApp(createErrorNotification('Templating', 'Error updating options: ' + error.message)));
```
torkelo, line 649: These two path !awaitUpdateOptions  & else looks very similar, one is done with await the other is not, can you explain the difference?

torkelo (approved):
Looks great! added two comments on things that confused me

public/app/features/variables/state/actions.ts
hugohaggmark: It might just be a copy-paste error, will check. Thanks for finding this

public/app/features/variables/state/actions.ts
hugohaggmark: Yeah, it's ugly. In the first case we don't want to resolve variableAdapters.get(...).updateOptions but in the other we want to. Any suggestions on how to solve this. Could be two separate methods I guess?

aocenas:
Just FYI some time ago I added this https://github.com/grafana/grafana/blob/eb734a2f6e939f29b6e3e92e8d3e0a0f6c1f6e4c/packages/grafana-ui/src/utils/useDelayedSwitch.ts#L18 to deal with loading indicator flickering. It basically delays the transitions a bit so if the load is really quick it won't even switch and when it's switched on it makes sure to be switched on at least some amount of time. I wanted to incorporate it into some spinner component but did not have time. But I think it can still be handy if somebody has time to use it.

public/app/features/variables/state/actions.ts
torkelo: not sure I see the difference, one path is returning a promise of variableAdapters.get(...).updateOption,  the other is using await. keyword, both should return a similar promise with the same effect

hugohaggmark: Hmm not sure, the first case is used in a Promise.all so I figured we wanted to run all Promises in that Promise.all in parallel. If I was to use await in within the Promise.all one wouldn't the Promise.all become sequential?

hugohaggmark: Hmm, I think I'm wrong in my thinking here

mckn (approved):
I think it looks great! Awesome work!

One thing that was a bit weird for me is that the example dashboard that you linked loaded super fast with the proper spinners attached to the different components. But I have another older dashboard that get stuck on the loading screen instead.

https://gist.github.com/mckn/4664db75773bcc081cc860e8b45274f5

hugohaggmark:
> I think it looks great! Awesome work!
> 
> One thing that was a bit weird for me is that the example dashboard that you linked loaded super fast with the proper spinners attached to the different components. But I have another older dashboard that get stuck on the loading screen instead.
> 
> https://gist.github.com/mckn/4664db75773bcc081cc860e8b45274f5

I think you need to change the data source to gdev-slow-prometheus :)

PR #27919, opened 2020-09-30
Drone: Don't build enterprise Windows installer unless downstream

aknuds1:
Drone: Fix enterprise master builds, since they shouldn't try to build the Windows installer unless triggered by OSS.

### DIFF
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -832,7 +832,7 @@ def get_windows_steps(edition, version_mode, is_downstream=False):
             ],
         },
     ]
-    if version_mode == 'master':
+    if version_mode == 'master' and (edition != 'enterprise' or is_downstream):
         installer_commands = [
             '$$gcpKey = $$env:GCP_KEY',
             '[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($$gcpKey)) > gcpkey.json',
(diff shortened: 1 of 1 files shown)


scripts/lib.star
```diff
@@ -842,13 +842,10 @@ def get_windows_steps(edition, version_mode, is_downstream=False):
             'rm gcpkey.json',
             'cp C:\\App\\nssm-2.24.zip .',
             '.\\grabpl.exe windows-installer --edition {} --build-id $$env:{}'.format(edition, build_no),
+            '$$fname = ((Get-Childitem grafana*.msi -name) -split "`n")[0]',
+            'gsutil cp $$fname gs://grafana-downloads/{}/{}/'.format(edition, version_mode),
+            'gsutil cp "$$fname.sha256" gs://grafana-downloads/{}/{}/'.format(edition, version_mode),
```
papagian, line 847: out of curiosity: why only in the 2nd command the source file should be quoted?

papagian (commented):
Generally LGTM, I have a question

scripts/lib.star
aknuds1: @papagian Good question - it's necessary in this case, to stop PowerShell from interpreting it as getting property `sha256` of `$fname`. Make sense?

papagian: ah you mean for making the concatenation working; it makes sense

papagian (approved):

PR #27920, opened 2020-09-30
Drone: Add Slack failure notifier

aknuds1:
Add Slack failure notifier to Drone. Not added for PR builds for now, since the Slack webhook could easily be leaked, until we protect the Drone configuration.

### DIFF
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -9,6 +9,7 @@ dockerize_version = '0.6.1'
 wix_image = 'grafana/ci-wix:0.1.1'
 
 def pr_pipelines(edition):
+    version_mode = 'pr'
     services = [
         {
             'name': 'postgres',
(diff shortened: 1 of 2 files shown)


scripts/lib.star
```diff
@@ -286,6 +315,11 @@ def init_steps(edition, platform, is_downstream=False, install_deps=True):
         },
     ]
 
+    if version_mode == 'master':
+        steps.append(slack_step(channel='grafana-ci-failures'))
```
marefr, line 319: I've already setup a webhook secret for the #grafana-ci slack channel. If you rather would like to use this one let me know and I'll delete my channel and webhook token

marefr (approved):
Great. Started working on this, but other things came in between so never opened any PR. LGTM, but please see comment

scripts/lib.star
aknuds1: @marefr Aha... I saw that secret and got confused where it came from, so I replaced it with my own. What channel had you set up already?

aknuds1: Merging now, since we can discuss which webhook to use separately :)

PR #27921, opened 2020-09-30
Instrumentation: Removes invalid chars from label names

bergquist:
skip instrumenting datasources if datasource
names only contains invalid chars

### DIFF
--- a/pkg/infra/metrics/metrics.go
+++ b/pkg/infra/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"errors"
+	"fmt"
 	"runtime"
 	"strings"
 
--- a/pkg/infra/metrics/metrics.go
+++ b/pkg/infra/metrics/metrics.go
@@ -1,7 +1,9 @@
 package metrics
 
 import (
+	"errors"
 	"runtime"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -588,4 +590,26 @@ func newCounterStartingAtZero(opts prometheus.CounterOpts, labelValues ...string
 	counter.Add(0)
 
 	return counter
+}
+
+// SanitizeLabelName removes all invalid chars from the label name.
+// If the label name is empty or contains only invalid chars, it
+// will return an error.
+func SanitizeLabelName(name string) (string, error) {
+	if len(name) == 0 {
+		return "", errors.New("label name cannot be empty")
+	}
+
+	out := strings.Builder{}
+	for i, b := range name {
+		if (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || (b >= '0' && b <= '9' && i > 0) {
+			out.WriteRune(b)
+		}
+	}
+
+	if out.Len() == 0 {
+		return "", errors.New("label name only contains invalid chars")
+	}
+
+	return out.String(), nil
 }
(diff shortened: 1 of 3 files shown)


pkg/infra/metrics/metrics.go
```diff
@@ -589,3 +591,25 @@ func newCounterStartingAtZero(opts prometheus.CounterOpts, labelValues ...string
...
+
+	out := strings.Builder{}
+	for i, b := range name {
+		if (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || b == '_' || (b >= '0' && b <= '9' && i > 0) {
+			out.WriteRune(b)
+		}
+	}
+
+	if out.Len() == 0 {
+		return "", errors.New("label name only contains invalid chars")
```
aknuds1, line 611: IMO, including the erroneous name would help in debugging?
```suggestion
		return "", fmt.Errorf("label name only contains invalid chars: %q", name)
```

pkg/infra/metrics/metrics_test.go
```diff
@@ -0,0 +1,29 @@
...
+		{input: "", expected: "", err: true},
+		{input: ";;;", expected: "", err: true},
+	}
+
+	for _, tc := range testcases {
+		got, err := SanitizeLabelName(tc.input)
+		if tc.err {
+			assert.Error(t, err)
+		} else {
+			assert.Equal(t, tc.expected, got)
```
aknuds1, line 26: I would also make sure that `err` didn't happen:
```suggestion
			require.NoError(t, err)
			assert.Equal(t, tc.expected, got)
```

aknuds1 (approved):
LGTM, but see comments.

aknuds1:
Should we add the cherry-pick needed label? Also didn't know we were making more 7.1.x releases.

pkg/infra/metrics/metrics_test.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func TestLabelNameSanitization(t *testing.T) {
+	testcases := []struct {
+		input    string
+		expected string
+		err      bool
+	}{
+		{input: "job", expected: "job"},
+		{input: "job._loal['", expected: "job_loal"},
```
marefr, line 16: Does it work with spaces as well? Personally would replace spaces with underscore (_). WDYT?

```suggestion
		{input: "job._loal['", expected: "job_loal"},
		{input: "Data source'", expected: "Datasource"},
```

pkg/models/datasource_cache.go
```diff
@@ -72,7 +73,14 @@ type dataSourceTransport struct {
 
 func instrumentRoundtrip(datasourceName string, next http.RoundTripper) promhttp.RoundTripperFunc {
 	return promhttp.RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
-		datasourceLabel := prometheus.Labels{"datasource": datasourceName}
+		datasourceLabelName, err := metrics.SanitizeLabelName(datasourceName)
```
marefr, line 76: Maybe we should lowercase this label?

marefr (changes requested):
LGTM, but added a couple comments/suggestions

pkg/models/datasource_cache.go
bergquist: Uppercase is allowed in Prometheus https://github.com/prometheus/common/blob/master/model/labels.go#L86

Lowercasing would increase the chance of collision so I would prefer to not do that.

pkg/infra/metrics/metrics_test.go
bergquist: Agree.

marefr (approved):
LGTM

PR #27925, opened 2020-09-30
Drone: Fix publishing of next NPM packages

aknuds1:
Try to fix publishing of next NPM packages. According to @mckn, this should be the way to do it.

The new version of the build pipeline tool builds the NPM packages together with the rest of the frontend.


scripts/circle-release-next-packages.sh
```diff
@@ -64,23 +64,7 @@ else
 
   for PACKAGE in "${PACKAGES[@]}"
   do
-    start=$(date +%s%N)
-    yarn workspace @grafana/"${PACKAGE}" run build
-    runtime=$((($(date +%s%N) - start)/1000000))
-    if [ "${DRONE_BRANCH}" == "master" ]; then
-      exit_if_fail ./scripts/ci-metrics-publisher.sh "grafana.ci-buildtimes.${DRONE_STEP_NAME}.$PACKAGE=$runtime"
```
mckn, line 71: In the new version of `grabpl` do we publish the same metrics if the build fails on master?

mckn (approved):
One small question but overall I think this looks great!

There are a couple of reasons why we want to build the packages in a couple of different scenarios:

1. To be able to generate the latest version of the packages-api docs we need to build the packages so we can extract the proper meta data that is needed to generate the packages-api docs.

2. At multiple times we have merged PRs that are green but they fails in master because we can't publish the npm packages. The reason could be a lot of different things but we do not detect it in the PR because we currently do not build the npm packages. By doing this we will detect it before we merge a PR to master.

3. When building the website we need to have all the documentation files in place so that part of the build can verify that all the links is connected properly. So we need to generate the docs before building the website and to be able to do that we need to build the npm packages.

scripts/circle-release-next-packages.sh
aknuds1: @mckn Doesn't this logic publish build times per package? I don't really see how eventual failure matters.

That said, it isn't currently implemented in the build pipeline tool. I have a task to port build metrics to it.

mckn: hmm, I agree. @dprokop do you remember why we shipped these metrics?

scripts/circle-release-next-packages.sh
```diff
@@ -64,23 +64,7 @@ else
 
   for PACKAGE in "${PACKAGES[@]}"
   do
-    start=$(date +%s%N)
-    yarn workspace @grafana/"${PACKAGE}" run build
```
stevenvachon, line 68: We aren't running the build anymore now

PR #27927, opened 2020-09-30
Drone: Don't ignore failure to publish next NPM packages

aknuds1:
Configure Drone to _not_ ignore failure to publish next NPM packages, since this step works now and it's better to be notified if it fails.

NB: We might want to implement retries for this step (in the build pipeline tool), to avoid random failures.


hugohaggmark (approved):
LGTM!

PR #27930, opened 2020-09-30
Build: added steps to generate packages api documentation.

### LINKED ISSUE #27661: Docs: tsdoc @links are not rendered in docs
**What happened**: [The `@link` included in `TemplateSrv.replace`'s tsdoc](https://github.com/grafana/grafana/blob/master/packages/grafana-runtime/src/services/templateSrv.ts#L17) is not rendered into the [corresponding docs page](https://grafana.com/docs/grafana/latest/packages_api/runtime/templatesrv/#replace-method). I imagine this also happens for all other `@link`s.

**What you expected to happen**: The docs page would say:
> Replace the values within the target string. See also [InterpolateFunction](https://grafana.com/docs/grafana/latest/packages_api/data/interpolatefunction/)

(note the link)

- `cd docs && make docs`
- open http://localhost:3002/docs/grafana/latest/packages_api/runtime/templatesrv/#replace-method

**Anything else we need to know?**: I think the last time I was messing around with tsdoc, it wasn't able to render out `@link`s. So maybe this is still impossible. Would certainly be a bummer, though, and at the very least it should be rendering the non-hyperlinked name.

**Environment**:
- Grafana version: master
- User OS & Browser: N/A
- Grafana plugins: N/A
- Others: N/A

mckn:
This PR will make sure that we generate the packages docs before running the build website step. Also for the PR pipeline we will lint the documentation of the packages code to make sure we are decreased the amount of undocumented code.

The reason why we want to do this is to make sure that when we add code to the `@grafana/*` packages we also add proper documentation.

**Todo:**
- [x] Fix issues in grafana/data that causes the docs generation to fail.
- [x] Make sure we fail the `generate-frontend-package-docs` if a failure occur when extracting the docs.

Fixes: https://github.com/grafana/grafana/issues/27661


scripts/lib.star
```diff
@@ -45,6 +45,7 @@ def pr_pipelines(edition):
         e2e_tests_server_step(),
         e2e_tests_step(),
         build_storybook_step(edition),
+        generate_packages_docs(edition=edition, lint=True),
```
aknuds1, line 48: Package docs should be better English
```suggestion
        generate_package_docs(edition=edition, lint=True),
```

scripts/lib.star
```diff
@@ -84,6 +85,7 @@ def master_steps(edition, is_downstream=False):
         e2e_tests_step(),
         build_storybook_step(edition=edition),
         publish_storybook_step(edition=edition),
+        generate_packages_docs(edition=edition, lint=False),
```
aknuds1, line 88: Package docs should be better English
```suggestion
        generate_package_docs(edition=edition, lint=False),
```

scripts/lib.star
```diff
@@ -441,6 +443,38 @@ def build_frontend_step(edition, is_downstream=False):
         ],
     }
 
+def generate_packages_docs(edition, lint=False):
```
aknuds1, line 446: Package docs should be better English
```suggestion
def generate_package_docs(edition, lint=False):
```

scripts/lib.star
```diff
@@ -441,6 +443,38 @@ def build_frontend_step(edition, is_downstream=False):
...
+    if edition == 'enterprise':
+        return None
+
+    if lint:
+        return {
+            'name': 'generate-packages-docs',
```
aknuds1, line 452:
```suggestion
            'name': 'generate-package-docs',
```

scripts/lib.star
```diff
@@ -441,6 +443,38 @@ def build_frontend_step(edition, is_downstream=False):
...
+                    'from_secret': 'grafana_misc_stats_api_key',
+                },
+            },
+            'commands': [
+                './scripts/ci-reference-docs-metrics.sh | ./bin/grabpl publish-metrics $${GRAFANA_MISC_STATS_API_KEY}',
+            ],
+        }
+    else:
+        return {
+            'name': 'generate-packages-docs',
```
aknuds1, line 468:
```suggestion
            'name': 'generate-package-docs',
```

scripts/lib.star
```diff
@@ -441,6 +443,38 @@ def build_frontend_step(edition, is_downstream=False):
...
+            'image': build_image,
+            'depends_on': [
+                'build-frontend'
+            ],
+            'environment': {
+                'GRAFANA_MISC_STATS_API_KEY': {
+                    'from_secret': 'grafana_misc_stats_api_key',
+                },
```
aknuds1, line 460: This is a problem, since this secret won't be shared with PRs, for security reasons.

aknuds1 (changes requested):

scripts/lib.star
mckn: aha, so lets remove this one for now. I think it is enough just to fail the build.

aknuds1:
I see one big problem, which is assuming the API key will be shared with PRs. It currently isn't, since it would easily leak to 3rd parties.

scripts/lib.star
aknuds1: Yeah, just testing should be enough for PRs, as you say.

scripts/lib.star
```diff
@@ -443,29 +443,24 @@ def build_frontend_step(edition, is_downstream=False):
         ],
     }
 
-def generate_packages_docs(edition, lint=False):
+def generate_package_docs(edition, lint=False):
     if edition == 'enterprise':
         return None
 
     if lint:
```
aknuds1, line 450: Can't you remove the `lint` condition now, since both branches are the same (AFAICT)?

aknuds1 (commented):
Left a question, I think .drone.yml also needs to be re-generated.

scripts/lib.star
mckn: Almost the same :D They run different scripts. The first one will only build the other will build and lint.

aknuds1: @mckn Yeah I noticed when checking locally, that there is a slight difference.

aknuds1:
@mckn Can you run `starlark drone convert` and commit the result? .drone.yml is still stale.

mckn:
> @mckn Can you run `starlark drone convert` and commit the result? .drone.yml is still stale.

yes, on it 👍

scripts/lib.star
```diff
@@ -441,6 +443,33 @@ def build_frontend_step(edition, is_downstream=False):
         ],
     }
 
+def generate_package_docs(edition, lint=False):
```
aknuds1, line 446: Can we just merge the command from this step into the build-docs-website step instead? I don't see any benefit to complicating the build graph with another step.

mckn: Yes, I will change that 👍

aknuds1: Never mind, leave it be for now since the build image is (probably) incompatible in build-docs-website.

aknuds1 (approved):

scripts/lib.star
```diff
@@ -497,6 +499,26 @@ def build_frontend_step(edition, is_downstream=False):
...
+    if lint:
+        script = './scripts/ci-reference-docs-lint.sh'
+    else:
+        script = './scripts/ci-reference-docs-build.sh'
+
+    return {
+        'name': 'build-frontend-docs',
+        'image': build_image,
+        'depends_on': [
+            'build-frontend'
```
aknuds1, line 515: Could this depend on 'test-frontend' instead, so it can run in parallel with build-frontend? Otherwise it will slow down CI.

mckn: It cannot, we need to get the packages built before continue with this one. The only way we can do that is to move the build packages to its own step instead of having it as part of the build-frontend.

aknuds1:
The build fails.

.drone.yml
```diff
@@ -143,6 +143,13 @@ steps:
   depends_on:
   - package
 
+- name: build-frontend-docs
+  image: grafana/build-container:1.2.27
+  commands:
+  - ./scripts/ci-reference-docs-lint.sh ci
+  depends_on:
+  - build-frontend
```
aknuds1, line 151: Does it have to run _after_ build-frontend? Can it execute in parallel instead, for speed?

scripts/ci-reference-docs-build.sh
```diff
@@ -17,4 +24,4 @@ yarn packages:docsToMarkdown
 
 # cleaning packages
 echo "cleaning up packages build files..."
-lerna run clean
+yarn packages:clean
```
aknuds1, line 27: I think this should be moved into the if block starting at line 8, since I don't think this cleanup should be part of CI (it will only slow things down and possibly break a parallel step).

scripts/ci-reference-docs-lint.sh
```diff
@@ -3,49 +3,42 @@
...
 if [ ! -d "$REPORT_PATH" ]; then
   # this script needs to be run after the packages have been built and the api-extractor has completed.
   # shellcheck source=/scripts/ci-reference-docs-build.sh
-  . "$BUILD_SCRIPT_PATH"
+  if ! . "$BUILD_SCRIPT_PATH" "$BUILD_MODE";
+    then
+      echo "Failed to build packages and extract docs" >&2
+      exit 1
+    else
+      echo "Successfully build packages and extracted docs"
```
aknuds1, line 27:
```suggestion
      exit 1
    else
      echo "Successfully built packages and extracted docs"
```

aknuds1 (commented):
Found one thing I think should be done differently, and I wonder if the step should run in parallel to the build-frontend step, for speed.

.drone.yml
mckn: yes, it is depending on the packages have been built. So if you remember we added that to the `build-frontend` build step.

scripts/ci-reference-docs-build.sh
mckn: ok ok, sounds good! I will remove it.

.drone.yml
```diff
@@ -449,6 +457,7 @@ steps:
   - cd /hugo && make prod
   depends_on:
   - initialize
+  - build-frontend-docs
```
aknuds1, line 460: I think the build-frontend-docs step is missing for the master pipeline.

scripts/pr.star
```diff
@@ -42,6 +43,7 @@ def pr_pipelines(edition):
         e2e_tests_server_step(),
         e2e_tests_step(),
         build_storybook_step(edition=edition, ver_mode=ver_mode),
+        build_frontend_docs_step(edition=edition, lint=True),
```
aknuds1, line 46: You need to add this to scripts/master.star also.

aknuds1 (changes requested):

.drone.yml
mckn: There are no need to build that one on the master build so I just skipped it. But do you think we should build it there as well? We are running it in the `grafana -> website` sync.

mckn: Great catch @aknuds1 !

aknuds1 (approved):
LGTM!

PR #27931, opened 2020-09-30
Transformations: Handles non timeseries data with same field name

### LINKED ISSUE #27781: Transform outer join - error when the metric name of the two tables is the same
I have Grafana 7.1.5 and the data is in Elasticsearch 7.X 
When make two queries to apply transform outer join from one field, but the join repeat the values of one table twice because the metric name is the same.
The first query is to count when have one condition and the second for another condition like "error" and "correct"

image

if not make the outer join the values of counter are different  between the two queries:

image

image

If tried to make transform organize fields, only appear one metric, because the metric fields have the same name:

image

hugohaggmark:

Fixes #27781

Not sure this fixes https://github.com/grafana/grafana/issues/27781#issuecomment-698686395 but it should fix https://github.com/grafana/grafana/issues/27781#issuecomment-698791122

### DIFF
--- a/packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts
+++ b/packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts
@@ -346,4 +346,61 @@ describe('SeriesToColumns Transformer', () => {
       },
     ]);
   });
+
+  // https://github.com/grafana/grafana/issues/27781
+  it('joins if metric name is the same and frames have no names', () => {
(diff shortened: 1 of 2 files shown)


mckn (approved):
Looks really great!

hugohaggmark:
@ryantxu I could use some help with the failing tests (see build). Seems like `calculateFieldTransformer` and `ensureColumnsTransformer` doesn't like the changes in this PR.

hugohaggmark:
Closing this in favor of #27963

PR #27936, opened 2020-09-30
CI: Notify failures in separate pipeline

marefr:
Changes slack notifications to execute in a separate pipeline, triggered by failure and depending on the other master  pipelines.

### DIFF
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -146,7 +146,7 @@ def master_pipelines(edition):
             depends_on=['build-master', 'windows-master',], install_deps=False, version_mode=version_mode,
         ))
 
-        notifyTrigger = { 'status': ['failure'] }
+        notifyTrigger = dict(trigger, status = ['failure'])
         pipelines.append(notifyPipeline(
             name='notify-master', slackChannel='grafana-ci-notifications', trigger=notifyTrigger,
             depends_on=['build-master', 'windows-master', 'publish-master'],
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -146,9 +146,9 @@ def master_pipelines(edition):
             depends_on=['build-master', 'windows-master',], install_deps=False, version_mode=version_mode,
         ))
 
-        notifyTrigger = dict(trigger, status = ['failure'])
+        notify_trigger = dict(trigger, status = ['failure'])
         pipelines.append(notifyPipeline(
-            name='notify-master', slackChannel='grafana-ci-notifications', trigger=notifyTrigger,
+            name='notify-master', slackChannel='grafana-ci-notifications', trigger=notify_trigger,
             depends_on=['build-master', 'windows-master', 'publish-master'],
         ))
     if edition == 'enterprise':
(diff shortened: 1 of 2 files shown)


scripts/lib.star
```diff
@@ -145,6 +145,12 @@ def master_pipelines(edition):
             name='publish-master', edition=edition, trigger=trigger, steps=publish_steps,
             depends_on=['build-master', 'windows-master',], install_deps=False, version_mode=version_mode,
         ))
+
+        notifyTrigger = dict(trigger, status = ['failure'])
```
aknuds1, line 149: Can you write in snake case instead of camel case, since it's standard in Starlark?

aknuds1 (approved):
Looks good I think, but Starlark should have snake case, just like Python.

scripts/lib.star
marefr: Sure will fix. I thought you could do `trigger + { 'status': ['failure']` similar to jsonnet, but it didn't work so used dict instead. Do you know?

marefr:
@aknuds1 does it make sense to have trigger of pipeline by branch, by event and by status?

aknuds1:
@marefr have to look into it after eating dinner.

marefr:
@aknuds1 no worries. I'm off for dinner as well. Pushed some enterprise/downstream fixes as well

scripts/lib.star
```diff
@@ -204,6 +216,22 @@ def pipeline(
 
     return pipeline
 
+def notifyPipeline(name, slack_channel, trigger, depends_on=[]):
```
aknuds1, line 219:
```suggestion
def notify_pipeline(name, slack_channel, trigger, depends_on=[]):
```

scripts/lib.star
```diff
@@ -145,6 +145,12 @@ def master_pipelines(edition):
             name='publish-master', edition=edition, trigger=trigger, steps=publish_steps,
             depends_on=['build-master', 'windows-master',], install_deps=False, version_mode=version_mode,
         ))
+
+        notify_trigger = dict(trigger, status = ['failure'])
+        pipelines.append(notifyPipeline(
```
aknuds1, line 150:
```suggestion
        pipelines.append(notify_pipeline(
```

scripts/lib.star
```diff
@@ -165,6 +171,12 @@ def master_pipelines(edition):
             version_mode=version_mode,
         ))
 
+        notify_trigger = dict(trigger, status = ['failure'])
+        pipelines.append(notifyPipeline(
```
aknuds1, line 175:
```suggestion
        pipelines.append(notify_pipeline(
```

aknuds1 (approved):
LGTM

scripts/lib.star
aknuds1: @marefr `dict` is probably fine. There's also a dict literal merging syntax in Python, but not sure if Starlark supports it.

aknuds1:
> @aknuds1 does it make sense to have trigger of pipeline by branch, by event and by status?

I think the original trigger by branch/push makes sense at least? Unless I'm mistaken, it means to listen for pushes to master.

PR #27945, opened 2020-09-30
Elasticsearch: Support multiple pipeline aggregations for a query

### LINKED ISSUE #15741: Cannot perform bucket script on moving average/derivative aggregations
6.0.0
Elasticsearch 6.6.1
Ubuntu
### What did you do?
Because grafana renders so many points on a chart (Kibanas default maximum is 200 data points per chart (time wise), visulizations that have very fast changing counts are hard to interpret as their lines become very messy within grafana (I do not want to change the minimum time interval within each chart). I also want to perform a 'count per second query' with some data using the new bucket script agg that finally became available within grafana 6.0.0.

With Kibana, as shown below, I can solve both of these by doing a moving average of count, then performing a bucket script on the moving average in order to get a smoother events per second graph:
image
However, in grafana, the bucket script does not allow a moving average to become a variable:
image

### LINKED ISSUE #18325: [Elasticsearch] Derivative of metric cannot be selected as variable in bucket script
For my use case I want to show inbound and outbound network traffic on the same panel.  Inbound as a positive number, outbound as a negative number.  
I need to use a bucket script to convert the derivative of outbound traffic to a negative value.
However the derivative does not show up as an available metric in the selection box.

image

I would expect to be able to select the derivative in the bucket script.  Additionally the same seems to apply to the bucket script itself as well.  Meaning you cannot chain two bucket scripts together either.

- Grafana version: 6.2.5
- Data source type & version: Elasticsearch 7.2
- OS Grafana is installed on: RHEL 7.6
- User OS & Browser: Chrome

simianhacker:

This PR enables chaining multiple pipeline aggregations together. With this PR users will be able to use the results of one pipeline aggregation in the results of another.  For example, if a user wanted to convert a counter to a positive rate they could so something like this:

image

This also allows user to take the second derivative using Elasticsearch's backend

image

I've also added a check to the `getPipelineAggregations` method to filter out any pipelines that are in the ancestor tree or itself to avoid recursions in Elasticsearch.

Fixes #18325 
Fixes #15741

None


Elfo404 (changes requested):
Thanks for the PR @simianhacker!

I think this is a good starting point but unfortunately it breaks the editor whenever removing metrics on which other metrics depend on.

I would say a good approach would be, when a  metric is removed, to recursively remove dependant metrics in https://github.com/grafana/grafana/blob/65d04688061b56c26dcd7a3fd075f42f6dd3db12/public/app/plugins/datasource/elasticsearch/metric_agg.ts#L202

or alternatively, ensure the editor is in a valid state by clearing dependant's fields.
Personally I'd rather just remove dependants, any opinion on that @torkelo ?

I guess the same applies whenever changing metric type, in that case would be nice to also update dependants.

simianhacker:
@Elfo404 Thanks for the review!

Looks like removing a metric in Master is also kind of broken. 

1. Starting with `max of system.network.in.bytes`, add `derivative` of "max of system.network.in.bytes"
1. Delete `max of system.network.in.bytes` aggregation, UI is left in broken state.

In this scenario, it leaves the derivative with no way to add additional aggregations. I can fix this along with what you identified above so you're never left in a situation where there are no metrics or the ability add more. In the situation where the user would be left with ZERO metrics, I will just add the default `Count` back so they can start over.

simianhacker:
> I guess the same applies whenever changing metric type, in that case would be nice to also update dependants.

@Elfo404 I'm not sure what needs to be done here. I was looking at the original code (on Master) and it currently doesn't update the dependents when changing the metric type. Do you have steps to reproduce the (bad) behavior and what the new behavior should be?

Also, I pushed changes to ensure the dependent's are appropriately removed. Also these changes should prevent the scenerio where you delete the "last" metric and are left with a broken visualization. Let me know if there is anything else I can do.

Elfo404 (changes requested):
@simianhacker
I tested it and it works now, thanks!

> Do you have steps to reproduce the (bad) behavior and what the new behavior should be?

Basically labels (descriptions for metrics that depend on others) are not being regenerated.
You can see it in the gif here: 
output

I wasn't sure it was like this on master but apparently it is and it's not related to your changes.

Also, we are currently migrating the ES query editor to React, so this kind of UI issues will be most likely tackled all at once.

> Also, I pushed changes to ensure the dependents are appropriately removed. Also, these changes should prevent the scenario where you delete the "last" metric and are left with a broken visualization

That's awesome, thanks a lot! This adds a lot of useful information when the migration to react will happen!

There are just a few typos, but the code looks good 🚀

simianhacker:
@Elfo404 Thanks for the response! I see what you're saying about the labels not updating. I will take a look and see if there is a quick and easy fix, if so I will include it as well. I'm willing to bet just by the nature of how React issues updates based on data changing that those problems will likely resolve themselves. It's good to hear there's a React migration effort underway.

simianhacker:
@Elfo404 I looked into why the label (`metric-segment` directive) not updating, that element is a child of the `metric-segment-model` directive. When you change one of the other metrics, like `Count` to `Average`, there isn't any data that triggers the update in the sibling directives since those are tied via the `id` of the metric (which doesn't change in that situation). 

Since you mentioned y'all are in the process or refactoring those components in React, I agree with deferring this feature/bug/glitch. The future code needs to make sure a change in one segment triggers a render in the siblings.

Elfo404 (approved):

Elfo404:
Thanks a lot for your contribution @simianhacker!

simianhacker:
@Elfo404 Thanks for the quick review!

PR #27947, opened 2020-10-01
consistent panel link when panel repeat is used

### LINKED ISSUE #27946: Templating: Share repeated panel in collapsed row results in panel not found
**What happened**:
Panel link is broken. Cause error like `Panel with id 6 not found`.

Open the panel which is shared by link.

1. create new dashboard
2. add query type template variable with `On Dashboard Load` option and `Multi-value` option
3. check one template variable value
4. add a panel with panel repeat option, select the step 2 template variable as panel repeat variable
5. move repeated panels inside collapsed row
6. save dashboard
7. expand row
8. check multiple template variable value
9. get panel link which are generated by 8 step (not first panel)
10. open panel link in another browser tab

I share reproducible dashboard, the Prometheus endpoint is `http://demo.robustperception.io:9090/`.
https://gist.github.com/mtanda/af673739c996e387726d4153c9d16f00

Open this link will reproduce the bug.
http://localhost:3000/d/oFyir-KMk/test-panel-repeat?orgId=1&var-repeat_key=All&from=1601502591231&to=1601524191232&viewPanel=6

**Environment**:
- Grafana version: 7.3.0-pre (0c70308870e1748063b66274941772deba8cb76d)
- User OS & Browser: Firefox
- Grafana plugins:
- Others:

mtanda:
This bug cause error when open link or render panel image.
Could be critical bug if user create dynamic dashboard by panel repeat.

Fixes https://github.com/grafana/grafana/issues/27946

By fixing variable value by scoped variable, panel count becomes only one, and repeatedPanelId should point correct panel.
The panel repeat key variable is overwrited when link is shared, this behavior is changed.

### DIFF
--- a/packages/grafana-data/src/types/panel.ts
+++ b/packages/grafana-data/src/types/panel.ts
@@ -113,6 +113,7 @@ export interface PanelModel<TOptions = any> {
   /** Version of the panel plugin */
   pluginVersion?: string;
   scopedVars?: ScopedVars;
+  repeatPanelId?: number;
 }
 
 /**
--- a/public/app/features/dashboard/components/ShareModal/utils.ts
+++ b/public/app/features/dashboard/components/ShareModal/utils.ts
@@ -31,6 +31,12 @@ export function buildParams(
 
   if (panel && !params.editPanel) {
     params.viewPanel = panel.id;
+    if (panel.scopedVars) {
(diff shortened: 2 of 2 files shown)


torkelo:
Not sure this solution feels correct. 

Not sure we can support this use case the way the collapsed rows work, feel like something more fundamental needs to be changed, like evaluating panel repeats even for collapsed rows

PR #27949, opened 2020-10-01
Drone: Upgrade build pipeline tool

aknuds1:
Upgrade build pipeline tool in Drone, in order to try to fix frontend metrics publishing.

Basically, frontend metrics publishing doesn't outright fail at the moment, but the metrics aren't published properly for some reason that's still to be pinpointed.

Also: Getting rid of one use of `npx`, since it shouldn't be used in a CI context (due to dynamically executing scripts).


marefr (approved):
LGTM

PR #27951, opened 2020-10-01
Plugins: Fix loading of backend plugins

marefr:
The referenced PR introduced a new dependency/import graph by importing the infra/metrics package to the models package. This broke the order of which services are registered with our dependency injection library and made the backend plugin manager crash.

Ref #27921 

### DIFF
--- a/pkg/infra/metrics/metrics.go
+++ b/pkg/infra/metrics/metrics.go
@@ -1,10 +1,7 @@
 package metrics
 
 import (
-	"errors"
-	"fmt"
 	"runtime"
-	"strings"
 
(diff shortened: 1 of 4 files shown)


aknuds1 (approved):
Wow, that was a surprising effect of that PR.

bergquist:
:facepalm: I guess this should be caught by integrations tests. Tricky one. We could try to serialize the dependency graph and save it as a golden test file so we know it's changed in a PR.

bergquist (approved):
LGTM

marefr:
@bergquist that's a good idea short term. However to fix this permanently I would like to just tear out the whole dependency injection library. Ping @sakjur

PR #27955, opened 2020-10-01
Drone: Don't use npx

aknuds1:
Get rid of another use of `npx` from Drone, since we don't want npx' ability to _dynamically_ execute NPM scripts. We want to use the _installed_ `lerna` script every time.

Also replace CircleCI references with Drone.


hugohaggmark (approved):
LGTM!

PR #27963, opened 2020-10-01
FieldDisplayName: Make field display names unique if field name exists more than once

### LINKED ISSUE #27781: Transform outer join - error when the metric name of the two tables is the same

torkelo:
This was a bit trickier than I thought. And not sure if this logic is correct or what side effects it has (existing transforms seem to work without modification). 

I had to modify seriesToColumns as MutableDataFrame.addField does not add the instance you pass to the function but creates a new field instance. 

Given two data frames with fields:
type, value 

Join now gives us: 

Screenshot from 2020-10-01 14-08-05

Fixes #27781

Alternative to #27931

### DIFF
--- a/packages/grafana-data/src/transformations/transformers/seriesToColumns.ts
+++ b/packages/grafana-data/src/transformations/transformers/seriesToColumns.ts
@@ -68,7 +68,7 @@ export const seriesToColumnsTransformer: DataTransformerInfo<SeriesToColumnsOpti
     });
 
     for (const item of allFields) {
-      resultFrame.addField(item.newField);
+      item.newField = resultFrame.addField(item.newField);
     }
 
     const keyFieldTitle = getFieldDisplayName(resultFrame.fields[0], resultFrame);
--- a/packages/grafana-data/src/dataframe/ArrowDataFrame.test.ts
+++ b/packages/grafana-data/src/dataframe/ArrowDataFrame.test.ts
@@ -43,7 +43,7 @@ describe('Read/Write arrow Table to DataFrame', () => {
       ],
     });
 
-    const table = grafanaDataFrameToArrowTable(frame);
+    const table = grafanaDataFrameToArrowTable(frame, true);
     expect(table.length).toEqual(frame.length);
 
     // Now back to DataFrame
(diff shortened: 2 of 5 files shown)


packages/grafana-data/src/transformations/transformers/seriesToColumns.ts
```diff
@@ -68,7 +68,7 @@ export const seriesToColumnsTransformer: DataTransformerInfo<SeriesToColumnsOpti
     });
 
     for (const item of allFields) {
-      resultFrame.addField(item.newField);
+      item.newField = resultFrame.addField(item.newField);
```
hugohaggmark, line 71: Maybe we should add the same test case from my PR?

hugohaggmark (approved):
LGTM! I think we should copy the test case from my PR https://github.com/grafana/grafana/blob/176cb648ac2c0cfc584280607301102e7ef9e1c6/packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts#L351

packages/grafana-data/src/transformations/transformers/seriesToColumns.ts
torkelo: yea, I need to add a test case for seriesToColumn, as without the above line things break quite badly

torkelo: added test

torkelo:
Not sure most have, but some definitely.

And for a tranformer like organize fields it has definitely. And that is the main way to rename fields , and it currently cannot distinguish between two fields with same display name

torkelo:
and without unique names, the whole field overrides system sort of does not work

ryantxu (approved):
:shrug: likely better to have this constraint

mckn (approved):
I think this looks good! Great work @torkelo!

If I don't recall incorrectly the transforms I have been involved in will probably pick the first time field as the default one (if they have the same name). This should not have an impact on the merge transform since it uses the `field.name`.

PR #27965, opened 2020-10-01
Drone: Fix releasing of next NPM packages

aknuds1:
Fix releasing of next NPM packages under Drone.


stevenvachon (approved):
LGTM; with the understanding that the hash is generated and injected into the *package.json* file as part of the build step instead.

aknuds1:
@stevenvachon I tested that, and it looks fine.

mckn:
> LGTM; with the understanding that the hash is generated and injected into the _package.json_ file as part of the build step instead.

Yes, that is correct. I can update the comments to make that more obvious.

aknuds1:
Is it ready now @mckn?

mckn:
> Is it ready now @mckn?

`  # Frontend packages have already been versioned and built by the 
  # build-frontend step in drone. We will only unpublish the previous 
  # canary version and publish the current built version as the new 
  # latest canary build.`

aknuds1:
@mckn I don't have any opinion on the comments.

mckn (approved):

PR #27968, opened 2020-10-01
WIP: CloudWatch Logs queue

### LINKED ISSUE #27969: CloudWatch Logs: Implement query queueing
Amazon CloudWatch Logs imposes certain limits, or quotas, on the usage of various API endpoints and features. One of these limits is the limit on query concurrency, which has a default of 4, but can be increased to 10 on request. If a user attempts to start a query while the maximum number of concurrent queries are running, an error will be returned. In order to prevent CloudWatch Logs panels, or CloudWatch Logs queries sent in Explore, from returning errors in the case where the concurrent queries limit is reached, it is necessary to implement a queueing system that only dispatches a “Start Query” request when we are under the concurrent query limit.

kaydelaney:
Amazon CloudWatch Logs imposes certain limits, or quotas, on the usage of various API endpoints and features. One of these limits is the limit on query concurrency, which has a default of 4, but can be increased to 10 on request. If a user attempts to start a query while the maximum number of concurrent queries are running, an error will be returned. In order to prevent CloudWatch Logs panels, or CloudWatch Logs queries sent in Explore, from returning errors in the case where the concurrent queries limit is reached, it is necessary to implement a queueing system that only dispatches a “Start Query” request when we are under the concurrent query limit.
In service of this, this PR changes the architecture of the CloudWatch datasource from one that periodically sends `GetQueryResults` proxied through the grafana backend, to a websocket based architecture using the currently in-progress Grafana Live feature.
A lot of the design in this PR is exploratory and experimental, and in fact is broken right now as parts need to be rewritten to accommodate some recent changes to Grafana Live, but most of the main ideas are there.

See #27969

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,6 +68,7 @@ require (
 	github.com/robfig/cron v0.0.0-20180505203441-b41be1df6967
 	github.com/robfig/cron/v3 v3.0.0
 	github.com/russellhaering/goxmldsig v0.0.0-20200902171629-2e1fbc2c5593 // indirect
+	github.com/satori/go.uuid v1.2.1-0.20181028125025-b2ce2384e17b
 	github.com/smartystreets/goconvey v1.6.4
 	github.com/stretchr/testify v1.6.1
 	github.com/teris-io/shortid v0.0.0-20171029131806-771a37caa5cf
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	cloud.google.com/go/storage v1.8.0
 	github.com/BurntSushi/toml v0.3.1
 	github.com/VividCortex/mysqlerr v0.0.0-20170204212430-6c6b55f8796f
-	github.com/aws/aws-sdk-go v1.33.12
+	github.com/aws/aws-sdk-go v1.34.26
 	github.com/benbjohnson/clock v0.0.0-20161215174838-7dc76406b6d3
 	github.com/bradfitz/gomemcache v0.0.0-20190913173617-a41fca850d0b
 	github.com/centrifugal/centrifuge v0.11.0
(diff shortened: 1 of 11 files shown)


ryantxu:
IIUC, this does two things:
1. make a queue to limit queries
2. adds a new websocket endpoint 

Rather than adding a new websocket, I think we should use the new live server (feature flag for 7.3).  The main benefit is that there is a single websockt for a client and we can then use multiple channels in the same socket.

The other thing we need to figure out is how multiple clients can share the same streaming results, or if each client needs its own stream centrifugo has user channels:
https://centrifugal.github.io/centrifugo/server/channels/#user-channel-boundary

kaydelaney:
Just to explain the rough architecture here:
A channel `SocketChannel` has been added to the `HTTPServer`. Upon being initialized, a function `runRequests` is started in another thread where it waits for messages to `SocketChannel`. Each message contains a `channelName`, and a channel. Upon receiving a message, a new thread is started where messages from this new channel are published to `channelName`  via `hs.Live.Publish`. This is done until the channel closes.

Messages get sent to `SocketChannel` through the following mechanism:
First, a http request is made by the client to `'/api/tsdb/wsquery`. Upon receiving this request, the backend generates a channel name (uuid) and responds to the http request with this channel name. The client receives this channel name and waits for additional query responses through the channel.
At the same time, a new (go) channel is made on the backend and passed to the datasource to handle via `WebSocketQuery()`. That same channel and its name are sent through `SocketChannel` as mentioned above. The datasource on the backend sends query results through the go channel mentioned previously until the query is completed, at which point the channel is closed and the previously created thread completes.
As far as I know this should only use the one websocket, just with a different channel for each query request sent by a client.

ryantxu:
Ahh -- I see.  I am reluctant to add the new `/api/tsdb/wsquery` because it will be akward for plugins, can we do the same approach with either:
* DataQuery -- return a DataFrame with channel info in the metadata?  For many queries I assume you may want to "query for last X events, then stream new ones from channel Y"

or

* if that is not appropriate and we need something more custom, use the existing `/datasources/:id/resources/*` call

This is a bit akward since cloudwatch is not a plugin, and the infrastructure should look more standard!

kaydelaney:
Closing in favor of #28176

Issue #25573, opened 2020-06-12
Revamp GitLab OAuth2 Authentication


vladdy:
**What would you like to be added**:
Hi! I'd like to offer to get rid of API calls during OAuth2 authentication with GitLab as the service currently provides all the necessary information via the OpenID endpoint (https://docs.gitlab.com/ee/integration/openid_connect_provider.html#shared-information).

**Why is this needed**:
The current implementation requires `api` or `read_user` scopes to authenticate a user, however it appears to be excessive as `openid` and `profile` should be enough.

Would you be open for a PR for the above change?

marefr:
> Would you be open for a PR for the above change?

@vladdy we're always open to PR's, but if we accept it is a later discussion. The feature request sounds valid though

sakjur:
> [...]should apply to all different oauth provider

Are scopes standardized in OpenID Connect? From a pure OAuth perspective, this change sounds to me like it's beyond the standard and in the IdP specific configuration, so it would make sense to apply specifically for GitLab imo

vladdy:
> [...]should apply to all different oauth provider

@marefr, can you please clarify why do you think it should be applied to all different providers? AFAIK, the groups is not a standard claim (https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims). My reasoning for proposing the change is to avoid giving unnecessary permissions to the OAuth2 tokens used by Grafana with GitLab as IdP, which should improve the security of already existing functionality.

PR #27976, opened 2020-10-01
Use read_api scope for GitLab OAuth

vin01:
Hi Team!

`read_api` seems to be the minimal scope currently which can be used, it shall be preferred over `api` which grants complete read/write access.

- https://gitlab.com/gitlab-org/gitlab/-/merge_requests/28944#note_322904691
- https://gitlab.com/gitlab-org/gitlab/-/issues/21909

gitlab_scopes

`read_api` is also more permissive than required apparently, `openid`/`profile` seem to be the way forward: https://github.com/grafana/grafana/issues/25573

Thanks!


achatterjee-grafana (approved):
Thanks for adding the change!

PR #27977, opened 2020-10-01
Plugins: Fix panic when using complex dynamic URLs in app plugin routes 

cinaglia:

This PR fixes a panic (`runtime error: invalid memory address or nil pointer dereference`) when setting up an app plugin route with a complex enough templated URL (e.g. `{{if .JsonData.apiHost}}{{.JsonData.apiHost}}{{else}}https://example.com{{end}}`).

This happens because the plugin proxy is blindly calling `url.Parse(route.URL)` before interpolating the string. I propose we always interpolate it before parsing the URL.

Partial stacktrace:
```
/usr/local/Cellar/go/1.15/libexec/src/runtime/panic.go:212 (0x404feb2)
  panicmem: panic(memoryError)
/usr/local/Cellar/go/1.15/libexec/src/runtime/signal_unix.go:720 (0x404fd32)
  sigpanic: panicmem()
/Users/victorcinaglia/workspace/grafana/pkg/api/pluginproxy/pluginproxy.go:63 (0x4ee337e)
  NewApiPluginProxy.func1: req.URL.Scheme = targetURL.Scheme
/usr/local/Cellar/go/1.15/libexec/src/net/http/httputil/reverseproxy.go:241 (0x48055cc)
  (*ReverseProxy).ServeHTTP: p.Director(outreq)
/Users/victorcinaglia/workspace/grafana/pkg/api/app_routes.go:62 (0x503dd90)
  AppPluginRoute.func1: proxy.ServeHTTP(c.Resp, c.Req.Request)
/usr/local/Cellar/go/1.15/libexec/src/reflect/value.go:475 (0x409fec6)
  Value.call: call(frametype, fn, args, uint32(frametype.size), uint32(retOffset))
/usr/local/Cellar/go/1.15/libexec/src/reflect/value.go:336 (0x409f3b8)
...
```

This PR also refactors the code a bit to remove an unused function as well as to share the `AddHeaders` function between both datasource & app plugin proxies.

Let me know if you'd prefer if the refactoring was done in a separate PR.


pkg/api/pluginproxy/pluginproxy.go
```diff
@@ -82,33 +65,9 @@ func NewApiPluginProxy(ctx *models.ReqContext, proxyPath string, route *plugins.
...
-			targetURL, err := url.Parse(interpolatedURL)
-			if err != nil {
-				ctx.JsonApiErr(500, "Could not parse custom url: %v", err)
-				return
-			}
-			req.URL.Scheme = targetURL.Scheme
-			req.URL.Host = targetURL.Host
-			req.Host = targetURL.Host
-			req.URL.Path = util.JoinURLFragments(targetURL.Path, proxyPath)
+		if err := AddHeaders(&req.Header, route, data); err != nil {
```
marefr, line 68: Due to other suggestions.

```suggestion
		if err := addHeaders(&req.Header, route, data); err != nil {
```

pkg/api/pluginproxy/pluginproxy_test.go
```diff
@@ -142,7 +169,7 @@ func getPluginProxiedRequest(ctx *models.ReqContext, cfg *setting.Cfg, route *pl
 	}
 	proxy := NewApiPluginProxy(ctx, "", route, "", cfg)
 
-	req, err := http.NewRequest(http.MethodGet, route.URL, nil)
+	req, err := http.NewRequest(http.MethodGet, "", nil)
```
marefr, line 172: Did this cause a problem? Just trying to understand why you changed it and/or if it really doesn't matter.

marefr (changes requested):
Sorry for keeping you waiting. Looks good I think. Added a few suggestions and comments.

aknuds1 (commented):
Generally looks good to me, but agree with @marefr's comments.

pkg/api/pluginproxy/pluginproxy_test.go
cinaglia: Yes, the issue I was running into was that when `route.URL` is an interpolated string, the call to `http.NewRequest` would fail to parse the URL parameter with the following error (`parse "{{if .JsonData.apiHost}}{{.JsonData.apiHost}}{{else}}https://example.com{{end}}": first path segment in URL cannot contain colon`).

This is somewhat confusing. The `req` object is the request that is being proxied, so in practice the URL would actually be along the lines of `/api/plugin-proxy/grafana-simple-app/api/v4/endpoint`. I went ahead and changed from an empty string to something like that URL to hopefully make that clearer.  Let me know if you think a comment would be better (or any other suggestions you might have).

cinaglia:
Thanks for the feedback!

> This can be lowercase/non-exported since it doesn't seem to be used outside this package.

Good catch! I also went ahead and lower cased `InterpolateString` which doesn't seem to be used externally either. Can you think of any issues I might be failing to anticipate (not hugely familiar with the codebase)?

aknuds1:
Going to have another look, forgot in the meantime :)

pkg/api/pluginproxy/pluginproxy_test.go
```diff
@@ -142,7 +169,7 @@ func getPluginProxiedRequest(ctx *models.ReqContext, cfg *setting.Cfg, route *pl
 	}
 	proxy := NewApiPluginProxy(ctx, "", route, "", cfg)
 
-	req, err := http.NewRequest(http.MethodGet, route.URL, nil)
+	req, err := http.NewRequest(http.MethodGet, "/api/plugin-proxy/grafana-simple-app/api/v4/alerts", nil)
```
aknuds1, line 172: What's the purpose of this change? Just asking because I'd like to understand :)

aknuds1 (approved):
AFAICT, this PR makes total sense, I just have a question (see comment).

pkg/api/pluginproxy/pluginproxy_test.go
aknuds1: Update: Never mind, I found your reply to @marefr earlier clearing this up. I think an explanatory code comment would be helpful here though.

aknuds1:
@marefr You think this is good now? AFAICT it's all good.

marefr (approved):
LGTM, but would like to take it for a test spin as well just to be sure.

aknuds1:
I have fixed the conflicts locally; if you're interested, I can push them to your branch @cinaglia.

marefr:
@aknuds1 please do

aknuds1:
@marefr @cinaglia I pushed my conflict resolutions.

marefr (approved):
Tested very quickly with azure and cloud monitoring. Lets get this merged

cinaglia:
@aknuds1 thanks for resolving the conflicts! Sorry for not doing it myself – I'm just coming back from a short vacation.

PR #27980, opened 2020-10-01
Storybook: Fix checkbox spacing

### LINKED ISSUE #27065: Storybook: Fix checkbox spacing
Compared to the other form elements the label and description text have a different line-height so it’s spaced differently. They inherit body 1.5 line-height, the other components use a 1.25 line-height. 

And we have a `<br>` in between, which is not required.

Lundbyit:
* Remove unecessary <br>
* add line-height to description to align it with other form component styles

Suggestion on how to fix #27065

Fixes #27065 

Not sure if the line-height should be in the checkbox description styles or in label description styles.

### DIFF
--- a/packages/grafana-ui/src/components/Forms/Checkbox.tsx
+++ b/packages/grafana-ui/src/components/Forms/Checkbox.tsx
@@ -25,6 +25,7 @@ export const getCheckboxStyles = stylesFactory((theme: GrafanaTheme) => {
     description: cx(
       labelStyles.description,
       css`
+        line-height: 1.25px;
         padding-left: ${theme.spacing.formSpacingBase}px;
       `
     ),
@@ -117,12 +118,7 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
         />
         <span className={styles.checkmark} />
         {label && <span className={styles.label}>{label}</span>}
-        {description && (
-          <>
-            <br />
-            <span className={styles.description}>{description}</span>
-          </>
-        )}
+        {description && <span className={styles.description}>{description}</span>}
       </label>
     );
   }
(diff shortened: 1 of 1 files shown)


packages/grafana-ui/src/components/Forms/Checkbox.tsx
```diff
@@ -25,6 +25,7 @@ export const getCheckboxStyles = stylesFactory((theme: GrafanaTheme) => {
     description: cx(
       labelStyles.description,
       css`
+        line-height: 1.25px;
```
Clarity-89, line 28:
```suggestion
        line-height: 1.25;
```

Clarity-89 (changes requested):
Thank you for your contribution! 🎉   It looks good, just one minor thing to fix: we use unitless values for line height instead of pixels.

packages/grafana-ui/src/components/Forms/Checkbox.tsx
peterholmberg: was thinking if we should use the line heights in the `theme.typography.lineHeight` ?

Clarity-89: We don't have `line-height` 1.25 in the themes, the closest is 1.1, which also works I think.

peterholmberg: I'd suggest using that so we don't introduce custom line-heights

Lundbyit: Thanks for the pointer. I'll fix that

Clarity-89 (approved):
Great job!

Lundbyit:
Thanks! :) And thanks for the help with this PR

PR #27988, opened 2020-10-01
Explore: respect min_refresh_interval

### LINKED ISSUE #27494: min_refresh_interval not work for explore
**What happened**:
When I set min_refresh_interval=1m, and use explore page. I found the min refresh interval is still 5s.
image

Should only show >= 1m refresh interval options.
As above
No
**Environment**:
- Grafana version: 7.1.3
- Data source type & version: any
- User OS & Browser: MacOS & Chrome
- Grafana plugins: Explore
- Others:

elliotpryde:
Stops us using the unvalidated default time intervals in [RefreshPicker.tsx](https://github.com/grafana/grafana/blob/6145bf77cf4462af04353001cd7eda369e15b235/packages/grafana-ui/src/components/RefreshPicker/RefreshPicker.tsx) and instead passing intervals from [RunButton.tsx](https://github.com/grafana/grafana/blob/49b5fc4b9a904a4f53f7da46637ed72a75ab3eb0/public/app/features/explore/RunButton.tsx) which have been validated against the `min_refresh_interval` by [TimeSrv.ts](https://github.com/grafana/grafana/blob/c89ad9b03883085c35f88e05bcec25ecdacef3e7/public/app/features/dashboard/services/TimeSrv.ts).

Fixes #27494

I suspect that the RunButton component may not be the best place to define a set of default intervals, but I'm not sure where they should go.


zoltanbedi:
Hi @elliotpryde, thanks for your contribution. I think this looks great and works as expected. 
You can import the defaultIntervals from grafana/ui like this `import { defaultIntervals } from '@grafana/ui/src/components/RefreshPicker/RefreshPicker';`
Also please update the documentation in https://github.com/grafana/grafana/blob/master/docs/sources/administration/configuration.md#min_refresh_interval by mentioning that it affects explore as well from now.

zoltanbedi (approved):
Thank you for contributing.

hugohaggmark:
> Hi @elliotpryde, thanks for your contribution. I think this looks great and works as expected.
> You can import the defaultIntervals from grafana/ui like this `import { defaultIntervals } from '@grafana/ui/src/components/RefreshPicker/RefreshPicker';`
> Also please update the documentation in https://github.com/grafana/grafana/blob/master/docs/sources/administration/configuration.md#min_refresh_interval by mentioning that it affects explore as well from now.

@zoltanbedi @elliotpryde maybe we should export `defaultIntervals` properly so it is accessible directly via `@grafana/ui`?

hugohaggmark (approved):
LGTM!

zoltanbedi:
Seems like the ci is not starting for some reason. can you please try and merge master to your branch and push it?

docs/sources/administration/configuration.md
```diff
@@ -552,6 +552,8 @@ Number dashboard versions to keep (per dashboard). Default: `20`, Minimum: `1`.
 This prevents users from setting the dashboard refresh interval of a lower than given interval. Per default this is 5 seconds.
 The interval string is a possibly signed sequence of decimal numbers, followed by a unit suffix (ms, s, m, h, d), e.g. `30s` or `1m`.
 
+As of Grafana v7.3, this also limits the refresh interval options in the explore section.
```
achatterjee-grafana, line 555: As of Grafana v7.3, this also limits the refresh interval options in Explore.

achatterjee-grafana (commented):
Added some copy-edit suggestions.

kylebrandt:

PR #27997, opened 2020-10-02
Migration: TestData Query Editor

### LINKED ISSUE #27526: TestData: Migrate TestData Query Editor to react
Would be really useful to migrate test data query editor to react so we can more easily add things and integrate the static data frame definition feature of https://github.com/marcusolsson/grafana-static-datasource into the testdata backend scenarios as being able to define a data frame is very useful for testing scenarios (e2e tests, dev test dashboards,  and replicating issues) 

If this query editor was in react we could also then use the new annotation support that is using the normal query editor: 

https://github.com/grafana/grafana/pull/27375

Clarity-89:

- Migrate TestData Query Editor to React
- Use new Inline form components
- Minor UI updates

Screenshots:
Screenshot 2020-10-02 at 13 53 03
Screenshot 2020-10-02 at 13 32 22
Screenshot 2020-10-02 at 13 32 44

(Wide screen)
Screenshot 2020-10-02 at 13 33 13

Fixes https://github.com/grafana/grafana/issues/27526

There is an error in console when adding new point via `Manual Entry`, which happens because the `datapoints` value is `null`. Couldn't get to the bottom of it. It also happens in production, with the old editor.
Screenshot 2020-10-02 at 13 30 32


torkelo:
Awesome job! Will take if for test spin later

Clarity-89:
Noticed that a few e2e tests got broken because of the new Select, which cannot be selected by cypress. @hugohaggmark do you know if we have any way to test our `Select` in e2e tests?

hugohaggmark:
> Noticed that a few e2e tests got broken because of the new Select, which cannot be selected by cypress. @hugohaggmark do you know if we have any way to test our `Select` in e2e tests?

I think I've added some `aria-label` for the Select, check where this is used by e2e https://github.com/grafana/grafana/blob/0ffd9a9a3c5c632be811ea36364b1c76269ea184/packages/grafana-e2e-selectors/src/selectors/components.ts#L129

Clarity-89:
> > Noticed that a few e2e tests got broken because of the new Select, which cannot be selected by cypress. @hugohaggmark do you know if we have any way to test our `Select` in e2e tests?
> 
> I think I've added some `aria-label` for the Select, check where this is used by e2e
> 
> https://github.com/grafana/grafana/blob/0ffd9a9a3c5c632be811ea36364b1c76269ea184/packages/grafana-e2e-selectors/src/selectors/components.ts#L129

Thank you! I'll check it out.

torkelo (approved):
Looks good, awesome job!

PR #28007, opened 2020-10-02
FieldConfig: Implementation slider editor (#27592)

### LINKED ISSUE #27592: Field config: implement slider editor
In the new graph panel, there are plenty of field config properties that are candidates to be controlled via slider editor. An example of such property is fill-opacity:

image

Another is line width, point radius, etc.

For easier control of such properties it would be nice to have a slider type of config editor rather than using select component. 

From API perspective this would be a following method in both `FieldConfigEditorBuilder` and `PanelOptionsEditorBuilder` builders:

```
.addSlider({
  settings: {
    min: number,
    max: number,
    step: number,
  },
  ...
})
```

Settings have a similar API to `NumberFieldConfigSettings`, but the range and step are required.

TODO:
- [ ] Let @marcusolsson and plugins team know the new control is added to the API
- [ ] Include info about new API in change log

Ref #27572

isaozlerfm:
This PR implements the Slider component

https://github.com/grafana/grafana/issues/27592


hugohaggmark:
I unassigned myself because I think @dprokop will have a look

dprokop:
@isaozlerfm thanks a lot for the contribution, I'll have a look at it later today!

packages/grafana-ui/src/components/OptionsUI/slider.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useCallback } from 'react';
+import { FieldConfigEditorProps, SliderFieldConfigSettings } from '@grafana/data';
+import { Slider } from '../Slider/Slider';
+
+export const SliderValueEditor: React.FC<FieldConfigEditorProps<number | number[], SliderFieldConfigSettings>> = ({
```
dprokop, line 5: I'm not convinced about the typing here.

I assume the `number[]` typing is to enable range sliders. But I would opt for range slider (`addRangeSlider`) to be a separate method on both `FieldConfigEditorBuilder` and  `PanelOptionsEditorBuilder`. 

The SlideValueEditor is basically a specific UI implementation of numeric option hence it should have pretty much the same interface and implementation as [numeric option](https://github.com/grafana/grafana/blob/62aa15f76074d028599feb0c628cf13f533c6245/packages/grafana-ui/src/components/OptionsUI/number.tsx#L10-L9):

```
  const slider: StandardEditorsRegistryItem<number> = {
    id: 'slider',
    name: 'Slider', 
    description: 'Allows numeric values input',
    editor: SliderValueEditor as any,
  };
```

And then the range slider:

```
  const rangeSlider: StandardEditorsRegistryItem<[number, number]> = {
    id: 'rangeSlider',
    name: 'Range slider', 
    description: 'Allows numeric range input',
    editor: RangeSliderValueEditor as any,
  };
```

The distinction between those two will simplify the logic behind the initial value.

packages/grafana-ui/src/components/OptionsUI/slider.tsx
```diff
@@ -0,0 +1,28 @@
...
+}) => {
+  const { settings } = item;
+  const onValueAfterChange = useCallback(
+    (values: number | number[]) => {
+      onChange(values);
+    },
+    [onChange]
+  );
+  const initialValue =
+    typeof value === 'number' ? [value] : typeof value === 'string' ? [+value] : Array.isArray(value) ? value : [0];
```
dprokop, line 18: See my previous comment about range slider / initial value

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -230,6 +231,13 @@ export const getStandardOptionEditors = () => {
     editor: NumberValueEditor as any,
   };
 
+  const slider: StandardEditorsRegistryItem<number, number[]> = {
+    id: 'slider',
+    name: 'Slider',
+    description: 'Allows slider input returning an array of numeric min & max values',
```
dprokop, line 237: I think slider should return just a number, as number input does. min/max should be responsibility of range slider.

packages/grafana-data/src/field/overrides/processors.ts
```diff
@@ -24,6 +24,18 @@ export const numberOverrideProcessor = (
...
+  min: number;
+  max: number;
+  step?: number;
+}
+
+export const sliderOverrideProcessor = (
+  value: any,
+  _context: FieldOverrideContext,
+  _settings?: SliderFieldConfigSettings
+) => value as number | number[];
```
mckn, line 37: Question: in what scenario could we return an array of numbers?

dprokop (changes requested):
@isaozlerfm thanks again for the contibution, it's great! 

I have left a couple of comments, especially regarding the input type being accepted by slider (number|number[]). An important thing about the slider input to remember that it's a **variation of number input and should have the same interface and behavior**. This means that is should also support float/integers as the number input does.

I have also checked with our design team if we already have a design for the slider editor, and yes, we do!

Please take a look at the screenshot below:
image

In your implementation, I find it confusing for the value to be rendered on top of the slider handle. I think this value should be shown in the editable input next to the slider as in the design.

It would be great if you could implement this design for the slider control.

mckn (commented):
Overall I think the code looks great! Nice work!

I have some small nits:

- In what scenario do we need the `number[]` as a value from the slider input?
- It would be nice to use this on at least one place in this PR so we actually start using it.
- I did a quick test where I used it in the `Fill area opacity` and it looks like in the picture below. I would make sure that the slider has the correct margin so it is aligning with the other input components in the panel editor.

Skärmavbild 2020-10-05 kl  14 28 49

packages/grafana-data/src/field/overrides/processors.ts
dprokop: @mckn it's for the range slider I assume. See my comment here: https://github.com/grafana/grafana/pull/28007#discussion_r499558978 :)

packages/grafana-ui/src/components/OptionsUI/slider.tsx
mckn: Aha, I agree on making two different components out of this to make both of them simpler and easier to understand.

dprokop:
Ah, and for a matter of testing feel free to replace selects in [graph3 plugin](https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph3/module.tsx) (fill opacity, line width, point size) with the field editor you are introducing:)

packages/grafana-ui/src/components/OptionsUI/slider.tsx
hugohaggmark: I agree with this as well, the Variable system has shown that having string | string[] as a type introduces a lot of confusion and bugs, great suggestion @dprokop 🌟

isaozlerfm: While working on this I had the same thoughts, but didn't want to opt out the range feature. But I als o think it's better to reiterate once the initial slider component is implemented.

packages/grafana-data/src/field/overrides/processors.ts
isaozlerfm: Hi @mckn thx for you feedback. Do you have a specific component in mind (ref. feedback point 2)?

packages/grafana-ui/src/utils/standardEditors.tsx
isaozlerfm: I will adjust the code so wel get just retrieve a number value 👍

packages/grafana-data/src/field/overrides/processors.ts
isaozlerfm: I think a good use case for starting to use this is the [graph3 plugin](https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph3/module.tsx) as @dprokop suggested

isaozlerfm:
Thx guys for all the feedback;

to sum up and check if I got you guys correctly;
- we opt out the range slider feature and only focus on the slider input for now
- add an input field as in the designs
- implement the component in the graph3 plugin as for start
- and some minor tweaks related to the range input feature changes

packages/grafana-data/src/field/overrides/processors.ts
mckn: Yes, I agree! Go with that one!

dprokop:
Hi @isaozlerfm, anything we can help you with to push this work forward? :)

isaozlerfm:
> Hi @isaozlerfm, anything we can help you with to push this work forward? :)

finished the Pareto plugin last week :) this week I will finish this 👍

packages/grafana-ui/src/components/Slider/Slider.tsx
```diff
@@ -106,33 +28,52 @@ export const Slider: FunctionComponent<Props> = ({
...
   const isHorizontal = orientation === 'horizontal';
   const theme = useTheme();
   const styles = getStyles(theme, isHorizontal);
-  const RangeWithTooltip = createSliderWithTooltip(Range);
+  const SliderWithTooltip = SliderComponent;
+  const [slidervalue, setSliderValue] = useState<number>(value || min);
+  const onSliderChange = (v: number) => {
+    setSliderValue(v);
+    !!onChange && onChange(v);
+  };
```
dprokop, line 41: useCallback maybe?

packages/grafana-ui/src/components/Slider/Slider.tsx
```diff
@@ -106,33 +28,52 @@ export const Slider: FunctionComponent<Props> = ({
...
+  const onSliderInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    let v = +e.target.value;
+
+    v > max && (v = max);
+    v < min && (v = min);
+
+    setSliderValue(v);
+    !!onChange && onChange(v);
```
dprokop, line 49:
```suggestion
    if (onChange) {
      onChange(v)
    }
```

packages/grafana-ui/src/components/Slider/Slider.tsx
```diff
@@ -106,33 +28,52 @@ export const Slider: FunctionComponent<Props> = ({
...
 }) => {
   const isHorizontal = orientation === 'horizontal';
   const theme = useTheme();
   const styles = getStyles(theme, isHorizontal);
-  const RangeWithTooltip = createSliderWithTooltip(Range);
+  const SliderWithTooltip = SliderComponent;
+  const [slidervalue, setSliderValue] = useState<number>(value || min);
+  const onSliderChange = (v: number) => {
+    setSliderValue(v);
+    !!onChange && onChange(v);
```
dprokop, line 40:
```suggestion
    if (onChange) {
      onChange(v)
    }
```

packages/grafana-ui/src/components/Slider/Slider.tsx
```diff
@@ -106,33 +28,52 @@ export const Slider: FunctionComponent<Props> = ({
...
+  };
+  const onSliderInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    let v = +e.target.value;
+
+    v > max && (v = max);
+    v < min && (v = min);
+
+    setSliderValue(v);
+    !!onChange && onChange(v);
+    !!onAfterChange && onAfterChange(v);
```
dprokop, line 50: if (onAfterChange) {
      onAfterChange(v)
    }

packages/grafana-ui/src/components/Slider/Slider.tsx
```diff
@@ -106,33 +28,52 @@ export const Slider: FunctionComponent<Props> = ({
...
   const theme = useTheme();
   const styles = getStyles(theme, isHorizontal);
-  const RangeWithTooltip = createSliderWithTooltip(Range);
+  const SliderWithTooltip = SliderComponent;
+  const [slidervalue, setSliderValue] = useState<number>(value || min);
+  const onSliderChange = (v: number) => {
+    setSliderValue(v);
+    !!onChange && onChange(v);
+  };
+  const onSliderInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
dprokop, line 42: useCallback maybe?

packages/grafana-ui/src/components/Slider/RangeSlider.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  //   expect(wrapper.html()).toContain('aria-valuenow="10"');
+  //   expect(wrapper.html()).toContain('aria-valuenow="20"');
+  // });
+
+  // it('renders correct contents with a value', () => {
+  //   const wrapper = mount(<RangeSlider {...sliderProps} value={[15, 10]} />);
+  //   expect(wrapper.html()).toContain('aria-valuenow="[15, 10]"');
+  //   expect(wrapper.html()).not.toContain('aria-valuenow="20"');
+  //   expect(wrapper.html()).not.toContain('aria-valuenow="10"');
+  // });
```
dprokop, line 31: Can you please remove this code and move the range slider input work into a separate PR? Thanks!

packages/grafana-ui/src/components/Slider/RangeSlider.test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { RangeSlider, Props } from './RangeSlider';
+import { mount } from 'enzyme';
```
dprokop, line 3: Please use testing-library/react instead? We are discontinuing using Enzyme.

packages/grafana-data/src/utils/OptionsUIBuilders.ts
```diff
@@ -136,6 +149,16 @@ export class PanelOptionsEditorBuilder<TOptions> extends OptionsUIRegistryBuilde
     });
   }
 
+  addSliderInput<TSettings>(
+    config: PanelOptionsEditorConfig<TOptions, TSettings & SliderFieldConfigSettings, number | number[]>
```
dprokop, line 153:
```suggestion
    config: PanelOptionsEditorConfig<TOptions, TSettings & SliderFieldConfigSettings, number>
```

dprokop (changes requested):
@isaozlerfm outstanding work, thanks so much for nailing this! Added few comments and I think addressing those should make this PR mergable. Also - could you please sync with master?

Adding @ivanahuckova as a reviewer as this work touches Reach History UI.

ryantxu:
It may be late to chime in...  and I am a bit of a broken record with "Antd is a really nice library" -- but I really like most of the options in https://ant.design/components/slider/

isaozler:
> It may be late to chime in... and I am a bit of a broken record with "Antd is a really nice library" -- but I really like most of the options in https://ant.design/components/slider/

saw a comment in `uiElements.tsx` about "getting rid of AntD UI library that was used by default", so don't know what the vision here is

packages/grafana-ui/src/components/Slider/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  onChange?: (value: number) => void;
+  onAfterChange?: (value?: number) => void;
+}
+
+export interface RangeSliderProps {
+  min: number;
+  max: number;
+  orientation?: Orientation;
+  /** Set current positions of handle(s). If only 1 value supplied, only 1 handle displayed. */
+  value?: number[];
```
mckn, line 22: Do we have any scenario where thus could be more than 2 values? I'm thinking that we might be able to have it as `from: number, to:number`.

mckn (approved):
I think it looks really good! One small question regarding the API for the range slider. But I think this is a great start so I would say focus on getting this merged and then extend the API more in future PRs if needed.

ivanahuckova (approved):
Really good job and thanks so much for your contribution 🧡. I have tested new RangeSlider with the Rich history and all looks good. LGTM!

isaozlerfm:
> P

exactly 👍  thx

isaozlerfm:
> I think it looks really good! One small question regarding the API for the range slider. But I think this is a great start so I would say focus on getting this merged and then extend the API more in future PRs if needed.

exactly 👍 thx

packages/grafana-ui/src/components/Slider/RangeSlider.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { cx } from 'emotion';
+import { Global } from '@emotion/core';
+import { useTheme } from '../../themes/ThemeContext';
+import { getStyles } from './styles';
+import { RangeSliderProps } from './types';
+
+/**
+ * @public
+ *
+ * RichHistoryQueriesTab uses this Range Component
```
dprokop, line 12: Think we can skip this comment. We should tend to add comments that are meaningful for someone consuming grafana/ui APIs. If I were developing a new plugin and wanted to use this component I would find this information unnecessary.

dprokop (approved):
Looks great, thank you one more time @isaozlerfm!

PR #28010, opened 2020-10-02
Fixes #27989 - support periods in ES annotation field names

### LINKED ISSUE #27989: Elasticsearch annotations don't work with field names that contain periods
summary by @gabor: in elasticsearch annotations, when entered field-names contain periods (`.`), we assume it's a nested structure, and recursively try to find the real field. so if it is a not-nested field that has a `.` in it's name, it does not work. even if we change this, we need to be careful with backward-compatibility. (the related code: https://github.com/grafana/grafana/blob/66b54e2dbd10fd399e2034b3367bbb579b8a0635/public/app/plugins/datasource/elasticsearch/datasource.ts#L330-L341 )

---
**What happened**:

I'm using an elasticsearch datasource for annotations and would like to use my fields for the "start" and "end" time. The field names are `p4.start_time` and `p4.end_time`.

Note - periods are allowed, but are probably not a great idea. However, because many datasources may already contain fields with period names it would be really useful to support either automatically handling it or escaping the periods somehow with quotes or other escape character.

Note: the "beat" tools that the elastic group created makes fairly heavy use of field names containing periods.

The field values work in the annotation

Create an elasticsearch document with text and/or date fields that contain periods (but are not "object" type). Attempt to use this document as an annotation's start, end, or data field value.

**Environment**:
- Grafana version: 
v7.1.1 (3039f9c3bd)

- Data source type & version:
Elasticsearch annotation (built in)

- OS Grafana is installed on:
Ubuntu / Kubernetes

- User OS & Browser:
Chrome

- Grafana plugins:
none

seanhoughton:
This fix allows  elasticsearch annotation fields to have periods in field names. Users must use quotes around the fields with periods to prevent resolving sub-object fields.

For example
```
my."field.with.dots".value
```

Fixes #27989

This change allows users to reference fields with dots in them in the Elasticsearch annotation configuration by adding quotes around the portions of the value that should not have periods resolved as sub-objects.

Notes:

I adapted some code from another project in this PR and it may have to be re-written to be compatible with whatever open-source licensing requirements that code had. This is kind of a draft / proof-of-concept. If this looks good I can re-write the code.

Also, I couldn't find any tests that exercise the annotation query code. If it exists and I missed it let me know and I'll add some coverage for this change.

### DIFF
--- a/docs/sources/features/datasources/elasticsearch.md
+++ b/docs/sources/features/datasources/elasticsearch.md
@@ -193,6 +193,8 @@ for annotation events.
 | Text     | Event description field.                                                                                                                   |
 | Tags     | Optional field name to use for event tags (can be an array or a CSV string).                                                               |
 
+Note: field names with dot characters (i.e. `.`) must be surrounded in quotes. For example: `item."field.with.dots".value`
+
 ## Querying Logs (BETA)
 
 > Only available in Grafana v6.3+.
--- a/public/app/plugins/datasource/elasticsearch/datasource.ts
+++ b/public/app/plugins/datasource/elasticsearch/datasource.ts
@@ -167,6 +167,44 @@ export class ElasticDatasource extends DataSourceApi<ElasticsearchQuery, Elastic
     });
   }
 
+  // https://github.com/elgs/splitargs/blob/master/splitargs.js
+  private splitOnDot(input: string) {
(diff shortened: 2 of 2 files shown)


davkal:
Good stuff! Could you add a test as well for the split dots logic? (Or include it via npm, since the source has a test.)

PR #28011, opened 2020-10-02
Dashboard: Add Datetime local (No date if today) option in panel axes' units

### LINKED ISSUE #27803: Datetime local (No date if today)
It would be nice to get the Datetime local (No date if today) unit.

**Why is this needed**:
There is "No date if today" option available in Datetime ISO and Datetime US units, but not in Datetime local unit.

triptu:

This adds one more option in the units dropdown to skip the date part if date is today. It'll make date as local option consistent with the other two formatting options being provided.

image

Fixes #27803 

I'm not sure how to write tests for this as the date format may vary depending on the locale. Let me know if there is a better way than what I have currently done.


packages/grafana-data/src/valueFormats/dateTimeFormatters.ts
```diff
@@ -370,18 +370,13 @@ export const dateTimeAsIsoNoDateIfToday = toDateTimeValueFormatter('YYYY-MM-DD H
...
-    localTimeFormat({
-      year: 'numeric',
-      month: '2-digit',
-      day: '2-digit',
-      hour: '2-digit',
-      minute: '2-digit',
-      second: '2-digit',
-    })
-  );
-}
```
torkelo, line 384: this was done in a function for a reason, as the local stuff access Intl api which is not available from node environments

triptu: Makes sense I have updated the PR.

kaydelaney:
Rather than having additional options with "No date if today" for each choice, perhaps a toggle with that option would be more appropriate?

PR #22383, opened 2020-02-22
Generic OAuth: Add support for requiring role

burakksglu:
**What this PR does / why we need it**:

This PR adds a 'login restriction by role' functionality for Generic Auth role mapping. This way if "required_role" is enabled from config, we can allow login only for 'Admin', 'Editor' and 'Viewer' roles. If it is disabled "auto_assign_org_role" is still active and works.

**Which issue(s) this PR fixes**:

This PR fixes the issue of the entry of those who do not have a role. For example, if a user from Keycloak has no role for Grafana client and he tries to login the client anyway, he can still login with some role. What we want is in that situation he shouldn't be able to login that Grafana client. 

Therefore this feature adds a functionality for those who wants to restrict login by role for Generic Auth.

**Special notes for your reviewer**:

We tested this feature on a local Keycloak server and it works as intended. 

**If 'role_required' is true from sample.ini config file:**

_1- User has a Grafana role from Keycloak ( Admin/Editor/Viewer ) he can login to the client.
2- User doesn't have a Grafana role from Keycloak but has something else ( Guest,user,etc. ) he can't login to the client.
3- User doesn't have ANY role from Keycloak, he can't login to the client._ 

**If 'role_required' is false from sample.ini config file:**

_Login works as previous, if auto_assign is enable it assigns corresponding role for the user and let's the user in to client._

Since this is my first PR to Grafana community and I am not a Go expert, I do need a help on documentation and other necessary actions for this PR. 

Thanks!

PR #28021, opened 2020-10-03
OAuth: Add optional strict parsing of role_attribute_path

### LINKED ISSUE #23218: Auth: Support for restricting access if certain criteria aren't met
I'd like to see this PR added in the upcoming releases: grafana/grafana/pull/22383

This PR adds a 'login restriction by role' functionality for Generic Auth role mapping. This way if "required_role" is enabled from config, we can allow login only for 'Admin', 'Editor' and 'Viewer' roles. If it is disabled "auto_assign_org_role" is still active and works.

**Why is this needed**:

This PR fixes the issue of the entry of those who do not have a role. For example, if a user from Keycloak has no role for Grafana client and he tries to login the client anyway, he can still login with some role. What we want is in that situation he shouldn't be able to login that Grafana client.

Therefore this feature adds a functionality for those who wants to restrict login by role for Generic Auth.

**Please include your OAuth provider and possible version**:

We have used and tested with Keycloak v9.0.0.

**Why is there a problem with someone in your organization getting Viewer access to Grafana?**:

We've lots of Grafana instances in a docker environment. In Keycloak you create users under realms. Then you can create clients (applications) under this realm, give access to these clients to users under this realm again. So let's say I gave a role to user for Grafana_1 but not on the Grafana_2. That user, without any role from Keycloak, can still login to Grafana_2. This is what we tried to restrict. Because these are different instances, not related to each other and shouldn't be seen by another user from other instance.

 **Have you considered disable allow_sign_up and adding users manually in Grafana? If so, why is that solution not optimal for you given restricting access for certain users is a big problem for you?**:

Yes we tried that. It was not manageable in our use case. We have different teams in our organizations, different instances for different sites etc. Managing and organizing 30+ or 50+ Grafana instances by hand is very time consuming. In Keycloak we add those user just by once. We create only clients with roles with them in Keycloak and give access to specific team or user.

### LINKED ISSUE #26626: When user fails role_attribute_path, user is still granted "Viewer" role
**What happened**:
* Created a `role_attribute_path` jmespath that maps specific oauth `/userInfo` to `Admin` and `Viewer` roles.
* If a user with oauth authentication fails this authorization check, the user is by default given the `Viewer` role anyway.

* I expect that if a user succeeds in authentication, but fails the `role_attribute_path` authorization check, that the user is given an unauthorized role.

Simple Repro Config
1) Configure oauth login with an oauth system of your choice
2) Create a `role_attribute_path` rule that provides someone with a specific email address the `Admin` role and fails for any other email address.
3) Login with a different user/email than what was configured in step 2 which successfully authenticates with the configured oauth system from step 1.
4) Verify that you have `Viewer` access even though you failed all `role_attribute_path` checks.

**Environment**:
- Grafana version: 7.0
- OS Grafana is installed on: published docker image
- User OS & Browser: n/a
- Grafana plugins: n/a
- Others:

klausenbusk:

It is currently not possible to deny users access to Grafana when using [generic OAuth authentication](https://grafana.com/docs/grafana/latest/auth/generic-oauth/). 

Fixes #23218


achatterjee-grafana (dismissed):
Thank you for contributing.

sakjur (dismissed):
(no changes requested, but I don't want this to be merged before we've discussed this a little)

I'd personally rather see a boolean that makes the parsing of roles strict and denies any non-role results 🤔
It seems bad that say someone accidentally writing `unauthorised` would give the users access.

So something like:

```
role_attribute_strict = true
```

meaning that anything except for `Viewer`, `Editor`, or `Admin` would be unauthorized.

alexanderzobnin:
@sakjur Yes, it definitely makes sense.

sakjur:
@klausenbusk What do you think about changing this to add an optional strict mode instead of hard-coding the deny access role?

klausenbusk:
> @klausenbusk What do you think about changing this to add an optional strict mode instead of hard-coding the deny access role?

It sounds like a better approach. I will try to implement it in the weekend (I don't have much time this week :/).

klausenbusk:
I haven't got time to test it (yet), but how does it looks?

sakjur:
Based on just looking at it it looks really good! Pretty much exactly like I'd do it myself.

Let me know when you've tested this and I'll test it on my end. After we've tested this, we'll also need to document the option and then I'd say it's pretty much ready for merge.

marefr:
@klausenbusk i closed your referenced issue as a duplicate of #23218. We've already closed the #22383 related to #23218.

@sakjur based on #22383 we need to discuss what to do with this one.

@klausenbusk we'll discuss this and come back with an answer.

sakjur:
@marefr :+1: I wasn't aware of https://github.com/grafana/grafana/pull/22383#issuecomment-606071605 I am a little concerned that we're making perfect to the enemy of the good here. I think this is a change for the better, making a feature (optionally) work the way it should've worked all along.

We're going to have to deprecate/support the `role_attribute_path` as we're moving to a better authentication solution. The difference between having a `role_attribute_strict` next to it as well isn't all too significant.

achatterjee-grafana (commented):
Looks good to me, but Dev will approve for merge.

marefr:
@sakjur lets pick this up after 7.3 stable has been released and discuss it

klausenbusk:
> @sakjur lets pick this up after 7.3 stable has been released and discuss it

Any update?

sakjur:
Yes, @marefr and I discussed this last week.

Our conclusion is that we (it's on my todo list) should update the documentation to make it clear that the default role is being used when an invalid role is returned and that that's something that might change in the future.

Auth in general is in a very messy state at the moment, and it is one of the areas we'd like to work on for 8.0, which means we want to take this PR into the context of that work. I'm still much in favor of adopting this behavior as I think it is the way it should work, but we want to make sure that regardless of if you use Grafana with OAuth, LDAP, vendor-specific auth, or whatnot we would like to handle it in the same way going forward. We don't want to further extend the gap between the different mechanisms at the moment (security rationale excluded — which is in a sense an edge-case for this particular feature, but still [in mine and Marcus's opinion] on the side of enhancement rather than bug fix).

In short, we won't be merging this PR for 7.4 to make sure we're not undoing it in 8.0 following soon after.

sakjur:
Being a bit further down the line with our internal discussions: this is a good idea and a good implementation thereof. Thank you for contributing to Grafana, @klausenbusk 🙏 

This PR looks good and can make it into Grafana 8.0 if its updated to be based on the most recent master and add documentation for the new option (for both Okta and Generic OAuth).

klausenbusk:
Rebased, added some documentation, squashed and I rewrote the commit message.

This is a reference in the Okta docs:
> Also, check Generic OAuth page for [JMESPath examples]({{< relref "generic-oauth.md/#jmespath-examples" >}}).

So I only described the strict mode in the generic oauth docs, perhaps it should be documented at both pages?

Not tested, but I think it is ready for review.

docs/sources/auth/generic-oauth.md
```diff
@@ -194,6 +194,8 @@ To ease configuration of a proper JMESPath expression, you can test/evaluate exp
 
 ### Role mapping
 
+By default the user is assigned the role `Viewer` if `role_attribute_path` doesn't return a role for the user, this is not always desirable and can be disabled by setting `role_attribute_strict = true`, which will deny the user access if no role is returned.
```
sakjur, line 197:
```suggestion
By default the user is assigned the role `Viewer` if `role_attribute_path` doesn't return a role for the user, this is not always desirable and can be disabled by setting `role_attribute_strict = true`, which will deny the user access if no role or an invalid role is returned.
```

I'm not reviewing the documentation, our docs squad will do that, but I think this might be worth mentioning 🙂

sakjur (approved):
Code looks good, I'll wait for someone in the documentations team to review the docs for this :+1:

docs/sources/auth/generic-oauth.md
```diff
@@ -194,6 +194,8 @@ To ease configuration of a proper JMESPath expression, you can test/evaluate exp
 
 ### Role mapping
 
+By default the user is assigned the role `Viewer` if `role_attribute_path` doesn't return a role for the user, this is not always desirable and can be disabled by setting `role_attribute_strict = true`, which will deny the user access if no role or an invalid role is returned.
```
achatterjee-grafana, line 197:
```suggestion
By default, if  the`role_attribute_path` does not return a role, the user is assigned the `Viewer` role. You can disable the role assignment by setting `role_attribute_strict = true`. It denies user access if no role or an invalid role is returned.
```

achatterjee-grafana (changes requested):
Added doc change request.

achatterjee-grafana (approved):

aocenas:
Hey @klausenbusk there is an issue with the build which is fixed on master. Can you pull the latest master, merge it with your branch and push that commit? That should fix the CI failure.

klausenbusk:
> Can you pull the latest master, merge it with your branch and push that commit?

Sure, done. I normally prefer rebasing but whatever works :)

burakksglu:
Glad to see this feature will come to Grafana in the future. We've used Keycloak to do this like mentioned in the #22383 but it wasn't the best solution. Thank you @klausenbusk & Grafana team!

Issue #4355, opened 2016-03-14
Datasource: Live Websocket based data source


torkelo:
**UPDATE**: Published [design document](https://docs.google.com/document/d/15VgeOQSVJ8aBkVPlv6KodtwiPboW_XrbA-lKOft5Uk8/edit?usp=sharing) and requesting comments.

--

allow for live data to be pushed to graphs and panels via Grafana HTTP api. 

for example: 

```http
POST /api/streams/:<stream_name>/
{
"type": "time-series",
"delta": true,
"metric": "name",
"tags": {
"host": "server-1",
"data-center": "us-east-02",
},
"datapoints": [
]
}
```

digrich:
@torkelo Do we have an documentation on this? What should be the data source API to support websocket streaming on grafana? In one of your video, you were mentioning that Intel's snap integration supports real time push of data, does that uses websocket?

torkelo:
no, that one uses streaming features in snap http api.  But the snap data source is using a feature in Grafana that allows data sources to push data to panels (instead of panels only doing pull)

mtanda:
I create a datasource plugin by using Grafana Live. (I'm not sure this is good example.)
https://github.com/mtanda/grafana-prometheus-pull-datasource

To create datasource plugin, you should learn about RxJS.
https://github.com/Reactive-Extensions/RxJS

digrich:
Thanks. I will take a look.

fti7:
@torkelo Do we have a new planned Milestone Version for this important Feature request? :-)

torkelo:
no not yet

davkal:
We're currently spec'ing this out. Here is a [design document](https://docs.google.com/document/d/15VgeOQSVJ8aBkVPlv6KodtwiPboW_XrbA-lKOft5Uk8/edit?usp=sharing). I request you all comment there.

lssilva:
This would be a great feature to have.

lssilva:
Has there been progress in this direction?

torkelo:
Sorry, stale bot was wrongly configured (should only have acted on PRs)

Scavallarin:
Hi! There has been any news into this? From what i saw there is only this so far:
https://github.com/mtanda/grafana-prometheus-pull-datasource
It would be such an amazing component really useful!
Thanks

ryantxu:
This is now part of grafana-live:  https://grafana.com/docs/grafana/latest/live/

PR #28022, opened 2020-10-03
Live: support real time measurements (alpha)

ryantxu:
This PR adds an initial alpha implementation for generic real-time measurement support.

This PR currently supports:
1. POST a measurement batch to the `api/live/grafana/measurements/{path}` endpoint
1. the values are broadcast to everyone listening on this channel
1. the front end will buffer these points in a circular buffer.  
1. values can be added to panels using the `-- Grafana --` datasource
1. improves the (alpha) live panel, so we can send data to channels

This is still early and experimental, however I think it is worth merging for 7.3.  The interface that should get the most attention is:
https://github.com/grafana/grafana/pull/28022/files#diff-1fbde96172d0110468fcb647d365f52729f61cf72856c91ab6113d611a946b1cR33

There is one *massive* issue with streaming as currently implemented.  We do not yet handle unsubscribing!  We *could* unsubscribe when the rxjs observable is done -- however, this is not great for lots of standard features where you navigate from one dashboard to another and expect to see the same data but displayed differently.

The easiest way to see somethign is to use the "Grafana Live" option under testdata:  This will add a point to the graph every 2s

image


seanlaff:
Excellent! :)

packages/grafana-runtime/src/measurement/query.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const channel = live.getChannel<LiveMeasurements>(addr);
+  const getController = channel?.config?.getController;
+  return getController ? getController() : undefined;
+}
+
+/**
+ * When you know the stream will be managed measurments
+ */
+export function getLiveMeasurmentsObserver(
```
ryantxu, line 33: @hugohaggmark this is the function I think needs the most attention.

pkg/api/api.go
```diff
@@ -388,6 +388,11 @@ func (hs *HTTPServer) registerRoutes() {
 			annotationsRoute.Post("/graphite", reqEditorRole, bind(dtos.PostGraphiteAnnotationsCmd{}), Wrap(PostGraphiteAnnotation))
 		})
 
+		// The live handler will manage its own routes
+		if hs.Live != nil {
+			apiRoute.Any("/live/*", hs.LiveHTTPHandler)
```
ryantxu, line 393: I tried using:
```suggestion
			apiRoute.Any("/live/*", Wrap(hs.LiveHTTPHandler))
```
but the circular imports to use any response helpers was too difficult to sort out

packages/grafana-runtime/src/measurement/query.ts
```diff
@@ -0,0 +1,72 @@
...
+          rsp.state = LoadingState.Streaming;
+        } else if (isLiveChannelStatusEvent(evt)) {
+          if (evt.error != null) {
+            rsp.error = rsp.error;
+            rsp.state = LoadingState.Error;
+          }
+        }
+        return { ...rsp }; // send event on all status messages
+      })
+    );
```
hugohaggmark, line 71: nit: this looks fine but everytime you introduce an if statement in RxJs operator you should always ask yourself if that operator can be divided into 2 streams using `filter` and then merge the streams in the return.

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -6,54 +6,63 @@ import {
...
+    }
+    // With a single query just return the results
+    if (queries.length === 1) {
+      return queries[0];
+    }
+    if (queries.length > 1) {
+      // HELP!!!
+      return queries[0];
+    }
+    return of(); // nothing
```
hugohaggmark, line 57: Maybe something like this?
```suggestion
     return forkJoin(queries).pipe(map((results:DataFrame[]) => ... )))
```

public/app/core/components/TransformersUI/ReduceTransformerEditor.tsx
```diff
@@ -1,42 +1,103 @@
...
+    },
+    [onChange, options]
+  );
+
+  const onToggleTime = useCallback(() => {
+    onChange({
+      ...options,
+      includeTimeField: !options.includeTimeField,
+    });
+  }, [onChange, options]);
```
hugohaggmark, line 50: I don't understand how changes in this file relates to changes in live? Maybe open in a separate PR?

public/app/plugins/datasource/grafana/datasource.ts
hugohaggmark: I did something similar in https://github.com/grafana/grafana/pull/28221 so maybe we should brea this out into a utility function?

public/app/plugins/panel/live/LiveChannelEditor.tsx
```diff
@@ -135,7 +141,7 @@ export class LiveChannelEditor extends PureComponent<Props, State> {
               <Label>Path</Label>
               <Select
                 options={paths}
-                value={paths.find(s => s.value === path) || path || ''}
+                value={paths.find(s => s.value === path) ?? (path ? { label: path, value: path } : undefined)}
```
hugohaggmark, line 144: nit: maybe break out this value entries into functions?

public/app/core/components/TransformersUI/ReduceTransformerEditor.tsx
ryantxu: dooh -- that is part of https://github.com/grafana/grafana/pull/27875

these three PRs are all required to make some good demos work ;)

hugohaggmark (commented):
This is massive work but I can't follow all the changes because I don't understand the overall picture :) so I think this PR could probably be broken up into smaller pieces or maybe it would be easier to follow if there were steps on how to test this "live" in Grafana?

Is this something that is being demoed at ObsCon because at some places it doesn't feel that clean and it doesn't build?

aknuds1:
> This is massive work but I can't follow all the changes because I don't understand the overall picture :) so I think this PR could probably be broken up into smaller pieces or maybe it would be easier to follow if there were steps on how to test this "live" in Grafana?
> 
> Is this something that is being demoed at ObsCon because at some places it doesn't feel that clean and it doesn't build?

I'm also wondering if this has to go into 7.3.0-beta1 already, since it's a huge PR.

pkg/api/live_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	c.JSON(404, util.DynMap{})
+}
+
+func getChannelAddress(path string) (live.ChannelAddress, bool) {
+	idx := strings.Index(path, "/api/live/")
+	if idx < 0 {
+		return live.ChannelAddress{}, false
+	}
+	return live.ParseChannelAddress(path[(idx + len("/api/live/")):]), true
+}
```
aknuds1, line 58: Can we just inline this function? LiveHTTPHandler is already relatively short and I don't see any benefit from the abstraction.

pkg/models/live.go
```diff
@@ -15,12 +15,18 @@ type ChannelHandler interface {
...
 	// Called when something writes into the channel.  The returned value will be broadcast if len() > 0
 	OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error)
+
+	// Called from the HTTP API
+	DoChannelHTTP(c *ReqContext, channel string)
 }
 
-// ChannelHandlerProvider -- this should be implemented by any core feature
-type ChannelHandlerProvider interface {
+// ChannelNamespaceHandler -- this should be implemented by any core feature
```
aknuds1, line 23: Documentation comments should be sentences.
```suggestion
// ChannelNamespaceHandler should be implemented by any core feature.
```

pkg/models/live.go
```diff
@@ -15,12 +15,18 @@ type ChannelHandler interface {
...
+type ChannelNamespaceHandler interface {
 	// This is called fast and often -- it must be synchrnozed
 	GetHandlerForPath(path string) (ChannelHandler, error)
+
+	// Called from the HTTP API
```
aknuds1, line 28:
```suggestion
	// DoNamespaceHTTP is called from the HTTP API
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,35 @@
...
+
+// NOTE:  this likely shoud go in the Plugin SDK since it will be useful from plugins
+
+// Measurement is a single measurement value
+type Measurement struct {
+	Name   string                      `json:"name,omitempty"`
+	Time   int64                       `json:"time,omitempty"`   // units are usually ms, but depend on the channel
+	Values map[string]interface{}      `json:"values,omitempty"` // typically number or string
+	Config map[string]data.FieldConfig `json:"config,omitempty"` // optional list of field configs
+	Labels map[string]string           `json:"labels,omitempty"` // labels are applied to all values
```
aknuds1, line 13: The comments should go above and be on sentence form. When we enable better linting, it will catch this.

pkg/models/measurement.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// MeasurementActionClear means all existing values should be remoed before adding
+	MeasurementActionClear MeasurementAction = "clear"
+)
+
+// MeasurementBatch is a collection of measurments all sent at once
+type MeasurementBatch struct {
+	Action   MeasurementAction `json:"action,omitempty"`   // default action is append
+	Measures []Measurement     `json:"measures,omitempty"` // batch of measurments
+	Capacity int64             `json:"capacity,omitempty"` // The suggested size of the client buffer
```
aknuds1, line 34: The comments should go above and be on sentence form. When we enable better linting, it will catch this.

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
-// ChannelIdentifier is the channel id split by parts
-type ChannelIdentifier struct {
-	Scope     string // grafana, ds, or plugin
-	Namespace string // feature, id, or name
-	Path      string // path within the channel handler
+// ChannelAddress is the channel id split by parts
+type ChannelAddress struct {
+	Scope     string `json:"scope,omitempty"`     // grafana, ds, or plugin
+	Namespace string `json:"namespace,omitempty"` // feature, id, or name
+	Path      string `json:"path,omitempty"`      // path within the channel handler
```
aknuds1, line 11: The comments should go above and be on sentence form. When we enable better linting, it will catch this.

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
+	if length > 2 {
+		identifier.Path = parts[2]
+	}
+	return identifier
+}
+
+// IsValid checks if all parts of the address are valid
+func (id *ChannelAddress) IsValid() bool {
+	return id.Scope != "" && id.Namespace != "" && id.Path != ""
+}
```
aknuds1, line 35: The receiver name `id` is confusing here.
```suggestion
func (ca *ChannelAddress) IsValid() bool {
	return ca.Scope != "" && ca.Namespace != "" && ca.Path != ""
}
```

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
+
+// ToChannelID converts this to a single string
+func (id *ChannelAddress) ToChannelID() string {
```
aknuds1, line 38: The receiver name id is confusing here.
```suggestion
func (ca *ChannelAddress) ToChannelID() string {
```

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
+	return id.Scope + "/" + id.Namespace + "/" + id.Path
```
aknuds1, line 39:
```suggestion
	return ca.Scope + "/" + ca.Namespace + "/" + ca.Path
```

pkg/services/live/features/broadcast.go
```diff
@@ -1,32 +1,86 @@
...
+}
 
 // GetHandlerForPath called on init
-func (g *BroadcastRunner) GetHandlerForPath(path string) (models.ChannelHandler, error) {
-	return g, nil // for now all channels share config
+func (b *BroadcastRunner) GetHandlerForPath(path string) (models.ChannelHandler, error) {
+	return b, nil // for now all channels share config
+}
+
+// DoNamespaceHTTP called from the http api
```
aknuds1, line 22:
```suggestion
// DoNamespaceHTTP is called from the HTTP API.
```

pkg/services/live/features/broadcast.go
```diff
@@ -1,32 +1,86 @@
...
+func (b *BroadcastRunner) DoNamespaceHTTP(c *models.ReqContext) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": "BroadcastRunner",
```
aknuds1, line 25:
```suggestion
		"Unsupported": "BroadcastRunner",
```

pkg/services/live/features/broadcast.go
```diff
@@ -1,32 +1,86 @@
...
 // OnPublish called when an event is received from the websocket
-func (g *BroadcastRunner) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
+func (b *BroadcastRunner) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
 	// expect the data to be the right shape?
 	return e.Data, nil
 }
+
+// DoChannelHTTP accepts POST from anywhere
+func (b *BroadcastRunner) DoChannelHTTP(c *models.ReqContext, channel string) {
+	if c.Req.Method == "POST" {
```
aknuds1, line 51: It's better/more idiomatic to invert the `if` here and fail and return if the method isn't "POST". It will make the code more readable due to reduced nesting.

pkg/services/live/features/dashboard.go
```diff
@@ -17,21 +18,21 @@ type dashboardEvent struct {
...
-	}
+	Publisher models.ChannelPublisher
 }
 
 // GetHandlerForPath called on init
 func (g *DashboardHandler) GetHandlerForPath(path string) (models.ChannelHandler, error) {
 	return g, nil // all dashboards share the same handler
 }
 
+// DoNamespaceHTTP called from the http api
```
aknuds1, line 29:
```suggestion
// DoNamespaceHTTP is called from the HTTP API.
```

pkg/services/live/features/dashboard.go
```diff
@@ -52,13 +53,20 @@ func (g *DashboardHandler) OnPublish(c *centrifuge.Client, e centrifuge.PublishE
 	return e.Data, nil
 }
 
+// DoChannelHTTP called from the http api
```
aknuds1, line 56:
```suggestion
// DoChannelHTTP is called from the HTTP API.
```

pkg/services/live/features/dashboard.go
```diff
@@ -17,21 +18,21 @@ type dashboardEvent struct {
...
 
 // GetHandlerForPath called on init
 func (g *DashboardHandler) GetHandlerForPath(path string) (models.ChannelHandler, error) {
 	return g, nil // all dashboards share the same handler
 }
 
+// DoNamespaceHTTP called from the http api
+func (g *DashboardHandler) DoNamespaceHTTP(c *models.ReqContext) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": "DashboardHandler",
```
aknuds1, line 32:
```suggestion
		"Unsupported": "DashboardHandler",
```

pkg/services/live/features/dashboard.go
```diff
@@ -52,13 +53,20 @@ func (g *DashboardHandler) OnPublish(c *centrifuge.Client, e centrifuge.PublishE
 	return e.Data, nil
 }
 
+// DoChannelHTTP called from the http api
+func (g *DashboardHandler) DoChannelHTTP(c *models.ReqContext, channel string) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": channel,
```
aknuds1, line 59:
```suggestion
		"Unsupported": channel,
```

packages/grafana-runtime/src/measurement/query.ts
torkelo: why is that preferable? Seems more complex & inefficient? Can understand it if you after the filtering need to do different async stream processing

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+	logger = log.New("live.features") // scoped to all features?
+)
+
+// MeasurementsRunner will simply broadcast all events to `grafana/broadcast/*` channels
+// This makes no assumptions about the shape of the data and will broadcast it to anyone listening
+type MeasurementsRunner struct {
+	Publisher models.ChannelPublisher
+}
+
+// GetHandlerForPath called on init
```
aknuds1, line 22:
```suggestion
// GetHandlerForPath is called on init.
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+func (m *MeasurementsRunner) GetHandlerForPath(path string) (models.ChannelHandler, error) {
+	return m, nil // for now all channels share config
+}
+
+// DoNamespaceHTTP called from the http api
```
aknuds1, line 27:
```suggestion
// DoNamespaceHTTP is called from the HTTP API.
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+func (m *MeasurementsRunner) DoNamespaceHTTP(c *models.ReqContext) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": "MeasurementsRunner",
```
aknuds1, line 30:
```suggestion
		"Unsupported": "MeasurementsRunner",
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+	})
+}
+
+// GetChannelOptions called fast and often
```
aknuds1, line 34:
```suggestion
// GetChannelOptions gets channel options.
// It gets called fast and often.
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+	return centrifuge.ChannelOptions{}
+}
+
+// OnSubscribe for now allows anyone to subscribe to any dashboard
+func (m *MeasurementsRunner) OnSubscribe(c *centrifuge.Client, e centrifuge.SubscribeEvent) error {
+	// anyone can subscribe
+	return nil
+}
+
+// OnPublish called when an event is received from the websocket
```
aknuds1, line 45:
```suggestion
// OnPublish is called when an event is received from the websocket.
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+func (m *MeasurementsRunner) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
+	// currently generic... but should be more strict
+	// logger.Debug("GOT: %s", e.Channel)
+	return e.Data, nil
+}
+
+// DoChannelHTTP called from the http api
```
aknuds1, line 52:
```suggestion
// DoChannelHTTP is called from the HTTP API.
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+func (m *MeasurementsRunner) DoChannelHTTP(c *models.ReqContext, channel string) {
+	if c.Req.Method == "POST" {
```
aknuds1, line 54: See my previous comment about returning early on error.

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+				"error":   err.Error(),
+			})
+			return
+		}
+
+		msg := &models.MeasurementBatch{}
+		err = json.Unmarshal(body, &msg)
+		if err != nil {
+			c.JSON(500, util.DynMap{
+				"message": "body must be measurment batch",
```
aknuds1, line 68:
```suggestion
				"message": "body must be measurement batch",
```

pkg/services/live/features/measurements.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		c.JSON(200, util.DynMap{
+			"message": "OK",
+		})
+		return
+	}
+
+	c.JSON(400, util.DynMap{
+		"unsuppoted?": channel,
```
aknuds1, line 90:
```suggestion
		"unsupported?": channel,
```

pkg/services/live/features/testdata.go
```diff
@@ -18,18 +18,19 @@ type testdataRunner struct {
...
 
-// CreateTestdataSupplier Initialize a dashboard handler
-func CreateTestdataSupplier(p models.ChannelPublisher) TestdataSupplier {
-	return TestdataSupplier{
-		publisher: p,
-	}
+// DoNamespaceHTTP called from the http api
+func (g *TestdataSupplier) DoNamespaceHTTP(c *models.ReqContext) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": "TestdataSupplier",
```
aknuds1, line 32:
```suggestion
		"Unsupported": "TestdataSupplier",
```

pkg/services/live/features/testdata.go
```diff
@@ -82,11 +84,11 @@ func (g *testdataRunner) OnPublish(c *centrifuge.Client, e centrifuge.PublishEve
...
 
-type randomWalkMessage struct {
-	Time  int64
-	Value float64
-	Min   float64
-	Max   float64
+// DoChannelHTTP called from the http api
+func (g *testdataRunner) DoChannelHTTP(c *models.ReqContext, channel string) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": channel,
```
aknuds1, line 90:
```suggestion
		"Unsupported": channel,
```

pkg/services/live/features/testdata.go
```diff
@@ -105,20 +114,20 @@ func (g *testdataRunner) runRandomCSV() {
...
+		bytes, err := json.Marshal(&msg)
 		if err != nil {
 			logger.Warn("unable to marshal line", "error", err)
 			continue
 		}
 
 		err = g.publisher(g.channel, bytes)
 		if err != nil {
-			logger.Warn("write", "channel", g.channel, "line", line)
+			logger.Warn("write", "channel", g.channel, "line", measure)
```
aknuds1, line 130:
```suggestion
			logger.Warn("write", "channel", g.channel, "measure", measure)
```

pkg/services/live/live.go
```diff
@@ -245,39 +253,50 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
...
+// GetChannelNamespace gives threadsafe access to the channel
+func (g *GrafanaLive) GetChannelNamespace(scope string, name string) (models.ChannelNamespaceHandler, error) {
+	if scope == "grafana" {
+		p, ok := g.GrafanaScope.Features[name]
 		if ok {
-			return p.GetHandlerForPath(id.Path)
+			return p, nil
 		}
-		return nil, fmt.Errorf("Unknown feature: %s", id.Namespace)
+		return nil, fmt.Errorf("Unknown feature: %s", name)
```
aknuds1, line 271: I think quoting helps debugging, in case of spaces in name.
```suggestion
		return nil, fmt.Errorf("Unknown feature: %q", name)
```

pkg/services/live/live.go
```diff
@@ -245,39 +253,50 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
...
 	}
 
-	if id.Scope == "ds" {
-		return nil, fmt.Errorf("todo... look up datasource: %s", id.Namespace)
+	if scope == "ds" {
+		return nil, fmt.Errorf("todo... look up datasource: %s", name)
```
aknuds1, line 275:
```suggestion
		return nil, fmt.Errorf("todo... look up datasource: %q", name)
```

pkg/services/live/live.go
```diff
@@ -245,39 +253,50 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
...
+		p, ok := plugins.Plugins[name]
 		if ok {
 			h := &PluginHandler{
 				Plugin: p,
 			}
-			return h.GetHandlerForPath(id.Path)
+			return h, nil
 		}
-		return nil, fmt.Errorf("unknown plugin: %s", id.Namespace)
+		return nil, fmt.Errorf("unknown plugin: %s", name)
```
aknuds1, line 286:
```suggestion
		return nil, fmt.Errorf("unknown plugin: %q", name)
```

pkg/services/live/live.go
```diff
@@ -245,39 +253,50 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
...
+	}
+
+	return nil, fmt.Errorf("invalid scope: %s", scope)
```
aknuds1, line 289:
```suggestion
	return nil, fmt.Errorf("invalid scope: %q", scope)
```

pkg/services/live/pluginHandler.go
```diff
@@ -16,6 +17,13 @@ func (g *PluginHandler) GetHandlerForPath(path string) (models.ChannelHandler, e
 	return g, nil // all dashboards share the same handler
 }
 
+// DoNamespaceHTTP called from the http api
```
aknuds1, line 20:
```suggestion
// DoNamespaceHTTP is called from the HTTP API.
```

pkg/services/live/pluginHandler.go
```diff
@@ -16,6 +17,13 @@ func (g *PluginHandler) GetHandlerForPath(path string) (models.ChannelHandler, e
...
+func (g *PluginHandler) DoNamespaceHTTP(c *models.ReqContext) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": "TestdataSupplier",
```
aknuds1, line 23:
```suggestion
		"Unsupported": "TestdataSupplier",
```

pkg/services/live/pluginHandler.go
```diff
@@ -30,3 +38,10 @@ func (g *PluginHandler) OnSubscribe(c *centrifuge.Client, e centrifuge.Subscribe
 func (g *PluginHandler) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
 	return e.Data, nil // broadcast any event
 }
+
+// DoChannelHTTP called when an event is received from the websocket
+func (g *PluginHandler) DoChannelHTTP(c *models.ReqContext, channel string) {
+	c.JSON(400, util.DynMap{
+		"Unsupportedd": channel,
+	})
```
aknuds1, line 46:
```suggestion
// DoChannelHTTP is called when an event is received from the websocket.
func (g *PluginHandler) DoChannelHTTP(c *models.ReqContext, channel string) {
	c.JSON(400, util.DynMap{
		"Unsupported": channel,
	})
```

aknuds1 (commented):
I have to admit I don't understand yet what this does, and I would like to be able to, but see my comments/suggestions for now.

marefr:
Does this aiming to close #4355?

torkelo:
yes, it's very much that

packages/grafana-runtime/src/measurement/query.ts
hugohaggmark: I can see you point Torkel and maybe this is not the perfect use case because of efficiency and because we don't do anything combined. But usually the streams tend to grow as all other code and then I've found that keeping the logic outside the map functions is easier to read and easier to refactor. 

I was just trying to share some of the insights I found working with larger RxJs implementations, I hold no strong opinions about this.

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -105,3 +114,32 @@ export class GrafanaDatasource extends DataSourceApi<GrafanaQuery> {
     return Promise.resolve();
   }
 }
+
+// Note that the query does not actually matter
+function getRandomWalk(request: DataQueryRequest): Observable<DataQueryResponse> {
+  const { intervalMs, maxDataPoints, range, requestId } = request;
+
+  // Yes, this implementaiton ignores multiple targets!  But that matches exisitng behavior
```
wbrowne, line 122:
```suggestion
  // Yes, this implementation ignores multiple targets!  But that matches existing behavior
```

public/app/plugins/panel/live/module.tsx
```diff
@@ -1,14 +1,37 @@
...
     name: 'Channel',
     editor: LiveChannelEditor,
     defaultValue: {},
   });
+
+  builder
+    .addRadio({
+      path: 'message',
+      name: 'Show Message',
+      description: 'Display the last message recieved on this channel',
```
wbrowne, line 20:
```suggestion
      description: 'Display the last message received on this channel',
```

public/app/plugins/panel/live/module.tsx
```diff
@@ -1,14 +1,37 @@
...
+          { value: MessageDisplayMode.Auto, label: 'Auto' },
+          { value: MessageDisplayMode.None, label: 'None' },
+        ],
+      },
+      defaultValue: MessageDisplayMode.JSON,
+    })
+    .addBooleanSwitch({
+      path: 'publish',
+      name: 'Show Publish',
+      description: 'Display the a form to publish values',
```
wbrowne, line 34: Not sure which is meant here - `Display the form to publish values` / `Display a form to publish values`?

wbrowne (commented):
Have tested a bit and it seems to work as expected. Will continue testing for a bit more.

Found some spelling/grammar mistakes in the meantime

pkg/models/live.go
```diff
@@ -15,12 +15,18 @@ type ChannelHandler interface {
 
 	// Called when something writes into the channel.  The returned value will be broadcast if len() > 0
 	OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error)
+
+	// Called from the HTTP API
```
aknuds1, line 19:
```suggestion
	// DoChannelHTTP is called from the HTTP API
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
+package models
+
+import "github.com/grafana/grafana-plugin-sdk-go/data"
+
+// NOTE:
+// this likely shoud go in the Plugin SDK since it will be useful from plugins
```
aknuds1, line 6:
```suggestion
// this likely should go in the Plugin SDK since it will be useful from plugins
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Measurement is a single measurement value
+type Measurement struct {
+	// the measurement name
```
aknuds1, line 10:
```suggestion
	// Name of the measurement.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Name string `json:"name,omitempty"`
+
+	// units are usually ms, but depend on the channel
```
aknuds1, line 13:
```suggestion
	// Time is the measurement time. Units are usually ms, but depends on the channel
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Time int64 `json:"time,omitempty"`
+
+	// typically number or string
```
aknuds1, line 16:
```suggestion
	// Values is the measurement's values. The value type is typically number or string.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Values map[string]interface{} `json:"values,omitempty"`
+
+	// optional list of field configs
```
aknuds1, line 19:
```suggestion
	// Config is an optional list of field configs.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Config map[string]data.FieldConfig `json:"config,omitempty"`
+
+	// labels are applied to all values
```
aknuds1, line 22: What is the key and value in this map?
```suggestion
	// Labels are applied to all values.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	// MeasurementActionReplace means new values should replace any existing values
+	MeasurementActionReplace MeasurementAction = "replace"
+
+	// MeasurementActionClear means all existing values should be remoed before adding
+	MeasurementActionClear MeasurementAction = "clear"
+)
+
+// MeasurementBatch is a collection of measurments all sent at once
+type MeasurementBatch struct {
+	// default action is append
```
aknuds1, line 42:
```suggestion
	// Action is the action in question, the default is append.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Action MeasurementAction `json:"action,omitempty"`
+
+	// batch of measurments
```
aknuds1, line 45:
```suggestion
	// Measures is the array of measurements.
```

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+	Measures []Measurement `json:"measures,omitempty"`
```
aknuds1, line 46: Should this field be renamed to measurements, to be consistent?? The inconsistent terminology is confusing.

pkg/models/measurement.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// The suggested size of the client buffer
```
aknuds1, line 48:
```suggestion
	// Capacity is the suggested size of the client buffer
```

aknuds1 (changes requested):
See suggestions.

pkg/api/live_handler.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"strings"
+
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/grafana/grafana/pkg/services/live"
+	"github.com/grafana/grafana/pkg/util"
+)
+
+// LiveHTTPHandler will process any requets to /api/live/*
```
aknuds1, line 11:
```suggestion
// LiveHTTPHandler will process any requests to /api/live/*.
```

aknuds1 (changes requested):

aknuds1:
@ryantxu How can one test this PR so the `LiveHTTPHandler` method gets hit? I tested as you describe in the PR description, i.e. via gdev-testdata/Grafana Live, but that doesn't send HTTP requests.

pkg/api/live_handler.go
```diff
@@ -0,0 +1,58 @@
...
+func (hs *HTTPServer) LiveHTTPHandler(c *models.ReqContext) {
```
aknuds1, line 12: I don't see how this handler gets exercised from the frontend, I'd need to know how it's supposed to be used before I can approve.

aknuds1 (changes requested):

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
+	Namespace string `json:"namespace,omitempty"` // feature, id, or name
+	Path      string `json:"path,omitempty"`      // path within the channel handler
 }
 
-// ParseChannelIdentifier parses the parts from a channel id:
+// ParseChannelAddress parses the parts from a channel id:
 //   ${scope} / ${namespace} / ${path}
-func ParseChannelIdentifier(id string) (ChannelIdentifier, error) {
+func ParseChannelAddress(id string) ChannelAddress {
+	identifier := ChannelAddress{}
```
aknuds1, line 17:
```suggestion
	addr := ChannelAddress{}
```

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
 	parts := strings.SplitN(id, "/", 3)
-	if len(parts) == 3 {
-		return ChannelIdentifier{
-			Scope:     parts[0],
-			Namespace: parts[1],
-			Path:      parts[2],
-		}, nil
+	length := len(parts)
+	if length > 0 {
+		identifier.Scope = parts[0]
```
aknuds1, line 21:
```suggestion
	length := len(parts)
	if length > 0 {
		addr.Scope = parts[0]
```

pkg/services/live/channel.go
```diff
@@ -1,27 +1,40 @@
...
 	}
-	return ChannelIdentifier{}, fmt.Errorf("Invalid channel id: %s", id)
+	if length > 1 {
+		identifier.Namespace = parts[1]
+	}
+	if length > 2 {
+		identifier.Path = parts[2]
+	}
+	return identifier
```
aknuds1, line 29:
```suggestion
	if length > 1 {
		addr.Namespace = parts[1]
	}
	if length > 2 {
		addr.Path = parts[2]
	}
	return addr
```

aknuds1 (changes requested):

pkg/services/live/live.go
```diff
@@ -232,9 +240,9 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
 	}
 
 	// Parse the identifier ${scope}/${namespace}/${path}
-	id, err := ParseChannelIdentifier(channel)
-	if err != nil {
-		return nil, err
+	id := ParseChannelAddress(channel)
+	if !id.IsValid() {
+		return nil, fmt.Errorf("invalid channel")
```
aknuds1, line 245: Much better for debugging if the invalid channel is included:
```suggestion
		return nil, fmt.Errorf("invalid channel: %q", channel)
```

pkg/services/live/live.go
```diff
@@ -245,39 +253,50 @@ func (g *GrafanaLive) GetChannelHandler(channel string) (models.ChannelHandler,
...
+// GetChannelNamespace gives threadsafe access to the channel
+func (g *GrafanaLive) GetChannelNamespace(scope string, name string) (models.ChannelNamespaceHandler, error) {
+	if scope == "grafana" {
+		p, ok := g.GrafanaScope.Features[name]
 		if ok {
-			return p.GetHandlerForPath(id.Path)
+			return p, nil
 		}
-		return nil, fmt.Errorf("Unknown feature: %s", id.Namespace)
+		return nil, fmt.Errorf("Unknown feature: %q", name)
```
aknuds1, line 271:
```suggestion
		return nil, fmt.Errorf("unknown feature: %q", name)
```

aknuds1 (changes requested):

packages/grafana-runtime/src/measurement/types.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export interface MeasurementsQuery {
+  name?: string;
+  labels?: Labels;
+  fields?: string[]; // only include the fields with these names
+}
+
+/**
+ * Channels that recieve Measurements can collect them into frames
```
aknuds1, line 60:
```suggestion
 * Channels that receive Measurements can collect them into frames
```

public/app/features/live/features.ts
```diff
@@ -39,7 +47,13 @@ export function registerLiveFeatures() {
       },
       getSupportedPaths: () => [broadcastConfig],
     },
-    description: 'Broadcast will send/recieve any events on a channel',
+    description: 'Broadcast will send/recieve any JSON object in a channel',
```
aknuds1, line 50:
```suggestion
    description: 'Broadcast will send/receive any JSON object in a channel',
```

public/app/features/live/features.ts
```diff
@@ -39,7 +47,13 @@ export function registerLiveFeatures() {
...
+  });
+
+  grafanaLiveCoreFeatures.register({
+    name: 'measurements',
+    support: new LiveMeasurmentsSupport(),
```
aknuds1, line 55:
```suggestion
    support: new LiveMeasurementsSupport(),
```

public/app/plugins/datasource/grafana/components/QueryEditor.tsx
```diff
@@ -23,18 +31,137 @@ export class QueryEditor extends PureComponent<Props> {
...
+        description: `Connected to ${channel}`,
+      };
+      channels.push(currentChannel);
+    }
+
+    if (!measurements) {
+      measurements = {};
+    }
+    const names: Array<SelectableValue<string>> = [
+      { value: '', label: 'All measurements', description: 'show every measurment streamed to this channel' },
```
aknuds1, line 69:
```suggestion
      { value: '', label: 'All measurements', description: 'show every measurement streamed to this channel' },
```

public/app/plugins/datasource/grafana/components/QueryEditor.tsx
```diff
@@ -23,18 +31,137 @@ export class QueryEditor extends PureComponent<Props> {
...
+      <>
+        <div className="gf-form">
+          <InlineField label="Channel" grow={true} labelWidth={labelWidth}>
+            <Select
+              options={channels}
+              value={currentChannel || ''}
+              onChange={this.onChannelChange}
+              allowCustomValue={true}
+              backspaceRemovesValue={true}
+              placeholder="Select measurments channel"
```
aknuds1, line 114:
```suggestion
              placeholder="Select measurements channel"
```

public/app/plugins/datasource/grafana/components/QueryEditor.tsx
```diff
@@ -23,18 +31,137 @@ export class QueryEditor extends PureComponent<Props> {
...
+            />
+          </InlineField>
+        </div>
+        {channel && (
+          <div className="gf-form">
+            <InlineField label="Measurement" grow={true} labelWidth={labelWidth}>
+              <Select
+                options={names}
+                value={names.find(v => v.value === measurements?.name) || names[0]}
+                onChange={this.onMeasurmentNameChanged}
```
aknuds1, line 127:
```suggestion
                onChange={this.onMeasurementNameChanged}
```

public/app/plugins/datasource/grafana/components/QueryEditor.tsx
```diff
@@ -23,18 +31,137 @@ export class QueryEditor extends PureComponent<Props> {
...
+                isClearable={true}
+                noOptionsMessage="Filter by name"
+                formatCreateLabel={(input: string) => `Show: ${input}`}
+                isSearchable={true}
+              />
+            </InlineField>
+          </div>
+        )}
+
+        <FeatureInfoBox title="Grafana Live - Measurments" featureState={FeatureState.alpha}>
```
aknuds1, line 140:
```suggestion
        <FeatureInfoBox title="Grafana Live - Measurements" featureState={FeatureState.alpha}>
```

public/app/plugins/datasource/grafana/components/QueryEditor.tsx
```diff
@@ -23,18 +31,137 @@ export class QueryEditor extends PureComponent<Props> {
...
+          <p>
+            This supports real-time event streams in grafana core. This feature is under heavy development. Expect the
```
aknuds1, line 142:
```suggestion
            This supports real-time event streams Ggrafana core. This feature is under heavy development. Expect the
```

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -6,54 +6,63 @@ import {
   DataQueryResponse,
   DataSourceApi,
   DataSourceInstanceSettings,
+  LiveChannelScope,
 } from '@grafana/data';
 
-import { GrafanaQuery, GrafanaAnnotationQuery, GrafanaAnnotationType } from './types';
-import { getBackendSrv, getTemplateSrv, TemplateSrv, toDataQueryResponse } from '@grafana/runtime';
+import { GrafanaQuery, GrafanaAnnotationQuery, GrafanaAnnotationType, GrafanaQueryType } from './types';
+import { getBackendSrv, getTemplateSrv, toDataQueryResponse, getLiveMeasurmentsObserver } from '@grafana/runtime';
```
aknuds1, line 13:
```suggestion
import { getBackendSrv, getTemplateSrv, toDataQueryResponse, getLiveMeasurementsObserver } from '@grafana/runtime';
```

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -6,54 +6,63 @@ import {
...
+    const queries: Array<Observable<DataQueryResponse>> = [];
+    for (const target of request.targets) {
+      if (target.hide) {
+        continue;
+      }
+      if (target.queryType === GrafanaQueryType.LiveMeasurements) {
+        const { channel, measurements } = target;
+        if (channel) {
+          queries.push(
+            getLiveMeasurmentsObserver(
```
aknuds1, line 34:
```suggestion
            getLiveMeasurementsObserver(
```

public/app/plugins/datasource/testdata/datasource.ts
```diff
@@ -17,9 +17,16 @@ import {
...
+  LiveChannelScope,
 } from '@grafana/data';
 import { Scenario, TestDataQuery } from './types';
-import { getBackendSrv, toDataQueryError, getTemplateSrv, TemplateSrv } from '@grafana/runtime';
+import {
+  getBackendSrv,
+  toDataQueryError,
+  getTemplateSrv,
+  TemplateSrv,
+  getLiveMeasurmentsObserver,
```
aknuds1, line 28:
```suggestion
  getLiveMeasurementsObserver,
```

packages/grafana-runtime/src/measurement/collector.test.ts
```diff
@@ -0,0 +1,250 @@
+import { MeasurementCollector } from './collector';
+import { MeasurmentAction } from './types';
```
aknuds1, line 2:
```suggestion
import { MeasurementAction } from './types';
```

packages/grafana-runtime/src/measurement/collector.test.ts
```diff
@@ -0,0 +1,250 @@
...
+            },
+          },
+        },
+        "name": "test",
+        "refId": undefined,
+      }
+    `);
+
+    collector.addBatch({
+      action: MeasurmentAction.Replace,
```
aknuds1, line 170:
```suggestion
      action: MeasurementAction.Replace,
```

packages/grafana-runtime/src/measurement/collector.test.ts
```diff
@@ -0,0 +1,250 @@
...
+            },
+          },
+        },
+        "name": "test",
+        "refId": undefined,
+      }
+    `);
+
+    collector.addBatch({
+      action: MeasurmentAction.Clear,
```
aknuds1, line 245:
```suggestion
      action: MeasurementAction.Clear,
```

packages/grafana-runtime/src/measurement/collector.ts
```diff
@@ -0,0 +1,203 @@
...
+  FieldType,
+  DataFrame,
+  matchAllLabels,
+  parseLabels,
+  CircularVector,
+  ArrayVector,
+} from '@grafana/data';
+import { Measurement, MeasurementBatch, LiveMeasurements, MeasurementsQuery, MeasurmentAction } from './types';
+
+interface MeasurmentCacheConfig {
```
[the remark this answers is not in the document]
aknuds1: I noticed the same.

packages/grafana-runtime/src/measurement/collector.ts
```diff
@@ -0,0 +1,203 @@
...
+  append?: 'head' | 'tail';
+  capacity?: number;
+}
+
+/** This is a cache scoped to a the measurement name */
+export class MeasurmentCache {
```
aknuds1, line 20:
```suggestion
export class MeasurementCache {
```

packages/grafana-runtime/src/measurement/collector.ts
```diff
@@ -0,0 +1,203 @@
...
+  readonly frames: Record<string, CircularDataFrame> = {}; // key is the labels
+
+  constructor(public name: string, private config: MeasurmentCacheConfig) {
```
aknuds1, line 23:
```suggestion
  constructor(public name: string, private config: MeasurementCacheConfig) {
```

aknuds1 (changes requested):
Please rewrite measurment as measurement.

aknuds1 (changes requested):

aknuds1 (changes requested):
I've committed my previous suggestions, so will try reviewing again tomorrow, now that I know more about how it works.

public/app/plugins/panel/live/module.tsx
ryantxu: dooh -- either, but not both!

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -73,7 +73,7 @@ export function concatenateFields(data: DataFrame[], opts: ConcatenateTransforme
         } else if (opts.frameNameMode === ConcatenateFrameNameMode.Label) {
           copy.labels = { ...f.labels };
           copy.labels[frameNameLabel] = frame.name;
-        } else if (!copy.name || copy.name === 'Value') {
+        } else if (!copy.name || copy.name === TIME_SERIES_VALUE_FIELD_NAME) {
```
ryantxu, line 76: kinda unrelated... but it took me a while to track down why stuff was behaving weird with "value" rather than "Value"

public/app/plugins/panel/live/LiveChannelEditor.tsx
```diff
@@ -141,7 +141,7 @@ export class LiveChannelEditor extends PureComponent<Props, State> {
               <Label>Path</Label>
               <Select
                 options={paths}
-                value={paths.find(s => s.value === path) ?? (path ? { label: path, value: path } : undefined)}
+                value={findPathOption(paths, path)}
```
hugohaggmark, line 144: Nice!

hugohaggmark:
> There is one _massive_ issue with streaming as currently implemented. We do not yet handle unsubscribing! We _could_ unsubscribe when the rxjs observable is done -- however, this is not great for lots of standard features where you navigate from one dashboard to another and expect to see the same data but displayed differently.

I'm not sure if this is the _massive_ issue you're mentioning Ryan but it seems like every TestData panel I add that uses Grafana Live starts pushing websocket messages and it never stops.
live-feature

Changing it to another scenario will not stop it either
live-feature2

ryantxu:
> I'm not sure if this is the massive issue you're mentioning Ryan 

Yes -- I would like to solve this in a follow on PR.  The logic will be a little tricky and it would get lost in this one.   On the frontend, we do not want to stop listening immediately when unsubscribed because we need to keep state on refresh and when navigating from page to page.  My plan is to add a watchdog check that will cleanup any open connections that are not listened to

pkg/models/live.go
```diff
@@ -17,9 +17,10 @@ type ChannelHandler interface {
 	OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error)
 }
 
-// ChannelHandlerProvider -- this should be implemented by any core feature
-type ChannelHandlerProvider interface {
-	// This is called fast and often -- it must be synchrnozed
+// ChannelNamespaceHandler should be implemented by all core features.
+type ChannelNamespaceHandler interface {
```
aknuds1, line 21: I suggest we rename the interface as follows, since the current name doesn't reflect what it does:
```suggestion
// ChannelHandlerFactory should be implemented by all core features.
type ChannelHandlerFactory interface {
```

aknuds1 (approved):
Looks good AFAICT, but please see suggestion about renaming an interface so it's more understandable.

aknuds1:
Thanks for applying my suggestion, fixed up the rest of the code correspondingly :)

ryantxu:
nice -- I was about to do it and got conflicts :)

aknuds1:
Fails on e2e tests.

torkelo:
> In the frontend, we do not want to stop listening immediately when unsubscribed because we need to keep state on refresh and when navigating from page to page 

Can you describe this more? Feels strange to leave things running and pushing data like this and not using RXJS unsubscribe to clean up / remove centrifuge channel subscriptions. 

Does centrifuge keep state? Why do we need another background clean-up process for centrifuge channels?

public/app/plugins/datasource/grafana/datasource.ts
```diff
@@ -6,54 +6,63 @@ import {
...
+    }
+    // With a single query just return the results
+    if (queries.length === 1) {
+      return queries[0];
+    }
+    if (queries.length > 1) {
+      // HELP!!!
+      return queries[0];
+    }
+    return of(); // nothing
```
torkelo, line 57: not sure I understand this part, and since we are returning getLiveMeasurementsObserver and it calls getStream on the stream handler, when this is unsubscribed by Grafana why is not the centrifuge channel subscription also then removed?

ryantxu:
> Why do we need another background clean-up process for centrifuge channels?

The primary use case I am hoping to solve is using grafana in the frontend to collect points that are streamed to it.  If we disconnect and reconnect every time an rxjs subscription ends/starts we will lose info.  This happens on dashboard refresh, setting a template variable, navigating pages etc.  We want to be able to connect to a stream and keep using it as we navigate around various dashboards using the same data.  In https://github.com/grafana/grafana/pull/28416, I am exploring how to best solve this --- but it looks like we are not actually unsubscribing everything :(  still need to dig

faxm0dem:
I'm confused about the streaming support implementation.
How does this PR relate to https://grafana.com/docs/grafana/latest/developers/plugins/build-a-streaming-data-source-plugin/ ?

PR #28025, opened 2020-10-04
Runtime: Fix handling of short-lived background services

### LINKED ISSUE #26364: Registry: Allow short-lived background services
**What happened**:
When starting a background service by implementing `Run(context.Context) error` and immediately returning `nil` (to avoid running the service) the service will stop other services from starting up.

It is possible to work-around this by idly waiting for `<-ctx.Done()` when Grafana is shutting down, but it'd be reasonable to allow services to shut themselves down without returning an error and not break unrelated services.

These lines are likely responsible:

https://github.com/grafana/grafana/blob/6619cc4b816e6778b68a46ba6c1404f7af845bb4/pkg/cmd/grafana-server/server.go#L129-L132

We should avoid triggering shutdown mode unless we actually get a `<-ctx.Done` signal.

The service is cleanly shut down.

Implement a BackgroundService and return `nil` as the first step of the service execution.

(cc @marefr)

ahlaw:

Moves shutdown mode flag setting so that shutdown mode is not triggered when services shut down without returning an error.

Fixes [#26364](https://github.com/grafana/grafana/issues/26364)

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -179,10 +179,10 @@ func (s *Server) Run() (err error) {
 			}
 
 			err := service.Run(s.context)
-			// Mark that we are in shutdown mode
-			// So no more services are started
-			s.shutdownInProgress = true
 			if err != nil {
+				// Mark that we are in shutdown mode
+				// So no more services are started
+				s.shutdownInProgress = true
 				if err != context.Canceled {
 					// Server has crashed.
 					s.log.Error("Stopped "+descriptor.Name, "reason", err)


pkg/server/server.go
```diff
@@ -179,10 +179,10 @@ func (s *Server) Run() (err error) {
 			}
 
 			err := service.Run(s.context)
-			// Mark that we are in shutdown mode
-			// So no more services are started
-			s.shutdownInProgress = true
 			if err != nil {
+				// Mark that we are in shutdown mode
+				// So no more services are started
+				s.shutdownInProgress = true
```
aknuds1, line 185: @sakjur Isn't there more that needs doing than just this?

torkelo:
Think this should be fine, but not 100% sure

pkg/server/server.go
sakjur: I just dived into this a little bit and it seems like this should work.
I'll check out the code and play around a little to verify, but I think this change should be mergeable.

aknuds1: @sakjur So `s.shutdownInProgress` only needs to be set if an error occurred?

sakjur (approved):
Tested this locally with both returning `nil` and an error on service initialization. Both were handled correctly (nil continued with the start of the Grafana service and the error aborted it with "A service failed").

LGTM

aknuds1 (approved):
Think I see the issue solved by this change, and it makes sense.

marefr (approved):
LGTM

aknuds1:
Thanks for the contribution!

PR #28027, opened 2020-10-05
Docs: Update Elasticsearch and InfluxDB logs info

### LINKED ISSUE #27215: Documentation Elasticsearch datasource logs explore
Hello in https://grafana.com/docs/grafana/latest/features/datasources/elasticsearch/#querying-logs-beta
it said that we can switch between logs and metrics with the button but this button no longer exist.
https://grafana.slack.com/archives/C05675Y4F/p1596658860428300?thread_ts=1596646902.426300&cid=C05675Y4F

ahlaw:

Adds a note to the Elasticsearch and InfluxDB data source documentation informing users of Grafana 7.1 [update](https://grafana.com/docs/grafana/latest/guides/whats-new-in-v7-1/#explore-modes-unified) where query modes were unified and the metrics/logs switcher was removed.

Fixes [#27215](https://github.com/grafana/grafana/issues/27215)

### DIFF
--- a/docs/sources/features/datasources/elasticsearch.md
+++ b/docs/sources/features/datasources/elasticsearch.md
@@ -197,13 +197,8 @@ for annotation events.
 
 > Only available in Grafana v6.3+.
 
-Querying and displaying log data from Elasticsearch is available via [Explore]({{< relref "../../explore" >}}).
-
-![](/img/docs/v63/elasticsearch_explore_logs.png)
-
-Select the Elasticsearch data source, change to Logs using the Metrics/Logs switcher, and then optionally enter a lucene query into the query field to filter the log messages.
-
-Finally, press the `Enter` key or the `Run Query` button to display your logs.
+Querying and displaying log data from Elasticsearch is available in [Explore]({{< relref "../../explore" >}}), and in the [logs panel]({{< relref "../../panels/visualizations/logs-panel.md" >}}) in dashboards. 
+Select the Elasticsearch data source, and then optionally enter a lucene query to display your logs.
 
 ### Log Queries
 
--- a/docs/sources/features/datasources/influxdb.md
+++ b/docs/sources/features/datasources/influxdb.md
@@ -180,12 +180,8 @@ You can view the interpolated version of a query with the Query Inspector.
 
 > Only available in Grafana v6.3+.
 
-Querying and displaying log data from InfluxDB is available via [Explore]({{< relref "../../explore" >}}).
-
-![](/img/docs/v63/influxdb_explore_logs.png)
-
-Select the InfluxDB data source, change to Logs using the Metrics/Logs switcher,
-and then use the `Measurements/Fields` button to display your logs.
+Querying and displaying log data from InfluxDB is available in [Explore]({{< relref "../../explore" >}}), and in the [logs panel]({{< relref "../../panels/visualizations/logs-panel.md" >}}) in dashboards. 
+Select the InfluxDB data source, and then enter a query to display your logs.
 
 ### Log Queries
 


docs/sources/features/datasources/elasticsearch.md
```diff
@@ -203,6 +203,8 @@ Querying and displaying log data from Elasticsearch is available via [Explore]({
 
 Select the Elasticsearch data source, change to Logs using the Metrics/Logs switcher, and then optionally enter a lucene query into the query field to filter the log messages.
```
zoltanbedi, line 204: I've imagined that we should rewrite this sentence and replace the images that show the switcher but maybe @oddlittlebird have some input about this?

achatterjee-grafana, line 204: It is a very long sentence. Change to:

Select the Elasticsearch data source then change to Logs using the Metrics/Logs switcher. Optionally enter a lucene query in the query field to filter log messages.

docs/sources/features/datasources/influxdb.md
```diff
@@ -187,6 +187,8 @@ Querying and displaying log data from InfluxDB is available via [Explore]({{< re
 Select the InfluxDB data source, change to Logs using the Metrics/Logs switcher,
 and then use the `Measurements/Fields` button to display your logs.
 
+> **Note:** As of Grafana 7.1, the Metrics/Logs switcher has been removed. Instead, Explore will choose the correct visualization method given the query response.
```
achatterjee-grafana, line 190: Remove "Instead,"

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -203,6 +203,8 @@ Querying and displaying log data from Elasticsearch is available via [Explore]({
...
 
+> **Note:** As of Grafana 7.1, the Metrics/Logs switcher has been removed. Instead, Explore will choose the correct visualization method given the query response.
```
achatterjee-grafana, line 206: Remove "Instead,"

docs/sources/features/datasources/elasticsearch.md
ahlaw: In that case, maybe we use the same format as in the [Loki](https://grafana.com/docs/grafana/latest/features/datasources/loki/#querying-logs) section (image removed)?

zoltanbedi: yeah that sounds good to me. Thank you for helping with this.

aocenas:
@oddlittlebird I am wondering if having different notes with version for some functionality makes sense. We already have docs versioned and you can switch between versions so it feels to me like each docs version should describe current version and does not have to explain how it worked in old versions.

We have this already sometimes when we say this feature is available since X.X version, but if we say this feature works like this for version 7.0 different for 7.2, like this for 7.4 etc it may not help the docs to be more readable or understandable.

oddlittlebird:
Yeah, I don't love it, but right now there are two reasons for it:

- People lag in changing Grafana versions, sometimes a lot. 
- The notes are often added for functionality that is future state or in beta, and then the note is never removed.

At some point, we'll go through and make a coherent policy on that, but for now it is littered throughout and that's just how it is.

ahlaw:
I don't believe log querying is beta anymore for InfluxDB or Elasticsearch? Should we remove that and the subsequent version note for both in this PR as well?

zoltanbedi:
> I don't believe log querying is beta anymore for InfluxDB or Elasticsearch? Should we remove that and the subsequent version note for both in this PR as well?

No, let's just leave it as it is right now. We will take care of those later.

zoltanbedi (approved):
This looks good to me. Thanks for helping with this. Let's have one more approval from @achatterjee-grafana or @oddlittlebird and we can merge this.

docs/sources/features/datasources/elasticsearch.md
```diff
@@ -197,13 +197,8 @@ for annotation events.
...
 > Only available in Grafana v6.3+.
 
-Querying and displaying log data from Elasticsearch is available via [Explore]({{< relref "../../explore" >}}).
-
-![](/img/docs/v63/elasticsearch_explore_logs.png)
-
-Select the Elasticsearch data source, change to Logs using the Metrics/Logs switcher, and then optionally enter a lucene query into the query field to filter the log messages.
-
-Finally, press the `Enter` key or the `Run Query` button to display your logs.
+Querying and displaying log data from Elasticsearch is available via [Explore]({{< relref "../../explore" >}}), and with the [logs panel]({{< relref "../../panels/visualizations/logs-panel.md" >}}) in dashboards.
```
achatterjee-grafana, line 200: Suggestion: replace "via" with "in"

"and in the [logs panel] ..."

docs/sources/features/datasources/influxdb.md
```diff
@@ -180,12 +180,8 @@ You can view the interpolated version of a query with the Query Inspector.
 
 > Only available in Grafana v6.3+.
 
-Querying and displaying log data from InfluxDB is available via [Explore]({{< relref "../../explore" >}}).
-
-![](/img/docs/v63/influxdb_explore_logs.png)
-
-Select the InfluxDB data source, change to Logs using the Metrics/Logs switcher,
-and then use the `Measurements/Fields` button to display your logs.
+Querying and displaying log data from InfluxDB is available via [Explore]({{< relref "../../explore" >}}), and with the [logs panel]({{< relref "../../panels/visualizations/logs-panel.md" >}}) in dashboards.
```
achatterjee-grafana, line 183: Same suggestion as above.

achatterjee-grafana (approved):
Approving. I did make a couple of suggestions for you. Thank you.

docs/sources/features/datasources/elasticsearch.md
ahlaw: I agree with replacing "via" with "in", but not sure about the logs panel because querying is done in the query editor which is not _in_ the logs panel. Rather, I imagine the semantics being that "displaying log data is available in dashboards _with_ the [use of] logs panel".

achatterjee-grafana: Thank you for the explanation. Can you rephrase using either "through" or "from" whichever you think is a better fit? With is not the preposition to use here.

ahlaw: On second thought, I think "in" would be the most appropriate here since it makes it clear that the logs panel is _where_ the functionality can be found!

achatterjee-grafana: Thank you for taking the time to think over and making the documentation "just right" :)

PR #28028, opened 2020-10-05
Live: cleanup and simple changes

ryantxu:
Extracting some cleanup code from #28022 it will be easier to review.  The changes here are:

1. Rather than calling `getChannel(scope, namespace, path)`, uses an object `getChannel({scope, namespace, path});
2. measurement typos
3. labels matching utilities


hugohaggmark (approved):
Looks good!

PR #28033, opened 2020-10-05
Elasticsearch: Migrate queryeditor to React

Elfo404:
MIgrates the Elasticsearch data source query editor to React

I wasn't able to keep the number of commits low to make the review easier, sorry for that.

I limited the changes to the query editor focusing on type-safety, touching the query builder, response parser, and the datasource to fix some type inconsistencies or make types a bit stricter, but they will probably need some more love in a follow-up PR.

I feel there are still some missing needed tests, I'll add them here or in a follow-up PR.

To better manage query state I'm using a reducer-like pattern where the reducer doesn't mutate the state but emits an `onChange` event after every dispatched action. It adds a bit of boilerplate code but I feel like it makes working with complex  "state" easier.


ivanahuckova (commented):
As mentioned on daily, I agree that this makes sense and it is going to make working with state much easier. 👍

public/test/core/redux/reducerTester.ts
```diff
@@ -66,9 +66,9 @@ export const reducerTester = <State>(): Given<State> => {
     disableDeepFreeze = false
   ): When<State> => {
     reducerUnderTest = reducer;
-    initialState = { ...state };
+    initialState = state;
     if (!disableDeepFreeze) {
-      initialState = deepFreeze(initialState);
+      deepFreeze(initialState);
```
hugohaggmark, line 71: Great catch ⭐

public/test/core/redux/reducerTester.ts
```diff
@@ -66,9 +66,9 @@ export const reducerTester = <State>(): Given<State> => {
     disableDeepFreeze = false
   ): When<State> => {
     reducerUnderTest = reducer;
-    initialState = { ...state };
+    initialState = state;
```
hugohaggmark, line 69: I think we need to be a bit more clever here. If this is an object, I think we should use cloneDeep from lodash to clone the state.

public/test/core/redux/reducerTester.ts
hugohaggmark: I think we need to be a bit more clever here too. Only deepFreeze if this is an object.

public/test/core/redux/reducerTester.ts
hugohaggmark: I used the spread syntax before but I realise that using cloneDeep would be better, this means that when I use reducerTester I don't need to worry about spreading the state as a consumer. WDYT?

Elfo404: my first take on this was to deep clone the state object (using `JSON.parse(JSON.stringify())` for simplicity, but the same apply to cloneDeep) but it's not clear to me why we want it.

as far as I understood, `initialState` is frozen to "enforce" immutability, so if you don't opt-out from this you should be fine with the freeze leaking over the original object, right?

> this means that when I use reducerTester I don't need to worry about spreading the state as a consumer

Not sure why it would be needed, can you explain please?

public/test/core/redux/reducerTester.ts
Elfo404: i think it make sense to freeze array as well as this is valid:
```ts
const obj = [1, 2, 3];
obj.somethingElse = 4;
```

what do you think?

public/test/core/redux/reducerTester.ts
hugohaggmark: I'll try to explain this, you could end up in false positive tests if you use the initialState passed to reducerTester in your expects. This is not entirely a freeze issue but a way to prevent false positive tests in certain circumstances. But this is just me being super cautious you need to do pretty bad things to get false positives and as we're using redux toolkit this might never happen. 

```ts
      const dummyAction = createAction<string>('dummyAction');

      const reducer = (state: any, action: AnyAction): any => {
        state.newProp = 'A'; // an example where we add a new prop to state
        return state;
      };

      const initialState = { prop: 'A' };
      reducerTester()
        .givenReducer(reducer, initialState, true, true) // we disable freeze
        .whenActionIsDispatched(dummyAction('some string'))
        .thenStateShouldEqual(initialState); // because initialState is the exact same instance as the reducer state this will not fail but it should as we've added a new prop
```

public/test/core/redux/reducerTester.ts
hugohaggmark: Arrays are object so yes ;)

public/test/core/redux/reducerTester.ts
Elfo404: ooh, didn't think about that situation. Thanks!

torkelo:
Took this for a test, awesome job!  

Some of the UX feels a bit worse, especially in terms of alignment of form elements.

Screenshot from 2020-11-27 11-11-01

Old editor
Screenshot from 2020-11-27 11-05-57

and I feel the eye icon/action works better as icon button part of the Metric label rather than a label element somewhere at the end (next to plus/minus). Potentially both eye & delete would work better maybe there? Example:

Screenshot from 2020-11-27 11-18-20

torkelo:
Another minor UX issue: 
* spacing between chevron down and the text to the right is too small (compare to current editor)
* The new form switch is not designed for inline forms, we need to create an inline form wrapper (not sure if it exists) so that it looks like the old inline form switch (propper spacing and background and border like an input)
Screenshot from 2020-11-27 11-20-33

Elfo404:
@torkelo Thanks for taking this for a ride!

> and I feel the eye icon/action works better as icon button part of the Metric label rather than a label element somewhere at the end (next to plus/minus). Potentially both eye & delete would work better maybe there? Example:

Putting the 'eye' icon there was also what I thought initially, but I got feedback that replicating what we already have in explore (for any datasource) might have made more sense:
image

But I agree on the alignment issue, I'll take a look at it.

Elfo404:
@torkelo also, regarding
>and does not look perfectly vertically centered (too high)

on which browser are you experiencing this? On Chrome 86 & Firefox 84 (after fixing the horizontal spacing) it seems fine to me:
chrome:
Screenshot 2020-12-02 at 10 14 55

FF:
Screenshot 2020-12-02 at 10 15 24

torkelo:
> Putting the 'eye' icon there was also what I thought initially, but I got feedback that replicating what we already have in explore (for any datasource) might have made more sense:

I do not agree with that. that is for the full query and is part of the query row: 

Screenshot from 2020-12-02 12-10-22

I think explore will start using this query row header soon as well. 

I feel the remove and eye elements get lost this way 
Screenshot from 2020-11-27 11-11-01

This way the eye & trash are aligned on all metric rows and more easily found  
Screenshot from 2020-11-27 11-05-57

Elfo404:
@torkelo applied your suggestions and it looks way better :)

- add (plus) button is kept at the end of the first metric / bucket aggregation row
- the remove button is always visible (but disabled when needed) to avoid some jumps when adding / removing metrics and bucket aggregations

Kapture 2020-12-02 at 14 40 51

torkelo:
@Elfo404 awesome!  now just make the same width on the first & second segment and it should be :1st_place_medal:

Elfo404:
@torkelo Way better.
FYI, segments do not have a `width` or `minWidth` prop, but I don't think is needed since it's pretty straightforward to style it via css.

Screenshot 2020-12-02 at 16 05 24

torkelo:
@Elfo404 it has a labelWidth , but missing a minLabelWidth

torkelo:
but could also use a utility css class name, but we should have standards grid based minWidth & width css classes (emotion based), that we can reuse

torkelo:
we have width-n and min-width-x , defined https://github.com/grafana/grafana/blob/master/public/sass/utils/_widths.scss

Elfo404:
@torkelo i think labelWidth (or minLabelWidth) applies to `InlineField`s and not to Segments, I'm already using it in each metric settings, to keep the sizing consistent.

Screenshot 2020-12-02 at 16 19 13

In the "main row" there's no label associated with the Segment; Am I missing something? 🤔

torkelo:
as sorry you are rigtht, then you can use css class names

public/app/plugins/datasource/elasticsearch/components/QueryEditor/QueryEditorRow.tsx
```diff
@@ -0,0 +1,91 @@
...
+      className={css`
+        display: block;
+        margin-bottom: ${theme.spacing.xs};
+      `}
+    >
+      <div className={cx(flex)}>
+        <div className={cx(getInlineLabelStyles(theme, 17).label)}>
+          <legend
+            className={css`
+              font-size: ${theme.typography.size.sm};
```
torkelo, line 43: we have some inline styles in this function that can be moved to the getStyles function. (I added it now for the IconButton styles)

torkelo (approved):
Really great job on this one, a lot of code but super clean, well structured & unit tests. 

A bit big but PRs get that sometimes :)

Elfo404:
Just rebased on top of master because of a typescript error appeared when upgrading to 4.1.2

aocenas (commented):
Yeah agree the code and structuring is super clean, good job on that. Going to do some manual testing.

public/app/plugins/datasource/elasticsearch/components/QueryEditor/BucketAggregationsEditor/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export const BucketAggregationsEditor: FunctionComponent<Props> = ({ nextId }) => {
+  const dispatch = useDispatch<BucketAggregationAction>();
+  const { bucketAggs } = useQuery();
+  const totalBucketAggs = bucketAggs?.length || 0;
+
+  return (
+    <>
+      {bucketAggs!.map((bucketAgg, index) => (
```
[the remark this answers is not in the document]
Elfo404: That's very true and I also thought about it, the reason I didn't focus on avoiding rerenderings due to `useQuery` is that (almost?) every time we use it we also use `useDispatch`, which uses ultimately depends on the `onChange` callback that is passed to the Query Editor.
The problem is that `onChange` also changes at every query change, which would basically nullify every optimization made by avoiding using `useQuery`.

> If so having it as a non optional prop would make it more obvious and you would not be able to use it without the bucketAggs

but I agree on this, it would probably be much cleaner. I'll give it a second pass.

public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/useDescription.ts
```diff
@@ -0,0 +1,40 @@
+import { extendedStats } from '../../../../query_def';
+import { MetricAggregation } from '../aggregations';
+
+const hasValue = (value: string) => (object: { value: string }) => object.value === value;
+
+// FIXME: All the defaults and validations down here should be defined somewhere else
+// as they are also the defaults that are gonna be applied to the query.
+// In the previous version, the same methos was taking care of describing the settings and setting defaults.
```
zoltanbedi, line 8:
```suggestion
// In the previous version, the same method was taking care of describing the settings and setting defaults.
```

public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  );
+};
+
+interface ExtendedStatSettingProps {
+  stat: ExtendedStat;
+  onChange: (checked: boolean) => void;
+  value: boolean;
+}
+const ExtendedStatSetting: FunctionComponent<ExtendedStatSettingProps> = ({ stat, onChange, value }) => {
+  // this is needed for the htmlFor prop in the label so that clicking the lable will toggle the switch state.
```
zoltanbedi, line 121:
```suggestion
  // this is needed for the htmlFor prop in the label so that clicking the label will toggle the switch state.
```

public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/MovingAverageSettingsEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { useDispatch } from '../../../../hooks/useStatelessReducer';
+import { movingAvgModelOptions } from '../../../../query_def';
+import { isEWMAMovingAverage, isHoltMovingAverage, isHoltWintersMovingAverage, MovingAverage } from '../aggregations';
+import { changeMetricSetting } from '../state/actions';
+
+interface Props {
+  metric: MovingAverage;
+}
+
+// The way we handle changes for those settings is not ideal compaared to the other components in the editor
```
zoltanbedi, line 12:
```suggestion
// The way we handle changes for those settings is not ideal compared to the other components in the editor
```

public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/MetricEditor.tsx
```diff
@@ -0,0 +1,120 @@
...
+// it doesn't make sense to show it in the type picker when there is no non-pipeline-aggregation previously selected
+// as they work on the outputs produced from other aggregations rather than from documents or fields.
+// This means we should filter them out from the type picker if there's no other "basic" aggregation before the current one.
+const isBasicAggregation = (metric: MetricAggregation) => !metricAggregationConfig[metric.type].isPipelineAgg;
+
+const getTypeOptions = (
+  previousMetrics: MetricAggregation[],
+  esVersion: number
+): Array<SelectableValue<MetricAggregationType>> => {
+  // we'll include Pipeline Aggeregations only if at least one previous metric is a "Basic" one
```
zoltanbedi, line 47:
```suggestion
  // we'll include Pipeline Aggregations only if at least one previous metric is a "Basic" one
```

public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/MetricEditor.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const includePipelineAggregations = previousMetrics.some(isBasicAggregation);
+
+  return (
+    Object.entries(metricAggregationConfig)
+      // Only showing metrics type supported by the configured version of ES
+      .filter(([_, { minVersion = 0, maxVersion = esVersion }]) => {
+        // TODO: Double check this
+        return esVersion >= minVersion && esVersion <= maxVersion;
+      })
+      // Filtering out Pipeline Aggragations if there's no basic metric selected before
```
zoltanbedi, line 57:
```suggestion
      // Filtering out Pipeline Aggregations if there's no basic metric selected before
```

aocenas (approved):
Looks good, there are some weird things but all looks like something that was already before the refactor and not a regression so this seems to be ready to merge.

public/app/plugins/datasource/elasticsearch/components/QueryEditor/BucketAggregationsEditor/index.tsx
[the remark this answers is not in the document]
aocenas: As I said no reason to do anything about it now in this PR, as you say there are probably other bottlenecks anyway and it is not a big issue.

public/app/plugins/datasource/elasticsearch/components/AddRemove.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface Props {
+  index: number;
+  elements: any[];
+  onAdd: () => void;
+  onRemove: () => void;
+}
+
+/**
+ * A component used to show add & remove buttons for mutable lists ov values. Wether to show or not the add or the remove buttons
```
zoltanbedi, line 13:
```suggestion
 * A component used to show add & remove buttons for mutable lists of values. Wether to show or not the add or the remove buttons
```

zoltanbedi (approved):
Very good job Gio! Huge refactor. Only found some typos other than that all good. 🥇

hugohaggmark:
This introduces a parse bug because we forgot to use `parsedQuery` in the calls to `getFields` and `getTerms` as described in #30002:
https://github.com/grafana/grafana/blob/375e8e4fd051a5c821c2b477634c73166d0694ab/public/app/plugins/datasource/elasticsearch/datasource.ts#L710
https://github.com/grafana/grafana/blob/375e8e4fd051a5c821c2b477634c73166d0694ab/public/app/plugins/datasource/elasticsearch/datasource.ts#L716

PR #28043, opened 2020-10-05
Alerting: Return proper status code when trying to create alert notification channel with duplicate name or uid

### LINKED ISSUE #20089: When creating notification channel with non-unique uid or name should not return HTTP status 500
**What happened**:
We have a service that creates programmatic dashboards and alerts in Grafana. One step is to check if a notification channel with a certain uid (same as the email address) exists. If it does not exist, the service will try to create it. 
A handful of times, we have run into the service returning a 500 despite the unique uid and the notification channel not existing if you try to look it up with that uid. 
If you then check for a notification channel with the same name (different uid) and delete that, then try making the channel again with the email address as the uid it will work. Then the service we have can find the notification channel as expected.

If a notification channel has the same name but a unique UID, the api should be able to create the new notification channel. 

Have an email notification channel that exists with the name you're going to use.

First check if a notification channel exists with the uid you want:
curl -H "Authorization: Bearer" http://grafana/api/alert-notifications/uid/team+alerts@company.com
{"message":"Alert notification not found"}%

Run an API call to create a new notification channel with the same name but a unique uid (uid is the email address in this case):
curl -H "Content-Type: application/json" -H "Authorization: Bearer" -d @alert.json -X POST http://grafana/api/alert-notifications
{"message":"Failed to create alert notification"}%

**Environment**:
- Grafana version: v6.3.5
- Data source type & version:
- OS Grafana is installed on: linux, ubuntu
- User OS & Browser: mac & chrome
- Grafana plugins:
- Others:

jgulick48:
…tification where the name or UID already exists.

When trying to create an Alert Notification where the uid or name are already in use grafana currently returns a 500 status code.  This adds two new errors that are returned from the sqlstore and the api will now return a 409 (conflict) status code. 

Fixes #20089


pkg/api/alerting.go
```diff
@@ -273,6 +273,9 @@ func CreateAlertNotification(c *models.ReqContext, cmd models.CreateAlertNotific
 	cmd.OrgId = c.OrgId
 
 	if err := bus.Dispatch(&cmd); err != nil {
+		if err == models.ErrAlertNotificationWithSameNameExists || err == models.ErrAlertNotificationWithSameUIDExists {
```
aknuds1, line 276: It's better to use `errors.Is` than comparing errors directly, since otherwise you'll miss wrapped errors:
```suggestion
		if errors.Is(err, models.ErrAlertNotificationWithSameNameExists) || errors.Is(err, models.ErrAlertNotificationWithSameUIDExists) {
```

aknuds1 (changes requested):

aknuds1 (approved):
LGTM, but looks as if you need to fix a compilation error.

marefr:
I think errors package is missing an import in api/alerting.go. @aknuds1 maybe you could help out here getting this merged?

aknuds1:
I've fixed the build, do you mind if I push my commit to your branch @jgulick48?

jgulick48:
> I've fixed the build, do you mind if I push my commit to your branch @jgulick48?

Go ahead. Sorry I haven't had time to circle back on this one.

aknuds1:
> > I've fixed the build, do you mind if I push my commit to your branch @jgulick48?
> 
> Go ahead. Sorry I haven't had time to circle back on this one.

Done.

aknuds1:
Thanks for contributing!

PR #27972, opened 2020-10-01
Graph NG: annotations display

dprokop:
This introduces AnnotationsPlugin (no support for region annotations yet!) to validate the idea of rendering series agnostic data on the visualization. There's been a couple of bumps on the road to this state, but this should be a good reference for rendering exemplars later on.

Ref https://github.com/grafana/grafana/issues/27848

PR #28050, opened 2020-10-06
Graph NG: Use annotations data observable

dprokop:
Fixes failing master after merging #27972 without syncing with master first


aknuds1 (approved):

hugohaggmark (approved):
LGTM! The only thing I don't really like about useObservable is the fact that you can't use dependencies.

PR #28057, opened 2020-10-06
Prometheus: Add support for Exemplars

zoltanbedi:

This PR adds support for exemplars in line graph.

You can use https://github.com/grafana/tns/tree/callum-exemplars-env to try it out with. ~Don't forget to enable alpha plugins to see it work with the new graph panel (uPlot)# only needed for dashboards~ No need for this as the panel was renamed to Time series and is in beta now

## Setting up tns demo
1. `git clone https://github.com/grafana/tns.git`
1. `cd tns`
1. `git checkout callum-exemplars-env`
1. `cd production`
1. `docker-compose up`

It might fail because it has grafana in it as well. You don't need that, just use the local one.

## Setup to see exemplars

1. Set up tns demo as mentioned above
1. Add a new Jaeger data source with an URL pointing to http://localhost:8004
1. Add a new Prometheus data source with an URL pointing to http://localhost:8003 and Exemplars setting like this
Screenshot 2021-01-12 at 11 32 56
1. Run this query `histogram_quantile(0.95, sum(rate(tns_request_duration_seconds_bucket[5m])) by (le))` in Explore with Exemplars enabled in the query editor

## Todo

- [x] Move and write tests for sampling logic in result transformer
- [x] Graph legend toggler (https://github.com/grafana/grafana/pull/29575)
- [x] Styling issues
  - [x] Legend in explore should stretch (shouldn't show a scrollbar) ~TBD~ We decided to not have this anymore
  - [x] Graph width should be equal to panel width. It should fill the panel.
  - [x] Values show outside of the container
- [x] Prometheus setting for go to traces should work like derived field in Loki
- [x] Fix tests goes oom for some reason
- [x] Show context menu (https://github.com/grafana/grafana/pull/29745)
- [x] ~Add second y axis for exemplars?~ won't do this
- [x] Pass exemplar as DataLink?
- [x] ~Rewrite git history with proper authors to fix cla issue?~ I say merge this with admin rights as it will be squashed anyway


public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -57,7 +64,23 @@ export function transform(
...
   const prometheusResult = response.data.data;
 
-  if (!prometheusResult.result) {
+  if (isExemplarData(prometheusResult)) {
+    const events: AnnotationEvent[] = [];
+    prometheusResult.forEach(exemplarData => {
+      const data = exemplarData.exemplars.map(exemplar => {
+        return {
+          time: exemplar.timestamp,
+          text: getText(exemplar, transformOptions.exemplarTraceIDDestination),
```
ryantxu, line 73: do exemplars have a Y value?  It is actually helpful to use the `AnnotationEvent` interface here?

zoltanbedi: > do exemplars have a Y value?
They have a timestamp and a value so I assume that's what you mean by y value.
> It is actually helpful to use the AnnotationEvent interface here?
What do you mean? It helps with intellisense and type checking for sure.,

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+          {`Showing only ${MAX_NUMBER_OF_TIME_SERIES} time series. `}
+          <span
+            className={cx([style.showAllTimeSeries])}
+            onClick={() => setShowAllTimeSeries(true)}
+          >{`Show all ${data.series.length}`}</span>
+        </div>
+      )}
+
+      <Collapse label="Graph" loading={data.state === LoadingState.Loading} isOpen>
+        <GraphPanel
```
torkelo, line 68: ah, nice you opted to reuse the dashboard panel, was worried this would duplicate a lof the code from that 

CC @dprokop

public/app/plugins/datasource/prometheus/result_transformer.tsx
```diff
@@ -127,21 +133,38 @@ function getText(
...
+        <tr><td>Value</td><td>${exemplar.value}</td></tr>
+      </table>
     </div>`;
-  if (exemplarTraceIDDestination) {
-    traceID = exemplar.labels[exemplarTraceIDDestination.name];
-    const href = exemplarTraceIDDestination.url.replace('${value}', traceID);
-    const anchorElement = `<a href="${href}" rel="noopener" target="_blank">Go to ${traceID}</a>`;
-    return template + anchorElement;
-  }
+
```
torkelo: This feels wonky, I thought this exemplar data would be part of the DataFrame a the trace id link be done via data link. So all this table rendering could be done by Grafana and shared between all data sources. But understand that this is just for a demo

torkelo (commented):
left a few comments

public/app/features/explore/ExploreGraphNGPanel.tsx
dprokop: Yup, we discussed that and the goal is to replace this with Graph component ltr on rather than graph panel, as this one's API is too verbose

public/app/plugins/datasource/prometheus/result_transformer.tsx
zoltanbedi: Yep I think that is the future for this that I only need to pass the DataFrame with datalink and I don't have to pass an html like this.

public/app/features/explore/Explore.tsx
```diff
@@ -364,19 +363,13 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
                       )}
                       {!showStartPage && (
                         <>
-                          {showMetrics && (
-                            <ExploreGraphPanel
-                              ariaLabel={selectors.pages.Explore.General.graph}
-                              series={graphResult}
+                          {showMetrics && graphResult && (
```
aocenas, line 366: Shouldn't showMetrics be true only in case graphResult is true?

public/app/features/explore/state/reducers.ts
```diff
@@ -512,7 +512,7 @@ export const processQueryResponse = (
     loading: loadingState === LoadingState.Loading || loadingState === LoadingState.Streaming,
     update: makeInitialUpdateState(),
     showLogs: !!logsResult,
-    showMetrics: !!graphResult,
+    showMetrics: Boolean(graphResult?.length),
```
aocenas, line 515: Not sure but I think this may prevent showing a `No data` message? I mean if we get graphResult with zero length we should probably show an empty graph or something. Similar for logs if we get logsResult with zero length maybe the logs panel should show some `No data` state. At the same time not sure how that is implemented now and maybe we have actually lost that info when unifying the interface.

public/app/plugins/datasource/prometheus/result_transformer.tsx
```diff
@@ -94,6 +128,46 @@ export function transform(
...
+  let traceIDComponent: any;
+  if (exemplarTraceIDDestination) {
+    const traceID = exemplar.labels[exemplarTraceIDDestination.name];
+    const tag = ReactDOMServer.renderToString(<Tag name={traceID} colorIndex={6} />);
+    const href = exemplarTraceIDDestination.url.replace('${value}', traceID);
+    const anchorElement = `<a href="${href}" rel="noopener" target="_blank">${tag}</a>`;
+    traceIDComponent = anchorElement;
+  }
+
+  const template = `
```
aocenas, line 145: Can't this be a react element?

zoltanbedi: It can but in the end it will be sent as a string right now.

aocenas: Why as a string? Does the new Graph API accept only strings for this?

zoltanbedi: it does https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph3/plugins/ExemplarMarker.tsx#L54

aocenas: @dprokop is that to allow non react code to use this? Although I would assume that even in Angular code you could pass on a react element or not?

public/app/features/explore/ExploreGraphNGPanel.tsx
dprokop: GraphNG (exposed from grafana/ui) component merged to master https://github.com/grafana/grafana/pull/28290

public/app/plugins/datasource/prometheus/result_transformer.tsx
dprokop: It's to allow html as as far as I see the annotation can contain html. 

@zoltanbedi ExemplarMarker is a copy of an annotation marker. Don't get too attached to it and feel free to adjust the markup/code to your needs. I agree with @aocenas that this should probably be a react element rather than hardcoded markup.

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,93 @@
+import { AbsoluteTimeRange, dateTime, GrafanaTheme, LoadingState, PanelData, TimeZone } from '@grafana/data';
+import { getTemplateSrv } from '@grafana/runtime';
+import { Collapse, Icon, useTheme } from '@grafana/ui';
+import { GraphPanel } from 'app/plugins/panel/graph3/GraphPanel';
+import { css, cx } from 'emotion';
+import React, { useState } from 'react';
+
+const MAX_NUMBER_OF_TIME_SERIES = 20;
+
+const getStyles = (theme: GrafanaTheme) => ({
```
dprokop, line 10: We tend to put `getStyles` at the bottom of the file

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,93 @@
...
+  width: number;
+  absoluteRange: AbsoluteTimeRange;
+  timeZone: TimeZone;
+  onUpdateTimeRange: (absoluteRange: AbsoluteTimeRange) => void;
+}
+
+export function ExploreGraphNGPanel({ width, data, timeZone, absoluteRange, onUpdateTimeRange }: Props) {
+  const [showAllTimeSeries, setShowAllTimeSeries] = useState(false);
+  const theme = useTheme();
+  const style = getStyles(theme);
```
dprokop, line 43: You can replace this with [useStyles builtin hook](https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md#styling-with-theme)

public/app/plugins/datasource/prometheus/types.ts
```diff
@@ -93,6 +117,13 @@ export function isMatrixData(result: MatrixOrVectorResult): result is PromMatrix
   return 'values' in result;
 }
 
+export function isExemplarData(result: PromData): result is PromExemplarData[] {
+  if (result == null || !Array.isArray(result)) {
+    return false;
+  }
+  return 'exemplars' in (result as any)[0];
```
dprokop, line 129: TIL `in` operator 👍

public/app/features/explore/Explore.tsx
zoltanbedi: this check is only to let TypeScript happy.

public/app/features/explore/state/reducers.ts
zoltanbedi: Yes you are right. Nice catch.

public/app/plugins/datasource/prometheus/result_transformer.tsx
zoltanbedi: Change text on ExemplarMarker to component.

public/app/plugins/datasource/prometheus/result_transformer.ts
```diff
@@ -57,7 +65,73 @@ export function transform(
...
   const prometheusResult = response.data.data;
 
-  if (!prometheusResult.result) {
+  if (isExemplarData(prometheusResult)) {
+    const events: ExemplarsDataFrameViewDTO[] = [];
+    prometheusResult.forEach(exemplarData => {
+      const data = exemplarData.exemplars.map(exemplar => {
+        return {
+          time: exemplar.scrapeTimestamp,
+          component: getExemplarInfoComponent(
```
aocenas, line 74: @dprokop @zoltanbedi Was looking at a way to allow internal links for this and wanted to ask if passing a component will be allowed for all plugins? If so we will have to have some interface that this should conform to and pass some additional props to it. Mainly a function that will turn a dataLink into clickable link with proper href and onClick handler in case it's an internal link.

public/app/plugins/panel/graph3/plugins/ExemplarsPlugin.tsx
```diff
@@ -17,19 +9,16 @@ interface ExemplarsPluginProps {
 }
 
 // Type representing exemplars data frame fields
-interface ExemplarsDataFrameViewDTO {
+export interface ExemplarsDataFrameViewDTO {
```
aocenas, line 12: Why has this type DataFrame in a name? Does not seem like we turn it into a DataFrame actually, do we? And even if we do I don't think it needs to have DataFrame in a name because of it.

dprokop: It's not DataFrame that's referenced here but [`DataFrameView` ](https://github.com/grafana/grafana/blob/c0762b6ddc60dc2c251cd5c1428a031448c357fe/packages/grafana-data/src/dataframe/DataFrameView.ts#L16) which accepts a generic that represents the shape of the "viewed" DataFrame

aocenas: But you need a DataFrame to create DataFrameView don't you? I mean this is sort of nitpicky but I thought this to be a naming bug that was forgotten or something. Like I would expect this to be something like `ExemplarsData` and whether it is later turned into DataFrame or DataFrameView in some way to be sort of irrelevant.

public/app/plugins/datasource/prometheus/components/PromExemplarField.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { PromQuery } from '../types';
+
+interface Props {
+  query: PromQuery;
+  onChange: (value: PromQuery) => void;
+}
+
+const onExemplarsChange = ({ query, onChange }: Props) => (e: React.ChangeEvent<HTMLInputElement>) => {
+  const exemplar = e.target.checked;
+  onChange({ ...query, exemplar: exemplar });
```
ivanahuckova, line 13:
```suggestion
  const exemplar = e.target.checked;
  onChange({ ...query, exemplar });
```

or 

```suggestion
  onChange({ ...query, exemplar: e.target.checked });
```

Elfo404 (commented):
Left some comment for now, still testing this properly :)

public/app/plugins/datasource/prometheus/components/PromExploreExtraField.tsx
```diff
@@ -4,17 +4,21 @@ import { css, cx } from 'emotion';
...
+import { PromExemplarField } from './PromExemplarField';
 
 export interface PromExploreExtraFieldProps {
   queryType: string;
   stepValue: string;
+  query: PromQuery;
   onStepChange: (e: React.SyntheticEvent<HTMLInputElement>) => void;
   onKeyDownFunc: (e: React.KeyboardEvent<HTMLInputElement>) => void;
   onQueryTypeChange: (value: string) => void;
+  onChange: (value: PromQuery) => void;
```
ivanahuckova, line 17: I would name this `onExemplarsChange` as we have also other onChange props (`onQueryTypeChange, onStepChange`). So it is more clear.

public/app/plugins/datasource/prometheus/configuration/ExemplarsSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+  );
+  const styles = useStyles(getExemplarsSettingsStyles);
+
+  return (
+    <>
+      <h3 className="page-heading">Exemplars</h3>
+
+      <InlineField label="Internal link" labelWidth={24}>
+        <div className={styles.switch}>
+          <Switch value={isInternalLink} onChange={ev => setIsInternalLink(ev.currentTarget.checked)} />
```
ivanahuckova, line 22: The styling of this toggle is different than rest of the toggles. Border is missing.
image

public/app/plugins/datasource/prometheus/datasource.ts
```diff
@@ -28,9 +28,11 @@ import { expandRecordingRules } from './language_utils';
 import { getQueryHints } from './query_hints';
 import { getOriginalMetricName, renderTemplate, transform } from './result_transformer';
 import {
+  ExemplarTraceIDDestination,
```
ivanahuckova, line 31: nit pick, not sure if this can be change, but following the rest of the formatting, I think this should be `ExemplarTraceIdDestination`.

packages/grafana-ui/src/components/uPlot/geometries/EventsCanvas.tsx
```diff
@@ -1,18 +1,18 @@
...
-import { useRefreshAfterGraphRendered } from '../hooks';
 
-interface EventsCanvasProps<T> {
+interface EventsCanvasProps {
   id: string;
   events: DataFrame[];
-  renderEventMarker: (event: T) => React.ReactNode;
-  mapEventToXYCoords: (event: T) => { x: number; y: number } | undefined;
+  renderEventMarker: (event: DataFrame, index: number) => React.ReactNode;
+  mapEventToXYCoords: (event: DataFrame, index: number) => { x: number; y: number } | undefined;
```
aocenas, line 12: The `event` name for dataFrame is misleading should probably be `dataFrame`. Before it meant single row in a dataFrame of events, but then I would confuse it with DOM click event or something like that.

public/app/plugins/datasource/prometheus/datasource.ts
```diff
@@ -56,6 +58,7 @@ export class PrometheusDatasource extends DataSourceApi<PromQuery, PromOptions>
   queryTimeout: string;
   httpMethod: string;
   languageProvider: PrometheusLanguageProvider;
+  exemplarTraceIDDestination: ExemplarTraceIDDestination | undefined;
```
aocenas, line 61: So right now we can only add one datalink for the exemplars right?

public/app/plugins/panel/timeseries/plugins/ExemplarMarker.tsx
```diff
@@ -47,23 +67,36 @@ export const ExemplarMarker: React.FC<ExemplarMarkerProps> = ({ time, text, tags
       >
         <div ref={annotationPopoverRef} className={styles.wrapper}>
           <div className={styles.header}>
-            {/*<span className={styles.title}>{exemplar.title}</span>*/}
-            {time && <span className={styles.time}>{time}</span>}
+            <span className={styles.time}>Exemplar</span>
```
ivanahuckova, line 70:
```suggestion
            <span className={styles.title}>Exemplar</span>
```

e2e/suite1/specs/explore.spec.ts
```diff
@@ -25,6 +25,6 @@ e2e.scenario({
       });
 
     const canvases = e2e().get('canvas');
-    canvases.should('have.length', 2);
+    canvases.should('have.length', 1);
```
ivanahuckova, line 28: Wondering/not sure why this needs to be changed?

aocenas (commented):
Did not run it but the code looks good have just small comments.

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+  };
+
+  const onLegendClick = useCallback(
+    (event: GraphNGLegendEvent) => {
+      const configWithOverrides = hideSeriesConfigFactory(event, { defaults: fieldConfig, overrides: [] }, data);
+      const applied = applyFieldOverrides({
+        fieldConfig: configWithOverrides,
+        data,
+        timeZone,
+        replaceVariables: value => value, // TODO: replace with real replace
```
ivanahuckova, line 75: `/ / TODO: replace with real replace` What does this mean? Is it for template variables? Cause we don't have any in Explore.

ivanahuckova (commented):
Great Job Zoltan. 🚀 The readability of this code was just amazing! 🔝
I have left some minor feedback/questions/suggestions in the comments. Couple of general questions (some of them not directly related to this PR, I am just being curious and you might have answers):

- Not related to this, but does context menu work only when you click exactly on the circle/point on purpose? Cause in the current graph, it works if you click anywhere on graph line, and here it is sometimes hard to find the exact point. 

- For the exemplars on the top I can't see full info, but when I scroll a little bit down, the exemplars pop-up appears correctly on the bottom
image
image

- I would add a tooltip to Exemplars button that provides more info. Because when it is toggled "on" and you don't run histogram queries, Exemplars don't show (which makes sense, but if you are a new user, you might be wondering if something is wrong).

public/app/plugins/datasource/prometheus/configuration/ExemplarsSettings.tsx
Elfo404: Good catch! i think @torkelo added the component in https://github.com/grafana/grafana/pull/30162

public/app/features/explore/ExploreGraphNGPanel.tsx
zoltanbedi: You are right. We don't need this.

e2e/suite1/specs/explore.spec.ts
zoltanbedi: This test was failing because of this. The old graph panel had 2 canvases 🤷‍♂️

public/app/features/explore/Explore.tsx
```diff
@@ -293,7 +293,6 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
       split,
       queryKeys,
       graphResult,
-      loading,
```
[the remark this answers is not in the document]
zoltanbedi: Good catch!

public/app/plugins/datasource/prometheus/components/PromExemplarField.tsx
zoltanbedi: I like version 1.

public/app/plugins/datasource/prometheus/components/PromExploreExtraField.tsx
zoltanbedi: I'm not sure, this is the onChange handler for the query editor. Maybe it would make sense to move the on change handlers to this component and only pass the onChange and query props down? What do you think?

public/app/plugins/datasource/prometheus/configuration/ExemplarsSettings.tsx
zoltanbedi: yep

public/app/plugins/datasource/prometheus/datasource.ts
zoltanbedi: agree

zoltanbedi:
Thanks for the feedbacks @ivanahuckova !

> Not related to this, but does context menu work only when you click exactly on the circle/point on purpose? Cause in the current graph, it works if you click anywhere on graph line, and here it is sometimes hard to find the exact point.

@dprokop is this something that you want to fix?

> I would add a tooltip to Exemplars button that provides more info. Because when it is toggled "on" and you don't run histogram queries, Exemplars don't show (which makes sense, but if you are a new user, you might be wondering if something is wrong).

I'm absolutely up to this but not sure what should be in the tooltip. What do you think would make sense there?

packages/grafana-ui/src/components/uPlot/geometries/EventsCanvas.tsx
zoltanbedi: makes sense.

public/app/plugins/datasource/prometheus/datasource.ts
zoltanbedi: Changed it to support multiple link configuration

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,165 @@
...
+  timeZone,
+  absoluteRange,
+  onUpdateTimeRange,
+  isLoading,
+  annotations,
+  splitOpenFn,
+}: Props) {
+  const theme = useTheme();
+  const [showAllTimeSeries, setShowAllTimeSeries] = useState(false);
+  const [fieldConfig, setFieldConfig] = useState<FieldConfigSource>({
```
mckn, line 59: Me and @zoltanbedi talked back and forth about moving this to redux and apply the fieldOverrides in the runQuery pipeline via decorators. What would be the benefit of doing that compared to keep it like this @torkelo @dprokop.

One benefit that I can think of is that we would be able to use the fieldOverrides on other parts of Explore. But please jump in here with some more feedback since I don't have good knowledge about the Explore code.

torkelo: > Me and @zoltanbedi talked back and forth about moving this to redux and apply the fieldOverrides in the runQuery pipeline via decorators. What would be the benefit of doing that compared to keep it like this 

As it's already done there for table (sort of via setting display processor). 

And doing there would be it is only done on data updates, not on other rendering triggers. But if its ieasier to use local graph component state & memoize the applyfield overrides call in the explore graph component then I am fine doing it inside the component if that is quicker/easier/faster now

mckn: @zoltanbedi I guess it is up to you :)

zoltanbedi: We thought that we leave it like this for now and will refactor it later if needed. Thanks for the input anyway!

public/app/features/explore/ExploreGraphNGPanel.tsx
```diff
@@ -0,0 +1,180 @@
...
+    return getFieldLinksForExplore({ field, rowIndex, splitOpenFn, range: timeRange });
+  };
+
+  return (
+    <>
+      {dataFramesWithConfig.length > MAX_NUMBER_OF_TIME_SERIES && !showAllTimeSeries && (
+        <div className={cx([style.timeSeriesDisclaimer])}>
+          <Icon className={style.disclaimerIcon} name="exclamation-triangle" />
+          {`Showing only ${MAX_NUMBER_OF_TIME_SERIES} time series. `}
+          <span
```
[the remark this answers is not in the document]
zoltanbedi: Not sure. This is how it was in the old graph component. Should we change this to a button? @jessover9000 
Screenshot 2021-01-13 at 13 33 11

packages/grafana-ui/src/components/Logs/FieldLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const getLinkStyles = () => {
+  return {
+    tag: css`
+      margin-left: 6px;
+      font-size: 11px;
+      padding: 2px 6px;
+    `,
+  };
+};
```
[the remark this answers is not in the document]
zoltanbedi: I have changed this to take up as much space as needed. Also changed the tag to a button.

public/app/features/explore/ExploreGraphNGPanel.tsx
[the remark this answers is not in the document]
jessover9000: I lack a bit of context here. Where does this notification box appear exactly and what do users have to do to trigger it?

ivanahuckova:
>I'm absolutely up to this but not sure what should be in the tooltip. What do you think would make sense there?

I don't think I have enough knowledge about exemplars for this. Do you know if they are only part of histogram queries? Or this is only because of the dev environment we were testing this on and exemplars would be part of all queries? I would probably ask Prometheus-people if they could think of a hint that could be useful for users and if they think there should be one.

dprokop:
> Not related to this, but does context menu work only when you click exactly on the circle/point on purpose? Cause in the current graph, it works if you click anywhere on graph line, and here it is sometimes hard to find the exact point.

> @dprokop is this something that you want to fix?

@ivanahuckova @zoltanbedi I don't think the context menu in Explore works by clicking anywhere in the graph TBH. I've just checked against play, and it only works when clicking on the point/close to the point. Please correct me if I'm wrong

public/app/plugins/datasource/prometheus/components/PromExploreExtraField.tsx
ivanahuckova: >Maybe it would make sense to move the on change handlers to this component and only pass the onChange and query props down? What do you think?

Yes, I 🧡 the idea When I was refactoring & adding the query type picker to Loki, I did the same thing for [LokiOptionFields](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/loki/components/LokiOptionFields.tsx), which is basically same as PrometheusExploreExtraField. But it defnitely doesn't have to be part of this PR, we can refactor it after this is merged, separately.

ivanahuckova:
>@ivanahuckova @zoltanbedi I don't think the context menu in Explore works by clicking anywhere in the graph TBH. I've just checked against play, and it only works when clicking on the point/close to the point. Please correct me if I'm wrong

@dprokop see the gif below. You are righ that you can't click anywhere, but you have largerer area where you can click. In the GraphNG, you have to click exactly on the dot. Wondering, if the area could be larger - at least couple pixels. 
Kapture 2021-01-13 at 14 32 09

dprokop:
>@dprokop see the gif below. You are right that you can't click anywhere, but you have larger area where you can click. In the GraphNG, you have to click exactly on the dot. Wondering, if the area could be larger - at least couple pixels.

@ivana, yups, let me see if I can easily increse the click area :)

public/app/features/explore/ExploreGraphNGPanel.tsx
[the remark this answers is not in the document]
zoltanbedi: So this is in Explore shown when we have more than 20 series in the graph.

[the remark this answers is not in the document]
ivanahuckova: But this is how it currently looks like in the current graph panel as well. So no change here, right?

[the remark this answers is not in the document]
Elfo404: I was only thinking about the semantics of the element rather than its style given that, in fact, it acts like a button.
But it's definitely not blocking given also that this is what we already have

ivanahuckova (approved):
I think it looks awesome. I have played around with the graph and exemplars and everything I noticed was either fixed or is being worked on. Really good job. This is going to rock! 🤘

packages/grafana-ui/src/components/uPlot/geometries/EventsCanvas.tsx
```diff
@@ -1,18 +1,18 @@
...
-import { useRefreshAfterGraphRendered } from '../hooks';
 
-interface EventsCanvasProps<T> {
+interface EventsCanvasProps {
   id: string;
   events: DataFrame[];
-  renderEventMarker: (event: T) => React.ReactNode;
-  mapEventToXYCoords: (event: T) => { x: number; y: number } | undefined;
+  renderEventMarker: (dataFrame: DataFrame, index: number) => React.ReactNode;
+  mapEventToXYCoords: (dataFrame: DataFrame, index: number) => { x: number; y: number } | undefined;
```
dprokop, line 12: Yups, this change makes more sense, thanks

public/app/plugins/datasource/prometheus/datasource.ts
```diff
@@ -194,6 +198,17 @@ export class PrometheusDatasource extends DataSourceApi<PromQuery, PromOptions>
         rangeTarget.instant = false;
         instantTarget.range = true;
 
+        // Create exemplar query
+        if (target.exemplar) {
+          const exemplarTarget = cloneDeep(target);
```
dprokop, line 203: Why do we need the deep clone here? I assume the target is still being issued, and exemplars is just another query based on the original target.

public/app/plugins/panel/timeseries/TimeSeriesPanel.tsx
```diff
@@ -47,10 +52,14 @@ export const TimeSeriesPanel: React.FC<TimeSeriesPanelProps> = ({
...
-      <TooltipPlugin mode={options.tooltipOptions.mode as any} timeZone={timeZone} />
+      <TooltipPlugin mode={options.tooltipOptions.mode} timeZone={timeZone} />
       <ZoomPlugin onZoom={onChangeTimeRange} />
       <ContextMenuPlugin timeZone={timeZone} replaceVariables={replaceVariables} />
-      {data.annotations ? <ExemplarsPlugin exemplars={data.annotations} timeZone={timeZone} /> : <></>}
+      {data.annotations ? (
+        <ExemplarsPlugin exemplars={data.annotations} timeZone={timeZone} getFieldLinks={getFieldLinks} />
+      ) : (
+        <></>
+      )}
```
dprokop, line 62: uhm, is this correct (data.annotations)?

dprokop (approved):
@zoltanbedi I've tested it and went through the code - everything seems fine to me. Left some minor questions only. Great work

public/app/plugins/panel/timeseries/TimeSeriesPanel.tsx
zoltanbedi: It is. Because of this https://github.com/grafana/grafana/blob/32758e7cf0f9118f8f301a7720c8ad1b6cb237e0/public/app/features/query/state/runRequest.ts#L62-L63

PR #20365, opened 2019-11-13
Echo: mechanism for collecting custom events lazily

dprokop:
This PR proposes mechanism for collecting front-end performance metrics. *Echo service* enables metrics consumption (for now it collects [basic front end perf metrics](https://developers.google.com/web/fundamentals/performance/user-centric-performance-metrics) like first paint, dom content loaded, [time to interactive](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/preview#heading=h.k1h25blerz3i))

This is a first step of a bigger endeavour I would like to take to collect front-end performance stats in Hosted Grafana.

In this PR:
- Added `Echo` service and `PerformanceBackend`. Performance backend does not send collected events anywhere yet.
- Added [tti-polyfill](https://github.com/GoogleChromeLabs/tti-polyfill) to collect TTI perf metrix
- Changed the way JS chunks are rendered in index.html template (https://github.com/grafana/grafana/pull/20365/files#diff-34712c43373283c700a1c3d6ff0c9d31R312)

### About Echo
*Echo* in general is an API for consuming *events* and distributing them between *backends* which implement the logic for i.e. sending events to an endpoint. *Echo* flushes the collected metrics once per configured interval (10s by default). 

Important thing to remember is that Echo is intended to work with events that *do not need to be processed immediately, but eventually*.

### Creating backends
Echo's backend needs to implement `EchoBackend<TEvent | TOptions>` interface where:
- `TEvent` are events supported by backend
- `TOptions` are config options for  backend

Backends should be placed under `app/core/services/echo/backends` path.

#### Registering backends
In order for Echo to use backend, you need to register it first. To do so use `registerEchoBackend` function as soon as possible, ideally in `app/index.ts`

#### Supporting multiple event types in a single backend
Echo's backend can specify `supportedEvents` array which defines what event types given backend handles. Echo uses this array to decide if event should be passed to particular backend. If `supportedEvents` is not defined, all events will be passed to the backend.

### TODO
- [x] Implement basic metrics
- [x] Implement client API for metrics consumption
- [ ] Implement service for metrics retrieval and storage (separate PR)
- [ ] Enable `PerformanceConsumer` for HostedGrafana (separate PR)
- [ ] Add docs about Echo and building on top of it (separate PR)

PR #28073, opened 2020-10-07
Logging: Log frontend errors

### LINKED ISSUE #27993: Log frontend errors
Opt-in, configurable way to capture frontend javascript errors and log them.

**Why is this needed**:
There's currently no visibility into errors occuring on frontend, so we rely on manual testing and user feedback to spot problems. It would be great to have frontend error logs as well for discoverability and debugging.

**Relevant links**

- [related post on grafana google group](https://groups.google.com/u/1/g/grafana-developers/c/SodA1v4I1BA)
- [public google doc summarizing the problem, proposed implementation and next actions](https://docs.google.com/document/d/14Bua0_zaJjVC6W_k1BplE2XceLoot5CofTjLOG9EHZ8/edit#)

**Todos**
- [x] base implementation: initialize Sentry on frontend and log it's events on backend
- [x] instrument `ErrorBoundary` in `@grafana/ui` to add component stack to error context, log this in backend
- [ ] configure inbound filters to ignore error spam from extensions, old browsers, etc. [see documentation](https://docs.sentry.io/clients/javascript/tips/)
- [x] add configurable rate limiting to backend API (what's a good way to do it?)
- [x] make sample rate configurable 
- [x] sourcemap support

**Questions / invstigations / ideas**
- [ ] investigate [Sentry breadcrumbs feature](https://docs.sentry.io/platforms/javascript/enriching-events/breadcrumbs/). Is it useful?
- [x] should we log user email / id?
   * Yes!
- [ ] what other metadata is useful and log-worthy for frotnend errors?
- [ ] expose some instrumenation for logging errors/messages via `@grafana/runtime`? Can it be Sentry-specific or should a layer of abstraction be added, to make ditching Sentry easier later?
- [x] what documentation is needed?
- [ ] it might be more robust to build separate entrypoint js file for setting up sentry and load it before loading the main js file, thereby having a better chance of logging working in case main file errors out. Not sure yet if worth the effort and overhead
- [x] whats best way to handle stacktraces with newlines, keeping go logger & loki in mind

domasx2:
There's currently no visibility into errors occuring on frontend, so we rely on manual testing and user feedback to spot problems. It would be great to have frontend error logs as well for discoverability and debugging.

This PR adds an opt-in Sentry integration that can either send events to Sentry, or to grafana backend to be logged to stdout.
Currently it enables default integrations, logging unhandled promise rejections, toplevel errors (`window.onerror`) and async erros occurring in callbacks for `setTimeout` and similar.

This is currently very much POC quality yet

Things that remain @todo for this PR:
- [x] tests

More todos to be implemented via separate PRs on main issue: #27993

How config looks like:
```
[sentry]
# Should Sentry be initialized
enabled = false

# Sentry DSN if you wanna send events to Sentry
dsn =

# Custom endpoint to send sentry events to. if this is configured, dsn will be ignored and events push to this endpoint. Default endpoint will log frontend errors to stdout. Set to empty if using dsn
custom_endpoint = /log_frontend
```

Sample of logged error:

> EROR[10-07|11:07:08] Error: wat                               logger=frontend url="http://localhost:3000/a/grafana-alerting-ui-app/?tab=alerts" user_agent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.121 Safari/537.36" event_id=7e4a2a31165444e7817c1c37141414b6 stacktrace="Error: wat\n  at _next (http://localhost:3000/public/build/AppRootPage.ea8cc955e8f27a4e2d4d.js:29:194)\n  at asyncGeneratorStep (http://localhost:3000/public/build/AppRootPage.ea8cc955e8f27a4e2d4d.js:27:103)\n  at Generator.forEach.e.<computed> [as next] (chrome-extension://hejbmebodbijjdhflfknehhcgaklhano/window/testing-library.js:90:126001)\n  at Generator._invoke (chrome-extension://hejbmebodbijjdhflfknehhcgaklhano/window/testing-library.js:90:125397)\n  at u (chrome-extension://hejbmebodbijjdhflfknehhcgaklhano/window/testing-library.js:90:125644)\n  at AppRootPage._callee$ (http://localhost:3000/public/build/AppRootPage.ea8cc955e8f27a4e2d4d.js:133:22)\n  at AppRootPage.push../node_modules/react/cjs/react.development.js.Component.setState (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:433909:16)\n  at Object.enqueueSetState (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:392719:5)\n  at scheduleUpdateOnFiber (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:402434:9)\n  at flushSyncCallbackQueue (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:390907:3)\n  at flushSyncCallbackQueueImpl (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:390919:7)\n  at runWithPriority$2 (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:390874:10)\n  at unstable_runWithPriority (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:407709:12)\n  at ? (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:390924:24)\n  at performSyncWorkOnRoot (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:402995:11)\n  at workLoopSync (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:403396:22)\n  at performUnitOfWork (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:403420:12)\n  at beginWork$$1 (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:404505:7)\n  at invokeGuardedCallback (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:379165:31)\n  at Object.invokeGuardedCallbackDev (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:379110:16)\n  at HTMLUnknownElement.sentryWrapped (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:82438:23)\n  at HTMLUnknownElement.callCallback (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:379061:14)\n  at beginWork$1 (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:398973:16)\n  at updateMemoComponent (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:396913:14)\n  at updateSimpleMemoComponent (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:397010:10)\n  at updateFunctionComponent (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:397072:20)\n  at renderWithHooks (http://localhost:3000/public/build/vendors~app.ea8cc955e8f27a4e2d4d.js:394985:18)\n  at AppRoot (http://localhost:3000/public/plugins/grafana-alerting-ui-app/module.js:38265:11)" user_email=admin@localhost

Demo:
error logging

#27993

This is POC level, also my first more serious foray into main grafana codebase and go language.  Any and all feedback, suggestions, advice, criticism is much appreciated


conf/defaults.ini
```diff
@@ -825,3 +825,14 @@ use_browser_locale = false
 
 # Default timezone for user preferences. Options are 'browser' for the browser local timezone or a timezone name from IANA Time Zone database, e.g. 'UTC' or 'Europe/Amsterdam' etc.
 default_timezone = browser
+
+
+[sentry]
```
bergquist, line 830: I think we should rename this to "frontend logging" or something like that if we replace the sentry lib.
maybe make this a setting under [log] instead?

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/getsentry/sentry-go"
+	"github.com/grafana/grafana/pkg/infra/log"
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/inconshreveable/log15"
+)
+
+var frontendLogger = log.New("frontend")
+
+type FrontendSentryExceptionValue struct {
```
aknuds1, line 15: There's no need to export this:
```suggestion
type frontendSentryExceptionValue struct {
```

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+type FrontendSentryException struct {
+	Values []FrontendSentryExceptionValue `json:"values,omitempty"`
+}
+
+type FrontendSentryEvent struct {
+	*sentry.Event
+	Exception *FrontendSentryException `json:"exception,omitempty"`
+}
+
+func (value *FrontendSentryExceptionValue) FmtMessage() string {
```
aknuds1, line 30:
```suggestion
func (value *frontendSentryExceptionValue) FmtMessage() string {
```

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+
+var frontendLogger = log.New("frontend")
+
+type FrontendSentryExceptionValue struct {
+	Value      string            `json:"value,omitempty"`
+	Type       string            `json:"type,omitempty"`
+	Stacktrace sentry.Stacktrace `json:"stacktrace,omitempty"`
+}
+
+type FrontendSentryException struct {
```
aknuds1, line 21:
```suggestion
type frontendSentryException struct {
```

bergquist (commented):
Great start! 

I suggest you move the todo list to an issue and implement this in multiple small pull requests since this is behind a feature flag anyway :)

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (value *FrontendSentryExceptionValue) FmtStacktrace() string {
+	var stacktrace = value.FmtMessage()
+	for _, frame := range value.Stacktrace.Frames {
+		stacktrace += fmt.Sprintf("\n  at %s (%s:%v:%v)", frame.Function, frame.Filename, frame.Lineno, frame.Colno)
+	}
+	return stacktrace
+}
+
+func (exception *FrontendSentryException) FmtStacktraces() string {
```
aknuds1, line 42:
```suggestion
func (exception *frontendSentryException) FmtStacktraces() string {
```

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+type FrontendSentryEvent struct {
+	*sentry.Event
+	Exception *FrontendSentryException `json:"exception,omitempty"`
+}
+
+func (value *FrontendSentryExceptionValue) FmtMessage() string {
+	return fmt.Sprintf("%s: %s", value.Type, value.Value)
+}
+
+func (value *FrontendSentryExceptionValue) FmtStacktrace() string {
```
aknuds1, line 34:
```suggestion
func (value *frontendSentryExceptionValue) FmtStacktrace() string {
```

pkg/setting/setting_sentry.go
```diff
@@ -0,0 +1,16 @@
+package setting
+
+type Sentry struct {
+	Enabled        bool   `json:"enabled"`
+	Dsn            string `json:"dsn"`
```
aknuds1, line 5: [Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case.](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
```suggestion
	DSN            string `json:"dsn"`
```

bergquist:
> should we log user email / id?

I think that would be very useful. You can get those values from `c *models.ReqContext` in the backend handler. 

> whats best way to handle stacktraces with newlines, keeping go logger & loki in mind

I guess this is solved in explore when rendering the error? ping @grafana/observability-squad

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,88 @@
...
+var frontendLogger = log.New("frontend")
+
+type FrontendSentryExceptionValue struct {
+	Value      string            `json:"value,omitempty"`
+	Type       string            `json:"type,omitempty"`
+	Stacktrace sentry.Stacktrace `json:"stacktrace,omitempty"`
+}
+
+type FrontendSentryException struct {
+	Values []FrontendSentryExceptionValue `json:"values,omitempty"`
```
aknuds1, line 22:
```suggestion
	Values []frontendSentryExceptionValue `json:"values,omitempty"`
```

pkg/setting/setting_sentry.go
```diff
@@ -0,0 +1,16 @@
...
+	CustomEndpoint string `json:"customEndpoint"`
+}
+
+func (cfg *Cfg) readSentryConfig() {
+	sentryRaw := cfg.Raw.Section("sentry")
+	cfg.Sentry = Sentry{
+		Enabled:        sentryRaw.Key("enabled").MustBool(true),
+		Dsn:            sentryRaw.Key("dsn").String(),
```
aknuds1, line 13:
```suggestion
		DSN:            sentryRaw.Key("dsn").String(),
```

aknuds1 (approved):
Generally looks good to me, but spotted some nits.

conf/defaults.ini
domasx2: `dsn` is sentry-sepcific, but we can call it `sentry_dsn`

- [ ] change  frontend logging config

wbrowne:
Just wanted to say this is awesome ❤️  Big fan of sentry!

public/app/index.ts
```diff
@@ -1,4 +1,5 @@
 import app from './app';
 
+app.initSentry();
```
ryantxu, line 3: @dprokop -- can you take a look?  How close/different is this from the existing echo srv?  could we add a new interface to that?

pkg/api/api.go
```diff
@@ -436,5 +436,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/api/snapshots-delete/:deleteKey", reqSnapshotPublicModeOrSignedIn, Wrap(DeleteDashboardSnapshotByDeleteKey))
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
 
+	// Frontend logs
+	r.Post("/log_frontend", bind(FrontendSentryEvent{}), Wrap(hs.LogFrontendMessage))
```
ryantxu, line 440: I'm not wild about this endpoint... the only `_` in any URL... maybe just "/log"?  or "/log/frontend"?

ryantxu:
Looks great!   I agree with @bergquist that this should get split into a few issues.  Lets find the minimal first steps required in this PR and move most of your TODOs to later issues :)

My biggest concerns are making sure that long term this:
* plays nice with a hosted grafana logging solution :)  loki etc
* does not duplicate too much of the echo server stuff already deployed

pkg/api/api.go
domasx2: `/log` sounds good, don't think there's any scenario to log non frontend stuff
- [ ] rename endpoint

aknuds1 (approved):

domasx2:
Thanks @bergquist , @ryantxu !

> Lets find the minimal first steps required in this PR and move most of your TODOs to later issues :)
For now I moved the todos out of this PR into the relevant issue. Not sure what is the prefered way for grafana, is it ok to keep track of progress on a single issue? Or do I create several ones and somehow group them (project?)?

> * plays nice with a hosted grafana logging solution :)  loki etc
HG is the main target for this currently. My understanding based on some discussion on related HG issue is that logging to stdout is the way to go. We can then filter out frotnend logs in loki by `logger=frontend` 

> * does not duplicate too much of the echo server stuff already deployed
I looked at this briefly and it seems orthogonal, but any insight is appreciated

packages/grafana-data/src/types/config.ts
```diff
@@ -56,6 +56,17 @@ export interface LicenseInfo {
   stateInfo: string;
 }
 
+/**
+ * Describes Sentry integration config
+ *
+ * @public
+ */
+export interface Sentry {
```
dprokop, line 64: nit
```suggestion
export interface SentryConfig {
```

public/app/index.ts
dprokop: @ryantxu my thinking about Sentry reporter was to use EchoSrv as suggested on Slack @domasx2. 

[EchoSrv](https://github.com/grafana/grafana/blob/9e3def4214327dc1efd74fd68913c32f46c24524/packages/grafana-runtime/src/services/EchoSrv.ts#L16) attaches [default meta information to every event](https://github.com/grafana/grafana/blob/f48169633ce9a9387d468081140b1b1a6e2c5ad6/public/app/core/services/echo/Echo.ts#L70) that can be used to construct sentry events when flushing those to the backend, so the initSentry function that @domasx2 you have in this PR can be moved to [EchoBackend](https://github.com/grafana/grafana/blob/9e3def4214327dc1efd74fd68913c32f46c24524/packages/grafana-runtime/src/services/EchoSrv.ts#L52-L51) implementation for Sentry. 

@domasx2 if you want to read more about EchoSrv please take a look at the PR that introdced it https://github.com/grafana/grafana/pull/20365. For a custom async reporter you can have a look ar [MetaAnalyticsBackend](https://github.com/grafana/grafana-enterprise/blob/master/src/public/meta-analytics/MetaAnalyticsBackend.ts) from enterprise.

Not sure if Sentry's transport actually buffers the events before sending, but I think this would be useful for error reporting, especially for the ones that we catch, as those probably do not need to be reported immediately. 

@domasx2 let me know if there is anything unclear here or if me and @ryantxu  are missing something.

dprokop (commented):
Good start!

packages/grafana-data/src/types/config.ts
domasx2: - [ ] rename

public/app/index.ts
domasx2: Thanks @dprokop, I do have concerns:

* `initSentry` does more than configure transport - it wraps global functions, adds various listeners and so on to intercept errors. Also, this is best done as early as possible. Are you sure that an `EchoBackend` is the right place for this? I got the impression that it's only a transport layer

* I looked around, but don't see if `EchoSrv` will flush events if window is about to be closed. I think this is important for error reporting

* Delayed reporting is a concern for logging purposes imho, especially with regards to loki - we can't backdate a log timestamp. Events will appear significantly later than they happened. It's not possible to make it perfect, but would like to make a best effort
I guess we could log original timestamp as a separate field, but then it's not so convenient to investigate

* General unease about relying on system that is being monitored for the implementation of the monitoring :) In my experience error (and metrics!) reporting should be as decoupled as possible. I was even thinking we could configure webpack to spit out a separate entrypoint with Sentry setup that would preceed the main app

Anywyay, let me know what you think. If that's not convincing, I'll try to use echo backend and see how it goes

aocenas:
Just to understand this, is this PR changing anything for development? I would assume you already can see the errors in browser console or am I mistaken?

domasx2:
> Just to understand this, is this PR changing anything for development? I would assume you already can see the errors in browser console or am I mistaken?

Nope, no change for development. This is for collecting errors & possibly other logs from prod instances running on users browsers

public/app/index.ts
dprokop: > initSentry does more than configure transport - it wraps global functions, adds various listeners and so on to intercept errors. Also, this is best done as early as possible. Are you sure that an EchoBackend is the right place for this? I got the impression that it's only a transport layer

EchoSrv allows collecting events and report those to backends of interest on an interval (it basically calls flush on each registered backend). It's the Echo backend that is the transport. Echo is pretty much the first thing initialized in the app, so if Sentry was Echo backend then you get this without the need for `initSentry` function. Since backends are basically classes you could move the sentry config you are doing currently in the `initSentry` function to SentryBackend's constructor and achieve the same behavior.

An additional benefit of Echo is that if needed we can easily implement another Echo backend to consume errors and report to different endpoint/3rd party fairly easily as echo backend consume event types they are interested in (https://github.com/grafana/grafana/blob/f48169633ce9a9387d468081140b1b1a6e2c5ad6/public/app/core/services/echo/Echo.ts#L62). And we can expose `reportError` function, similar to [reportPerformance](https://github.com/grafana/grafana/blob/178bb1d3abbf45c62203cb1bff054bb474a9b4c2/public/app/core/services/echo/EchoSrv.ts#L4-L3) to allow imperative error collection which I think might be interesting for us as well.

> I looked around, but don't see if EchoSrv will flush events if window is about to be closed. I think this is important for error reporting

Yes, this is good point and something that we need to add to Echo.

> Delayed reporting is a concern for logging purposes imho, especially with regards to loki - we can't backdate a log timestamp. Events will appear significantly later than they happened. It's not possible to make it perfect, but would like to make a best effort
I guess we could must log original timestamp as a separate field, but then it's not so convenient to investigate

[Each Echo event has timestamp attached to it](https://github.com/grafana/grafana/blob/f48169633ce9a9387d468081140b1b1a6e2c5ad6/public/app/core/services/echo/Echo.ts#L70). I'm not familiar with how Loki treat logs, but is it the timestamp is added to the log when it's reported? Can't you use the timestamp passed with the log as a source for log timing?


> General unease about relying on system that is being monitored for the implementation of the monitoring :) In my experience error (and metrics!) reporting should be as decoupled as possible. I was even thinking we could configure webpack to spit out a separate entrypoint with Sentry setup that would preceed the main app

I agree and I think this sounds like a good idea going forward.

domasx2: > Since backends are basically classes you could move the sentry config you are doing currently in the initSentry function to SentryBackend's constructor and achieve the same behavior. An additional benefit of Echo is that if needed we can easily implement another Echo backend to consume errors and report to different endpoint/3rd party fairly easily as echo backend consume event types they are interested in.

I get the benefits of a decoupled event ingestion & procesing. 
What would we do if Grafana user wants to report to Sentry backend? In current version we don't configure any custom transport, just let Sentry do it's thing. Should we then still have Sentry dump to `EchoSrv` and somehow marry Sentry internal transport with `SentryEchoBackend`?

> And we can expose reportError function, similar to reportPerformance to allow imperative error collection which I think might be interesting for us as well.

Sentry exposes `captureError` for this, and we must use it rather than pushing to `EchoSrv` direclty to have Sentry gather metadata & instantiate the event.

> I'm not familiar with how Loki treat logs, but is it the timestamp is added to the log when it's reported? Can't you use the timestamp passed with the log as a source for log timing?

My understanding is that Loki currently requires timestamps to be in-order. So it only makes sense to log with timestamp generated at the time the log is ingested in backend, to be in order with other backend logs. There's no way to sort out order for events coming asynchrnously from multiple browsers anyway. I'll try to clarify this with Loki guys though

> I agree and I think this sounds like a good idea going forward.
So all in all, move Sentry init to separate entrypooint? With or without EchoSrv, what's the conclusion? :)

domasx2: Regarding logs I had a chat with Ed:
In summary, surrently logs have to be in-order. This restrition might be lifted in next 3-6 months.  A workaround would be to have log stream per browser session, but is a bad use case for Loki due to high cardinality.
So currenly sounds like best option is to use the timestamp of when the log is output to stdout at backend, so it's in order with other logs for the same stream

dprokop: @domasx2 ok, I think I  see the complete constraints around this work. Correct me if I'm wrong but Sentry will basically send the captured error as soon as it occured, right?

Going forward I think we should do the following:

1. Move sentry configuration (basically initSentry function) into Echo backend. 
2. This backend, since we will be relying on Sentry's transport, would implement `flush` as a noop function and encapsulate the Sentry's `captureError` function within Sentry's backend `addEvent` implementation and expose it indirectly via `reportError` function, similar to what we do with `reportPerformance`. This will enable us in the future to have a different/additional backend for error reports 
3. I wonder if we could hijack automatic Sentry's error capturing so that it also adds events to Sentry's Echo backend if we would like to report the errors to some other storage independently from Sentry's transport. **I don't think we need to do this now though**!

> So all in all, move Sentry init to separate webpack entrypoint? With or without EchoSrv, what's the conclusion? :)

I think you can try this out, taking the initEchoSrv out from the main app and initializing this in a separate chunk.

Does ^^ make sense? I'm pushing towards using Echo (even though not utilizing it's flush behavior now), so that all captured "events" go through the same pipeline.

domasx2: Ok, if we can send errors immediately, I'm onboard with using EchoSrv :) 

If I understand correctly, you suggest that `SentryBackend.addEvent` would call `Sentry.captureError`. This has a flaw however - only errors reported via `addEvent` would make it to `EchoSrv`, but not erorrs captured by Sentry itself through the various hooks it sets up on promises, `window.onerror`, etc.

Then I propose to invert it: I'll override Sentry transport to dump all events into `EchoSrv`, and make `SentryBackend` encapsulate `SentryTransport` to do the http sending. This way all errors will go through `EchoSrv`. Makes sense?

domasx2: - [ ] send all events to `EchoSrv`, attach backends to consume them. Can flush immediately

domasx2:
Hi, @dprokop  @ryantxu  
Other than taking care of misc PR comments, I refactored Sentry integration to route events through `EchoSrv`. 
Basically sentry events are routed  like this now:
`error source` -> `Sentry` -> `EchoSrvTransport` -> `EchoSrv` -> `SentryEchoBackend` -> `CustomEndpointTransport` -> `backend`

Also added tests. Please take a look if it makes sense to you, thanks!

conf/defaults.ini
```diff
@@ -559,6 +559,20 @@ facility =
 # Syslog tag. By default, the process' argv[0] is used.
 tag =
 
+[log.frontend]
```
aknuds1, line 562: You'll need to add these changes in conf/sample.ini also, since it has to be in sync with conf/defaults.ini.

conf/defaults.ini
```diff
@@ -559,6 +559,20 @@ facility =
...
+
+# Should Sentry be initialized
+enabled = false
+
+# Sentry DSN if you wanna send events to Sentry. In this case, set custom_endpoint to empty
+sentry_dsn =
+
+# Custom endpoint to send sentry events to. if this is configured, dsn will be ignored and events push to this endpoint. Default endpoint will log frontend errors to stdout
```
aknuds1, line 570:
```suggestion
# Custom endpoint to send Sentry events to. If this is configured, DSN will be ignored and events push to this endpoint. Default endpoint will log frontend errors to stdout.
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,143 @@
...
+		})
+
+		sc.m.Post(sc.url, sc.defaultHandler)
+		sc.fakeReqWithParams("POST", sc.url, map[string]string{}).exec()
+		fn(sc, logs)
+	})
+}
+
+func TestFrontendLoggingEndpoint(t *testing.T) {
+	ts, _ := time.Parse("2006-01-02T15:04:05.000Z", "2020-10-22T06:29:29.078Z")
```
aknuds1, line 40:
```suggestion
	ts, err := time.Parse("2006-01-02T15:04:05.000Z", "2020-10-22T06:29:29.078Z")
	require.NoError(t, err)
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	Convey("FrontendLoggingEndpoint", t, func() {
```
aknuds1, line 42: Please rewrite without Convey, since we're converting to the standard library for tests.

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	glog "github.com/grafana/grafana/pkg/infra/log"
+	"github.com/grafana/grafana/pkg/models"
+	log "github.com/inconshreveable/log15"
+	. "github.com/smartystreets/goconvey/convey"
```
aknuds1, line 12: This should be removed since we're dropping Convey.

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+type logScenarioFunc func(c *scenarioContext, logs []*log.Record)
+
+func logSentryEventScenario(desc string, event FrontendSentryEvent, fn logScenarioFunc) {
+	Convey(desc, func() {
+		logs := []*log.Record{}
+		frontendLogger.SetHandler(log.FuncHandler(func(r *log.Record) error {
+			logs = append(logs, r)
+			return nil
+		}))
```
aknuds1, line 23: Please add a `t.Cleanup()` function to reset the handler after the test is done, so tests don't leak state between each other.

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func indexOf(arr []interface{}, item string) int {
+	for i := 0; i < len(arr); i++ {
+		if arr[i] == item {
+			return i
+		}
+	}
+	return -1
+}
```
aknuds1, line 137:
```suggestion
func indexOf(arr []interface{}, item string) int {
	for i, elem := range arr {
		if elem== item {
			return i
		}
	}
	return -1
}
```

aknuds1 (changes requested):

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,90 @@
...
+	Value      string            `json:"value,omitempty"`
+	Type       string            `json:"type,omitempty"`
+	Stacktrace sentry.Stacktrace `json:"stacktrace,omitempty"`
+}
+
+type frontendSentryException struct {
+	Values []frontendSentryExceptionValue `json:"values,omitempty"`
+}
+
+type FrontendSentryEvent struct {
```
aknuds1, line 25: This doesn't need exporting.
```suggestion
type frontendSentryEvent struct {
```

pkg/api/frontend_logging.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if len(event.User.Email) > 0 {
+		ctx["user_email"] = event.User.Email
+		ctx["user_id"] = event.User.ID
+	}
+
+	return ctx
+}
+
+func (hs *HTTPServer) LogFrontendMessage(c *models.ReqContext, event FrontendSentryEvent) Response {
```
aknuds1, line 67: This doesn't need exporting.
```suggestion
func (hs *HTTPServer) logFrontendMessage(c *models.ReqContext, event FrontendSentryEvent) Response {
```

aknuds1 (changes requested):

conf/defaults.ini
domasx2: huh, was sure I did that. thanks!
- [ ] update sample.ini

pkg/api/frontend_logging_test.go
domasx2: - [ ] add `t.Cleanup`

aknuds1:
Conflicts also need fixing BTW, in case you haven't noticed.

pkg/api/frontend_logging_test.go
domasx2: Thanks, I'm newb to `go` here, what is the alternative?
- [ ] don't use Convey

domasx2:
> Conflicts also need fixing BTW, in case you haven't noticed.

Yup, didn't want to spend time continuously keeping the PR branch up to date until there was feedback

pkg/api/frontend_logging_test.go
domasx2: ahh, standard lib, gotcha

aknuds1: @domasx2 If you want, I can make the changes and push them to your branch?

domasx2: thanks a lot @aknuds1, but I'd rather do it myself and learn

pkg/api/frontend_logging_test.go
```diff
@@ -137,8 +142,8 @@ func indexOf(arr []interface{}, item string) int {
 	return -1
 }
 
-func soContextContains(logRecord *log.Record, label string, value interface{}) {
-	So(logRecord.Ctx, ShouldContain, label)
+func assertContextContains(t *testing.T, logRecord *log.Record, label string, value interface{}) {
+	assert.Contains(t, logRecord.Ctx, label)
 	labelIdx := indexOf(logRecord.Ctx, label)
-	So(logRecord.Ctx[labelIdx+1], ShouldEqual, value)
+	assert.Equal(t, logRecord.Ctx[labelIdx+1], value)
```
aknuds1, line 148: With `assert.Equal` the expectation comes before the actual value.
```suggestion
	assert.Equal(t, value, logRecord.Ctx[labelIdx+1])
```

conf/defaults.ini
```diff
@@ -560,6 +560,20 @@ facility =
 # Syslog tag. By default, the process' argv[0] is used.
 tag =
 
+[log.frontend]
+
```
aknuds1, line 564: Nit, but could you drop this newline? We don't have them after section headers elsewhere in this file.

conf/sample.ini
```diff
@@ -551,6 +551,21 @@
 # Syslog tag. By default, the process' argv[0] is used.
 ;tag =
 
+[log.frontend]
+
```
aknuds1, line 555: Nit, but could you drop this newline? We don't have them after section headers elsewhere in this file.

conf/sample.ini
```diff
@@ -551,6 +551,21 @@
...
+# Sentry DSN if you wanna send events to Sentry. In this case, set custom_endpoint to empty
+;sentry_dsn =
+
+# Custom endpoint to send Sentry events to. If this is configured, DSN will be ignored and events push to this endpoint. Default endpoint will log frontend errors to stdout.
+;custom_endpoint = /log
+
+# Rate of events to be reported between 0 (none) and 1 (all), float
+;sample_rate = 1.0
+
+
```
aknuds1, line 568: Nit, but could you remove this extra blank line? It also isn't in conf/defaults.ini.

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+			return nil
+		}))
+		t.Cleanup(func() {
+			frontendLogger.SetHandler(origHandler)
+		})
+
+		sc := setupScenarioContext("/log")
+		hs := HTTPServer{}
+
+		sc.defaultHandler = Wrap(func(w http.ResponseWriter, c *models.ReqContext) Response {
```
aknuds1, line 33: Is there a reason not to use sc.handlerFunc instead? That hook is called by sc.defaultHandler if I'm not mistaken.

aknuds1 (approved):
LGTM with some nits. Nicely cleaned up! :)

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+
+		logSentryEventScenario(t, "Should log received error event", errorEvent, func(sc *scenarioContext, logs []*log.Record) {
+			assert.Equal(t, sc.resp.Code, 200)
```
aknuds1, line 95: With assert.Equal, the expectation comes before the actual value.
```suggestion
			assert.Equal(t, 200, sc.resp.Code)
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+			assert.Equal(t, len(logs), 1)
```
aknuds1, line 96:
```suggestion
			assert.Len(t, logs, 1)
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+				Request:   &request,
+				Timestamp: ts,
+				Message:   "hello world",
+				User:      user,
+			},
+			nil,
+		}
+
+		logSentryEventScenario(t, "Should log received message event", messageEvent, func(sc *scenarioContext, logs []*log.Record) {
+			assert.Equal(t, sc.resp.Code, 200)
```
aknuds1, line 120:
```suggestion
			assert.Equal(t, 200, sc.resp.Code)
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+			assert.Equal(t, len(logs), 1)
+			assert.Equal(t, logs[0].Msg, "hello world")
+			assert.Equal(t, logs[0].Lvl, log.LvlInfo)
```
aknuds1, line 123:
```suggestion
			assert.Len(t, logs, 1)
			assert.Equal(t, "hello world", logs[0].Msg)
			assert.Equal(t, log.LvlInfo, logs[0].Lvl)
```

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+			return i
+		}
+	}
+	return -1
+}
+
+func assertContextContains(t *testing.T, logRecord *log.Record, label string, value interface{}) {
+	assert.Contains(t, logRecord.Ctx, label)
+	labelIdx := indexOf(logRecord.Ctx, label)
+	assert.Equal(t, logRecord.Ctx[labelIdx+1], value)
```
aknuds1, line 148:
```suggestion
	assert.Equal(t, value, logRecord.Ctx[labelIdx+1])
```

aknuds1 (approved):
Found some more nits.

conf/sample.ini
domasx2: - [ ] remove blank line

conf/sample.ini
domasx2: - [ ] remove whitespace

pkg/api/frontend_logging_test.go
domasx2: I reused this piece of code from another test without without polishing it too much. But now that I look at it, don't need to set either `handlerFunc` or `defaultHandler` here, as not using any of the helper scenarios that actually use these. Updated code to reflect it

aknuds1: Great :)

pkg/api/frontend_logging_test.go
```diff
@@ -0,0 +1,149 @@
...
+	t.Run("FrontendLoggingEndpoint", func(t *testing.T) {
+		request := sentry.Request{
+			URL: "http://localhost:3000/",
+			Headers: map[string]string{
+				"User-Agent": "Chrome",
+			},
+		}
+
+		user := sentry.User{
+			Email: "geralt@kaermorhen.com",
```
dprokop, line 57: :D

package.json
```diff
@@ -206,6 +206,9 @@
   "dependencies": {
     "@grafana/slate-react": "0.22.9-grafana",
     "@reduxjs/toolkit": "1.3.4",
+    "@sentry/browser": "^5.25.0",
+    "@sentry/types": "^5.24.2",
+    "@sentry/utils": "^5.24.2",
```
dprokop, line 211: Nit: please use fixed versions

dprokop (approved):
@domasx2 i think this looks good now and should be ready to merge. Left one nit only

package.json
domasx2: done!

PR #28079, opened 2020-10-07
Update bot comments

marcusolsson:

Updates the comments for the grafanabot.

A couple of things that we might want to put back, but clarified/rephrased:

- “See also our issue reporting”: Why should they see this? Are we insinuating that they should’ve read the guidelines before posting? Are we suggesting that they might want to report a bug instead?
- “Happy coding”: The user that created the issue might not be a programmer. If we want to sign off the comment, we should have something that’s applicable to any contributor.

### DIFF
--- a/.github/commands.json
+++ b/.github/commands.json
@@ -3,7 +3,7 @@
     "type": "label",
     "name": "type/question",
     "action": "close",
-    "comment": "Please ask your question on [community.grafana.com/](https://community.grafana.com/). See also our [issue reporting](https://github.com/grafana/grafana/blob/master/CONTRIBUTING.md#report-bugs) guidelines.\n\nHappy Coding!"
+    "comment": "Please ask your question on [community.grafana.com/](https://community.grafana.com/)."
   },
   {
     "type": "comment",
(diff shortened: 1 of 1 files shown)


torkelo:
If we label an issue a question it's clear they have not read our guidelines for issues, so that feels like an excellent thing to link to. Maybe we need to have better info on issue guidelines to link to?

torkelo:
“Happy coding"  was from vscode, maybe we can come up with something similar that fits Grafana better?

marcusolsson:
> If we label an issue a question it's clear they have not read our guidelines for issues, so that feels like an excellent thing to link to. Maybe we need to have better info on issue guidelines to link to?

Then maybe we could frame it in a way that makes it clear what the benefits are to the person reporting the issue? Something like this maybe?:

_To avoid having your issue closed in the future, please read our guidelines on [CONTRIBUTING](https://github.com/grafana/grafana/blob/master/CONTRIBUTING.md)._

marcusolsson:
> “Happy coding" was from vscode, maybe we can come up with something similar that fits Grafana better?

Happy graphing? :)

torkelo:
> Happy graphing?
:+1:

torkelo (approved):

PR #28082, opened 2020-10-07
Bugfix: adding polyfills for supporting intl.DateTimeFormat in older browsers.

### LINKED ISSUE #28070: Samsung Tizen TV (1280) - Not Loading
**What happened**:

image

https://github.com/grafana/grafana/blame/61bd33c24187184b058571c42f0e5f5129d22320/packages/grafana-data/src/datetime/formats.ts#L91 introduced code which breaks Samsung Tizen TVs (ver 1280)

It not to be broken

Load the login page on a Samsung Tizen TV

Console debugging shows which lead us here.
image

Seems the Tizen supports a subset of the Intl namespace, a fix could be as simple as changing line 91 of that patch to

```
  if (!window.Intl || !Intl.DateTimeFormat.prototype.formatToParts) {
```

I have also had some success with polyfilling support...

**Environment**:
- Grafana version: v7.2.0 (efe4941ee3)
- User OS & Browser: TIZEN Web App

mckn:
Adds a set of polyfills that will give older browsers intl.DateTimeFormat support which we need to handle the new date time formatting: https://github.com/grafana/grafana/blame/61bd33c24187184b058571c42f0e5f5129d22320/packages/grafana-data/src/datetime/formats.ts

Fixes #28070

I need help verifying that this works for e.g. Samsung Tizen TV (1280)


dprokop (approved):
Looks good! Unfortunately no Samsung TV to verify on :(

mckn:
I am closing this one in favor of the following: #28086

The reason is that when adding the polyfill we also need to include the locale and timeZone data for all the locales/timeZones that we want to support. That makes this a lot of extra code that needs to be bundled just to support old browsers.

So lets just fallback to the default formatting for older browsers instead of adding a polyfill.

PR #28083, opened 2020-10-07
Drone: Add release pipelines

aknuds1:
Add Drone pipelines for release and test release builds, porting them from CircleCI.

I would recommend primarily reviewing .drone.yml, as opposed to the *.star source files, since you can see the generated Drone configuration there. Basically, I'm adding two "super pipelines" (calling them this because each consists of multiple individual pipelines), one for doing releases and one for simulating releases (to test the former).

Once the PR is merged I plan to test by executing the test-release (super) pipeline, and see how it goes basically :P

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,12 +106,6 @@ jobs:
     executor: grafana-build
     steps:
       - run:
-          name: Exit if enterprise and forked PR
-          command: |
-            if [[ "<< parameters.edition >>" == "enterprise" && -n "$CIRCLE_PR_NUMBER" ]]; then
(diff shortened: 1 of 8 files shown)


kylebrandt (approved):
Hopefully it works :crossed_fingers: 

I'm not sure how some of this translates into processes. You may want to consider setting the API key environment variable by putting it before the command (so it is only available to that command and its sub-processes) instead of export - but maybe it works out that anyways.

aknuds1:
@kylebrandt It's a good point you're making. I don't the Drone config uses `export` though, only the Circle stuff? In CircleCI, AFAIK, the exports are only visible to grabpl.

PR #28085, opened 2020-10-07
Alerting: Add labels to name when converting data frame to series

### LINKED ISSUE #28068: Group by "tag" not available in alert data associated with Azure Log Analytics query
For InfluxDB datasources, I've had a lot of success creating alerts that cover several targets (ie. servers, etc.) by using a single query with a "group by" clause. In this case, the alert result exposes the "tag" from the "group by" clause in various parts of the result to delineate them. As an example, here is an InfluxDB query and corresponding alert result that demonstrates how the "datacenter" tag is bubbled up in the alert results:

Query
image

Alert Result
image

**What happened**:
However, I've noticed that similar queries in Azure Monitor/Log Analytics using a "summarize .. by" do not expose the "tag" being grouped on. Consequently, when alert notifications are sent out, it is unclear which targets are failing to pass the alert criteria. As an example, the attached query performs a "summarize" by "Computer".  In the alert results, you can see that the "Computer" tag is captured in one portion of the query result. However, it is not exposed in the portions of the result (like the data series name or message) that I assume are used by the notifications. **I would expect the Azure Monitor-based alert to provide the "tag" data in the notification message(s), just like the InfluxDB datasource does.**

Query:
image

Alert Result:
image

Notification
image

**Environment**:
- Grafana version: v7.2.0
- Data source type & version: Azure Monitor v0.3.0

kylebrandt:
When converting a data frame with time series to a tsdb.TimeSeries this makes it so the labels (if present) are appended to the series name.

In turn, this makes it so it shows up in the message of each Eval match when testing (EvalMatch.Metric):

image

Then, I believe notifiers like slack the iterate over the matches will display them:

https://github.com/grafana/grafana/blob/a2e676f6755faf2322326f4d3c308e71193da91b/pkg/services/alerting/notifiers/slack.go#L202-L207

Fixes #28068

**Other Notes**:

I also moved the conversion function from the tsdb to the alerting package since that is the only place it is used, and it is getting stuff specific to alerting.

Example slack notification:

image


pkg/services/alerting/conditions/query.go
```diff
@@ -294,3 +295,63 @@ func validateToValue(to string) error {
...
+		ts := &tsdb.TimeSeries{
+			Name:   field.Name,
+			Points: make(tsdb.TimeSeriesPoints, field.Len()),
+		}
+
+		if field.Labels != nil {
+			ts.Tags = field.Labels.Copy()
+			// Tags are appended to the name so they are eventually included in EvalMatch's Metric property
+			// for display in notifications.
+			ts.Name = fmt.Sprintf("%v {%v}", ts.Name, field.Labels.String())
```
marefr, line 339: I guess the idea is to have a fallback on name when not provided. But some data sources sets the name and/or display name and I wonder whether we only should fallback if there are no existing name/display name. Or maybe I'm missing something?

For example, see azure monitor if an alias is provided or not: https://github.com/grafana/grafana/blob/54c503dcc3d930a1bcb74d3c3293d41971e2a245/pkg/tsdb/azuremonitor/azuremonitor-datasource.go#L294-L312

marefr (commented):
Looks good. Awesome PR description btw :+1:  I did add a comment/question that I would like you to have a look at before I approve this.

pkg/services/alerting/conditions/query.go
```diff
@@ -294,3 +295,63 @@ func validateToValue(to string) error {
 	_, err := time.ParseDuration(to)
 	return err
 }
+
+// FrameToSeriesSlice converts a frame that is a valid time series as per data.TimeSeriesSchema()
+// to a TimeSeriesSlice.
+func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
+	tsSchema := frame.TimeSeriesSchema()
+	if tsSchema.Type == data.TimeSeriesTypeNot {
+		// If no fields, or only a time field, create an empty tsdb. with a single
```
papagian, line 304:
```suggestion
		// If no fields, or only a time field, create an empty tsdb.TimeSeriesSlice with a single
```

papagian (commented):
Generally LGTM; waiting for the reply to @marefr [question](https://github.com/grafana/grafana/pull/28085#discussion_r501323227) for approving it.

pkg/services/alerting/conditions/query.go
kylebrandt: @marefr Good call. I have updated it to use the displayName if it is set.

When using the Alias query field in the Azure Monitor service in the Azure Monitor data source, it sets the field config's display name:

Without Alias:

image

With an Alias ( e.g. `🧵 {{blobtype}}  🧵 {{tier}}`):

image

image

marefr (approved):
LGTM great

Issue #22209, opened 2020-02-14
New Stat Panel scalability improvement.


scroodj:
Hello,

First of all i would like to give my gratitude to you guys for that you are doing.
Thank you for a new Stat Panel arrived with 6.6 release. I see a big potential of using this panel.
Nevertheless i want to suggest some improvements.

When it comes to using it with just a few numbers of series it looks good:
![stat_panel_6_series](https://user-images.githubusercontent.com/6392717/74521332-8e654680-4f19-11ea-9bee-a575707b1212.PNG)

but with increasing number of output values(multiple series) it becomes not user friendly:
![stat_panel_multiple_series](https://user-images.githubusercontent.com/6392717/74521521-cec4c480-4f19-11ea-8c77-24b3c207c159.PNG)
 

**What would you like to be added**:
1. Use Vertical Scalability (multiply rows) instead of using Horizontal Scalability within a same panel. Example:
![stat_panel_multiple_series_multiple_rows](https://user-images.githubusercontent.com/6392717/74521704-2400d600-4f1a-11ea-9888-205ecbd7c963.PNG)
2. Control over the font size.
3. Ability to fix individual cube size.

torkelo:
We plan to add a grid layout mode, but not manual font size , that will not help

pkolyvas:
We're in the process of moving Grafana feature requests to [GitHub Discussions](https://community.grafana.com/t/were-moving-grafana-feature-requests-to-github-discussions/50341/3). This issue is being moved to [Discussions](https://github.com/grafana/grafana/discussions) or has been moved from our issues list, depending on where you're seeing this message. 

 We hope to use the discussion thread to continue building use cases, and additional evidence based on engagement/voting, in order to help us prioritize this feature request later on down the road. 

As with all things Grafana, we value your input and participation in this project and appreciate your patience during this transitional period. Looking forward to the discussion!

PR #28091, opened 2020-10-07
Stats/BarGauge: allow setting text size

### LINKED ISSUE #23305: Enable font size option for Stat panel
I would like the possibility to change the font-size for the field title in the Stat panel. Right now I am showing the status about my hosts networkadapters where I have the networkadapter names in the font-title. 
But the text is so small I can bearly see it unless I resize the panel to take the whole screen.
I also searched the form and found several other questions about this but none had been answered.

ryantxu:
Fixes #23305  and helps #22209

Currently, most panels do not allow setting font sizes directly, rather the font sizes depend on the text we are trying to display.  We do this automatically for a couple reasons:
* allowing a fixed size does not scale well when the dashboard resizes
* the number of distinct font size options is potentially large.  For example, the bar gauge has different sizes for title, text, unit in both horizontal and vertical mode.
* as panel structures change, migrating the explicit settings gets really awkward

Under the hood, the automatic calculations are all based on the *longest* strings the visualization has to show.  In this PR, we let people configure the string lengths explicitly rather than calculating them.

|  Automatic | Explicit   | 
|---|---|
|  image | image |

I went though a few options to enter the text length, and am finally using a select -- any better ideas?  Typing the text felt awkward with slow refresh rates and unclear response

image

TODO:

- [ ] Docs for stats panel  
- [ ] Docs for bar gauge panel

(I'll look at the docs once we agree on a reasonable structure/UI)


public/app/features/dashboard/components/PanelEditor/PanelOptionsEditor.tsx
```diff
@@ -57,11 +57,12 @@ export const PanelOptionsEditor: React.FC<PanelOptionsEditorProps<any>> = ({
               return null;
             }
 
-            const label = (
-              <Label description={e.description} category={e.category?.slice(1)}>
-                {e.name}
-              </Label>
-            );
+            const label =
```
ryantxu, line 60: This change will only show the label if name|description exists.  This allows me to make the `<Label ` explicit in the body depending on the option state

public/app/plugins/panel/stat/AlignmentFactorText.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Input } from '@grafana/ui';
+import { css } from 'emotion';
+import { config } from 'app/core/config';
+
+interface Props {
+  value?: string;
+  onChange: (value?: string) => void;
+}
+
+export const AlignmentFactorTextEditor: React.FC<Props> = ({ value, onChange }) => {
```
ryantxu, line 11: NOTE:  this was my direct text option, and should be removed if we stick with the `<Select` version

ryantxu:
@torkelo -- this is not great... but I am not sure what a better approach would be

public/app/plugins/panel/stat/AlignmentFactorLengthEditor.tsx
```diff
@@ -0,0 +1,78 @@
...
+      label: getLabelText(i),
+    };
+  });
+
+export const AlignmentFactorLengthEditor: React.FC<Props> = ({ value, onChange }) => {
+  const styles = {
+    textInput: css`
+      margin-bottom: 8px;
+    `,
+  };
```
hugohaggmark, line 32: nit: maybe move styles function outside of the component and use `useStyles` hook

public/app/plugins/panel/stat/AlignmentFactorLengthEditor.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const getTextLengthFrom = (txt: string) => {
+    if (!txt) {
+      return 0;
+    }
+    const len = parseInt(txt, 10);
+    if (isNaN(len)) {
+      return txt.length;
+    }
+    return len;
+  };
```
hugohaggmark, line 53: maybe move this one outside of the component?

hugohaggmark (commented):
I think the code looks great and I just left some minor nits. 

As a first time user of the UX I find the Explicit factors a bit complicated to understand. Can the 8 - xxxxx be visualized in another way? Maybe with a slider and showing in real time how the Font size is changed?

ryantxu:
> Explicit factors a bit complicated to understand.

:rofl: no kidding

I'm not sure the best solution... in the end I think it may be exposing explicit font sizes!  but that is a bigger project.  uggg

torkelo:
This was more complex than I imagined, do we need this many separate settings? I imaged it would be enough with title & value

ryantxu:
I removed the 7.3 label since this feels too weird, and I think we should likely just expose some standard real text sizes (like text and title)

ryantxu:
Closing this in favor of #29476

PR #28101, opened 2020-10-08
Live: remove admin pages, add alpha panel

ryantxu:
This removes a jankey admin page (hidden behind a feature flag) and moves the features to an alpha panel called live.

This is the first step, and the aim is really just basic debugging:
image

Down the road (not this PR, but soon), this panel will be improves to that it:
* can optionally show status and/or message
* format message as JSON, table, and possibly a markdown/html template
* option to POST into the channel


papagian (approved):
Backend-wise looks good

public/app/plugins/panel/live/LiveChannelEditor.tsx
```diff
@@ -0,0 +1,164 @@
...
+export class LiveChannelEditor extends PureComponent<Props, State> {
+  state: State = {
+    // scope: LiveChannelScope.Grafana,
+    // namespace: 'testdata',
+    // path: 'random-2s-stream',
+    namespaces: [],
+    paths: [],
+  };
+  // onTextChanged: ((event: FormEvent<HTMLInputElement>) => void) | undefined;
+  // onPublish: ((event: MouseEvent<HTMLButtonElement, MouseEvent>) => void) | undefined;
```
mckn, line 43: nit: maybe get rid of the commented code before merging this to master.

public/app/plugins/panel/live/LiveChannelEditor.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const { namespaces, paths } = this.state;
+    const { scope, namespace, path } = this.props.value;
+    const style = getStyles(config.theme);
+
+    return (
+      <>
+        <FeatureInfoBox
+          title="Grafana Live"
+          featureState={FeatureState.alpha}
+          // url={getDocsLink(DocsId.Transformations)}
```
mckn, line 115: nit: another thing that might be worth removing.

public/app/plugins/panel/live/LivePanel.tsx
```diff
@@ -0,0 +1,160 @@
...
+  unsubscribe = () => {
+    if (this.subscription) {
+      this.subscription.unsubscribe();
+      this.subscription = undefined;
+    }
+  };
+
+  async loadChannel() {
+    const addr = this.props.options?.channel;
+    if (isValidChannel(addr)) {
```
mckn, line 73: nit: if you convert the `isValidChannel` to a guard function then you don't have to do the `addr!` when fetching the channel.

mckn: nit: another thing that might be worth considering is to invert the if statement to prevent unnecessary nesting. So what you basically could do is to is change the the ` if (isValidChannel(addr))` to `if (!isValidChannel(addr)))` and do an early return.

mckn (approved):
I think it looks good! I have added some small nits to consider.

public/app/plugins/panel/live/LivePanel.tsx
ryantxu: nice -- thanks!

PR #28105, opened 2020-10-08
Plugins: do not remount app plugin on nav change

domasx2:
Calling `onNavChanged` from inside an app plugin for the first time causes the plugin to be unmounted & re-mounted, because it's parent changes in [AppRootPage.tsx](https://github.com/grafana/grafana/blob/master/public/app/features/plugins/AppRootPage.tsx#L87). This can be problematic, causing plugin to re-create redux store, re-issue "onload" data requests, flicker. 
Reparenting is a tricky problem, there's [an issue in react repo dating back to 2015](https://github.com/facebook/react/issues/3965).
This PR introduces [react-reverse-portal](https://github.com/httptoolkit/react-reverse-portal) which solves the issue using portals. Let me know if this makes sense to you, I can try to bang out a test for it


torkelo:
Not sure this makes sense (or seems a complex solution). 

If you switch page the app should unmount. Should be a way to not re-create redux store on-mount? check if it's already created?

torkelo:
Are you switching from an app page with nav model to one without?

domasx2:
> Are you switching from an app page with nav model to one without?

The other way around, initially app page has no nav model, but sets it on first load.

> If you switch page the app should unmount. Should be a way to not re-create redux store on-mount? check if it's already created?

Switching between pages did not unmount previously either. The problem is only with switching from no nav to nav, or vice versa, and seems a side-effect of implementation rather than intentional...
I'm not sure though why initial nav is not loaded from `plugin.json`, this config only affects sidebar? Seems odd

torkelo:
AppRootPage and the onNavChanged feel like a bit of a hack, but adding portal workarounds feels like hack on hack 

I did a POC of a react app router that would allow for more control over URL and sub pages. 
https://github.com/grafana/grafana/pull/26245

But we decided to wait with that until we switch main app to use react router (still on angular router for the main app). 

@ryantxu any ideas for what to do short term here?

ryantxu:
> feel like a bit of a hack, but adding portal workarounds feels like hack on hack

Lets all agree that is *is*/*was* a hack... but has let us move forward while we try to figure out react router ;)

My sense here is that we all know the real solution is react-router, but that is a big lift and will take some time and attention.  Meanwhile we are building apps full-speed and need whatever we can to make that better.  Since this does not change any interfaces for the apps, I think it is a good temporary step.

So I am +1 for this band-aid, and we will rip it all out when we are ready for react-router

hugohaggmark:
> > feel like a bit of a hack, but adding portal workarounds feels like hack on hack
> 
> Lets all agree that is _is_/_was_ a hack... but has let us move forward while we try to figure out react router ;)
> 
> My sense here is that we all know the real solution is react-router, but that is a big lift and will take some time and attention. Meanwhile we are building apps full-speed and need whatever we can to make that better. Since this does not change any interfaces for the apps, I think it is a good temporary step.
> 
> So I am +1 for this band-aid, and we will rip it all out when we are ready for react-router

I agree with Ryan so +1 from me too.

torkelo:
Not sure I agree, by never addressing app plugins and continuing with temporary hacks we just build more technical debt by having plugins built in this rocky foundation. React router is still not coming very soon. Q1 21 at the earliest. 

Is the problem that we need to load the page to get the initial nav model? Can we get the nav model from the AppPlugin before we load the page? is there a way to not have unmount be a big problem? Don't see how initializing a redux store onmount cannot be solved some other way. Redux stores live longer than a react component lifetime so should be able to know that it's already been initialized. 

Another option could be to add a static function to AppRootPage to get the initial nav model? We could define page routes in plugin.json as we do for angular pages. This we probably need to do anyway for react pages to have access control on pages and so the can be added to the main nav.

domasx2:
It's not a big or urgent problem, and we are working around it, I'm not looking to die on this hill ;)

It's just unexpected behaviour that I thought would be nice to fix - you generally don't expect your root component to be almost arbitrarily re-mounted. Redux store is just one example. It's a common pattern to add providers to top level component, that you expect to not be arbitrarily unmounted, and have these providers keep some state. Another isnigificant but odd behaviour related to this: if plugin root component throws, "error" nav model is created. Currently this causes plugin to be mounted again, and error gets thrown twice.

I don't see yet why this fix would count as adding technical debt - I don't think it adds to work to be done when we switch to react-router. 

Regarding alternatives, we can define pages in `plugin.json`, for example: https://github.com/grafana/alerting-ui-plugin/blob/master/src/plugin.json#L17. Currently they are used to populate sidebar items. Maybe we can then just use these for the initial nav? I'm not sure why currently they're not, maybe missing some context

torkelo:
Another idea maybe is to expose Page component from grafana-u so it can be used from the plugin page component?

ryantxu:
@domasx2 proposal improves the current behavior without changing any API.   

@torkelo your proposals chang the interfaces that we all agree should be replaced with direct react router access when that is possible

ryantxu:
Progress, not perfection :)

torkelo (approved):
Just afraid that if we always address the pain with another workaround. 

If we exposed the Page component to an app we would have no need for onNavChanged

domasx2:
Thanks @torkelo , I'll see if I can add a test to this as well.
I like the idea to expose Page via ui package, will see if I can find the time to work on it

domasx2:
@torkelo  @ryantxu  added a test reproducing the issue, I think nothing more to be done on this PR

torkelo (approved):

ryantxu (approved):
looks good -- i think it should target 7.3 also

PR #18818, opened 2019-08-30
FieldConfig: expose nullValueMode and respect Ignore for react panels

ryantxu:
Fixes: #18785

~~Totally untested, but I needed something simple to do :)~~ alas simple things never are 😢

PR #28118, opened 2020-10-08
Graph NG: Added support for connecting points when having null values.

mckn:
This makes the line connect in the graph even when having data points with null values in the data set.

Kapture 2020-10-08 at 15 23 31

I changed the `NullValuesMode` to an enum to prevent having 'magic strings' in the code. Not sure if this could cause any issues when serializing the field configuration.


packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -2,7 +2,11 @@ import React from 'react';
 import uPlot from 'uplot';
 import { DataFrame, FieldColor, TimeRange, TimeZone } from '@grafana/data';
 
-export type NullValuesMode = 'null' | 'connected' | 'asZero';
+export enum NullValuesMode {
+  null = 'null',
+  connected = 'connected',
+  asZero = 'asZero',
+}
```
dprokop, line 9: 👍

packages/grafana-ui/src/components/uPlot/utils.ts
```diff
@@ -62,6 +62,7 @@ export const preparePlotData = (data: DataFrame): uPlot.AlignedData => {
     let values = field.values.toArray();
 
     if (field.config.custom?.nullValues === 'asZero') {
+      console.log('replacing null with 0');
```
dprokop, line 65:
```suggestion
```

public/app/plugins/panel/graph3/GraphPanel.tsx
```diff
@@ -188,7 +188,7 @@ export const GraphPanel: React.FC<GraphPanelProps> = ({
     }
     if (seriesGeometry.length > 1) {
       geometries.push(
-        <SeriesGeometry key={`seriesGeometry-${scale}-${i}`} scaleKey={scale}>
+        <SeriesGeometry key={`seriesGeometry-${scale}-${i}`} scaleKey={scale} nullValues={customConfig?.nullValues}>
```
dprokop, line 191: This is going to work only if you have more than one geometry(line+points, points+area etc). If you config to display the Line only then this setting will not be applied. I think we need to extend the Line and Point geometry interface to accept property `nullValues` so that the span gaps of the line can be controlled independently from the SeriesGeometry as well :)

dprokop (commented):
Great start, left one important comment :)

public/app/plugins/panel/graph3/GraphPanel.tsx
mckn: ah, good point!

mckn: Would it be any drawbacks of always adding a SeriesGeometry group regardless of how many series there are?

dprokop: No, not really. But since the graph primitives are to be used by plugin devs as well, I thought it would be nice to have possibility to make the composition as plain as possible.

But when I think about it it actually adds unnecessary complexity. I think you suggestion makes sense to always require wrapping geomerties in SeriesGeometry - let's do it!

torkelo:
Not sure null value option should be part of the series rendering logic. It might need to be earlier than that and shared with any reducer calculations that are done for the legend. 

It's tricky. @ryantxu  & I explored making null value mode a standard field config option over a year ago, https://github.com/grafana/grafana/pull/18818 But struggled with some behaviors.

torkelo:
But this might be the best solution still.

Think we need to update the wording a bit. 
But think we can improve option name & the different radio options.

mckn:
> Not sure null value option should be part of the series rendering logic. It might need to be earlier than that and shared with any reducer calculations that are done for the legend.
> 
> It's tricky. @ryantxu & I explored making null value mode a standard field config option over a year ago, #18818 But struggled with some behaviors.

This is great feedback! Still kind of new to this part of the code base but I will dig in to the PR you referenced.

packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -2,7 +2,11 @@ import React from 'react';
 import uPlot from 'uplot';
 import { DataFrame, FieldColor, TimeRange, TimeZone } from '@grafana/data';
 
-export type NullValuesMode = 'null' | 'connected' | 'asZero';
+export enum NullValuesMode {
+  null = 'null',
```
leeoniya, line 6: @mckn 

imo, "Gaps" would be more clear here. "display null as null" doesnt really convey what the result would look like.

ryantxu: +1  We should change that in the UI.  This property has been around for a long long time, so it is hard to change, but we can certainly change what is displayed in the UI

ryantxu: why are you creating a new NullValuesMode object vs the one in https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/data.ts#L139

public/app/plugins/panel/graph3/GraphPanel.tsx
ryantxu: The null values mode is already on each field...  can we use it from there?
https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/dataFrame.ts#L58

packages/grafana-ui/src/components/uPlot/types.ts
mckn: Awesome feedback! Will have a look 👍

public/app/plugins/panel/graph3/GraphPanel.tsx
mckn: Great feedback, will have a look!

packages/grafana-ui/src/components/uPlot/types.ts
mckn: @ryantxu I did not know that we had that one :) But I have replaced it with the one you referenced.

mckn:
Would be awesome to get some more feedback before adding tests.

packages/grafana-ui/src/components/uPlot/types.ts
dprokop: +1 for the change in the UI

packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -132,4 +132,5 @@ export enum FieldConfigProperty {
   Mappings = 'mappings',
   Links = 'links',
   Color = 'color',
+  ModeNullValue = 'nullValueMode',
```
dprokop, line 135: I'm not 100% sure we want this as a standard property but it feels like it might make sense. @ryantxu @torkelo, what's your opinion on that?

torkelo: yea, not sure. Maybe not something all panels get but still standard, and can be opted into? Tricky part is should it affect reduced values in legend etc? Or for that should we make something like this but a transformation instead? This is could also be done as a transformation, but is a common problem and feels cumbersome to have to add a transform for this

mckn: @torkelo I was thinking about the possibility to have a transformer for this instead. But feels like it is so many steps (clicks) to get the expected behavior.

I don't have any strong opinions on this so point me in the right direction and I will apply the proper changes.

dprokop: So adding this as an std property has also a consequence of this null value mode being available for all React panels. Currently, we only have a way to explicitly say use all or only specified properties. There is no way to exclude some of the properties, which might be usefull in this case, so that 

I suggest having the null value mode as a custom property as it was before these changes.

torkelo: Interesting, nullValueMode is actually used in the fieldReducer, https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/transformations/fieldReducer.ts#L87  

But in many of the places it's used it's before field overrides has been applied so it wont have access to the final nullValueMode. hm..

leeoniya: > Or for that should we make something like this but a transformation instead? This is could also be done as a transformation, but is a common problem and feels cumbersome to have to add a transform for this

from my very lacking understanding of transforms, how would this work?

if the graph has a single series, then you can probably use a transform to remove the nulls, which will work as "connected" or coerce to 0s for "as zeros" behavior, or leave as-is for "as gaps".

for graphs with multiple series, only the "as zeros" coercion and "as gaps" (do nothing) transforms will work, but "connected" cannot be attained by a remove-nulls transform as in the single series case, since all series must be x-aligned and the same length.

i think the only meaningful transform here that will work without caveats is nulls -> zeros?

ryantxu: for uPlot, supporting "connected" mode will be super tricky.  that will take some thought, but I don't think can happen easily.

For reference, In plotly studio, they show options like this:
image

This is their UI representation for:
https://plotly.com/javascript/reference/scattergl/#scattergl-line-shape

leeoniya: > for uPlot, supporting "connected" mode will be super tricky

as long as alignment is maintained across all series (data frames are null-filled as needed), then it's already well-handled. the trickiness just migrates into the data alignment logic, no? 🤔

torkelo: But how would we solve connected in that case? Connecting one series would lead removing valid data points for another?

leeoniya: does this help?

https://jsfiddle.net/qa9bsxvf/

dprokop: Just to be totally clear, I think connecting points work great out of the box in uPlot. We had that already working well until I have refactored the config to use geometries, and missed reimplementing it. 

Ref plotly studio and line rendering @ryantxu - I think series paths rendering which can be relatively easily modified in uPlot (ref https://leeoniya.github.io/uPlot/demos/line-stepped.html, https://leeoniya.github.io/uPlot/demos/line-smoothing.html)

leeoniya: @dprokop 

as a bonus, https://github.com/leeoniya/uPlot/commit/1c634ea55312b132addd7f3674fd8f03305080cc exposes `self.paths` which makes it possible to dynamically re-use the internal path builder. this can be useful if you want to render stepped paths when zoomed in, but fall back to the default line path builder at lower zoom levels when the higher data density makes stepped rendering moot.

the stepped example now includes a demo of how to do this: https://leeoniya.github.io/uPlot/demos/line-stepped.html

mckn:
I will close this due too inactivity

leeoniya:
more specifically, this is superseded by https://github.com/grafana/grafana/pull/29512

mckn:
> more specifically, this is superseded by #29512

Ah, cool - I didn't know about that one.

PR #28126, opened 2020-10-08
AzureMonitor: Pass through null values instead of setting 0

### LINKED ISSUE #28001: Azure Monitor Metrics sometimes go to 0
**What happened**:
Azure Monitor Metrics sometimes have no value, indicated by the dotted lines in the Azure Portal:

Grafana returns this as 0 which results in dips on the graph:

Grafana should ignore the empty values and return null in the json response so the values can be ignored. 

Add an Azure Monitor Data Source and add a Storage Account Dashboard with the Used Capacity Metric.

-

**Environment**:
- Grafana version: v7.0.6 (9f4a172a76)
- Data source type & version: Azure Monitor / ?
- User OS & Browser: macOS 10.15.7 & Google Chrome 85
- Grafana plugins: -
- Others: -

kylebrandt:
For the Azure Monitor service in the Azure Monitor datasource, if values were null the backend set them to zero. They will now be null instead.

Before:
image

After:
image

(One could still use "Null point mode: null as zero" if one wants to make it look like it did before)

Fixes #28001

### DIFF
--- a/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
+++ b/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
@@ -288,7 +288,7 @@ func (e *AzureMonitorDatasource) parseResponse(queryRes *tsdb.QueryResult, amr A
 			labels[md.Name.LocalizedValue] = md.Value
 		}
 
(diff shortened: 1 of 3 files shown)


marefr (approved):
LGTM

al-lac (approved):

PR #28127, opened 2020-10-08
Update configure-docker.md

nvtkaszpir:
Bumpt container image to 7.2.x

Docs update.

None, just noticed it doing my own private docker-compose and just got triggered in the October!

Fixes #
None.

Not tested in given setup, but it realllllyyy works on my machine :D

### DIFF
--- a/docs/sources/administration/configure-docker.md
+++ b/docs/sources/administration/configure-docker.md
@@ -38,7 +38,7 @@ mkdir data # creates a folder for your data
 ID=$(id -u) # saves your user id in the ID variable
 
 # starts grafana with your user id and using the data folder
-docker run -d --user $ID --volume "$PWD/data:/var/lib/grafana" -p 3000:3000 grafana/grafana:5.1.0
+docker run -d --user $ID --volume "$PWD/data:/var/lib/grafana" -p 3000:3000 grafana/grafana:7.2.1
 ```
 
 ## Default paths


docs/sources/administration/configure-docker.md
```diff
@@ -38,7 +38,7 @@ mkdir data # creates a folder for your data
 ID=$(id -u) # saves your user id in the ID variable
 
 # starts grafana with your user id and using the data folder
-docker run -d --user $ID --volume "$PWD/data:/var/lib/grafana" -p 3000:3000 grafana/grafana:5.1.0
+docker run -d --user $ID --volume "$PWD/data:/var/lib/grafana" -p 3000:3000 grafana/grafana:7.2.1
```
zoltanbedi, line 41:
```suggestion
docker run -d --user $ID --volume "$PWD/data:/var/lib/grafana" -p 3000:3000 grafana/grafana
```
Would it make sense to remove the version from here?

nvtkaszpir: that would force people to use 'latest' version, which is not really that safe, It's better to run specific version of the container.

nvtkaszpir:
```  grafana:
    image: grafana/grafana:7.2.1
    container_name: grafana
    ports:
      - 3000:3000
    volumes:
      - ./grafana/config:/etc/grafana:ro
      - ./grafana/data:/var/lib/grafana:rw
```

Just tested wit above config with `grafana.ini` being an empty file, works :D

zoltanbedi (approved):
lgtm

oddlittlebird:
Thanks for contributing to the docs!

Issue #26836, opened 2020-08-06
Investigate storing dashboard JSON base64 encoded or as a blob


aknuds1:
**What happened**:
We store dashboard JSON in the database as un-encoded text, which is troublesome since encoding issues can occur (e.g. because the JSON isn't purely encoded in UTF-8), leading to data loss/corruption. To avoid such issues, and ease troubleshooting, we should store dashboard JSON base64 encoded instead. To support existing databases, we should add a column with base64 encoded content and support also databases lacking content in this column.

As an alternative we could also consider storing JSON in a blob type column instead, to avoid the whole text encoding problem.

We also need to figure out if there's any need to query the dashboard JSON directly as text.

Furthermore, I think we should test that the JSON is valid before storing it in the database, so we can be certain that what gets stored can also be deserialized. Could it be an option to store BSON (MongoDB uses this), or some other binary format for space efficiency and ensuring validity?

bergquist:
The Blob idea sounds interesting. I think that should be backward compatible and make it easier to extract all dashboards. We should definitely test that before we make any decision.

ying-jeanne:
For now i am going to close the issue, since the design was there since a while, there is no clear evidence that it causes a lot of issue. If it causes any problem in the future, please attach the use case and feel free to reopen the issue. :) glad to do some improvement after the use cases.

PR #28129, opened 2020-10-08
Snapshots: Store dashboard data encrypted in the database

### LINKED ISSUE #28065: Encrypt the dashboard data inside dashboard_snapshots
We would like the data inside dashboard_snapshots (the dashboard column) to be encrypted.
Maybe the same way secure json is encrypted inside the datasource table

**Why is this needed**:
This column could potentially contain private data

wbrowne:
Adds an encrypted column for dashboard JSON data for the dashboard snapshot entity. The change now means that if encrypted data already exists (data found in the new `dashboard_encrypted` column), we will prefer that as the source of truth over the pre-existing `dashboard` column. If no data exists there, we will return whatever is stored in the `dashboard` column. 

When we create new dashboard snapshots, we will store the encrypted data in `dashboard_encrypted` and an empty JSON payload (`{}`) in the `dashboard` column - as it is non-NULLABLE (at least for now)

Fixes #28065

This does not cover the migration from unencrypted to encrypted data - that will be done as a follow up


pkg/services/sqlstore/dashboard_snapshot.go
```diff
@@ -45,22 +49,32 @@ func CreateDashboardSnapshot(cmd *models.CreateDashboardSnapshotCommand) error {
...
-			OrgId:             cmd.OrgId,
-			UserId:            cmd.UserId,
-			External:          cmd.External,
-			ExternalUrl:       cmd.ExternalUrl,
-			ExternalDeleteUrl: cmd.ExternalDeleteUrl,
-			Dashboard:         cmd.Dashboard,
-			Expires:           expires,
-			Created:           time.Now(),
-			Updated:           time.Now(),
+		encode, err := cmd.Dashboard.Encode()
```
wbrowne, line 52: As I was making these changes it made me think of https://github.com/grafana/grafana/issues/26836 as we had been discussing it only recently. I'm not certain yet but after a brief look I think this will be base64 encoded and since we're already now storing it as a `blob`, we may be good 😄

pkg/api/dashboard_snapshot_test.go
```diff
@@ -198,6 +201,61 @@ func TestDashboardSnapshotApiEndpoint(t *testing.T) {
 					So(sc.resp.Code, ShouldEqual, 500)
 				})
 			})
+
+			Convey("Should be able to read a snapshot's un-encrypted data", func() {
```
marefr, line 205: This is a FYI if you didn't know or anyone told you. We want to avoid introducing new GoConvey tests and use the Go standard test library, for info [here](https://github.com/grafana/grafana/blob/master/contribute/style-guides/backend.md#testing).

For these tests that uses `loggedInUserScenarioWithRole` they're very tied to GoConvey so hard to not introduce new ones and re-writing the existing is probably quite a bit of work. But good to be aware of these things moving forward.

pkg/services/sqlstore/dashboard_snapshot.go
```diff
@@ -45,22 +49,32 @@ func CreateDashboardSnapshot(cmd *models.CreateDashboardSnapshotCommand) error {
...
+		if err != nil {
+			return err
 		}
 
-		_, err := sess.Insert(snapshot)
+		encryptedDashboard, err := util.Encrypt(encode, setting.SecretKey)
```
marefr, line 57: Right now the logic for encrypt/decrypt is spread out between sqlstore and api package. I would suggest introducing a new package `securedata` and related struct for holding the encrypted data  and provide methods for encrypt and decrypt. This would be very similar to the `securejsondata` package. 

Example usage of the securejsondata can be seen for datasources: 
https://github.com/grafana/grafana/blob/90303b5a03215f5dbcd3215a182f3251e0f0a53c/pkg/services/sqlstore/datasource.go#L140
https://github.com/grafana/grafana/blob/90303b5a03215f5dbcd3215a182f3251e0f0a53c/pkg/models/datasource_cache.go#L251

WDYT?

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -15,6 +18,12 @@ func TestDashboardSnapshotDBAccess(t *testing.T) {
 	Convey("Testing DashboardSnapshot data access", t, func() {
 		InitTestDB(t)
 
+		origSecret := setting.SecretKey
```
marefr, line 21: Related to other comment regarding GoConvey. These tests should be quite easy to convert to standard Go tests I think. However, this can be done as a separate follow-up PR if you want? 

Here's a good example of how we normally write standard tests using testify for assert/require: https://github.com/grafana/grafana/blob/master/pkg/infra/serverlock/serverlock_test.go

Some general testify 
- The `require` package you want to use if you want the test runner to fail directly (most common)
- The `assert` package` you can use if you have multiple asserts that doesn't depend on each other

marefr (changes requested):
Great work. Looks good in general, but added some comments/suggestions of improvements.

pkg/services/sqlstore/dashboard_snapshot.go
marefr: Thought I added a comment here. Well, here it goes again. Happened to notice the other day that we do encrypt and encode for OAuth tokens that we store in the database. But maybe that's due to the database type DB_Text used?

https://github.com/grafana/grafana/blob/90303b5a03215f5dbcd3215a182f3251e0f0a53c/pkg/services/sqlstore/user_auth.go#L275-L283

wbrowne: Hmm interesting.. Maybe it's better to keep them consistent then. Will explore that option and see how it turns out 😄

pkg/services/sqlstore/dashboard_snapshot.go
wbrowne: I like it 👍

pkg/api/dashboard_snapshot_test.go
```diff
@@ -198,6 +201,61 @@ func TestDashboardSnapshotApiEndpoint(t *testing.T) {
...
+
+			Convey("Should be able to read a snapshot's encrypted data", func() {
+				origSecret := setting.SecretKey
+				setting.SecretKey = "dashboard_snapshot_api_test"
+				t.Cleanup(func() {
+					setting.SecretKey = origSecret
+				})
+
+				dashboardId := 123
+				jsonModel, _ := simplejson.NewJson([]byte(fmt.Sprintf(`{"id":%d}`, dashboardId)))
```
aknuds1, line 229:
```suggestion
				jsonModel, err := simplejson.NewJson([]byte(fmt.Sprintf(`{"id":%d}`, dashboardId)))
				So(err, ShouldBeNil)
```

pkg/services/sqlstore/dashboard_snapshot.go
aknuds1: Are we solving #26836 with this PR?

aknuds1 (approved):
Looks good AFAICT.

pkg/services/sqlstore/dashboard_snapshot.go
wbrowne: No not directly. In this case we're just focussing on the dashboards within `dashboard _snapshot`s, but now is a good opportunity to do the right thing and follow up later with the fix for #26836

aknuds1: @wbrowne Ah... That difference was lost on me.

pkg/api/dashboard_snapshot_test.go
wbrowne: Will take care of the `sql` tests in this PR but I reckon will follow up in the future for this one for the reasons you mentioned

pkg/components/securedata/securedata.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"encoding/base64"
+
+	"github.com/grafana/grafana/pkg/setting"
+	"github.com/grafana/grafana/pkg/util"
+)
+
+type SecureData string
+
+func EncryptAndEncode(data []byte) (SecureData, error) {
```
wbrowne, line 12: Was thinking instead to have these functions as `Encrypt` and `Decrypt` and rename `SecureData` to `SecureEncodedData`. Opinions?

pkg/components/securedata/securedata.go
```diff
@@ -0,0 +1,26 @@
...
+	encrypted, err := util.Encrypt(data, setting.SecretKey)
+	if err != nil {
+		return "", err
+	}
+	return SecureData(base64.StdEncoding.EncodeToString(encrypted)), nil
+}
+
+func (s SecureData) DecodeAndDecrypt() ([]byte, error) {
+	decoded, err := base64.StdEncoding.DecodeString(string(s))
```
aknuds1, line 21: Might as well avoid the encoding to string:
```suggestion
	decoded, err := base64.StdEncoding.Decode(s)
```

pkg/components/securedata/securedata.go
```diff
@@ -0,0 +1,26 @@
+package securedata
+
+import (
+	"encoding/base64"
+
+	"github.com/grafana/grafana/pkg/setting"
+	"github.com/grafana/grafana/pkg/util"
+)
+
+type SecureData string
```
aknuds1, line 10: Might it be more efficient to make this `[]byte` instead, to avoid having to convert to string?

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
 
 	"github.com/grafana/grafana/pkg/components/simplejson"
 	"github.com/grafana/grafana/pkg/models"
 	"github.com/grafana/grafana/pkg/setting"
 )
 
 func TestDashboardSnapshotDBAccess(t *testing.T) {
-	Convey("Testing DashboardSnapshot data access", t, func() {
-		InitTestDB(t)
+	t.Helper()
```
aknuds1, line 17: Are you sure we should be calling this here?? What good does it do?

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
+			}),
+			UserId: 1000,
+			OrgId:  1,
+		}
+		err := CreateDashboardSnapshot(&cmd)
+		require.NoError(t, err)
+
+		t.Run("Should be able to get snapshot by key", func(t *testing.T) {
+			query := models.GetDashboardSnapshotQuery{Key: "hej"}
+			err = GetDashboardSnapshot(&query)
```
aknuds1, line 41:
```suggestion
			err := GetDashboardSnapshot(&query)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
+			require.NoError(t, err)
+
+			assert.NotNil(t, query.Result)
+
+			dashboard, err := query.Result.DashboardJSON()
+			require.NoError(t, err)
+			assert.Equal(t, dashboard.Get("hello").MustString(), "mupp")
```
aknuds1, line 48: assert.Equal takes the wanted value before the actual one.
```suggestion
			assert.Equal(t, "mupp", dashboard.Get("hello").MustString())
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
+			require.NoError(t, err)
 
-				So(query.Result, ShouldNotBeNil)
+			t.Run("Should return all the snapshots", func(t *testing.T) {
 
-				dashboard, err := query.Result.DashboardJSON()
-				So(err, ShouldBeNil)
-				So(dashboard.Get("hello").MustString(), ShouldEqual, "mupp")
+				assert.NotNil(t, query.Result)
+				assert.Equal(t, len(query.Result), 1)
```
aknuds1, line 62:
```suggestion
				assert.Len(t, query.Result, 1)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
-					}
-					err := SearchDashboardSnapshots(&query)
-					So(err, ShouldBeNil)
-
-					So(query.Result, ShouldNotBeNil)
-					So(len(query.Result), ShouldEqual, 1)
-				})
+			t.Run("Should return all the snapshots", func(t *testing.T) {
+				assert.NotNil(t, query.Result)
+				assert.Equal(t, len(query.Result), 1)
```
aknuds1, line 76:
```suggestion
				assert.Len(t, query.Result, 1)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
-						SignedInUser: &models.SignedInUser{OrgRole: models.ROLE_EDITOR, UserId: 1000},
-					}
-					err := SearchDashboardSnapshots(&query)
-					So(err, ShouldBeNil)
-
-					So(query.Result, ShouldNotBeNil)
-					So(len(query.Result), ShouldEqual, 1)
-				})
+			t.Run("Should return all the snapshots", func(t *testing.T) {
+				assert.NotNil(t, query.Result)
```
aknuds1, line 75:
```suggestion
				require.NotNil(t, query.Result)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
-					}
-					err := SearchDashboardSnapshots(&query)
-					So(err, ShouldBeNil)
-
-					So(query.Result, ShouldNotBeNil)
-					So(len(query.Result), ShouldEqual, 0)
-				})
+			t.Run("Should not return any snapshots", func(t *testing.T) {
+				assert.NotNil(t, query.Result)
+				assert.Equal(t, len(query.Result), 0)
```
aknuds1, line 90:
```suggestion
				require.NotNil(t, query.Result)
				assert.Empty(t, query.Result)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
-					So(err, ShouldBeNil)
-
-					So(query.Result, ShouldNotBeNil)
-					So(len(query.Result), ShouldEqual, 0)
-				})
+				err := SearchDashboardSnapshots(&query)
+				require.NoError(t, err)
+
+				assert.NotNil(t, query.Result)
+				assert.Equal(t, len(query.Result), 0)
```
aknuds1, line 116:
```suggestion
				require.NotNIl(t, query.Result)
				assert.Empty(t, query.Result)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
+			require.NoError(t, err)
 
-				decrypted, err := util.Decrypt(cmd.Result.DashboardEncrypted, setting.SecretKey)
-				So(err, ShouldBeNil)
+			decrypted, err := cmd.Result.DashboardEncrypted.DecodeAndDecrypt()
+			require.NoError(t, err)
 
-				So(bytes.Equal(decrypted, original), ShouldEqual, true)
-			})
+			assert.Equal(t, bytes.Equal(decrypted, original), true)
```
aknuds1, line 127: Shouldn't this also work?
```suggestion
			assert.Equal(t, original, decrypted, original)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
 			SignedInUser: &models.SignedInUser{OrgRole: models.ROLE_ADMIN},
 		}
 		err = SearchDashboardSnapshots(&query)
-		So(err, ShouldBeNil)
+		require.NoError(t, err)
 
-		So(len(query.Result), ShouldEqual, 1)
-		So(query.Result[0].Key, ShouldEqual, notExpiredsnapshot.Key)
+		assert.Equal(t, len(query.Result), 1)
+		assert.Equal(t, query.Result[0].Key, nonExpiredSnapshot.Key)
```
aknuds1, line 153:
```suggestion
		assert.Len(t, query.Result, 1)
		assert.Equal(t, nonExpiredSnapshot.Key, query.Result[0].Key)
```

pkg/services/sqlstore/dashboard_snapshot_test.go
```diff
@@ -5,169 +5,169 @@ import (
...
```
aknuds1, line 166:
```suggestion
		assert.Len(t, query.Result, 1)
		assert.Equal(t, nonExpiredSnapshot.Key, query.Result[0].Key)
```

pkg/services/sqlstore/migrations/dashboard_snapshot_mig.go
```diff
@@ -66,6 +66,6 @@ func addDashboardSnapshotMigrations(mg *Migrator) {
 	}))
 
 	mg.AddMigration("Add encrypted dashboard json column", NewAddColumnMigration(snapshotV5, &Column{
-		Name: "dashboard_encrypted", Type: DB_Blob, Nullable: true,
+		Name: "dashboard_encrypted", Type: DB_Text, Nullable: true,
```
aknuds1, line 69: Why was it had to be a text column exactly?

aknuds1 (approved):
Generally looks good AFAICT (and thanks for converting the tests!), but left suggestions for improvement plus a question or two.

pkg/services/sqlstore/migrations/dashboard_snapshot_mig.go
wbrowne: It hasn't be fully decided yet - I wanted to experiment with it as a text field but I think considering the information I've found it might be best suitable as a blob - as it's not directly JSON but encrypted JSON.

According to official MySQL docs at least:
>Many encryption and compression functions return strings for which the result might contain arbitrary byte values. If you want to store these results, use a column with a VARBINARY or BLOB binary string data type. This avoids potential problems with trailing space removal or character set conversion that would change data values, such as may occur if you use a nonbinary string data type (CHAR, VARCHAR, TEXT).

https://dev.mysql.com/doc/refman/8.0/en/encryption-functions.html

aknuds1: Thanks for clarifying. I don't see any upside to storing as TEXT even if it's JSON though? It's so easy to get encoding errors, which is why I made that issue.

marefr (approved):
Great work. Much cleaner using blob :+1: LGTM

PR #28135, opened 2020-10-09
Fix: Dynamically add ShowWhen fields to the form watch

### LINKED ISSUE #28132: Notifier ShowWhen struct not working as expected
**What happened**:
I am creating a new notifier and want to provide a way for users to select an Auth Provider, either a Credential Profile, AWS ARN, or Access Key & Secret. These same options are available for customers setting up the CloudWatch DataSource today. 

In the code for the notifier, I used the [ShowWhen](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/notifier.go#L80-L84) struct to display the relevant auth fields,  based on the Auth Provider the user selects. 

The fields displayed using `ShowWhen` only appear when the other required fields (Notification Channel Name and SNS Topic) are *not* already filled in. Once I remove either of these fields the ShowHide elements appear on the page. 

Further, if fill in the Topic and name, and change the Auth Provider selection, the `ShowWhen` elements are not refreshed. If I remove either the Name or the Topic and change the Auth Provider selection the page does automatically refresh the boxes.

The `ShowWhen` elements for the Pushover notifier channel do not have either of these issues. [[link](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/notifiers/pushover.go#L153-L174)]

**Environment**:
- Grafana version: 7.3.0
- OS Grafana is installed on: mac
- User OS & Browser: 
   * macOS High Sierra 10.13.6
Reproduced the issue in the following browsers:
   * Safari 13.1.2
   * Chrome 86.0.4240.75
   * Firefox Quantum 68.11.0esr 
   
- Grafana plugins: n/a
- Others:

peterholmberg:
Dynamically add fields to be watched by the Form component.

Fixes #28132

### DIFF
--- a/public/app/features/alerting/components/NotificationChannelForm.tsx
+++ b/public/app/features/alerting/components/NotificationChannelForm.tsx
@@ -36,9 +36,22 @@ export const NotificationChannelForm: FC<Props> = ({
 }) => {
   const styles = getStyles(useTheme());
 
+  /*
+   Finds fields that have dependencies on other fields and removes duplicates.
+   Needs to be prefixed with settings.
+  */
+  const fieldsToWatch =
+    new Set(
+      selectedChannel?.options
+        .filter(o => o.showWhen.field)
+        .map(option => {
+          return `settings.${option.showWhen.field}`;
+        })
+    ) || [];
+
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


torkelo (approved):
LGTM

PR #28136, opened 2020-10-09
Add docs for IconButton

jessover9000:
**Implementation** (Checklist meant for the person implementing the component)
I lack the technical knowledge to check some of these, @peterholmberg or @Clarity-89 could you look at this?

- [x] Component has a story in Storybook.
- [x] Props and naming follows [our style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/frontend.md).
- [x] It is extendable (rest props are spread, styles with className work, and so on).
- [x] Uses [theme for spacing, colors, and so on](https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md).
- [x] Works with both light and dark theme. – **I think it does but it's really hard to check in storybook because the background doesn't switch with the docs theming switcher**

**Documentation**

- [x] Properties are documented.
- [x] Use cases are described.
- [x] Code examples for the different use cases.
- [x] Dos and don'ts.
- [x] Styling guidelines, specific color usage (if applicable).

### DIFF
--- a/packages/grafana-ui/src/components/IconButton/IconButton.story.tsx
+++ b/packages/grafana-ui/src/components/IconButton/IconButton.story.tsx
@@ -12,7 +12,9 @@ export default {
   component: IconButton,
   decorators: [withCenteredStory],
   parameters: {
-    docs: {},
+    docs: {
+      page: mdx,
+    },
   },
 };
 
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/IconButton/IconButton.story.tsx
```diff
@@ -5,6 +5,7 @@ import { withCenteredStory } from '../../utils/storybook/withCenteredStory';
 import { useTheme } from '../../themes/ThemeContext';
 import { GrafanaTheme } from '@grafana/data';
 import { IconSize, IconName } from '../../types';
+import mdx from './IconButton.mdx';
```
Clarity-89, line 8: You still need to enable the mdx by adding `page: mdx,` inside the docs on line 15.

Clarity-89 (approved):

PR #28155, opened 2020-10-09
Docs: Add view settings and view stats

oddlittlebird:
Starting to fill gaps in our Administration documentation. This adds instructions for the Stats and Settings tab in the Server Admin menu.

### DIFF
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -17,6 +17,8 @@ Grafana has a number of configuration options that you can specify in a `.ini` c
 
 > **Note:** You must restart Grafana for any configuration changes to take effect.
 
+To see all settings currently applied to the Grafana server, refer to [View server settings]({{< relref "view-server-settings.md" >}}).
+
 ## Config file locations
 
 _Do not_ change `defaults.ini`! Grafana defaults are stored in this file. Depending on your OS, make all configuration changes in either `custom.ini` or `grafana.ini`.
@@ -89,9 +91,7 @@ export GF_PLUGIN_GRAFANA_IMAGE_RENDERER_RENDERING_IGNORE_HTTPS_ERRORS=true
 
 ## Variable expansion
 
-> Only available in Grafana 7.1+.
-
-> For any changes to `conf/grafana.ini` (or corresponding environment variables) to take effect, you must restart Grafana.
+> **Note:** Only available in Grafana 7.1+.
 
 If any of your options contains the expression `$__<provider>{<argument>}`
 or `${<environment variable>}`, then they will be processed by Grafana's
--- a/docs/sources/administration/view-server-settings.md
+++ b/docs/sources/administration/view-server-settings.md
@@ -1,0 +1,26 @@
++++
+title = "View server settings"
+description = "How to view server settings in the Grafana UI"
+keywords = ["grafana", "configuration", "server", "settings"]
+type = "docs"
+[menu.docs]
+name = "Configuration"
+identifier = "config"
+parent = "admin"
+weight = 350
++++
+
+# View Grafana server settings
+
+If you are a Grafana server admin, then you can use the Settings tab to view the settings applied to your Grafana server by the [Configuration]({{< relref "configuration.md" >}}) file and any environmental variables.
+
+> **Note:** Only Grafana server admins can access the **Server Admin** menu. For more information about about admin permissions, refer to [Grafana server admin]({{< relref "../permissions/_index.md" >}}).
+
+## View server settings
+
+1. Log in to your Grafana server with an account that has the Grafana Admin flag set.
+1. Hover your cursor over the **Server Admin** (shield) icon in the side menu and then click the **Settings** tab.
+
+## Available settings
+
+For a full list of server settings, refer to [Configuration]({{< relref "configuration.md" >}}).
--- a/docs/sources/administration/view-server-settings.md
+++ b/docs/sources/administration/view-server-settings.md
@@ -12,7 +12,7 @@ weight = 350
 
 # View Grafana server settings
 
-If you are a Grafana server admin, then you can use the Settings tab to view the settings applied to your Grafana server by the [Configuration]({{< relref "configuration.md" >}}) file and any environmental variables.
+If you are a Grafana server admin, then you can use the Settings tab to view the settings applied to your Grafana server by the [Configuration]({{< relref "configuration.md#config-file-locations" >}}) file and any environmental variables.
 
 > **Note:** Only Grafana server admins can access the **Server Admin** menu. For more information about about admin permissions, refer to [Grafana server admin]({{< relref "../permissions/_index.md" >}}).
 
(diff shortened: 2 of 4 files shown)


docs/sources/administration/configuration.md
```diff
@@ -89,9 +91,7 @@ export GF_PLUGIN_GRAFANA_IMAGE_RENDERER_RENDERING_IGNORE_HTTPS_ERRORS=true
 
 ## Variable expansion
 
-> Only available in Grafana 7.1+.
-
-> For any changes to `conf/grafana.ini` (or corresponding environment variables) to take effect, you must restart Grafana.
```
papagian, line 94: I think that restarting grafana for the new settings to be effective is important information and it should stay unless it has been moved somewhere else in the docs.

docs/sources/administration/view-server-settings.md
```diff
@@ -0,0 +1,26 @@
...
+[menu.docs]
+name = "Configuration"
+identifier = "config"
+parent = "admin"
+weight = 350
++++
+
+# View Grafana server settings
+
+If you are a Grafana server admin, then you can use the Settings tab to view the settings applied to your Grafana server by the [Configuration]({{< relref "configuration.md" >}}) file and any environmental variables.
```
papagian, line 15:
```suggestion
If you are a Grafana server admin, then you can use the Settings tab to view the settings applied to your Grafana server by the [Configuration]({{< relref "configuration.md#config-file-locations" >}}) file and any environmental variables.
```

docs/sources/administration/view-server-settings.md
```diff
@@ -0,0 +1,26 @@
...
+
+> **Note:** Only Grafana server admins can access the **Server Admin** menu. For more information about about admin permissions, refer to [Grafana server admin]({{< relref "../permissions/_index.md" >}}).
+
+## View server settings
+
+1. Log in to your Grafana server with an account that has the Grafana Admin flag set.
```
papagian, line 21: I wasn't aware that there was such a flag! I think that this is only visible in `User Information` tab under `Server Admin` which is accessible only be administrators. So for non administrators this flag is practically invisible.
I would suggest changing to something like: `Log in to your Grafana server with an account with administrator permissions`.

docs/sources/administration/view-server-stats.md
```diff
@@ -0,0 +1,39 @@
...
+
+# View Grafana server stats
+
+If you are a Grafana server admin, then you can view useful statistics about your Grafana server in the Stats tab.
+
+> **Note:** Only Grafana server admins can access the **Server Admin** menu. For more information about about admin permissions, refer to [Grafana server admin]({{< relref "../permissions/_index.md" >}}).
+
+## View server stats
+
+1. Log in to your Grafana server with an account that has the Grafana Admin flag set.
```
papagian, line 17: I would suggest changing to something like: `Log in to your Grafana server with an account with administrator permissions`.
See my respective comment for the server settings.

docs/sources/administration/view-server-stats.md
```diff
@@ -0,0 +1,39 @@
...
+1. Hover your cursor over the **Server Admin** (shield) icon in the side menu and then click the **Stats** tab.
+
+## Available stats
+
+The following statistics are displayed in the Stats tab:
```
papagian, line 22: It's worth mentioning here somehow that if users belong to several organisations then each user is counted once (for the highest role in any org) so as Total users = admins + editors + viewers. Same for active users.
For more details you can check the respective [PR](https://github.com/grafana/grafana/pull/26777) and [issue](https://github.com/grafana/grafana/issues/26744).

papagian (changes requested):
LGTM; I have some suggestions!

docs/sources/administration/configuration.md
oddlittlebird: Yes, which is why we already mention it prominently on line 18 in this same document.

docs/sources/administration/view-server-settings.md
oddlittlebird: Nope, because we have two kinds of administrators: organization administrators and server administrators. And this is specific to Grafana server admins.

docs/sources/administration/view-server-stats.md
oddlittlebird: Same as above. We have two administrator types (three if you're talking about Cloud), so we have to specify which admin type we mean here.

docs/sources/administration/view-server-stats.md
oddlittlebird: I did not know that! Please take a look at my example and let me know if it makes sense.

For active users, do you need to log in under the particular organization in order to count as a user? For example, if you didn't log in to two of your four organizations, would you not count as an active user for those organizations?

papagian: AFAIK active users are considered all the users that have signed in [the last month](https://github.com/grafana/grafana/blob/0bfdf79968e926aa436dd29ce17e582d06f93643/pkg/services/sqlstore/stats.go#L22). They are counted once regardless on how many organisations belong. Their role is the highest on all the organisations.

So for your example, assuming that Sofia is the only user then it should be:

- Total users     1
- Total admins    1 (the highest role of Sofia is Admin)
- Total editors   0
- Total viewers   0

@sakjur can you confirm that?

oddlittlebird: Thanks, I clarified the text.

sakjur: Yes! And these stats are global for the instance, not per org.

docs/sources/administration/view-server-stats.md
oddlittlebird: Correction: We have four administrator types if you count Cloud.

docs/sources/administration/view-server-settings.md
oddlittlebird: Correction: Three kinds of administrators: organization admins, server admins, and team admins.

docs/sources/administration/view-server-stats.md
```diff
@@ -0,0 +1,50 @@
...
+- Total snapshots
+- Total dashboard tags
+- Total starred dashboards
+- Total alerts
+
+## Counting users
+
+If a user belongs to several organizations, then that user is counted once for each organization as a user in the highest organization role they are assigned.
+
+For example, if Sofia is a Viewer in two organizations, an Editor in two organizations, and and Admin in three organizations, then she would be reflected in the stats as:
```
papagian, line 47:
```suggestion
For example, if Sofia is a Viewer in two organizations, an Editor in two organizations, and Admin in three organizations, then she would be reflected in the stats as:
```

papagian (approved):
LGTM!

docs/sources/administration/view-server-settings.md
sakjur: In addition to those three, you can also have admin rights over a folder.

sakjur (approved):
:+1:

docs/sources/administration/view-server-settings.md
oddlittlebird: Yes, but that is admin permissions, not an admin role.

PR #28159, opened 2020-10-10
Dependabot: Check actions daily

jlosito:
**What this PR does / why we need it**: This PR allows dependabot to submit PRs with version bumps to the GitHub actions which this project uses. I don't think it's necessarily needed but rather nice to have.

**Which issue(s) this PR fixes**: There aren't any issues which this PR fixes. I don't believe there's anything necessarily wrong.

**Special notes for your reviewer**: Here's the link to the docs. https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


xlson:
I don't feel like I understand the implications of this PR well enough to offer any kind of review. If you want we can chat about it and you can explain what the considerations are?

torkelo:
Yea, will this have any effect on the current packages the dependabot does cover ?

jlosito:
I believe dependabot sends updates when there's security concerns but doesn't necessarily notify when there's package updates. This PR would send updates when there's version changes, for example action/checkout@v1 to action/checkout@v2, to any GitHub actions which this project uses only. To put it differently, this PR would not send updates if there were version changes upstream to any of Go modules that this project uses.

Below is an example of a PR from dependabot which this code change would cause.

https://github.com/ubuntu/microk8s/pull/1615

aknuds1:
> I believe dependabot sends updates when there's security concerns but doesn't necessarily notify when there's package updates. This PR would send updates when there's version changes, for example action/checkout@v1 to action/checkout@v2, to any GitHub actions which this project uses only. To put it differently, this PR would not send updates if there were version changes upstream to any of Go modules that this project uses.
> 
> Below is an example of a PR from dependabot which this code change would cause.
> 
> [ubuntu/microk8s#1615](https://github.com/ubuntu/microk8s/pull/1615)

@jlosito Can you please answer the following, I can't see from the above that you do:

> Yea, will this have any effect on the current packages the dependabot does cover ?

jlosito:
The following are some actions that are used within this project.
- checkout@v2
- checkout@v1
- setup-node@v1

I believe the v1 instance of the checkout action will try to be upgraded to v2. The same goes for setup-node. I see that there exists a v2 for setup-node.

Yes, I believe this PR will have an effect on the actions that this project currently uses. 

Note that this PR only allows dependabot to submit PRs for actions that are out of date such as checkout@v1, but the PR from dependabot could always be denied.

aknuds1:
@jlosito You're not answering though if your configuration of dependabot affects which packages it will attempt to upgrade, you are only referring to GitHub Actions. Take for example [this PR](https://github.com/grafana/grafana/pull/27507), which upgrades an NPM package.

jlosito:
I'm a little confused by the question. This PR only configures dependabot to check actions, not any of the npm packages. Whatever settings were already set for npm packages will continue to work as they were. Nothing is being changed for npm packages or any other packages other than actions.

aknuds1:
@jlosito Thanks, that's the sort of clarification I wanted at least. It's not obvious to someone not familiar with Dependabot configuration what is affected by the configuration you've written.

jlosito:
I believe dependabot only submits those types of PRs similair to the one you've linked when there's a good reason such as a security vulnerability. The config change purposed here is not only for security vulnerabilities but any version increase, only for action packages. If this project also wanted to keep each of its npm packages in synced with the latest stable version the change would look something like the following

```
- package-ecosystem: "npm"
  directory: "/"
    schedule:
      interval: "daily"
```

torkelo (approved):

PR #28161, opened 2020-10-10
Added DOT(.) at end of line

ashishagarwal06:
Hi - I have observed Dot(.) is missing in many places of paragraphs in document. So raising this PR.

Added Dot(.) to below line
This page shows how to use regex to filter/modify values in the variable dropdown.

Thanks,
Ashish

Fixes #


achatterjee-grafana (approved):
Good catch! Thank you.

PR #28162, opened 2020-10-10
Alerting: Log warnings for obsolete notifiers when extracting alerts and remove frequent error log messages

### LINKED ISSUE #27755: Alerting: move spammy error message about obsolete notifiers
**What happened**:
[This error](https://github.com/grafana/grafana/commit/1e543d5ac12d7a47310e25768a88fb2ab0cf7820) indicates that an alert's JSON references an obsolete notifier. This can be an issue if this is the only notifier defined for this alert since in that case the alert will not be delivered. Therefore we don't want to completely remove it so users will be notified that there is something wrong with their alert definition. However this error is too spammy since it's produced every time the rule is evaluated. We have to warn the users when the alert is saved instead.

**Environment**:
- Grafana version:
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

papagian:

- ~removes spammy error for obsolete notifiers when evaluating alerts~
- ~logs alert warnings for obsolete notifiers when extracting alerts~
- ~introduces type `alertValidator`~
- ~modifies `getAlertFromPanels` and `extractAlerts` to accept  an arbitrary number of `alertValidator` (of different severity) instead of a single `validateAlertFunc`~

**Updated description**
After the [feedback](https://github.com/grafana/grafana/pull/28162#issuecomment-707337205) I got I have reverted the previews modifications and
- I've introduced `ErrAlertNotificationFailedTranslateUniqueID` error which is returned when no notifier is found for a specific ID
- I've modified `extractAlerts`, `getAlertFromPanels`, `NewRuleFromDBAlert` to get an additional boolean parameter for enabling/disabling logging when the above error is returned
- I've reduced the log message level to a warning

Note that in the code (executed when saving a dashboard), `extractAlerts` is called twice: once from `buildSaveDashboardCommand` and once for `updateAlerting`. Therefore I've parameterised `extractAlerts` so as warnings for each alert are logged just once.
https://github.com/grafana/grafana/blob/a5d9196a5387a7fb05f12142cdcf98b6da23e5af/pkg/services/dashboards/dashboard_service.go#L219-L254

Fixes #27755

For testing it one can edit an alert in the dashboard JSON and add a non-existing notifier like:
```
"notifications": [
          {
            "id": 2046
          }
]
```

With this fix there should not be any logging errors for failed notifier translation to UID when the alert is evaluated
but if the user navigates to the dashboard and tries to save the dashboard
then there should be a warning in the logs.


pkg/services/alerting/extractor.go
```diff
@@ -193,8 +194,30 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
+						validationErrors.WriteString("\n")
+					}
+					validationErrors.WriteString(reason)
+				case alertWarning:
+					if validationWarnings.Len() > 0 {
+						validationErrors.WriteString("\n")
+					}
+					validationWarnings.WriteString(reason)
+				}
+			}
```
aknuds1, line 214: It's generally more readable, and idiomatic, if you reduce nesting:
```suggestion
			if ok {
				continue
			}
			
			switch validator.aSeverity {
			case alertError:
				if validationErrors.Len() > 0 {
					validationErrors.WriteString("\n")
				}
				validationErrors.WriteString(reason)
			case alertWarning:
				if validationWarnings.Len() > 0 {
					validationErrors.WriteString("\n")
				}
				validationWarnings.WriteString(reason)
			}
```

pkg/services/alerting/extractor.go
```diff
@@ -203,16 +226,49 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
 	return alerts, nil
 }
 
-func validateAlertRule(alert *models.Alert) bool {
-	return alert.ValidToSave()
+func validateAlertRule(alert *models.Alert) (ok bool, reason string) {
+	ok = alert.ValidToSave()
+	if !ok {
+		reason = fmt.Sprintf("Panel id is not correct, alertName=%v, panelId=%v", alert.Name, alert.PanelId)
```
aknuds1, line 232: I think it's best to standardize on writing "ID" for identifier (id means [something else](https://en.wikipedia.org/wiki/Id,_ego_and_super-ego)):
```suggestion
		reason = fmt.Sprintf("Panel ID is not correct, alertName=%v, panelId=%v", alert.Name, alert.PanelId)
```

pkg/services/alerting/extractor.go
```diff
@@ -249,6 +305,15 @@ func (e *DashAlertExtractor) extractAlerts(validateFunc func(alert *models.Alert
 // ValidateAlerts validates alerts in the dashboard json but does not require a valid dashboard id
 // in the first validation pass.
 func (e *DashAlertExtractor) ValidateAlerts() error {
-	_, err := e.extractAlerts(func(alert *models.Alert) bool { return alert.OrgId != 0 && alert.PanelId != 0 })
+	_, err := e.extractAlerts(alertValidator{
+		aFunc: func(alert *models.Alert) (ok bool, reason string) {
+			ok = alert.OrgId != 0 && alert.PanelId != 0
+			if !ok {
+				reason = fmt.Sprintf("Panel id is not correct, alertName=%v, panelId=%v", alert.Name, alert.PanelId)
```
aknuds1, line 312:
```suggestion
				reason = fmt.Sprintf("Panel ID is not correct, alertName=%v, panelId=%v", alert.Name, alert.PanelId)
```

aknuds1 (commented):
Leaving some comments for now, will look more closely.

pkg/services/alerting/extractor.go
```diff
@@ -203,16 +226,49 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
+		if id, err := jsonModel.Get("id").Int64(); err == nil {
+			_, err := translateNotificationIDToUID(id, alert.OrgId)
+			if err != nil {
+				ok = false
+				if warnings.Len() > 0 {
+					warnings.WriteString("\n")
+				}
+				warnings.WriteString(fmt.Sprintf("Alert contains notification identified by incorrect id, alertName=%v, panelId=%v, notificationId=%v", alert.Name, alert.PanelId, id))
+			}
+		}
```
aknuds1, line 250: It's best to avoid nesting:
```suggestion
		id, err := jsonModel.Get("id").Int64()
		if err != nil {
			continue
		}
		
		if _, err := translateNotificationIDToUID(id, alert.OrgId); err == nil {
			continue
		}

		ok = false
		if warnings.Len() > 0 {
			warnings.WriteString("\n")
		}
		warnings.WriteString(fmt.Sprintf("Alert contains notification identified by incorrect id, alertName=%v, panelId=%v, notificationId=%v", alert.Name, alert.PanelId, id))	
```

pkg/services/alerting/extractor.go
```diff
@@ -193,8 +194,30 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
+					}
+					validationWarnings.WriteString(reason)
+				}
+			}
+		}
+		if validationErrors.String() != "" {
+			return nil, ValidationError{Reason: validationErrors.String()}
+		}
+		if validationWarnings.String() != "" {
+			e.log.Debug(validationWarnings.String())
```
aknuds1, line 220: Considering these are warnings, should they be logged with the warning level?

pkg/services/alerting/extractor.go
```diff
@@ -74,7 +75,7 @@ func copyJSON(in json.Marshaler) (*simplejson.Json, error) {
 	return simplejson.NewJson(rawJSON)
 }
 
-func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json, validateAlertFunc func(*models.Alert) bool) ([]*models.Alert, error) {
+func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json, validators ...alertValidator) ([]*models.Alert, error) {
```
aknuds1, line 78: Can we rename this to getAlertsFromPanels considering it's returning an array of alerts?

pkg/services/alerting/extractor.go
```diff
@@ -203,16 +228,53 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
 
-func validateAlertRule(alert *models.Alert) bool {
-	return alert.ValidToSave()
+func validateAlertRule(alert *models.Alert) (ok bool, reason string) {
+	ok = alert.ValidToSave()
+	if !ok {
+		reason = fmt.Sprintf("Dashboard ID, Org ID or Panel ID is not correct, alertName=%v, panelId=%v, orgId=%v, dashboardId=%v", alert.Name, alert.PanelId, alert.OrgId, alert.DashboardId)
+	}
+	return ok, reason
+}
```
aknuds1, line 237: I don't see any point in having two return values, when we can instead just check if the returned reason is empty or not (it also reduces the possible bugs, since otherwise you could have `!ok` and an empty string):
```suggestion
func validateAlertRule(alert *models.Alert) string {
	if alert.ValidToSave() {
		return ""
	}

	return fmt.Sprintf("Dashboard ID, Org ID or Panel ID is not correct, alertName=%v, panelId=%v, orgId=%v, dashboardId=%v", alert.Name, alert.PanelId, alert.OrgId, alert.DashboardId)
}
```

pkg/services/alerting/alert_validator.go
```diff
@@ -0,0 +1,15 @@
...
+
+type validatorSeverity int
+
+const (
+	alertWarning validatorSeverity = iota
+	alertError
+)
+
+type alertValidator struct {
+	aFunc     func(*models.Alert) (bool, string)
```
aknuds1, line 13: Don't see any point in returning also a bool, since that should correspond to whether the reason is empty or not:
```suggestion
	aFunc     func(*models.Alert) string
```

pkg/services/alerting/extractor.go
```diff
@@ -203,16 +228,53 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
+		}
+
+		ok = false
+		if warnings.Len() > 0 {
+			warnings.WriteString("\n")
+		}
+		warnings.WriteString(fmt.Sprintf("Alert contains notification identified by incorrect ID, alertName=%v, panelId=%v, notificationId=%v", alert.Name, alert.PanelId, id))
+	}
+	reason = warnings.String()
+	return ok, reason
```
aknuds1, line 259:
```suggestion
func validAlertJSON(alert *models.Alert) reason string {
	warnings := strings.Builder{}
	for _, v := range alert.Settings.Get("notifications").MustArray() {
		jsonModel := simplejson.NewFromAny(v)
		id, err := jsonModel.Get("id").Int64()
		if err != nil {
			continue
		}

		if _, err := translateNotificationIDToUID(id, alert.OrgId); err == nil {
			continue
		}

		if warnings.Len() > 0 {
			warnings.WriteString("\n")
		}
		warnings.WriteString(fmt.Sprintf("Alert contains notification identified by incorrect ID, alertName=%v, panelId=%v, notificationId=%v", alert.Name, alert.PanelId, id))
	}
	return warnings.String()
```

pkg/services/alerting/extractor.go
```diff
@@ -249,6 +311,15 @@ func (e *DashAlertExtractor) extractAlerts(validateFunc func(alert *models.Alert
...
 func (e *DashAlertExtractor) ValidateAlerts() error {
-	_, err := e.extractAlerts(func(alert *models.Alert) bool { return alert.OrgId != 0 && alert.PanelId != 0 })
+	_, err := e.extractAlerts(alertValidator{
+		aFunc: func(alert *models.Alert) (ok bool, reason string) {
+			ok = alert.OrgId != 0 && alert.PanelId != 0
+			if !ok {
+				reason = fmt.Sprintf("Org ID or Panel ID is not correct, alertName=%v, panelId=%v, orgId=%v", alert.Name, alert.PanelId, alert.OrgId)
+			}
+			return ok, reason
+		},
```
aknuds1, line 321:
```suggestion
		aFunc: func(alert *models.Alert) reason string {
			if alert.OrgId == 0 || alert.PanelId == 0 {
				return fmt.Sprintf("Org ID or Panel ID is not correct, alertName=%v, panelId=%v, orgId=%v", alert.Name, alert.PanelId, alert.OrgId)
			}

			return ""
		},
```

pkg/services/alerting/extractor.go
```diff
@@ -193,8 +194,32 @@ func (e *DashAlertExtractor) getAlertFromPanels(jsonWithPanels *simplejson.Json,
...
 
-		if !validateAlertFunc(alert) {
-			return nil, ValidationError{Reason: fmt.Sprintf("Panel id is not correct, alertName=%v, panelId=%v", alert.Name, alert.PanelId)}
+		validationErrors := strings.Builder{}
+		validationWarnings := strings.Builder{}
+		for _, validator := range validators {
+			ok, reason := validator.aFunc(alert)
+			if ok {
+				continue
+			}
```
aknuds1, line 203:
```suggestion
			reason := validator.aFunc(alert)
			if reason == "" {
				continue
			}
```

aknuds1 (approved):
Generally looks good AFAICT, but left suggestions for improving the code.

pkg/services/alerting/extractor.go
papagian: thanks! I have changed it to return an error; do you mind to have a fresh look?

aknuds1: @papagian Have you pushed the change? I see the same code as before.

papagian: yes, that's the commit: ad933863e32057cf02caabaa5799e3c14033770d

aknuds1 (approved):
LGTM, but left a simplifying suggestion.

marefr (changes requested):
Been trying to make sense of the existing code, but I have a hard time doing that. Maybe I'm too tired? Feels like something is wrong with the alert rule extractor in general, but I cannot pinpoint exactly what. In general, the alert rule extractor behaves differently based on who is calling it if I understand it correctly which makes it hard to understand and possibly maintain. Here'e my first thoughts/suggestions of changes possibly needed. 

Starting at the bottom, the sqlstore function for looking up alert notification by uid doesn't return a "typed" error so it's almost impossible to differentiate if there's a real database error or if the error is due to notification channel couldn't be found:
https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/services/sqlstore/alert_notification.go#L183-L189

Mentioned function is called by https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/services/alerting/rule.go#L186 

Which is called by (and debug logging that uid lookup failed) https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/services/alerting/rule.go#L176 

And finally called and logged with error level if there's an error returned https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/services/alerting/rule.go#L143

To start with, removing the error and debug log would hide a possible real database problem. This would be the first thing I would suggest changing.

WDYT? Please provide feedback, maybe it's me who is confused.

papagian:
> Been trying to make sense of the existing code, but I have a hard time doing that. Maybe I'm too tired? Feels like something is wrong with the alert rule extractor in general, but I cannot pinpoint exactly what. In general, the alert rule extractor behaves differently based on who is calling it if I understand it correctly which makes it hard to understand and possibly maintain. Here'e my first thoughts/suggestions of changes possibly needed.

AFAICT the first time it attempts to save a dashboard and the dashboard does not have an ID
`extractAlerts` using this for validating alerts:
https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/services/alerting/extractor.go#L249-L254

for subsequent updates it calls this for validating alerts:
https://github.com/grafana/grafana/blob/957c88eaca0bfd666ad525ef607df4d60c4c4ea2/pkg/models/alert.go#L92-L93

So I have introduce an new validator for the alert JSON that is executed only during updating dashboards
(new dashboards are not expected to contain invalid notifiers)

Hope that this clarifies it a bit!

papagian:
> To start with, removing the error and debug log would hide a possible real database problem. This would be the first thing I would suggest changing.

good point! I will fix it.

marefr:
@papagian thanks, that's helpful. Regarding the existing code I'm probably responsible for introducing that :smile: 

Regarding the new validation rule `validAlertJSON` which calls `translateNotificationIDToUID`. The extractor already calls
https://github.com/grafana/grafana/blob/4acbcd7053d0549aaa53f982199dc31490478a94/pkg/services/alerting/extractor.go#L191

Which in turn could call `translateNotificationIDToUID`, creating multiple database calsl for the same thing. This is a code smell to me. Not sure how easy it is to fix though. Did you consider adding an additional argument to `NewRuleFromDBAlert` - maybe that would be a rather easy way to decide if the uid to id translation errors should be logged or not and maybe doesn't need as many changes to `DashAlertExtractor`? I realize now when I've written this that the `DashAlertExtractor` calls `NewRuleFromDBAlert` so maybe a bad idea?

But the idea would be to optimize for this so it doesn't log errors every time it reads the alert rule from the database, right?
https://github.com/grafana/grafana/blob/4acbcd7053d0549aaa53f982199dc31490478a94/pkg/services/alerting/reader.go#L39

papagian:
@marefr 
I reverted the previous modifications and I followed your suggestions.
Please check once more!

pkg/services/alerting/rule.go
```diff
@@ -140,7 +140,13 @@ func NewRuleFromDBAlert(ruleDef *models.Alert) (*Rule, error) {
 		if id, err := jsonModel.Get("id").Int64(); err == nil {
 			uid, err := translateNotificationIDToUID(id, ruleDef.OrgId)
 			if err != nil {
-				logger.Error("Unable to translate notification id to uid", "error", err.Error(), "dashboardId", model.DashboardID, "alertId", model.ID, "panelId", model.PanelID, "notificationId", id)
+				if !errors.Is(err, models.ErrAlertNotificationFailedTranslateUniqueID) {
+					return nil, ValidationError{Reason: "Failed to tranlate notification id to uid:" + err.Error(), DashboardID: model.DashboardID, AlertID: model.ID, PanelID: model.PanelID}
```
papagian, line 144: I'm not 100% sure here if in case of an actual database problem we want to cause the alert to fail or just log the error.

marefr: So if this error is returned it would end up here when reading alert rules from the database :
https://github.com/grafana/grafana/blob/c8658f3ee8408d3e7db7615740a30d3d5cfc1f90/pkg/services/alerting/reader.go#L40

That's a good question actually.The engine will continue to call `fetch()`, as seen below, so on one hand it will be "retried" on the other hand it may take a while for it to be "retried":
https://github.com/grafana/grafana/blob/c8658f3ee8408d3e7db7615740a30d3d5cfc1f90/pkg/services/alerting/engine.go#L85

Given there's a cache as well in sqlstore and having notifier id instead of uid should be less likely I think we want to log the error instead of returning it. WDYT?

pkg/services/alerting/rule.go
```diff
@@ -140,7 +140,13 @@ func NewRuleFromDBAlert(ruleDef *models.Alert) (*Rule, error) {
...
+				}
+
+				if logTranslationFailures {
+					logger.Warn("Unable to translate notification id to uid", "error", err.Error(), "dashboardId", model.DashboardID, "alertId", model.Name, "panelId", model.PanelID, "notificationId", id)
```
marefr, line 148: We already know here that the error is `models.ErrAlertNotificationFailedTranslateUniqueID` so don't need to include that generic error message. Why did you change `model.ID` to `model.Name`? In case the rule is not saved to the database and have an id of zero?  

```suggestion
					logger.Warn("Unable to translate notification id to uid", "dashboardId", model.DashboardID, "alertId", model.ID, "panelId", model.PanelID, "notificationId", id)
```

pkg/services/sqlstore/alert_notification.go
```diff
@@ -185,7 +185,7 @@ func getAlertNotificationUidInternal(query *models.GetAlertNotificationUidQuery,
 	}
 
 	if len(results) == 0 {
-		return fmt.Errorf("Alert notification [ Id: %v, OrgId: %v ] not found", query.Id, query.OrgId)
+		return models.ErrAlertNotificationFailedTranslateUniqueID
```
marefr, line 188: Looking at the code I'm unsure if any other consumer than `NewRuleFromDBAlert` are calling this. Since the name of the function is `getAlertNotificationUidInternal` maybe the returned error could reuse the existing `models.ErrAlertNotificationNotFound`. If there's only one consumer of this function I guess we can stick with returning `models.ErrAlertNotificationFailedTranslateUniqueID`. WDYT?

marefr (changes requested):
Great work! This is a big improvement, much easier to reason about and understand. To fix the actual problem I think this is the best we can possibly do, for now.  

I left some minor suggestions/comments/questions. Please have a look at those. But think we should be able to merge this soon.

pkg/services/alerting/rule.go
papagian: No, I've changed it because I think that a user can remember easier the name instead of the id.
But I'll keep your suggestion with the ID as it used to be before.

bergquist (approved):
:+1: 

I think this is as good as it gets unless we decide to refactor even more which doesn't make sense given alerting ng :)

pkg/services/alerting/rule.go
papagian: I agree that logging the error is preferable; I'll change it.

pkg/services/alerting/rule.go
papagian: Another reason for logging `model.Name` instead of `model.ID` is that if `NewRuleFromDBAlert` is called for an alert returned from a database query contains the ID:
https://github.com/grafana/grafana/blob/0b9eae7f84152e4187ced34f5c62886c7d9cf012/pkg/services/alerting/reader.go#L39

but when it is called with for an alert extracted from a dashboard JSON does not:
https://github.com/grafana/grafana/blob/0b9eae7f84152e4187ced34f5c62886c7d9cf012/pkg/services/alerting/extractor.go#L191
because the id is not stored in the JSON settings, so this is empty:
https://github.com/grafana/grafana/blob/0b9eae7f84152e4187ced34f5c62886c7d9cf012/pkg/services/alerting/extractor.go#L131

pkg/services/sqlstore/alert_notification.go
papagian: @marefr 
do you still want to return `models.ErrAlertNotificationNotFound` here? personally, I would prefer to differentiate them.

marefr: Let's stick with `ErrAlertNotificationFailedTranslateUniqueID` for now

marefr (approved):
Tested this now and it works as expected.

Testing alert rule generates warning :heavy_check_mark: However duplicate warnings, but lets disregard that for now:
WARN[10-22|12:56:09] Unable to translate notification id to uid logger=alerting.testnotification dashboardId=2916 alert="Spammy alert 1" panelId=23763571993 notificationId=2046
WARN[10-22|12:56:09] Unable to translate notification id to uid logger=alerting.testnotification dashboardId=2916 alert="Spammy alert 1" panelId=23763571993 notificationId=2046

Saving dashboard generates warning :heavy_check_mark: :
WARN[10-22|12:57:48] Unable to translate notification id to uid logger=alerting.testnotification dashboardId=2916 alert="Spammy alert 1" panelId=2 notificationId=2046

Evaluating the alert rule doesn't log an error :heavy_check_mark: 

LGTM great :+1:

kylebrandt:
removed CP label because I have created the above backport branch instead

PR #28168, opened 2020-10-10
Chore: Upgrade Node to 12.19.0

aknuds1:
SSIA.


hugohaggmark (approved):
LGTM! Do you know if we still use https://github.com/grafana/grafana-build-container?  If not consider removing this line:
https://github.com/grafana/grafana/blob/a4f37d08c510aae305cdbb5fec45c6af3b836bcd/packages/grafana-toolkit/src/cli/tasks/nodeVersionChecker.ts#L70

If we're still using https://github.com/grafana/grafana-build-container then I think that needs to be updated as well. Thank you.

aknuds1:
> LGTM! Do you know if we still use https://github.com/grafana/grafana-build-container? If not consider removing this line:
> https://github.com/grafana/grafana/blob/a4f37d08c510aae305cdbb5fec45c6af3b836bcd/packages/grafana-toolkit/src/cli/tasks/nodeVersionChecker.ts#L70

I don't think we use it any longer, no, as you can see it's archived. Thanks for the heads up.

PR #28171, opened 2020-10-11
TemplateSrv: Fix interpolating strings with object variables

### LINKED ISSUE #28125: Error occurs when using global variable $__user in MySQL query variable
**What happened**: After place __user global variable in field "Query" with MySQL datasource and leave this field, Grafana send two alert: "Validation - e.replace is not a function",  and "Templating - Template variables could not be initialized: e.replace is not a function"

**What you expected to happen**: User Grafana user id in MySQL query

**How to reproduce it (as minimally and precisely as possible)**: Create dashboard variable, choose Type: Query. Select data source MySQL. In Query field place "SELECT $__user;" and leave field Query.

**Anything else we need to know?**: no

**Environment**:
- Grafana version: Grafana v7.2.0 (efe4941ee3)
- Data source type & version: build in MySQL datasource 
- OS Grafana is installed on: docker
- User OS & Browser: Windows 10 + Chrome
- Grafana plugins:
- Others:
Снимок

torkelo:
Fixes issue in custom value format functions that expects a string or array of strings.
When using our new built in `__user` variable for example that is an object
these custom value formats can crash.

This ensures that we always transform the variable to a string before passing
it to a formatter.

Fixes #28125

### DIFF
--- a/public/app/features/templating/template_srv.test.ts
+++ b/public/app/features/templating/template_srv.test.ts
@@ -687,4 +687,35 @@ describe('templateSrv', () => {
       expect(target).toBe('2020-07');
     });
   });
+
+  describe('handle objects gracefully', () => {
+    beforeEach(() => {
+      initTemplateSrv([{ type: 'query', name: 'test', current: { value: { test: 'A' } } }]);
+    });
+
+    it('should not pass object to custom function', () => {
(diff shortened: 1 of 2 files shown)


hugohaggmark (approved):
LGTM!

PR #28176, opened 2020-10-11
CloudWatch Logs queue and websocket support

kaydelaney:
See #27969 
Opened to replace #27968

Amazon CloudWatch Logs imposes certain limits, or quotas, on the usage of various API endpoints and features. One of these limits is the limit on query concurrency, which has a default of 4, but can be increased to 10 on request. If a user attempts to start a query while the maximum number of concurrent queries are running, an error will be returned. In order to prevent CloudWatch Logs panels, or CloudWatch Logs queries sent in Explore, from returning errors in the case where the concurrent queries limit is reached, it is necessary to implement a queueing system that only dispatches a “Start Query” request when we are under the concurrent query limit.
In service of this, this PR changes the architecture of the CloudWatch datasource from one that periodically sends `GetQueryResults` proxied through the grafana backend, to a websocket based architecture using the currently in-progress Grafana Live feature.


pkg/models/live.go
```diff
@@ -20,7 +20,7 @@ type ChannelHandler interface {
 // ChannelHandlerProvider -- this should be implemented by any core feature
 type ChannelHandlerProvider interface {
 	// This is called fast and often -- it must be synchrnozed
```
aknuds1, line 22: Do you mind fixing this typo?
```suggestion
	// This is called fast and often -- it must be synchronized.
```

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	return nil, fmt.Errorf("channel with name '%s' not found", name)
+}
+
+func deleteResponseChannel(name string) {
+	delete(responseChannels, name)
+}
+
+// GetHandlerForPath called on init
```
aknuds1, line 66:
```suggestion
// GetHandlerForPath gets called on init.
```

pkg/services/live/live.go
```diff
@@ -14,8 +14,9 @@ import (
 )
 
 var (
-	logger   = log.New("live")
-	loggerCF = log.New("live.centrifuge")
+	logger             = log.New("live")
+	loggerCF           = log.New("live.centrifuge")
+	dsHandlerProviders = map[string]models.ChannelHandlerProvider{}
```
aknuds1, line 19: Can this be an attribute of GrafanaLive instead? Global variables are generally best avoided.

pkg/services/live/live.go
```diff
@@ -272,14 +277,28 @@ func (g *GrafanaLive) initChannel(id ChannelIdentifier) (models.ChannelHandler,
...
 		}
 		return nil, fmt.Errorf("unknown plugin: %s", id.Namespace)
 	}
 
 	return nil, fmt.Errorf("invalid scope: %s", id.Scope)
 }
 
+var handlerMu = sync.Mutex{}
+
+func RegisterHandler(datasourceID string, channelHandlerProvider models.ChannelHandlerProvider) error {
```
aknuds1, line 290: I wonder if this could be a method on GrafanaLive instead of free function, so it can avoid using global variables?

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -40,6 +43,19 @@ type datasourceInfo struct {
...
+
+	ec2Client           ec2iface.EC2API
+	rgtaClient          resourcegroupstaggingapiiface.ResourceGroupsTaggingAPIAPI
+	logsClientsByRegion map[string]cloudwatchlogsiface.CloudWatchLogsAPI
+	mtx                 sync.Mutex
+
+	queuesByRegion map[string](chan bool)
+	queueLock      sync.Mutex
+}
+
```
aknuds1, line 58: Why move this struct? If you don't have to move it, it makes the diff larger than it has to be.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -50,26 +66,22 @@ const logStreamIdentifierInternal = "__logstream__grafana_internal__"
...
 }
 
-// cloudWatchExecutor executes CloudWatch requests.
-type cloudWatchExecutor struct {
-	*models.DataSource
+func init() {
+	globalExecutor := newExecutor()
+	tsdb.RegisterTsdbQueryEndpoint("cloudwatch", func(ds *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
+		return globalExecutor, nil
+	})
```
aknuds1, line 82: I would revert this change, since it only makes the code more complicated. I'm also wondering why the function is moved, since it makes the diff larger.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +154,69 @@ func (e *cloudWatchExecutor) getRGTAClient(region string) (resourcegroupstagging
...
+	e.queueLock.Lock()
+	defer e.queueLock.Unlock()
+
+	if queue, ok := e.queuesByRegion[region]; ok {
+		return queue, nil
+	}
+
+	concurrentQueriesQuota, err := e.fetchConcurrentQueriesQuota(region)
+	if err != nil {
+		plog.Info("Could not fetch quota")
```
aknuds1, line 176: Why not return an error here? The value of concurrentQueriesQuota should be inconsistent in this case anyhow.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +154,69 @@ func (e *cloudWatchExecutor) getRGTAClient(region string) (resourcegroupstagging
...
+	}
+
+	concurrentQueriesQuota, err := client.GetServiceQuota(&servicequotas.GetServiceQuotaInput{
+		ServiceCode: aws.String("logs"),
+		QuotaCode:   aws.String("L-32C48FBB"),
+	})
+
+	if err != nil {
+		return defaultConcurrentQueries, err
+	}
```
aknuds1, line 198: Error checks should be co-located with the error's origin, for readability:
```suggestion
	concurrentQueriesQuota, err := client.GetServiceQuota(&servicequotas.GetServiceQuotaInput{
		ServiceCode: aws.String("logs"),
		QuotaCode:   aws.String("L-32C48FBB"),
	})
	if err != nil {
		return defaultConcurrentQueries, err
	}
```

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +154,69 @@ func (e *cloudWatchExecutor) getRGTAClient(region string) (resourcegroupstagging
...
+
+	defaultConcurrentQueriesQuota, err := client.GetAWSDefaultServiceQuota(&servicequotas.GetAWSDefaultServiceQuotaInput{
+		ServiceCode: aws.String("logs"),
+		QuotaCode:   aws.String("L-32C48FBB"),
+	})
+
+	if err != nil {
+		return defaultConcurrentQueries, err
+	}
```
aknuds1, line 211:
```suggestion
	defaultConcurrentQueriesQuota, err := client.GetAWSDefaultServiceQuota(&servicequotas.GetAWSDefaultServiceQuotaInput{
		ServiceCode: aws.String("logs"),
		QuotaCode:   aws.String("L-32C48FBB"),
	})
	if err != nil {
		return defaultConcurrentQueries, err
	}
```

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+const (
+	maxAttempts         = 8
+	initialPollInterval = 500 * time.Millisecond
+)
+
+var (
+	channelMu        = sync.Mutex{}
+	responseChannels = map[string]chan *tsdb.Response{}
```
aknuds1, line 36: Can we avoid adding global variables?

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+	if responseChannel, ok := responseChannels[name]; ok {
+		return responseChannel, nil
+	}
+
+	return nil, fmt.Errorf("channel with name '%s' not found", name)
+}
+
+func deleteResponseChannel(name string) {
+	delete(responseChannels, name)
+}
```
aknuds1, line 64: I would inline this function, it's too simple to warrant the abstraction.

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+// GetHandlerForPath called on init
+func (supplier *LogQueryRunnerSupplier) GetHandlerForPath(path string, publisher models.ChannelPublisher) (models.ChannelHandler, error) {
+	return &LogQueryRunner{
+		channelName: path,
+		publish:     publisher,
+		running:     make(map[string]bool),
+	}, nil
+}
+
+// GetChannelOptions called fast and often
```
aknuds1, line 75:
```suggestion
// GetChannelOptions gets channel options.
It's called fast and often.
```

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+func (g *LogQueryRunner) GetChannelOptions(id string) centrifuge.ChannelOptions {
+	return centrifuge.ChannelOptions{}
+}
+
+// OnSubscribe for now allows anyone to subscribe to any dashboard
```
aknuds1, line 80: Can you rewrite this to a sentence describing what OnSubscribe does?

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,269 @@
...
+func (g *LogQueryRunner) OnSubscribe(c *centrifuge.Client, e centrifuge.SubscribeEvent) error {
+	if _, ok := g.running[e.Channel]; !ok {
+		g.running[e.Channel] = true
+		go g.publishResults(e.Channel)
+	}
+
+	return nil
+}
+
+// OnPublish called when an event is received from the websocket
```
aknuds1, line 90:
```suggestion
// OnPublish is called when an event is received from the websocket.
```

aknuds1 (commented):
I left some questions and suggestions.

pkg/tsdb/cloudwatch/cloudwatch.go
kaydelaney: Wanted to have all the struct definitions at the top of the file

pkg/tsdb/cloudwatch/cloudwatch.go
kaydelaney: Because rather than having everything fail it seemed to make more sense to just use a hardcoded default (4 in this case), something that's possible since certain permissions are required on the AWS end to retrieve the quota values.

pkg/models/live.go
kaydelaney: Most of these comments are copy-pasted, but I can fix them, sure

pkg/services/live/live.go
kaydelaney: This was initially what I wanted to do, but since `Live` is a property of the HTTP server, I wasn't sure how best to expose it so that it can be used, for example, by the CloudWatch data source

pkg/tsdb/cloudwatch/cloudwatch.go
aknuds1: @kaydelaney Please don't move things around if not necessary, since it makes it more difficult to diff and see what has changed.

pkg/tsdb/cloudwatch/cloudwatch.go
kaydelaney: This change is necessary since otherwise a new executor is created for each query request, which would mean query state isn't persisted between queries

pkg/services/live/live.go
aknuds1: Hmm.. I guess it should be looked into.

pkg/tsdb/cloudwatch/cloudwatch.go
aknuds1: @kaydelaney Ah, I didn't consider that it might be intentional to have a persisted closure variable.

pkg/services/live/live.go
kaydelaney: Yeah, I'll discuss with @ryantxu . Definitely agree that having global variables isn't ideal

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -50,26 +66,22 @@ const logStreamIdentifierInternal = "__logstream__grafana_internal__"
...
+	globalExecutor := newExecutor()
+	tsdb.RegisterTsdbQueryEndpoint("cloudwatch", func(ds *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
+		return globalExecutor, nil
+	})
 
-	ec2Client           ec2iface.EC2API
-	rgtaClient          resourcegroupstaggingapiiface.ResourceGroupsTaggingAPIAPI
-	logsClientsByRegion map[string]cloudwatchlogsiface.CloudWatchLogsAPI
-	mtx                 sync.Mutex
+	live.RegisterHandler("CloudWatch", &LogQueryRunnerSupplier{})
```
ryantxu, line 84: I don't really understand the workflow here -- this looks like when cloudwatch starts, it sets up an endpoint with live.  I would expect something more like

When live gets a request to: `/ds/22/path`, live will look up datasource 22, find the plugin that owns it and ask it for the config

I'm not sure the best way to have plugins register their support, for cloudwatch it does not really matter, but we soon want to do this for gRPC based plugins also

public/app/plugins/datasource/cloudwatch/plugin.json
```diff
@@ -5,6 +5,7 @@
   "category": "cloud",
   "metrics": true,
   "logs": true,
+  "live": true,
```
ryantxu, line 8: Is this necessary now?  I think we will need something like it eventually, but lets be careful about how we support it.  Ideally it can actually be driven by the existence of interfaces and backend support.

pkg/services/live/live.go
ryantxu: ideally the datasource handlers are tied to their instance, not registered ad-hoc.

ryantxu (commented):
This is looking good -- yes @aknuds1 , we should avoid globals if possible!

@kaydelaney perhaps, this should use the (still TBD) plugin endpoint rather than datasource endpoint.  For datasource, the thought was that it would be scoped to an instance, but plugin is scoped to the whole plugin.  So in this case it would be:

`plugin/cloudwatch/$path`

ryantxu:
As a quick-and-dirty, make this work for 7.3 approach, what about:

```
diff --git a/pkg/services/live/live.go b/pkg/services/live/live.go
index ee7156d53d..cfe38a6a87 100644
--- a/pkg/services/live/live.go
+++ b/pkg/services/live/live.go
@@ -276,6 +276,10 @@ func (g *GrafanaLive) GetChannelNamespace(scope string, name string) (models.Cha
        }

        if scope == "plugin" {
+               if name == "cloudwatch" {
+                       return &CloudwatchPluginHandler{}, nil
+               }
+
                p, ok := plugins.Plugins[name]
                if ok {
                        h := &PluginHandler{
```

Then this PR can focus on the cloudwatch behavior, and we can solve making this more generic later?

pkg/tsdb/cloudwatch/cloudwatch.go
aknuds1: @kaydelaney As I said before, please move this and `newExecutor` back to their original locations, since moving them around makes it difficult to review the real changes. If you want to move them, please discuss it in another PR.

aknuds1 (changes requested):
Please move cloudWatchExecutor and newExecutor back to their original places in cloudwatch.go, so it's practical to review the changes proposed in the PR.

pkg/tsdb/cloudwatch/live.go
aknuds1: Please make these into struct fields, so we can avoid globals.

pkg/tsdb/cloudwatch/live.go
aknuds1: On reading the code more closely, inline it and avoid the global.

aknuds1 (changes requested):

aknuds1:
If you want, I can help you make the changes necessary to avoid the globals.

pkg/tsdb/cloudwatch/cloudwatch.go
kaydelaney: There is a map `dsHandlerProviders`, where the datasourceID (in this case, "CloudWatch") and the corresponding `ChannelHandlerProvider` are stored when `RegisterHandler` is called. Later when`initChannel` is called, the `Namespace` parameter of `id` is used as a key for `dsHandlerProviders` to retrieve the handler provider, after which `GetHandlerForPath` is called.

public/app/plugins/datasource/cloudwatch/plugin.json
kaydelaney: Not necessary, think this is just a holdover from an earlier version of Live

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +154,67 @@ func (e *cloudWatchExecutor) getRGTAClient(region string) (resourcegroupstagging
...
+	e.queueLock.Lock()
+	defer e.queueLock.Unlock()
+
+	if queue, ok := e.queuesByRegion[region]; ok {
+		return queue, nil
+	}
+
+	concurrentQueriesQuota, err := e.fetchConcurrentQueriesQuota(region)
+	if err != nil {
+		plog.Info("Could not fetch quota")
```
aknuds1, line 176: Please make it explicit instead, since when a function returns an error you shouldn't trust the other values it returns:
```suggestion
		plog.Info("Could not fetch quota")
		concurrentQueriesQuota = defaultConcurrentQueries
```

aknuds1 (changes requested):

marefr:
Is this supposed to go into 7.3? I don't want to ruin the party, but a few considerations from my part. Live is currently behind a feature toggle as far as I know. I don't think we have a good story for websockets and running Grafana in HA mode meaning that the current approach will fail to use CloudWatch logs if you use a setup with multiple Grafana instances. 

> this should use the (still TBD) plugin endpoint rather than datasource endpoint.

I also agree with this.

What other options, if any, have you considered besides websockets? Did you consider a client side queue?

aknuds1:
@marefr I am also wondering if this should be delayed until after 7.3, since it's not well understood what changes are introduced and there is a risk of breaking things.

kaydelaney:
@marefr Re: HA mode, this has been discussed with Amazon and the conclusion was that while not ideal, it's still better to have queueing for non-HA setups than to not have it at all.
As for other options, yes, others have been considered. Not sure how a client-side queue would work with this specific case, since the purpose of the queue is to prevent overloading of the CloudWatch Logs API tied to a data source

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -50,15 +53,21 @@ const logStreamIdentifierInternal = "__logstream__grafana_internal__"
 var plog = log.New("tsdb.cloudwatch")
 var aliasFormat = regexp.MustCompile(`\{\{\s*(.+?)\s*\}\}`)
 
+const defaultConcurrentQueries = 4
+
 func init() {
+	globalExecutor := newExecutor()
 	tsdb.RegisterTsdbQueryEndpoint("cloudwatch", func(ds *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
-		return newExecutor(), nil
+		return globalExecutor, nil
```
aknuds1, line 61: What's the benefit of this change? Why is it necessary for this PR?

marefr:
@kaydelaney 

> Re: HA mode, this has been discussed with Amazon and the conclusion was that while not ideal, it's still better to have queueing for non-HA setups than to not have it at all.

Okay :+1: Does this mean that it wouldn't work at all in HA or kind of work with the risk of being hit by query concurrency?

> As for other options, yes, others have been considered. Not sure how a client-side queue would work with this specific case, since the purpose of the queue is to prevent overloading of the CloudWatch Logs API tied to a data source

I don't have the all the technical details, but I noticed you do query something like concurrent queries quota and thought maybe a polling solution could work by returning the quota to the client after each query.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +154,67 @@ func (e *cloudWatchExecutor) getRGTAClient(region string) (resourcegroupstagging
...
+	if err != nil {
+		return defaultConcurrentQueries, err
+	}
+
+	if defaultConcurrentQueriesQuota != nil && defaultConcurrentQueriesQuota.Quota != nil {
+		return int(*defaultConcurrentQueriesQuota.Quota.Value), nil
+	}
+
+	return defaultConcurrentQueries, nil
+}
```
aknuds1, line 216: After playing with the code locally, I think it makes more sense for this function to not return an error, and instead log a warning about what exactly went wrong. The returned error will only be logged anyhow, so it's pointless.

```suggestion
func (e *cloudWatchExecutor) fetchConcurrentQueriesQuota(region string) int {
	client, err := e.getServiceQuotasClient(region)
	if err != nil {
		plog.Warn("Could not get service quota client")
		return defaultConcurrentQueries
	}

	cqq, err := client.GetServiceQuota(&servicequotas.GetServiceQuotaInput{
		ServiceCode: aws.String("logs"),
		QuotaCode:   aws.String("L-32C48FBB"),
	})
	if err != nil {
		plog.Warn("Could not get service quota")
		return defaultConcurrentQueries
	}

	if cqq != nil && cqq.Quota != nil && cqq.Quota.Value != nil {
		return int(*cqq.Quota.Value)
	}

	plog.Warn("Could not get service quota")

	dcqq, err := client.GetAWSDefaultServiceQuota(&servicequotas.GetAWSDefaultServiceQuotaInput{
		ServiceCode: aws.String("logs"),
		QuotaCode:   aws.String("L-32C48FBB"),
	})
	if err != nil {
		plog.Warn("Could not get default service quota")
		return defaultConcurrentQueries
	}

	if defaultConcurrentQueriesQuota != nil && defaultConcurrentQueriesQuota.Quota != nil {
		return int(*defaultConcurrentQueriesQuota.Quota.Value)
	}

	plog.Warn("Could not get default service quota")
	return defaultConcurrentQueries
}
```

aknuds1 (changes requested):

pkg/tsdb/cloudwatch/cloudwatch.go
aknuds1: @kaydelaney Thanks, have you considered if it could break any of the existing functionality though?

kaydelaney: The only potential issue I can think of at the moment involves how the cloudwatch logs clients are cached at the moment. Currently there's no invalidation mechanism so if some changes were made to the datasource by a user (default region, credentials, etc.) those changes wouldn't take effect unless the backend is restarted, which obviously isn't ideal. This could be fixed by just getting rid of the client caching though.

pkg/tsdb/cloudwatch/live.go
kaydelaney: Maybe I'm misunderstanding but why would inlining avoid using a global variable?

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,246 @@
...
+	return response, nil
+}
+
+func (e *cloudWatchExecutor) sendQueriesToChannel(ctx context.Context, queryContext *tsdb.TsdbQuery, responseChannel chan *tsdb.Response) {
+	eg, ectx := errgroup.WithContext(ctx)
+
+	for _, query := range queryContext.Queries {
+		query := query
+		eg.Go(func() error {
+			return e.startQuery(ectx, responseChannel, query, queryContext.TimeRange)
```
aknuds1, line 144: Should be made clear it's for the live functionality.
```suggestion
			return e.startLiveQuery(ectx, responseChannel, query, queryContext.TimeRange)
```

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if err := eg.Wait(); err != nil {
+		plog.Error(err.Error())
+	}
+
+	close(responseChannel)
+}
+
+func (e *cloudWatchExecutor) startQuery(ctx context.Context, responseChannel chan *tsdb.Response, query *tsdb.Query, timeRange *tsdb.TimeRange) error {
```
aknuds1, line 155: Should be made clear it's for the live functionality.
```suggestion
func (e *cloudWatchExecutor) startLiveQuery(ctx context.Context, responseChannel chan *tsdb.Response, query *tsdb.Query, timeRange *tsdb.TimeRange) error {
```

aknuds1 (changes requested):

pkg/services/live/live.go
```diff
@@ -257,22 +258,30 @@ func (g *GrafanaLive) initChannel(id ChannelIdentifier) (models.ChannelHandler,
...
 		if ok {
-			return p.GetHandlerForPath(id.Path)
+			return p.GetHandlerForPath(id.Path, g.Publish)
 		}
 		return nil, fmt.Errorf("Unknown feature: %s", id.Namespace)
 	}
 
 	if id.Scope == "ds" {
-		return nil, fmt.Errorf("todo... look up datasource: %s", id.Namespace)
+		if id.Scope == "ds" {
```
papagian, line 267: This condition is exactly the same as the above. Did you mean something else?

pkg/tsdb/cloudwatch/live.go
kaydelaney: To briefly outline the workflow:
A client makes an HTTP request with the query, and query type `liveLogAction`. The `cloudWatchExecutor` `Query` method is called, which then calls `executeLiveLogQuery`. From there, a UUID is generated, and a corresponding go channel is made, through which query responses will be sent and consumed. A response to the client's HTTP request is made which includes that UUID. At the same time, `StartQuery` requests are made to AWS, followed by periodic `GetQueryResults` requests, the results of which are sent to the go channel previously created.
When the client subscribes to the centrifuge channel, `OnSubscribe` is called and the responses from the corresponding go channel are taken and sent through the centrifuge channel back to the client until the go channel closes (i.e. the query has completed).

The issue is, there doesn't seem to be a simple way (ignoring global variables) to let `LogQueryRunner` have access to `cloudWatchExecutor`, since `LogQueryRunner` isn't created until a client subscribes and `GetHandlerForPath` is called. Furthermore there doesn't seem to be an easy way to access the executor - earlier I thought maybe since it's set in a registry using `RegisterTsdbQueryEndpoint` there might be a corresponding function to retrieve it, which there is (`getTsdbQueryEndpointFor`) but it isn't exposed outside the `tsdb` package.
I think it would make things easier if that function _were_ exposed, but perhaps there's a reason it's not. Any thoughts?

pkg/services/live/live.go
kaydelaney: Good catch, sloppy copy/paste on my part

pkg/tsdb/cloudwatch/live.go
aknuds1: Thanks @kaydelaney. It's too late for me tonight, but I can look more at this tomorrow.

pkg/tsdb/cloudwatch/live.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return nil
+}
+
+func (e *cloudWatchExecutor) executeLiveLogQuery(ctx context.Context, queryContext *tsdb.TsdbQuery) (*tsdb.Response, error) {
+	responseChannelName := uuid.Must(uuid.NewV4()).String()
+	responseChannel := make(chan *tsdb.Response)
+	if err := addResponseChannel("plugin/cloudwatch/"+responseChannelName, responseChannel); err != nil {
+		return nil, err
+	}
```
ryantxu, line 128: This will create a new channel for each request -- this may be fine, but also means that concurrent users all make individual queries to the data source.  If the query is "show me the latest logs", ideally we can make a *single* request to cloudwatch, and share the results with everyone making the same query.  Given that queries come from dashboards, it is very likely to have everyone executing the same query!

If there this is an easy change, base64(part of the query + salt)?, i think we should try soon -- if not multiple queries is fine till we figure out a reasonable strategy.

ryantxu (commented):
@kaydelaney this is looking great.  I don't know the query semantics enough to know if we should try to reuse results or execute new ones for each client.

I am excited to see this getting used by someone not me :)

ryantxu:
@marefr 

> Is this supposed to go into 7.3? I don't want to ruin the party, but a few considerations from my part. 

I hope so, yes.  Getting real alpha usage is super valuable.


> Live is currently behind a feature toggle as far as I know. I don't think we have a good story for websockets and running Grafana in HA mode 

What is this replacing?  IE, is there something working in 7.2 that will *stop* working in 7.3 if the live toggle is not set?  If the answer is yes, then we need to be very careful, but if this is all new things added to 7.3, I think we just need the appropriate caveats.

In 7.4, we can support HA websockets using [the centrifuge redis engine](https://github.com/centrifugal/centrifuge/blob/master/_examples/redis_engine/main.go#L92)  -- reasonable HA support is a prerequisite for removing the feature flag, that said the sooner we get experience with non-HA behavior the better (and hosted grafana is not HA anyway!)

ryantxu:
> What other options, if any, have you considered besides websockets? Did you consider a client side queue?

@marefr this PR is not adding websockets, it is using the infrastructure provided by grafana live.

Grafana live is implemented using https://github.com/centrifugal/centrifuge -- from the client, we connect to the server using [sockjs](https://github.com/centrifugal/centrifuge-js#sockjstransports), this will then pick the best transport for the connection.  

In general I am super happy that this pushes on how we can support streaming.  Obviously lots of details and patterns to develop ;)

daniellee:
It would be great to get this into 7.3 but there are a lot of comments here and it's getting hard to see what's left to do for Kay. Can @aknuds1 and @marefr summarize the remaining feedback into two categories: what needs to be fixed so the PR can be merged and what can be fixed during the beta period.

I'm seeing the following:
- Concerns about HA mode
- Global variables
- Which interface to use: plugin endpoint or data source endpoint
- @ryantxu 's [feedback](https://github.com/grafana/grafana/pull/28176#discussion_r504878186) on channels

Thanks everyone for doing such a thorough review for Kay.

pkg/tsdb/cloudwatch/live.go
kaydelaney: In theory I like the idea of being able to share results among multiple users, but since query results are informed by start and end time ranges, in practice I imagine it's unlikely enough that many users would have the exact time ranges & queries that I'm reluctant to optimize for that use case.

ryantxu: :+1: -- it really depends on what a normal query is and how long they are expected to run. 

I think the current state is good to merge (likely after some changes from @aknuds1), but we should revist some channel options in 7.3+:
* private channels -- with the current query model, we are creating a new channel for each request, but it is only ever sent to one person.  I think there is a method that we only publish results back to an explicit client Id
* consider reuse -- for splunk logs, we will certainly want to reuse the query results for all connected clients (queries are very expensive), I assume we can more easily improve the pattern here before adding the complication for how to real plugins

ryantxu (approved):
This looks good to me, and is good progress towards a hard new pattern.  Wait for some feedback from @aknuds1, but I think this is worth getting into 7.3

pkg/services/live/live.go
ryantxu: @aknuds1 this will require routing that does not yet exist.  Although centrifuge lets us publish data to any channel, we want to add more clear constraints and structure so that portions of the channel namespace are "owend" by various parts of the system:

The thee scopes we have defined (but not yet implemented)

* /live/grafana/${feature}
* /live/ds/${datasourceId | name?}
* /live/plugin/${pluginId}

This PR now uses hardcodes its access to `plugin/cloudwatch/*`  For the particular approach in this PR (start a runner from a query, and listen to results on a channel) that is totally fine and can be improved without issue.

aknuds1:
Moving this to the next 7.3 milestone, since there's no time to review before the release.

kaydelaney:
> Moving this to the next 7.3 milestone, since there's no time to review before the release.

I think only fixes are supposed to go into 7.3, but I could be wrong. Can you confirm @hugohaggmark ?

aknuds1:
> > Moving this to the next 7.3 milestone, since there's no time to review before the release.
> 
> I think only fixes are supposed to go into 7.3, but I could be wrong. Can you confirm @hugohaggmark ?

@kaydelaney Since this is behind a feature flag, I guess it's not a big deal? Otherwise we can also make a beta2. I think it's preferable to rushing this into v7.3.0 without proper vetting.

hugohaggmark:
> > Moving this to the next 7.3 milestone, since there's no time to review before the release.
> 
> I think only fixes are supposed to go into 7.3, but I could be wrong. Can you confirm @hugohaggmark ?

You're correct @kaydelaney, I don't think we should force any bigger features into 7.3 unless there are very strong opinions for adding them.

aknuds1:
@hugohaggmark @kaydelaney Maybe there's a misunderstanding. I moved it into the 7.3.0 milestone, which will either be the next 7.3.0 beta or stable. I don't think it's an issue getting a feature behind a flag into either of those. I also discussed with @ryantxu yesterday about another PR of his, and he agrees with the reasoning.

hugohaggmark:
My bad, I didn't go through the code so I didn't realize all these changes were behind a feature toggle. In that case it should be fine to add it.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -187,20 +193,12 @@ func (e *cloudWatchExecutor) getCWClient(region string) (cloudwatchiface.CloudWa
 }
 
 func (e *cloudWatchExecutor) getCWLogsClient(region string) (cloudwatchlogsiface.CloudWatchLogsAPI, error) {
-	e.mtx.Lock()
-	defer e.mtx.Unlock()
-
-	if logsClient, ok := e.logsClientsByRegion[region]; ok {
-		return logsClient, nil
-	}
```
aknuds1, line 195: This change of no longer caching the logs client has an effect also on users not enabling the feature flag. Why is this change made? Will it have a negative effect on performance?

kaydelaney: I described why this change was made [here](https://github.com/grafana/grafana/pull/28176#discussion_r504032077)
I don't foresee a negative effect on performance as, as far as I can tell, the caching wasn't functioning correctly anyway, since a new executor was created for each request.

aknuds1: Aha, thanks for pointing me to the explanation. I'm still unsure if the caching should be removed via this PR though, as it's for a completely different topic. I'll discuss it with the others.

pkg/tsdb/cloudwatch/cloudwatch.go
marefr: I think this one could be quite a big risk. Have a feeling that cloudwatch datasource have stored temporary data on cloudwatch executor which have been used further down the call chain. Last time this happened was recently when we fixed a bug for custom metrics. Introducing this could mean that it has to be rewritten to the "thread safe"

I obviously don't have all the details and understanding. My focus (and Arve's) is to be able to merge this for 7.3.

Thinking out loud a bit here, but would like to get some better understandings so please let me know if I'm completely off.

Feels like changing to a "globalExecutor" was made because you need to have one instance of `queuesByRegion map[string](chan bool)` which I guess is the queue used for "handling" logs queries. To me it does look a bit strange  to only have one queue by region, but I don't know all the details. What happens if you have multiple cloudwatch data sources targeting different accounts - is there a risk pushing out logs from wrong account to a user?

My solution to not having a "globalExecutor" would be to define a new struct or interface that can be provided as dependency when creating a new `cloudWatchExecutor`. Example code:

```go
func init() {
  myLogQueueService := newLogQueueService()
  
  tsdb.RegisterTsdbQueryEndpoint("cloudwatch", func(ds *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
    return newExecutor(myLogQueueService), nil
  }
}
```

To take this a step further we can define a new struct/interface than handles everything related to a "liveLogAction". Maybe you share a lot of code between "logAction" and "liveLogAction" and this would not be optimal, but maybe there's some kind of middle way abstraction between "logQueue" and "liveLogActionHandler". In addition we could maybe ask the live package from cloudwatch to register the handler instead of the opposite order like now. Example code :

```go
func init() {
  myliveLogActionExecutor := newLiveLogActionExecutor()
  // register live handler
  live.RegisterPluginHandler("cloudwatch", myliveLogActionExecutor)
  tsdb.RegisterTsdbQueryEndpoint("cloudwatch", func(ds *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
    return newExecutor(myliveLogActionExecutor), nil
  }
}

func (e *cloudWatchExecutor) Query(ctx context.Context, dsInfo *models.DataSource, queryContext *tsdb.TsdbQuery) (*tsdb.Response, error) {
  ...
  case "liveLogAction":
    return e.liveLogActionExecutor.Execute(ctx, queryContext)
  ...
}
```

Going this path the liveLogAction handling will be completely separated from existing cloudwatch implementation and by that reduces risk and it also simplifies things according to me. Obviously more work to get in these changes, but I think they're needed to get this into 7.3.

@kaydelaney @aknuds1 I suggest that we meet tomorrow, if needed, and sync on/discuss the path forward for getting this merged. WDYT?

aknuds1: I'm up for a meet today :)

aknuds1:
I pushed a merge with master fixing merge conflicts, hope you don't mind.

pkg/tsdb/cloudwatch/cloudwatch.go
kaydelaney: > Have a feeling that cloudwatch datasource have stored temporary data on cloudwatch executor which have been used further down the call chain

Spent some more time looking into this and while I think it should be fine for the cloudwatch logs side of things, I hadn't considered the metrics side of things and struct properties like `ec2Client` which get set and changed in a way that wouldn't be threadsafe with a global executor.

> Feels like changing to a "globalExecutor" was made because you need to have one instance of `queuesByRegion map[string](chan bool)` which I guess is the queue used for "handling" logs queries. To me it does look a bit strange to only have one queue by region, but I don't know all the details. What happens if you have multiple cloudwatch data sources targeting different accounts - is there a risk pushing out logs from wrong account to a user?

Re: Having multiple cloudwatch data sources targeting different accounts and how this might affect things, I had actually made a note for myself to investigate that but promptly forgot about it, so thank you for bringing it up, as it's a very good point. The queues would have to be by region and by account, so I suppose we could still have a map with the channels as before, but rather than just using the region as string, something like
`queueKey :=  region + "-" + accountID` would work better. To clarify though, as it is now it wouldn't result in the wrong logs getting pushed to users, it just means requests wouldn't be dispatched as quickly as they could be.

marefr: Thanks :+1: It didn't really go as planned in regards to meet today, but @aknuds1 and me have started to look into a couple of tweaks of how we can improve. Lets sync up on Monday.

aknuds1 (commented):
LGTM, but have to test.

aknuds1 (approved):
Tested and looks good to go AFAICT! Very happy we were able to resolve this one in the end!

aknuds1:
@kaydelaney I switched to [Google's](https://github.com/google/uuid) UUID implementation since I found mentions that the satori library is unmaintained since years, and supposedly has critical flaws.

marefr (approved):
Pushed a couple minor changes. Please have a look at those before merging.
- Ran some basic tests of CloudWatch and seems to work as expected :heavy_check_mark: 
- Tried to run logs query, but didn't find any data. I expect you would test this.

Great work! This feels much better :heart:

ChinthakaJay:
Hi, Did this resolved the issue? I am using Grafana v7.5.1 (4dffb875cf) and have 10 dashboards with 10 alerts configured for CloudWatch Logs. 

I see this exception roughly at 10 min intervals, 

`Error: tsdb.HandleRequest() error LimitExceededException: Account maximum query concurrency limit of [10] reached. (Service: AWSLogs; Status Code: 400; Error Code: LimitExceededException; Request ID: c71c35b3-b3ea-40e9-9312-548d66053dea; Proxy: null)`

My alerts are configured to evaluate every 10 min.
And I am also getting this in random panels when my dashboards refresh. However when i retry the query it works.

`Object
status:500
statusText:"Internal Server Error"
data:Object
message:"Metric request error"
config:Object
method:"POST"
url:"api/tsdb/query"
data:Object
retry:0
headers:Object
hideFromInspector:false
message:"Metric request error"`

PR #28180, opened 2020-10-12
Azure: Fixes cancellation of requests with different Azure sources

### LINKED ISSUE #28114: Panel with queries from different sources not working
When a panel configured with query from different source in my case query A is application insights and query B is azure log anlytics. In query inspector no data is returned. It seems for different sources grafana is making two requests to 
/api/ds/query one for application insights and one for log anytics i can see in browser one query is getting blocked. Tried in both chrome and firefox same result chrome its showing cancelled. This was working fine before , i am not sure which version was running before as this was in kubernetes and when kubernetes cluster upgraded and restarted i think it downloaded a new container version. Attached the screenshot of network call from browser

Grafana version  v7.2.0

grafana

hugohaggmark:

Fixes #28114


torkelo (approved):
LGTM

PR #28190, opened 2020-10-12
Alerting: remove obsolete notifiers (after user confirmation)

papagian:
Following #28162 that detects n the backend obsolete alert notifiers when a dashboard is saved,
this fix introduces a confirmation modal for removing the invalid notifier
when a user navigates to the alert tab and an obsolete notifier is detected.

Screenshot 2020-10-12 at 12 39 48

Fixes #

The reason that a dashboard can contain invalid notifiers it that
when the notifier is deleted, the notifier references are not deleted from the dashboard JSON as well.

### DIFF
--- a/public/app/features/alerting/AlertTabCtrl.ts
+++ b/public/app/features/alerting/AlertTabCtrl.ts
@@ -225,6 +225,18 @@ export class AlertTabCtrl {
       // fallback to using id if uid is missing
       if (!model) {
         model = _.find(this.notifications, { id: addedNotification.id });
+        if (!model) {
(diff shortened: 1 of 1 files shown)


hugohaggmark (approved):
LGTM! 
nit: Would be nice to add a test for this if possible

papagian:
> LGTM!
> nit: Would be nice to add a test for this if possible

thanks! Is there a similar test where I could have a look?

hugohaggmark:
> > LGTM!
> > nit: Would be nice to add a test for this if possible
> 
> thanks! Is there a similar test where I could have a look?

Sorry but I thought we already had tests for AlertTabCtrl.

PR #28195, opened 2020-10-12
Datasource: Prefix SigV4 json data keys with sigV4

marefr:
Prefixes SigV4 json data and secure json data keys with `sigV4` to make sure they don't collide with any other settings. Also minor fixes to UI.

Ref #27552 

### DIFF
--- a/packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
+++ b/packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
@@ -84,7 +84,7 @@ export const SigV4AuthSettings: React.FC<HttpSettingsProps> = props => {
 
   return (
     <>
-      <h3 className="page-heading">Sigv4 Details</h3>
+      <h6>SigV4 Auth Details</h6>
       <div className="gf-form-group">
         <div className="gf-form-inline">
           <div className="gf-form">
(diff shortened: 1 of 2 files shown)


packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
```diff
@@ -97,21 +97,17 @@ export const SigV4AuthSettings: React.FC<HttpSettingsProps> = props => {
             <Select
               className="width-30"
               value={authProviderOptions.find(
-                authProvider => authProvider.value === dataSourceConfig.jsonData.authType
+                authProvider => authProvider.value === dataSourceConfig.jsonData.sigV4AuthType
               )}
               options={authProviderOptions}
-              defaultValue={dataSourceConfig.jsonData.authType || 'keys'}
+              defaultValue={dataSourceConfig.jsonData.sigV4AuthType || 'keys'}
```
marefr, line 103: Fallback to default doesn't seem to work:
image

I also think that default should be "default"/AWS SDK Default

wbrowne: Weird.. must have got messed up from switching from the original design to the existing CW config editor. Looks like that's even changed again so we might be able to just make them consistent presuming CW works as expected.

Agree regarding the default 👍

wbrowne (approved):
Checked locally and it looks good 👍

packages/grafana-ui/src/components/DataSourceSettings/SigV4AuthSettings.tsx
marefr: Okay I'll merge this one for now and handle changes in upcoming PR when CloudWatch auth changes have been merged

PR #28211, opened 2020-10-13
Azure Datasource: WIP - Using data proxy timeout as Prefer header

### LINKED ISSUE #28199: Use 'Prefer' header on requests to Azure Monitor datasource

On HTTP requests to the Azure Monitor data source, include a request header

Prefer: wait=time-to-wait-in-seconds

Pass through setting as configured by _dataProxy.timeout_

**Why is this needed**:

As per the documentation, this allows you to set the maximum timeout when calling the API. 

https://dev.loganalytics.io/documentation/Using-the-API/Timeouts

This will help ensure timeouts are correctly honoured across system-boundaries.

dylan-asos:

Adds a Prefer header to Azure Monitor data source HTTP requests, which allows passing the Data Proxy timeout setting to the Azure Monitor API.

For example, the default data proxy time of '30 seconds' will be passed as 

Prefer: wait=30

which will reduce the query timeout [from 3 minutes to 30 seconds](https://dev.loganalytics.io/documentation/Using-the-API/Timeouts).

Fixes #28199

### DIFF
--- a/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
+++ b/pkg/tsdb/azuremonitor/azuremonitor-datasource.go
@@ -248,6 +248,7 @@ func (e *AzureMonitorDatasource) createRequest(ctx context.Context, dsInfo *mode
 
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("User-Agent", fmt.Sprintf("Grafana/%s", setting.BuildVersion))
+	req.Header.Set("Prefer", fmt.Sprintf("wait=%d", setting.DataProxyTimeout))
 
 	pluginproxy.ApplyRoute(ctx, req, proxyPass, azureMonitorRoute, dsInfo)
 
--- a/pkg/tsdb/azuremonitor/applicationinsights-datasource.go
+++ b/pkg/tsdb/azuremonitor/applicationinsights-datasource.go
@@ -229,6 +229,7 @@ func (e *ApplicationInsightsDatasource) createRequest(ctx context.Context, dsInf
 	}
 
 	req.Header.Set("User-Agent", fmt.Sprintf("Grafana/%s", setting.BuildVersion))
+	req.Header.Set("Prefer", fmt.Sprintf("wait=%d", setting.DataProxyTimeout))
 
 	pluginproxy.ApplyRoute(ctx, req, proxyPass, appInsightsRoute, dsInfo)
 
--- a/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
+++ b/pkg/tsdb/azuremonitor/azure-log-analytics-datasource.go
@@ -212,6 +212,7 @@ func (e *AzureLogAnalyticsDatasource) createRequest(ctx context.Context, dsInfo 
 
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("User-Agent", fmt.Sprintf("Grafana/%s", setting.BuildVersion))
+	req.Header.Set("Prefer", fmt.Sprintf("wait=%d", setting.DataProxyTimeout))
 
 	// find plugin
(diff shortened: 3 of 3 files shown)


aknuds1:
Does it make sense to include tests for this?

pkg/tsdb/azuremonitor/azuremonitor-datasource.go
```diff
@@ -248,6 +248,7 @@ func (e *AzureMonitorDatasource) createRequest(ctx context.Context, dsInfo *mode
 
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("User-Agent", fmt.Sprintf("Grafana/%s", setting.BuildVersion))
+	req.Header.Set("Prefer", fmt.Sprintf("wait=%d", setting.DataProxyTimeout))
```
kylebrandt, line 251: Are you sure that this applies to the azure monitor metrics service / api? I do see documented for the analytics and insights. So perhaps it should be removed here?

Also this should perhaps also be added to insights-analytics-datasource.go ?

kylebrandt (changes requested):
I wondering if you can maybe be more specific about the behavior change you are after?

Generally, I don't think this will change the behavior. Because:
 1. We send a request, and that request will have a timeout of X
 2. The remote azure system after getting the request will set the same timeout of X

Given that 1 starts before 2, the timeout on the request in Grafana is going to hit before the timeout on the server side. Therefore we are still hitting the deadline exceeded on the client and will not get the gateway timeout from the server. I'm not sure if this will stop the query on Azure side or not (as I am not sure what happens when the deadline is exceeded, rst maybe?) 

So if we really need this depending on what it gets us, we probably need a datasource timeout setting, and maybe warn if that is equal or less than the datasource proxy

pkg/tsdb/azuremonitor/azuremonitor-datasource.go
dylan-asos: I can only find it documented for analytics and insights - the metrics API would be less prone to variations in query times, so yes will remove from here + look at insights-analytics-datasource

kylebrandt:
@dylan-asos Thank you for the additional info, in particular the query audition information. So I think it is a good change, but the  ... sort of race condition? Still needs to be resolved. Basically, the server side timeout needs to be some amount less than the client timeout, or we will still get `client.Timeout execeed' error message because the client will timeout before the server if they are the same.

dylan-asos:
@kylebrandt yep makes sense. keep it simple, wait time = (proxy timeout - 1 sec)?

kylebrandt:
> @kylebrandt yep makes sense. keep it simple, wait time = (proxy timeout - 1 sec)?

I think it should be an data source setting because:
 - Do not know what people's latency are from their Grafana instance to azure, 1 second might not be long enough
 - One may wish to have different data sources with timeouts for different cases. For example, a group of users that is allowed to make long running queries, and a group that is not (and Enterprise has data source permissions).

For the setting, I think we should:
 - Make it optional configuration, if not defined default the header to dataproxy timeout-1s. (edge case, do not set header if that would make the timeout 0).
 - Warn in the datasource settings save validation if it is equal to or greater than the data proxy timeout
 - If at runtime it is equal to or greater than, override it and log a warning (to handle the case where someone may change the system setting in the future, without being aware of the data source setting)

dylan-asos:
@kylebrandt all sounds good, but becomes a bigger piece of work than this PR in it's current state - will take a little while.  

Will mark this as WIP.

cheekp:
Is it now possible to configure the Prefer wait for Log Analytics API calls some how. The data proxy timeout does not seem to have an effect in v8.2.5. At a 30 seconds timeout we only see 499 errors, so the root cause or extent of the query delay against the API is not known.

PR #28216, opened 2020-10-13
Card: Add new Card component

### LINKED ISSUE #28209: Grafana-UI: Add Card component
Implement the [designs](https://www.figma.com/file/ITCsaXtsANM17oRtU6GQS9/Component-Playground?node-id=712%3A0) for Card/List component and add it to grafana-ui.

Clarity-89:
- Implement designs for the Card component.
- use the new Card component for `AlertRuleItem`

Fixes https://github.com/grafana/grafana/issues/28209

Screenshots:

##### Full card view
Screenshot 2020-10-14 at 16 58 41

Screenshot 2020-10-14 at 16 59 19

###### Basic

Screenshot 2020-10-14 at 16 58 50
Screenshot 2020-10-14 at 17 42 25

##### With tags
Screenshot 2020-10-14 at 16 59 40
Screenshot 2020-10-14 at 16 59 55

##### Alert rule item list
Screenshot 2020-10-14 at 15 58 39
Screenshot 2020-10-14 at 17 00 29


packages/grafana-ui/src/components/Card/Card.mdx
```diff
@@ -0,0 +1,277 @@
+import { Meta, Preview, Props } from "@storybook/addon-docs/blocks";
+import { Card } from "./Card";
+import { Button } from '../Button';
+import { IconButton } from '../IconButton/IconButton';
+import promLogo from '../../../../../public/app/plugins/datasource/prometheus/img/prometheus_logo.svg';
```
dprokop, line 5: I'm not 100% sure this will allow storybook to be built? Maybe instead you could use something like https://placeholder.com/

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+      {children}
+    </a>
+  ) : (
+    <>{children}</>
+  );
+};
+
+export interface Props extends ContainerProps {
+  /** Main heading for the Card **/
+  heading: ReactNode;
```
dprokop, line 45:
```suggestion
  title: ReactNode;
```

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /** Additional data about the card. If array is supplied, elements will be rendered with vertical line separator */
+  metaData?: ReactNode | ReactNode[];
```
dprokop, line 47:
```suggestion
  metadata?: ReactNode | ReactNode[];
```

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /** Card description text */
+  description?: string;
+  /** List of tags to display in the card */
+  tags?: string[];
+  /** Optional callback for tag onclick event */
+  onTagClick?: OnTagClick;
+  /** Indicates if the card and all its actions can be interacted with */
+  disabled?: boolean;
+  /** Image or icon to be displayed on the let side of the card */
+  mediaContent?: ReactNode;
```
dprokop, line 57: Not a fan of the `mediaContent` name. Maybe just `image`?

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+  tooltip,
+  secondaryActions = [],
+  tag,
+  href,
+  onClick,
+  className,
+  ...htmlProps
+}) => {
+  const hasActions = Boolean(actions.length || secondaryActions.length);
+  const disableHover = disabled || actions.length > 1;
```
dprokop, line 86: Is this correct? I see BasicCard story has a pointer cursor, which does not make much sense. I think you should also check for `href` property here.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+            </div>
+          )}
+        </div>
+      </CardInner>
+      <div className={styles.overlay} />
+    </CardContainer>
+  );
+};
+
+const getStyles = (theme: GrafanaTheme, disabled = false, disableHover = false) => {
```
dprokop, line 131: Please use `useStyles` hook or `stylesFactory` for memoization

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+  className,
+  ...htmlProps
+}) => {
+  const hasActions = Boolean(actions.length || secondaryActions.length);
+  const disableHover = disabled || actions.length > 1;
+  const theme = useTheme();
+  const styles = getStyles(theme, disabled && !actions.length, disableHover);
+  // Join meta data elements by '|'
+  const meta = Array.isArray(metaData)
+    ? (metaData as ReactNode[]).filter(Boolean).reduce((prev, curr, i) => [
```
dprokop, line 91: Why are fou using `filter` here?

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+import { css, cx } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { useTheme, styleMixins } from '../../themes';
+import { Tooltip, PopoverContent } from '../Tooltip/Tooltip';
+import { OnTagClick } from '../Tags/Tag';
+import { TagList } from '../Tags/TagList';
+
+interface ContainerProps extends HTMLAttributes<HTMLOrSVGElement> {
+  /** Customise the container html element for the card. Defaults to div */
+  tag?: keyof JSX.IntrinsicElements;
```
dprokop, line 11: I'm not convinced about this customization. IMO Card should always be div, and additional behaviours / layout should be transparent to card. By this I especially mean that we should do:

```tsx
<ul>
  <li><Card .... /></li>
  ...
</ul>
```
Rather than:
```tsx
<ul>
  <Card tag="li" .... />
  ...
</ul>
```

It's easier to reason about and removes the unnecessary "complexity" of the CardContainer component.

@jackw WDYT?

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+  onClick,
+  className,
+  ...htmlProps
+}) => {
+  const hasActions = Boolean(actions.length || secondaryActions.length);
+  const disableHover = disabled || actions.length > 1;
+  const theme = useTheme();
+  const styles = getStyles(theme, disabled && !actions.length, disableHover);
+  // Join meta data elements by '|'
+  const meta = Array.isArray(metaData)
```
dprokop, line 90: Think this reduce is `useMemo` worthy.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,222 @@
...
+    ? (metaData as ReactNode[]).filter(Boolean).reduce((prev, curr, i) => [
+        prev,
+        <span key={`separator_${i}`} className={styles.separator}>
+          |
+        </span>,
+        curr,
+      ])
+    : metaData;
+  const onCardClick = disableHover ? () => {} : onClick;
```
dprokop, line 99:
```suggestion
  const onCardClick = useCallback((e) => {
    if(disableHover) {
      return null;
    }
    
    onClick(e)
  } , [onClick, disableHover]);
```

or sth similar. Otherwise, you are creating noop with every render.

dprokop (commented):
Looks good @Clarity-89, great work. I've left a few comments worth considering.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: `title` is an [html attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) so that's why I decided not to override it here.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: Yeah, altho it could be also an icon, but maybe image could work.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: Since metaData can be an array, there might be smth like `metaData={['Prometheus', condition && 'some string']}` and if the condition is false the element will be undefined but it's still an array element and will create an unnecessary separator. So this way we filter falsy values from the array.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: The argument makes sense but at the same time I'm not a huge fan of this added layout complexity in case we need to render card as a list element.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: Good catch1 It'd also check if `onClick` is provided.

packages/grafana-ui/src/components/Card/Card.tsx
dprokop: Ah, you are right! Thanks

packages/grafana-ui/src/components/Card/Card.mdx
Clarity-89: Just ran `yarn storybook:build` and it was successful, not sure if it's gonna work in CI tho 🤔

dprokop: It will if the build is OK locally. But I still think **we should not reference core** from grafana-ui.

packages/grafana-ui/src/components/Card/Card.tsx
dprokop: Yeah, but with the `tag` property you are introducing a new pattern to the API. I cannot recall any situation when we allow DOM customization this way and we always delegate the layout responsibility to the outer component rather than encapsulating it within a component. 

I suggest dropping this property and plain div always for the card.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <div className={styles.actionRow}>
+              {!!actions.length && (
+                <div className={styles.actions}>{actions.map(action => cloneElement(action, { disabled }))}</div>
+              )}
+              {!!secondaryActions.length && <div className={styles.secondaryActions}>{secondaryActions}</div>}
+            </div>
+          )}
+        </div>
+      </CardInner>
+      <div className={styles.overlay} />
```
dprokop, line 131: What is this `overlay`?

Clarity-89: It's used for disabling the whole card and also giving it a disabled background.
Screenshot 2020-10-20 at 11 44 59

dprokop: Ah, ok. Do we need an additional dom element for that? Think you could just apply the appropriate styling to the root dom element of the card, couldn't you?

Clarity-89: Yeah, but we also need to disable all click events on the underlined elements, for ex. if the there's no or only one primary action. Could try to use pseudo elements for that but don't think it'll work.

dprokop: what about `pointer-events: none`?

Clarity-89: Good point! There's already `pointer-events: none` for the container in disabled state, so I guess just styling `::after` pseudo-selector instead of using an additional div will work here.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,227 @@
...
+          <div className={styles.heading}>{heading}</div>
+          {meta && <div className={styles.metadata}>{meta}</div>}
+          {!!tags.length && <TagList tags={tags} onClick={onTagClick} className={styles.tagList} />}
+          {description && <p className={styles.description}>{description}</p>}
+          {hasActions && (
+            <div className={styles.actionRow}>
+              {!!actions.length && (
+                <div className={styles.actions}>{actions.map(action => cloneElement(action, { disabled }))}</div>
+              )}
+              {!!secondaryActions.length && <div className={styles.secondaryActions}>{secondaryActions}</div>}
```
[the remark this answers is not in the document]
Clarity-89: This whole disable logic gets quite complicated, have to check if it can be simplified.

packages/grafana-ui/src/components/Card/Card.tsx
jackw: > delegate the layout responsibility to the outer component rather than encapsulating it within a component
I'm a little late to the table here but I just wanted to say I agree with this approach. The less it knows the better. If the card needs to be rendered in a list, I'd expect a List or ListItem component to handle that piece of layout.

torkelo: agree with Dominik that controlling what tag is used feels unnecessary and not something we do with any other component

torkelo:
Do we need this in 7.3? (just wonder why it's in 7.3 milestone)

Clarity-89:
> Do we need this in 7.3? (just wonder why it's in 7.3 milestone)

Would be nice to have it in 7.3 so we can start using it in some of the enterprise app plugins.

torkelo:
ah, ok. That makes sense. But this updates the Alert list to use the new cards. So not ideal to merge late stage beta. Need to double check that we are not introducing any bugs in the alert list

jessover9000:
Testing the design, I wonder: Is it just me or is the new card design only implemented in the alerting area so far?

My feedback on the alerting area:
- I find the tooltips on the action buttons a bit jarring, especially because one opens to the bottom and the other to the right. Do we really need them here now that we have button text?
- Icon alignment: Why is there added spacing to the left between margin and actual icon? The div around the icon is 40px wide, I wonder why because it looks like it's messing with the alignment. I think it would look better if the spacing would be equal to the left and the top. If the bounding box fit the icon, the spacing between icon and card text would also be smaller which I think would be okay – right now the asymmetrical whitespace looks a bit wrong to me.

Other than that, it looks good! Well done, Alex 👍

Clarity-89:
> ah, ok. That makes sense. But this updates the Alert list to use the new cards. So not ideal to merge late stage beta. Need to double check that we are not introducing any bugs in the alert list

I can move the AlertRule commits to a separate PR, so they can be reviewed later?

Clarity-89:
> Is it just me or is the new card design only implemented in the alerting area so far?

Yeah, I only updated the AlertRuleItem to test out the new component.

> * I find the tooltips on the action buttons a bit jarring, especially because one opens to the bottom and the other to the right. Do we really need them here now that we have button text?

Good point, I don't think those tooltips are needed anymore.

> * Icon alignment: Why is there added spacing to the left between margin and actual icon? The div around the icon is 40px wide, I wonder why because it looks like it's messing with the alignment. I think it would look better if the spacing would be equal to the left and the top. If the bounding box fit the icon, the spacing between icon and card text would also be smaller which I think would be okay – right now the asymmetrical whitespace looks a bit wrong to me.

Will need to check this, but I think there might be some spacing coming from the icon component itself.

mckn:
Great work @Clarity-89! Would it be possible to split the component into several smaller ones? I guess this depends on how they will be used but it would be nice to get several components with simpler APIs.

But maybe have a look and see if it is possible to create an ActionCard and make Card a bit simpler. I don't know if this makes sense. I am just thinking out loud here :)

torkelo (changes requested):
See that this changes the secondary button design as well, need some time to test this more.

torkelo:
@mckn I am not sure this should be multiple components, the whole point is to have one resource card component that is very generic & extensible so we can have consistency in designs of item lists. The API does not look that complex to me, are you proposing having multiple different Card components like CardWithActions, CardWithImage CardWithImageAndActions , that would be messy?, or do you have something else in mind for how to break this up?  

I guess the other approach would be different child components. 

```jsx
<Card header="Hello">
  <Card.Figure>
    ....
  </Card.Figure>
  <Card.Badges>
   ...
  </Card.Badges>
  <Card.Actions>
    ...
  </Card.Actions>
<Card>
```

That could be an option, and for the Main card component to look up children and check type and place then in the right place in the markup tree

Clarity-89:
> @mckn I am not sure this should be multiple components, the whole point is to have one resource card component that is very generic & extensible so we can have consistency in designs of item lists. The API does not look that complex to me, are you proposing having multiple different Card components like CardWithActions, CardWithImage CardWithImageAndActions , that would be messy?, or do you have something else in mind for how to break this up?

Yeah, I also don't think we need multiple components for that. If you need an AlertCard, you can use the card component and provide it all the necessary props for Alert, if that is then used in multiple places, we can add it to grafana-ui as a separate component, based on top of Card.

> I guess the other approach would be different child components.
> 
> ```js-jsx
> <Card header="Hello">
>   <Card.Figure>
>     ....
>   </Card.Figure>
>   <Card.Badges>
>    ...
>   </Card.Badges>
>   <Card.Actions>
>     ...
>   </Card.Actions>
> <Card>
> ```
> 
> That could be an option, and for the Main card component to look up children and check type and place then in the right place in the markup tree

Agree, this could work, abstracting some of the props into separate components seems to be cleaner, at least at the first glance.  It did feel a bit weird to me that card doesn't accept `children`, but I didn't see a good place to use them.

torkelo (commented):
A) New dark button
The new dark button continue to now not working for me :(  

It's not that I think the current gray button is great, but something about this black gradient button feels out of place and is inconsistent with the reset of design and buttons. And is inconsistent with how secondary button looks in the light theme. 

B) Tooltips
I really do not like our tooltip component, and especially here where it just says same thing as the button, hover state makes button look disabled, and a tooltip appears with same word. Remove tootip or replace with native title 
Screenshot from 2020-10-26 15-37-49

Alternative to black button, tweaked gray button that stands out a bit more:
Screenshot from 2020-10-26 16-18-33

Clarity-89:
> A) New dark button
> The new dark button continue to now not working for me :(
> 
> It's not that I think the current gray button is great, but something about this black gradient button feels out of place and is inconsistent with the reset of design and buttons. And is inconsistent with how secondary button looks in the light theme.

I think the reasoning fo the new design was that the current secondary button blends with the card's background.

> B) Tooltips
> I really do not like our tooltip component, and especially here where it just says same thing as the button, hover state makes button look disabled, and a tooltip appears with same word. Remove tootip or replace with native title
> Screenshot from 2020-10-26 15-37-49

Jess also commented about this and i'll just remove the tooltip from alert items.

> Alternative to black button, tweaked gray button that stands out a bit more:
> Screenshot from 2020-10-26 16-18-33

@jessover9000 any feedback on this?

torkelo:
Testing this design for real now not just via figma makes me think that maybe these cards should be more responsive and have different layout for wide screens. The layout is really good for smaller screens (super important), but look bad with so much horizontal space wasted 

Screenshot from 2020-10-27 12-57-31

Clarity-89:
Seems like this needs more tweaking and discussion, so probably better to remove it from 7.3. milestone.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,245 @@
...
+  /** Link to redirect to on card click. If provided, the Card inner content will be rendered inside `a` */
+  href?: string;
+  /** On click handler for the Card */
+  onClick?: () => void;
+}
+
+/**
+ * Generic card component
+ *
+ * @public
```
torkelo, line 77: can we change this to alpha so we can play around with it more before we mark this a beta or public

torkelo (approved):
But I do think this is mergable, we can iterate on implementation and design. 

Just maybe not mark it the component as public

Clarity-89:
> But I do think this is mergable, we can iterate on implementation and design.
> 
> Just maybe not mark it the component as public

Ok, but what about the AlertRuleItem then? Should I move it to a separate PR?

torkelo:
No let's keep it but let's keep it 7.4 so we can iterate on this

packages/grafana-ui/src/components/Card/Card.mdx
```diff
@@ -0,0 +1,282 @@
...
+  tags={['test', 'data', 'testdata']}
+/>
+```
+
+<Preview>
+  <Card
+    heading="Test dashboard"
+    metadata={['Folder: Test', 'Views: 100']}
+    tags={['test', 'data', 'testdata']}
+    onTagClick={(tag) => console.log('clicked tag:', tag) }
```
torkelo, line 45: this onTagClick is feels very wrong to me, clear indicator this is best handled via child component that uses Tag component direct?

Clarity-89: Agree, switching to composing child components sounds like a good idea here.

Issue #6877, opened 2016-12-08
[Feature request] Add query builder to make template variables with complex datasource queries


drewboswell:
Datasources used: Graphite, Elasticsearch, Zabbix, InfluxDB
OS: Ubuntu, SLES, Solaris x64, Solaris Sparc v9

When doing queries on graphite, template queries are easy. For other datasources like elasticsearch providing the lucene query is hard, or other brutal queries is hard. It would be ideal to add the query builders to the templating panel, just like they are in the main queries in panels.

hugohaggmark:
This feature request is one step closer to being implemented with #28217.

pkolyvas:
We're in the process of moving Grafana feature requests to [GitHub Discussions](https://community.grafana.com/t/were-moving-grafana-feature-requests-to-github-discussions/50341/3). I'm closing feature requests that have seen limited from the community and/or the Grafana team. 

Should this feature still be of interest to you, please re-create it in our [Discussions area](https://github.com/grafana/grafana/discussions). In doing so the community can continue to contribute to exploring, confirming and sharing their use cases (and more).

As with all things Grafana, we value your input and participation in this project and appreciate your patience during this transitional period. Should you wish to connect to discuss anything Grafana-related, I have an open door policy and [my contact details are on my GitHub profile page](https://github.com/pkolyvas).

Issue #18105, opened 2019-07-12
Variables do not protect against self-reference when using ${var_name:option} syntax


CRBarber3:
**What happened**:

I accidentally created a variable that referenced itself using `${var_name:option}` syntax. 
This causes the dashboard to get stuck loading if you attempt to open it with the self-referencing variable. (loading wheel)Services 

The real issue is that I don't see any way to recover the dashboard if this occurs. (from the frontend at least)

**What you expected to happen**:

There appears to be protection from self-referencing variables if you just directly put `$var_name` in the query. I would expect this protection to extend to the `${var_name:option}` syntax.

i.e. I would expect to see this:
![image](https://user-images.githubusercontent.com/14188213/61163347-6f15bf00-a4db-11e9-8615-c81e40062964.png)

**How to reproduce it (as minimally and precisely as possible)**:

1. Create new dashboard
2. Dashboard Settings
3. Variables > Add Variable
    Name: test
    Type: Query
    Data source: --Grafana--
    Query: `SELECT ${test:csv}`
4. Add
5. Save (give it a name)
6. Save

**Anything else we need to know?**:

This is a user issue but I feel the protection is important because the dashboard appears unrecoverable (at least as far as I can tell, from the front end).

**Environment**:
- Grafana version: Grafana v6.2.5 (6082d19)
- Data source type & version: PostgreSQL (and the default possibly all)
- OS Grafana is installed on: Windows 7/10
- User OS & Browser: Windows 7/10
- Grafana plugins: none
- Others:

hugohaggmark:
We have introduced this check, not sure in which version of Grafana but I can't repro it on Grafana v7.4.0-pre (db637a3630). This check will only work for data sources that represent their query with a string. This will check will not work for data sources that represent their query with complex objects. 

Thank You to you for taking the time to create this issue!

PR #28217, opened 2020-10-13
Variables: Adds new Api that allows proper QueryEditors for Query variables

### LINKED ISSUE #27746: Variables: Make variable queries cancellable with Observables

hugohaggmark:
So after a discussion with Torkel the proposed solution in #27752 didn't check all the boxes. So I restarted a bit and this is what I came up with, sorry for the very rough PR there's lots of things to polish.

If you look in TestData, CloudMonitoring & Prometheus data sources you can see examples of the usage. The typing is working and enforces data source plugins in the right direction. 

I started with a Variable system specific __QueryEditorProps__ but I found that to be britle moving forward so instead I made this change to QueryEditorProps to be able to support another query besides the one defined in the DataSourceApi:
image

Fixes #27746
Relates #6877


packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,96 @@
...
+    datasource: {
+      editor: ComponentType<VariableQueryEditorProps<any, TQuery, TOptions, TQuery>>;
+    };
+  };
+}
+
+export interface VariableQueryEditorProps<
+  DSType extends DataSourceApi<TQuery, TOptions>,
+  TQuery extends DataQuery = DataQuery,
+  TOptions extends DataSourceJsonData = DataSourceJsonData,
```
torkelo, line 86: think you should be able to infer these two

torkelo: see https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/datasource.ts#L20

hugohaggmark:
This is still missing a lot of tests so I'm going those tomorrow :)

hugohaggmark:
__Some thoughts__

- The experimental group/tags
image still use `metricfindQuery` in the data sources, this isn't super clean any suggestions? I'd love to deprecate the experimental group/tags soon

- This PR will upgrade variable queries for TestData & Promethues so they become proper objects of DataQuery instead of string. This means there is no downgrading these queries for previous versions of Grafana. 

- This PR will not make variable queries for Prometheus och Cloudmonitoring cancellable all the way. Only the observables are cancelled not the network request. This will have to fixed in a followup PR

public/app/plugins/datasource/prometheus/variables.ts
```diff
@@ -0,0 +1,55 @@
...
+  StandardVariableSupport,
+} from '@grafana/data';
+import { getTemplateSrv, TemplateSrv } from '@grafana/runtime';
+
+import { PrometheusDatasource } from './datasource';
+import { PromQuery } from './types';
+import PrometheusMetricFindQuery from './metric_find_query';
+import { getTimeSrv, TimeSrv } from '../../../features/dashboard/services/TimeSrv';
+
+export class PrometheusVariableSupport implements StandardVariableSupport<PrometheusDatasource> {
```
aocenas, line 17: Is there some comparison what it looked like before? Or didn't Prometheus support variables query before?

packages/grafana-data/src/types/datasource.ts
```diff
@@ -310,12 +320,13 @@ export interface MetadataInspectorProps<
...
-  TOptions extends DataSourceJsonData = DataSourceJsonData
+  TOptions extends DataSourceJsonData = DataSourceJsonData,
+  TVQuery extends DataQuery = TQuery
 > {
   datasource: DSType;
-  query: TQuery;
+  query: TVQuery;
   onRunQuery: () => void;
-  onChange: (value: TQuery) => void;
+  onChange: (value: TVQuery) => void;
```
aocenas, line 329: This looks a bit weird to me. Do we need both TQuery and TVQuery? Like is there any moment you would specify both? I assume not because in that case TQuery won't be actually used anywhere.

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,93 @@
...
+> {
+  /**
+   * Use this property if you want to use the standard query editor supplied in the Variable system
+   */
+  standard?: StandardVariableSupport<DSType, TQuery, TOptions>;
+
+  /**
+   * Use this property if you want to use a custom query and query editor
+   */
+  custom?: CustomVariableSupport<DSType, VariableQuery, TQuery, TOptions>;
```
aocenas, line 36: In addition to this do you have to supply the component somewhere?

aocenas:
Maybe I am totally off, not sure I understand the code much, but couldn't we use something like this https://github.com/grafana/grafana/blob/fe423644aef7b307b8222019023deae555d83143/packages/grafana-data/src/types/datasource.ts#L457 to mark request as coming from variables config page so DS can decide whether to treat it in different way? I assume you can opt into that by providing custom editor. Do we also need an opt in for the legacy editor?

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Use this property if you want to use the same query editor as the data source
+   */
+  datasource?: DataSourceVariableSupport<DSType, TQuery, TOptions>;
```
aocenas, line 41: Do we need this? You can just supply the same editor as a custom editor right? That should do the same thing doesn't it?

public/app/plugins/datasource/prometheus/variables.ts
hugohaggmark: This is brand new, Prometheus used metricFindQuery for the variable queries before. [With this change we still use metricFindQuery for the experimental tags query see my comment](https://github.com/grafana/grafana/pull/28217#issuecomment-714232445).

packages/grafana-data/src/types/datasource.ts
hugohaggmark: You're right, but unfortunately TQuery is inferred by DataSourceAPI so you can't tell the QueryEditor to use a different query than the one declared in DataSourceAPI. We need this for custom variable support where you can have a custom TVQuery. 

But I'm opened for other ways to solve this. I started by creating my own QueryEditorProps for variables, but that would mean they would have to be in sync.

packages/grafana-data/src/types/variables.ts
hugohaggmark: Not sure I understand your question. The interface tells you need to define an editor and a custom query method like so:
```ts
export interface CustomVariableSupport<
  DSType extends DataSourceApi<TQuery, TOptions>,
  VariableQuery extends DataQuery = any,
  TQuery extends DataQuery = DataSourceQueryType<DSType>,
  TOptions extends DataSourceJsonData = DataSourceOptionsType<DSType>
> {
  editor: ComponentType<QueryEditorProps<DSType, TQuery, TOptions, VariableQuery>>;
  query: (request: DataQueryRequest<VariableQuery>) => Observable<DataQueryResponse>;
}
```

packages/grafana-data/src/types/variables.ts
hugohaggmark: I see what you mean but not really, because for the data source variable support we'll use the `query` function from the data source so you only need to declare the editor. An improvement I can see is that we could probably resolve the editor from the data source plugin instead.

hugohaggmark:
> Maybe I am totally off, not sure I understand the code much, but couldn't we use something like this

So depending on which variable support the data source uses, the variable system will adapt and call the "right" query method. The "magic" is in the [queryRunners.ts](https://github.com/grafana/grafana/blob/cdebe1e7f56ad4b49e24b15d82cd2dab335a308d/public/app/features/variables/query/queryRunners.ts) and those are used in 
[VariableQueryRunner](https://github.com/grafana/grafana/blob/cdebe1e7f56ad4b49e24b15d82cd2dab335a308d/public/app/features/variables/query/variableQueryRunner.ts#L114)


> 
> https://github.com/grafana/grafana/blob/fe423644aef7b307b8222019023deae555d83143/packages/grafana-data/src/types/datasource.ts#L457

Yes, I started with this [in my last PR ](https://github.com/grafana/grafana/pull/27752) but we decided CoreApp shouldn't be used for this.

> 
> to mark request as coming from variables config page so DS can decide whether to treat it in different way? I assume you can opt into that by providing custom editor. Do we also need an opt in for the legacy editor?

Legacy should just work, no need to opt in, so if you don't upgrade to the new api you would have [to use the legacy way](https://github.com/grafana/grafana/blob/cdebe1e7f56ad4b49e24b15d82cd2dab335a308d/packages/grafana-data/src/types/datasource.ts#L86) of adding any custom QueryEditor.

packages/grafana-data/src/types/variables.ts
dprokop, line 36: @hugohaggmark would it make any sense, instead of having standard, custom and data source properties here, to have this interface defined as: 

```ts
export interface VariableSupport<
  DSType extends DataSourceApi<TQuery, TOptions>,
  TQuery extends DataQuery = DataSourceQueryType<DSType>,
  TOptions extends DataSourceJsonData = DataSourceOptionsType<DSType>,
  VariableQuery extends DataQuery = any
> {
  type: StandardVariableSupport<...> | CustomVariableSupport<...> | DataSourceVariableSupport<...>
}
```
 I imagine this will require some casting and typeguards down the road, but we would avoid complexity of deciding about the precedence of configuration. 

I'm thinking a bit from the POV of a consumer of the API who has to decide what kind of variable support he/she needs and can get confused with multiple options available as top-level properties of this interface. For me, if I don't know the underlying logic of the precedence of configured variables support, _this API is an invitation to configure more than one property_ and get surprised by the editor being chosen as the correct one in the UI.  I hope you understand what I mean.

dprokop (commented):
@hugohaggmark this is a massive work, amazing! I think I need to to spend a little bit more time with this PR, but after looking at it briefly I find the entire proposal understandable. I have left one concern I have about the VariablesSupport that you may find worth considering (if you haven't considered it already).

packages/grafana-data/src/types/variables.ts
hugohaggmark: Yes, I totally understand what you mean and that is something I thought about. I think we would need to add a `type` prop in each sub variable support as well to make sure the guards work. I think this is worth considering, so I'll look into this next week. Thank you @dprokop!

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Implement this interface in a data source plugin to use the query editor in the data source plugin for Query variables
+ *
+ * @alpha -- experimental
+ */
+export interface DataSourceVariableSupport<
+  DSType extends DataSourceApi<TQuery, TOptions>,
+  TQuery extends DataQuery = DataSourceQueryType<DSType>,
+  TOptions extends DataSourceJsonData = DataSourceOptionsType<DSType>
+> {
+  editor: ComponentType<QueryEditorProps<DSType, TQuery, TOptions, TQuery>>;
```
torkelo, line 83: this could just be a boolean, useDefaultQueryEditor?

torkelo (commented):
Took time to review this now, it's a lot to take in :) but looks really good. 

A) Some concern in the increased complexity QueryEditorProps type and the overall complexity of VariableSupport that will make it pretty hard for plugin authors to know what to do. Not sure what our options are though. A fluent api & class for VariableSupport that guides user on what options they have? But the comments on the interfaces are good so might be good enough. 

B) Since StandardVariableSupport uses LegacyQueryEditor, should StandardVariableSupport maybe be LegacyVariableSupport? 

Need to dig into this a big more, but awesome job!

daniellee (approved):
This is just a shallow review. As requested, I tested this with the Stackdriver/Cloud Monitoring data source and it works perfectly. The code changes to the CloudMonitoring data source look good to me too.

packages/grafana-data/src/types/variables.ts
hugohaggmark: @dprokop I've tried out the suggested approach by you and I think it makes the api a bit clearer.

packages/grafana-data/src/types/variables.ts
hugohaggmark: @torkelo I've tried out the suggested approach by Dominik and I think it makes the api a bit clearer.

public/app/features/variables/query/variableQueryRunner.ts
```diff
@@ -0,0 +1,199 @@
...
+      runner
+        .runRequest(runnerArgs, request)
+        .pipe(
+          filter(() => {
+            // lets check if we started another batch during the execution of the observable. If so we just want to abort the rest.
+            const afterUid = getState().templating.transaction.uid;
+            return beforeUid === afterUid;
+          }),
+          first(data => data.state === LoadingState.Done || data.state === LoadingState.Error),
+          toMetricFindValues(),
```
dprokop, line 127: If I understand the flow correctly- when the query results in an error state- we will still go through the entire pipe here, right? Is there a way to avoid that somehow and handle the error earlier?

dprokop (commented):
Just wow, the amount of improvements here is massive. I also think the API with 
```
variables?:
    | StandardVariableSupport<DataSourceApi<TQuery, TOptions>>
    | CustomVariableSupport<DataSourceApi<TQuery, TOptions>>
    | DataSourceVariableSupport<DataSourceApi<TQuery, TOptions>>;
```

is easier to digest now:) Great work @hugohaggmark ! Left few nits you may consider or ignor if you disagree :)

public/app/features/variables/query/queryRunners.ts
```diff
@@ -0,0 +1,171 @@
...
+      return variable.query;
+    }
+
+    throw new Error("Couldn't create a target with supplied arguments.");
+  }
+
+  runRequest({ datasource, variable, searchFilter, timeSrv }: RunnerArgs, request: DataQueryRequest) {
+    if (!hasLegacyVariableSupport(datasource)) {
+      return getEmptyMetricFindValueObservable();
+    }
```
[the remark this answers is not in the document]
hugohaggmark: That was my initial intention, but then I'll not get the extra benefits from the TypeGuard with the correct type inferred. Same goes for the rest of these comments.

public/app/features/variables/query/variableQueryRunner.ts
hugohaggmark: Good catch, I didn't bother with data.state === LoadingState.Error in this PR. Will look into that!

public/app/features/variables/query/operators.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  describe('validateVariableSelection', () => {
+    describe('when called', () => {
+      it('then the correct observable should be created', done => {
+        const variable = queryBuilder()
+          .withId('query')
+          .build();
+        const dispatch = jest.fn().mockResolvedValue({});
+        const observable = of(undefined).pipe(
+          validateVariableSelection({ variable, dispatch, searchFilter: 'A search filter' })
+        );
```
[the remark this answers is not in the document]
hugohaggmark: I started off with private functions within VariableQueryRunner but then I needed to add tests so I made them public but finally I decided that they could be separate functions instead. Not sure this is a good pattern or not?

public/app/app.ts
```diff
@@ -47,10 +47,11 @@ import { reportPerformance } from './core/services/echo/EchoSrv';
 import { PerformanceBackend } from './core/services/echo/backends/PerformanceBackend';
 import 'app/routes/GrafanaCtrl';
 import 'app/features/all';
-import { getStandardFieldConfigs, getStandardOptionEditors, getScrollbarWidth } from '@grafana/ui';
+import { getScrollbarWidth, getStandardFieldConfigs, getStandardOptionEditors } from '@grafana/ui';
 import { getDefaultVariableAdapters, variableAdapters } from './features/variables/adapters';
 import { initDevFeatures } from './dev';
 import { getStandardTransformers } from 'app/core/utils/standardTransformers';
+import { setVariableQueryRunner, VariableQueryRunner } from './features/variables/query/variableQueryRunner';
```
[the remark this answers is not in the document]
hugohaggmark: Yepp, good catch!

public/app/features/dashboard/state/runRequest.ts
```diff
@@ -97,7 +97,11 @@ export function processResponsePacket(packet: DataQueryResponse, state: RunningQ
  *  Will emit a loading state if no response after 50ms
  *  Cancel any still running network requests on unsubscribe (using request.requestId)
  */
-export function runRequest(datasource: DataSourceApi, request: DataQueryRequest): Observable<PanelData> {
+export function runRequest(
+  datasource: DataSourceApi,
+  request: DataQueryRequest,
+  queryFunction?: typeof datasource.query
```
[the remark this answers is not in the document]
hugohaggmark: It does, I didn't think it would work either. Maybe it's a bug and I should use the proper type instead?

public/app/features/variables/editor/LegacyVariableQueryEditor.tsx
```diff
@@ -0,0 +1,48 @@
...
+    (event: React.FormEvent<HTMLTextAreaElement>) => {
+      onChange(event.currentTarget.value, event.currentTarget.value);
+    },
+    [onChange]
+  );
+
+  return (
+    <div className="gf-form">
+      <span className="gf-form-label width-10">Query</span>
+      <textarea
```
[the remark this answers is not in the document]
hugohaggmark: Good suggestion, thank you!

public/app/features/variables/pickers/shared/VariableLink.tsx
```diff
@@ -71,6 +72,22 @@ const VariableLinkText: FC<Pick<Props, 'tags' | 'text'>> = ({ tags, text }) => {
...
+    [onCancel]
+  );
+
+  return (
+    <Tooltip content="Cancel query">
+      <Icon className="spin-clockwise" name="sync" size="xs" onClick={onClick} />
+    </Tooltip>
+  );
+};
+
```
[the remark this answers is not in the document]
hugohaggmark: I'll have a look if it's a good fit.

packages/grafana-data/src/types/variables.ts
```diff
@@ -0,0 +1,68 @@
...
+  DataQueryRequest,
+  DataQueryResponse,
+  DataSourceApi,
+  DataSourceJsonData,
+  DataSourceOptionsType,
+  DataSourceQueryType,
+  QueryEditorProps,
+} from './datasource';
+
+export type VariableSupportType = 'standard' | 'custom' | 'datasource';
```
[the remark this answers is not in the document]
dprokop: I think the type becomes easier to use with enum. This may be a personal preference, but I find enums more discoverable at least with Webstorm intellisense.

Type:
image

Enum:
image

public/app/features/dashboard/state/runRequest.ts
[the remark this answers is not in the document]
dprokop: I don't think it's a bug. If it infers it correctly then it's great!

public/app/features/variables/query/queryRunners.ts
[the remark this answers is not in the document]
dprokop: I see, thanks for explaining.

public/app/features/variables/pickers/shared/VariableLink.tsx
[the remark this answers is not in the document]
hugohaggmark: Don't think this is a good fit right now.
image

packages/grafana-data/src/types/variables.ts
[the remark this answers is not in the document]
hugohaggmark: Switch to enums and changed interfaces to classes instead, so plugin authors automatically get correct type when selecting class to inherit. Let me know what you think.

public/app/features/variables/query/variableQueryRunner.ts
hugohaggmark: Fixed!

hugohaggmark:
Remember to fix the check related to #18105 before merging, as this will not work as is.

packages/grafana-data/src/types/datasource.ts
```diff
@@ -79,6 +80,9 @@ export class DataSourcePlugin<
     return this;
   }
 
+  /*
+   * @deprecated -- prefer using {@link CustomVariableSupport or DataSourceVariableSupport}
```
mckn, line 84:
```suggestion
   * @deprecated -- prefer using {@link CustomVariableSupport} or {@link DataSourceVariableSupport}
```

not 100% but I think we need to change the syntax like this.

mckn (approved):
Looks great! Go go go!

packages/grafana-data/src/types/datasource.ts
hugohaggmark: Ok, will look into it

torkelo (approved):
So much code & tests, all beautiful :) 

The problem is really tricky, so the solution here is also not trivial. Feel like it's a bit too complex than it needs to be but think the only way to simplify it would be to break backward compatabiltity and rewrite even more parts, so feel like this is a really good step, but wonder how we can simplify and unify the different modes more in the future and how that can simplify things (like force unified query editor and model).  Would be nice to get rid of the TVQuery on QueryEditorProps but see how that is needed as it stands

PR #28219, opened 2020-10-13
Plugins: Track plugin signing errors and expose them to the frontend

### LINKED ISSUE #27871: Plugins: report unsigned plugins in the UI
Currently when a datasource plugin has an invalid signature, the system logs an error and then ignores the plugin.  We can do better than that.

The plugin should be added in the UI and given a flag that says it is unsigned and will not run.  The UI should include instructions on how to fix it so it will run.

wbrowne:
Adds the possibility to track and capture plugin signing errors that are returned via the API, so that the frontend can do something useful with that information IE help a user figure out that something is wrong with their plugin and how they can fix it, etc.

With this change we track signing errors as high level error codes, which are returned through the API (new `/plugins/errors` endpoint). [Check out the frontend code that uses this new API.](https://github.com/grafana/grafana/pull/28469)

Fixes #27871


packages/grafana-data/src/types/plugin.ts
```diff
@@ -23,6 +23,12 @@ export enum PluginSignatureStatus {
   unsigned = 'unsigned', // no MANIFEST file
 }
 
+export enum PluginErrorCode {
+  invalid = 'invalid',
+  modified = 'modified',
```
ryantxu, line 28: are these about the signature?  lets make sure to scope the names to include that.
```suggestion
  invalidSignature = 'invalidSignature',
  signatureModified = 'signatureModified',
```

Other error stats we will want to expose will be:
* plugin health check is failing
* plugin is in some crash loop and we won't run it anymore
* (maybe) 'unavailable' or something like that -- when we expect an external gRPC connection that has not yet connected

packages/grafana-data/src/types/plugin.ts
```diff
@@ -23,6 +23,12 @@ export enum PluginSignatureStatus {
...
+  unsigned = 'unsigned',
```
ryantxu, line 29: as I think more... do we need different error codes for all the possible signature states?  Can we just have one that says "there is a problem with the signature" and then we look at the signature state (already available) for more details

pkg/plugins/error.go
```diff
@@ -0,0 +1,21 @@
+package plugins
+
+const (
+	UNSIGNED ErrorCode = iota
+	MODIFIED
+	INVALID
```
ryantxu, line 6: all three can be a single error code right?  and then we use the existing status for more details on the exact signature error.

packages/grafana-data/src/types/plugin.ts
wbrowne: Yeah I had thought about that and had opted for more specificity at first as it allows us to better guide the user - but if we can afford to simplify that then that's great!

pkg/api/plugins.go
```diff
@@ -112,6 +112,10 @@ func (hs *HTTPServer) GetPluginList(c *models.ReqContext) Response {
 			Signature:     pluginDef.Signature,
 		}
 
+		if pluginDef.Errors != nil {
```
aknuds1, line 115: It's generally recommended to check if a slice is empty rather than nil, since it will work both on empty and nil slices:
```suggestion
		if len(pluginDef.Errors) > 0 {
```

pkg/api/plugins.go
```diff
@@ -164,6 +168,10 @@ func GetPluginSettingByID(c *models.ReqContext) Response {
 		Signature:     def.Signature,
 	}
 
+	if def.Errors != nil {
```
aknuds1, line 171:
```suggestion
	if len(def.Errors) > 0 {
```

pkg/plugins/error.go
aknuds1, line 6: You shouldn't use all-caps for anything but initialisms. Where are these constants used?

```suggestion
	Unsigned ErrorCode = iota
	Modified
	Invalid
```

pkg/plugins/error.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+type ErrorCode int
+
+func (e ErrorCode) String() string {
+	return [...]string{"unsigned", "modified", "invalid"}[e]
```
aknuds1, line 12: I don't think it's safe to assume that `e` starts at 0. Also, what is the purpose of `...`? I tried without, and it's the same result. Never saw that syntax before.

```suggestion
	return []string{"unsigned", "modified", "invalid"}[e]
```

packages/grafana-data/src/types/plugin.ts
wbrowne: Ah I see what you mean now. I didn't want to start having this state depend on other information in this first pass really, since this is introducing a design change to the plugin flow which may lead to weird side effects. Treating this error code as the source of truth allows us to plug something on top without interfering too much with what's already there.

But technically yeah we could just roll them up into one and see how that works out. Will wait to hear what the other folks think

packages/grafana-data/src/types/plugin.ts
wbrowne: Makes sense. Will need them to match up with the backend enums too.

Regarding the other error states - they sound reasonable too. Can see them having a place here when necessary

aknuds1 (changes requested):
Backend generally looks good to me, but there are some issues.

pkg/api/plugins.go
wbrowne: This is a pointer so `len()` won't work here. Unless you believe this field type should change?

pkg/plugins/error.go
wbrowne: If that's the case then we have some inconsistencies. Maybe something we can add to the linter?

In this case I think they can be un-exported

pkg/plugins/error.go
wbrowne: This seemed to be a simpler solution than the pattern we seem to follow. For example:

```go
func (p permissionType) String() string {
	names := map[uint8]string{
		uint8(USER):   "user",
		uint8(TEAM):   "team",
		uint8(EDITOR): "editor role",
		uint8(VIEWER): "viewer role",
	}
	return names[uint8(p)]
}
```

It would be safe to assume `e` starts at 0 due to using `iota` as the base for the enum, no?

Aye so the purpose of `...` is the difference between is creating a slice vs an array. I want an array because this list will never change

pkg/api/plugins.go
aknuds1: @wbrowne Oh yeah, never mind. I couldn't tell the type from GitHub UI.

pkg/plugins/error.go
aknuds1: We would have to enable the typecheck linter, but it uncovers a ton of issues, so would be a huge amount of work.

pkg/plugins/error.go
aknuds1: It's not a simpler pattern since it assumes the following, which could break:

* The enum values start at 0 (this _could_ change)
* The enum values have a certain order (this could most definitely change)

Your suggestion just looks more complex than a switch statement. You also have to check if `p` is missing from the map. One benefit of the switch statement is that an analysis tool could tell if you're missing a branch.

pkg/plugins/models.go
```diff
@@ -74,6 +74,8 @@ type PluginBase struct {
 	GrafanaNetHasUpdate bool   `json:"-"`
 
 	Root *PluginBase
+
+	Errors *PluginErrors
```
aknuds1, line 78: Why the wrapping struct, which doesn't bring any benefit that I can tell, instead of:

```suggestion
	Errors []PluginError
```

aknuds1 (changes requested):

pkg/plugins/models.go
wbrowne: Yeah no benefit yet. We're discussing more in detail about what info will be required so will keep this as is for now, but if nothing changes in requirements I'll simplify

aknuds1 (approved):
Looks good AFAICT.

pkg/plugins/models.go
aknuds1: OK, thanks for the clarification.

pkg/plugins/plugins.go
```diff
@@ -229,10 +229,10 @@ func (pm *PluginManager) scan(pluginDir string, requireSigned bool) error {
...
 
 		pm.log.Debug("Found plugin", "id", plugin.Id, "signature", plugin.Signature, "hasRoot", plugin.Root != nil)
-		if !scanner.validateSignature(plugin) {
-			pm.log.Debug("Not adding plugin since it lacks a valid signature", "id", plugin.Id,
-				"signature", plugin.Signature)
-			continue
+		signingError := scanner.validateSignature(plugin)
+		if signingError != nil {
+			pm.log.Warn("Plugin has signature errors", "id", plugin.Id,
+				"signature", plugin.Signature, "status", signingError.ErrorCode.String())
```
marefr, line 235: Changing this will obviously result in that unsigned or invalid signature for plugins will allow frontend plugin static content to be loaded in the UI and backend plugins to start running in the backend. Returning an error here before was an easy way out to stop all of this from happening.

With current suggestion of changes I don't really see how/where we need to put code to make sure a loaded plugin cannot be used. In general I'm worried that we spread the logic of whether a plugin is able/allowed to run/be used throughout the code base. Feels like we're missing some kind of abstraction maybe?

Instead of calling `loader.Load` below if valid signature is required and there are errors can we load a "shadow plugin" instead that's just basically plugin meta data and nothing else?

Or alternatively, would it help/be easier to add a new API endpoint where you can query for failed plugins and only get back a list of basic plugin meta data such as id/name, plugin state and the reason it failed to start? If so a different map of failed plugins can be kept in the backend by the plugins package? A bit hard to know if this is possible without knowing what the idea with the UI is.

pkg/plugins/error.go
```diff
@@ -0,0 +1,32 @@
...
+		panic(fmt.Sprintf("Unrecognized error code %d", e))
+	}
+}
+
+type PluginError struct {
+	ErrorCode
+}
+
+type PluginErrors struct {
+	PluginErrors []PluginError
```
marefr, line 31: Do we have need of handling multiple errors? Right now feels like it can only be one. Maybe can simplify code to only handle one error?

marefr (commented):
In general some of the changes looks good, but please see comments. Maybe I've misunderstood something or missed some changes? 

Another interesting thing is what would happen or what we would expect in the following scenario. I don't have a good answer besides the user experience will be bad I think.
- You have a dashboard B using data source plugin A
- You have alert rules C using data source plugin A
- You fail to load data source plugin A when Grafana starts
- A user visiting dashboard B
- Alert rules C evaluates for data source plugin A

pkg/plugins/models.go
```diff
@@ -82,7 +82,7 @@ func (pb *PluginBase) registerPlugin(base *PluginBase) error {
 	}
 
 	if !strings.HasPrefix(base.PluginDir, setting.StaticRootPath) {
-		plog.Info("Registering plugin", "name", pb.Name)
+		plog.Info("Registering plugin", "name", pb.Id)
```
aknuds1, line 85: Labeling the ID as "name" looks wrong:
```suggestion
		plog.Info("Registering plugin", "id", pb.Id)
```

pkg/plugins/plugins.go
aknuds1: I share @marefr's concerns. Are they still valid?

pkg/plugins/plugins.go
```diff
@@ -269,12 +270,21 @@ func (pm *PluginManager) scan(pluginDir string, requireSigned bool) error {
 
 		jsonParser := json.NewDecoder(reader)
 
-		// Load the full plugin, and add it to manager
-		if err := loader.Load(jsonParser, plugin, scanner.backendPluginManager); err != nil {
-			return err
+		if signingError != nil {
+			phantomLoader := &PhantomPlugin{*plugin}
```
aknuds1, line 274: Does this resolve @marefr's concern above, wrt. loading unsigned plugins?

aknuds1 (approved):
Looks good AFAICT, but found a misnamed label in a log statement, please commit my suggestion there. Also left a couple of questions.

pkg/plugins/plugins.go
papagian: I'm not at all familiar with this code but given this:
https://github.com/grafana/grafana/blob/b705329d6a761b6c70bd730d48b3446ea4129e3d/pkg/plugins/plugins.go#L234-L238

is it ever possible this condition to be true?

wbrowne: Good eye 😉  You are correct! I just didn't want to remove the phantom loader until I was sure it can be deleted. And now it can thankfully!

wbrowne:
@aknuds1 @marefr instead of changing the behaviour of what/how we scan plugins in this change, now this is just tracking those backend plugin errors and exposing them via a new API. For full context I recommend checking out the frontend PR mentioned in the description block above.

This is scheduled for 7.3 so while there could likely be improvements, it's good to bear that in mind when reviewing 😉

pkg/api/plugins.go
```diff
@@ -364,6 +364,17 @@ func (hs *HTTPServer) getCachedPluginSettings(pluginID string, user *models.Sign
 	return query.Result, nil
 }
 
+func (hs *HTTPServer) GetPluginErrorsList(c *models.ReqContext) Response {
+	pluginErrors := make([]plugins.PluginErrorInfo, 0)
```
marefr, line 368: Would move this to a function in plugins package to get the list of errors ready to use

pkg/plugins/plugins.go
```diff
@@ -29,6 +29,7 @@ var (
 	StaticRoutes []*PluginStaticRoute
 	Apps         map[string]*AppPlugin
 	Plugins      map[string]*PluginBase
+	Errors       map[string]*PluginError
```
marefr, line 32: Could maybe make this private to the package if you add a public function to retrieve list of errors.

pkg/plugins/plugins.go
```diff
@@ -228,9 +230,11 @@ func (pm *PluginManager) scan(pluginDir string, requireSigned bool) error {
 		}
 
 		pm.log.Debug("Found plugin", "id", plugin.Id, "signature", plugin.Signature, "hasRoot", plugin.Root != nil)
-		if !scanner.validateSignature(plugin) {
-			pm.log.Debug("Not adding plugin since it lacks a valid signature", "id", plugin.Id,
-				"signature", plugin.Signature)
+		signingError := scanner.validateSignature(plugin)
+		if signingError != nil {
+			pm.log.Warn("Plugin has signature errors", "id", plugin.Id,
```
marefr, line 235: Guess the errors are already logged with error level somewhere so I guess that was the reason it used debug level here before. Maybe @aknuds1 has input on this. 

Would maybe change message to "failed to validate plugin signature"

pkg/api/frontendsettings.go
```diff
@@ -173,6 +173,7 @@ func (hs *HTTPServer) getFrontendSettingsMap(c *models.ReqContext) (map[string]i
 			"sort":          getPanelSort(panel.Id),
 			"skipDataQuery": panel.SkipDataQuery,
 			"state":         panel.State,
```
marefr, line 175: Needed for the frontend changes?

marefr (dismissed):
Great work. Looks great, but requested some minor changes.

pkg/plugins/plugins.go
wbrowne: Yeah actually was just an oversight - will return to Debug and update the message 👍

pkg/api/frontendsettings.go
wbrowne: Exactly. In order to render labels for panels in the visualizations view

marefr: Understood

pkg/plugins/models.go
```diff
@@ -197,3 +197,8 @@ func NewEnabledPlugins() EnabledPlugins {
 		Apps:        make([]*AppPlugin, 0),
 	}
 }
+
+type PluginErrorInfo struct {
```
aknuds1, line 201: I would recommend fusing this type with `PluginError`, since there's no point in having both.

aknuds1 (approved):
Looks good AFAICT, but I don't see any reason for having both `PluginError` and `PluginErrorInfo`. They could be narrowed down to one type without any drawbacks unless I'm missing the obvious.

pkg/plugins/models.go
wbrowne: Wasn't sure if there was a pattern to distinguish between internal model and DTO, but in this case they're quite simple and not the intended final result - so thanks for the suggestion 👍

aknuds1: @wbrowne TBH I don't know either, just going off intuition :D

marefr (approved):
Awesome LGTM

Took it for a quick spin and when enabled mode production I got the expected returned from /api/plugins/errors,

wbrowne:
woohoo!

marefr:
Maybe consider "add to changelog" label?

PR #28220, opened 2020-10-13
FieldColor: Per panel settings to filter out supported modes

torkelo:
Not super happy with this. 

feels a bit messy to have standardOptions (array of enums), and standardOptionDefaults (map of enum -> default obj), and standardOptionSettings: (map of enum to setting obj). 

But with I can set field color settings and have access to them in the color editor UI so I can filter the dropdown. 

The problem next is how to change the current value of config.color.mode when you switch visualization. Could do something custom for just this property.


mckn (approved):
I'm not really following what you refer to in the description but when testing and going through the code I think it looks really good!

mckn:
> I'm not really following what you refer to in the description but when testing and going through the code I think it looks really good!

Btw: do a sync with master to make this PR build.

PR #28221, opened 2020-10-13
Azure: Fixes merging of requests with different Azure sources

### LINKED ISSUE #28114: Panel with queries from different sources not working

hugohaggmark:
When I fixed #28114 with #28180 I didn't realize all the details of this data source. This PR tries to merge data from different Azure sources gracefully but as I don't know all the scenarios here I would appreciate help testing this. 

Relates #28114

### DIFF
--- a/public/app/plugins/datasource/grafana-azure-monitor-datasource/datasource.ts
+++ b/public/app/plugins/datasource/grafana-azure-monitor-datasource/datasource.ts
@@ -7,7 +7,6 @@ import {
   DataFrame,
   DataQueryRequest,
   DataQueryResponse,
-  DataQueryResponseData,
   DataSourceApi,
   DataSourceInstanceSettings,
   LoadingState,
@@ -50,7 +49,7 @@ export default class Datasource extends DataSourceApi<AzureMonitorQuery, AzureDa
     this.optionsKey = optionsKey;
   }
 
-  query(options: DataQueryRequest<AzureMonitorQuery>): Observable<DataQueryResponseData> {
+  query(options: DataQueryRequest<AzureMonitorQuery>): Observable<DataQueryResponse> {
     const byType: Record<AzureQueryType, DataQueryRequest<AzureMonitorQuery>> = ({} as unknown) as Record<
       AzureQueryType,
       DataQueryRequest<AzureMonitorQuery>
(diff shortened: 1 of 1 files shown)


aknuds1:
@kylebrandt Are you the best person to help Hugo here?

public/app/plugins/datasource/grafana-azure-monitor-datasource/datasource.ts
```diff
@@ -98,16 +101,22 @@ export default class Datasource extends DataSourceApi<AzureMonitorQuery, AzureDa
...
+      return forkJoin(obs).pipe(
+        map((results: DataQueryResponse[]) => {
+          const data: DataFrame[] = [];
+          for (const result of results) {
+            for (const frame of result.data) {
+              data.push(frame);
+            }
+          }
+
+          return { state: LoadingState.Done, data };
```
hugohaggmark, line 115: The typings for this function `Observable<DataQueryResponseData>` seem incorrect when looking at what we are returning. But I just followed the return type that already existed in the code.

hugohaggmark:
> @kylebrandt Are you the best person to help Hugo here?

Kyle found some issues with my previous fix so I add him too :)

kylebrandt (approved):
Works on my machine

image

public/app/plugins/datasource/grafana-azure-monitor-datasource/datasource.ts
ryantxu: weird -- it should be `Observable<DataDataQueryResponse>` I think, if that compiles and runs, we should change it

ryantxu (approved):
looks good, but I think we should fix the type if it does not break anythign

public/app/plugins/datasource/grafana-azure-monitor-datasource/datasource.ts
hugohaggmark: Yes, I fixed it

PR #28227, opened 2020-10-13
Build: handles verify-repos failures correctly

xlson:

Make sure `verify-repos.sh` handles failures so that it useful even when there are verification issues.

Fixes #


aknuds1 (approved):
So the idea is you want execution to continue if a step fails?

xlson:
Yes, the failurs will turn up in the lack of a version. It could be improved further, but I think I'd rather rewrite the whole thing in Go without using docker in that case.

aknuds1:
Yeah, I'm definitely for rewriting to Go :)

PR #28229, opened 2020-10-13
Explore: Add link to logs from trace span

aocenas:
Adds a link in the span that will open logs in Loki based on the process tags.
Screenshot from 2020-10-13 18-50-25

Right now this is pretty much hardcoded and no user config is available.

To see this you need to enable a feature toggle:
```
[feature_toggles]
enable = traceToLogs
```


packages/grafana-runtime/src/services/dataSourceSrv.ts
```diff
@@ -19,6 +19,16 @@ export interface DataSourceSrv {
...
+
+  /**
+   * Get all data sources
+   */
+  getAll(): DataSourceInstanceSettings[];
+
+  /**
+   * Get all data sources except for internal ones that usually should not be listed like mixed data source.
+   */
+  getExternal(): DataSourceInstanceSettings[];
```
aocenas, line 31: @dprokop @torkelo does it make sense to add it here? Otherwise I would have to use the grafana main package instead of runtime and these seems like reasonable to have here like creating a 3rd party "mixed" datasource.

torkelo: yes, makes sense :+1:

public/app/features/explore/state/actions.ts
```diff
@@ -696,6 +700,10 @@ export function splitOpen<T extends DataQuery = any>(options?: { datasourceUid:
       rightState.queryKeys = [];
       urlState.queries = [];
       rightState.urlState = urlState;
+      if (options.range) {
```
aocenas, line 703: Before we let this to be initialised automatically from left split.

packages/grafana-data/src/utils/dataLinks.ts
```diff
@@ -62,6 +63,7 @@ export function mapInternalLinkToExplore(
           onClickFn?.({
             datasourceUid: link.internal!.datasourceUid,
             query: interpolatedQuery,
+            range,
```
aocenas, line 66: This sort of affects the links from logs in loki but as this range is from the left split, which is what would have been used for the right anyway, this should not change the behaviour of log links.

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBarRow.tsx
```diff
@@ -437,6 +441,30 @@ export class UnthemedSpanBarRow extends React.PureComponent<SpanBarRowProps> {
               </span>
               <small className={styles.endpointName}>{rpc ? rpc.operationName : operationName}</small>
             </a>
+            {createSpanLink &&
```
aocenas, line 444: We add the links here if they are specified. Bit ugly with the function wrapper and all so should be refactored later on. There are already some other links in the same space so this uses the same style.

aocenas:
There is an unpleasant but with the resizer and the hover functionality that I am not sure how to fix right now and so will have to wait for separate PR.
Peek 2020-10-13 18-49

ivanahuckova (approved):
Besides the UI bug, all looks and works good. Thanks for helpful comments and adding the tests. As this is behind the feature toggle, it doesn't need any docs update.  So LGTM!

PR #28235, opened 2020-10-13
NamedColors: Named colors refactors

torkelo:
Tried to get theme colors (like panel background, etc) into the named palette, discovered a really bad performance issue with getColorFromHexRgbOrName and how it looked up values from the named palette. 

```
export const getColorDefinitionByName = (name: Color): ColorDefinition => {
  return flatten(Array.from(getNamedColorPalette().values())).filter(definition => definition.name === name)[0];
};
```

The inner function was called about 5000 times per refresh for a simple dashboard. So rewrote this part to be just an index lookup instead. 

Another problem with getColorFromHexRgbOrName was that the themeType was optional so a lot of places did not pass the theme in resulting in the wrong colors for the theme being used. So the new function getColorForTheme (please come with a better name), changes the signature from theme type to full theme object so we can access colors from the theme when we build the colors map. Also it's now required.

### DIFF
--- a/packages/grafana-data/src/field/fieldColor.ts
+++ b/packages/grafana-data/src/field/fieldColor.ts
@@ -1,5 +1,5 @@
 import { FALLBACK_COLOR, Field, FieldColorModeId, GrafanaTheme, Threshold } from '../types';
-import { classicColors, getColorFromHexRgbOrName, RegistryItem } from '../utils';
+import { classicColors, getColorForTheme, RegistryItem } from '../utils';
 import { Registry } from '../utils/Registry';
 import { interpolateRgbBasis } from 'd3-interpolate';
 import { fallBackTreshold } from './thresholds';
@@ -28,7 +28,7 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
       getCalculator: (_field, theme) => {
         return (_value, _percent, threshold) => {
           const thresholdSafe = threshold ?? fallBackTreshold;
-          return getColorFromHexRgbOrName(thresholdSafe.color, theme.type);
+          return getColorForTheme(thresholdSafe.color, theme);
         };
       },
     },
@@ -54,7 +54,7 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
     // }),
     new FieldColorSchemeMode({
       id: FieldColorModeId.PaletteClassic,
-      name: 'By series: Classic palette',
+      name: 'By series / Classic palette',
       //description: 'Assigns color based on series or field index',
       isContinuous: false,
       isByValue: false,
@@ -62,7 +62,7 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
     }),
     new FieldColorSchemeMode({
       id: 'continuous-GrYlRd',
-      name: 'By value: Green Yellow Red (gradient)',
+      name: 'By value / Green Yellow Red (gradient)',
       //description: 'Interpolated colors based value, min and max',
       isContinuous: true,
       isByValue: true,
@@ -104,7 +104,7 @@ export class FieldColorSchemeMode implements FieldColorMode {
       return this.colorCache;
     }
 
-    this.colorCache = this.colors.map(c => getColorFromHexRgbOrName(c, theme.type));
+    this.colorCache = this.colors.map(c => getColorForTheme(c, theme));
     return this.colorCache;
   }
 
@@ -149,6 +149,6 @@ export function getFieldColorMode(mode?: FieldColorModeId): FieldColorMode {
 
 function getFixedColor(field: Field, theme: GrafanaTheme) {
   return () => {
-    return getColorFromHexRgbOrName(field.config.color?.fixedColor ?? FALLBACK_COLOR, theme.type);
+    return getColorForTheme(field.config.color?.fixedColor ?? FALLBACK_COLOR, theme);
   };
 }
--- a/packages/grafana-data/src/field/fieldColor.ts
+++ b/packages/grafana-data/src/field/fieldColor.ts
@@ -61,7 +61,7 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
       colors: classicColors,
     }),
     new FieldColorSchemeMode({
-      id: FieldColorModeId.ContinousGrYlRd,
+      id: 'continuous-GrYlRd',
       name: 'By value: Green Yellow Red (gradient)',
       //description: 'Interpolated colors based value, min and max',
       isContinuous: true,
@@ -72,7 +72,7 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
 });
 
 interface FieldColorSchemeModeOptions {
-  id: FieldColorModeId;
+  id: string;
   name: string;
   description?: string;
   colors: string[];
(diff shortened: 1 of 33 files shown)


packages/grafana-data/src/utils/namedColorsPalette.ts
```diff
@@ -65,61 +67,59 @@ const buildColorDefinition = (
   isPrimary: !!isPrimary,
 });
 
-export const getColorDefinitionByName = (name: Color): ColorDefinition => {
+export function getColorDefinitionByName(name: Color): ColorDefinition {
```
ryantxu, line 70: :heart: I am happy every time I see these constants replaced!

ryantxu (approved):
touches everything!   looks good.

packages/grafana-data/src/utils/namedColorsPalette.ts
hugohaggmark: Interesting, why `function` is preferable over `const` in this case (there are some arguments I've heard when creating funcs inside a Class)? I always try to follow the style of the existing functions but for green field files I always use `const`. I don't have strong opinions about this but maybe we should have a code style guideline for this so we keep the code consistent?

packages/grafana-data/src/field/fieldColor.ts
```diff
@@ -32,55 +32,47 @@ export const fieldColorModeRegistry = new Registry<FieldColorMode>(() => {
...
-        'purple',
-        'orange',
-        'dark-blue',
-        'dark-red',
-        'dark-yellow',
-        'dark-purple',
-        'dark-orange',
-      ],
-    }),
+    // new FieldColorSchemeMode({
```
hugohaggmark, line 35: nit: maybe remove this?

packages/grafana-data/src/utils/namedColorsPalette.ts
```diff
@@ -49,6 +49,8 @@ export type ColorDefinition = {
 };
 
 let colorsPaletteInstance: Map<Hue, ColorDefinition[]>;
+let colorsMap: Record<Color, string> | undefined;
+let colorsMapTheme: GrafanaTheme | undefined;
```
hugohaggmark, line 53: nit: to me these global lets indicates we should have a wrapping function or class instead.

packages/grafana-ui/src/components/BarGauge/BarGauge.tsx
```diff
@@ -567,11 +567,12 @@ export function getBarGradient(props: Props, maxSize: number): string {
...
  */
 export function getValueColor(props: Props): string {
-  const { theme, value } = props;
+  const { value } = props;
   if (value.color) {
     return value.color;
   }
-  return getColorFromHexRgbOrName('gray', theme.type);
+
+  return 'gray';
```
hugohaggmark, line 575: nit: could these "magic" color strings be part of a enum or const?

hugohaggmark (approved):
Wow, very nice work on this and great perf catch! Just left some nit for consideration.

Tried it out at some places and found this but I don't know if this is an existing issue:
When choosing SingleColor for a BarGauge the text changes color but not the fill for the Gauge. Not sure this is an issue?
image

packages/grafana-data/src/utils/namedColorsPalette.ts
ryantxu: It may just be personal taste :)  arrow function in classes/objects will bind `this` for you, so that is great.  

I find it less clear to define function as a const... when it is really intended to be a function.

torkelo: I find functions much better, they get hoisted to the top so order is not important. It's more readable and clear that you are looking at function and not a variable. 

It used to be that normal functions had better stack traces as the function had a name instead of being an anonymous function but not sure if this is the case anymore.

packages/grafana-data/src/field/fieldColor.ts
torkelo: yea, just want to add it back soon, just the order and the colors need to be figured out :)

packages/grafana-ui/src/components/uPlot/geometries/configGetters.ts
```diff
@@ -1,10 +1,10 @@
 import { AreaProps, LineProps, PointProps } from './types';
 import tinycolor from 'tinycolor2';
-import { getColorFromHexRgbOrName } from '@grafana/data';
 
 export const getAreaConfig = (props: AreaProps) => {
+  // TODO can we pass therem here? or make sure color is already correct?
   const fill = props.fill
-    ? tinycolor(getColorFromHexRgbOrName(props.color))
```
torkelo, line 7: @dprokop found this part in the new graph component, but do not have access to the theme here so can't get the correct color

packages/grafana-data/src/utils/namedColorsPalette.ts
hugohaggmark: Here is a short example of what happens after TypeScript transpiler: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZiEMC8MAUBKFA+GBvAWACgZSYAnAUygFdywYBGAbmIF9Wji4axgoAluBgAjAIblM+YmQrU6DFu2LFgAGzEQIMAGKIAQhOkkyCJKinJchLidIcVd0RKk3ZsmWTbsgA

torkelo: yes, that highlights why I never like arrow functions in classes as well :) 
unless you have to for a callback to bind this

packages/grafana-ui/src/components/uPlot/geometries/configGetters.ts
dprokop: yes, great catch!

packages/grafana-ui/src/components/BarGauge/BarGauge.tsx
torkelo: fixed :)

packages/grafana-data/src/utils/namedColorsPalette.ts
dprokop: Interesting, why? These are module-scoped, not sure I understand the benefit of introducing another abstraction over those.

@torkelo one thing I would consider ad some point is to start namespacing the utilities. Currently we expose all of these as top level exports. Think it would possibliy be easier for the API consumers to import namespaces first and then explore the APIs. But this is fairly big change I think, something that we can think about for major version

packages/grafana-data/src/utils/namedColorsPalette.ts
dprokop: I prefer `function` over `const` and the primary reason is that with `const` syntax is hard(hacks needed) to create generic functions in `.tsx` files.

dprokop (approved):
Plenty of goodies, awesome work Torkel!

dprokop:
@torkelo I think we need to add deprecation notice to grafana-ui changelog when we release this

packages/grafana-data/src/utils/namedColorsPalette.ts
hugohaggmark: Hacks needed?

aocenas: > with const syntax is hard(hacks needed) to create generic functions
@dprokop what do you mean by that?

I like arrow functions for higher order functions ie `() => () =>` but is a weak preference because as Torkel says the lack of hoisting can be problematic and the missing function names in stack traces.

PR #28236, opened 2020-10-13
Instrumentation: Add counters and histograms for database queries

bergquist:
# DO NOT USE THIS FEATURE FLAG IT CAN CORRUPT YOUR DATABASE

----

This PR adds a fake database driver that wraps the real one and adds pre/post hooks for database queries. Using this we can get metrics and logs for all database requests issued by Grafana. It's quite many, to be honest.

I've tested it with sqlite, MySQL and Postgres and all seems to work as expected. 

This wrapper doesn't handle errors due to not being able to connect to the database. So we might have to find something for that as well. 

Adding this behind a feature flag cause I would like to get it released in a patch release quite soon. These metrics are very important for our Hosted Grafana service. 

Use this feature flag to enable the feature
```ini
[feature_toggles]
enable = database_metrics
```

Set this log filter to enable debug logs for this feature
```ini
[log]
filters = DBMetrics:debug
```


pkg/services/sqlstore/db_hook.go
```diff
@@ -0,0 +1,69 @@
...
+		Name:      "database_queries_total",
+		Help:      "The total amount of Database queries",
+	}, []string{"status"})
+
+	databaseQueryDuration = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Namespace:  "grafana",
+		Name:       "database_queries_duration",
+		Help:       "Database query duration",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+	}, []string{"status"})
```
torkelo, line 30: so nice! :heart:

torkelo (commented):
To get some good DB metrics makes me really happy! Looks good

pkg/services/sqlstore/db_hook.go
bergquist: Yea I think this will be good once it's been refactored a bit :)

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	driverWithHooks := dbType + "WithHooks"
+	sql.Register(driverWithHooks, sqlhooks.Wrap(d, &databaseQueryWrapper{}))
+	core.RegisterDriver(driverWithHooks, &databaseQueryWrapperParser{dbType: dbType})
+	return driverWithHooks
+}
+
+// databaseQueryWrapper satisfies the sqlhook.databaseQueryWrapper interface
+// which allow use to wrapp all SQL queries with a `Before` & `After` hook.
```
papagian, line 68:
```suggestion
// which allow use to wrap all SQL queries with a `Before` & `After` hook.
```

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+		Buckets:   prometheus.ExponentialBuckets(0.0001, 4, 9),
+	}, []string{"status"})
+
+	prometheus.MustRegister(databaseQueryCounter, databaseQueryHistogram)
+
+	logger = log.New("DBMetrics")
+}
+
+// WrapDatabaseDriverWithHooks creates a fake database driver that
+// executes Pre and post functions which we use to gather metrics about
```
papagian, line 45:
```suggestion
// executes pre and post functions which we use to gather metrics about
```

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// databaseQueryWrapper satisfies the sqlhook.databaseQueryWrapper interface
+// which allow use to wrapp all SQL queries with a `Before` & `After` hook.
+type databaseQueryWrapper struct{}
+
+// databaseQueryWrapperKey is used as key to save values in `context.Context`
+type databaseQueryWrapperKey struct{}
+
+// Before hook will print the query with it's args and return the context with the timestamp
```
papagian, line 74:
```suggestion
// Before hook will print the query with its args and return the context with the timestamp
```

papagian (commented):
I love it! I found some small misspellings.

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+	databaseQueryHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "grafana",
+		Name:      "database_queries_duration_seconds",
+		Help:      "Database query histogram",
+		Buckets:   prometheus.ExponentialBuckets(0.0001, 4, 9),
+	}, []string{"status"})
+
+	prometheus.MustRegister(databaseQueryCounter, databaseQueryHistogram)
+
+	logger = log.New("DBMetrics")
```
aknuds1, line 41: Can we move the initialization out of `init`, and to where `logger` is declared? I think we should only use `init` where it's technically necessary.

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/lib/pq"
+	"github.com/mattn/go-sqlite3"
+	"github.com/prometheus/client_golang/prometheus"
+	"xorm.io/core"
+)
+
+var (
+	databaseQueryCounter   *prometheus.CounterVec
+	databaseQueryHistogram *prometheus.HistogramVec
+	logger                 log.Logger
```
aknuds1, line 22: Personally, I think it would be better (wrt. readability and complexity) to initialize `logger` right off the bat.
```suggestion
	logger                 = log.New("DBMetrics")
```

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+
+	logger = log.New("DBMetrics")
+}
+
+// WrapDatabaseDriverWithHooks creates a fake database driver that
+// executes Pre and post functions which we use to gather metrics about
+// database queries.
+func WrapDatabaseDriverWithHooks(dbType string) string {
+	var d driver.Driver
+
```
aknuds1, line 49: I think we can get rid of this variable declaration. It isn't used until its definition on line 56 in any case.
```suggestion
```

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+// After hook will get the timestamp registered on the Before hook and print the elapsed time
+func (h *databaseQueryWrapper) After(ctx context.Context, query string, args ...interface{}) (context.Context, error) {
+	begin := ctx.Value(databaseQueryWrapperKey{}).(time.Time)
+	databaseQueryCounter.WithLabelValues("success").Inc()
+	databaseQueryHistogram.WithLabelValues("success").Observe(time.Since(begin).Seconds())
+	logger.Debug("query finished", "status", "success", "elapsed time", time.Since(begin), "sql", query)
+	return ctx, nil
+}
+
+// OnError instances will be called if any error happens
```
aknuds1, line 88: "instances" doesn't make any sense to me in this context.
```suggestion
// OnError will be called if any error happens
```

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+	status := "error"
+	// https://golang.org/pkg/database/sql/driver/#ErrSkip
+	if err == nil || err == driver.ErrSkip {
+		status = "success"
+	}
+
+	begin := ctx.Value(databaseQueryWrapperKey{}).(time.Time)
+	databaseQueryCounter.WithLabelValues(status).Inc()
+	databaseQueryHistogram.WithLabelValues(status).Observe(time.Since(begin).Seconds())
+	logger.Debug("query finished", "status", status, "elapsed time", time.Since(begin), "sql", query, "error", err)
```
aknuds1, line 99: I think it's best to call `time.Since(begin)` once and assign to a variable here:
```suggestion
	elapsed := time.Since(begin)
	databaseQueryHistogram.WithLabelValues(status).Observe(elapsed.Seconds())
	logger.Debug("query finished", "status", status, "elapsed time", elapsed, "sql", query, "error", err)
```

aknuds1 (approved):
Generally looks good to me, but please see comments/suggestions.

aknuds1:
Is this really supposed to go into 7.1.6? Not 7.3?

pkg/services/sqlstore/database_wrapper.go
bergquist: Agree

pkg/services/sqlstore/database_wrapper.go
bergquist: :+1:

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,109 @@
...
+func WrapDatabaseDriverWithHooks(dbType string) string {
+	drivers := map[string]driver.Driver{
+		migrator.SQLITE:   &sqlite3.SQLiteDriver{},
+		migrator.MYSQL:    &mysql.MySQLDriver{},
+		migrator.POSTGRES: &pq.Driver{},
+	}
+
+	d, exist := drivers[dbType]
+	if !exist {
+		return dbType
```
marefr, line 54: Shouldn't this be treated as an error?

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,109 @@
...
+// databaseQueryWrapper satisfies the sqlhook.databaseQueryWrapper interface
+// which allow us to wrap all SQL queries with a `Before` & `After` hook.
+type databaseQueryWrapper struct{}
+
+// databaseQueryWrapperKey is used as key to save values in `context.Context`
+type databaseQueryWrapperKey struct{}
+
+// Before hook will print the query with its args and return the context with the timestamp
+func (h *databaseQueryWrapper) Before(ctx context.Context, query string, args ...interface{}) (context.Context, error) {
+	return context.WithValue(ctx, databaseQueryWrapperKey{}, time.Now()), nil
```
marefr, line 72: What happens if you do multiple queries using the "same" context? Or will queries always execute synchronously meaning that After always will be called directly after Before?

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,109 @@
...
+	databaseQueryHistogram.WithLabelValues("success").Observe(elapsed.Seconds())
+	logger.Debug("query finished", "status", "success", "elapsed time", elapsed, "sql", query)
+	return ctx, nil
+}
+
+// OnError will be called if any error happens
+func (h *databaseQueryWrapper) OnError(ctx context.Context, err error, query string, args ...interface{}) error {
+	status := "error"
+	// https://golang.org/pkg/database/sql/driver/#ErrSkip
+	if err == nil || err == driver.ErrSkip {
```
marefr, line 89: Do we want to do anything special regarding these kind of errors?
https://github.com/grafana/grafana/blob/95a199344322918874283ce361e8b2af7d450760/pkg/services/sqlstore/transactions.go#L45

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/lib/pq"
+	"github.com/mattn/go-sqlite3"
+	"github.com/prometheus/client_golang/prometheus"
+	"xorm.io/core"
+)
+
+var (
+	databaseQueryCounter   *prometheus.CounterVec
+	databaseQueryHistogram *prometheus.HistogramVec
+	logger                 log.Logger = log.New("DBMetrics")
```
marefr, line 22: I think we in general use lowercase logger names. We already have a global `sqlstore` logger in this package. Maybe rename this `sqlstore.metrics`? And maybe move into field of `databaseQueryWrapper`?

marefr (changes requested):
Great. In general looks good. Left some comments.

What's the idea with the feature toggle, remove that in the future? If not, maybe should consider moving that into database settings similar to existing one we already have for logging queries?
https://grafana.com/docs/grafana/latest/administration/configuration/#log_queries

Given the existing log of queries feature, would make sense to me to either remove that one or either skip logging of queries in the new database wrapper. Just to make sure we don't have two features sort of doing the same thing.

pkg/services/sqlstore/database_wrapper.go
bergquist: :+1:

pkg/services/sqlstore/database_wrapper.go
bergquist: Good question. Since I want to introduce this in a patch release I prefer to be a bit defensive. 
If it doesn't exist it will just fall back to use the normal DB driver rather than the wrapped driver. Once we remove the feature flag this should definitely return an error as it's a configuration error.

pkg/services/sqlstore/database_wrapper.go
bergquist: Each query will get its own copy of the context since `WithValue` copies the context rather than adding anything to it. 
The values are added in this function and the caller overrides the old context. https://github.com/gchaincl/sqlhooks/blob/master/sqlhooks.go#L139

pkg/services/sqlstore/database_wrapper.go
bergquist: I think `driver.ErrSkip` gets ignored in `xorm`. I'm not sure :)

pkg/services/sqlstore/database_wrapper.go
marefr: Got it

pkg/services/sqlstore/database_wrapper.go
marefr: Okay :+1:

pkg/services/sqlstore/database_wrapper.go
marefr: My question was related to the retry logic we have for sqlite. I guess not

marefr (approved):
LGTM

kylebrandt:

PR #28237, opened 2020-10-14
Transformations: Add Concatenate fields transformer

ryantxu:
In combination with the [reduce transformer improvements](https://github.com/grafana/grafana/pull/27875), we can now process results more effectively.  

image


packages/grafana-data/src/transformations/transformers/concat.test.ts
```diff
@@ -0,0 +1,136 @@
+import { toDataFrame } from '../../dataframe/processDataFrame';
+import { concatenateFields, ConcatenateFrameNameMode } from './concat';
+
+export const simpleABC = toDataFrame({
+  name: 'ABC',
+  fields: [
+    { name: 'A', values: [1] },
+    { name: 'B', values: [1, 2] },
+    { name: 'C', values: [1, 2, 3] },
+  ],
```
torkelo, line 10: is this a valid configuration? (ie field with not the same lengths)

packages/grafana-data/src/transformations/transformers/concat.test.ts
```diff
@@ -0,0 +1,136 @@
...
+});
+
+describe('Concat Transformer', () => {
+  it('dropping frame name', () => {
+    const frame = concatenateFields([simpleABC, simpleXYZ], { frameNameMode: ConcatenateFrameNameMode.Drop });
+    expect(frame.length).toBe(3);
+    expect(frame.fields.map(f => ({ name: f.name, labels: f.labels }))).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "labels": undefined,
```
torkelo, line 29: it's annoying that the inline snapshot prints out undefined fields :(

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { DataTransformerID } from './ids';
+import { DataTransformerInfo } from '../../types/transformations';
+import { DataFrame, Field } from '../../types/dataFrame';
+import { ArrayVector } from '../../vector';
+
+export enum ConcatenateFrameNameMode {
+  Drop = 'drop',
+  FieldName = 'field',
+  Label = 'label',
```
torkelo, line 11: jsdoc descriptions for what these mean would be great

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -0,0 +1,88 @@
...
+  frameNameLabel?: string;
+}
+
+export const concatenateTransformer: DataTransformerInfo<ConcatenateTransformerOptions> = {
+  id: DataTransformerID.concatenate,
+  name: 'Concatenate fields',
+  description:
+    'Combine all fields into a single frame.  Values will be appended with undefined values if not the same length.',
+  defaultOptions: {
+    frameNameMode: ConcatenateFrameNameMode.FieldName,
```
torkelo, line 25: it feels like this mode controls field naming more than frame naming :)

packages/grafana-data/src/transformations/transformers/concat.test.ts
ryantxu: `toDataFrame` will sort that out (making them all the same length), but your comment points out that I am not testing what I though I was.  The two frames should be of different lenghts to make sure they end up the same after concat

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -0,0 +1,88 @@
...
+          copy.labels[frameNameLabel] = frame.name;
+        } else if (!copy.name || copy.name === 'Value') {
+          copy.name = frame.name;
+        } else {
+          copy.name = `${frame.name} · ${f.name}`;
+        }
+      }
+      fields.push(copy);
+    }
+  }
```
torkelo, line 69: nitpick: please newline after a long big if statement

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+/**
+ * @internal only exported for tests
+ */
+export function concatenateFields(data: DataFrame[], opts: ConcatenateTransformerOptions): DataFrame {
+  let sameLength = true;
+  let maxLength = data[0].length;
+  const frameNameLabel = opts.frameNameLabel ?? 'frame';
+  let fields: Field[] = [];
```
torkelo, line 46: nitpick: newline here to separate the for statement from the preceding variable inits

packages/grafana-data/src/transformations/transformers/concat.ts
```diff
@@ -0,0 +1,88 @@
...
+        return f;
+      }
+      const values = f.values.toArray();
+      values.length = maxLength;
+      return {
+        ...f,
+        values: new ArrayVector(values),
+      };
+    });
+  }
```
torkelo, line 83: nitpick: last one newline here so the return is not part of the preceding if statement

packages/grafana-data/src/transformations/transformers/concat.test.ts
ryantxu: I updated the test so that ABC is lenght 2, and XYZ is length 3 -- after concat, the total length of all fields is 3

torkelo (approved):
Looks good! some naming suggestions & nitpicks. 
Sorry for the newline comments, don't want to be annoying but I really struggle reading functions when the outer scope does not have paragraphs. I can buy no new line in nested for & if statements but the outer main function scope I really need paragraphs in order to read.

ryantxu:
> Sorry for the newline comments,

No worries -- I struggle to have any style opinions while working in a few languages at the same time.

Are there any possible lint rules we could use to enforce this?  or is it more a feeling?  Ideally we can just enforce it and remove it from the list of things we discuss :)

PR #28256, opened 2020-10-14
Gauge: Improve text sizing and support non threshold color modes

### LINKED ISSUE #19636: Gauge font rendering does not scale correctly
**What happened**:
Long text when using range text mappings on a gauge do not scale properly, resulting in the text overlapping the gauge itself. The scaling seems to hit a point where the gauge itself widens the value area, while also increasing the text size, resulting in only certain panel sizes being viable for use with long text in value mappings. The only workaround is to make the panel extremely large, or extremely small, or switch back to the singlestat (which is what I had to do in my case).

Text scaling needs to be reviewed to adapt to all sizes appropriately, perhaps by pre-processing of the text width in line with the gauge to prevent the text appearing so large it overlaps the gauge itself; or allow manual font size similar to the legacy singlestat panel gauge (which I personally prefer, as it allows for consistent text sizes).

Create a relatively small gauge panel, and create a mapping that results in long text (e.g. "Very Poor"). The text overlaps the gauge itself depending on the panel size (simply adjust the panel size in varying width or height and this is readily observed with some sizes resulting in overlap).

I have markers displayed on the graph also.
**Environment**:
- Grafana version: 6.4.1
- Data source type & version: InfluxDB 1.7.8
- OS Grafana is installed on: Raspbian Stretch (debian)
- User OS & Browser: Windows 10, Firefox 69.0.2
- Grafana plugins: Standard (no changes)
- Others: Screenshot attached

image

torkelo:
Fixes #19636

Also adds color.mode check

### DIFF
--- a/packages/grafana-ui/src/components/Gauge/Gauge.test.tsx
+++ b/packages/grafana-ui/src/components/Gauge/Gauge.test.tsx
@@ -3,7 +3,7 @@ import { shallow } from 'enzyme';
 
 import { Gauge, Props } from './Gauge';
 import { getTheme } from '../../themes';
-import { ThresholdsMode, FieldConfig } from '@grafana/data';
+import { ThresholdsMode, FieldConfig, FieldColorModeId } from '@grafana/data';
 
 jest.mock('jquery', () => ({
   plot: jest.fn(),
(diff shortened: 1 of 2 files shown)


jackw:
LGTM! 🚀

jackw (approved):

PR #28257, opened 2020-10-14
Variables: Fixes so constants set from url get completed state

### LINKED ISSUE #28215: Templating: No error when Datasource does not exists on dashboard load
I had a dashboard with variable loading data from Postgres datasource. Variable update method is set to on Dashboard load.

I renamed the datasource from `A` to `B` but variable still tries to load data from datasource `A` (that does not exists anymore).

image

**However, no error is shown.**

----------------

It is necessary to access dashboard through url with variables included

https://localhost/grafana/d/hash/dbname?var-name=a

Accessing this displays the error properly https://localhost/grafana/d/hash/dbname

image

v7.3 dev

hugohaggmark:
When adding Loading state to variables I forgot the set value from url path during the initial load of a dashboard. That meant that some variables would never receive the loading state completed.

Fixes #28215


mckn (approved):
I think this looks good! Easier to follow then the previous version and I can't reproduce the issue I had yesterday when getting stuck in the loading state described in the issue.

Awesome work @hugohaggmark!

nit: just make sure the build is green.

hugohaggmark:
> I think this looks good! Easier to follow then the previous version and I can't reproduce the issue I had yesterday when getting stuck in the loading state described in the issue.
> 
> Awesome work @hugohaggmark!
> 
> nit: just make sure the build is green.

Thank you @mckn for taking the time to try this out ⭐   and I'm going to fix the build 🙈

PR #28265, opened 2020-10-14
SQLStore: Run tests as integration tests

papagian:
Continuation of work #26858 (I don't have permissions to push to this branch) following [this](https://github.com/grafana/grafana/pull/26858#issuecomment-692661118) suggestion

Fixes #


pkg/services/sqlstore/migrator/postgres_dialect.go
```diff
@@ -140,6 +140,37 @@ func (db *Postgres) CleanDB() error {
...
+		case "dashboard_acl":
+			// keep default dashboard permissions
+			if _, err := sess.Exec(fmt.Sprintf("DELETE FROM %v WHERE dashboard_id != -1 AND org_id != -1;", db.Quote(table.Name))); err != nil {
+				return errutil.Wrapf(err, "failed to truncate table %q", table.Name)
+			}
+			if _, err := sess.Exec(fmt.Sprintf("ALTER SEQUENCE %v RESTART WITH 3;", db.Quote(fmt.Sprintf("%v_id_seq", table.Name)))); err != nil {
+				return errutil.Wrapf(err, "failed to reset table %q", table.Name)
+			}
+		default:
+			if table.Name == "" {
```
kylebrandt, line 158: I think this should be in case statement above, `case "":`, instead of nested in the default?

papagian: Good point

pkg/services/sqlstore/migrator/dialect.go
```diff
@@ -45,6 +45,7 @@ type Dialect interface {
 	PostInsertId(table string, sess *xorm.Session) error
 
 	CleanDB() error
+	TruncateDB() error
```
kylebrandt, line 48: None of the existing interface methods have comments, but maybe this would be a good time to start? In particular, what is the difference between CleanDB and TruncateDB seems like a worthy comment. 

Maybe TruncateDBTables is more specific?

kylebrandt (approved):
LGTM pending case statement fix (and maybe a comment/rename as your desecration)

sakjur (approved):
It's awesome how this brings runtime down from 20 minutes+ to 1.5 minutes 😄

pkg/services/sqlstore/migrator/sqlite_dialect.go
```diff
@@ -85,6 +86,39 @@ func (db *Sqlite3) CleanDB() error {
...
+	defer sess.Close()
+
+	for _, table := range tables {
+		switch table.Name {
+		case "dashboard_acl":
+			// keep default dashboard permissions
+			if _, err := sess.Exec(fmt.Sprintf("DELETE FROM %q WHERE dashboard_id != -1 AND org_id != -1;", table.Name)); err != nil {
+				return errutil.Wrapf(err, "failed to truncate table %q", table.Name)
+			}
+			if _, err := sess.Exec("UPDATE sqlite_sequence SET seq = 2 WHERE name = '%s';", table.Name); err != nil {
```
papagian, line 107: @kylebrandt @sakjur 
thank you for approving it!
However, I've introduced this change (together with @kylebrandt's suggestions) since the last time you checked so if you don't mind please have a look.
The change is basically for setting the `dashboard_acl` sequence to the correct value; note that no tests did complain before. I guess that's because SQLite is smart handling it but still it's good to fix it:
```
sqlite> UPDATE sqlite_sequence SET seq = 0 WHERE name = 'dashboard_acl';
sqlite> select id from dashboard_acl ;
1
2
3
4
sqlite> select * from sqlite_sequence where name = 'dashboard_acl';
dashboard_acl|0
sqlite> insert into dashboard_acl(org_id, dashboard_id, user_id, permission, created, updated) values (1, 138, 3, 4, '2020-10-15 10:24:48', '2020-10-15 10:24:48');
sqlite> select id from dashboard_acl ;
1
2
5
3
4
sqlite> select * from sqlite_sequence where name = 'dashboard_acl';
dashboard_acl|5
```

sakjur (approved):
LGTM.

Not of consequence, since the exact auto incremented value shouldn't matter to any test, but why do we restart the Postgres counter to 3 and the MySQL/SQLite counter to 2?

pkg/services/sqlstore/migrator/mysql_dialect.go
```diff
@@ -138,7 +138,9 @@ func (db *Mysql) CleanDB() error {
 	return nil
 }
 
-func (db *Mysql) TruncateDB() error {
+// TruncateDBTables truncates all the tables
```
kylebrandt, line 141: Should be periods at the end of this and the following sentence (and in the other occurrences)

papagian:
> LGTM.
> 
> Not of consequence, since the exact auto incremented value shouldn't matter to any test, but why do we restart the Postgres counter to 3 and the MySQL/SQLite counter to 2?

good point, I think that in PostgreSQL it should be set to the next value while in the others to the highest id in the table.

for [MySQL](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html):
> You cannot reset the counter to a value less than or equal to the value that is currently in use. For both InnoDB and MyISAM, if the value is less than or equal to the maximum value currently in the AUTO_INCREMENT column, the value is reset to the current maximum AUTO_INCREMENT column value plus one.

for [postgreSQL](https://www.postgresql.org/docs/13/sql-altersequence.html):
>  The optional clause RESTART [ WITH restart ] changes the current value of the sequence. This is similar to calling the setval function with is_called = false: the specified value will be returned by the next call of nextval.

for [SQLite](https://sqlite.org/fileformat2.html#seqtab):
> the largest INTEGER PRIMARY KEY ever inserted into that table is in the sqlite_sequence.seq field

kylebrandt:

PR #28271, opened 2020-10-14
Elasticsearch: Use 'source' as inline script key for versions 5.6+

### LINKED ISSUE #23766: Elasticsearch: Scripted field deprecated key names
**What happened**:

In version 5.6.0 of ElasticSearch there was a change of scripted fields keys, "inline" was changed to "source", and "stored" to "id". Grafana is still using the "inline" script when using a scripted field, which triggers a Deprecated warning in ElasticSearch. The support for old key names will eventually get removed, which will break Grafana graphs which use scripted fields.

Example:
Setting script value to:
`_value / 60
`Results in a query like:
`body:"{"search_type":"query_then_fetch","ignore_unavailable":true,"index":["flows-*"]} {"size":0,"query":{"bool":{"filter":[{"range":{"@timestamp":{"gte":1587525071624,"lte":1587535871624,"format":"epoch_millis"}}},{"query_string":{"analyze_wildcard":true,"query":"netflow.ipv4_src_addr: x.x.x.x"}}]}},"aggs":{"3":{"terms":{"field":"netflow.ipv4_src_addr","size":10,"order":{"_key":"desc"},"min_doc_count":1},"aggs":{"2":{"date_histogram":{"interval":"1m","field":"@timestamp","min_doc_count":0,"extended_bounds":{"min":1587525071624,"max":1587535871624},"format":"epoch_millis"},"aggs":{"1":{"sum":{"field":"netflow.in_bits","missing":0,"script":{"inline":"_value / 60"}}}}}}}}} "`

The graph works, but ElasticSearch logs a warning:
`[2020-04-20T12:47:42,994][WARN ][o.e.d.c.ParseField       ] [node0] Deprecated field [inline] used, expected [source] instead
`

Grafana has to use "inline" for ElasticSearch prior to 5.6.0 and "source" for ElasticSearch version 5.6.0+.

Reference: https://www.elastic.co/guide/en/elasticsearch/reference/5.6/release-notes-5.6.0.html
Related GitHub change: https://github.com/elastic/elasticsearch/pull/25127

No deprecated warnings.

- create a graph in Grafana from ElasticSearch
- select metric: sum
- click options
- put in a script, like: `_value / 1`
- open the graph
- watch elasticsearch logs

**Environment**:
- Grafana version: 6.7.1
- Data source type & version: ElasticSearch 7.1
- OS Grafana is installed on: Debian Linux 10
- User OS & Browser: OSX 10.15.3
- Grafana plugins: N/A
- Others:

fabsrc:

Sets `source` as key for inline aggregation scripts for all ES versions 5.6+. `inline` (current key, but deprecated in version 5.6) will be used for all versions <5.6.
This should fix the deprecation warnings in the ES deprecation logs.

Fixes #23766

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/query_builder.ts
+++ b/public/app/plugins/datasource/elasticsearch/query_builder.ts
@@ -318,11 +318,6 @@ export class ElasticQueryBuilder {
         metricAgg = { field: metric.field };
       }
 
-      if (metric.inlineScript) {
-        const keyName = this.esVersion < 56 ? 'inline' : 'source';
-        metricAgg.script = { [keyName]: metric.inlineScript };
-      }
-
       for (const prop in metric.settings) {
         if (metric.settings.hasOwnProperty(prop) && metric.settings[prop] !== null) {
           metricAgg[prop] = metric.settings[prop];
--- a/public/app/plugins/datasource/elasticsearch/query_builder.ts
+++ b/public/app/plugins/datasource/elasticsearch/query_builder.ts
@@ -318,6 +318,11 @@ export class ElasticQueryBuilder {
         metricAgg = { field: metric.field };
       }
 
+      if (metric.inlineScript) {
+        const keyName = this.esVersion < 56 ? 'inline' : 'source';
+        metricAgg.script = { [keyName]: metric.inlineScript };
+      }
+
       for (const prop in metric.settings) {
         if (metric.settings.hasOwnProperty(prop) && metric.settings[prop] !== null) {
           metricAgg[prop] = metric.settings[prop];
--- a/public/app/plugins/datasource/elasticsearch/metric_agg.ts
+++ b/public/app/plugins/datasource/elasticsearch/metric_agg.ts
@@ -115,8 +115,20 @@ export class ElasticMetricAggCtrl {
           break;
         }
       }
-      if ($scope.aggDef.supportsInlineScript && $scope.settingsLinkText === '') {
-        $scope.settingsLinkText = 'Options';
+      if ($scope.aggDef.supportsInlineScript) {
+        // I know this stores the inline script twice
+        // but having it like this simplifes the query_builder
+        const inlineScript = $scope.agg.inlineScript;
+        if (inlineScript) {
+          const inlineScriptKey = $scope.esVersion < 56 ? 'inline' : 'source';
+          $scope.agg.settings.script = { [inlineScriptKey]: inlineScript };
+        } else {
+          delete $scope.agg.settings.script;
+        }
+
+        if ($scope.settingsLinkText === '') {
+          $scope.settingsLinkText = 'Options';
+        }
       }
     };
 
--- a/public/app/plugins/datasource/elasticsearch/metric_agg.ts
+++ b/public/app/plugins/datasource/elasticsearch/metric_agg.ts
@@ -120,7 +120,8 @@ export class ElasticMetricAggCtrl {
         // but having it like this simplifes the query_builder
         const inlineScript = $scope.agg.inlineScript;
         if (inlineScript) {
-          $scope.agg.settings.script = { inline: inlineScript };
+          const inlineScriptKey = $scope.esVersion < 56 ? 'inline' : 'source';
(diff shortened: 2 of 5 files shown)


zoltanbedi:
Thanks for contributing @fabsrc this looks good to me. Can you confirm this @Elfo404 please?

Elfo404 (changes requested):
Hi @fabsrc! thanks for the PR!

This looks good, but do you mind adding a test for this in `public/app/plugins/datasource/elasticsearch/specs/query_builder.test.ts` to check whether the correct key is used for both `< 5.6` and `>= 5.6`?

We are in the process of refactoring this data source and this kind of changes are hard to keep track of without failing tests :)

fabsrc:
Since this logic was outside the `ElasticQueryBuilder` there was no easy way of testing it.
I moved the logic into the `ElasticQueryBuilder#build` method now, which seems like a better place for it anyway. All the other conditional logic based on ES versions is also located there.

public/app/plugins/datasource/elasticsearch/query_builder.ts
```diff
@@ -318,6 +318,11 @@ export class ElasticQueryBuilder {
         metricAgg = { field: metric.field };
       }
 
+      if (metric.inlineScript) {
```
Elfo404, line 321: Just to be functionally equivalent to the previous version, what do you think about adding this?

```suggestion
      const supportsInlineScript = !!queryDef.metricAggTypes.find(metricDef => metricDef.value === metric.type)
        ?.supportsInlineScript;

      if (supportsInlineScript && metric.inlineScript) {
```

the reason is that when changing metric type from the query editor we don't erase previously set properties from `settings`, which means you can set a script on a metric that supports it, then change the metric type to something that doesn't support it and the `inlineScript` would maintain the previous value. without this extra check the query builder would add `script: { 'inline/source': ...}` to every metric instead to only those that supports it.

Elfo404:
@fabsrc I was testing this out, and i'm not sure why, but with your changes the query generated when creating alerts does not include the script. I'll try to investigate a bit more later on, but could you please double check?

fabsrc:
@Elfo404 I checked the issue with the alerts. Seems like the server side query parser uses the model (not the query from the `query_builder.ts`) to construct the ES query. And since the inline script logic does not exist there, the query is sent to ES without the inline script.

The solutions I see here would be to implement the logic on the server side as well or just go back to the previous solution and maybe add a function to `query_def.ts` which returns the inline script settings object based on the ES version.
I would prefer the latter as it adds less complexity. E.g.

```ts
export function getInlineScriptSetting(inlineScript: string, esVersion: number) {
  const keyName = esVersion < 56 ? 'inline' : 'source';
  return { [keyName]: inlineScript };
}
```

which is then called like this in the controller:

```ts
$scope.agg.settings.script = queryDef.getInlineScriptSetting(inlineScript, $scope.esVersion);
```

Elfo404:
@fabsrc 
> The solutions I see here would be to implement the logic on the server side as well or just go back to the previous solution and maybe add a function to `query_def.ts` which returns the inline script settings object based on the ES version.
> I would prefer the latter as it adds less complexity. E.g.

I would generally agree, but this would require users to manually update all of their queries if they want the setting to be effectively applied (especially if at some point in the future the deprecated key is going to be removed) and as a general rule we would like users to actively maintain (as in fixing) their dashboards as little as possible.

So maybe in this case decoupling what we store to what is actually sent to ES would be the best way to go. What do you think?

fabsrc:
I was a bit worried about the code duplication in the frontend and the backend query builder. But if both are tested well enough, there should be no issue.

I'll try to adjust the server logic then as well.

Elfo404:
Hi @fabsrc, any plan on picking this up? 
I'm closing the PR due to inactivity, but feel free to reopen/ping me if you have any updates.

fabsrc:
> Hi @fabsrc, any plan on picking this up?
> I'm closing the PR due to inactivity, but feel free to reopen/ping me if you have any updates.

Aah, kind of forgot about this one.
I have some WIP for the backend changes, but still needs some tweaking.
Will reopen the PR once it's finished.

PR #24247, opened 2020-05-04
Backend plugins: Allow backend component for apps

marefr:
**What this PR does / why we need it**:
Allow an app plugin to have a backend component.

PR #28272, opened 2020-10-14
App Plugins: Add support for backend

joanlopez:

As far as I know there is no way to include backend code for `"type": "app"` plugins. However, it'd be nice if we add support for backend for this type of plugins for the same reasons as for `"type": "datasource"` ones. Imagine the case when you need to consume an external API with complex auth mechanism or even with a Go SDK, as an example. This way, we'll be able to expose internal app endpoints to do it on the backend side. 

I basically copy-pasted almost the same code as for `datasource` plugins, but I'm not really sure if everything there is really needed, especially the `grpcplugin.PluginStartFuncs` parameter passed into the `grpcplugin.NewBackendPlugin` function, as it seems to be there for legacy reasons. So, feedback is more than welcome.

PS: I tested it manually with an `app` plugin and it worked! 🎉 

Thanks!!!


ryantxu:
@wbrowne -- can you take a look at this in the context of the plugin manager?

@joanlopez what are you using this for?  I assume you are defining resource routes, not query responses

marefr:
We did close my PR https://github.com/grafana/grafana/pull/24247  basically doing the same thing before 7.0. There are some interesting discussions/context in that PR. Given some discussions about whether all plugin types should be able to have a backend there's also some discussions going on in regards to supporting a single backend for multiple frontend plugins. 

As Ryan mentions only resource calls makes sense for apps in the current shape of plugin types. 

Without knowing why you need this I'm just throwing this out as well for context; In addition, to better support apps with a backend we probably want to pick this up again in the sdk which basically gives developer better support for building a backend app plugin: https://github.com/grafana/grafana-plugin-sdk-go/pull/149/commits/cfe3572348359056525a05abc602097af33a582d

joanlopez:
Thanks for your feedback @ryantxu @marefr 🙏 Here my thoughts on this 👇 

<br/>

> @joanlopez what are you using this for? I assume you are defining resource routes, not query responses

> As Ryan mentions only resource calls makes sense for apps in the current shape of plugin types.

Exactly, it's specifically thought for resource routes. As I mentioned on the PR description, I'm wasn't sure about which parts of the previous code are really needed to add support for (only) resource routes.

> Without knowing why you need this I'm just throwing this out as well for context; In addition, to better support apps with a backend we probably want to pick this up again in the sdk which basically gives developer better support for building a backend app plugin: grafana/grafana-plugin-sdk-go@cfe3572

I agree that it might have sense, especially to improve the developer experience. However, it wouldn't be enough, right? I mean, currently you cannot define resource routes for `app` plugins, and these changes won't make it possible, if I'm not wrong (just guessing).

> We did close my PR #24247 basically doing the same thing before 7.0. There are some interesting discussions/context in that PR. Given some discussions about whether all plugin types should be able to have a backend there's also some discussions going on in regards to supporting a single backend for multiple frontend plugins.

Sorry, but I cannot understand clearly what's the plan. Is it currently possible to add resource routes without introducing new changes? We'd like to introduce them, but available for all the different plugin types?


<br/>
<br/>

Finally, and to give you as much context as possible, I'm current solving this issue by adding a `datasource` one within the `app` just to expose multiple resource routes. So, actually I can consume these endpoints without the need of adding a new `datasource` but it still represents an available datasource which can be confusing in terms of user experience.

marefr:
For reference, we're currently discussing this.

pkg/plugins/app_plugin.go
```diff
@@ -67,6 +75,18 @@ func (app *AppPlugin) Load(decoder *json.Decoder, base *PluginBase, backendPlugi
 		return err
 	}
 
+	if app.Backend {
+		cmd := ComposePluginStartCommand(app.Executable)
+		fullpath := path.Join(app.PluginDir, cmd)
+		factory := grpcplugin.NewBackendPlugin(app.Id, fullpath, grpcplugin.PluginStartFuncs{
+			OnLegacyStart: app.onLegacyPluginStart,
+			OnStart:       app.onPluginStart,
+		})
```
marefr, line 84:
```suggestion
		factory := grpcplugin.NewBackendPlugin(app.Id, fullpath, grpcplugin.PluginStartFuncs{})
```

pkg/plugins/app_plugin.go
```diff
@@ -111,3 +131,23 @@ func (app *AppPlugin) initApp() {
...
+
+func (app *AppPlugin) onPluginStart(pluginID string, client *grpcplugin.Client, logger log.Logger) error {
+	if client.DataPlugin != nil {
+		tsdb.RegisterTsdbQueryEndpoint(pluginID, func(dsInfo *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
+			return wrapper.NewDatasourcePluginWrapperV2(logger, app.Id, app.Type, client.DataPlugin), nil
+		})
+	}
+
+	return nil
+}
```
marefr, line 153:
```suggestion
```

marefr (dismissed):
Added some suggestions. Basically the additional start callbacks are only needed for data source backend plugins.

marefr (approved):
LGTM

pkg/plugins/app_plugin.go
```diff
@@ -67,6 +72,15 @@ func (app *AppPlugin) Load(decoder *json.Decoder, base *PluginBase, backendPlugi
 		return err
 	}
 
+	if app.Backend {
+		cmd := ComposePluginStartCommand(app.Executable)
+		fullpath := path.Join(app.PluginDir, cmd)
```
aknuds1, line 77: Need to use `filepath.Join` for filesystem paths.
```suggestion
		fullpath := filepath.Join(app.PluginDir, cmd)
```

pkg/plugins/app_plugin.go
```diff
@@ -67,6 +72,15 @@ func (app *AppPlugin) Load(decoder *json.Decoder, base *PluginBase, backendPlugi
...
+		factory := grpcplugin.NewBackendPlugin(app.Id, fullpath, grpcplugin.PluginStartFuncs{})
+		if err := backendPluginManager.Register(app.Id, factory); err != nil {
+			return errutil.Wrapf(err, "Failed to register backend plugin")
```
aknuds1, line 80: Go errors should not be capitalized:
```suggestion
			return errutil.Wrapf(err, "failed to register backend plugin")
```

aknuds1 (changes requested):

aknuds1:
Is it possible to include a test for this?

pkg/plugins/app_plugin.go
joanlopez: Does that mean that data source backend plugins aren't working on Windows? 
I just copy-pasted the code from `pkg/plugins/datasource_plugin.go`, where `path` is used.

I can simply replace the import, of course. But I'm a bit concerned now, since I copied these lines from a code that's supposed to be working properly.

What do you think? @aknuds1  Thanks!!!

aknuds1: @joanlopez I haven't tested, but the code in pkg/plugins/ should be fixed also to use `filepath.Join` unless I'm missing the obvious.

joanlopez:
> Is it possible to include a test for this?

Honestly, I cannot see a clear behavior to test. But I'm completely open to feedback & suggestions.
Note that no plugin implementation has tests on its `Load` method. The thing is that almost all the code is testing for the `PluginBase` data structure. And here, I cannot see a test that could bring us considerable value related to the few lines added.

What would you test? The only reasonable thing to test here is the call to `backendPluginManager.Register`, but I cannot clearly see how to assert it? Just testing it's being called? Call the `grpcplugin.NewBackendPlugin` function with the same arguments on the test (copy paste these lines)? Can you shed me some light, please? Thanks!!! 🙏

pkg/plugins/app_plugin.go
aknuds1: @marefr @papagian Can you confirm whether the code in pkg/plugins/datasource_plugin.go should be fixed to use [filepath.Join](https://golang.org/pkg/path/filepath/#Join) instead of path.Join? So long as it's a filesystem path, we should use filepath.Join, for it to work also on Windows, so unless there's something particular going on here, filepath.Join should be the right choice.

aknuds1:
@joanlopez I'm not familiar enough with the code to be able to suggest a test. I'm just wondering if there should be something, it would for example be great if we could catch this potential Windows bug (we are supposed to start testing on Windows also under CI).

pkg/plugins/app_plugin.go
joanlopez: Regarding the package documentation (`path` vs `filepath`), you're completely right.

However, there are multiple `path.Join` usages within the `plugins` package. 
Specifically, some of these usages are exactly for the same use case `fullpath := path.Join(r.PluginDir, cmd)`.
So, I guess we can replace all of these, but I'd like to get your confirmation.

joanlopez: Regarding how to test it, probably we can add a new method into the `PluginBase` or create a new isolated function (as  `ComposePluginStartCommand`) specifically to get the plugin full path, but it won't prevent future misusages of `path.Join`. 

Thoughts?

marefr: @aknuds1 @joanlopez I think you're correct, but we haven't had a problem with this before. Therefore I would suggest to handle this in a separate PR to get this PR merged for 7.3 stable and unblock Joan.

aknuds1: @marefr I don't think we need to fix the other instances in this PR, but I think my suggestion above should be committed to avoid introducing a new bug.

aknuds1 (approved):
LGTM

pkg/plugins/app_plugin.go
joanlopez: @aknuds1 I do agree! I replaced the `path.Join` usage on this PR.

</br>

> I'm still wondering if we should try to get a test in here.

Also interested in your thoughts 🤔 😇

aknuds1: @joanlopez It depends on the deadline, which milestone is it targeting?

marefr:
Data source plugin loading and others lacks tests as well. Therefore would suggest handling this in a separate PR, but if you want to add a test here you're more than welcome to do it. We can create a follow up issue with changes identified maybe?

PR #28276, opened 2020-10-15
BackendSrv: Reloads page instead of redirect on Unauthorized Error

### LINKED ISSUE #27874: Loss of login session (cookie expiry) also causes loss of URL
**Current (undesired) behavior**
I keep a number of Grafana tabs open at all times. Grafana regularly loses my session when a cookie expires, which isn't unusual. Unfortunately this means that when I log back in, I'm directed to the Home screen rather than the URLs I had been at previously and have to spend some time reconfiguring my tabs to have the right dashboards.

**Desired behavior**
"Re-authenticate" behavior maintains the originally requested URL and sends the user there once logged in. This actually works as expected if I have a URL for a dashboard and I'm not logged in—navigating to that URL with a "hard" page request from the browser gets intercepted with the "login" page, but then proceeds to the correct page. However, an expired cookie and AJAX refresh (not full-page refresh) sends me to a login page that does not then proceed to the original page.

**Steps to reproduce the problem**
1. Go to a dashboard 
2. Delete Grafana cookie 
3. Refresh dashboard with Grafana dashboard "refresh" button (not browser refresh) 
4. Log in when prompted

Version: grafana v7.1.5

hugohaggmark:
Are there any edge cases here that I missed?

Fixes #27874

### DIFF
--- a/public/app/core/services/backend_srv.ts
+++ b/public/app/core/services/backend_srv.ts
@@ -2,11 +2,11 @@ import { from, merge, MonoTypeOperatorFunction, Observable, Subject, Subscriptio
 import { catchError, filter, map, mergeMap, retryWhen, share, takeUntil, tap, throwIfEmpty } from 'rxjs/operators';
 import { fromFetch } from 'rxjs/fetch';
 import { v4 as uuidv4 } from 'uuid';
-import { BackendSrv as BackendService, BackendSrvRequest, FetchResponse, FetchError } from '@grafana/runtime';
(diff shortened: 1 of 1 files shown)


torkelo (approved):
Think this should be good. Did you manage test and verify that this worked?

hugohaggmark:
> Think this should be good. Did you manage test and verify that this worked?

Yes it worked, I tested with 
- deleting the cookie 
- logging out in a separate tab.

hugohaggmark:
Should we move this into beta1?

hugohaggmark:
I changed the milestone to beta1, because we might catch any edge cases before stable.

PR #28277, opened 2020-10-15
Docs: improved github action that syncs docs to website

mckn:
In this PR I have done two things:

1. I added a cache step that only will re-install the node_modules when yarn.lock file have been changed. You can read more about the approach here: https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh

2. Looks like the current version of the sync job will fail if there is no changes. Since we generate the docs we can end up in a situation that files in the grafana packages have changed (which will trigger this action) but the generated docs have not.

**Note:** it has a dependency on this PR: https://github.com/grafana/website-sync/pull/2


hugohaggmark:
> **What this PR does / why we need it**:
> In this PR I have done two things:
> 
> 1. I added a cache step that only will re-install the node_modules when yarn.lock file have been changed. You can read more about the approach here: https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh
> 2. Looks like the current version of the sync job will fail if there is no changes. Since we generate the docs we can end up in a situation that files in the grafana packages have changed (which will trigger this action) but the generated docs have not.

The first point feels like a piece of logic I hope we use in all our "other" builds. I'm not sure I can contribute in reviewing this but have you tested this and it works?

hugohaggmark (approved):
LGTM!

aknuds1 (approved):
Looks good AFAICT.

robbymilo (approved):

PR #28283, opened 2020-10-15
Drone: Use DRONE_COMMIT_REF to get release tag

aknuds1:
Fix Drone release pipeline, by making it use $DRONE_COMMIT_REF to get release tag, as $DRONE_TAG shouldn't be set (because we don't use tag event trigger).


marefr (approved):
LGTM

papagian (approved):

PR #28284, opened 2020-10-15
Jaeger: timeline collapser to show icons

### LINKED ISSUE #25326: Tracing UI collapse/uncollapse span button.

When using the jaeger UI I often do this:

- Collapse all spans.
- Uncollapse one depth at the time, until I'm sastisfied.

This is done by using those button:

image

I would like those button in Grafana too ! 

**Why is this needed**:

Unfortunately I have to use jaeger for big traces as it's too tedious to find what I want.

/cc @aocenas

zoltanbedi:
Fixes #25326

### DIFF
--- a/packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
+++ b/packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
@@ -12,18 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
-import { css } from 'emotion';
+import { Tooltip } from '@grafana/ui';
 import cx from 'classnames';
-
-import { UITooltip, UIIcon } from '../../uiElementsContext';
+import { css } from 'emotion';
+import React from 'react';
+import DoubleRightIcon from 'react-icons/lib/fa/angle-double-right';
+import RightIcon from 'react-icons/lib/fa/angle-right';
 import { createStyle } from '../../Theme';
 
 const getStyles = createStyle(() => {
   return {
-    TraceTimelineViewer: css`
-      border-bottom: 1px solid #bbb;
-    `,
     TimelineCollapser: css`
       align-items: center;
       display: flex;
@@ -31,13 +29,11 @@ const getStyles = createStyle(() => {
       justify-content: center;
       margin-right: 0.5rem;
     `,
-    tooltipTitle: css`
-      white-space: pre;
-    `,
     btn: css`
       color: rgba(0, 0, 0, 0.5);
       cursor: pointer;
       margin-right: 0.3rem;
+      font-size: 1.5rem;
       padding: 0.1rem;
       &:hover {
         color: rgba(0, 0, 0, 0.85);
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -56,40 +52,31 @@ type CollapserProps = {
...
-          <UIIcon type="double-right" onClick={onCollapseAll} className={styles.btn} />
-        </UITooltip>
-      </div>
-    );
-  }
+  return (
+    <div className={styles.TimelineCollapser} data-test-id="TimelineCollapser">
+      <Tooltip content="Expand +1" placement="top">
+        <span>
+          <RightIcon onClick={onExpandOne} className={cx(styles.btn, styles.btnExpanded)} />
```
Elfo404, line 62: Any reason why you are not using [`IconButton` from `@grafana/ui`](https://developers.grafana.com/ui/latest/index.html?path=/story/buttons-iconbutton--simple)?

packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -56,40 +52,31 @@ type CollapserProps = {
...
+        </span>
+      </Tooltip>
+      <Tooltip content="Collapse +1" placement="top">
+        <span>
+          <RightIcon onClick={onCollapseOne} className={styles.btn} />
+        </span>
+      </Tooltip>
+      <Tooltip content="Expand All" placement="top">
+        <span>
+          <DoubleRightIcon onClick={onExpandAll} className={cx(styles.btn, styles.btnExpanded)} />
```
Elfo404, line 72: same as before, but maybe for this icon you need to add it to `packages/grafana-ui/src/types/icon.ts`

packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
zoltanbedi: hah I didn't know that I can use that as well. good catch.

packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -56,40 +36,27 @@ type CollapserProps = {
   onExpandAll: () => void;
 };
 
-function getTitle(value: string) {
+export default function TimelineCollapser(props: CollapserProps) {
```
hugohaggmark, line 39: Out of curiosity why isn't this declared as a functional component? Also I think we try to avoid default exports.

hugohaggmark (approved):
LGTM!

hugohaggmark:
If this is PR is for 7.3 remember to add cherry-pick needed label :)

Elfo404 (approved):
Nice!

packages/jaeger-ui-components/src/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
zoltanbedi: What do you mean? It is a functional component.

hugohaggmark: Hmm, I thought we had a style guideline for React functional components to be like `TimelineCollapser: FC<CollapserProps>` but we don't so my bad.

hugohaggmark: To clarify I knew this is a functional component but I didn't see the specific FC return type so that's why I asked because I thought we did have a style guideline :)

aocenas: @hugohaggmark I remember that when I was using `const Component: FC<CollapserProps> = () => {...}` It did not pick up the component name from the variable name (like in the react dev tools for example) and I had to manually add `displayName` which was a bit tedious. Is that still happening or was that fixed in some way?

hugohaggmark: @aocenas thanks, I've seen so many const Compontent: FC<...> lately I just thought that was our standard.

PR #28286, opened 2020-10-15
OAuth: Configurable user name attribute

### LINKED ISSUE #28224: OAuth: configurable user Name (displayName) attribute
It's possible to configure email attribute name for OAuth users to use attribute other than default `Email` for user's email. It would be nice also to make user name attribute configurable as well (now grafana uses default `name` or `display_name` field).

**Why is this needed**:
This can make OAuth more flexible.

alexanderzobnin:
This PR adds configuration options for specifying custom attribute name or path for user name.

This PR fixes #28224 

### DIFF
--- a/docs/sources/auth/generic-oauth.md
+++ b/docs/sources/auth/generic-oauth.md
@@ -80,6 +80,8 @@ Customize user login using `login_attribute_path` configuration option. Order of
 
 You can customize the attribute name used to extract the ID token from the returned OAuth token with the `id_token_attribute_name` option.
 
+Customize user's displayed name using `name_attribute_path` configuration option. Order of operations is the same as for the `login_attribute_path` option.
+
 ## Set up OAuth2 with Auth0
 
 1. Create a new Client in Auth0
--- a/docs/sources/auth/generic-oauth.md
+++ b/docs/sources/auth/generic-oauth.md
@@ -80,8 +80,6 @@ Customize user login using `login_attribute_path` configuration option. Order of
 
 You can customize the attribute name used to extract the ID token from the returned OAuth token with the `id_token_attribute_name` option.
 
-Customize user's displayed name using `name_attribute_path` configuration option. Order of operations is the same as for the `login_attribute_path` option.
-
 ## Set up OAuth2 with Auth0
 
 1. Create a new Client in Auth0
--- a/pkg/login/social/generic_oauth.go
+++ b/pkg/login/social/generic_oauth.go
@@ -250,7 +250,7 @@ func (s *SocialGenericOAuth) extractFromToken(token *oauth2.Token) *UserInfoJson
 
 	data.rawJSON = rawJSON
 	data.source = "token"
-	s.log.Debug("Received id_token", "raw_json", string(data.rawJSON), "data", data)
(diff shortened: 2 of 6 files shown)


pkg/login/social/generic_oauth.go
```diff
@@ -24,6 +24,8 @@ type SocialGenericOAuth struct {
 	emailAttributeName   string
 	emailAttributePath   string
 	loginAttributePath   string
+	nameAttributeName    string
```
sakjur, line 27: Why do we add both `nameAttributeName` and `nameAttributePath`? Isn't the `name` variant of other properties just a legacy setting from before we supported JMESPaths?

pkg/login/social/generic_oauth.go
```diff
@@ -107,12 +109,9 @@ func (s *SocialGenericOAuth) UserInfo(client *http.Client, token *oauth2.Token)
...
 		if userInfo.Name == "" {
-			if data.Name != "" {
-				s.log.Debug("Setting user info name from name field")
-				userInfo.Name = data.Name
-			} else if data.DisplayName != "" {
-				s.log.Debug("Setting user info name from display name field")
-				userInfo.Name = data.DisplayName
+			userInfo.Name = s.extractUserName(data)
+			if userInfo.Name != "" {
+				s.log.Debug("Set user info name from extracted name", "name", userInfo.Name)
```
papagian, line 114: I think that this debug message is redundant given that each special case in `extractUserName` is logged.

pkg/login/social/generic_oauth.go
alexanderzobnin: Hm, good question, I'll check it out. Looks like for login we only have `loginAttributePath` option.

papagian (approved):
LGTM, it would be nice to add some tests!

pkg/login/social/generic_oauth.go
alexanderzobnin: @marefr What do you think?

xlson (approved):

docs/sources/auth/generic-oauth.md
```diff
@@ -80,6 +80,8 @@ Customize user login using `login_attribute_path` configuration option. Order of
 
 You can customize the attribute name used to extract the ID token from the returned OAuth token with the `id_token_attribute_name` option.
 
+Customize user's displayed name using `name_attribute_path` configuration option. Order of operations is the same as for the `login_attribute_path` option.
```
sakjur, line 83: Unless we'll squeeze this into 7.3, we should split out the docs changes to a separate PR that's merged together with 7.4-beta1 because our docs release pipeline publishes all docs changes immediately.

```suggestion
You can set the user's display name with JMESPath using the `name_attribute_path` configuration option. It operates the same way as the `login_attribute_path` option.

> **Note:** `name_attribute_path` is added in 7.4.
```

sakjur (approved):
I left a comment about the docs, we can either split docs out into a separate PR in which case this is :+1: from me, otherwise we should wait for a technical writer to give a LGTM from their perspective.

docs/sources/auth/generic-oauth.md
alexanderzobnin: Yes, forget about this, good catch!

sakjur (approved):

PR #28295, opened 2020-10-15
main-view hidden when search is opened

### LINKED ISSUE #28266: Panel Alerts sit above Dashboard Search
**What happened**:
Whilst browsing through dashboards - panels without installed plugins render their error alerts above the dashboard search. They have no way of being closed and obscure text and prevent clicking to switch dashboard.

Kapture 2020-10-14 at 16 42 11

For the alerts to be hidden by the dashboard search so they don't prevent reading dashboard names or selecting a dashboard to view.

- Open a dashboard with a panel that uses a plugin that isn't installed.
- Open the dashboard search

It looks like the [grafana-ui/alert](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Alert/Alert.tsx#L96-L98) component has a z-index set which is higher than the dashboard search z-index. A solution could be that the call site dictates the layer an alert lives on rather than this css property living inside this component?

z-indexes can become unwieldy over time. In the past I've tackled scaling z-indexes across applications by using an array of strings and retrieving a strings index for use as the value. This approach allows a design system to define names for each layer, the code can be explicit in its usage, and z-index values are kept as small/low as possible. It also means that if new layers are introduced they can be slotted in where needed and the existing indexes will update as necessary.

I don't know if this idea is something that could help us keep check of our z-indexes in the future but thought I'd throw it out there for consideration.

**Environment**:
- Grafana version: Grafana v7.3.0-pre (3fed829b8b)
- User OS & Browser: Mac OSX Chrome
- Grafana plugins: N/A
- Others: N/A

maknik:
Alerts has higher z-index than content in search-wrapper and when alert appears on dashboard and user open the search, alert is still visible and covers space. Due to the fact that changing the z-index may cause a lot of problems, I propose to make all main-view hidden when search is open

Fixes #28266 

### DIFF
--- a/public/app/features/search/components/SearchWrapper.tsx
+++ b/public/app/features/search/components/SearchWrapper.tsx
@@ -4,9 +4,10 @@ import { UrlQueryMap } from '@grafana/data';
 import { getLocationQuery } from 'app/core/selectors/location';
 import { updateLocation } from 'app/core/reducers/location';
 import { connectWithStore } from 'app/core/utils/connectWithReduxStore';
-import { StoreState } from 'app/types';
+import { CoreEvents, StoreState } from 'app/types';
 import DashboardSearch from './DashboardSearch';
 import { defaultQueryParams } from '../reducers/searchQueryReducer';
+import appEvents from 'app/core/app_events';
 
 interface OwnProps {
   search?: string | null;
@@ -35,7 +36,9 @@ export const SearchWrapper: FC<Props> = memo(({ search, folder, updateLocation }
         partial: true,
       });
     }
+    appEvents.emit(CoreEvents.toggleSearchOpened);
   };
+  isOpen && appEvents.emit(CoreEvents.toggleSearchOpened);
 
   return isOpen ? <DashboardSearch onCloseSearch={closeSearch} folder={folder} /> : null;
 });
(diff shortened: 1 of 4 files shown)


torkelo:
not sure this is the right way to fix this. The problem is that the Alert component has z-index when it should not. The z-index should be on the application notifications container (where they are used and where z-index is needed)

maknik:
> not sure this is the right way to fix this. The problem is that the Alert component has z-index when it should not. The z-index should be on the application notifications container (where they are used and where z-index is needed)

As I understand after investigation, Alert component is used in many cases for example in case "Unexcpected error has occure" and this Alert is above anything, even search layer so it has to have z-index. The solution in this case is to customize Alert and allow not to use z-index for PluginLoadingError or as I did, hide the container when search layer is present to not bother about any z-index rules. If you prefer solution with modify Alert component I can modify my PR.

torkelo:
yes, but the zindex should not be on the Alert component, it should be on the containers where it needs to be placed on top of things

maknik:
But in case of general component like `ErrorBoundaryAlert` which wrapping other components and show Alert when something is wrong, the solution of removing z-index from Alert will broke this concept.

torkelo:
No it won't the z-index can be set on the container for app alerts , it should not be on the Alert component, the component needs to be able to be used in different scenarios

torkelo:
closing this over https://github.com/grafana/grafana/pull/28834

PR #28299, opened 2020-10-15
Drone: Use ${DRONE_TAG} in release pipelines, since it should work

aknuds1:
From the release build of v7.3.1-beta1, it did look as if ${DRONE_TAG} is available after all, so this PR makes the release pipelines use it (additionally fixes a bug). There's also some mixin linting enabled, since .drone.yml hadn't been re-generated on master.


kylebrandt (approved):

kylebrandt:
Just curious, what does the double `$$` do?

aknuds1:
> Just curious, what does the double `$$` do?

@kylebrandt Good question! It's necessary to escape `$` in .drone.yml since Drone uses it to substitute its own variables, f.ex. `${DRONE_TAG}`.

aknuds1:
@bergquist This one [fails](https://drone.grafana.net/grafana/grafana/4700/1/6) due to the mixin linting you introduced.

kylebrandt:

PR #28308, opened 2020-10-16
Drone: Fix grafana-mixin linting

aknuds1:
Fix the linting of grafana-mixin, plus a syntax error in Drone Starlark code.

@bergquist already OK-ed moving the grafana-mixin build logic into scripts.

### DIFF
--- a/scripts/build/ci-build/Dockerfile
+++ b/scripts/build/ci-build/Dockerfile
@@ -155,7 +155,11 @@ RUN cd /tmp && \
     tar xf x86_64-linux-musl-cross.tgz && \
     rm x86_64-linux-musl-cross.tgz
 
-RUN go get -u github.com/mgechev/revive@v1.0.2
-RUN mv ${GOPATH}/bin/revive /usr/local/bin/
+RUN go get -u github.com/mgechev/revive@v1.0.2 && \
+  mv ${GOPATH}/bin/revive /usr/local/bin/ && \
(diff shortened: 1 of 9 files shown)


scripts/build/ci-build/Dockerfile
```diff
@@ -155,7 +155,11 @@ RUN cd /tmp && \
     tar xf x86_64-linux-musl-cross.tgz && \
     rm x86_64-linux-musl-cross.tgz
 
-RUN go get -u github.com/mgechev/revive@v1.0.2
-RUN mv ${GOPATH}/bin/revive /usr/local/bin/
+RUN go get -u github.com/mgechev/revive@v1.0.2 && \
```
bergquist, line 158: how often do we rebuild this image? I dont know how frequent changes we have for mixtool tbh

bergquist (approved):
LGTM 

My comment can be fixed once if we find a problem with it.

scripts/build/ci-build/Dockerfile
aknuds1: @bergquist We don't have a particular schedule for it. If it should need frequent updates, we should automate rebuilding it via Drone. BTW, I ran into a problem installing mixtool with the -u flag, since some unkwon package has a resolution issue. That shows how important it is to install dependencies in the build image and not during each build, since it's extremely fragile and can stop our builds.

kylebrandt:

mahdiansyah1 (approved):

PR #28321, opened 2020-10-16
TestData: Remove useEffect that triggeres query on component load

torkelo:
Fixes so that testdata does not issue queries when the query editor is loaded, and fixes so the scenario api is not called for every query


hugohaggmark (approved):
LGTM!

public/app/plugins/datasource/testdata/QueryEditor.tsx
```diff
@@ -46,10 +46,6 @@ export const QueryEditor = ({ query, datasource, onChange, onRunQuery }: Props)
     onRunQuery();
   };
 
-  useEffect(() => {
-    onUpdate(query);
-  }, []);
-
```
Clarity-89, line 52: Hmm, strange, I remember the graphs weren't rendered properly on page reload after saving dashboard, so that's why I added this call on mount, but I just checked it again and it works without it too.

PR #28323, opened 2020-10-16
BackendSrv: Fixes queue countdown when unsubscribe is before response

hugohaggmark:
We recently found out that in some edge cases the [outer subscribe of the request queue](https://github.com/grafana/grafana/blob/8db5d750d0cc8d2f5a09fef3d86a1f2355a5de60/public/app/core/services/backend_srv.ts#L91) was called [before the inner response](https://github.com/grafana/grafana/blob/8db5d750d0cc8d2f5a09fef3d86a1f2355a5de60/public/app/core/services/backend_srv.ts#L78) was done. This made so the queue to never ended. 

Fixes #


torkelo (approved):
Looks good!

PR #28333, opened 2020-10-16
Graph: trying to dynamically detect size for axis

mckn:
I am not happy with this one at all. It somewhat works but it feels like a hack. It would have been better to do this within the uPlot library to properly measure the size of the "longest/tallest" label for each axis.

Have a look and see if this is something we want to use prior to the demo. If not I will probably close this one.

### DIFF
--- a/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
+++ b/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
@@ -103,7 +103,9 @@ export const GraphNG: React.FC<GraphNGProps> = ({ data, children, ...plotProps }
     scales.push(<Scale key="scale-x" scaleKey="x" time />);
   }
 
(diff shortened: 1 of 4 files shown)


dprokop:
@leeoniya do you have any plans on introducing dynamic axis sizing into uPlot? It would be super neat if we didn't have to specify the size statically. In a lot of cases the default value (60) makes the graph in the panel misaligned: 
image

In the new graph panel we do have a control over it via field config so that the User can actually adjust the size. But I think this will still produce some glitches in the UI long term.

mckn:
Closing for doing this in the uPlot library

leeoniya:
i've opened https://github.com/leeoniya/uPlot/issues/333 for discussion. let me know what you think.

mckn:
@leeoniya super, thanks! 🙏

PR #28334, opened 2020-10-16
AlertingNG: remove warn/crit from eval prototype

kylebrandt:
and misc cleanup

want to drop the warn/crit from the design. While it can be useful, it also complicates things. The complications span across all the components that make up the alerting logic so I think it will make the behavior hard to implement and hard for users to always follow.
Also, when thinking about making alerting so it can integrate with different IRMs, having more particular notions in the system makes it more likely to not fit with IRMs in a logical way.

This is a prototype behind a feature flag

### DIFF
--- a/pkg/services/ngalert/eval.go
+++ b/pkg/services/ngalert/eval.go
@@ -188,7 +188,7 @@ func (ng *AlertNG) LoadAlertCondition(dashboardID int64, panelID int64, conditio
 	return &condition, nil
 }
 
-// Execute runs Conditions' expressions or queries.
+// Execute runs the Condition's expressions or queries.
 func (c *Condition) Execute(ctx AlertExecCtx, fromStr, toStr string) (*ExecutionResults, error) {
 	result := ExecutionResults{}
 	if !c.IsValid() {
--- a/pkg/services/ngalert/eval.go
+++ b/pkg/services/ngalert/eval.go
@@ -1,4 +1,4 @@
-// Package eval executes the condition for an alert defintion, evaluates the condition results, and
+// Package eval executes the condition for an alert definition, evaluates the condition results, and
 // returns the alert instance states.
 package eval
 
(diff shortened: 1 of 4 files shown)


pkg/services/ngalert/eval.go
```diff
@@ -176,14 +184,14 @@ func (ng *AlertNG) LoadAlertConditions(dashboardID int64, panelID int64, conditi
...
-	return &conditions, nil
+	condition.RefID = conditionRefID
+	return &condition, nil
 }
 
-// Execute runs the WarnCondition and CritCondtion expressions or queries.
-func (conditions *Conditions) Execute(ctx AlertExecCtx, fromStr, toStr string) (*ExecutionResult, error) {
-	result := ExecutionResult{}
-	if !conditions.IsValid() {
+// Execute runs Conditions' expressions or queries.
```
kylebrandt, line 191:
```suggestion
// Execute runs the Condition's expressions or queries.
```

papagian (approved):
Nice!

pkg/services/ngalert/eval.go
```diff
@@ -33,66 +36,71 @@ func init() {
...
 	Ctx context.Context
 }
 
-// At least Warn or Crit condition must be non-empty
-type Conditions struct {
-	Condition string `json:"condition"`
+// Condition contains backend expressions and queries and the RefID
+// of the query or expression that will be evaluated.
+type Condition struct {
+	RefID string `json:"condition"`
```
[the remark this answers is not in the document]
kylebrandt: nope oversight

kylebrandt, line 54:
```suggestion
	RefID string `json:"refId"`
```

pkg/services/ngalert/eval.go
```diff
@@ -33,66 +36,71 @@ func init() {
...
 
-type EvalResult struct {
+// Result contains the evaluated state of an alert instance
+// identified by its labels.
+type Result struct {
 	Instance data.Labels
 	State    State // Enum
 }
 
+// State is a enum of the evaluation state for an alert instance.
```
aknuds1, line 79:
```suggestion
// State is an enum of the evaluation state for an alert instance.
```

aknuds1 (approved):
Looks good AFAICT. You go Kyle.

PR #28343, opened 2020-10-17
Templating: Replace all '$tag' in tag values query

krzysdabro:
Allows use of multiple `$tag` in tag values query.
Before this change only the first `$tag` would be replaced with value.

This PR allows users to use queries such as:
```
SHOW TAG VALUES FROM "test" WITH KEY = "host" WHERE $timeFilter AND ("tag1" =~ /^$tag$/ OR "tag2" =~ /^$tag$/)
```


torkelo (approved):

torkelo:
just FYI the tag values feature is going away in v8 (it has always been a experimental)

PR #28349, opened 2020-10-18
API: Query database from /api/health endpoint

### LINKED ISSUE #28214: The `/api/health` should query the database
It currently only checks if the database connection exists which will be true if Grafana connects using a SQL proxy even if the database itself is down. I suggest that we send a real query to the database instead and cache the result for 30 seconds to avoid causing to much traffic on the database

ceh:

Query the database from the `/api/health` endpoint by issuing a `SELECT` query to check if the database is healthy, rather than just verifying the connection with `Ping`.

Fixes #28214

I was a bit curious about the in-process bus, and what caching alternatives that is available, so I took a stab at fixing the issue.

What do you think of the approach @bergquist?

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -1,0 +1,21 @@
+package api
+
+import (
+	"time"
+
+	"github.com/grafana/grafana/pkg/bus"
+	"github.com/grafana/grafana/pkg/models"
+)
+
+func (hs *HTTPServer) databaseHealthy() bool {
+	const cacheKey = "db-healthy"
+
+	if cached, found := hs.CacheService.Get(cacheKey); found {
+		return cached.(bool)
+	}
+
+	healthy := bus.Dispatch(&models.GetDBHealthQuery{}) == nil
+
+	hs.CacheService.Set(cacheKey, healthy, time.Second*5)
+	return healthy
+}
--- a/pkg/api/health_test.go
+++ b/pkg/api/health_test.go
@@ -1,0 +1,190 @@
+package api
+
+import (
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/grafana/grafana/pkg/bus"
+	"github.com/grafana/grafana/pkg/infra/localcache"
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/grafana/grafana/pkg/setting"
+	"github.com/stretchr/testify/require"
+	macaron "gopkg.in/macaron.v1"
+)
+
+func TestHealthAPI_Version(t *testing.T) {
+	m, _ := setupHealthAPITestEnvironment(t)
+	setting.BuildVersion = "7.4.0"
+	setting.BuildCommit = "59906ab1bf"
+
+	bus.AddHandler("test", func(query *models.GetDBHealthQuery) error {
+		return nil
+	})
+
+	req := httptest.NewRequest(http.MethodGet, "/api/health", nil)
+	rec := httptest.NewRecorder()
+	m.ServeHTTP(rec, req)
+
+	require.Equal(t, 200, rec.Code)
+	expectedBody := `
+		{
+			"database": "ok",
+			"version": "7.4.0",
+			"commit": "59906ab1bf"
+		}
+	`
+	require.JSONEq(t, expectedBody, rec.Body.String())
+}
+
+func TestHealthAPI_AnonymousHideVersion(t *testing.T) {
+	m, hs := setupHealthAPITestEnvironment(t)
+	hs.Cfg.AnonymousHideVersion = true
+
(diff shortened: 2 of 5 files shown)


pkg/api/health.go
```diff
@@ -0,0 +1,25 @@
...
+
+	healthy := hs.databaseHealthy()
+
+	hs.CacheService.Set(cacheKey, healthy, time.Second*5)
+	return healthy
+}
+
+func (hs *HTTPServer) databaseHealthy() bool {
+	return bus.Dispatch(&models.GetDBHealthQuery{}) == nil
+}
```
aknuds1, line 25: Please just inline this method, since it's too simple to extract as a separate function (doesn't outweigh the cost of the abstraction).

pkg/api/health.go
```diff
@@ -0,0 +1,25 @@
+package api
+
+import (
+	"time"
+
+	"github.com/grafana/grafana/pkg/bus"
+	"github.com/grafana/grafana/pkg/models"
+)
+
+func (hs *HTTPServer) databaseHealthyWithCache() bool {
```
aknuds1, line 10: Please inline `databaseHealthyWithCache` and rename this one to drop the WithCache suffix.
```suggestion
func (hs *HTTPServer) databaseHealthy() bool {
```

pkg/api/health.go
```diff
@@ -0,0 +1,25 @@
...
+	cacheKey := "db-healthy"
```
aknuds1, line 11:
```suggestion
	const cacheKey = "db-healthy"
```

pkg/api/health_test.go
```diff
@@ -0,0 +1,190 @@
...
+	expectedBody := `
+		{
+			"database": "ok"
+		}
+	`
+	require.JSONEq(t, expectedBody, rec.Body.String())
+}
+
+func TestHealthAPI_DatabaseHealthy(t *testing.T) {
+	cacheKey := "db-healthy"
```
aknuds1, line 64:
```suggestion
	const cacheKey = "db-healthy"
```

pkg/api/health_test.go
```diff
@@ -0,0 +1,190 @@
...
+	`
+	require.JSONEq(t, expectedBody, rec.Body.String())
+
+	healthy, found = hs.CacheService.Get(cacheKey)
+	require.True(t, found)
+	require.True(t, healthy.(bool))
+}
+
+func TestHealthAPI_DatabaseUnhealthy(t *testing.T) {
+	cacheKey := "db-healthy"
```
aknuds1, line 95:
```suggestion
	const cacheKey = "db-healthy"
```

pkg/api/health_test.go
```diff
@@ -0,0 +1,190 @@
...
+	`
+	require.JSONEq(t, expectedBody, rec.Body.String())
+
+	healthy, found = hs.CacheService.Get(cacheKey)
+	require.True(t, found)
+	require.False(t, healthy.(bool))
+}
+
+func TestHealthAPI_DatabaseHealthCached(t *testing.T) {
+	cacheKey := "db-healthy"
```
aknuds1, line 126:
```suggestion
	const cacheKey = "db-healthy"
```

pkg/api/http_server.go
```diff
@@ -410,7 +410,7 @@ func (hs *HTTPServer) apiHealthHandler(ctx *macaron.Context) {
 		data.Set("commit", setting.BuildCommit)
 	}
 
-	if err := bus.Dispatch(&models.GetDBHealthQuery{}); err != nil {
+	if !hs.databaseHealthyWithCache() {
```
aknuds1, line 413:
```suggestion
	if !hs.databaseHealthy() {
```

aknuds1 (changes requested):
Generally looks pretty good AFAICT, with some suggestions for improvement. I think @bergquist should give feedback though since he requested the change.

pkg/api/health.go
```diff
@@ -0,0 +1,25 @@
...
+
+	if cached, found := hs.CacheService.Get(cacheKey); found {
+		return cached.(bool)
+	}
+
+	healthy := hs.databaseHealthy()
+
+	hs.CacheService.Set(cacheKey, healthy, time.Second*5)
```
bergquist, line 19:
```suggestion
	hs.CacheService.Set(cacheKey, healthy, time.Second*10)
```

bergquist: I think the default value can be a bit higher. Looks good otherwise feature-wise.

bergquist (commented):
Looks good feature-wise :)

ceh:
Thanks for the review @aknuds1! I've updated the PR, please take another look.

papagian (approved):
Great work! thank you for contributing.

aknuds1 (approved):
LGTM, nice one :)

aknuds1:
Thanks for the contribution!

PR #28357, opened 2020-10-19
OAuth: Able to skip auto login

### LINKED ISSUE #25993: Allow local user login even with oauth_auto_login = true
Forgive me if this is already possible, I did a little bit of digging and couldn't find anything...

It would be nice to have the ability to bypass the `oauth_auto_login` functionality using a URL query string (maybe `?login` or `?no_sso`)

**Why is this needed**:

We want end users to always use oauth, but there are times when we also want to use a local account for superuser maintenance.

Being able to login from https://grafana.example.com/?login would let us keep the streamlined auto-login experience for users while giving a local admin a way into the system for maintenance.

_Yes, we could grant admin privileges to an oauth user, but we'd prefer to keep all oauth users on the `editor` role and use a separate account for admin duties._

alexanderzobnin:
This PR adds special `internal` param to the login path which makes it possible to disable OAuth auto login and log in with internal grafana user. Usage: `grafana.example.com/login?disableAutoLogin` or `grafana.example.com/login?disableAutoLogin=true`.

Fixes #25993

In order to prevent duplicated user sessions, all OAuth providers (and SAML also in the Enterprise) disabled when `internal` flag used.

### DIFF
--- a/pkg/api/login.go
+++ b/pkg/api/login.go
@@ -83,7 +83,7 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
 
 	urlParams := c.Req.URL.Query()
 	if _, disableAutoLogin := urlParams["disableAutoLogin"]; disableAutoLogin {
-		hs.log.Info("Using internal login, skipping OAuth auto login")
+		hs.log.Debug("Auto login manually disabled")
 		c.HTML(200, getViewIndex(), viewData)
 		return
 	}
--- a/pkg/api/login.go
+++ b/pkg/api/login.go
@@ -82,7 +82,7 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
 	}
 
 	urlParams := c.Req.URL.Query()
-	if _, isInternal := urlParams["internal"]; isInternal {
+	if _, disableAutoLogin := urlParams["disableAutoLogin"]; disableAutoLogin {
 		hs.log.Info("Using internal login, skipping OAuth auto login")
 		c.HTML(200, getViewIndex(), viewData)
 		return
--- a/pkg/api/login.go
+++ b/pkg/api/login.go
@@ -81,6 +81,13 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
 		return
 	}
 
+	urlParams := c.Req.URL.Query()
+	if _, isInternal := urlParams["internal"]; isInternal {
+		hs.log.Info("Using internal login, skipping OAuth auto login")
+		c.HTML(200, getViewIndex(), viewData)
+		return
+	}
+
 	enabledOAuths := make(map[string]interface{})
 	for key, oauth := range setting.OAuthService.OAuthInfos {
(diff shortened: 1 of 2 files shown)


pkg/api/login_test.go
```diff
@@ -500,6 +500,39 @@ func TestLoginOAuthRedirect(t *testing.T) {
...
+		Enabled:      true,
+		AllowSignup:  true,
+		Name:         "github",
+	}
+	setting.OAuthAutoLogin = true
+	sc.m.Get(sc.url, sc.defaultHandler)
+	sc.fakeReqNoAssertions("GET", sc.url).exec()
+
+	// Shouldn't redirect to the OAuth login URL
+	assert.Equal(t, sc.resp.Code, 200)
```
sakjur, line 533: How much work would it be to add another condition to check that we're being redirected (status code would be sufficient) when we want that to happen?

pkg/api/login.go
```diff
@@ -81,6 +81,13 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
 		return
 	}
 
+	urlParams := c.Req.URL.Query()
+	if _, isInternal := urlParams["internal"]; isInternal {
```
sakjur, line 85: I'm not sure about the `internal` name, not sure what to call it, but I'm thinking something like something like `noRedirect`?

pkg/api/login.go
```diff
@@ -81,6 +81,13 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
...
+		hs.log.Info("Using internal login, skipping OAuth auto login")
```
sakjur, line 86:
```suggestion
		hs.log.Debug("Redirect manually overridden, skipping auto login")
```

Mostly Info->Debug, since this isn't really of interest to the server admin unless something is wonky, but also generalized the language a little. You can still manually click the OAuth provider when you force not redirecting, tried to reflect that in the new language.

sakjur (changes requested):
I think it looks mostly good, but I don't think we should call the parameter `internal` since you can still use the OAuth provider by manually initializing that login, you're just skipping the automatic redirect.

pkg/api/login_test.go
alexanderzobnin: There's a test above that checks auto login (redirected to the `/login/github/`).

pkg/api/login.go
alexanderzobnin: `noRedirect` sounds like we only disable redirect, but allow to login with options other than login/password. But in this case OAuth disabled on the login page, so user can only use login/password.

alexanderzobnin: Maybe something like `disableAutoLogin` or `skipAutoLogin`?

sakjur: > But in this case OAuth disabled on the login page, so user can only use login/password.

Ah, I didn't know that. In that case `internal` makes a little more sense to me.

`disableAutoLogin` sounds good to me.

sakjur (approved):

xlson (approved):
LGTM!

PR #28360, opened 2020-10-19
Grafana/ui: auto focus threshold editor input

### LINKED ISSUE #25479: Creating new Threshold should autofocus the threshold input
Currently, when a new threshold is created the focus stays with the _Add threshold_ button when user clicks it (see video below). 

To improve the usability of the thresholds editor it would be great if creating a new threshold would automatically focus the new threshold's input so that you can type in the value straight away.

Kapture 2020-10-19 at 08 43 51

jackw:
This PR improves the UX of the new threshold editor by autofocusing a new threshold input so that the user can set a value straight away.

Kapture 2020-10-19 at 11 52 14

Fixes #25479

Cleaned up the commented out `margin-bottom` to silence stylelint.

### DIFF
--- a/packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
+++ b/packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
@@ -38,7 +38,7 @@ interface State {
 }
 
 export class ThresholdsEditor extends PureComponent<Props, State> {
-  private inputRef: React.RefObject<HTMLInputElement>;
+  private latestThresholdInputRef: HTMLInputElement | null;
 
   constructor(props: Props) {
     super(props);
@@ -47,8 +47,12 @@ export class ThresholdsEditor extends PureComponent<Props, State> {
     steps[0].value = -Infinity;
 
     this.state = { steps };
-    this.inputRef = React.createRef();
+    this.latestThresholdInputRef = null;
   }
+
+  setLatestThresholdInputRef = (node: HTMLInputElement | null) => {
+    this.latestThresholdInputRef = node;
+  };
 
   onAddThreshold = () => {
     const { steps } = this.state;
@@ -71,8 +75,8 @@ export class ThresholdsEditor extends PureComponent<Props, State> {
     sortThresholds(newThresholds);
 
     this.setState({ steps: newThresholds }, () => {
-      if (this.inputRef.current) {
-        this.inputRef.current.focus();
+      if (this.latestThresholdInputRef) {
+        this.latestThresholdInputRef.focus();
       }
       this.onChange();
     });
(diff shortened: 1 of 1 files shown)


mckn (approved):
Kapaaow! Looks great!

packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
```diff
@@ -167,6 +175,7 @@ export class ThresholdsEditor extends PureComponent<Props, State> {
         key={isPercent.toString()}
         onChange={(event: ChangeEvent<HTMLInputElement>) => this.onChangeThresholdValue(event, threshold)}
         value={threshold.value}
+        ref={this.inputRef}
```
dprokop, line 178: Hm, since this is rendered in a loop, won't it create some weird edge cases when the current ref does not actually refer to the last added input?

dprokop (approved):
Looks good, but left one question.

packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
jackw: Yep, the behaviour you're describing is the reason why I ended up with the `focus` occurring in the `setState` callback. On mount the ref will point to the lowest editable threshold input but on clicking the add threshold button it tracks the last added and I've yet to find a scenario where it doesn't. However I'm now wondering if an alternative approach would be to keep track of every input as an array of refs and then focus the last added? What do you think?

I tried a few other approaches around using `autoFocus` and letting react do the work but they all caused varying issues when switching tabs and the tree was torn down or rebuilt.

dprokop: Ok, I see. And what about applying the ref to the topmost input only?

packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
```diff
@@ -38,15 +38,22 @@ interface State {
 }
 
 export class ThresholdsEditor extends PureComponent<Props, State> {
+  private latestThresholdInputRef: HTMLInputElement | null;
```
dprokop, line 41: One last question - why are you not using `React.createRef` for this?

dprokop (approved):
OK - this is great now and works as expected!

packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.tsx
jackw: Hmm... I think I read the react docs a little too literally around when to use callback refs. I've gone back to using `createRef`. 👍

PR #28364, opened 2020-10-19
Instrumentation: Add histogram for request duration

bergquist:
While summaries are ok when running single instances they are not really useful when running multiple instances. 

I think we can remove the `http_request_summary` in 7.3 as histograms are better in most situations. 

This will add ~ 240 series for active instance which currently exposes ~ 550 series


pkg/middleware/request_metrics.go
```diff
@@ -23,7 +24,17 @@ func init() {
 		},
 	)
 
-	prometheus.MustRegister(httpRequestsInFlight)
+	httpRequestDurationHistogram = prometheus.NewHistogramVec(
```
aknuds1, line 27: Is there any reason why httpRequestDurationHistogram is initialized only after declaration? Why not initialize at the same time of declaring the variable?

aknuds1 (approved):
Looks good AFAICT, but it looks to me as if the variable `httpRequestDurationHistogram` should be initialized at once and not in a later step.

pkg/middleware/request_metrics.go
bergquist: I have to admit that that's the pattern I've seen in other places. But that doesn't mean it makes sense. I'll double-check.

aknuds1: Thanks :) I think it's safer to initialize on declaration, and this should be the pattern unless there's a reason not to.

marefr (approved):
Agree with Arve's feedback, but besides that looks good. Having this histogram by default is intended?

bergquist:
Punting this to 7.3

bergquist:
This feature requires the `http_request_histogram` feature flag to be enabled

marefr (approved):
LGTM

kylebrandt:

Issue #28350, opened 2020-10-18
Azure: Problem with Microsoft.NetApp/../../volumes


seanluce:
When trying to create a panel for Azure metric 'Microsoft.NetApp/netAppAccounts/capacityPools/volume' the below error message is displayed:

**The Resource 'Microsoft.NetApp/netAppAccounts/capacityPools' under resource group 'my.rg' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix**

Seems like the API call is missing '/volumes'.

Creating panels for Azure metric 'Microsoft.NetApp/netAppAccounts/capacityPools' works as expected.

torkelo:
Please include all the network requests made by Grafana when building the query & the responses. Not sure why this particular metric would fail. Maybe some response from the Azure API that is wrong?

seanluce:
Thank you, Torkelo. I am not sure how to include the information that you have requested.

The problem occurs when I try to select a 'Metric Namespace' from the dropdown after selecting a volume.

Picture of error: https://ibb.co/D1mhB6B

It is a bit suspect that the error does not reference the path to '../volumes' but only to the parent resource '../capacityPools' considering that the resource in question is a volume.

migel0:
Hi
I hope give you some light.

It seems Grafama request/looking for volumes into   'Microsoft.NetApp/netAppAccounts/capacityPools
instead
/Microsoft.NetApp/netAppAccounts/${NetApp accountNAME}/capacityPools

An example of the request made over Azure Portal could be
![image](https://user-images.githubusercontent.com/42206588/96715420-d63f4600-13a3-11eb-9da2-6af5b7171601.png)

```
relativeUrl: "/subscriptions/43534-543534-543-534-53453/resourceGroups/MygreatnetappResorucegroup/providers/Microsoft.NetApp/netAppAccounts/MyNetappGreatresourcename/capacityPools/cp-standard/volumes/mymarvelous-vol-/providers/microsoft.Insights/metrics?timespan=2020-10-20T11:00:00.000Z/2020-10-21T11:15:00.000Z&interval=PT15M&metricnames=XregionReplicationRelationshipTransferring&aggregation=average&metricNamespace=microsoft.netapp%2Fnetappaccounts%2Fcapacitypools%2Fvolumes&autoadjusttimegrain=true&validatedimensions=false&api-version=2019-07-01"
```

wbrowne:
[...]

PR #28369, opened 2020-10-19
Fixed capitalization of 'volumes'

seanluce:
'Microsoft.NetApp/netAppAccounts/capacityPools/Volumes'  -> 'Microsoft.NetApp/netAppAccounts/capacityPools/volumes'

Fixes #


kylebrandt:
Does this actually fix an error, or does it only update it match the documentation? (Not to say it is not good either way, but do not have the resource handy to test against, and curious if the calls are case sensitive or not).

seanluce:
Possibly related to: https://github.com/grafana/grafana/issues/28350

https://docs.microsoft.com/en-us/azure/templates/microsoft.netapp/2019-07-01/netappaccounts/capacitypools/volumes#property-values

seanluce:
> Does this actually fix an error, or does it only update it match the documentation? (Not to say it is not good either way, but do not have the resource handy to test against, and curious if the calls are case sensitive or not).

I am trying to figure out how I can test this. I do not that regardless of Azure Subscription, I cannot get a panel created for Azure NetApp Files' Volumes.

kylebrandt (approved):
Thank you

PR #28370, opened 2020-10-19
PanelEdit: Prevent the preview pane to be resized further than window height

### LINKED ISSUE #27990: PanelEdit: Protect against resizing query region beyond viewport
**What happened**:
DIVs at the bottom in EDIT are not scrollable when the graph itself takes much of the screen's real estate. Additionally, when you accidentally resized the graph to be very large, you lose the resize handle so you can't resize it back.

grafana_scroll_bug

Scrolls. Set max height of the graphs/panels in Edit view

Edit any panel, make the widget large.

Reproducible on all browsers, even on https://play.grafana.org/d/000000012/grafana-play-home?editPanel=2&orgId=1

peterholmberg:
Sets a maxSize on the SplitPane which prevents the topPane from being resized beyond the window height. Also store the value as a percent so if you resize the window the topPane will adjust after the new window size (needs refresh though). 

Fixes #27990

The needs refresh part doesn't feel super, but not sure how to solve this without a rerender loop?


dprokop:
> The needs refresh part doesn't feel super, but not sure how to solve this without a rerender loop?

@peterholmberg I think there are two ways.
1. Debouncing - but won't be bulletproof
2. Faking post-window resize event. Quick sketch of how this could look:

```ts
const rafToken = useRef<number>();

window.addEventListener('resize', () => {
  if(rafToken.current !== undefined) {
    window.cancelAnimationFrame(rafToke.current);
  }
  window.requestAnimationFrame(() => {
    // do the resize thing
  });
});
```

dprokop (commented):
@peterholmberg this seems to work as expected, nice!

Since you are on it, maybe you could also fix the issue with the side pane? You can end up in a situation when the side pane cannot be resized if you make it full width, see the video:

Kapture 2020-10-20 at 10 20 19

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -131,11 +131,16 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
-    updatePanelEditorUIState({
-      [targetPane]: size,
-    });
+    if (pane === Pane.Top) {
+      updatePanelEditorUIState({
+        topPaneSize: `${(size / window.innerHeight) * 100}%`,
+      });
+    } else {
+      updatePanelEditorUIState({
+        rightPaneSize: size,
```
dprokop, line 141: Would be awesome to have percentages here as well to be aligned with top pane sizing.

jackw (approved):
LGTM! 🚀

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -58,19 +58,48 @@ interface DispatchProps {
...
+    window.removeEventListener('resize', this.updateSplitPaneSize);
   }
 
+  updateSplitPaneSize = () => {
+    if (this.rafToken.current !== undefined) {
+      window.cancelAnimationFrame(this.rafToken.current!);
+    }
+    window.requestAnimationFrame(() => {
+      this.setState({ rightPanelSize: this.props.uiState.rightPaneSize });
+      this.setState({ topPanelSize: this.props.uiState.topPaneSize });
```
dprokop, line 99: Can we do in a single pass?
```suggestion
      this.setState({ 
        rightPanelSize: this.props.uiState.rightPaneSize,
        topPanelSize: this.props.uiState.topPaneSize
      });
```

peterholmberg: Gah, yes! No idea what happened there :smile:

dprokop (commented):
Seems to work as expected! I have noticed few things we need to followup on but in the context of this PR this one is relevant:

When resizing the side pane's min width(300px) is not being respected and the pane is being resized below that limit:
Kapture 2020-10-21 at 09 50 41


I think we could maybe collapse the side pane when there is no space available to show it? @torkelo @peterholmberg  WDYT?

Additional problems I've noticed and will open separate issues:
1. There's a lot of stuttering when resizing the panel editor window
2. Bottom pane's tabs are rendered incorrectly when size of the window is small:
image

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -62,15 +66,28 @@ type Props = OwnProps & ConnectedProps & DispatchProps;
...
 
   componentWillUnmount() {
     this.props.panelEditorCleanUp();
+    window.removeEventListener('resize', this.updateSplitPaneSize);
   }
 
+  updateSplitPaneSize = () => {
+    if (this.rafToken.current !== undefined) {
+      window.cancelAnimationFrame(this.rafToken.current!);
+    }
```
dprokop, line 85: This is never executed as you are never setting the rafToken. You need to assign it with rAF result for this to work.

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -62,15 +66,28 @@ type Props = OwnProps & ConnectedProps & DispatchProps;
 
 export class PanelEditorUnconnected extends PureComponent<Props> {
   querySubscription: Unsubscribable;
+  rafToken = createRef<number>();
 
   componentDidMount() {
     this.props.initPanelEditor(this.props.sourcePanel, this.props.dashboard);
+
+    window.addEventListener('resize', debounce(this.updateSplitPaneSize, 100));
```
dprokop, line 74: I think the debounce is not necessary here. We should use **either** debounce or rAF

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -131,11 +148,16 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
     }
 
-    const targetPane = pane === Pane.Top ? 'topPaneSize' : 'rightPaneSize';
     const { updatePanelEditorUIState } = this.props;
-    updatePanelEditorUIState({
-      [targetPane]: size,
-    });
+    if (pane === Pane.Top) {
+      updatePanelEditorUIState({
+        topPaneSize: `${(size / window.innerHeight) * 100}%`,
```
dprokop, line 154: I suggest keeping numbers in the state rather than strings. Can't see any benefit in having percentage strings persisted. What is interesting from the component rendering POV is the ratio between the pane size vs window size. This way we can easily switch between % and px if we decide to.

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -313,10 +344,21 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
     const { uiState } = this.props;
 
+    // Limit options pane width to 90% of screen.
+    const maxWidth = window.innerWidth * 0.9;
+
+    // Need to handle when width is relative. ie a percentage of the viewport
+    const width =
+      typeof uiState.rightPaneSize === 'string'
+        ? (parseFloat(uiState.rightPaneSize) / 100) * window.innerWidth
+        : uiState.rightPaneSize;
```
dprokop, line 354: With my comment above you can just assume the persisted state is stored as numbers and change this condition to check whether the number is <1 (storing ration) or >1 (storing actual width)

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
dprokop: I wasn't precise enough - instead of percentages I have meant pane size/ window size ratio

dprokop (approved):

PR #28371, opened 2020-10-19
Docs: updated cmd to build docs locally to generate docs prior to building site.

mckn:
Adds a make command to generate docs prior to building the docs locally.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,13 +2,20 @@
 
 IMAGE = grafana/docs-base:latest
 
-docs:
+docs-generate-frontend:
+	yarn install --pure-lockfile --no-progress
+	../scripts/ci-reference-docs-build.sh
+	yarn run packages:clean
+
+docs-no-generate-frontend:
+	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
+
+docs: docs-generate-frontend
 	docker pull ${IMAGE}
 	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
 
-docs-no-pull:
-	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
+docs-no-pull: docs-generate-frontend docs-no-generate-frontend
 
-docs-test:
+docs-test: docs-generate-frontend
 	docker pull ${IMAGE}
 	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest --rm -it $(IMAGE) /bin/bash -c 'make prod'
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,7 +4,9 @@ IMAGE = grafana/docs-base:latest
 
 docs-generate-frontend:
 	yarn install --pure-lockfile --no-progress
-	../scripts/ci-reference-docs-build.sh
+	yarn run packages:build
+	yarn run packages:docsExtract
+	yarn run packages:docsToMarkdown
 	yarn run packages:clean
 
 docs-no-generate-frontend:
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,9 +1,9 @@
-.PHONY: docs docs-test
+.PHONY: docs docs-test docs-no-pull docs-generate-frontend docs-no-generate-frontend
 
 IMAGE = grafana/docs-base:latest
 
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -2,13 +2,20 @@
 
 IMAGE = grafana/docs-base:latest
 
-docs:
-	docker pull ${IMAGE}
+docs-generate-frontend:
+	yarn install --pure-lockfile --no-progress
+	../scripts/ci-reference-docs-build.sh
```
aknuds1, line 7: Can we please avoid using a CI script in Makefile? I think it's a bad idea to ask people to use CI scripts locally, as those scripts might be changed to do something that shouldn't happen locally, exactly for the reason that they're meant for CI.

I would instead just extract the `yarn` commands from that script.

docs/Makefile
```diff
@@ -2,13 +2,22 @@
 
 IMAGE = grafana/docs-base:latest
 
-docs:
-	docker pull ${IMAGE}
+docs-generate-frontend:
+	yarn install --pure-lockfile --no-progress
```
aknuds1, line 6: Is `--no-progress` useful locally? Should think a human user would like to see progress.

docs/Makefile
```diff
@@ -2,13 +2,22 @@
...
+	yarn run packages:docsExtract
+	yarn run packages:docsToMarkdown
+	yarn run packages:clean
+
+docs-no-generate-frontend:
 	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
 
-docs-no-pull:
+docs: docs-generate-frontend
+	docker pull ${IMAGE}
```
aknuds1, line 16: Can we use the `$()` form to be consistent with the rest of the file?
```suggestion
	docker pull $(IMAGE)
```

docs/Makefile
```diff
@@ -2,13 +2,22 @@
 
 IMAGE = grafana/docs-base:latest
 
-docs:
-	docker pull ${IMAGE}
+docs-generate-frontend:
```
aknuds1, line 5: You need to add this to .PHONY.

docs/Makefile
```diff
@@ -2,13 +2,22 @@
...
+	yarn install --pure-lockfile --no-progress
+	yarn run packages:build
+	yarn run packages:docsExtract
+	yarn run packages:docsToMarkdown
+	yarn run packages:clean
+
+docs-no-generate-frontend:
```
aknuds1, line 12: You need to add this to .PHONY.

docs/Makefile
```diff
@@ -2,13 +2,22 @@
...
 	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
 
-docs-no-pull:
+docs: docs-generate-frontend
```
aknuds1, line 15: You need to add this to .PHONY.

docs/Makefile
```diff
@@ -2,13 +2,22 @@
...
+	docker pull ${IMAGE}
 	docker run -v $(shell pwd)/sources:/hugo/content/docs/grafana/latest -p 3002:3002 --rm -it $(IMAGE) /bin/bash -c 'make server'
 
-docs-test:
+docs-no-pull: docs-generate-frontend docs-no-generate-frontend
```
aknuds1, line 19: You need to add this to .PHONY.

aknuds1 (changes requested):

docs/Makefile
mckn: Sure, I have not changed this but why not :)

aknuds1: OK, I see it's inconsistent already, but mostly parens are used.

aknuds1 (approved):

docs/Makefile
```diff
@@ -1,14 +1,23 @@
...
 IMAGE = grafana/docs-base:latest
 
-docs:
-	docker pull ${IMAGE}
+docs-generate-frontend:
+	yarn install --pure-lockfile
+	yarn run packages:build
+	yarn run packages:docsExtract
+	yarn run packages:docsToMarkdown
+	yarn run packages:clean
```
aknuds1, line 10: Already approved, but got a question, can we move this to a `yarn` script to make things easier? Then CI could also use the same `yarn` script.

mckn: hmm, yeah - I think that should be possible. Maybe do it in a follow up PR? If that is ok?

aknuds1: Yeah sure, fine by me.

xlson (approved):
Thanks for fixing this!

@AgnesToulet can you test if this works on windows too?

PR #28375, opened 2020-10-19
Plugins: Compose filesystem paths with filepath.Join

aknuds1:
Fix a bug I saw in plugins code of using `path.Join` instead of `filepath.Join` to join filesystem paths. Also fixing it in some other spots I saw.


marefr (approved):
LGTM great

joanlopez (commented):
Cool, thanks @aknuds1!! 🙏

PR #28377, opened 2020-10-19
AlertingNG: manage and evaluate alert definitions via the API

papagian:
- Introduces API endpoints for creating/retrieving/updating/deleting alert definitions via the API
- Modifies the eval API to get `alertDefinitionId` query parameter and load the respective alert definition
- Move everything under `pkg/services/ngalert`

Fixes #

Remaining things to do:
- [ ] database tests
- [ ] api tests

Example curl command for creating alert definition:
`curl -H "Content-Type: application/json" -XPOST http://admin:admin@localhost:3000/api/alert-definitions/ -d @<file>.json`

Example [JSON](https://gist.github.com/papagian/508473945926a16cdacf23ff5f6a5dec) for creating alert definition

**Updated description**
I have switched from `tsdb.Query` to `backend.DataQuery`.
I have updated the above JSON to reflect the changes (basically to include timeRange).
~This requires grafana/grafana-plugin-sdk-go#230~

After merging with master there was an issue with service initialisation order. More specifically the grafana server would fail to start when trying to register a backend plugin and the `backendplugin.manager` was not yet initialised.
Therefore I have [introduced](https://github.com/grafana/grafana/pull/28377/commits/82987de2bd8cfee2980b8a45a199c94f04f36faf) a new service priority `MediumHigh` and assigned it in the `backendplugin.manager`


pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/grafana/grafana/pkg/tsdb"
+)
+
+// AlertDefinition is the model for alert definitions in Alerting NG.
+type AlertDefinition struct {
+	Id        int64
+	OrgId     int64
+	Name      string
+	Condition string
+	Data      []tsdb.Query
```
kylebrandt, line 17: I think we may want to make this a backend.DataQuery as soon as we can. That has the timerange attached to the query, which we will want to get in ASAP on the frontend.

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,133 @@
...
+		if dsName != "__expr__" {
+			datasourceID, err := query.Model.Get("datasourceId").Int64()
+
+			ds, err = ng.DatasourceCache.GetDatasource(datasourceID, signedInUser, skipCache)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		if ds == nil {
```
kylebrandt, line 99: I think `&& ! dsName == "expr"` since gel will not have a entry in the ds cache I don't believe.

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// intervalMS is calculated by the frontend
+		// should we do something similar?
+		if query.Model.Get("intervalMs").MustString() == "" { // GEL requires intervalMs inside the query JSON
+			query.Model.Set("intervalMs", 1000)
+		}
+
+		condition.QueriesAndExpressions = append(condition.QueriesAndExpressions, tsdb.Query{
+			RefId:         query.Model.Get("refId").MustString(""),
```
kylebrandt, line 126: query.RefId ?

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,133 @@
...
+
+		if ds == nil {
+			return nil, errors.New("No datasource reference found")
+		}
+
+		if dsName == "" {
+			query.Model.Set("datasource", ds.Name)
+		}
+
+		if query.Model.Get("datasourceId").MustString() == "" {
```
kylebrandt, line 107: This and org would MustInt64 and not string?

Also, I think they would come from datasource property, not the model?

kylebrandt: (this may not be relevant if switching to backend.dataquery

kylebrandt:
With a couple tweaks to ngalert:

```
-               if ds == nil {
+               if ds == nil && dsName != "__expr__" {
                        return nil, errors.New("No datasource reference found")
                }
 
@@ -101,11 +101,11 @@ func (ng *AlertNG) LoadAlertCondition(alertDefinitionID int64, signedInUser *mod
                        query.Model.Set("datasource", ds.Name)
                }
 
-               if query.Model.Get("datasourceId").MustString() == "" {
+               if query.Model.Get("datasourceId").MustInt64() == 0 {
                        query.Model.Set("datasourceId", ds.Id)
                }
 
-               if query.Model.Get("orgId").MustString() == "" { // GEL requires orgID inside the query JSON
+               if query.Model.Get("orgId").MustInt64() == 0 { // GEL requires orgID inside the query JSON
                        query.Model.Set("orgId", alertDefinition.OrgId)
                }
 
@@ -120,7 +120,7 @@ func (ng *AlertNG) LoadAlertCondition(alertDefinitionID int64, signedInUser *mod
                }
 
                condition.QueriesAndExpressions = append(condition.QueriesAndExpressions, tsdb.Query{
-                       RefId:         query.Model.Get("refId").MustString(""),
+                       RefId:         query.RefId,
                        MaxDataPoints: query.Model.Get("maxDataPoints").MustInt64(100),
                        IntervalMs:    query.Model.Get("intervalMs").MustInt64(1000),
                        QueryType:     query.Model.Get("queryType").MustString(""),
```

Was able to get this working with the following:

```

@alertId = 8

###

POST http://admin:admin@localhost:3000/api/alert-definitions/
Content-Type: application/json

{
    "name": "Kyle's first alert",
    "condition": {
        "refId": "A",
        "queriesAndExpressions": [
            {
                "model": {
                    "expression":"2 + 2 > 1",
                    "type":"math",
                    "datasource": "__expr__",
                    "datasourceId": -100
                },
                "refId":"A",
                "datasource": {
                    "name":"__expr__",
                    "id": -100,
                    "orgId": 1
                }
            }
        ]
    }
}

###
GET http://admin:admin@localhost:3000/api/alert-definitions/{{alertId}}

###
GET http://admin:admin@localhost:3000/api/alert-definitions/eval/{{alertId}}
```

image

(VSCode REST client seems handy https://marketplace.visualstudio.com/items?itemName=humao.rest-client)

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,136 @@
...
+			{Name: "data", Type: migrator.DB_Text, Nullable: false},
+		},
+	}
+	// create table
+	mg.AddMigration("create alert_definition table", migrator.NewAddTableMigration(alertDefinition))
+}
+
+// LoadAlertCondition returns a Condition object for the given alertDefintionId.
+func (ng *AlertNG) LoadAlertCondition(alertDefinitionID int64, signedInUser *models.SignedInUser, skipCache bool) (*eval.Condition, error) {
+	// get queries from the dashboard (because GEL expressions cannot be stored in alerts so far)
```
kylebrandt, line 76:
```suggestion
```

kylebrandt:
So it seems the current source of truth for the datasource is the datasource name (within the org).

Frontend gets it at:
https://github.com/grafana/grafana/blob/74d1d3c6a82872ed05e9122d3d47d684797c4f20/packages/grafana-runtime/src/utils/DataSourceWithBackend.ts#L74

In the case of alerting, it is the DashAlertExtractor:
https://github.com/grafana/grafana/blob/415638cddadfc0cf930b62568596a0acdd4cab35/pkg/services/alerting/extractor.go#L33

So I imagine this means the expected behavior of users is that they can create a new datasource, rename the old, and rename the new one to be the old name, and then things would work if they are basically the same.  I think for importing dashboards, we now have a way to pick datasources.

That being said, since our datasource cache is build on Id, I think we should save the Id and use that. Our current ds cache is based on ID, and I don't think we have a different one based name.

pkg/services/ngalert/eval/eval.go
```diff
@@ -99,93 +67,12 @@ func (c Condition) IsValid() bool {
...
-				if query.Get("maxDataPoints").MustString() == "" { // GEL requires maxDataPoints inside the query JSON
-					query.Set("maxDataPoints", 100)
-				}
-
-				// intervalMS is calculated by the frontend
-				// should we do something similar?
-				if query.Get("intervalMs").MustString() == "" { // GEL requires intervalMs inside the query JSON
-					query.Set("intervalMs", 1000)
-				}
+// AlertExecCtx is the context provided for executing an alert condition.§
```
kylebrandt, line 70:
```suggestion
// AlertExecCtx is the context provided for executing an alert condition.
```

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,188 @@
...
+		blob, err := json.Marshal(model)
+		if err != nil {
+			return nil, fmt.Errorf("unable to marshal query model %w", err)
+		}
+		rm := json.RawMessage{}
+		err = rm.UnmarshalJSON(blob)
+		if err != nil {
+			return nil, fmt.Errorf("unable to unmarshal query model %w", err)
+		}
+		query.JSON = rm
```
kylebrandt, line 183:
```suggestion
		if query.JSON, err = json.Marshal(model); err != nil {
			return nil, fmt.Errorf("unable to marshal query model %w", err)
		}
```

papagian: It was too early in the morning and didn't have a lot of coffee when I wrote this!

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,181 @@
...
+func (ng *AlertNG) LoadAlertCondition(alertDefinitionID int64, signedInUser *models.SignedInUser, skipCache bool) (*eval.Condition, error) {
+	getAlertDefinitionByIDQuery := GetAlertDefinitionByIDQuery{ID: alertDefinitionID}
+	if err := bus.Dispatch(&getAlertDefinitionByIDQuery); err != nil {
+		return nil, err
+	}
+	alertDefinition := getAlertDefinitionByIDQuery.Result
+
+	condition := eval.Condition{RefID: alertDefinition.Condition}
+	var ds *models.DataSource
+	for _, query := range alertDefinition.Data {
```
kylebrandt, line 91: I am thinking maybe we should move the fetching of the datasource id and the model modification to before saving? 

This way:
 - data source id is the source of truth (this maybe become UID in the future, but I do not think we want to worry about a new dscache etc right now)
 - Alert must have valid datasources / model to be saved.

(Also maybe not include orgid in the saving but keep it where it is, need to see what is going on there (Thinking if we send a QueryDataRequest, the orgId should be in that plugincontext for that, and GEL should get it from there).

Or maybe extract the validation parts so it can be revalidated before running, but not modified (not sure if that is needed).

papagian: > I am thinking maybe we should move the fetching of the datasource id and the model modification to before saving?

Yes, that makes more sense!

kylebrandt:
So perhaps more like:

```go
// AlertDefinition is the saved model for alert definitions in Alerting NG.
type AlertDefinition struct {
	ID    int64
	OrgID int64
	Name  string

	Condition             string
	QueriesAndExpressions []AlertQuery
}

// AlertQuery is the saved model for each query or expression
// in alert defintion.
type AlertQuery struct {
	RefID string

	RelativeTimeRange RelativeTimeRange

	DatasourceID int64

	// Custom query properties
	Model json.RawMessage
}

// GetEvalCondition returns a condition suitable for execution
// based on the saved AlertQuery
func (aq *AlertQuery) GetEvalCondition(now time.Time) *eval.Condition {
	// Copy Properties backend.DataQuery
	// Turn RelativeTimeRange into a time Range
	// Copy the AlertQuery's Model, update model in DataQuery as needed
	return nil
}

// RelativeTimeRange is the per query start and end time
// for requests.
type RelativeTimeRange struct {
	From time.Duration
	To   time.Duration
}
```

- I did not making a clear enough distinction between saving and runtime. 
- `backend.DataQuery` is designed for runtime execution. Also need account for saving, so `AlertDefintion` and `AlertQuery` are the Go representations of the saved model
-  `AlertQuery` I think may get its own table? Saving datasource ID, and relative time ranges.
- The model modifications are done at runtime, not saved (but properties we need can be validated before saving (and probably after to be defensive)).

pkg/services/ngalert/alert_definition.go
```diff
@@ -0,0 +1,71 @@
...
+		maxDatapoints, err := aq.getMaxDatapoints()
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve maxDatapoints from the model %w", err)
+		}
+		condition.QueriesAndExpressions = append(condition.QueriesAndExpressions, backend.DataQuery{
+			JSON:          model,
+			Interval:      time.Duration(intervalMs) * time.Millisecond,
+			RefID:         aq.RefID,
+			MaxDataPoints: maxDatapoints,
+			QueryType:     "",
```
kylebrandt, line 49: important we set this one for certain datasources I think

pkg/services/ngalert/alert_query.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (d *duration) UnmarshalJSON(b []byte) error {
+	var v interface{}
+	if err := json.Unmarshal(b, &v); err != nil {
+		return err
+	}
+	switch value := v.(type) {
+	case float64:
+		*d = duration(time.Duration(value))
```
kylebrandt, line 29: I think this would make it nanoseconds ... why would we have a string or float?

pkg/services/ngalert/alert_definition.go
papagian: I want to ask about this: this should be an additional field in AlertQuery or will be retrieved from the JSON data?

kylebrandt (approved):

pkg/services/ngalert/alert_query.go
papagian: I will change it to accept only float64 (seconds) for now and we can reconsider changing it if a string representation will be more convenient in the future.

pkg/services/ngalert/middleware.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func validateOrgAlertDefinition(c *models.ReqContext) {
+	id := c.ParamsInt64(":alertDefinitionId")
+	query := GetAlertDefinitionByIDQuery{ID: id}
+
+	if err := bus.Dispatch(&query); err != nil {
+		c.JsonApiErr(404, "Alert definition not found", nil)
+		return
+	}
```
aknuds1, line 15: Can we replace this bus usage with a method call?

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	"time"
+
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/grafana/grafana/pkg/services/ngalert/eval"
+)
+
+// AlertDefinition is the model for alert definitions in Alerting NG.
+type AlertDefinition struct {
+	Id        int64
+	OrgId     int64
```
aknuds1, line 14: Initialisms should have consistent case in Go.
```suggestion
	ID        int64
	OrgID     int64
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	OrgID        int64                `json:"-"`
+	Condition    Condition            `json:"condition"`
+	SignedInUser *models.SignedInUser `json:"-"`
+	SkipCache    bool                 `json:"-"`
+
+	Result *AlertDefinition
+}
+
+// IsValid validates a SaveAlertDefinitionCommand.
+// Returns always true.
```
aknuds1, line 59:
```suggestion
// Always returns true.
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	Condition    Condition            `json:"condition"`
+	SignedInUser *models.SignedInUser `json:"-"`
+	SkipCache    bool                 `json:"-"`
+
+	RowsAffected int64
+	Result       *AlertDefinition
+}
+
+// IsValid validates a UpdateAlertDefinitionCommand.
+// Returns always true.
```
aknuds1, line 78:
```suggestion
// IsValid validates an UpdateAlertDefinitionCommand.
// Always returns true.
```

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+	}
+	// create table
+	mg.AddMigration("create alert_definition table", migrator.NewAddTableMigration(alertDefinition))
+
+	// create indices
+	mg.AddMigration("add index alert_definition org_id", migrator.NewAddIndexMigration(alertDefinition, alertDefinition.Indices[0]))
+}
+
+// LoadAlertCondition returns a Condition object for the given alertDefintionId.
```
aknuds1, line 79:
```suggestion
// LoadAlertCondition returns a Condition object for the given alertDefintionID.
```

pkg/services/ngalert/validator.go
```diff
@@ -0,0 +1,32 @@
+package ngalert
+
+import (
+	"fmt"
+
+	"github.com/grafana/grafana/pkg/models"
+)
+
+// validate validates that the alert definition contains at least one alert query
```
aknuds1, line 9:
```suggestion
// validateAlertDefinition validates that the alert definition contains at least one alert query
```

pkg/services/ngalert/validator.go
```diff
@@ -0,0 +1,32 @@
...
+// and that alert queries refer to existing datasources.
+func (ng *AlertNG) validateAlertDefinition(alertDefinition *AlertDefinition, signedInUser *models.SignedInUser, skipCache bool) error {
+	if len(alertDefinition.Data) == 0 {
+		return fmt.Errorf("no queries and expressions are found")
```
aknuds1, line 13:
```suggestion
		return fmt.Errorf("no queries or expressions are found")
```

aknuds1 (commented):
Looks mostly good to me, but please see my suggestions/questions.

pkg/services/ngalert/models.go
papagian: This is a model so I'm afraid I cannot change it because of the go-xorm naming convention.

I think I've tried also changing the `OrgId` -> `OrgID` but then it was looking for column `org_i_d`.

aknuds1: > This is a model so I'm afraid I cannot change it because of the go-xorm naming convention.

@papagian Oh yeah, it's no biggie. I think there's a tag directive you can provide though, to instruct xorm wrt. the translated name.

aknuds1 (changes requested):
Looks better to me! Let's see if we can drop the bus usages though.

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/grafana/grafana/pkg/registry"
+	"github.com/grafana/grafana/pkg/services/datasources"
+	"github.com/grafana/grafana/pkg/services/sqlstore"
+	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
+	"github.com/grafana/grafana/pkg/setting"
+)
+
+// AlertNG is the service for evaluating the condition of an alert definition.
+type AlertNG struct {
+	Bus             bus.Bus                  `inject:""`
```
aknuds1, line 21: We can drop this now.

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	Id        int64
+	OrgId     int64
+	Name      string
+	Condition string
+	Data      []AlertQuery
+}
+
+var (
+	// ErrAlertDefinitionNotFound is an error for an unknown alert definition.
+	ErrAlertDefinitionNotFound = fmt.Errorf("could not find alert definition")
```
aknuds1, line 22: This doesn't need exporting.
```suggestion
	// errAlertDefinitionNotFound is an error for an unknown alert definition.
	errAlertDefinitionNotFound = fmt.Errorf("could not find alert definition")
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// GetAlertDefinitionByIDQuery is the query for retrieving/deleting an alert definition by ID.
+type GetAlertDefinitionByIDQuery struct {
```
aknuds1, line 26: This type can be dropped. If you want, I can push my changes where it's no longer in use.

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	ID    int64
+	OrgID int64
+
+	Result *AlertDefinition
+}
+
+type DeleteAlertDefinitionByIDQuery struct {
```
aknuds1, line 33: This type can be dropped. If you want, I can push my changes where it's no longer in use.

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	ID    int64
+	OrgID int64
+
+	RowsAffected int64
+}
+
+// Condition is the structure used by storing/updating alert definition commmands
+type Condition struct {
```
aknuds1, line 41: This doesn't need exporting.
```suggestion
// condition is used for storing/updating alert definitions.
type condition struct {
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+// SaveAlertDefinitionCommand is the query for saving a new alert definition.
+type SaveAlertDefinitionCommand struct {
+	Name         string               `json:"name"`
+	OrgID        int64                `json:"-"`
+	Condition    Condition            `json:"condition"`
+	SignedInUser *models.SignedInUser `json:"-"`
+	SkipCache    bool                 `json:"-"`
+
+	Result *AlertDefinition
+}
```
aknuds1, line 56: saveAlertDefinitionCommand can be trimmed down to only provide params to the HTTP endpoint:
```suggestion
// saveAlertDefinitionCommand contains parameters for saving a new alert definition.
type saveAlertDefinitionCommand struct {
	Name         string               `json:"name"`
	Condition    Condition            `json:"condition"`
}
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+type UpdateAlertDefinitionCommand struct {
+	ID           int64                `json:"-"`
+	Name         string               `json:"name"`
+	OrgID        int64                `json:"-"`
+	Condition    Condition            `json:"condition"`
+	SignedInUser *models.SignedInUser `json:"-"`
+	SkipCache    bool                 `json:"-"`
+
+	RowsAffected int64
+	Result       *AlertDefinition
```
aknuds1, line 74: This can be trimmed down to only provide parameters to the HTTP endpoint:
```suggestion
// UpdateAlertDefinitionCommand contains parameters for updating an existing alert definition.
type UpdateAlertDefinitionCommand struct {
	Name         string               `json:"name"`
	Condition    Condition            `json:"condition"`
```

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// IsValid validates an UpdateAlertDefinitionCommand.
+// Always returns true.
+func (cmd *UpdateAlertDefinitionCommand) IsValid() bool {
+	return true
+}
+
+type EvalAlertConditionCommand struct {
```
aknuds1, line 83: This type can be removed.

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+	Condition eval.Condition `json:"condition"`
+	Now       time.Time      `json:"now"`
+}
+
+type ListAlertDefinitionsCommand struct {
```
aknuds1, line 88: This type can be dropped.

pkg/services/ngalert/ngalert.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	// create table
+	mg.AddMigration("create alert_definition table", migrator.NewAddTableMigration(alertDefinition))
+
+	// create indices
+	mg.AddMigration("add index alert_definition org_id", migrator.NewAddIndexMigration(alertDefinition, alertDefinition.Indices[0]))
+}
+
+// LoadAlertCondition returns a Condition object for the given alertDefinitionID.
+func (ng *AlertNG) LoadAlertCondition(alertDefinitionID int64, signedInUser *models.SignedInUser, skipCache bool) (*eval.Condition, error) {
```
aknuds1, line 79: This doesn't need to be public:
```suggestion
// loadAlertCondition returns a Condition object for the given alertDefinitionID.
func (ng *AlertNG) loadAlertCondition(alertDefinitionID int64, signedInUser *models.SignedInUser, skipCache bool) (*eval.Condition, error) {
```

pkg/services/ngalert/alert_definition.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"time"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana/pkg/services/ngalert/eval"
+)
+
+// GetEvalCondition returns a condition suitable for execution
+// based on the saved AlertDefinition.
+func (alertDefinition *AlertDefinition) GetEvalCondition(now time.Time) (*eval.Condition, error) {
```
aknuds1, line 13: Doesn't have to be public:
```suggestion
// getEvalCondition returns a condition suitable for execution
// based on the saved AlertDefinition.
func (alertDefinition *AlertDefinition) getEvalCondition(now time.Time) (*eval.Condition, error) {
```

pkg/services/ngalert/alert_query.go
```diff
@@ -0,0 +1,250 @@
...
+func (rtr *RelativeTimeRange) IsValid() bool {
+	return rtr.From > rtr.To
+}
+
+func (rtr *RelativeTimeRange) toTimeRange(now time.Time) backend.TimeRange {
+	return backend.TimeRange{
+		From: now.Add(-time.Duration(rtr.From)),
+		To:   now.Add(-time.Duration(rtr.To)),
+	}
+}
```
aknuds1, line 52: Doesn't need to be public:
```suggestion
// relativeTimeRange is the per query start and end time
// for requests.
type relativeTimeRange struct {
	From duration
	To   duration
}

// IsValid checks that From duration is greater than To duration.
func (rtr *relativeTimeRange) IsValid() bool {
	return rtr.From > rtr.To
}

func (rtr *relativeTimeRange) toTimeRange(now time.Time) backend.TimeRange {
	return backend.TimeRange{
		From: now.Add(-time.Duration(rtr.From)),
		To:   now.Add(-time.Duration(rtr.To)),
	}
}
```

pkg/services/ngalert/alert_query.go
```diff
@@ -0,0 +1,250 @@
...
+
+// AlertQuery represents a single query associated with an alert definition.
+type AlertQuery struct {
```
aknuds1, line 55: Doesn't have to be public:
```suggestion
// alertQuery represents a single query associated with an alert definition.
type alertQuery struct {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+		alertDefinitions.Post("/eval", middleware.ReqSignedIn, binding.Bind(EvalAlertConditionCommand{}), api.Wrap(ng.ConditionEval))
+		alertDefinitions.Get("/:alertDefinitionId", validateOrgAlertDefinition, api.Wrap(ng.GetAlertDefinitionEndpoint))
+		alertDefinitions.Delete("/:alertDefinitionId", validateOrgAlertDefinition, api.Wrap(ng.DeleteAlertDefinitionEndpoint))
+		alertDefinitions.Post("/", middleware.ReqSignedIn, binding.Bind(SaveAlertDefinitionCommand{}), api.Wrap(ng.CreateAlertDefinitionEndpoint))
+		alertDefinitions.Put("/:alertDefinitionId", validateOrgAlertDefinition, binding.Bind(UpdateAlertDefinitionCommand{}), api.Wrap(ng.UpdateAlertDefinitionEndpoint))
+	})
+}
+
+// ConditionEval handles POST /api/alert-definitions/eval.
+func (ng *AlertNG) ConditionEval(c *models.ReqContext, dto EvalAlertConditionCommand) api.Response {
```
aknuds1, line 32: Doesn't have to be public:
```suggestion
// conditionEval handles POST /api/alert-definitions/eval.
func (ng *AlertNG) conditionEval(c *models.ReqContext, dto EvalAlertConditionCommand) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+		return api.Error(400, "Failed to encode result dataframes", err)
+	}
+
+	return api.JSON(200, util.DynMap{
+		"instances": instances,
+	})
+}
+
+// §AlertDefinitionEval handles GET /api/alert-definitions/eval/:dashboardId/:panelId/:refId".
+func (ng *AlertNG) AlertDefinitionEval(c *models.ReqContext) api.Response {
```
aknuds1, line 71: Doesn't have to be public:
```suggestion
// alertDefinitionEval handles GET /api/alert-definitions/eval/:dashboardId/:panelId/:refId".
func (ng *AlertNG) alertDefinitionEval(c *models.ReqContext) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+		return api.Error(400, "Failed to encode result dataframes", err)
+	}
+
+	return api.JSON(200, util.DynMap{
+		"instances": instances,
+	})
+}
+
+// GetAlertDefinitionEndpoint handles GET /api/alert-definitions/:alertDefinitionId.
+func (ng *AlertNG) GetAlertDefinitionEndpoint(c *models.ReqContext) api.Response {
```
aknuds1, line 118: Doesn't have to be public:
```suggestion
// getAlertDefinitionEndpoint handles GET /api/alert-definitions/:alertDefinitionId.
func (ng *AlertNG) getAlertDefinitionEndpoint(c *models.ReqContext) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if err := ng.Bus.Dispatch(&query); err != nil {
+		return api.Error(500, "Failed to get alert definition", err)
+	}
+
+	return api.JSON(200, &query.Result)
+}
+
+// DeleteAlertDefinitionEndpoint handles DELETE /api/alert-definitions/:alertDefinitionId.
+func (ng *AlertNG) DeleteAlertDefinitionEndpoint(c *models.ReqContext) api.Response {
```
aknuds1, line 133: Doesn't have to be public:
```suggestion
// deleteAlertDefinitionEndpoint handles DELETE /api/alert-definitions/:alertDefinitionId.
func (ng *AlertNG) deleteAlertDefinitionEndpoint(c *models.ReqContext) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if err := ng.DeleteAlertDefinitionByID(&query); err != nil {
+		return api.Error(500, "Failed to delete alert definition", err)
+	}
+
+	return api.JSON(200, util.DynMap{"affectedRows": query.RowsAffected})
+}
+
+// UpdateAlertDefinitionEndpoint handles PUT /api/alert-definitions/:alertDefinitionId.
+func (ng *AlertNG) UpdateAlertDefinitionEndpoint(c *models.ReqContext, cmd UpdateAlertDefinitionCommand) api.Response {
```
aknuds1, line 149: Doesn't have to be public:
```suggestion
// updateAlertDefinitionEndpoint handles PUT /api/alert-definitions/:alertDefinitionId.
func (ng *AlertNG) updateAlertDefinitionEndpoint(c *models.ReqContext, cmd UpdateAlertDefinitionCommand) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if err := ng.UpdateAlertDefinition(&cmd); err != nil {
+		return api.Error(500, "Failed to update alert definition", err)
+	}
+
+	return api.JSON(200, util.DynMap{"affectedRows": cmd.RowsAffected, "id": cmd.Result.Id})
+}
+
+// CreateAlertDefinitionEndpoint handles POST /api/alert-definitions.
+func (ng *AlertNG) CreateAlertDefinitionEndpoint(c *models.ReqContext, cmd SaveAlertDefinitionCommand) api.Response {
```
aknuds1, line 162: Doesn't have to be public:
```suggestion
// createAlertDefinitionEndpoint handles POST /api/alert-definitions.
func (ng *AlertNG) createAlertDefinitionEndpoint(c *models.ReqContext, cmd SaveAlertDefinitionCommand) api.Response {
```

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if err := ng.SaveAlertDefinition(&cmd); err != nil {
+		return api.Error(500, "Failed to create alert definition", err)
+	}
+
+	return api.JSON(200, util.DynMap{"id": cmd.Result.Id})
+}
+
+// ListAlertDefinitions handles GET /api/alert-definitions.
+func (ng *AlertNG) ListAlertDefinitions(c *models.ReqContext) api.Response {
```
aknuds1, line 175: Doesn't have to be public:
```suggestion
// listAlertDefinitions handles GET /api/alert-definitions.
func (ng *AlertNG) listAlertDefinitions(c *models.ReqContext) api.Response {
```

pkg/services/ngalert/database.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	return &alertDefinition, nil
+}
+
+// DeleteAlertDefinitionByID is a handler for deleting an alert definition.
+// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
+func (ng *AlertNG) DeleteAlertDefinitionByID(query *DeleteAlertDefinitionByIDQuery) error {
```
aknuds1, line 23: Doesn't have to be public:
```suggestion
// deleteAlertDefinitionByID is a handler for deleting an alert definition.
// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
func (ng *AlertNG) deleteAlertDefinitionByID(query *DeleteAlertDefinitionByIDQuery) error {
```

pkg/services/ngalert/database.go
```diff
@@ -0,0 +1,120 @@
...
+			return err
+		}
+		query.RowsAffected = rowsAffected
+		return nil
+	})
+}
+
+// GetAlertDefinitionByID is a handler for retrieving an alert definition from that database by its ID.
+// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
+func (ng *AlertNG) GetAlertDefinitionByID(query *GetAlertDefinitionByIDQuery) error {
```
aknuds1, line 41: Doesn't have to be public:
```suggestion
// getAlertDefinitionByID is a handler for retrieving an alert definition from that database by its ID.
// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
func (ng *AlertNG) getAlertDefinitionByID(query *GetAlertDefinitionByIDQuery) error {
```

pkg/services/ngalert/database.go
```diff
@@ -0,0 +1,120 @@
...
+		if err != nil {
+			return err
+		}
+		query.Result = alertDefinition
+		return nil
+	})
+}
+
+// SaveAlertDefinition is a handler for saving a new alert definition.
+func (ng *AlertNG) SaveAlertDefinition(cmd *SaveAlertDefinitionCommand) error {
```
aknuds1, line 53: Doesn't have to be public:
```suggestion
// saveAlertDefinition is a handler for saving a new alert definition.
func (ng *AlertNG) saveAlertDefinition(cmd *SaveAlertDefinitionCommand) error {
```

pkg/services/ngalert/database.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		cmd.Result = alertDefinition
+		return nil
+	})
+}
+
+// UpdateAlertDefinition is a handler for updating an existing alert definition.
+// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
+func (ng *AlertNG) UpdateAlertDefinition(cmd *UpdateAlertDefinitionCommand) error {
```
aknuds1, line 81: Doesn't have to be public:
```suggestion
// updateAlertDefinition is a handler for updating an existing alert definition.
// It returns models.ErrAlertDefinitionNotFound if no alert definition is found for the provided ID.
func (ng *AlertNG) updateAlertDefinition(cmd *UpdateAlertDefinitionCommand) error {
```

pkg/services/ngalert/database.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		cmd.Result = alertDefinition
+		cmd.RowsAffected = affectedRows
+		return nil
+	})
+}
+
+// GetAlertDefinitions is a handler for retrieving alert definitions of specific organisation.
+func (ng *AlertNG) GetAlertDefinitions(cmd *ListAlertDefinitionsCommand) error {
```
aknuds1, line 109: Doesn't have to be public:
```suggestion
// getAlertDefinitions is a handler for retrieving alert definitions of specific organisation.
func (ng *AlertNG) getAlertDefinitions(cmd *ListAlertDefinitionsCommand) error {
```

pkg/services/ngalert/eval/eval.go
```diff
@@ -0,0 +1,184 @@
...
+// of the query or expression that will be evaluated.
+type Condition struct {
+	RefID string `json:"refId"`
+
+	QueriesAndExpressions []backend.DataQuery `json:"queriesAndExpressions"`
+}
+
+// ExecutionResults contains the unevaluated results from executing
+// a condition.
+type ExecutionResults struct {
```
aknuds1, line 27: Doesn't have to be public:
```suggestion
// executionResults contains the unevaluated results from executing
// a condition.
type executionResults struct {
```

pkg/services/ngalert/eval/eval.go
```diff
@@ -0,0 +1,184 @@
...
+	AlertDefinitionID int64
+
+	Error error
+
+	Results data.Frames
+}
+
+// Results is a slice of evaluated alert instances states.
+type Results []Result
```
aknuds1, line 36:
```suggestion
// results is a slice of evaluated alert instances states.
type results []Result
```

pkg/services/ngalert/eval/eval.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Result contains the evaluated state of an alert instance
+// identified by its labels.
+type Result struct {
+	Instance data.Labels
+	State    State // Enum
+}
+
+// State is an enum of the evaluation state for an alert instance.
```
aknuds1, line 45: Don''t have to be public:
```suggestion
// result contains the evaluated state of an alert instance
// identified by its labels.
type result struct {
	Instance data.Labels
	State    state // Enum
}

// state is an enum of the evaluation state for an alert instance.
```

pkg/services/ngalert/eval/eval.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// Alerting is the eval state for an alert instance condition
+	// that evaluated to false.
+	Alerting
+)
+
+func (s State) String() string {
+	return [...]string{"Normal", "Alerting"}[s]
+}
+
```
aknuds1, line 61:
```suggestion
	normal state = iota

	// alerting is the eval state for an alert instance condition
	// that evaluated to false.
	alerting
)

func (s state) String() string {
	return [...]string{"Normal", "Alerting"}[s]
}

```

aknuds1 (commented):
There are some bus usages left, and there are many symbols that can be unexported. I can push the changes if you want?

pkg/services/ngalert/eval/alert_query.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	// override model
+	model, err := aq.getModel()
+	if err != nil {
+		return err
+	}
+	aq.Model = model
+
+	if ok := aq.RelativeTimeRange.isValid(); !ok {
```
kylebrandt, line 285: Currently backend expression nodes don't have a time range. They might, but if they do, it might be as parameters to whatever is being called. So for now a valid time range is not needed expected for the expression queries, just datasource ones.

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+func (ng *AlertNG) updateAlertDefinitionEndpoint(c *models.ReqContext, cmd updateAlertDefinitionCommand) api.Response {
+	cmd.ID = c.ParamsInt64(":alertDefinitionId")
+	cmd.SignedInUser = c.SignedInUser
+	cmd.SkipCache = c.SkipCache
+
+	if err := ng.updateAlertDefinition(&cmd); err != nil {
+		return api.Error(500, "Failed to update alert definition", err)
+	}
+
+	return api.JSON(200, util.DynMap{"affectedRows": cmd.RowsAffected, "id": cmd.Result.Id})
```
kylebrandt, line 158: It seems for me when updating an alert the Id is coming back as zero (for an alert that is not id 0).

pkg/services/ngalert/eval/eval.go
```diff
@@ -0,0 +1,196 @@
...
+func EvaluateExecutionResult(results *ExecutionResults) (Results, error) {
+	evalResults := make([]result, 0)
+	labels := make(map[string]bool)
+	for _, f := range results.Results {
+		rowLen, err := f.RowLen()
+		if err != nil {
+			return nil, fmt.Errorf("unable to get frame row length: %w", err)
+		}
+		if rowLen > 1 {
+			return nil, fmt.Errorf("invalid frame %q: row length: %v", f.Name, rowLen)
```
kylebrandt, line 154: I hit this one pointing at a time series queries instead of the reduction query:

```
{
  "error": "invalid frame \"\": row length: 9",
  "message": "Failed to evaluate results"
}
```

Took me a bit to figure out where/why.  I think maybe we need to wrap this with an generic error explaining that the results returned were not in a valid format. Also maybe referencing the condition/refId might help. 

With this specific one, we should probably also state the expected length in the details of the error (0 or 1)

pkg/services/ngalert/models.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type deleteAlertDefinitionByIDQuery struct {
+	ID    int64
+	OrgID int64
+
+	RowsAffected int64
+}
+
+// Condition is the structure used by storing/updating alert definition commmands
```
aknuds1, line 40:
```suggestion
// condition is the structure used by storing/updating alert definition commmands
```

aknuds1 (approved):
Looks good AFAICT, with some nits.

pkg/services/ngalert/alert_definition.go
```diff
@@ -0,0 +1,13 @@
+package ngalert
+
+// preSave sets datasource and loads the updated model for each alert query.
+func (alertDefinition *AlertDefinition) preSave() error {
+	for i, q := range alertDefinition.Data {
+		err := q.PreSave(alertDefinition.OrgId)
+		if err != nil {
+			return err
```
kylebrandt, line 8: I think we should wrap the error with the RefID of the alert query in this loop so when there are problems you can see which query has the issue. (or perhaps i and queryRefId)

kylebrandt:
Couple things with error handling and extra timerange, but https://gist.github.com/kylebrandt/58b18d783b52d369bfa0c6cadd393a7c seems to work, which is a mix of 2 ADX queries and backend expressions - and produces multiple instances.

pkg/services/ngalert/api.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/grafana/grafana/pkg/util"
+)
+
+func (ng *AlertNG) registerAPIEndpoints() {
+	ng.RouteRegister.Group("/api/alert-definitions", func(alertDefinitions routing.RouteRegister) {
+		alertDefinitions.Get("", middleware.ReqSignedIn, api.Wrap(ng.listAlertDefinitions))
+		alertDefinitions.Get("/eval/:alertDefinitionId", ng.validateOrgAlertDefinition, api.Wrap(ng.alertDefinitionEval))
+		alertDefinitions.Post("/eval", middleware.ReqSignedIn, binding.Bind(evalAlertConditionCommand{}), api.Wrap(ng.conditionEval))
+		alertDefinitions.Get("/:alertDefinitionId", ng.validateOrgAlertDefinition, api.Wrap(ng.getAlertDefinitionEndpoint))
+		alertDefinitions.Delete("/:alertDefinitionId", ng.validateOrgAlertDefinition, api.Wrap(ng.deleteAlertDefinitionEndpoint))
```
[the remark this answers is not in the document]
papagian: good point; I think it's redundant because if not signed-in then `c.OrgId` is zero so this condition becomes false and `403 Forbidden` is returned:
https://github.com/grafana/grafana/blob/b59045aa21f1aa68e829d62a48814b0348de8a03/pkg/services/ngalert/middleware.go#L16

pkg/services/ngalert/eval/alert_query_test.go
```diff
@@ -0,0 +1,296 @@
...
+				require.Equal(t, tc.expectedMaxPoints, maxDataPoints)
+			})
+
+			t.Run("can update model intervalMs (if missing)", func(t *testing.T) {
+				intervalMS, err := tc.alertQuery.getIntervalMS()
+				require.NoError(t, err)
+				require.Equal(t, intervalMS, tc.expectedIntervalMS)
+			})
+
+			t.Run("can get the updated model with the default properties (if missing)", func(t *testing.T) {
```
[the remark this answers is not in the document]
papagian: Given [this](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) I think that is fine to have hierarchical tests and on top of it one can run only a subset of the sub-tests like:
`go test -v ./pkg/services/ngalert/eval -run AlertQuery//can_recognize_if_it\'s_an_expression`

PR #28379, opened 2020-10-19
Component/NewsPanel: Add rel="noopener" to NewsPanel links

kaydelaney:
Adds `rel = "noopener"` to `NewsPanel` links as suggested by Lighthouse [best practices](https://web.dev/external-anchors-use-rel-noopener/).


dprokop (approved):
👏

PR #28381, opened 2020-10-19
Explore: Fix date formatting in url for trace logs link

aocenas:
Need to format the string in UTC. Also changes the time range buffers for the logs search so it's startTime - 1h -> endTime + 5s


joe-elliott (approved):
brilliant!

kylebrandt:

PR #28387, opened 2020-10-20
Instrumentation: Remove status label from database duration histogram

bergquist:
We have counters for success/error, that should be enough. 


marefr (approved):
LGTM

kylebrandt:

PR #28390, opened 2020-10-20
Plugins: Don't exit on duplicate plugin

### LINKED ISSUE #28324: Plugins: Failed to load multiple versions of plugins
**What happened**:
Upgraded to 7.3.0-beta1 and Grafana failed to start with the following error message:
```
EROR[10-16|09:58:48] Server shutdown                          logger=server reason="Service init failed: failed to scan external plugins directory '/home/marcus/go/src/github.com/grafana/grafana/data/plugins': Plugin with ID \"grafana-worldmap-panel\" already exists"
Service init failed: failed to scan external plugins directory '/home/marcus/go/src/github.com/grafana/grafana/data/plugins': Plugin with ID "grafana-worldmap-panel" already exists
[Bra] 10-16 09:58:48 [ WARN] Fail to execute command: ./bin/grafana-server [-packaging=dev cfg:app_mode=development] - exit status 1
```

Grafana to start without throwing error and exit.

1. Install raintank-worldping-app plugin
1. Install grafana-worldmap-panel plugin
1. Start Grafana

**Environment**:
- Grafana version: 7.3.0-beta1 / master (6002df580f)
- Grafana plugins: raintank-worldping-app, grafana-worldmap-panel

aknuds1:
Don't exit when a duplicate plugin is detected, just warn of the error.

Fixes #28324.


papagian (approved):
LGTM; I've tested it and resolves the issue!

marefr (approved):
I did talk to @daniellee about these special cases. Here's a summary:

[Azure Monitor external plugin](https://grafana.com/grafana/plugins/grafana-azure-monitor-datasource) at first and later moved into core:
 - The reason you always get the core azure plugin is because core plugins are scanned first (before external) when loading plugins so no we've probably never had any specific version logic besides this 
 - The expected outcome would be to get a log message when trying to load the external azure plugin, but should not reject Grafana to start

Worldmap and worldping:
 - If you have installed worldping and worldmap plugins separately the execpted outcome is that the "latest" version should be used. 
 - The bundled worldmap is an old version and the separately installed worldmap panel is in general a higher version. It was noticed as a problem that you had to install the worldmap panel separately before being able to use the worldping app why worldmap panel was bundled into the dist directory of the app. It was hard to keep the worldping app up to date with latest worldmap panel why the special case of allowing an external worldmap panel of later version to be installed and by that override the bundled worldmap panel.
 - However, we don't have any logic in place to make sure we pick the highest version. 
 - Seems like we're just lucky here that the worldmap panel is scanned before the worldping app :) 

Syntethic monitoring app:
 - Seems to have [worldmap as a dependency](https://github.com/grafana/synthetic-monitoring-app/blob/767d746ab43f216f28fa7b86a829cf7b4f238dfe/src/plugin.json#L55-L60), but doesn't bundle it.

Based on above I've verified that with these changes it should work as before:
```
WARN[10-21|11:40:37] Plugin is duplicate                      logger=plugins error="Failed to register plugin: plugin with ID \"grafana-azure-monitor-datasource\" already loaded from \"/home
/marcus/go/src/github.com/grafana/grafana/public/app/plugins/datasource/grafana-azure-monitor-datasource\""                                                                                   
WARN[10-21|11:40:37] Plugin is duplicate                      logger=plugins error="plugin with ID \"grafana-worldmap-panel\" already loaded from \"/home/marcus/go/src/github.com/grafana/gra
fana/data/plugins/grafana-worldmap-panel/dist\""                                                 
WARN[10-21|11:40:37] Some plugins failed to load              logger=plugins errors="[Failed to register plugin: plugin with ID \"grafana-azure-monitor-datasource\" already loaded from \"/ho
me/marcus/go/src/github.com/grafana/grafana/public/app/plugins/datasource/grafana-azure-monitor-datasource\" plugin with ID \"grafana-worldmap-panel\" already loaded from \"/home/marcus/go/s
rc/github.com/grafana/grafana/data/plugins/grafana-worldmap-panel/dist\"]"
```

I was a bit confused at first of the log messages since I thought it "failed to load from referenced path", but it was already loaded from. Nit, maybe room for improvement? 

In general a bit tricky to write unit/integrations tests to cover all of the above - manual testing is still the best, but the test you've added is still good/needed to verify the error handling :+1: 

LGTM great :+1:

PR #28391, opened 2020-10-20
CloudWatch: Fix custom metrics

### LINKED ISSUE #28075: Cloudwatch custom metric results in server error when query metricFindQuery
**What happened**:

Querying aws cloudwatch custom metrics results in a 500er server error when try to get the list of available metric names.

grafanabug

Server log 

```
/grafana.....de ecs/grafana/f05454fc888745d99c4f82f261266df8 t=2020-10-07T09:30:21+0000 lvl=eror msg="Request Completed" logger=context userId=2 orgId=1 uname=markuman method=POST path=/api/tsdb/query status=500 remote_addr=10.103.11.8 time_ms=27 size=34 referer="https://grafana......de/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22AWS%20CloudWatch%22,%7B%22namespace%22:%22fluentd_metrics%22,%22metricName%22:%22%22,%22expression%22:%22%22,%22dimensions%22:%7B%7D,%22region%22:%22default%22,%22id%22:%22%22,%22alias%22:%22%22,%22statistics%22:%5B%22Average%22%5D,%22period%22:%22%22,%22matchExact%22:true%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D"

```

Browser Payload

```json
{"from":"1602059279729","to":"1602062879731","queries":[{"refId":"metricFindQuery","intervalMs":1,"maxDataPoints":1,"datasourceId":2,"type":"metricFindQuery","subtype":"metrics","region":"eu-central-1","namespace":"fluentd_metrics"}]}
```

Browser response 

```json
{"message":"Metric request error"}
```

Autocomplete existings metric names and its dimensions.  
When a dimension is set, the autocomplete works again for its value.

Try to access aws cloudwatch custom metric.

It works in the past.  But I cannot say for sure in which version it failed.  
Out past used versions. I believe 7.0.x it works fine.  
```
7.0.0
7.0.3
7.0.6
7.1.3
```

**Environment**:
- Grafana version: 7.2.0
- OS Grafana is installed on: Official Docker Container
- User OS & Browser: Firefox 81

aknuds1:
Fix querying of CloudWatch custom metrics. I've tested the change, and it works (I can query custom metrics).

Fixes #28075.

### DIFF
--- a/pkg/tsdb/cloudwatch/metric_find_query.go
+++ b/pkg/tsdb/cloudwatch/metric_find_query.go
@@ -39,8 +39,8 @@ type customMetricsCache struct {
 	Cache  []string
 }
 
-var customMetricsMetricsMap = make(map[string]*customMetricsCache)
-var customMetricsDimensionsMap = make(map[string]*customMetricsCache)
+var customMetricsMetricsMap = make(map[string]map[string]map[string]*customMetricsCache)
+var customMetricsDimensionsMap = make(map[string]map[string]map[string]*customMetricsCache)
 var metricsMap = map[string][]string{
 	"AWS/ACMPrivateCA":            {"CRLGenerated", "Failure", "MisconfiguredCRLBucket", "Success", "Time"},
 	"AWS/AmazonMQ":                {"BurstBalance", "ConsumerCount", "CpuCreditBalance", "CpuUtilization", "CurrentConnectionsCount", "DequeueCount", "DispatchCount", "EnqueueCount", "EnqueueTime", "EstablishedConnectionsCount", "ExpiredCount", "HeapUsage", "InactiveDurableTopicSubscribersCount", "InFlightCount", "JobSchedulerStorePercentUsage", "JournalFilesForFastRecovery", "JournalFilesForFullRecovery", "MemoryUsage", "NetworkIn", "NetworkOut", "OpenTransactionCount", "ProducerCount", "QueueSize", "ReceiveCount", "StorePercentUsage", "TempPercentUsage", "TotalConsumerCount", "TotalDequeueCount", "TotalEnqueueCount", "TotalMessageCount", "TotalProducerCount", "VolumeReadOps", "VolumeWriteOps"},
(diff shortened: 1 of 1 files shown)


pkg/tsdb/cloudwatch/metric_find_query.go
```diff
@@ -39,8 +39,8 @@ type customMetricsCache struct {
 	Cache  []string
 }
 
-var customMetricsMetricsMap = make(map[string]map[string]map[string]*customMetricsCache)
-var customMetricsDimensionsMap = make(map[string]map[string]map[string]*customMetricsCache)
+var customMetricsMetricsMap = make(map[string]*customMetricsCache)
```
papagian, line 42: Since this is a hot fix and, as we agreed we want to keep it simple, I suggest to not introduce this change here; it's not necessary for fixing the custom metrics.

marefr (changes requested):
Please do refactoring/code simplification in a separate PR. I have a hard time see what changes are actually related to the bug fix

papagian:
> Please do refactoring/code simplification in a separate PR. I have a hard time see what changes are actually related to the bug fix

For reference the original problem (in master) was [here](https://github.com/grafana/grafana/blob/0e17a15fbd2cabe65087207b347d5a21f95206bd/pkg/tsdb/cloudwatch/metric_find_query.go#L713); the `dsInfo.Namespace` is empty; `getDSInfo()` [does not](https://github.com/grafana/grafana/blob/0e17a15fbd2cabe65087207b347d5a21f95206bd/pkg/tsdb/cloudwatch/cloudwatch.go#L430) set it.

papagian (approved):
LGTM

marefr (commented):
Changes looks good to me. @papagian you have more context so relying on you. 

To me it looks like the Namespace field of `datasourceInfo` now can be removed since no code is using it any longer - is that correct? If so, I would suggest removing it since the reason it was added in first place is no longer there.

aknuds1:
> Changes looks good to me. @papagian you have more context so relying on you.
> 
> To me it looks like the Namespace field of `datasourceInfo` now can be removed since no code is using it any longer - is that correct? If so, I would suggest removing it since the reason it was added in first place is no longer there.

@marefr I was also wondering if it could be removed. Haven't checked though, can try cleaning it in another PR.

papagian:
> > Changes looks good to me. @papagian you have more context so relying on you.
> > To me it looks like the Namespace field of `datasourceInfo` now can be removed since no code is using it any longer - is that correct? If so, I would suggest removing it since the reason it was added in first place is no longer there.
> 
> @marefr I was also wondering if it could be removed. Haven't checked though, can try cleaning it in another PR. There's lots of potential for code scrubbing in CW.

I think that it's safe to remove it and it's good suggestion but also agree that there is a lot of room for improvements so let's remove it in the forthcoming refactoring.

marefr (approved):
Sounds good :+1:

PR #28393, opened 2020-10-20
Explore: Support wide data frames

### LINKED ISSUE #27275: Explore does not Graph wide time series data
Explore to graph [wide time series dataframes](https://grafana.com/docs/grafana/latest/developers/plugins/data-frames/#wide-format). The table panel will Graph a "wide" dataframe, for example with a Frame structure like:

```json
[
  {
    "length": 2,
    "fields": [
      {
        "name": "Time",
        "type": "time",
        "config": {}
      },
      {
        "name": "",
        "type": "number",
        "config": {},
        "labels": {
          "host": "a"
        }
      },
      {
        "name": "",
        "type": "number",
        "config": {},
        "labels": {
          "host": "b"
        }
      }
    ]
  }
]
```

We only see the table:

image

However, the Graph panel will display it:

image

**Why is this needed**:
Azure datasource, and in the future SQL will return time series in a single frame like this.

**Reference**:

Docs Reference: https://grafana.com/docs/grafana/latest/developers/plugins/data-frames/#wide-format

Base64 Frame used in the example with the gdev data source:

```
QVJST1cxAAD/////GAIAABAAAAAAAAoADgAMAAsABAAKAAAAFAAAAAAAAAEDAAoADAAAAAgABAAKAAAACAAAAFAAAAACAAAAKAAAAAQAAAB4/v//CAAAAAwAAAAAAAAAAAAAAAUAAAByZWZJZAAAAJj+//8IAAAADAAAAAAAAAAAAAAABAAAAG5hbWUAAAAAAwAAACQBAACMAAAABAAAAPr+//8UAAAAaAAAAGgAAAAAAAADaAAAAAIAAAAoAAAABAAAAOz+//8IAAAADAAAAAAAAAAAAAAABAAAAG5hbWUAAAAADP///wgAAAAYAAAADAAAAHsiaG9zdCI6ImIifQAAAAAGAAAAbGFiZWxzAAAAAAAACv///wAAAgAAAAAAAAAAAH7///8UAAAAaAAAAGgAAAAAAAADaAAAAAIAAAAoAAAABAAAAHD///8IAAAADAAAAAAAAAAAAAAABAAAAG5hbWUAAAAAkP///wgAAAAYAAAADAAAAHsiaG9zdCI6ImEifQAAAAAGAAAAbGFiZWxzAAAAAAAAjv///wAAAgAAAAAAAAASABgAFAAAABMADAAAAAgABAASAAAAFAAAAEQAAABMAAAAAAAACkwAAAABAAAADAAAAAgADAAIAAQACAAAAAgAAAAQAAAABAAAAFRpbWUAAAAABAAAAG5hbWUAAAAAAAAAAAAABgAIAAYABgAAAAAAAwAEAAAAVGltZQAAAAAAAAAA/////+gAAAAUAAAAAAAAAAwAFgAUABMADAAEAAwAAAAwAAAAAAAAABQAAAAAAAADAwAKABgADAAIAAQACgAAABQAAAB4AAAAAgAAAAAAAAAAAAAABgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAADAAAAAgAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAAkRNqJoqALxYYbbEemIAvFgAAAAAAAPA/AAAAAAAAAEAAAAAAAAAIQAAAAAAAABBAEAAAAAwAFAASAAwACAAEAAwAAAAQAAAALAAAADgAAAAAAAMAAQAAACgCAAAAAAAA8AAAAAAAAAAwAAAAAAAAAAAAAAAAAAAAAAAKAAwAAAAIAAQACgAAAAgAAABQAAAAAgAAACgAAAAEAAAAeP7//wgAAAAMAAAAAAAAAAAAAAAFAAAAcmVmSWQAAACY/v//CAAAAAwAAAAAAAAAAAAAAAQAAABuYW1lAAAAAAMAAAAkAQAAjAAAAAQAAAD6/v//FAAAAGgAAABoAAAAAAAAA2gAAAACAAAAKAAAAAQAAADs/v//CAAAAAwAAAAAAAAAAAAAAAQAAABuYW1lAAAAAAz///8IAAAAGAAAAAwAAAB7Imhvc3QiOiJiIn0AAAAABgAAAGxhYmVscwAAAAAAAAr///8AAAIAAAAAAAAAAAB+////FAAAAGgAAABoAAAAAAAAA2gAAAACAAAAKAAAAAQAAABw////CAAAAAwAAAAAAAAAAAAAAAQAAABuYW1lAAAAAJD///8IAAAAGAAAAAwAAAB7Imhvc3QiOiJhIn0AAAAABgAAAGxhYmVscwAAAAAAAI7///8AAAIAAAAAAAAAEgAYABQAAAATAAwAAAAIAAQAEgAAABQAAABEAAAATAAAAAAAAApMAAAAAQAAAAwAAAAIAAwACAAEAAgAAAAIAAAAEAAAAAQAAABUaW1lAAAAAAQAAABuYW1lAAAAAAAAAAAAAAYACAAGAAYAAAAAAAMABAAAAFRpbWUAAAAAQAIAAEFSUk9XMQ==
```

Go snippet to make a frame an encoded Frame like that (in case the time being far in the past is annoying):

```go
package main

import (
	"encoding/base64"
	"fmt"
	"log"
	"time"

	"github.com/grafana/grafana-plugin-sdk-go/data"
)

func main() {
	frame := data.NewFrame("",
		data.NewField("Time", nil, []time.Time{time.Now().UTC().Add(-time.Minute), time.Now().UTC()}),
		data.NewField("", data.Labels{"host": "a"}, []float64{1, 2}),
		data.NewField("", data.Labels{"host": "b"}, []float64{3, 4}),
	)
	b, err := frame.MarshalArrow()
	if err != nil {
		log.Fatalf("failed to encode frame: %v", err)
	}
	sEnc := base64.StdEncoding.EncodeToString(b)
	fmt.Println(sEnc)
}
```

aocenas:
Fixes: https://github.com/grafana/grafana/issues/27275

Explore did not handle dataFrames with single time field and multiple numeric fields. Instead plugins had to return each time series in separate dataFrame which was an unnecessary complication.

### DIFF
--- a/public/app/features/explore/state/actions.ts
+++ b/public/app/features/explore/state/actions.ts
@@ -1,5 +1,5 @@
 // Libraries
-import { map, throttleTime } from 'rxjs/operators';
+import { map, mergeMap, throttleTime } from 'rxjs/operators';
 import { identity } from 'rxjs';
 import { PayloadAction } from '@reduxjs/toolkit';
 import { DataSourceSrv } from '@grafana/runtime';
@@ -84,7 +84,7 @@ import {
 } from './actionTypes';
 import { getTimeZone } from 'app/features/profile/state/selectors';
 import { getShiftedTimeRange } from 'app/core/utils/timePicker';
-import { updateLocation } from '../../../core/actions';
+import { notifyApp, updateLocation } from '../../../core/actions';
 import { getTimeSrv, TimeSrv } from '../../dashboard/services/TimeSrv';
 import { preProcessPanelData, runRequest } from '../../dashboard/state/runRequest';
 import { DashboardModel, PanelModel } from 'app/features/dashboard/state';
@@ -96,6 +96,7 @@ import {
   decorateWithLogsResult,
   decorateWithTableResult,
 } from '../utils/decorators';
+import { createErrorNotification } from '../../../core/copy/appNotification';
 
 /**
  * Adds a query row after the row with the given index.
@@ -427,6 +428,8 @@ export const runQueries = (exploreId: ExploreId): ThunkResult<void> => {
       queryResponse,
       querySubscription,
       history,
+      refreshInterval,
+      absoluteRange,
     } = exploreItemState;
 
     if (!hasNonEmptyQuery(queries)) {
@@ -473,47 +476,54 @@ export const runQueries = (exploreId: ExploreId): ThunkResult<void> => {
         // actually can see what is happening.
         live ? throttleTime(500) : identity,
         map((data: PanelData) => preProcessPanelData(data, queryResponse)),
-        decorateWithGraphLogsTraceAndTable(getState().explore[exploreId].datasourceInstance),
-        decorateWithGraphResult(),
-        decorateWithTableResult(),
-        decorateWithLogsResult(getState().explore[exploreId])
+        map(decorateWithGraphLogsTraceAndTable),
+        map(decorateWithGraphResult),
+        map(decorateWithLogsResult({ absoluteRange, refreshInterval })),
+        mergeMap(decorateWithTableResult)
       )
-      .subscribe(data => {
-        if (!data.error && firstResponse) {
-          // Side-effect: Saving history in localstorage
-          const nextHistory = updateHistory(history, datasourceId, queries);
-          const nextRichHistory = addToRichHistory(
-            richHistory || [],
-            datasourceId,
-            datasourceName,
-            queries,
-            false,
-            '',
-            ''
-          );
-          dispatch(historyUpdatedAction({ exploreId, history: nextHistory }));
-          dispatch(richHistoryUpdatedAction({ richHistory: nextRichHistory }));
+      .subscribe(
+        data => {
+          if (!data.error && firstResponse) {
+            // Side-effect: Saving history in localstorage
+            const nextHistory = updateHistory(history, datasourceId, queries);
+            const nextRichHistory = addToRichHistory(
+              richHistory || [],
+              datasourceId,
+              datasourceName,
+              queries,
+              false,
+              '',
+              ''
+            );
+            dispatch(historyUpdatedAction({ exploreId, history: nextHistory }));
+            dispatch(richHistoryUpdatedAction({ richHistory: nextRichHistory }));
 
-          // We save queries to the URL here so that only successfully run queries change the URL.
-          dispatch(stateSave());
-        }
-
-        firstResponse = false;
-
-        dispatch(queryStreamUpdatedAction({ exploreId, response: data }));
-
-        // Keep scanning for results if this was the last scanning transaction
-        if (getState().explore[exploreId].scanning) {
-          if (data.state === LoadingState.Done && data.series.length === 0) {
-            const range = getShiftedTimeRange(-1, getState().explore[exploreId].range);
-            dispatch(updateTime({ exploreId, absoluteRange: range }));
-            dispatch(runQueries(exploreId));
-          } else {
-            // We can stop scanning if we have a result
-            dispatch(scanStopAction({ exploreId }));
+            // We save queries to the URL here so that only successfully run queries change the URL.
+            dispatch(stateSave());
           }
+
+          firstResponse = false;
+
+          dispatch(queryStreamUpdatedAction({ exploreId, response: data }));
+
+          // Keep scanning for results if this was the last scanning transaction
+          if (getState().explore[exploreId].scanning) {
+            if (data.state === LoadingState.Done && data.series.length === 0) {
+              const range = getShiftedTimeRange(-1, getState().explore[exploreId].range);
+              dispatch(updateTime({ exploreId, absoluteRange: range }));
+              dispatch(runQueries(exploreId));
+            } else {
+              // We can stop scanning if we have a result
+              dispatch(scanStopAction({ exploreId }));
+            }
+          }
+        },
+        error => {
+          dispatch(notifyApp(createErrorNotification('Query processing error', error)));
+          dispatch(changeLoadingStateAction({ exploreId, loadingState: LoadingState.Error }));
+          console.error(error);
         }
-      });
+      );
 
     dispatch(queryStoreSubscriptionAction({ exploreId, querySubscription: newQuerySub }));
   };
--- a/public/app/features/explore/utils/decorators.test.ts
+++ b/public/app/features/explore/utils/decorators.test.ts
@@ -3,15 +3,12 @@ jest.mock('@grafana/data/src/datetime/formatter', () => ({
   dateTimeFormatTimeAgo: (ts: any) => 'fromNow() jest mocked',
 }));
 
-import { of } from 'rxjs';
 import {
   ArrayVector,
   DataFrame,
   DataQueryRequest,
-  DataSourceApi,
   FieldType,
   LoadingState,
-  observableTester,
   PanelData,
   TimeRange,
   toDataFrame,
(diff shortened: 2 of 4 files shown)


public/app/features/explore/state/actions.ts
```diff
@@ -473,47 +476,54 @@ export const runQueries = (exploreId: ExploreId): ThunkResult<void> => {
...
+            } else {
+              // We can stop scanning if we have a result
+              dispatch(scanStopAction({ exploreId }));
+            }
           }
+        },
+        error => {
+          dispatch(notifyApp(createErrorNotification('Query processing error', error)));
+          dispatch(changeLoadingStateAction({ exploreId, loadingState: LoadingState.Error }));
+          console.error(error);
```
aocenas, line 524: Added error handling here in case something happens during the processing.

public/app/features/explore/utils/decorators.ts
```diff
@@ -1,214 +1,187 @@
...
+        break;
+      default:
+        if (isTimeSeries(frame)) {
           graphFrames.push(frame);
-        } else if (shouldShowInVisualisationTypeStrict(frame, 'trace')) {
-          traceFrames.push(frame);
-        } else if (shouldShowInVisualisationTypeStrict(frame, 'table')) {
           tableFrames.push(frame);
-        } else if (isTimeSeries(frame, datasourceInstance?.meta.id)) {
-          if (shouldShowInVisualisationType(frame, 'graph')) {
```
aocenas, line 53: This could have been removed because if it had any visualisation type meta info we would already match it before. So here we are only in case it does not have any preferences for vis type. This means we did not need the matcher functions anymore.

public/app/features/explore/utils/decorators.ts
```diff
@@ -1,214 +1,187 @@
...
-      const graphFrames: DataFrame[] = [];
-      const tableFrames: DataFrame[] = [];
-      const logsFrames: DataFrame[] = [];
-      const traceFrames: DataFrame[] = [];
+/**
+ * When processing response first we try to determine what kind of dataframes we got as one query can return multiple
+ * dataFrames with different type of data. This is later used for type specific processing. As we use this in
+ * Observable pipeline, it decorates the existing panelData to pass the results to later processing stages.
+ */
+export const decorateWithGraphLogsTraceAndTable = (data: PanelData): ExplorePanelData => {
```
aocenas, line 25: Functions here were changed to just by a mapper functions. That should make it easier to test and reuse if needed with only tradeoff that we have to use them with `map()` when using with observables.

public/app/features/explore/utils/decorators.ts
```diff
@@ -1,214 +1,187 @@
...
-      const logsResult = { ...sortedNewResults, rows, series };
-
-      return { ...data, logsResult };
-    })
-  );
-
-function isTimeSeries(frame: DataFrame, datasource?: string): boolean {
-  // TEMP: Temporary hack. Remove when logs/metrics unification is done
-  if (datasource && datasource === 'cloudwatch') {
-    return isTimeSeriesCloudWatch(frame);
```
aocenas, line 184: We should not be needing this as right now the new test is handling wide dataFrames. @kaydelaney just for sure, the cloudWatch check would also pass dataFrames with time/numeric/string columns as timeSeries dataFrames. Is that how it should work? How do we handle the string columns in that case?

public/app/features/explore/utils/decorators.ts
hugohaggmark: This simplifies the tests because it removes RxJs and observableTester right?

public/app/features/explore/utils/decorators.test.ts
```diff
@@ -86,450 +84,337 @@ const createExplorePanelData = (args: Partial<ExplorePanelData>): ExplorePanelDa
...
-          const data = panelData.tableResult;
-          expect(data?.fields[0].display).toBe(displayFunctionMock);
-        },
-        done,
-      });
-    });
+  it('should process table type dataFrame', async () => {
+    const { table, emptyTable } = getTestContext();
+    const panelData = createExplorePanelData({ tableFrames: [table, emptyTable] });
+    const panelResult = await decorateWithTableResult(panelData).toPromise();
```
hugohaggmark, line 211: Moving away from OperatorFunctions introduces `toPromise` in tests which is in my point of view is a hack.

hugohaggmark (approved):
Great work on this, I can see why you made the change but I'm not sure I agree with it. 

I think using OperatorFunctions are a lot more clearer/cleaner as we're using them within an observable pipe. 

Having the decorators return either ExplorePanelData or Observable<ExplorePanelData> doesn't seem as clean. 

Don't let my strong opinions about this discourage you from proceeding with this change.

public/app/features/explore/utils/decorators.ts
aocenas: Yes. The table mapper still returns observable due to the transformer so that needs to be handled, but others don't do anything async so could be used as simple mappers. It's not a big change and I mostly done it as I was exploring how the code works right now. Also the table transform could probably use toArray and then toPromise in case it emits multiple times but I assume it's not necessary at the moment.

public/app/features/explore/utils/decorators.test.ts
hugohaggmark: Because it's an observable and you're treating it as a Promise.

aocenas:
>  think using OperatorFunctions are a lot more clearer/cleaner as we're using them within an observable pipe.

> Having the decorators return either ExplorePanelData or Observable doesn't seem as clean.

I agree that having different return types can seem a bit worse but to me personally I don't think we should make into observable things that do not need to be observables. For me it's the same as returning a promise when you don't need to. If the code is async then it should be async but if it isn't forcing it into promise or observable seems off to me, like having a parameter that isn't used.

It also forces some particular usage in this case having observable pipeline for code don't need it. I am all in for using observables to handle async cases as better abstraction then promises but we should not overuse it similar to how we don't return promises from every function even if it's used from async code.

Simpler functions are simpler to use and lifted if needed to use in both promise abstraction or observables or without any of that. It's not the same other way around as too much abstraction is usually harder to get rid of, like in case of testing where it needs special tools to which you need to know and learn separately adding a bit of overhead that again isn't necessary.

public/app/features/explore/utils/decorators.test.ts
aocenas: You can transform between those quite easily and in the test you are interested in end result and wan't to see what it is. You don't care here about particular way it is executed here because its 1 input and one observable output. Also for example the `subscribeAndExpectOnNext` which was used before would not handle multiple emissions well because if there were multiple values you still have single expect function and so you don't test different values as they are executed, you would have to have an array of expect functions and probably test if there are more values than expects to make sure all is tested. In that case just collecting the results, waiting for complete and check all the results is more straightforward and you did not loose any information.

This is a bit different from testing an observable pipeline where you have multiple input streams and need to check timing of the events. There handling it with manually mocking the timers is tedious and for example the marble tests make more sense but this is much more simpler case.

aocenas:
@hugohaggmark merging this so we can get the fix in the release but would like to continue the discussion as I did not think the refactor part would be a controversial thing.

hugohaggmark:
@aocenas no worries, I don't think it's controversial at all; I think we just have different points of view.

PR #28396, opened 2020-10-20
Docs: Fixing v51 link

andor-pierdelacabeza:

The "What's new in v5.1" doc had an HTML anchor to an element that has changed, so it has to be changed where it's referenced.

It used to be:
/installation/docker/#migration-from-a-previous-version-of-the-docker-container-to-5-1-or-later

and now it is:
/installation/docker/#migrate-to-v51-or-later

It changed on c344a3a66e47c180867fb77a869774b69af18a89


oddlittlebird (approved):
LGTM. Thanks for contributing to the docs!

PR #28405, opened 2020-10-20
Chore: Set timezone for tests to non utc.

aocenas:
Set the test timezone to something else than UTC. Using Pacific/Easter because it has DST, is 6h diff (should be more noticeable) and is something that probably has little chance that somebody from Grafana lives there so nobody should feel as getting preferential treatment :).

Otherwise the comment sums the reasoning for non UTC timezone
// 1. It makes sense for both CI tests and local tests to behave the same so issues are found earlier
// 2. Any wrong timezone handling could be hidden if we use UTC/GMT local time (which would happen in CI).

### DIFF
--- a/packages/grafana-data/src/datetime/timezones.test.ts
+++ b/packages/grafana-data/src/datetime/timezones.test.ts
@@ -2,28 +2,25 @@ import { getTimeZoneInfo } from './timezones';
 import { setTimeZoneResolver } from './common';
 
 describe('getTimeZoneInfo', () => {
-  // global timezone is set to UTC, see jest-config.js file
+  // global timezone is set to Pacific/Easter, see jest-config.js file
 
   describe('IANA canonical name of the timezone', () => {
     it('should resolve for default timezone', () => {
       setTimeZoneResolver(() => 'browser');
       const result = getTimeZoneInfo('', Date.now());
-      expect(result?.ianaName).toBe('Africa/Abidjan');
+      expect(result?.ianaName).toBe('Pacific/Easter');
     });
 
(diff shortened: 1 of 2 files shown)


packages/grafana-data/src/datetime/timezones.test.ts
```diff
@@ -2,28 +2,25 @@ import { getTimeZoneInfo } from './timezones';
...
 describe('getTimeZoneInfo', () => {
-  // global timezone is set to UTC, see jest-config.js file
+  // global timezone is set to Pacific/Easter, see jest-config.js file
 
   describe('IANA canonical name of the timezone', () => {
     it('should resolve for default timezone', () => {
       setTimeZoneResolver(() => 'browser');
       const result = getTimeZoneInfo('', Date.now());
-      expect(result?.ianaName).toBe('Africa/Abidjan');
+      expect(result?.ianaName).toBe('Pacific/Easter');
```
aocenas, line 11: This `Africa/Abidjan` was weird. I assume it's because utc isn't a real timezone so when moment tried to infer the timeozne it jest returned the first +00:00 timezone there is alphabetically.

mckn: Agree, that looks wieird.

mckn (approved):
I think this makes sense and looks really good!

dprokop (approved):
@aocenas I think this looks good, great catch about the UTC, and sorry for the trouble!

aocenas:
@dprokop no worries. Without that I wouldn't even know you can mock the timezone and it's definitely a good idea to have it aligned for testing.

PR #28412, opened 2020-10-20
Add SVG Cell in Table Component

miguelangellv:

That PR add SVG Image Cell to Table Component. Now, you can add SVG images, as Material Icons, with thresholds color and  value mapping support.

Is very usefull for show states (online/offline for example).

Support SVG imgaes

Fixes #

### DIFF
--- a/packages/grafana-ui/src/components/Table/SVGCell.tsx
+++ b/packages/grafana-ui/src/components/Table/SVGCell.tsx
@@ -1,0 +1,24 @@
+import React, { FC } from 'react';
+import { DisplayValue, Field } from '@grafana/data';
+import { TableStyles } from './styles';
+import { TableCellProps } from './types';
+
+export const SVGCell: FC<TableCellProps> = props => {
+  const { field, cell, tableStyles, cellProps } = props;
+
+  const displayValue = field.display!(cell.value);
+
+  const cellStyle = getCellStyle(tableStyles, field, displayValue);
+
+  return (
+    <div
+      {...cellProps}
+      className={cellStyle + ' ' + tableStyles.imageCell}
+      dangerouslySetInnerHTML={{ __html: displayValue.text }}
+    ></div>
+  );
+};
+
+function getCellStyle(tableStyles: TableStyles, field: Field, displayValue: DisplayValue) {
+  return tableStyles.buildCellContainerStyle(displayValue.color);
+}
--- a/packages/grafana-ui/src/components/Table/types.ts
+++ b/packages/grafana-ui/src/components/Table/types.ts
@@ -19,6 +19,7 @@ export enum TableCellDisplayMode {
   JSONView = 'json-view',
   BasicGauge = 'basic',
   Image = 'image',
+  SVG = 'svg',
 }
 
 export type FieldTextAlignment = 'auto' | 'left' | 'right' | 'center';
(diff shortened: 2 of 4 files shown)


torkelo:
Why does this need new cell? Can't it use the image cell?

MiguelAngelLV:
@torkelo 

Image cell use URL to image.  SVG cell use the SVG directly.

With SVG, the SVG color change using thresholds...

And not use  tag, use <svg> tag.

Could they be merged with a fews checks as "start with <svg" but... I think is more clean two cells.

packages/grafana-ui/src/components/Table/SVGCell.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  const displayValue = field.display!(cell.value);
+
+  const cellStyle = getCellStyle(tableStyles, field, displayValue);
+
+  return (
+    <div
+      {...cellProps}
+      className={cellStyle + ' ' + tableStyles.imageCell}
+      dangerouslySetInnerHTML={{ __html: displayValue.text }}
```
torkelo, line 17: we cannot allow this, not secure

torkelo:
> With SVG, the SVG color change using thresholds...

I do not see anything here that relates to thresholds

packages/grafana-ui/src/components/Table/SVGCell.tsx
MiguelAngelLV: I could add safe-html (https://www.npmjs.com/package/safe-html) to ensure that is only a SVG.

Ok?

ryantxu:
Would it make more sense to have an HTML cell type?  and then manage the values the same way we do for the text panel?  https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/text/TextPanel.tsx#L76

MiguelAngelLV:
Welll... Now it works as "Html Cell"..  I can adapt it...

MiguelAngelLV:
@ryankurte TextPanel only works with markdown... And I don't know if markdown support svg

packages/grafana-ui/src/components/Table/SVGCell.tsx
torkelo: we already have sanitize functions in grafana/data https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/text/sanitize.ts

MiguelAngelLV: Jums... I see that configuration is only for CSS attributes. So, I need create a new config por XSS with some tags.

I think... a, span, strong, italic, svg, img, h1-6.... any else?

dprokop:
@MiguelAngelLV thank you for the contribution! I don't think we need a special cell type to display SVGs. You can already do that using base64 encoded SVG and image cells.

MiguelAngelLV:
@dprokop the problem with base64 is that you can't change the color dynamically

peterholmberg:
Hi @MiguelAngelLV, we don't want to add more table cell types at this point without a strong need from the community, therefore we have decided to close this PR for now and encourage you to open a [feature request](https://github.com/grafana/grafana/issues/new?labels=type%3A+feature+request&template=2-feature_request.md) for this. If it gains traction and interest from the community we can reopen this pull request.

PR #28416, opened 2020-10-20
WIP/Live: disconnect orphaned channels

ryantxu:
This is a follow up to https://github.com/grafana/grafana/pull/28022 that aims to solve the key issue that subscriptions to channels exist after the stream has been unsubscribed.

In particular trying to solve this issue:
https://github.com/grafana/grafana/pull/28022#issuecomment-712830054


public/app/features/live/channel.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    const subA = chan.getStream().subscribe(noop);
+    expect(chan.getObserverCount()).toEqual(1);
+    const subB = chan.getStream().subscribe(noop);
+    expect(chan.getObserverCount()).toEqual(2);
+
+    subA.unsubscribe();
+    subB.unsubscribe();
+
+    expect(chan.getObserverCount()).toEqual(0);
```
torkelo, line 22: is there a way the channel can remove itself when the last subscriber unsubscribes? If you want to keep the channel open for some seconds after we could have a timeout that then checks if it still has no other subscribers?

ryantxu: yes -- I'll look into that

ryantxu:
part of #32311

PR #28417, opened 2020-10-20
TestData: multiple arrow requests should return multiple frames

ryantxu:
Fix a bug where only one result is returned from multiple queries.  When the refId and `key` are set, both frames are returned:

image


torkelo (approved):
LGTM

dprokop (approved):

PR #28419, opened 2020-10-21
Chore: Configure go-ruleguard via golangci-lint

aknuds1:
Configure golangci-lint to use [go-ruleguard](https://github.com/quasilyte/go-ruleguard) with some initial custom rules (copied from Damian Gryski's [semgrep-go](https://github.com/dgryski/semgrep-go) repo) to catch some Go anti-patterns.


pkg/bus/bus.go
```diff
@@ -143,9 +143,9 @@ func (b *InProcBus) Publish(msg Msg) error {
 
 	for _, listenerHandler := range listeners {
 		ret := reflect.ValueOf(listenerHandler).Call(params)
-		err := ret[0].Interface()
-		if err != nil {
-			return err.(error)
+		err, ok := ret[0].Interface().(error)
+		if ok && err != nil {
+			return err
 		}
```
sakjur, line 149: Thought, maybe it's a good thing to safe up by making sure that if it's _not_ an error we complain

```suggestion
		if ok && err != nil {
			return err
		} else if !ok {
			return fmt.Errorf("expected listener to return an error, got '%v'", reflect.TypeOf(ret[0].Interface()))
		}
```

pkg/middleware/auth_proxy.go
```diff
@@ -15,11 +15,11 @@ var header = setting.AuthProxyHeaderName
 func logUserIn(auth *authproxy.AuthProxy, username string, logger log.Logger, ignoreCache bool) (int64, *authproxy.Error) {
 	logger.Debug("Trying to log user in", "username", username, "ignoreCache", ignoreCache)
 	// Try to log in user via various providers
-	id, err := auth.Login(logger, ignoreCache)
-	if err != nil {
-		logger.Error("Failed to login", "username", username, "message", err.Error(), "error", err.DetailsError,
+	id, e := auth.Login(logger, ignoreCache)
```
sakjur, line 18: In case other reviewers wonder about this:
I looked this up, `e` is an `*auth.Error`, not `error`.

pkg/ruleguard.rules.go
```diff
@@ -0,0 +1,488 @@
+// +build ignore
+
+package gorules
```
sakjur, line 3: Are the commented out code something we've done or from the source?

Are we properly crediting the source of this rule file?

sakjur (commented):
Oh, cool, I didn't know that existed.
Are we crediting the original creator of these rules appropriately? I like the direction we're heading in here :+1:

papagian (approved):
LGTM

aknuds1:
> Oh, cool, I didn't know that existed.
> Are we crediting the original creator of these rules appropriately? I like the direction we're heading in here +1

@sakjur Good question - there's no license, should we do anything but perhaps credit Damian in the commit message?

pkg/ruleguard.rules.go
aknuds1: @sakjur Should we add a comment documenting the origin of the file?

pkg/middleware/auth_proxy.go
aknuds1: Yes, I suspect we should convert this code later to return an `error` implementation :)

sakjur: Yeah, I think that's out of scope though :)

aknuds1: @sakjur Absolutely :D

sakjur:
@aknuds1 Build fails because we need an import "fmt" as well

pkg/ruleguard.rules.go
aknuds1: Added a comment crediting origin.

sakjur (approved):
Still not sure how we should do attribution but :+1:

aknuds1:
> Still not sure how we should do attribution but +1

@sakjur Did you see the comment I added in pkg/ruleguard.rules.go?

sakjur:
Now I did, that looks like a good attribution :+1:
It'd be nice with explicit licensing as well to make sure we're in the clear, but it isn't being linked into the binary, so that's good.

aknuds1:
> Now I did, that looks like a good attribution +1
> It'd be nice with explicit licensing as well to make sure we're in the clear, but it isn't being linked into the binary, so that's good.

@sakjur What can we do about licensing though? I couldn't find any license in the repo.

sakjur:
We could ask upstream for a license, but as there are a few external contributors (IANAL but many of them seem to be fortunately not of consequence, spellfixes, minor changes which cannot be considered to be works in and of themselves) it's not necessarily easily done.

aknuds1:
I commented in an existing [issue](https://github.com/dgryski/semgrep-go/issues/16) (made 1 hour ago!) requesting the license.

dgryski:
A quick comment here about gating builds/merges on ruleguard failures:  I (on purpoose!) have not been so careful to avoid false positives in the rules I'm creating.  While I'd like to avoid the very obvious ones, some limitations of ruleguard/semgrep prevent full analysis required to *only* catch true issues.  Things like staticcheck and vet have a much higher bar.

aknuds1:
> A quick comment here about gating builds/merges on ruleguard failures: I (on purpoose!) have not been so careful to avoid false positives in the rules I'm creating. While I'd like to avoid the very obvious ones, some limitations of ruleguard/semgrep prevent full analysis required to _only_ catch true issues. Things like staticcheck and vet have a much higher bar.

@dgryski Interesting, thanks for sharing that info. I'm so new to go-ruleguard/semgrep that I'm unsure how they relate to other tools, e.g. staticcheck, vet. Guess I should dig into those once I have more time. Do you have any tips for learning about the alternatives available when it comes to Go static analysis?

dgryski:
Basically, staticcheck and vet should have virtually *no* false positives.  However, this means that they can't be used to flag suspicious code which is technically correct.  Also, adding custom checks is compicated.  ruleguard and semgrep fill this void by having a simpler format for writing checks, but it's also a less powerful query language meaning it's hard if not impossible to eliminate classes of false positives that would be possible to eliminate if you were writing a check with the full power of Go.  ruleguard and semgrep are trading simplicity of writing custom checks for accuracy.

aknuds1:
@dgryski Thanks for clarifying that! There's clearly a lot for me to dig into :)

marefr (approved):
Looks useful :+1: LGTM

PR #28420, opened 2020-10-21
Chore: Rewrite some tests to use testify

aknuds1:
Rewrite some tests to use testify assertions instead of the standard library directly.


papagian (approved):
LGTM

As a minor remark `sql_engine_test.go` is not converted to use testify as the title suggests.

aknuds1:
> LGTM
> 
> As a minor remark `sql_engine_test.go` is not converted to use testify as the title suggests.

@papagian Yeah, it's a two-step rocket, convert to framework assertions, then after converting from go-convey, to testify.

PR #28421, opened 2020-10-21
Chore: Use net.JoinHostPort

aknuds1:
Use [net.JoinHostPort](https://golang.org/pkg/net/#JoinHostPort) function to join host and port into network address. Apparently the benefit is it will deal with eventual IPv6 hosts. Semgrep caught these issues.

### DIFF
--- a/pkg/api/http_server.go
+++ b/pkg/api/http_server.go
@@ -107,7 +107,7 @@ func (hs *HTTPServer) Run(ctx context.Context) error {
 	hs.applyRoutes()
 
 	hs.httpSrv = &http.Server{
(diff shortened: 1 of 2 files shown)


sakjur (approved):

wbrowne (commented):
I wonder if there are any more references we should care about. Is this something we should then actively change if we come across it?

papagian (approved):
LGTM

aknuds1:
> I wonder if there are any more references we should care about. Is this something we should then actively change if we come across it?

@wbrowne I think that's all, except for one complicated case where net.JoinHostPort doesn't work. If we merge my PR to enable go-ruleguard, we can catch this automatically.

PR #28436, opened 2020-10-21
CloudWatch/Athena - valid metrics and dimensions.

kwarunek:

Separate Athena's  dimensions from the metrics. 

Grafana does not prompt with a dimension for CloudWatch/Athena

    t=2020-10-21T14:46:56+0200 lvl=eror msg="Metric request error" logger=context userId=534 orgId=275 uname=test error="unable to find dimension \"AWS/Athena\"" remote_addr=10.20.30.40

It turns out that dimensions are, incorrectly, combined with metrics. 

Change in accordance with https://docs.aws.amazon.com/athena/latest/ug/query-metrics-viewing.html.


pkg/tsdb/cloudwatch/metric_find_query.go
```diff
@@ -48,7 +48,7 @@ var metricsMap = map[string][]string{
...
-	"AWS/Athena":                  {"DataScannedInBytes", "EngineExecutionTime", "QueryPlanningTime", "QueryQueueTime", "QueryState", "QueryType", "ServiceProcessingTime", "TotalExecutionTime", "WorkGroup"},
+	"AWS/Athena":                  {"DataScannedInBytes", "EngineExecutionTime", "QueryPlanningTime", "QueryQueueTime", "ServiceProcessingTime", "TotalExecutionTime"},
```
papagian, line 51:
```suggestion
	"AWS/Athena":                  {"EngineExecutionTime", "ProcessedBytes", "QueryPlanningTime", "QueryQueueTime", "ServiceProcessingTime", "TotalExecutionTime"},
```

I think that `DataScannedInBytes` is no more around; at least it does appear in the docs.

papagian (changes requested):
Generally LGTM; I've spotted an additional metric and an (obsolete?) one, WDYT?

pkg/tsdb/cloudwatch/metric_find_query.go
kwarunek: good catch

papagian (approved):

kylebrandt:

PR #28446, opened 2020-10-21
Docs: Add survey content

oddlittlebird:
Add survey code snippet to various topics.

@robbymilo, after the latest Docker image update, the code snippets are not displaying on my local build. Any idea why?


achatterjee-grafana (approved):
LGTM! Can we move it higher to where the doc content ends? That will make it more visible.

PR #28465, opened 2020-10-22
Plugin page: update readme icon

Clarity-89:
The old `fa` icons are missing the right margin and look a bit too close to their text. Using icon name instead fixes the issue:

Before:
Screenshot 2020-10-22 at 13 03 14
After:
Screenshot 2020-10-22 at 13 03 51


kaydelaney (approved):
LGTM

PR #28480, opened 2020-10-22
AzureMonitor: Support decimal (as float64) type in analytics/logs

### LINKED ISSUE #28278: Azure Insights Analytics, unsupported analytics column type decimal
**What happened**:
When  using the `todecimal()` function in Insights Analytics queries, you get this error: 
`unsupported analytics column type decimal` 
image

Getting a result to the query. 

Create a new query using Azure Monitor data source, and Insights Analytics service. 
The query should return a number, use `todecimal()` on that number

Replacing `todecimal()` by `toreal()` works. 

**Environment**:
- Grafana version: 7.2.1
- Data source type & version: Azure Monitor
- OS Grafana is installed on: 
- User OS & Browser:
- Grafana plugins:
- Others:

kylebrandt:
Converters the kusto 128-bit decimal type into a 64-bit float64 instead of returning an error.

Fixes #28278

Following work already done in ADX plugin, which is very similar https://github.com/grafana/azure-data-explorer-datasource/blob/a710e2d67fd40fc79db4c1dc5f087168e19eaf07/pkg/azuredx/table.go#L245


pkg/tsdb/azuremonitor/azure-log-analytics-table-frame.go
```diff
@@ -194,3 +195,41 @@ var longConverter = data.FieldConverter{
...
+		if sOk {
+			out, err := strconv.ParseFloat(jS, 64)
+			if err != nil {
+				return nil, err
+			}
+			return &out, err
+		}
+
+		// As far as I can tell this always comes in a string, but this was in the
+		// ADX code, so leaving it in case does sometimes become a number somehow.
```
kylebrandt, line 224:
```suggestion
		// As far as I can tell this always comes in a string, but this is in the
		// ADX code, so leaving this in case values do sometimes become a number somehow.
```

pkg/tsdb/azuremonitor/testdata/loganalytics/7-log-analytics-all-types-table.json
```diff
@@ -38,7 +38,10 @@
           {
             "name": "XTimeSpan",
             "type": "timespan"
-          }
+          },
+          {
+            "name":"XDecimal",
+            "type":"decimal"}
```
aknuds1, line 44:
```suggestion
            "type":"decimal"
          }
```

aknuds1 (approved):
Sorry for the slow review, but LGTM. Just left a suggestion for some JSON cleanup.

PR #28481, opened 2020-10-22
Auth: Fix SigV4 request verification step for Amazon Elasticsearch Service

### LINKED ISSUE #28519: Sigv4 authentication returns 403 when running Grafana in Docker
**What happened**:

When attempting to set up an AWS Elasticsearch data source using the new AWS Sigv4 Auth feature, requests from a running Grafana Docker container to the Elasticsearch cluster fail with a 403 unauthorized error. 

The Elasticsearch cluster we are attempting to use as a datasource is running in a [private VPC](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-vpc.html), and has both [encryption at rest](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/encryption-at-rest.html) (encrypted with our own [CMK](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#master_keys)), and node-to-node encryption enabled. The security groups are properly configured to permit inbound network access to the cluster (if they weren't, of course, the requests would time out rather than returning a 403).

Screen Shot 2020-10-23 at 5 31 43 PM

I have turned on debug logging and nothing particularly helpful appears in the logs. Interestingly, when I select the **Explore** option in the panel menu and try to query the data source, I see the error `Credential should be scoped to correct service: 'es'`

Screen Shot 2020-10-23 at 5 31 53 PM

I have tested this on Grafana running as an ECS Fargate service using both the **AWS SDK Default** option with an IAM role that has access to the cluster, and with the **Access & Secret Key** option using an IAM user I created for testing, both with the same result. I've also tried in Grafana running in Docker locally on my Mac (using a VPN to to get network access to the cluster) using the **Access & Secret Key** option with those same static credentials, with the same result.

When I set up a proxy using [aws-es-proxy](https://github.com/abutaha/aws-es-proxy) and use the proxy URL for the data source, using both an IAM role and the static IAM user credential, requests to the cluster are successful. Also, if I enable open access in the cluster's access policy, requests from Grafana to the cluster are successful.

Adding an AWS Elasticsearch Cluster as a data source, then selecting the Sigv4 Auth option, and entering either an IAM role or static credentials should result in requests to the cluster completing successfully, rather than returning a 403 Unauthorized error.

- Start Grafana in a Docker container (the Alpine Linux version) with Sigv4 auth enabled
- Add an AWS Elasticsearch cluster as a data source
- Click the slider to enable Sigv4 Auth. Select one of the following options:
  - *Running as an ECS Fargate service*: **AWS SDK default**, enter the default region, then leave the **Assume Role ARN** blank to (I assume) use the task role assigned to the running ECS task:
    Screen Shot 2020-10-23 at 5 31 21 PM
  - *Running locally or as an ECS Fargate service*: **Access & Secret Key**, then enter the Access Key ID, Secret Access Key, and default region of an IAM user with access to the cluster:
    Screen Shot 2020-10-23 at 6 20 43 PM

- Click **Save & Test**

**Environment**:
- Grafana version: 7.3.0-beta2
- Data source type & version: AWS Elasticsearch 6.7 (VPC)
- OS Grafana is installed on: Docker (Alpine Linux version)
- User OS & Browser: macOS Chrome 86.0.4240.111
- Grafana plugins: Elasticsearch
- Others:

PR where this feature was introduced: https://github.com/grafana/grafana/pull/27552/

wbrowne:
Fixes the following:
- Ensures adding a valid AWS service name when creating request signature (`es` - Elasticsearch, `prometheus` - which is still TBD so is just a sane placeholder for now)
- Ensures that we encode necessary characters in the request URL if present
- Exclude certain headers (`X-Forwarded-For`) when signing as it breaks signature verification - based on the assumption that AWS ignore this header on their end
- Makes the AWS SDK default mechanism the default auth select option in the frontend

Fixes #28519.


packages/grafana-ui/src/components/DataSourceSettings/DataSourceHttpSettings.mdx
```diff
@@ -13,7 +13,7 @@ It is used in a `ConfigEditor` for data source plugins. You can find more exampl
 ### Example usage
 ```jsx
 export const ConfigEditor = (props: Props) => {
-  const { options, onOptionsChange, config } = props;
+  const { options, onOptionsChange } = props;
```
wbrowne, line 16: An incorrect prop I forgot to remove from the original SigV4 PR

pkg/models/sigv4.go
```diff
@@ -108,3 +123,26 @@ func (m *SigV4Middleware) credentials() (*credentials.Credentials, error) {
 
 	return nil, fmt.Errorf("unrecognized authType: %s", authType)
 }
+
+func replaceBody(req *http.Request) ([]byte, error) {
```
wbrowne, line 127: FYI this is code from the aws es proxy lib

marefr (approved):
Tested:
- Works as expected when configuring data source and testing it (GET index mappings).
- Works as expected when querying by using POST to the multi search endpoint which is being used in Grafana for querying (/api/datasources/proxy/230/_msearch?max_concurrent_shard_requests=5
- Works as expected when Elasticsearch data source executes in Grafana backend (created an alert rule to verify)

LGTM great work with this! :tada:

pkg/models/sigv4.go
```diff
@@ -63,15 +64,29 @@ func (m *SigV4Middleware) signRequest(req *http.Request) (http.Header, error) {
...
+
+	// if X-Forwarded-For header is present, omit during signing step as it breaks AWS request verification
+	forwardHeader := req.Header.Get("X-Forwarded-For")
+	if forwardHeader != "" {
+		req.Header.Del("X-Forwarded-For")
+
+		header, err := signer.Sign(req, bytes.NewReader(body), awsServiceNamespace(m.Config.DatasourceType), m.Config.Region, time.Now().UTC())
+
+		// reset pre-existing X-Forwarded-For header value
+		req.Header.Set("X-Forwarded-For", forwardHeader)
```
aknuds1, line 84: I think it would be much cleaner to move line 84 to 80, because you now have code executing after a potential error happens (line 81) and it requires thorough reading to see that there isn't a bug (since the standard pattern is to check and deal with the error immediately).

aknuds1: Never mind, on looking it through also locally!

PR #28490, opened 2020-10-23
Drone: Add version branch pipelines

### LINKED ISSUE #28464: Build version branches in CI
Version branches should be built by Drone when they get merged into. This is so we can catch errors quickly.

aknuds1:
Add Drone pipelines to build on pushes to version branches.

This pipeline topology is basically the same as for releases, except nothing gets published.

Fixes #28464.


marefr (approved):
I think it looks good :)

PR #28493, opened 2020-10-23
Performance: Make onChange passed to query editors non-anonymous

Elfo404:
the `onChange` handler is currently an anonymous function, this prevents query editors to opt-out from rerendering (eg. using `memo`), this is particularly important when on edit mode a dashboard is set to refresh automatically.

this change avoids passing an anonymous function to the query editors while retaining *almost* the same functionality. The only difference is that in order for this to work, query editors are required to mutate the whole query object every time they need to update a query (which imho would be a good practice to enforce anyway) but maybe there are better solutions.

eg:
```ts
// this won't work anymore
const MyEditor = ({query, onChange}) => {
  // [...]
  query.query = 'SELECT * FROM ...';
  onChange(query);
}

// this will:
const MyEditor = ({query, onChange}) => {
  // [...]
  query.query = 'SELECT * FROM ...';
  onChange({ ...query });
}
```

On the new Elastic Search datasource (being worked on) opting out from rerendering gives the folowing improvement:
(the benchmark is far from great and might need more iterations)
MacbookPro 2020, i5, Chrome, 4x CPU slowdown:
Before
before

After:
after

Flame chart (Query editor not re-rendering):
Screenshot 2020-10-05 at 23 44 10

Note that this won't automatically improve performance of every query editor as each one of them needs to manually opt-ot from rerendering ignoring props that are not relevant for them, as there are still other props that change at every render (like `timeRange` for instance) 

Fixes #

### DIFF
--- a/public/app/features/dashboard/panel_editor/QueryEditorRows.tsx
+++ b/public/app/features/dashboard/panel_editor/QueryEditorRows.tsx
@@ -42,27 +42,26 @@ export class QueryEditorRows extends PureComponent<Props> {
     panel.refresh();
   };
 
-  onChangeQuery(query: DataQuery, index: number) {
+  onChangeQuery = (query: DataQuery) => {
     const { queries, onChangeQueries } = this.props;
 
-    const old = queries[index];
+    // Assuming datasources won't mutate queries, we find the index of the query that was updated.
+    const prevQueryIndex = queries.findIndex((_, i, oldQueries) => oldQueries[i] !== queries[i]);
 
-    // ensure refId & datasource are maintained
-    query.refId = old.refId;
-    if (old.datasource) {
-      query.datasource = old.datasource;
-    }
-
-    // update query in array
     onChangeQueries(
       queries.map((item, itemIndex) => {
-        if (itemIndex === index) {
-          return query;
+        if (itemIndex === prevQueryIndex) {
+          // ensure refId & datasource are maintained
+          return {
+            ...query,
+            refId: queries[prevQueryIndex].refId,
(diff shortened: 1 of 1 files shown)


public/app/features/dashboard/panel_editor/QueryEditorRows.tsx
```diff
@@ -42,27 +42,26 @@ export class QueryEditorRows extends PureComponent<Props> {
     panel.refresh();
   };
 
-  onChangeQuery(query: DataQuery, index: number) {
+  onChangeQuery = (query: DataQuery) => {
     const { queries, onChangeQueries } = this.props;
 
-    const old = queries[index];
+    // Assuming datasources won't mutate queries, we find the index of the query that was updated.
+    const prevQueryIndex = queries.findIndex((_, i, oldQueries) => oldQueries[i] !== queries[i]);
```
aocenas, line 49: How is this supposed to work? `oldQueries` is `queries` the third arg is just the whole array you are iterating over. So this should always return -1 or am I missing something?

torkelo: we need the index, query editors SHOULD pass a new updated query model instance to onChange (Query editors should not modify the props query model instance)

Elfo404: It was clearly too late last night, sorry.
I didn't push the correct version, forget the whole `mutate queries` story, the actual solution was comparing queries' `refId` assuming query editors don't mutate it.

I'll update the PR

PR #28494, opened 2020-10-23
Drone: Don't build Windows installers for version branches

aknuds1:
Modify Drone config to not build Windows installers for version branches, since the required zip archives aren't uploaded. The builds break without this change.


wbrowne (approved):
LGTM from what I can tell 😉

PR #24999, opened 2020-05-22
Elastic: prevents datasource from throwing unexpected errors for invalid JSON

Estrax:
**What this PR does / why we need it**:

Adds additional error handling to GET requests in Elasticsearch datasource. Adds response parsability checks throwing errors when the response coming back is not parsable (e.g. being a plain text or HTML - when the datasource goes down while processing the request).

**Which issue(s) this PR fixes**:

Fixes #24843 and fixes #24857

PR #28498, opened 2020-10-23
Elasticsearch: Fix handling of errors when testing data source

marefr:
#24999 changed the error response payload, but the test data source
was not adapted to this change and broke the feature of displaying
any errors to the user in the UI. This change should resolve this
problem.

Ref #24999
Ref #28481 

Found when testing #28481


wbrowne (approved):
LGTM

PR #28521, opened 2020-10-24
GCS image uploader: Add tests

aknuds1:

* Rewrite the GCS image uploader to use interfaces in front of the Google SDK, to make the code testable (to avoid network requests to Google services)
* Add test mocks of the SDK interfaces
* Add a test for unsigned URLs
* Add a test for signed URLs

These tests should be really useful, since I recently had to stop a PR from accidentally removing the signed URL functionality.


kylebrandt (approved):
Looks good to me. I'm not clear on what generates the file? Do we need any comments somewhere on how/when one might regen that?

aknuds1:
> Looks good to me. I'm not clear on what generates the file? Do we need any comments somewhere on how/when one might regen that?

@kylebrandt Good point, maybe I should add some automation for regenerating the mock.

kylebrandt (changes requested):
Oops just noticed, I think you might be including an extra PR in here by mistake? https://github.com/grafana/grafana/pull/28525

aknuds1:
@kylebrandt Ugh, oops indeed.

aknuds1:
Will fix those issues.

aknuds1:
Rebased the branch to get rid of the unintended changes.

kylebrandt (approved):

PR #28525, opened 2020-10-25
gtime: Add ParseDuration function

aknuds1:
Add gtime.ParseDuration function, that operates with fixed durations independent of wall time. I've also changed usages of gtime.ParseInterval that I think should use wall time-independent durations. Please provide feedback if I got some of them wrong :)


sakjur:
This is a problem, I think the solution is to change the tests to tolerate the shifts rather than changing the definitions. I'm a little concerned that fixed definitions would break expectations and behavior in places. I'd have to take an extra look to be able to tell just what I think would be adverse consequences of this change.

NB: A year should be 13 months if a month is four weeks. https://en.m.wikipedia.org/wiki/Calendar_reform#13-month_calendars

aknuds1:
> This is a problem, I think the solution is to change the tests to tolerate the shifts rather than changing the definitions. I'm a little concerned that fixed definitions would break expectations and behavior in places. I'd have to take an extra look to be able to tell just what I think would be adverse consequences of this change.
> 
> NB: A year should be 13 months if a month is four weeks. https://en.m.wikipedia.org/wiki/Calendar_reform#13-month_calendars

@sakjur Yeah we do need to discuss it, I don't intend to rush this PR. It sounds fundamentally wrong to me though that a duration should be dependent on what the time is when you calculate it.

kylebrandt:
Apache arrow talks about "calendar interval":

```
enum IntervalUnit: short { YEAR_MONTH, DAY_TIME}
// A "calendar" interval which models types that don't necessarily
// have a precise duration without the context of a base timestamp (e.g.
// days can differ in length during day light savings time transitions).
// YEAR_MONTH - Indicates the number of elapsed whole months, stored as
//   4-byte integers.
// DAY_TIME - Indicates the number of elapsed days and milliseconds,
//   stored as 2 contiguous 32-bit integers (8-bytes in total).  Support
//   of this IntervalUnit is not required for full arrow compatibility.
table Interval {
  unit: IntervalUnit;
}
```

from https://github.com/apache/arrow/blob/7155cd5488310c15d864428252ca71dd9ebd3b48/format/Schema.fbs#L248

So if that is the case, perhaps the function should be changed such that it takes an argument for what time "now" is so we do get stable output.

Also I guess a comment explaining why we want a "calendar interval" here (or whatever we are calling it).

sakjur:
This is mostly used when parsing settings, so rather than feeding time.Now on initialization, a `CalendarInterval` type would allow us to apply any `time.Time` to the duration dynamically, which would keep the original intention, but create a stable interface.

I'm thinking something like:

```
type CalendarInterval struct {
years int32
months int8
days int8
milliseconds int32
}

func (i CalendarInterval) Apply(t0 time.Time) time.Time {
return t0.AddDate(i.years, i.months, i.days).Add(time.Duration(i.milliseconds) * time.Milliseconds)
}
```

aknuds1:
Just noting that our team discussed this, and we probably want to have both the notion of duration and calendar interval. This function should probably deal with durations, that have fixed definitions independent of wall time. In addition we might want a second function that deals with calendar intervals. This is all subject to further discussion though.

aknuds1:
@kylebrandt @sakjur Please have a look at the gtime.ParseDuration function I just added.

pkg/components/gtime/gtime.go
```diff
@@ -32,5 +32,47 @@ func ParseInterval(interval string) (time.Duration, error) {
...
+	}
+
+	return 0, fmt.Errorf("invalid duration %q", inp)
+}
+
+func parse(inp string) (int, string, error) {
+	result := dateUnitPattern.FindSubmatch([]byte(inp))
+	if len(result) != 3 {
+		dur, err := time.ParseDuration(inp)
+		return int(dur), "", err
```
sakjur, line 69: I'd return `time.Duration` here and cast it for `.AddDate` instead, since there are more paths where this is supposed to be a time.Duration all the time.

pkg/components/gtime/gtime.go
```diff
@@ -32,5 +32,47 @@ func ParseInterval(interval string) (time.Duration, error) {
...
+
+	switch period {
+	case "d":
+		return time.Duration(num*24) * time.Hour, nil
+	case "w":
+		return time.Duration(num*24*7) * time.Hour, nil
+	case "M":
+		return time.Duration(int64(float64(num*24)*(daysInAYear/12))) * time.Hour, nil
+	case "y":
+		return time.Duration(int64(float64(num*24)*daysInAYear)) * time.Hour, nil
```
sakjur, line 59:
```suggestion
		return time.Duration(float64(num*24)*daysInAYear/12) * time.Hour, nil
	case "y":
		return time.Duration(float64(num*24)*daysInAYear) * time.Hour, nil
```

- `float64` can be cast to `time.Duration` directly.
- Order of operation for `*` and `/` does not matter, removed parens.

https://play.golang.org/p/og8MyCyzqSB

pkg/components/gtime/gtime.go
```diff
@@ -32,5 +32,47 @@ func ParseInterval(interval string) (time.Duration, error) {
...
+	num, period, err := parse(inp)
+	if err != nil {
+		return 0, err
+	}
+	if period == "" {
+		return time.Duration(num), nil
+	}
+
+	// The average number of days in a year, using the Julian calendar
+	const daysInAYear = 365.25
```
sakjur, line 49: If we have time constants, we can then just multiply num with the relevant time constant and reduce the complexity of the maths.

Since `time.Duration` is nanosecond resolution, the loss of sub-nanoseconds when converting from float64 isn't too concerning even when multiplying with very big numbers 😉

```suggestion
	// The average number of days in a year, using the Julian calendar
	const daysInAYear = 365.25
	const day = 24 * time.Hour
	const week = 7 * day
	const year = time.Duration(float64(day) * daysInAYear)
	const month = year / 12
```

sakjur (commented):
📆 

Left some thoughts, but I think we could merge this as-is.
I haven't looked to see where we use ParseDuration and ParseInterval and if those makes sense overall, which is why I'm not adding a ✅

Maybe we should call these `ParseDurationRelative` and `ParseDurationFixed`? 🤔 It's a little vague what the difference between a duration and an interval is (I'd expect duration to be wall time and interval to be the fixed intervals based on the names, but this works the other way around).

aknuds1:
Thanks @sakjur, I've implemented your suggestions.

> Maybe we should call these `ParseDurationRelative` and `ParseDurationFixed`? thinking It's a little vague what the difference between a duration and an interval is (I'd expect duration to be wall time and interval to be the fixed intervals based on the names, but this works the other way around).

If you see @kylebrandt's [comment](https://github.com/grafana/grafana/pull/28525#issuecomment-716512612), there's prior art for referring to a base timestamp dependent as an interval. I think this makes sense, as I also (unlike you based on your comment) expect durations to be fixed and independent of any base timestamp.

sakjur (approved):
:+1: LGTM!

marefr (approved):
Make sense to me. LGTM

marefr:
Maybe some places in Enterprise that needs an update as well. Maybe something @sakjur can look into

PR #28531, opened 2020-10-26
Drone: Refactor logic

aknuds1:
Refactor version branch pipelines to reduce duplication.

There should be no practical differences in the generated Drone config (.drone.yml).


torkelo (approved):
Nice!

PR #28544, opened 2020-10-26
API: add login username in SendLoginLogCommand

AgnesToulet:
Add the username used for Grafana login in the `SendLoginLogCommand`. It's useful information when the login fails with a "user is disabled", "user lockout" or "invalid username and password" error. 

**Special notes for reviewers:**
I'm not sure about the labels I should use for this to be in 7.3.

### DIFF
--- a/pkg/api/login_test.go
+++ b/pkg/api/login_test.go
@@ -687,7 +687,7 @@ func TestLoginPostSendLoginLog(t *testing.T) {
 
 			cmd := testReceiver.cmd
 			assert.Equal(t, c.cmd.LogAction, cmd.LogAction)
-			assert.Equal(t, c.cmd.LoginUser, "admin")
+			assert.Equal(t, c.cmd.LoginUsername, "admin")
 			assert.Equal(t, c.cmd.HTTPStatus, cmd.HTTPStatus)
 			assert.Equal(t, c.cmd.Error, cmd.Error)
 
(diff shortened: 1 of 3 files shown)


xlson (approved):

pkg/api/login_test.go
```diff
@@ -687,6 +687,7 @@ func TestLoginPostSendLoginLog(t *testing.T) {
 
 			cmd := testReceiver.cmd
 			assert.Equal(t, c.cmd.LogAction, cmd.LogAction)
+			assert.Equal(t, c.cmd.LoginUsername, "admin")
```
aknuds1, line 690: I think you want to do the following? The second argument to assert.Equal is the expectation, and the third is what you got. What you got should be cmd.LoginUsername right?
```suggestion
			assert.Equal(t, "admin", cmd.LoginUsername)
```

aknuds1 (changes requested):
I think I spotted a bug in the test.

pkg/api/login_test.go
AgnesToulet: Yes exactly, I just fixed it, thanks! I'm still not used to put the expected value first.

AgnesToulet:
@aknuds1 Thanks, I fixed this. Do you know if the labels are good to have this in 7.3?

aknuds1 (approved):
LGTM

aknuds1:
> @aknuds1 Thanks, I fixed this. Do you know if the labels are good to have this in 7.3?

Yes, looks correct!

PR #28548, opened 2020-10-26
Plugins: Fix plugin URL paths on Windows

aknuds1:
Fix plugin URL paths on Windows.

Agnès discovered that plugin URLs had backslashes in them on Windows, and it looks to me as if this should be the culprits.

### DIFF
--- a/pkg/plugins/frontend_plugin.go
+++ b/pkg/plugins/frontend_plugin.go
@@ -55,7 +55,7 @@ func (fp *FrontendPluginBase) setPathsBasedOnApp(app *AppPlugin) {
 func (fp *FrontendPluginBase) handleModuleDefaults() {
 	if isExternalPlugin(fp.PluginDir) {
 		fp.Module = filepath.Join("plugins", fp.Id, "module")
-		fp.BaseUrl = filepath.Join("public/plugins", fp.Id)
+		fp.BaseUrl = path.Join("public/plugins", fp.Id)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


wbrowne (approved):
Was this broken just because it's an URL and it was treated as a filesystem path?

aknuds1:
> Was this broken just because it's an URL and it was treated as a filesystem path?

Yes. I haven't tested on Windows yet, but I hope these are the culprits. Looks like it from how it's supposed to be base URLs. Hopefully @AgnesToulet can confirm.

This bug happened because I fixed the opposite type of problem, i.e. that we were using `path.Join` for filesystem paths (`path.Join` is for network paths).

AgnesToulet (changes requested):
I still have issues with these changes. I don't really understand where and when these variables are used but this is working on Windows:
```
fp.Module = path.Join("plugins", fp.Id, "module")
fp.BaseUrl = filepath.Join("public/plugins", fp.Id)
```

aknuds1:
> I still have issues with these changes. I don't really understand where and when these variables are used but this is working on Windows:
> 
> ```
> fp.Module = path.Join("plugins", fp.Id, "module")
> fp.BaseUrl = filepath.Join("public/plugins", fp.Id)
> ```

@AgnesToulet Oh that's strange, maybe I should try to reproduce. Do I need to do anything special to reproduce on Windows?

AgnesToulet:
Nothing special, I just checked out your branch and tried to access one of the provisionned data source.

aknuds1:
@AgnesToulet Please have another look. AFAICT both those fields should have forward slashes.

AgnesToulet (approved):
Works on Windows! LGTM!

papagian:
For context: `fp.Module` is the JS import path imported here:
https://github.com/grafana/grafana/blob/3b9523fad7304c1e0abf9e87dfa002d30e54dcbf/public/app/features/plugins/plugin_loader.ts#L181-L191
 and `fp.BaseUrl` is a url so both should use '/' regardless of the running OS.

PR #28550, opened 2020-10-26
MetricFindQuery: Chore - fix the type for MetricFindValue

### LINKED ISSUE #28501: metricFindQuery - allow returning key/value pairs for the select
metricFindQuery only allows for returning an array of {text}. Allow array of {text, value}, which will in turn set the display value and value prop on the select option.

**Why is this needed**:
A typical case with selects is to set a display value with an underlying value/key that is used as the actual filter value.  In plugins this causes more effort/workarounds to use query type template variables as filters.

Same problem with ad hoc variables.

scottlepp:
The type for metricFindValue should include an optional value prop.

Fixes #28501

### DIFF
--- a/packages/grafana-data/src/types/datasource.ts
+++ b/packages/grafana-data/src/types/datasource.ts
@@ -499,6 +499,7 @@ export interface QueryHint {
 
 export interface MetricFindValue {
   text: string;
+  value?: string | number;
   expandable?: boolean;
 }
 


hugohaggmark:
Scott, could you change the merge to branch to master instead?

packages/grafana-data/src/types/datasource.ts
```diff
@@ -499,6 +499,7 @@ export interface QueryHint {
 
 export interface MetricFindValue {
   text: string;
+  value?: string | number;
```
hugohaggmark, line 502: Maybe we should just go with any here, WDYT?

hugohaggmark (approved):
LGTM, I left some nits for consideration

packages/grafana-data/src/types/datasource.ts
scottlepp: I really think string or number will be the only values that you would use for a select option value.  I don't think an object would work and possibly cause confusion.

scottlepp:
@hugohaggmark can I just merge here and cherry pick this to master?

hugohaggmark:
> @hugohaggmark can I just merge here and cherry pick this to master?

I don't know, seems like the wrong way around to me

scottlepp:
The problem is I branched off the release branch.  So I will need to branch off master.

hugohaggmark:
Gotcha, I don't know what the correct approach is here because v.7.3.x is our release branch but this change should be ok. So cherry-picking to master should be ok.

Issue #3441, opened 2015-12-06
ElasticSearch Query Editor - Scripted Metric Aggregation


shanielh:
I'm using a custom script aggregation in my elastic queries in Grafana 2.5, I can see why the "manual editor" has gone in 2.6 : I had to add the right level of aggregation in the "customized" editor and then move to the manual and insert my script (with the right aggregation names ("2", "1", etc')).

I used the manual editor just to replace the last aggregation level with a scripted metric aggregation. 

I think I'll manage to add a `scripted metric` aggregation with a button for settings that will allow to insert the four fields : `init_script`, `map_script`, `combine_script` and `reduce_script` inside a dialog.

What do you think?

Also, It might be nice to add a `custom` aggregation that will allow to insert the whole aggregation definition as a block, For aggregations that grafana doesn't support at the moment.

1tft:
That sounds really good. A section for a custom aggregation enables the user to do "everything" and he is not limited by a "simple" query editor. That seperates a pro tool from the rest (-;

aocenas:
There is feature request to reintroduce the raw query mode https://github.com/grafana/grafana/issues/3163.

PR #28554, opened 2020-10-27
Elasticsearch: Add Scripted Metric Aggregation

euniceek:

This PR adds support for Scripted Metrics Aggregation in Grafana’s Elasticsearch query editor. This option will give users the capability to define and execute their own metric aggregation using init/map/combine/reduce scripts, ultimately allowing more flexibility when calculating metrics. 

A new option **Scripted Metric** will be added under Metric in the Elasticsearch Query Editor. Users can fill out the four script parameters (init_script, map_script, combine_script, reduce_scrip) to get the desired metric output.  
Depending on the Elasticsearch version, different script parameters are required. All Elasticsearch version before v6.8 only requires map_script to be filled. However, from Elasticsearch v7.0, only init_script is optional, while all other script parameters are required. A pop-over will be presented in the user interface to clearly notify which script fields are optional depending on the version being used. 

Screen Shot 2020-10-23 at 6 36 52 PM

The current structure of the plugin is such that queries are sent to Elasticsearch whenever the query editor is updated, with query validation left as Elasticsearch’s responsibility. Query errors are returned from Elasticsearch and displayed to the user. There are two categories of errors that the user should be able to receive from the scripted metric aggregation fields: the user is either missing a required field, or the scripts failed at compile-time or runtime in Elasticsearch. Note that since queries are sent on every update, the user will immediately receive the missing field requirement error upon selecting scripted metric aggregation. This is intentional and consistent with the behaviour of the Bucket Script option. 

Resolves Grafana [issue 3441](https://github.com/grafana/grafana/issues/3441)

cc: @alolita

### DIFF
--- a/docs/sources/datasources/elasticsearch.md
+++ b/docs/sources/datasources/elasticsearch.md
@@ -123,6 +123,10 @@ Some metric aggregations are called Pipeline aggregations, for example, *Moving 
 
 ![](/img/docs/elasticsearch/pipeline_metrics_editor.png)
 
+## Scripting Language
+
+From Elasticsearch v5.0 and above, the default scripting language changed to *Painless*. For this scripting language, variables are referenced using params.<var>. For Elasticsearch version before v5.0, the default scripting language is *Groovy*, and variables are referenced using <var>. 
+
 ## Templating
 
 Instead of hard-coding things like server, application and sensor name in your metric queries you can use variables in their place.
--- a/docs/sources/datasources/elasticsearch.md
+++ b/docs/sources/datasources/elasticsearch.md
@@ -123,10 +123,10 @@ Some metric aggregations are called Pipeline aggregations, for example, *Moving 
 
 ![](/img/docs/elasticsearch/pipeline_metrics_editor.png)
 
-## Scripting Language
+## Scripting language
 
-Elasticsearch v5.0 and above uses *Painless* as the default scripting language. Variables are referenced using `params.<var>`. 
-Elasticsearch pre-v5.0 uses *Groovy* as the default scripting language if not changed. For *Groovy*, `<var>` is used to reference a variable.
+Elasticsearch v5.0 and above uses Painless as the default scripting language. Variables are referenced using `params.<var>`. 
+Elasticsearch pre-v5.0 uses Groovy as the default scripting language if not changed. For Groovy, `<var>` is used to reference a variable.
 
 ## Templating
 
--- a/docs/sources/datasources/elasticsearch.md
+++ b/docs/sources/datasources/elasticsearch.md
@@ -125,7 +125,8 @@ Some metric aggregations are called Pipeline aggregations, for example, *Moving 
 
 ## Scripting Language
 
-From Elasticsearch v5.0 and above, the default scripting language changed to *Painless*. For this scripting language, variables are referenced using params.<var>. For Elasticsearch version before v5.0, the default scripting language is *Groovy*, and variables are referenced using <var>. 
+Elasticsearch v5.0 and above uses *Painless* as the default scripting language. Variables are referenced using params.<var>. 
+Elasticsearch pre-v5.0 uses *Groovy* as the default scripting language if not changed. For *Groovy*, <var> is used to reference a variable.
 
 ## Templating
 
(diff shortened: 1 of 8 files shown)


zoltanbedi:
Hi @eunice98k, thanks for opening this pr. Code wise this looks good to me however kibana recently closed their issue to support this https://github.com/elastic/kibana/issues/2646 because of Elastics discourage using it https://github.com/elastic/elasticsearch/issues/63096 What do you think? Is this something you use a lot?

docs/sources/datasources/elasticsearch.md
```diff
@@ -123,6 +123,11 @@ Some metric aggregations are called Pipeline aggregations, for example, *Moving
 
 ![](/img/docs/elasticsearch/pipeline_metrics_editor.png)
 
+## Scripting Language
+
+Elasticsearch v5.0 and above uses *Painless* as the default scripting language. Variables are referenced using `params.<var>`.
```
oddlittlebird, line 128: Please remove the italics from these sections.

Also, all headings should be sentence case, so lowercase "language"in the h2.

achatterjee-grafana (commented):
Doc content looks OK, dev will need to approve at the code portion.

Cyberness:
> because of Elastics discourage using it

Elastic does not discourage using of Scripted Metrics Aggregation, what they discourage using is a Groove based script to implement scripted Metrics which (Groovy script) is indeed deprecated and recommendation is to use [Painless](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html) script instead. But the feature itself (Scripted Metrics) is not discourage by any means.

Cyberness:
> Hi @eunice98k, thanks for opening this pr. Code wise this looks good to me however kibana recently closed their issue to support this [elastic/kibana#2646](https://github.com/elastic/kibana/issues/2646) because of Elastics discourage using it [elastic/elasticsearch#63096](https://github.com/elastic/elasticsearch/issues/63096) What do you think? Is this something you use a lot?

`For these reasons, we wish to discourage our own stack features from using scripted-metric.` So, they basically just moving it from core of ES, not completely discourage using it. There are many cases when scripted metrics are useful because they provide more elegant or the only solution for a problem. Not supporting that kind of feature in Grafana just because someone cannot use it properly will force users to search for alternative tool that supports it. The fact that Kibana decided to artificially limit themselves on that regard basically means that people will search for alternative and can find it in Grafana. I think this is good.

eunice98k:
> > Hi @eunice98k, thanks for opening this pr. Code wise this looks good to me however kibana recently closed their issue to support this [elastic/kibana#2646](https://github.com/elastic/kibana/issues/2646) because of Elastics discourage using it [elastic/elasticsearch#63096](https://github.com/elastic/elasticsearch/issues/63096) What do you think? Is this something you use a lot?
> 
> `For these reasons, we wish to discourage our own stack features from using scripted-metric.` So, they basically just moving it from core of ES, not completely discourage using it. There are many cases when scripted metrics are useful because they provide more elegant or the only solution for a problem. Not supporting that kind of feature in Grafana just because someone cannot use it properly will force users to search for alternative tool that supports it. The fact that Kibana decided to artificially limit themselves on that regard basically means that people will search for alternative and can find it in Grafana. I think this is good.

Thank you for the feedbacks! I agree that this will give Grafana an advantage on this matter as there has been a lot of interest from users of this feature in Kibana. Adding onto this, a more specific use case where only scripted metrics can be used is ones such as counting numbers of overlapping minute-intervals from hourly documents stored. There is currently no other way to perform this calculation without stateful UDAF.

torkelo:
I think since Elastic discourage usage of scripted metric aggregations and they closed this feature request in Kibana we should do the same

spikhterev:
> I think since Elastic discourage usage of scripted metric aggregations and they closed this feature request in Kibana we should do the same

Having this feature merged will allow Grafana users to leverage Elasticsearch scripted metric aggregation for some advanced use cases. Without this feature, we currently were needed to use JSON Data Source plugin plus develop our own custom Elasticsearch API service to proxy and transform requests and parse the responses between the data source and Elasticsearch. This works, but adds additional limitations, like the inability to use Elastcisearch ad-hoc filters. 
Kibana team's reasoning to discontinue support of scripted metric aggregation sounds weird to us as end-users. Using the same logic they could drop some other existing features, like painless scripts support etc. as they could be adding some complexity. 
However, any awesome tool, which I believe Grafana is, should be allowing some advanced use cases and this could be an advantage of Grafana.

PR #28555, opened 2020-10-27
Use fetch API in InfluxDB data source

dprokop:
Replaces usages of datasourceRequest in InfluxDB data source in favor of fetch API.

Ref https://github.com/grafana/grafana/issues/27222

### DIFF
--- a/public/app/plugins/datasource/influxdb/datasource.ts
+++ b/public/app/plugins/datasource/influxdb/datasource.ts
@@ -22,7 +22,7 @@ import { InfluxQueryBuilder } from './query_builder';
 import { InfluxQuery, InfluxOptions, InfluxVersion } from './types';
 import { getTemplateSrv, TemplateSrv } from 'app/features/templating/template_srv';
 import { getBackendSrv, DataSourceWithBackend, frameToMetricFindValue } from '@grafana/runtime';
-import { Observable, from, throwError, of } from 'rxjs';
+import { Observable, throwError, of } from 'rxjs';
 import { FluxQueryEditor } from './components/FluxQueryEditor';
 import { catchError, map } from 'rxjs/operators';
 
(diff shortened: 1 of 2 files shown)


public/app/plugins/datasource/influxdb/datasource.ts
```diff
@@ -151,54 +153,56 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
...
-      const seriesList = [];
-      for (i = 0; i < data.results.length; i++) {
-        const result = data.results[i];
-        if (!result || !result.series) {
-          continue;
+    return this._seriesQuery(allQueries, options)
+      .toPromise()
+      .then((data: any): any => {
+        if (!data || !data.results) {
+          return [];
```
hugohaggmark, line 160: I think classicQuery should return an Observable instead and then we could remove __from__ creator here https://github.com/grafana/grafana/blob/316a7997a53882fe6acb37f4cffa0757bb674012/public/app/plugins/datasource/influxdb/datasource.ts#L80

and then we don't need to do `toPromise` here. WDTY?

public/app/plugins/datasource/influxdb/datasource.ts
```diff
@@ -459,46 +468,59 @@ export default class InfluxDatasource extends DataSourceWithBackend<InfluxQuery,
...
-                throw {
-                  message: 'InfluxDB Error: ' + errors[0].error,
-                  data,
-                };
+                return throwError(
+                  this.handleErrors({
+                    message: 'InfluxDB Error: ' + errors[0].error,
+                    data,
+                  })
+                );
```
hugohaggmark, line 486: This is returning an Observable but the other path is returning an array. Maybe use a simple `throw` instead?

hugohaggmark (commented):
Skilled work on this! I left some considerations I think could improve this.

dprokop:
@hugohaggmark thank you for your review. I've applied the changes you suggested!

hugohaggmark (approved):
LGTM!

PR #28558, opened 2020-10-27
Chore: use jest without grunt

zoltanbedi:

Using grunt in CI makes the output not readable for frontend tests. This pr leaves out grunt and uses jest instead. One thing that is not going to run is [sasslint](https://github.com/sasstools/sass-lint) which I think is not a big pain as we don't use sass files anymore for new components.

Maybe it would make sense to remove the commands and the grunt dependencies but I would do it altogether when the last grunt command is migrated as well.

This is also skipping now `no-focus-convey-tests` and `no-only-tests` @aocenas is this something we want to check in the future?


package.json
```diff
@@ -14,7 +14,7 @@
     "e2e:dev": "./e2e/start-and-run-suite dev",
     "jest": "jest --notify --watch",
     "lint": "eslint . --ext .js,.tsx,.ts --cache",
-    "jest-ci": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit -w ${TEST_MAX_WORKERS:-100%}",
+    "test": "mkdir -p reports/junit && export JEST_JUNIT_OUTPUT_DIR=reports/junit && jest --ci --reporters=default --reporters=jest-junit -w ${TEST_MAX_WORKERS:-100%}",
```
aknuds1, line 17: Are you sure you want the "test" script to use CI parameters? For example this script passes `--ci` to jest and makes a report directory. I'm not sure that the script should be the same for local testing and CI.

aknuds1 (commented):
In addition to my code comment, I am wondering if the "jest" target should be removed from scripts/grunt/options/exec.js?

package.json
zoltanbedi: Maybe I will do a poll about this but personally I've never used the test command locally. I used the jest command instead. Maybe change this back to jest-ci and rename the jest command to test?

aknuds1: I think it's best to follow the convention of having the -ci suffix for CI scripts. If you don't have a non-ci script for local usage, that's fine. I think it doesn't matter if it's called jest-ci or test-ci FWIW.

aknuds1 (approved):
LGTM

package.json
```diff
@@ -12,9 +12,9 @@
     "e2e": "./e2e/start-and-run-suite",
     "e2e:debug": "./e2e/start-and-run-suite debug",
     "e2e:dev": "./e2e/start-and-run-suite dev",
-    "jest": "jest --notify --watch",
+    "test": "jest --notify --watch",
```
aocenas, line 15: I would probably do something like `test` and `test:watch`. Personally I usually do `yarn test` just to make sure there are no errors in other parts of code but locally I just run the tests in the IDE and as. With `jest --watch` no tests are run until something changes which usually isn't what I personally want.

aocenas:
> This is also skipping now no-focus-convey-tests and no-only-tests @aocenas is this something we want to check in the future?

So I definitely think these are important. Basically if you do `it.only()` in jest tests it skips all other tests and nothing is run anymore while giving you green both locally and in CI which can be hard to notice (we had that issue already where for few days no tests were run in CI and nobody noticed)

zoltanbedi:
> > This is also skipping now no-focus-convey-tests and no-only-tests @aocenas is this something we want to check in the future?
> 
> So I definitely think these are important. Basically if you do `it.only()` in jest tests it skips all other tests and nothing is run anymore while giving you green both locally and in CI which can be hard to notice (we had that issue already where for few days no tests were run in CI and nobody noticed)

We can bring back the no only tests fairly easily with [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest/) (I know I love these plugins 😄) but what about the other one?

aocenas:
@zoltanbedi the other one is the same thing just for Go. You have the same issue there as you can do something similar with Convey tests (just different syntax)

aocenas:
But I think those can just be a separate steps in the pipeline and still run through grunt or something like that without much changes. Just put the jest tests outside of Grunt runner

contribute/developer-guide.md
```diff
@@ -87,7 +87,7 @@ The test suite consists of three types of tests: _Frontend tests_, _backend test
 We use [jest](https://jestjs.io/) for our frontend tests. Run them using Yarn:
 
 ```
-yarn jest
+yarn test
```
oddlittlebird, line 90: I think it was funnier before the change.

package.json
zoltanbedi: you can still run all tests by pressing `a`

aocenas (approved):

.drone.yml
```diff
@@ -66,6 +66,7 @@ steps:
 - name: test-backend
   image: grafana/build-container:1.2.28
   commands:
+  - yarn grunt no-focus-convey-tests
```
aknuds1, line 69: What does `yarn grunt no-focus-convey-tests` do? Why should it run for backend tests?

aknuds1: I think we can replace it with this:
```suggestion
  - [[ $(grep FocusConvey -R pkg | wc -l) -eq "0" ]] || exit 1
```

scripts/lib.star
```diff
@@ -441,7 +441,9 @@ def test_backend_step():
             'lint-backend',
         ],
         'commands': [
-            # First execute non-integration tests in parallel, since it should be safe
+            # First make sure that there are no tests with FocusConvey
+            'yarn grunt no-focus-convey-tests',
```
aknuds1, line 445: I think it's better if we avoid Yarn and Grunt :D
```suggestion
            '[[ $(grep FocusConvey -R pkg | wc -l) -eq "0" ]] || exit 1',
```

aknuds1 (approved):

scripts/lib.star
```diff
@@ -441,7 +441,9 @@ def test_backend_step():
             'lint-backend',
         ],
         'commands': [
-            # First execute non-integration tests in parallel, since it should be safe
+            # First make sure that there are no tests with FocusConvey
+            '[[ $(grep FocusConvey -R pkg | wc -l) -eq "0" ]] || exit 1',
```
aknuds1, line 445: This shell syntax should work under Bourne shell I think:
```suggestion
            '[ $(grep FocusConvey -R pkg | wc -l) -eq "0" ] || exit 1',
```

PR #28559, opened 2020-10-27
Developer guide: Update wrt. Windows

aknuds1:
Update developer guide with better instructions on how to build on Windows. I found out that building with Make isn't really an option on Windows, unless you use e.g. Git Bash. Running tests directly (`go test`) also doesn't currently work due to a linking error, so I instruct Windows users to go via build.go (`go run build.go test`).

### DIFF
--- a/contribute/developer-guide.md
+++ b/contribute/developer-guide.md
@@ -78,8 +78,8 @@ When you log in for the first time, Grafana asks you to change your password.
 
 The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
 
-In order to use the `make` command above, you need to install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
-If you don't want to install these and just want to build the back-end, you can do it as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
+You can simply build the back-end as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
+Alternately, if you wish to use the `make` command, install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
 
 ## Test Grafana
 
--- a/contribute/developer-guide.md
+++ b/contribute/developer-guide.md
@@ -78,7 +78,8 @@ When you log in for the first time, Grafana asks you to change your password.
 
 The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
 
-Unlike on Unix, unless you use a Unix shell (f.ex. Git Bash), you shouldn't use Make on Windows. Instead build the back-end as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
+In order to use the `make` command above, you need to install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
+If you don't want to install these and just want to build the back-end, you can do it as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
 
 ## Test Grafana
 
--- a/contribute/developer-guide.md
+++ b/contribute/developer-guide.md
@@ -76,7 +76,9 @@ When you log in for the first time, Grafana asks you to change your password.
 
 #### Building on Windows
 
-The Grafana backend includes Sqlite3 which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download).
+The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
+
+Unlike on Unix, unless you use a Unix shell (f.ex. Git Bash), you shouldn't use Make on Windows. Instead build the back-end as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
 
 ## Test Grafana
 
@@ -96,6 +98,13 @@ If you're developing for the backend, run the tests with the standard Go tool:
 
 ```
 go test -v ./pkg/...
+```
+
+#### On Windows
+Running the backend tests on Windows currently needs some tweaking, so use the build.go script:
+
+```
+go run build.go test
 ```
 
 ### Run end-to-end tests


contribute/developer-guide.md
```diff
@@ -76,7 +76,9 @@ When you log in for the first time, Grafana asks you to change your password.
 
 #### Building on Windows
 
-The Grafana backend includes Sqlite3 which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download).
+The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
+
+Unlike on Unix, unless you use a Unix shell (f.ex. Git Bash), you shouldn't use Make on Windows. Instead build the back-end as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
```
AgnesToulet, line 81:
```suggestion
In order to use the `make` command above, you need to install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
If you don't want to install these and just want to build the back-end, you can do it as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
```

AgnesToulet (commented):
I think `make` is pretty useful if you work on Grafana so I think we should talk about the two solutions with the same amount of information. I added some useful information for developers that would want to make the `make` commands work on Windows.

Also I don't have any trouble to run the tests both on Git Bash and PowerShell but I guess it's related to the issue that I can't replicate. Maybe the test part should be renamed "Troubleshooting tests on Windows", with an explication of which issue using the build.go script fixes?

aknuds1:
> I think `make` is pretty useful if you work on Grafana so I think we should talk about the two solutions with the same amount of information. I added some useful information for developers that would want to make the `make` commands work on Windows.

@AgnesToulet Thanks, I committed it.

> Also I don't have any trouble to run the tests both on Git Bash and PowerShell but I guess it's related to the issue that I can't replicate. Maybe the test part should be renamed "Troubleshooting tests on Windows", with an explication of which issue using the build.go script fixes?

My thinking was that since any build problems can change with time (the current issue might get fixed, and another might turn up), it would be a more future proof solution to ask people to test via build.go than to explain the current issue (the documentation will quickly go stale).

contribute/developer-guide.md
```diff
@@ -76,7 +76,10 @@ When you log in for the first time, Grafana asks you to change your password.
 
 #### Building on Windows
 
-The Grafana backend includes Sqlite3 which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download).
+The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
```
achatterjee-grafana, line 79: Change to: The Grafana backend includes SQLite which requires GCC to compile. As a result, to compile Grafana on Windows, you must intsall GCC. We recommend using [TDM-GCC](http://tdm-gcc.tdragon.net/download). If you use [Scoop](https://scoop.sh), you can install GCC through that.

contribute/developer-guide.md
```diff
@@ -76,7 +76,10 @@ When you log in for the first time, Grafana asks you to change your password.
...
+
+In order to use the `make` command above, you need to install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash).
```
achatterjee-grafana, line 81: In order to use the `make` command, install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash).

achatterjee-grafana (commented):
Added a few copy-edit suggestions.

contribute/developer-guide.md
oddlittlebird: FYI, I got a suspicious link warning when I clicked the link for Make for Windows.

Also, that program is from 2006. Do we really want to be recommending this to people?

aknuds1: @oddlittlebird To my knowledge, that's the only alternative for using Make on Windows.

aknuds1: @achatterjee-grafana I think my original wording is preferred because we don't want to instruct people to use Make, but inform them of _how_ they can do it if they wish. It's an alternative, not a requirement.

contribute/developer-guide.md
aknuds1: Why is "We recommend using TDM-GCC" better than "We recommend TDM-GCC"? I don't think it reads any better.

contribute/developer-guide.md
ChrisDGH: It would seem the link sends the browser to 'http' instead of 'https' which doesn't resolve.  It would seem to be a browser certificate warning.

aknuds1: Interesting that this page still doesn't have HTTPS. Will that have any security implications though for the intended use, i.e. to download Make? @ChrisDGH

ChrisDGH: When clicking on the Make for Windows, it goes to the http domain.  In there, when clicking on the dependency zip files, it forward to https://sourceforge.net/projects/gnuwin32/files/make/3.81/make-3.81-dep.zip/download which is https.  From a security perspective, the http may warn in the browser, but the links forward to https.

achatterjee-grafana: > @achatterjee-grafana I think my original wording is preferred because we don't want to instruct people to use Make, but inform them of _how_ they can do it if they wish. It's an alternative, not a requirement. Your suggestion looks more instructive to me, and thus more confusing than the current wording.
> 
> For example, I personally don't use Make on Windows to work with Grafana, but others (f.ex. @AgnesToulet) do.

@aknuds1, although I do not agree with you,  I will let this go. Product documentation is supposed to be instructive :)

aknuds1: This is not product documentation though, it's a guide for developers. Development isn't a straightforward procedure, there are many ways of arriving at a result.

contribute/developer-guide.md
```diff
@@ -76,7 +76,10 @@ When you log in for the first time, Grafana asks you to change your password.
 
 #### Building on Windows
 
-The Grafana backend includes Sqlite3 which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download).
+The Grafana backend includes SQLite which requires GCC to compile. So in order to compile Grafana on Windows you need to install GCC. We recommend [TDM-GCC](http://tdm-gcc.tdragon.net/download). Eventually, if you use [Scoop](https://scoop.sh), you can install GCC through that.
+
+In order to use the `make` command above, you need to install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
+If you don't want to install these and just want to build the back-end, you can do it as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
```
AgnesToulet, line 82:
```suggestion
You can simply build the back-end as follows: `go run build.go build`. The Grafana binaries will be in bin\\windows-amd64.
Alternately, if you wish to use the `make` command, install [Make for Windows](http://gnuwin32.sourceforge.net/packages/make.htm) and use it in a Unix shell (f.ex. Git Bash). 
```
I like @achatterjee-grafana suggestion, what about something like above? We don't recommend to run the `make` command but provide information if the user wants to do so.

aknuds1: @AgnesToulet thanks, this edit makes sense to me. Applied.

AgnesToulet (approved):
LGTM!

contribute/developer-guide.md
oddlittlebird: @aknuds1 , if you install a Linux subsystem on your Windows machine, then you can install and run make. I use the Ubuntu subsystem and have been able to run some make commands in the past. (That said, they also have occasionally failed due to permissions problems, but that could be because I'm a Linux noob.)

aknuds1: I haven't used the Linux subsystem so much on Windows, but it's essentially a VM isn't it, so you're actually on Linux? When you're doing that, you're not technically doing Windows development, so the Windows instructions don't apply as far as I understand.

Issue #21516, opened 2020-01-15
Backend plugins: Implement support for validating backend plugin config 


marefr:
Implement support for validating backend plugin config before saving app/datasource backend plugin settings. Any error/warning should be returned to client. If it fails settings should not be saved in database.

Blocked by: grafana/grafana-plugin-sdk-go#37
Epic: #23992

PR #28560, opened 2020-10-27
WIP: Backend: Introduce update handler mechanism

kaydelaney:
Creating this PR  purely as a starting point for discussion.
As far as I can tell, there is currently no way to execute an action on the backend in response to a datasource being updated, something which would be useful for caching pieces of data that rely on data source properties. For example in the case of the CloudWatch data source, it would be useful to fetch and cache an account ID when the corresponding datasource is updated.
Mainly I want to see if the approach implemented in this PR is the way to go about solving this issue, and would love to hear the perspective of those with more backend knowledge.

### DIFF
--- a/pkg/api/datasources.go
+++ b/pkg/api/datasources.go
@@ -13,8 +13,13 @@ import (
 	"github.com/grafana/grafana/pkg/models"
 	"github.com/grafana/grafana/pkg/plugins"
 	"github.com/grafana/grafana/pkg/plugins/datasource/wrapper"
+	"github.com/grafana/grafana/pkg/tsdb"
 	"github.com/grafana/grafana/pkg/util"
 )
+
+type UpdateHandler interface {
+	OnUpdate(dsInfo *models.DataSource, cmd *models.UpdateDataSourceCommand) error
+}
 
 var datasourcesLogger = log.New("datasources")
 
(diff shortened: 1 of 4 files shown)


aknuds1:
Interesting, thanks for getting the ball rolling on this :)

torkelo (commented):
this looks interesting, would be important to stress that onUpdate would not be called on all instances (in HA setup) so could not be used to clear a caches (unless store in data source table like is done here)

marefr:
The changes are somewhat in line with something I have thought about for the plugin SDK. Basically adding support for each data source to validate and/or provide defaults before a data source is saved/updated. Something I'm planning to look into soon hopefully.

@torkelo the idea is to provide additional data for the data source stored in the database to be used in upcoming requests. So will be HA-compatible I think. But we currently don't have HA support for live either.

marefr:
Here's the existing issue with links to another issue in the sdk: https://github.com/grafana/grafana/issues/21516

ryantxu:
interesting -- I would love a world where plugins are told their configs infrequently and when they change, then respond to requests.  The current model sends *everything* in every request (and includes a flag saying when it last changed)

 #21516 is just the beginning -- it would offer a method to ask if a new config is OK, but I don't see how that would affect the rest of the infra

marefr:
> The current model sends everything in every request (and includes a flag saying when it last changed)

Correct, that's the current approach. Makes it simple to reason about and works as expected in HA setup's without further effort. I guess you're thinking about use cases where a background task is running in a backend plugin?  We maybe have to do something eventually to support this use case better. 

> #21516 is just the beginning -- it would offer a method to ask if a new config is OK, but I don't see how that would affect the rest of the infra

It could handle migrations and also return default values and/or additional config to be saved in the database.

PR #28561, opened 2020-10-27
MetricFindQuery: Chore - Add value prop

### LINKED ISSUE #28501: metricFindQuery - allow returning key/value pairs for the select

scottlepp:
Add missing value prop to MetricFindValue type

Fixes #28501

### DIFF
--- a/packages/grafana-data/src/types/datasource.ts
+++ b/packages/grafana-data/src/types/datasource.ts
@@ -499,6 +499,7 @@ export interface QueryHint {
 
 export interface MetricFindValue {
   text: string;
+  value?: string | number;
   expandable?: boolean;
 }
 


hugohaggmark (approved):
LGTM!

tobiasgrabmeier:
Any idea, when this fix will be in one of the releases?

hugohaggmark:
> Any idea, when this fix will be in one of the releases?

@tobiasgrabmeier this should go into the 7.4.0 release

PR #28564, opened 2020-10-27
Migration: Remove LegacyForms from dashboard folder permissions

### LINKED ISSUE #25841: Chore: Remove LegacyForms from Grafana core
A followup on #22862

After the new forms were implemented we were left with a few places where we still used the LegacyForms. These need to go.

[See the current amount of LegacyForms in use](https://play.grafana.org/d/VEy5UXjmk/grafana-ci-code-stats?orgId=1)

To find where they are used, simply search `LegacyForms` in the codebase. Some of these might need some further UX work though. To handle possible API changes, have a look at Forms in [@grafana/ui](https://developers.grafana.com/ui/canary/index.html?path=/story/*)

- [x] Dashboard folder permissions #28564 
- [x] Dashboard feature #28707

There are still some (internal) components in @grafana/ui that does not import LegacyForms, but rather import the sepaate components as a relative path. These are the components:

- [ ] ClipboardButton
- [ ] TagsInput (Where is this even used?)
- [ ] ColorPicker (The input field and Y-axis Switch)

These need more extensive UX/UI work
- [ ] DataSourceHttpSettings
- [ ] CustomHeadersSettings

jackw:
This PR removes the usage of LegacyForms from Dashboard Folder Permissions view. It also restricts the line-height of the grafana-ui `<Select>` component option descriptions as wrapping components / css classes were affecting it.

Screenshot 2020-10-27 at 13 09 35

### DIFF
--- a/packages/grafana-ui/src/components/Select/getSelectStyles.ts
+++ b/packages/grafana-ui/src/components/Select/getSelectStyles.ts
@@ -43,6 +43,7 @@ export const getSelectStyles = stylesFactory((theme: GrafanaTheme) => {
       font-size: ${theme.typography.size.sm};
       color: ${theme.colors.textWeak};
       white-space: normal;
+      line-height: ${theme.typography.lineHeight.md};
     `,
     optionBody: css`
       label: grafana-select-option-body;
--- a/public/app/core/components/PermissionList/AddPermission.tsx
+++ b/public/app/core/components/PermissionList/AddPermission.tsx
@@ -63,7 +63,6 @@ class AddPermissions extends Component<Props, NewDashboardAclItem> {
   };
 
   onPermissionChanged = (permission: SelectableValue<PermissionLevel>) => {
-    console.log(permission);
     this.setState({ permission: permission.value! });
   };
 
@@ -87,7 +86,7 @@ class AddPermissions extends Component<Props, NewDashboardAclItem> {
     const newItem = this.state;
     const pickerClassName = 'min-width-20';
     const isValid = this.isValid();
-    console.log(newItem);
+
     return (
       <div className="cta-form">
         <button className="cta-form__close btn btn-transparent" onClick={onCancel}>
@@ -124,7 +123,6 @@ class AddPermissions extends Component<Props, NewDashboardAclItem> {
                   value={this.state.permission}
                   options={dashboardPermissionLevels}
                   onChange={this.onPermissionChanged}
-                  isOpen
                   width={25}
                 />
               </InlineField>
--- a/public/app/core/components/PermissionList/AddPermission.tsx
+++ b/public/app/core/components/PermissionList/AddPermission.tsx
@@ -1,8 +1,10 @@
 import React, { Component } from 'react';
+import { css } from 'emotion';
+import config from 'app/core/config';
 import { UserPicker } from 'app/core/components/Select/UserPicker';
 import { TeamPicker, Team } from 'app/core/components/Select/TeamPicker';
-import { Button, Form, InlineField, InlineFieldRow, Icon, Select } from '@grafana/ui';
-import { SelectableValue } from '@grafana/data';
+import { Button, Form, HorizontalGroup, Icon, Select, stylesFactory } from '@grafana/ui';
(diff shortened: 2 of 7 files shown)


dprokop (approved):
Looks good @jackw!

packages/grafana-ui/src/components/Select/getSelectStyles.ts
```diff
@@ -43,6 +43,7 @@ export const getSelectStyles = stylesFactory((theme: GrafanaTheme) => {
       font-size: ${theme.typography.size.sm};
       color: ${theme.colors.textWeak};
       white-space: normal;
+      line-height: ${theme.typography.lineHeight.md};
```
torkelo, line 46: is this not the default body line height?

public/app/core/components/PermissionList/AddPermission.tsx
```diff
@@ -88,56 +86,52 @@ class AddPermissions extends Component<Props, NewDashboardAclItem> {
...
+                  <UserPicker onSelected={this.onUserSelected} className={pickerClassName} />
+                </InlineField>
+              ) : null}
+
+              {newItem.type === AclTarget.Team ? (
+                <InlineField>
+                  <TeamPicker onSelected={this.onTeamSelected} className={pickerClassName} />
+                </InlineField>
+              ) : null}
+
```
torkelo, line 119: A label here that said "Can"  would be nice, to make it similar to table and not just be 3 selects in a row

packages/grafana-ui/src/components/Select/getSelectStyles.ts
jackw: Yep, that's correct. I guess it's more of a reset. My reasoning for adding it was due to `.filter-table` class used on the permissions list declares [a line-height](https://github.com/grafana/grafana/blob/master/public/sass/components/_filter-table.scss#L33). I considered reworking the table styles but thought it better if the dropdown descriptions looked consistent wherever they're used.

I spoke with @jessover9000 before adding it and we came to the conclusion that of the lineHeight theme values we have right now this one looks the best.

public/app/core/components/PermissionList/DisabledPermissionListItem.tsx
```diff
@@ -28,7 +27,7 @@ export default class DisabledPermissionListItem extends Component<Props, any> {
             <Select
               options={dashboardPermissionLevels}
               onChange={() => {}}
-              isDisabled={true}
+              disabled={true}
               className="gf-form-select-box__control--menu-right"
```
Clarity-89, line 31: Could we maybe also get rid of this class here? Doesn't seems like it has any effect.

packages/grafana-ui/src/components/Select/getSelectStyles.ts
torkelo: ok, yea that filter table line-height is probably not the best solution, but I see why this is needed now

public/app/core/components/PermissionList/AddPermission.tsx
```diff
@@ -88,59 +88,57 @@ class AddPermissions extends Component<Props, NewDashboardAclItem> {
...
+              </Button>
+            </HorizontalGroup>
+          )}
+        </Form>
       </div>
     );
   }
 }
 
+const getStyles = stylesFactory((theme: GrafanaTheme) => ({
```
dprokop, line 137: FYI: you don't need to use styles factory. You can define the style getter and use [`useStyles` hook](https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md#styling-with-theme) that will provide the theme automatically and guarantee memoization

dprokop (approved):

public/app/core/components/PermissionList/AddPermission.tsx
jackw: Maybe I'm misunderstanding but to the best of my knowledge hooks can't be used in class components. Did I miss something? 🤔

dprokop: Nope, it we me missing that this component is a class one. Ignore my comment :)

PR #28565, opened 2020-10-27
Short URL: Update last seen at when visiting a short URL

marefr:
Update last seen at when visiting a short URL

Ref #28248

This is a prerequisite for adding cleanup job in the future so important to include in 7.3 stable.

### DIFF
--- a/pkg/api/short_url.go
+++ b/pkg/api/short_url.go
@@ -3,6 +3,7 @@ package api
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"path"
 	"strings"
 
@@ -17,6 +18,9 @@ func (hs *HTTPServer) createShortURL(c *models.ReqContext, cmd dtos.CreateShortU
 	hs.log.Debug("Received request to create short URL", "path", cmd.Path)
 
 	cmd.Path = strings.TrimSpace(cmd.Path)
+
+	u, err := url.Parse(cmd.Path)
+	hs.log.Info("Parsed url", "path", cmd.Path, "u", u, "err", err)
 
 	if path.IsAbs(cmd.Path) {
 		hs.log.Error("Invalid short URL path", "path", cmd.Path)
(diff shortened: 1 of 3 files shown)


pkg/api/short_url.go
```diff
@@ -57,6 +57,11 @@ func (hs *HTTPServer) redirectFromShortURL(c *models.ReqContext) {
 		return
 	}
 
+	// Failure to update last seen at should still allow to redirect
```
wbrowne, line 60: I think when referring to the field `LastSeenAt`, it would be better to write it out in its field form IE `LastSeenAt` or `last_seen_at`, just so it doesn't confuse the sentence in the comment or error log

pkg/api/short_url.go
```diff
@@ -57,6 +57,11 @@ func (hs *HTTPServer) redirectFromShortURL(c *models.ReqContext) {
...
+	if err := hs.ShortURLService.UpdateLastSeenAt(c.Req.Context(), shortURL); err != nil {
```
wbrowne, line 61: Might be a nice candidate for an async function but probably overkill for v1 🤷

ivanahuckova (approved):
LGTM! 🎈

pkg/services/shorturls/short_url_service.go
```diff
@@ -42,6 +44,18 @@ func (s ShortURLService) GetShortURLByUID(ctx context.Context, user *models.Sign
 	return &shortURL, nil
 }
 
+func (s ShortURLService) UpdateLastSeenAt(ctx context.Context, shortURL *models.ShortUrl) error {
+	shortURL.LastSeenAt = getTime().Unix()
+	return s.SQLStore.WithTransactionalDbSession(ctx, func(dbSession *sqlstore.DBSession) error {
+		_, err := dbSession.Id(shortURL.Id).Update(shortURL)
```
papagian, line 50:
```suggestion
		_, err := dbSession.ID(shortURL.Id).Update(shortURL)
```
for fixing the deprecation warning.

aknuds1 (approved):

pkg/api/short_url.go
marefr:
```suggestion
	// Failure to update LastSeenAt should still allow to redirect
```

pkg/api/short_url.go
marefr: Yeah probably, but good idea :smile:

papagian (approved):
I've tested and works as expected.

PR #28570, opened 2020-10-27
Allow add panel for viewers_can_edit

clyang82:
Signed-off-by: clyang82 <chuyang@redhat.com>

Viewer role can be changed with the Grafana server setting viewers_can_edit. If you set this to true, then users with the Viewer role can also make transient dashboard edits. right now with `viewers_can_edit`, it can modify panels but not save the changes. In this PR, propose to allow this role can create panel but not save the changes.


torkelo (approved):
LGTM

PR #28578, opened 2020-10-27
Dashboard: Fix navigation from one SoloPanelPage to another one

### LINKED ISSUE #28574: Navigating between SoloPanelPages requires reload
**What happened**: When viewing a SoloPanel, navigating to another SoloPanel does not work without a full page reload. The page freezes with the message "Loading & initializing dashboard".

Navigation without reload can occurr, for example, by creating an annotation on dashboard 1 containing a link to dashboard 2.

**What you expected to happen**: Navigating from one SoloPanelPage to another using the angularjs router should be possible

- Create two dashboards (d1 and d2) with one panel each
- Lookup the d-solo urls using the Panel->Share->Embed function
- Add annotation with content `<a href="http://localhost:3000/d-solo/uid2/d2?panelId=1">test</a>` to dashboard 1
- Navigate to http://localhost:3000/d-solo/uid1/d1?panelId=1
- Click the link in the annotation
- The graph vanishes and is stuck at "Loading & initializing dashboard"

**My investigation**:
Upon navigation, the `componentDidUpdate` method of the `SoloPanelPage` is called:
https://github.com/grafana/grafana/blob/aa9134b3b42c8678efe609a0339a616f3c566aaf/public/app/features/dashboard/containers/SoloPanelPage.tsx#L53
However, `prevProps.dashboard` still contains the old dashboard, so the panel is not updated after
https://github.com/grafana/grafana/blob/aa9134b3b42c8678efe609a0339a616f3c566aaf/public/app/features/dashboard/containers/SoloPanelPage.tsx#L61
I would insert a check whether the uid of the dashboard has changed and update the panel if that is the case.

**Environment**:
- Grafana version: 7.2.2
- Data source type & version: - 
- OS Grafana is installed on: WSL Ubuntu 18.04
- User OS & Browser: Windows 10 + MS Edge
- Grafana plugins: - 
- Others: -

hborchardt:

The panel on the SoloPanelPage did not update after an internal navigation to another SoloPanelPage. The reason is that, because the props apparently are retained, in the `componentDidUpdate` function it was assumed that the dashboard was already fully loaded.

Fixes #28574 

I also added a test case, as there was none for the SoloPanelPage. I hope that it is helpful

### DIFF
--- a/public/app/features/dashboard/containers/SoloPanelPage.tsx
+++ b/public/app/features/dashboard/containers/SoloPanelPage.tsx
@@ -57,8 +57,8 @@ export class SoloPanelPage extends Component<Props, State> {
       return;
     }
 
-    // we just got the dashboard!
-    if (!prevProps.dashboard) {
+    // we just got a new dashboard
+    if (!prevProps.dashboard || prevProps.dashboard.uid !== dashboard.uid) {
       const panelId = parseInt(urlPanelId, 10);
 
       // need to expand parent row if this panel is inside a row
--- a/public/app/features/dashboard/containers/SoloPanelPage.tsx
+++ b/public/app/features/dashboard/containers/SoloPanelPage.tsx
@@ -13,7 +13,7 @@ import { initDashboard } from '../state/initDashboard';
 import { StoreState, DashboardRouteInfo } from 'app/types';
 import { PanelModel, DashboardModel } from 'app/features/dashboard/state';
 
-interface Props {
+export interface Props {
   urlPanelId: string;
   urlUid?: string;
   urlSlug?: string;
@@ -25,7 +25,7 @@ interface Props {
   dashboard: DashboardModel | null;
 }
 
-interface State {
+export interface State {
   panel: PanelModel | null;
   notFound: boolean;
 }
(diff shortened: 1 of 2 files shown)


torkelo (changes requested):
Looks good except for the added snapshot test (we do not use them anymore, only painful when updating and almost never catches bugs). 

And for react component testing we are migrating to react testing-library, trying to stop using enzyme.

hborchardt:
Ha okay, I didn't know that, I copied from the DashboardPage test. I will improve it.

dprokop (changes requested):
Looks good, echoing @torkelo request about tests :)

hborchardt:
Hopefully this test is better. I saw that doing a shallow render is discouraged in the testing-library, but it did not work for me with actually rendering a full panel in the test case and waiting for it to appear/change, so I hope you are fine with just checking that the panel arrives in the props of a mocked DashboardPage. I confirmed that the testcase was failing without my fix :)

torkelo (approved):
Looks good! 

Thanks for contributing! :heart:

public/app/features/dashboard/containers/SoloPanelPage.tsx
```diff
@@ -57,8 +57,8 @@ export class SoloPanelPage extends Component<Props, State> {
       return;
     }
 
-    // we just got the dashboard!
-    if (!prevProps.dashboard) {
+    // we just got a new dashboard
+    if (!prevProps.dashboard || prevProps.dashboard.uid !== dashboard.uid) {
```
torkelo, line 61: this can never happen, if you navigate to another dashboard there will be an unmount new SoloPage

PR #28592, opened 2020-10-28
Docs: Additional 7.3 upgrade notes

marefr:
Adds a couple upgrade notes for v7.3 regarding database migrations for user invites and snapshots. 

Not sure they make sense, but think it could be good to highlight these changes in case of any user having problems. Historically, we've had some problems related to creating a new table from an existing one and copy data from the old to the new so thought it could be a good idea to highlight these migrations.


docs/sources/installation/upgrading.md
```diff
@@ -297,4 +297,20 @@ The other authentication methods, _Access & secret key_ and _Credentials file_,
...
+| -------- | ------------- | ------------- |
+| Sqlite   | DATETIME      | INTEGER       |
+| MySQL    | DATETIME      | INT           |
+| Postgres | TIMESTAMP     | INTEGER       |
+
+> Please note that if downgrading Grafana to an earlier version, you have to manually change the data type of the _created_ and _updated_ columns back to _old data type_ , otherwise the user invite feature doesn't function as expected.
+
+### Snapshots database migration
+
+The database table _dashboard\_snapshot_, that stores dashboard snapshots, adds a new column _dashboard\_encrypted_ for storing an encrypted snapshot. Any existing snapshots is not migrated to be stored encrypted. Only snapshots created on this version and forward will be stored encrypted.
```
wbrowne, line 314:
```suggestion
The database table _dashboard\_snapshot_, that stores dashboard snapshots, adds a new column _dashboard\_encrypted_ for storing an encrypted snapshot. 
NOTE: Only snapshots created on Grafana 7.3 or later will use this column to store snapshot data as encrypted. Snapshots created before this version will be unaffected.
```

aknuds1 (approved):

wbrowne (approved):
👍

PR #28593, opened 2020-10-28
Docs: Data source provisioning and sigV4

marefr:
Document the sigv4 properties for data source provisioning.


docs/sources/administration/provisioning.md
```diff
@@ -175,6 +175,11 @@ Since not all datasources have the same configuration settings we only have the
 | maxOpenConns            | number  | MySQL, PostgreSQL and MSSQL                                      | Maximum number of open connections to the database (Grafana v5.4+)                          |
 | maxIdleConns            | number  | MySQL, PostgreSQL and MSSQL                                      | Maximum number of connections in the idle connection pool (Grafana v5.4+)                   |
 | connMaxLifetime         | number  | MySQL, PostgreSQL and MSSQL                                      | Maximum amount of time in seconds a connection may be reused (Grafana v5.4+)                |
+| sigV4AuthType                | string  | Elasticsearch                                                       | SigV4 auth provider. default/credentials/keys                                                     |
```
wbrowne, line 178: Wondering if this should be move up alongside the existing `Elasticsearch` rows so that everything is together?

marefr: I thoyght about this as well, but we may have other data sources supported as well :smile: But I can do it for now if you want?

wbrowne: True 😉  In that case we can collapse multiple datasources to one row, similar to the settings directly above here.

No strong feelings, just to me it feels like it could be clearer. Will leave it up to you!

wbrowne (approved):

PR #28597, opened 2020-10-28
Live: updated the reference to use lazy loaded Monaco in code editor.

mckn:
Currently we are importing the CodeEditor in a way that will include the Monaco editor by default in Grafana core. We should change to a proper import statement to load the component using the lazy loading of the Monaco editor.


dprokop (approved):
🔥

PR #28607, opened 2020-10-28
Chore: Migrate Dashboard List panel to React

### LINKED ISSUE #28491: Dashboards list migration to React

kaydelaney:
Migrates the Dashboard list panel to React as part of the ongoing angular to react migration.

Closes #28491


public/app/plugins/panel/dashlist/DashList.tsx
```diff
@@ -0,0 +1,142 @@
...
+interface DashboardGroup {
+  show: boolean;
+  header: string;
+  dashboards: Dashboard[];
+}
+
+export function DashList(props: PanelProps<DashListOptions>) {
+  const [dashboards, setDashboards] = useState<Record<number, Dashboard>>({});
+  useEffect(() => {
+    async function fetchDashboards() {
```
Clarity-89, line 22: Could we maybe use some functionality from `search_srv` here?

public/app/plugins/panel/dashlist/DashList.tsx
```diff
@@ -0,0 +1,142 @@
...
+    e.stopPropagation();
+
+    const isStarred = await getDashboardSrv().starDashboard(dash.id.toString(), dash.isStarred);
+    setDashboards(Object.assign({}, dashboards, { [dash.id]: { ...dash, isStarred } }));
+  };
+
+  const { showStarred, showRecentlyViewed, showHeadings, showSearch } = props.options;
+
+  const dashboardList = Object.values(dashboards);
+  const dashboardGroups: DashboardGroup[] = [
```
Clarity-89, line 95: Should this be wrapped in `useMemo` so it doesn't do all those operations on each render?

public/app/plugins/panel/dashlist/DashList.tsx
```diff
@@ -0,0 +1,142 @@
...
+      show: showSearch,
+    },
+  ];
+
+  return (
+    <CustomScrollbar autoHeightMin="100%" autoHeightMax="100%">
+      {dashboardGroups.map(
+        ({ show, header, dashboards }, i) =>
+          show && (
+            <div className="dashlist-section" key={`dash-group-${i}`}>
```
Clarity-89, line 118: Not sure if it's a part of the task, but would be nice to move the styles from these classes to emotion.

public/app/plugins/panel/dashlist/DashList.tsx
dprokop, line 22: @kaydelaney I'm not a big fan of having the function implementation in a component this way. I suggest extracting this function outside of the component (making it accept respective arguments) and instead of using the `useEffect` you could use https://github.com/streamich/react-use/blob/master/docs/useAsync.md instead

public/app/plugins/panel/dashlist/DashList.tsx
dprokop: Agree with @Clarity-89. Would be great if we could migrate these styles to Emotion

public/app/plugins/panel/dashlist/DashList.tsx
kaydelaney: Yes, good catch

public/app/plugins/panel/dashlist/DashList.tsx
peterholmberg: I suppose this could be a `PureComponent` as well?

Clarity-89: Then we won't be able to use hooks.

dprokop:
@kaydelaney I can only see **starred dashboards**, there is no recent dashboards list anymore:

image

public/sass/components/_panel_dashlist.scss
```diff
@@ -25,13 +26,31 @@
...
+  cursor: pointer;
+  z-index: 1;
 }
 
 .dashlist-folder {
   color: $text-color-weak;
   font-size: $font-size-xs;
 }
 
+.dashlist-title {
```
torkelo, line 38: where is this used?

torkelo:
Nice!

public/app/plugins/panel/dashlist/module.tsx
```diff
@@ -0,0 +1,71 @@
...
+        name: 'Query',
+        defaultValue: '',
+      })
+      .addCustomEditor({
+        path: 'folderId',
+        name: 'Folder',
+        id: 'folder-picker',
+        defaultValue: null,
+        editor: props => {
+          return <FolderPicker initialTitle="All" enableReset={true} onChange={({ id }) => props.onChange(id)} />;
```
Clarity-89, line 48:
```suggestion
          return <FolderPicker initialTitle="All" enableReset={true} onChange={({ id }) => props.onChange(id)} useNewForms />;
```

Clarity-89: This will make the picker look consistent with other fields:
Screenshot 2020-11-11 at 9 35 42

public/app/plugins/panel/dashlist/styles.ts
```diff
@@ -0,0 +1,57 @@
...
+    color: ${theme.colors.textWeak};
+  `,
+
+  dashlistSection: css`
+    margin-bottom: ${theme.spacing.d};
+    padding-top: 3px;
+  `,
+
+  dashlistLink: css`
+    ${styleMixins.listItem(theme)}
```
Clarity-89, line 18:
```suggestion
    ${styleMixins.listItem(theme)};
```

Clarity-89 (approved):
Great job! 🎉 Left a few minor comments to consider.

public/app/plugins/panel/dashlist/module.tsx
```diff
@@ -0,0 +1,71 @@
...
+  .setMigrationHandler((panel: PanelModel<DashListOptions> & Record<string, any>) => ({
+    showStarred: panel.options.showStarred ?? panel.starred,
+    showRecentlyViewed: panel.options.showRecentlyViewed ?? panel.recent,
+    showSearch: panel.options.showSearch ?? panel.search,
+    showHeadings: panel.options.showHeadings ?? panel.headings,
+    maxItems: panel.options.maxItems ?? panel.limit,
+    query: panel.options.query ?? panel.query,
+    folderId: panel.options.folderId ?? panel.folderId,
+    tags: panel.options.tags ?? panel.tags,
+  }));
```
torkelo, line 71: This needs a little bit more work as it has 2 bugs
* Need to cleanup / remove old properties from panel (using delete keyword).
* Need to only do this if pluginVersion < 7.3 

Both of these will sort of protect against the same bug

We had this bug in the TextPanel the release after it was introduced as then this migration handler is called again, and if the old properties remained it would overwrite the new options settings users had set. 

Se here for reference how it's done for TextPanel (after bug fix)
https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/text/textPanelMigrationHandler.ts#L17

torkelo (commented):
Took this for a spin! Works great. 

But I did find one thing, the migration handler has a bug that will be evident in 7.5

PR #28609, opened 2020-10-28
Build: support custom build tags

xlson:

Fixes #

### DIFF
--- a/build.go
+++ b/build.go
@@ -37,7 +37,7 @@ var (
 	libc    string
 	pkgArch string
 	version string = "v1"
-	buildTags []string = []string{}
+	buildTags []string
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -37,6 +37,7 @@ var (
 	libc    string
 	pkgArch string
 	version string = "v1"
+	buildTags []string = []string{}
```
aknuds1, line 40: The default value, a nil slice, is actually good enough. `doBuild` handles nil slices just like empty ones, which is the idiomatic way.
```suggestion
	buildTags []string
```

aknuds1 (approved):
LGTM

PR #28617, opened 2020-10-28
StatPanels: Change default stats option to "Last (not null)"

ryantxu:
This PR changes the default calculation from "mean" to "last (not null)" for the stats panels:
* stat
* gauge
* bar gauge

NOTE: this would not change the behavior for anything that is already saved, just what happens when you start editing a new panel.

I'm not sure trained people are on this behavior, so this may be a big deal :man_shrugging:  -- In general I am skeptical of calculating the average value across a set of search results.  It *may* be meaningful, but the mean of some aggregate values is often undefined.  

Thoughts?


dprokop:
@ryantxu I think this makes sense, but it definitely needs a changelog note

torkelo (approved):
Yes, been thinking the same as well

roidelapluie:
Yes! Definitively great :+1:

PR #28618, opened 2020-10-28
Elasticsearch: Add Support for Serial Differencing Pipeline Aggregation

### LINKED ISSUE #25621: ElasticSearch plugin cannot perform "serial_diff" aggregation
An aggregation type of "serial_diff"

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-serialdiff-aggregation.html

**Why is this needed**:

To draw comparison graphs to detect trends and alert on if there is a change w.r.t for example last week. The built-in time shift feature in Grafana cannot operate on single metrics, and many backends (prometheus, influx) already support this by writing a suitable query.

simianhacker:

This PR closes #25621 by adding support for the serial differencing pipeline aggregation. Per the original feature request:

> To draw comparison graphs to detect trends and alert on if there is a change w.r.t for example last week. The built-in time shift feature in Grafana cannot operate on single metrics, and many backends (prometheus, influx) already support this by writing a suitable query.

image

Fixes #25621

None

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/types.ts
+++ b/public/app/plugins/datasource/elasticsearch/types.ts
@@ -32,6 +32,7 @@ interface MetricConfiguration<T extends MetricAggregationType> {
   hasSettings: boolean;
   hasMeta: boolean;
   defaults: Omit<Extract<MetricAggregation, { type: T }>, 'id' | 'type'>;
+  transform?: Record<string, (value: any) => any>;
 }
 
 type BucketConfiguration<T extends BucketAggregationType> = {
--- a/public/app/plugins/datasource/elasticsearch/types.ts
+++ b/public/app/plugins/datasource/elasticsearch/types.ts
@@ -32,7 +32,6 @@ interface MetricConfiguration<T extends MetricAggregationType> {
   hasSettings: boolean;
   hasMeta: boolean;
   defaults: Omit<Extract<MetricAggregation, { type: T }>, 'id' | 'type'>;
-  transform?: Record<string, (value: any) => any>;
 }
 
 type BucketConfiguration<T extends BucketAggregationType> = {
--- a/pkg/tsdb/elasticsearch/models.go
+++ b/pkg/tsdb/elasticsearch/models.go
@@ -48,6 +48,7 @@ var metricAggType = map[string]string{
 	"moving_fn":      "Moving Function",
 	"cumulative_sum": "Cumulative Sum",
 	"derivative":     "Derivative",
+	"serial_diff":    "Serial Difference",
 	"bucket_script":  "Bucket Script",
 	"raw_document":   "Raw Document",
 }
@@ -68,6 +69,7 @@ var pipelineAggType = map[string]string{
 	"moving_fn":      "moving_fn",
 	"cumulative_sum": "cumulative_sum",
 	"derivative":     "derivative",
+	"serial_diff":    "serial_diff",
 	"bucket_script":  "bucket_script",
 }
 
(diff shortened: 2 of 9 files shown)


Elfo404 (changes requested):
Thanks for the PR! I tested it and it seems to work as expected. 

Would you mind adding a little test case for the query builder in `public/app/plugins/datasource/elasticsearch/specs/query_builder.test.ts` to check if the query gets built correctly?

Elfo404:
We are planning to merge this as soon as the react migration (https://github.com/grafana/grafana/pull/28033) is complete.

Thanks for the contribution!

public/app/plugins/datasource/elasticsearch/types.ts
```diff
@@ -32,6 +32,7 @@ interface MetricConfiguration<T extends MetricAggregationType> {
   hasSettings: boolean;
   hasMeta: boolean;
   defaults: Omit<Extract<MetricAggregation, { type: T }>, 'id' | 'type'>;
+  transform?: Record<string, (value: any) => any>;
```
simianhacker, line 35: @Elfo404 Is there a way to set the types up so the signature of `transform` matches the keys from the `settings` for an aggregation? OR would  we need to create 2 types of `MetricAggregationConfig`?

Elfo404: as far as I understood you need to transform what is saved in the query model (in this case `lag`) into a number because Elasticsearch expects it to be a number in the query, right?

while I think the solution is pretty neat (and I'm sure we can manage to make it type safe), this only works when the queries are issued by the frontend and won't support template variables. As we are planning to deprecate browser access mode (and therefore remove the query builder in a future release https://github.com/grafana/grafana/pull/29649) leveraging the plugin backend to build the query.

Given this, I had the idea of doing something similar by using a `format` function (like your `transform` one) in https://github.com/grafana/grafana/blob/82b21fe35e9bbab99bf3c185e17ab6834ef70776/public/app/plugins/datasource/elasticsearch/datasource.ts#L323  for all the `settings`, `meta` etc.. for each bucket and metric aggregation.

The formatting logic would have to be duplicated in the backend though as we would have to save the original string in the query model, which is less than ideal.

My suggestion is to drop that formatting logic for now, and simply treat `lag` as a number (like we already do for the MovingAverage Settings) while we find a better way to handle this in a more generic way, supporting template variables in the frontend and having a generic formatter in the backend, so instead of using `SettingField` something like: 

```tsx
<InlineField label="Lag">
  <Input
    onBlur={e => dispatch(changeMetricSetting(metric, 'lag', parseInt(e.target.value!, 10)))}
    defaultValue={metric.settings?.lag}
  />
</InlineField>
```

simianhacker: Sounds good... I went ahead and implemented your suggestion and removed the `transform` stuff.

simianhacker:
@Elfo404 This has been converted and should be ready for review.

marefr:
Does this work with Grafana Alerting? Otherwise backend code has to be implemented to support it. @Elfo404 have you checked?

Elfo404:
@marefr didn't have time to properly test this yet, but thanks for the heads up

simianhacker:
@marefr I added the support back in for alerting (with a test). It was originally in there but when I rebased against the React changes it got dropped. Thanks for catching that 👍

Elfo404 (approved):
LGTM! 🚀

marefr:
@simianhacker Thank you for contributing to Grafana!

PR #28620, opened 2020-10-28
Elasticsearch: Exclude pipeline aggregations from order by options

### LINKED ISSUE #19206: Elasticsearch: Should not be possible to select pipeline aggs in term order by
**What happened**:

I made a graph panel with ElasticSearch datasource with the following query:
image

As I wanted to select in the group by section to order the results based on the bucket script result, there was an option "Bucket Script select field"

When I checked the query inspector, I saw the following:
image

At the end of the query there's a  `"9":{"bucket_script":{"field":"select field"}}` part, which results in the following error: "Elasticsearch error: Unknown key for a VALUE_STRING in [9]: [field]."

When I select the order by field, the appropriate Elasticsearch query should be generated without Elasticsearch error.

**How to reproduce it**:

Create an ES query with bucket script aggregation, and try to order the results by that values.

**Environment**:
- Grafana version:  v6.3.5 
- Data source type & version: Elasticsearch 5.5
- User OS & Browser: Windows, Chrome

simianhacker:

This PR fixes a bug with the Elasticsearch datasource user experience. Elasticsearch is unable to sort by pipeline aggregations because they are only executed in the reduce phase on the coordinating node and so you do not have the information on the shard in order to be able to sort the shards buckets to pick the top N to send to the coordinating node.

Before

image

After

image

Fixes #19206 

None

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/query_def.ts
+++ b/public/app/plugins/datasource/elasticsearch/query_def.ts
@@ -246,7 +246,7 @@ export function getMovingAvgSettings(model: any, filtered: boolean) {
 export function getOrderByOptions(target: any) {
   const metricRefs: any[] = [];
   _.each(target.metrics, metric => {
-    if (metric.type !== 'count') {
+    if (metric.type !== 'count' && !isPipelineAgg(metric.type)) {
       metricRefs.push({ text: describeMetric(metric), value: metric.id });
     }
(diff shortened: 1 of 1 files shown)


Elfo404 (approved):
LGTM! 🚀 

Thanks a lot for the PR!

PR #28622, opened 2020-10-28
Prometheus: Fix copy paste behaving as cut and paste

### LINKED ISSUE #28571: Explore: Queries dissapear when interacting with multiple query windows
**What happened**:

dissapearing_labels

**Environment**:
- Grafana version:

master v7.4.0-5409pre (b3ac4eb899)

- Data source type & version:

prometheus

- OS Grafana is installed on:
- User OS & Browser:

windows

Firefox 82

Does not happen on Chrome

- Grafana plugins:
- Others:

aocenas:
Fixes: https://github.com/grafana/grafana/issues/28571

To reproduce it should just need to have 2 queries in explore and copy one part from one to another and then running the query (does not have to return data). The source query should be changed and the copied part should be deleted.

This is not a totally direct fix the issue seems to be that if this [line](https://github.com/grafana/grafana/blob/fd44c01675e54973370969dfb9e78f173aff7910/packages/grafana-ui/src/components/QueryField/QueryField.tsx#L107) runs at some point after the contents are pasted it will erase the copied part (from the source of the copy paste which is even stranger). cc @kaydelaney did you encounter something like this with Slate? Seems to be like their bug.

This fix mainly changes the behaviour of requesting metrics on time range change (which also happens on running query in case of relative range) so the offending line isn't even run after the initial mount.

This also fixes an issue with hints which we seem to have disabled after adding this hint `Dynamic label lookup is disabled for datasources with more than ${lookupMetricsThreshold} metrics.`.

### DIFF
--- a/public/app/plugins/datasource/prometheus/components/PromQueryField.test.tsx
+++ b/public/app/plugins/datasource/prometheus/components/PromQueryField.test.tsx
@@ -3,7 +3,7 @@ import RCCascader from 'rc-cascader';
 import React from 'react';
 import PromQlLanguageProvider, { DEFAULT_LOOKUP_METRICS_THRESHOLD } from '../language_provider';
 import PromQueryField, { groupMetricsByPrefix, RECORDING_RULES_GROUP } from './PromQueryField';
-import { DataSourceInstanceSettings } from '@grafana/data';
+import { DataSourceInstanceSettings, dateTime } from '@grafana/data';
 import { PromOptions } from '../types';
 import { fireEvent, render, screen } from '@testing-library/react';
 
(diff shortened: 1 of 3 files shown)


public/app/plugins/datasource/prometheus/components/PromQueryField.tsx
```diff
@@ -173,21 +166,27 @@ class PromQueryField extends React.PureComponent<PromQueryFieldProps, PromQueryF
...
+        {
+          from: prevProps.range.from.valueOf(),
+          to: prevProps.range.to.valueOf(),
+        }
+      );
+
+    if (languageProvider !== prevProps.datasource.languageProvider) {
+      // We reset this only on DS change so we do not flesh loading state on every rangeChange which happens on every
+      // query run if using relative range.
+      this.setState({
```
aocenas, line 183: This is the main fix that we are clearing this state only here not every time we refreshMetrics

public/app/plugins/datasource/prometheus/components/PromQueryField.tsx
```diff
@@ -298,17 +300,7 @@ class PromQueryField extends React.PureComponent<PromQueryFieldProps, PromQueryF
           ]
         : metricsByPrefix;
 
-    // Hint for big disabled lookups
```
aocenas, line 301: This fixes the hint as here it could end up null and rewriting any existing hints in state.

public/app/plugins/datasource/prometheus/components/PromQueryField.test.tsx
```diff
@@ -51,61 +51,89 @@ describe('PromQueryField', () => {
...
-    for (const node of listNodes) {
-      expect(changedMetrics).toContain(node.innerHTML);
-    }
+    // If we check the cascader right away it should be in loading state
+    let cascader = screen.getByRole('button');
+    expect(cascader.textContent).toContain('Loading');
+    checkMetricsInCascader(await screen.findByRole('button'), changedMetrics);
+  });
+
+  it('refreshes metrics when time range changes but dont show loading state', async () => {
```
aocenas, line 90: Testing slate itself is tricky so this is just indirect test and just checks if we get to loading state after timeRange change

zoltanbedi (approved):
lgtm

chancez:
I'm still seeing this in 7.3.1

zoltanbedi:
@chancez this is not released yet. Will be in 7.3.2.

aocenas:
@dprokop should we change the milestone here? As there is already this PR https://github.com/grafana/grafana/pull/28691 would that confuse the release scripts somwhow?

chancez:
@aocenas @zoltanbedi I saw it in the 7.3.1 release notes, so that should get fixed.

zoltanbedi:
> @aocenas @zoltanbedi I saw it in the 7.3.1 release notes, so that should get fixed.

Thanks @chancez, I've removed it from the changelog.

PR #28624, opened 2020-10-28
Chore: allow higher nodejs version than 12

zoltanbedi:

With node 12 going to maintenance soon and node 14 going to lts I think it is time to allow higher nodejs versions than 12.
node release table 

## In this pr:

- Set the engine to ">= 12" so one can use nodejs 14 or 15
- Removes the `grunt-contrib-compress` (it was not used and had a dependency that was not compatible with higher nodejs)
- Changes [node-sass](https://github.com/sass/node-sass) to [sass](https://github.com/sass/dart-sass) - this means that a lot of dependency is removed like `node-gyp@3` which still needed python2 to compile

From the node-sass readme
> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

I've tried out the build, start and test commands with node 15.0.1 and it all succeded.


stevenvachon (approved):

mckn (approved):
LGTM! Great work @zoltanbedi 🙌

aocenas:
@zoltanbedi I wonder if it make sense to have a spread of versions of this instead of just upgrade the node version. I fear that you can get into situations when locally you will use higher version than CI and then you will have CI failures that will be non obvious if you don't remember to check the node versions.

zoltanbedi:
The only problem I can see is a dependency that needs higher nodejs version for building what we have in ci but it is very unlikely right now as far as I can see.

HHHui:
Hi, when i ran`go run build.go package` command try to package my customized grafana, it runs the compress:release grunt task. But since the `grunt-contrib-compress` dependency was removed, the packaging was failed. Is there other way to do this? @zoltanbedi

zoltanbedi:
Ooh it seems like that it was used then 😄 interesting thing that we didn't notice this internally guess we don't use that command anymore. Before my next step (removing compress:release task or adding back grunt-contrib-compress). Why do you use this command? Is the docker image not working for you? I want to understand your use case.

PR #28625, opened 2020-10-29
Variables: Adds named capture groups to variable regex

### LINKED ISSUE #12039: [UX] editview=templating should link to regex documentation
@davkal 
Grafana 5.1.3

When editing a variable via https://play.grafana.org/d/000000029/prometheus-demo-dashboard?editview=templating&refresh=5m&orgId=1 I can mouse-over the `i` to see that this is optional, but I don't actually get a link to documentation, or even an online parser, for the regex flavour used.

### LINKED ISSUE #21076: prometheus template variable support __text __value, similar to mysql
template variable has diplay text and actual value, similar to mysql's text/value

**Why is this needed**:
we use ip when collecting data, but use domain when query metric grapha. we have another metric, which preserves the mapping between domain and ip.

we use ip to collect data because ip is more stable than domain. we may change domain.

**Possible solutions**
label_values(metric, label) add another param label_values(metric, text, value)

yenalsnavaj:
This PR allows 'text' and 'value' named capture groups to be specified in the regular expression for variable queries so that distinct values and display text can be extracted. This is simliar the __text and __value functionality available if using a SQL datasource but should work for all datasources - especially Prometheus, which is the tested use case. 

Fixes #21076
Fixes #12039

### DIFF
--- a/docs/sources/variables/filter-variables-with-regex.md
+++ b/docs/sources/variables/filter-variables-with-regex.md
@@ -79,4 +79,30 @@ demo.robustperception.io:9090
 demo.robustperception.io:9093
 demo.robustperception.io:9100
 ```
+
+## Filter and use named text and value capture groups
+
+This can be useful if the display text should differ from the value.
+
+List of options:
+
+```text
+kube_pod_labels{label_description="a-descriptive-name-for-pod1",namespace="default",pod="pod1"}
+kube_pod_labels{label_description="the-second-pod",namespace="default",pod="pod2"}
+kube_pod_labels{label_description="a-special-pod",namespace="default",pod="pod3"}
+```
+
+Regex:
+
+```regex
+/pod="(?<value>[^"]+)|label_description="(?<text>[^"]+)/g
+```
+
+Result:
+
+```text
+value   text
+pod1    a-descriptive-name-for-pod1
+pod2    the-second-pod
+pod3    a-sepcial-pod
 ```
--- a/docs/sources/variables/filter-variables-with-regex.md
+++ b/docs/sources/variables/filter-variables-with-regex.md
@@ -80,29 +80,34 @@ demo.robustperception.io:9093
 demo.robustperception.io:9100
 ```
 
-## Filter and use named text and value capture groups
+## Filter and modify using named text and value capture groups
 
-This can be useful if the display text should differ from the value.
+Using named capture groups it is possible to capture separate 'text' and 'value' parts from the options returned by the variable query. This allows the variable dropdown to contain a friendly name for each value that can be selected.
 
-List of options:
+For example, when querying the `node_hwmon_chip_names` prometheus metric the `chip_name` is a lot friendlier that the `chip` value. So the following variable query result:
 
 ```text
-kube_pod_labels{label_description="a-descriptive-name-for-pod1",namespace="default",pod="pod1"}
-kube_pod_labels{label_description="the-second-pod",namespace="default",pod="pod2"}
-kube_pod_labels{label_description="a-special-pod",namespace="default",pod="pod3"}
+node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_0",chip_name="enp216s0f0np0"} 1
+node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_1",chip_name="enp216s0f0np1"} 1
+node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_2",chip_name="enp216s0f0np2"} 1
+node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_3",chip_name="enp216s0f0np3"} 1
 ```
 
-Regex:
+Passed through the following Regex:
 
 ```regex
-/pod="(?<value>[^"]+)|label_description="(?<text>[^"]+)/g
+/chip_name="(?<text>[^"]+)|chip="(?<value>[^"]+)/g
 ```
 
-Result:
+Would produce the following dropdown:
 
 ```text
-value   text
-pod1    a-descriptive-name-for-pod1
-pod2    the-second-pod
-pod3    a-sepcial-pod
+Display Name          Value
+------------          -------------------------
+enp216s0f0np0         0000:d7:00_0_0000:d8:00_0
+enp216s0f0np1         0000:d7:00_0_0000:d8:00_1
+enp216s0f0np2         0000:d7:00_0_0000:d8:00_2
+enp216s0f0np3         0000:d7:00_0_0000:d8:00_3
 ```
+
+**NOTE:** only `text` and `value` capture group names are supported.
--- a/docs/sources/variables/filter-variables-with-regex.md
+++ b/docs/sources/variables/filter-variables-with-regex.md
@@ -82,9 +82,9 @@ demo.robustperception.io:9100
 
 ## Filter and modify using named text and value capture groups
 
-Using named capture groups it is possible to capture separate 'text' and 'value' parts from the options returned by the variable query. This allows the variable dropdown to contain a friendly name for each value that can be selected.
+Using named capture groups, you can capture separate 'text' and 'value' parts from the options returned by the variable query. This allows the variable drop-down list to contain a friendly name for each value that can be selected.
 
-For example, when querying the `node_hwmon_chip_names` prometheus metric the `chip_name` is a lot friendlier that the `chip` value. So the following variable query result:
+For example, when querying the `node_hwmon_chip_names` Prometheus metric, the `chip_name` is a lot friendlier that the `chip` value. So the following variable query result:
 
 ```text
 node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_0",chip_name="enp216s0f0np0"} 1
@@ -99,7 +99,7 @@ Passed through the following Regex:
 /chip_name="(?<text>[^"]+)|chip="(?<value>[^"]+)/g
 ```
 
-Would produce the following dropdown:
+Would produce the following drop-down list:
 
 ```text
 Display Name          Value
--- a/docs/sources/variables/filter-variables-with-regex.md
+++ b/docs/sources/variables/filter-variables-with-regex.md
@@ -110,4 +110,4 @@ enp216s0f0np2         0000:d7:00_0_0000:d8:00_2
 enp216s0f0np3         0000:d7:00_0_0000:d8:00_3
 ```
 
-**NOTE:** only `text` and `value` capture group names are supported.
+**Note:** Only `text` and `value` capture group names are supported.
--- a/public/app/features/variables/query/reducer.ts
+++ b/public/app/features/variables/query/reducer.ts
@@ -87,18 +87,14 @@ const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
 };
 
 const getAllMatches = (str: string, regex: RegExp): any => {
-  let results = {};
+    const results = {};
   let matches;
 
-  if (regex.global) {
-    while ((matches = regex.exec(str))) {
-      _.merge(results, matches);
-    }
-  } else {
-    if ((matches = regex.exec(str))) {
-      _.merge(results, matches);
-    }
-  }
+  do {
+    matches = regex.exec(str);
+    _.merge(results, matches);
+  } while (regex.global && matches);
+
   return results;
 };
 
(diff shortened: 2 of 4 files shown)


docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,30 @@ demo.robustperception.io:9090
 demo.robustperception.io:9093
 demo.robustperception.io:9100
 ```
+
+## Filter and use named text and value capture groups
+
+This can be useful if the display text should differ from the value.
```
oddlittlebird, line 85: What can be useful? Please explain what "this"is that you are referring to.

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,30 @@ demo.robustperception.io:9090
...
+
+List of options:
```
oddlittlebird, line 87: I am confused. Are these options an example, or are they common to all users? This section needs more context, as do the Regex and Result below.

oddlittlebird (commented):
I have a few suggestions for the documentation.

docs/sources/variables/filter-variables-with-regex.md
yenalsnavaj: Sure. I was following the format of the examples earlier the file. Agree that it could be improved - documentation is not my strong point ;) I’ll rework with expanded examples that relate tot the use case in the issue that it fixes

yenalsnavaj:
@oddlittlebird i've updated the documentation. Let me know what you think.

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,35 @@ demo.robustperception.io:9090
...
+node_hwmon_chip_names{chip="0000:d7:00_0_0000:d8:00_3",chip_name="enp216s0f0np3"} 1
+```
+
+Passed through the following Regex:
+
+```regex
+/chip_name="(?<text>[^"]+)|chip="(?<value>[^"]+)/g
 ```
+
+Would produce the following dropdown:
```
oddlittlebird, line 102:
```suggestion
Would produce the following drop-down list:
```

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,35 @@ demo.robustperception.io:9090
 demo.robustperception.io:9093
 demo.robustperception.io:9100
 ```
+
+## Filter and modify using named text and value capture groups
+
+Using named capture groups it is possible to capture separate 'text' and 'value' parts from the options returned by the variable query. This allows the variable dropdown to contain a friendly name for each value that can be selected.
```
oddlittlebird, line 85:
```suggestion
Using named capture groups, you can capture separate 'text' and 'value' parts from the options returned by the variable query. This allows the variable drop-down list to contain a friendly name for each value that can be selected.
```

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,35 @@ demo.robustperception.io:9090
...
+
+For example, when querying the `node_hwmon_chip_names` prometheus metric the `chip_name` is a lot friendlier that the `chip` value. So the following variable query result:
```
oddlittlebird, line 87:
```suggestion
For example, when querying the `node_hwmon_chip_names` Prometheus metric, the `chip_name` is a lot friendlier that the `chip` value. So the following variable query result:
```

docs/sources/variables/filter-variables-with-regex.md
```diff
@@ -79,4 +79,35 @@ demo.robustperception.io:9090
...
+```text
+Display Name          Value
+------------          -------------------------
+enp216s0f0np0         0000:d7:00_0_0000:d8:00_0
+enp216s0f0np1         0000:d7:00_0_0000:d8:00_1
+enp216s0f0np2         0000:d7:00_0_0000:d8:00_2
+enp216s0f0np3         0000:d7:00_0_0000:d8:00_3
+```
+
+**NOTE:** only `text` and `value` capture group names are supported.
```
oddlittlebird, line 113:
```suggestion
**Note:** Only `text` and `value` capture group names are supported.
```

oddlittlebird (commented):
Yes, that is much better! I added a little more polish, but now I can understand what this does. GREAT STUFF!

oddlittlebird:
Thanks, @yenalsnavaj! The code review will be a separate process, done by the developers. We appreciate the documentation coming up front, though!

public/app/features/variables/query/reducer.ts
```diff
@@ -86,6 +86,22 @@ const sortVariableValues = (options: any[], sortOrder: VariableSort) => {
...
+  if (regex.global) {
+    while ((matches = regex.exec(str))) {
+      _.merge(results, matches);
+    }
+  } else {
+    if ((matches = regex.exec(str))) {
+      _.merge(results, matches);
+    }
+  }
+  return results;
```
hugohaggmark, line 102: This looks ok, but having the nested else doesn't seem that clean. What about something like this
```suggestion
    const results = {};
  let matches;

  do {
    matches = regex.exec(str);
    _.merge(results, matches);
  } while (regex.global && matches);

  return results;
```

public/app/features/variables/query/reducer.ts
```diff
@@ -109,13 +125,24 @@ const metricNamesToVariableValues = (variableRegEx: string, sort: VariableSort,
...
-      if (!matches) {
+      matches = getAllMatches(value, regex);
+
+      if (_.isEmpty(matches)) {
+        continue;
+      }
+
+      if (matches.groups && matches.groups.value) {
+        value = matches.groups.value;
+      } else if (matches.groups && matches.groups.hasOwnProperty('value')) {
```
hugohaggmark, line 136: Do I understand this correctly that we ignore any falsy values here? Couldn't the value be assigned the text group value if that exists?

hugohaggmark (commented):
This is a powerful and great addition @yenalsnavaj 🌟  I've left some things to consider and also I think we need to advertize this powerful new feature somehow.

1. Maybe add more information to the placeholder here:
image
2. Would be great if icon tooltip here could include a link to the documentation you wrote
image

public/app/features/variables/query/reducer.ts
yenalsnavaj: I had considered that -- but as the value of the variable will almost always be used as a key is some other dashboard metric query I felt that it might introduce unwanted behaviour. I'm not set on that though so if you think it would be better we could include. Whatever is most consistent and understandable.

public/app/features/variables/query/reducer.ts
yenalsnavaj: Agree, much cleaner. Extracting the method was already part of a refactoring. I always need to push a bit further...

yenalsnavaj:
> This is a powerful and great addition @yenalsnavaj 🌟 I've left some things to consider and also I think we need to advertize this powerful new feature somehow.
> 
> 1. Maybe add more information to the placeholder here:
>    image
> 2. Would be great if icon tooltip here could include a link to the documentation you wrote
>    image

I'll take a look at adding something to the tool tip

public/app/features/variables/query/reducer.ts
hugohaggmark: I think we already do this in line 117 above, so my thought was to keep that behavior. https://github.com/grafana/grafana/blob/81699f07d3dc3d6706f7ea7e4a2a24a1db06a836/public/app/features/variables/query/reducer.ts#L117

hugohaggmark:
> > This is a powerful and great addition @yenalsnavaj 🌟 I've left some things to consider and also I think we need to advertize this powerful new feature somehow.
> > 
> > 1. Maybe add more information to the placeholder here:
> >    image
> > 2. Would be great if icon tooltip here could include a link to the documentation you wrote
> >    image
> 
> I'll take a look at adding something to the tool tip

Let me know if something is confusing, I just think this outstanding feature is super-useful, and it's a shame if users can't find it/use it.

yenalsnavaj:
> > > This is a powerful and great addition @yenalsnavaj 🌟 I've left some things to consider and also I think we need to advertize this powerful new feature somehow.
> > > 
> > > 1. Maybe add more information to the placeholder here:
> > >    image
> > > 2. Would be great if icon tooltip here could include a link to the documentation you wrote
> > >    image
> > 
> > 
> > I'll take a look at adding something to the tool tip
> 
> Let me know if something is confusing, I just think this outstanding feature is super-useful, and it's a shame if users can't find it/use it.

Makes sense - I can't see how to include a <Link/> <a> or whatever inside the <InlineFormLabel/> - any pointers?

hugohaggmark:
> > > > This is a powerful and great addition @yenalsnavaj 🌟 I've left some things to consider and also I think we need to advertize this powerful new feature somehow.
> > > > 
> > > > 1. Maybe add more information to the placeholder here:
> > > >    image
> > > > 2. Would be great if icon tooltip here could include a link to the documentation you wrote
> > > >    image
> > > 
> > > 
> > > I'll take a look at adding something to the tool tip
> > 
> > 
> > Let me know if something is confusing, I just think this outstanding feature is super-useful, and it's a shame if users can't find it/use it.
> 
> Makes sense - I can't see how to include a or whatever inside the - any pointers?

I think you can pass an React Element (with text and a link) to the tooltip prop here:
https://github.com/grafana/grafana/blob/45f311874de15e7b54e5e84b4ef2d24755406b2a/public/app/features/variables/query/QueryVariableEditor.tsx#L196

The placeholder can be changed here: https://github.com/grafana/grafana/blob/45f311874de15e7b54e5e84b4ef2d24755406b2a/public/app/features/variables/query/QueryVariableEditor.tsx#L203

yenalsnavaj:
@hugohaggmark just noticed that the htlm tags that I had included were stripped out so message didn't make sense. I had located files but wasn't sure how to include a link element - have figured it out now though so please let me know what you think. 

Also includes the updated tests and change in behaviour to use text capture if value capture doesn't match.

hugohaggmark (approved):
I think this looks great. Thank you for contributing @yenalsnavaj!

AnderssonPeter:
Does the 7.3.4 release contain this functionality?
The documentation states that it should exist in the 7.3.X releases https://grafana.com/docs/grafana/latest/variables/filter-variables-with-regex/#filter-and-modify-using-named-text-and-value-capture-groups

But i cant get it to work, i have the following nodes
* `node_meta{container_label_com_docker_swarm_node_id="7ixpxu9rftn1wfexyjfm1whwa",instance="10.0.15.16:9100",job="node-exporter",node_id="7ixpxu9rftn1wfexyjfm1whwa",node_name="owgeotest01.localdomain"}`
* `node_meta{container_label_com_docker_swarm_node_id="b1v7zostjjbl37ly5mronsjju",instance="10.0.15.15:9100",job="node-exporter",node_id="b1v7zostjjbl37ly5mronsjju",node_name="owgeotest02.localdomain"}`

And i have tried with the following regexes and none of them worked
* `/node_id="(?<value>[^"]+)".*node_name="(?<text>[^"]+)"/` - Just shows the id
* `/node_id="(?<value>[^"]+)".*node_name="(?<text>[^"]+)"/g` - just shows the id and i only get 1 result.
* `/node_id="(?<value>[^"]+)"|node_name="(?<text>[^"]+)"/g` - Just shows the id
* `/node_name="(?<text>[^"]+|node_id="(?<value>[^"]+)")"/g` - Just shows one name, value does not work (if i select nothing matches)
* `/node_name="(?<text>[^"]+|node_id="(?<value>[^"]+)")"/` - Shows the name of all nodes but values is lost (if i select nothing matches)

AnderssonPeter:
After looking at the commit for QueryVariableEditor.tsx it does not look like this change is included, it would be nice if the documentation had a warning.

hugohaggmark:
@AnderssonPeter sorry for the confusion, sometimes our docs get updated before the changes do. Especially when the feature contribution is external, we try to keep everything in the same PR. If you have a look in the milestone, you can see that we schedule this feature for the 7.4.0 realease.
image

Again sorry for the inconvenience and we'll try to better. You can open a PR that clarifies this in the doc [https://github.com/grafana/grafana/blob/master/docs/sources/variables/filter-variables-with-regex.md](https://github.com/grafana/grafana/blob/master/docs/sources/variables/filter-variables-with-regex.md).

AnderssonPeter:
Thanks for clearifying!

PR #28632, opened 2020-10-29
CloudWatch: Slight performance improvement

aknuds1:
Make slight CloudWatch performance improvement by specifying an array slice capacity. Also rename a function to reflect that it takes a collection of responses.


papagian (approved):
LGTM

Issue #28629, opened 2020-10-29
Plugins signing: Panels inside Signed App marked as Unsigned


mikhailredis:
**What happened**:

The panel, which is a part of Signed Application, marked as Unsigned.

![Screen Shot 2020-10-28 at 9 08 51 PM](https://user-images.githubusercontent.com/47795110/97524455-f3ef4b00-1961-11eb-97bf-241858acc0a8.png)

**What you expected to happen**:

The panel should be marked as Signed.

**How to reproduce it (as minimally and precisely as possible)**:

Install `redis-app` plug-in.

**Anything else we need to know?**:

**Environment**:
- Grafana version: Grafana 7.3
- Data source type & version: App, 1.0.1
- OS Grafana is installed on: Linux
- User OS & Browser: 
- Grafana plugins: Redis-app
- Others:

aknuds1:
Isn't this fundamentally the same bug as #28628? WDYT @hugohaggmark?

hugohaggmark:
Yes! /duplicate #28628

grafanabot:
Thanks for creating this issue! It looks like this has already been reported by another user. We’ve closed this in favor of the existing one. Please consider adding any details you think is missing to that issue.

To avoid having your issue closed in the future, please read our [CONTRIBUTING](https://github.com/grafana/grafana/blob/master/CONTRIBUTING.md) guidelines.

Happy graphing!

dprokop:
@hugohaggmark @aknuds1 this is not a duplicate. Apparently, when checking for the app plugin being signed or not the panel is not taken into consideration and its status is still unsigned. This is the meta of redis-cli panel:

```json
{
    "baseUrl": "public/plugins/redis-app",
    "hideFromList": false,
    "id": "redis-cli-panel",
    "info": {
        "author": {
            "name": "Redis Labs",
            "url": "https://oss.redislabs.com"
        },
        "description": "Redis CLI panel",
        "links": null,
        "logos": {
            "small": "public/plugins/redis-cli-panel/img/logo.svg",
            "large": "public/plugins/redis-cli-panel/img/logo.svg"
        },
        "build": {},
        "screenshots": [],
        "version": "%VERSION%",
        "updated": "%TODAY%"
    },
    "module": "plugins/redis-app/redis-cli-panel/module",
    "name": "Redis CLI",
    "signature": "unsigned",
    "skipDataQuery": false,
    "sort": 100,
    "state": ""
}
```

[...]

Issue #27364, opened 2020-09-03
App plugin marked as signed but data source can't start


alexanderzobnin:
**What happened**:
Signed application plugin with data source included can't run data source backend.

![Screenshot from 2020-09-03 12-30-57](https://user-images.githubusercontent.com/4932851/92133728-6257d700-ee11-11ea-8177-5cef5e3e84ab.png)

**What you expected to happen**:
Backend data source up and running.

**How to reproduce it (as minimally and precisely as possible)**:
Get the [latest Zabbix plugin version](https://github.com/alexanderzobnin/grafana-zabbix/releases/download/v4.0.1/alexanderzobnin-zabbix-app-4.0.1.zip), install it and run Grafana.

**Anything else we need to know?**:
That's happened because grafana scans plugins folder and loads plugins from every folder containing the `plugin.json` file. Then grafana checks every plugin for valid signature. So if app has tree similar to this:
```
example-app
|-- example-datasource
|  |-- plugin.json
|  |-- datasource_executable
|-- example-panel
|  |-- plugin.json
|-- plugin.json
|-- MANIFEST.txt
```
then datasource will be treated as unsigned, however plugin has correct MANIFEST file with all files insluded.

Logs from grafana:
```
t=2020-09-03T15:17:03+0000 lvl=dbug msg="Scanning external plugins directory" logger=plugins dir=/var/lib/grafana/plugins
t=2020-09-03T15:17:03+0000 lvl=dbug msg="Getting signature state of plugin" logger=plugins plugin=alexanderzobnin-zabbix-datasource
t=2020-09-03T15:17:03+0000 lvl=dbug msg="Invalid Plugin Signature" logger=plugins pluginID=alexanderzobnin-zabbix-datasource pluginD
ir=/var/lib/grafana/plugins/alexanderzobnin-zabbix-app/datasource-zabbix state=unsigned
t=2020-09-03T15:17:03+0000 lvl=eror msg="Failed to load plugin" logger=plugins error="plugin \"alexanderzobnin-zabbix-datasource\" i
s unsigned" pluginPath=/var/lib/grafana/plugins/alexanderzobnin-zabbix-app/datasource-zabbix
t=2020-09-03T15:17:03+0000 lvl=dbug msg="Getting signature state of plugin" logger=plugins plugin=alexanderzobnin-zabbix-app
[...]

PR #28638, opened 2020-10-29
Plugins: Fix descendent frontend plugin signature validation

wbrowne:
Frontend plugins (or non-backend however you want to look at it), that are nested within a plugin, unless they have a MANIFEST.txt - will get marked as `unsigned` by default. Later in the validation process we check the current plugin's signature status, and since it's non backend we exit and don't update the status. Thanks to @aknuds1's [fix](https://github.com/grafana/grafana/issues/27364) for a related issue, we just need to move the check for plugin nesting sooner as this will recognize the plugin can inherit a signature, and should do so - preventing us from maintaining the `unsigned` status which shows up on the frontend

TLDR; It moves the logic for checking for root plugin signature earlier as currently it exits too soon while checking if the plugin is backend or not.

Fixes #
https://github.com/grafana/grafana/issues/28629

### DIFF
--- a/pkg/plugins/plugins.go
+++ b/pkg/plugins/plugins.go
@@ -373,7 +373,6 @@ func (scanner *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
 
 // validateSignature validates a plugin's signature.
 func (s *PluginScanner) validateSignature(plugin *PluginBase) *PluginError {
-	// If the plugin has a parent plugin
 	if plugin.Root != nil {
 		// If a descendant plugin with invalid signature, set signature to that of root
 		if plugin.IsCorePlugin || plugin.Signature == PluginSignatureInternal {
--- a/pkg/plugins/plugins.go
+++ b/pkg/plugins/plugins.go
@@ -373,17 +373,7 @@ func (scanner *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
 
 // validateSignature validates a plugin's signature.
 func (s *PluginScanner) validateSignature(plugin *PluginBase) *PluginError {
-	// For the time being, we choose to only require back-end plugins to be signed
-	// NOTE: the state is calculated again when setting metadata on the object
-	if !plugin.Backend || !s.requireSigned {
-		return nil
-	}
-
-	if plugin.Signature == PluginSignatureValid {
-		s.log.Debug("Plugin has valid signature", "id", plugin.Id)
-		return nil
-	}
-
+	// If the plugin has a parent plugin
 	if plugin.Root != nil {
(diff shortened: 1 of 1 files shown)


pkg/plugins/plugins.go
```diff
@@ -373,17 +373,7 @@ func (scanner *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
 
 // validateSignature validates a plugin's signature.
 func (s *PluginScanner) validateSignature(plugin *PluginBase) *PluginError {
-	// For the time being, we choose to only require back-end plugins to be signed
-	// NOTE: the state is calculated again when setting metadata on the object
-	if !plugin.Backend || !s.requireSigned {
-		return nil
-	}
-
-	if plugin.Signature == PluginSignatureValid {
```
aknuds1, line 382: I think removing this is a bug, since the code line 392 will execute for plugins with valid signatures.

pkg/plugins/plugins.go
```diff
@@ -373,17 +373,7 @@ func (scanner *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
...
-		s.log.Debug("Plugin has valid signature", "id", plugin.Id)
-		return nil
-	}
-
+	// If the plugin has a parent plugin
```
aknuds1, line 376: I would remove this comment, since it's just overlapping with the one on 378.

aknuds1 (changes requested):

dprokop (approved):
Tested and it works fine! Thanks @wbrowne

pkg/plugins/plugins.go
wbrowne: It's not removed though, just moved. I essentially just moved the `if plugin.Root != nil {` block, but the diff comes out weird 😞

pkg/plugins/plugins.go
```diff
@@ -403,6 +392,17 @@ func (s *PluginScanner) validateSignature(plugin *PluginBase) *PluginError {
...
+	// For the time being, we choose to only require back-end plugins to be signed
+	// NOTE: the state is calculated again when setting metadata on the object
+	if !plugin.Backend || !s.requireSigned {
+		return nil
+	}
+
+	if plugin.Signature == PluginSignatureValid {
+		s.log.Debug("Plugin has valid signature", "id", plugin.Id)
+		return nil
+	}
```
aknuds1, line 404: I think this one should be moved back, since the check for this case is [here](https://github.com/grafana/grafana/blob/master/pkg/plugins/plugins.go#L396).

aknuds1 (changes requested):

aknuds1 (approved):
LGTM - tested and it works. Thanks!

PR #28642, opened 2020-10-29
Cloudwatch: Fix duplicate metric data

### LINKED ISSUE #28301: AWS CloudWatch with multiple queries are duplicating values in graph
When i use AWS Cloudwatch Data Source the values showed in graph become duplicated. I'm using multiple queries and filtering by instanceid. This issue was observed after upgrade to last update.

Above the example: 
image

zoltanbedi:

Fixes #28301

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/datasource.test.ts
+++ b/public/app/plugins/datasource/cloudwatch/datasource.test.ts
@@ -1,7 +1,7 @@
-import { CloudWatchDatasource } from './datasource';
-import { TemplateSrv } from '../../../features/templating/template_srv';
+import { DataQueryResponse, dateTime, DefaultTimeRange } from '@grafana/data';
 import { setBackendSrv } from '@grafana/runtime';
-import { DataQueryResponse, DefaultTimeRange } from '@grafana/data';
+import { TemplateSrv } from '../../../features/templating/template_srv';
+import { CloudWatchDatasource } from './datasource';
 
 describe('datasource', () => {
   describe('query', () => {
@@ -32,6 +32,33 @@ describe('datasource', () => {
         } as any)
         .toPromise()) as any;
       expect(response.data).toEqual([]);
+    });
+  });
+
+  describe('performTimeSeriesQuery', () => {
+    it('should return the same length of data as result', async () => {
+      const { datasource } = setup();
+      const awsRequestMock = jest.spyOn(datasource, 'awsRequest');
+      const buildCloudwatchConsoleUrlMock = jest.spyOn(datasource, 'buildCloudwatchConsoleUrl');
+      buildCloudwatchConsoleUrlMock.mockImplementation(() => '');
+      awsRequestMock.mockImplementation(async () => {
+        return {
+          results: {
+            a: { refId: 'a', series: [{ name: 'cpu', points: [1, 1] }], meta: { gmdMeta: '' } },
+            b: { refId: 'b', series: [{ name: 'memory', points: [2, 2] }], meta: { gmdMeta: '' } },
+          },
+        };
+      });
+      const response: DataQueryResponse = await datasource.performTimeSeriesQuery(
+        {
+          queries: [
+            { datasourceId: 1, refId: 'a' },
+            { datasourceId: 1, refId: 'b' },
+          ],
+        } as any,
+        { from: dateTime(), to: dateTime() } as any
+      );
+      expect(response.data.length).toEqual(2);
     });
   });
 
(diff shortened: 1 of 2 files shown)


papagian (approved):
Great! I had prepared also a fix for this but my solution was much more complicated than this.

public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -613,7 +613,6 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
           return {
             error: error || queryResult.error ? { message: queryResult.error } : null,
             data: [
-              ...data,
```
torkelo, line 616: this whole reduce function was very strange and hard to understand, seems like it can be simplified and some comments added to what is being done. Looping through the results and then looking up the queries might be an easier more intuitive way to write this. 

The reduce method on the queries and then for each query rebuild the whole data array was really messy and hard to follow for me, especially with no comments that explain what the code is trying to do

zoltanbedi: exactly my thoughts. Let me change this then.

torkelo: see frontend chan I think this function looks buggy, there seems to be no matching between query and result (what data frame the link is added to, it adds it to all, so the last query will overwrite all), if I read the code correctly

papagian: @zoltanbedi 
here is the solution I had prepared (I wasn't so much sure about the error handling, therefore I tried to imitate the existing approach):
```
  async performTimeSeriesQuery(request: MetricRequest, { from, to }: TimeRange): Promise<any> {
    try {
      const res: TSDBResponse = await this.awsRequest(TSDB_QUERY_ENDPOINT, request);
      const dataframes: DataFrame[] = toDataQueryResponse({ data: res }).data;
      if (!dataframes || dataframes.length <= 0) {
        return { data: [] };
      }

      const data = dataframes.map(frame => {
        const queryResult = res.results[frame.refId!];
        const error = queryResult.error ? { message: queryResult.error } : null;
        if (!queryResult) {
          return { frame, error };
        }

        const requestQuery = request.queries.find(q => q.refId === frame.refId!) as any;

        const link = this.buildCloudwatchConsoleUrl(
          requestQuery!,
          from.toISOString(),
          to.toISOString(),
          frame.refId!,
          queryResult.meta.gmdMeta
        );

        if (link) {
          for (const field of frame.fields) {
            field.config.links = [
              {
                url: link,
                title: 'View in CloudWatch console',
                targetBlank: true,
              },
            ];
          }
        }
        return { frame, error };
      });

      return {
        data: data.map(o => o.frame),
        error: data
          .map(o => o.error)
          .reduce((err, error) => {
            return err || error;
          }, null),
      };
    } catch (err) {
      if (/^Throttling:.*/.test(err.data.message)) {
        const failedRedIds = Object.keys(err.data.results);
        const regionsAffected = Object.values(request.queries).reduce(
          (res: string[], { refId, region }) =>
            (refId && !failedRedIds.includes(refId)) || res.includes(region) ? res : [...res, region],
          []
        ) as string[];

        regionsAffected.forEach(region => this.debouncedAlert(this.datasourceName, this.getActualRegion(region)));
      }

      if (err.data && err.data.message === 'Metric request error' && err.data.error) {
        err.data.message = err.data.error;
      }

      throw err;
    }
  }
```

torkelo: that looks way more reasonable & readable and looks you here match the result with the correct query and not just the last query

zoltanbedi: Thanks Sofia, this looks good.

public/app/plugins/datasource/cloudwatch/datasource.test.ts
```diff
@@ -35,6 +35,33 @@ describe('datasource', () => {
     });
   });
 
+  describe('performTimeSeriesQuery', () => {
+    it('should return the same length of data as result', async () => {
+      const { datasource } = setup();
+      const awsRequestMock = jest.spyOn(datasource, 'awsRequest');
+      const buildCloudwatchConsoleUrlMock = jest.spyOn(datasource, 'buildCloudwatchConsoleUrl');
+      buildCloudwatchConsoleUrlMock.mockImplementation(() => '');
```
aocenas, line 43: Do we need to replace this? seems like it does not do any networking and we could actually test if the function assigns the links correctly.

zoltanbedi: Didn't want to add it for this test. We could have another test for the links.

aocenas (approved):
Looks good

public/app/plugins/datasource/cloudwatch/datasource.test.ts
aocenas: No strong feelings, just felt we could easily test that if we already have this test.

Issue #28395, opened 2020-10-20
grafana/toolkit: Datasource with backend plugin template throws Typescript errors when building


dprokop:
**What happened**:
Created a new **data source with backend** plugin using grafana/toolkit (next version). When running `yarn build` the build failed due to the following errors:

![image](https://user-images.githubusercontent.com/2376619/96580404-821b5f80-12d8-11eb-8953-941940102664.png)

**What you expected to happen**:
I expect the plugin to build correctly.
**How to reproduce it (as minimally and precisely as possible)**:
1. Bootstrap a data source with backeng plugin using grafana/toolkit: `npx @grafana/toolkit@next plugin:create dataource-be`
2. Run `yarn install`
3. Run `yarn build`
4. See Typescript errors being thrown

**Environment**:
- Grafana version: doesn't matter
- Data source type & version:  doesn't matter
- OS Grafana is installed on: Mac OS Catalina (latest)
- User OS & Browser: Chrome
- Grafana plugins: -
- Others: -

dprokop:
Problem introduced most probably with field color mode type changes in https://github.com/grafana/grafana/pull/28039

dprokop:
@jackw confirms similar issue on latest channel

dprokop:
@jackw I guess when u used latest you might have encountered this: https://github.com/grafana/grafana/issues/28341 ?

jackw:
> @jackw I guess when u used latest you might have encountered this: #28341 ?

Yep, that's the error I was getting. 👍

scraton:
A solution I found from [here](https://github.com/grafana/github-datasource/blob/master/package.json) suggests the need to remove `node_modules/@grafana/data/node_modules`. I guess the types from that package interfere with the build process? Setting up my `package.json` to following got the build working for `7.2.2`:

```json
  "scripts": {
    "build": "rm -rf node_modules/@grafana/data/node_modules; grafana-toolkit plugin:build",
    "test": "rm -rf node_modules/@grafana/data/node_modules; grafana-toolkit plugin:test",
    "dev": "rm -rf node_modules/@grafana/data/node_modules; grafana-toolkit plugin:dev",
    "watch": "rm -rf node_modules/@grafana/data/node_modules; grafana-toolkit plugin:dev --watch"
  },
```

marefr:
Is this the same as https://github.com/grafana/grafana-starter-datasource-backend/issues/16?

dprokop:
[...]

PR #28657, opened 2020-10-29
Bump rxjs to 6.6.3

dprokop:
Addresses https://github.com/grafana/grafana/issues/28395

There is a rxjs type mismatch in https://github.com/grafana/grafana-starter-datasource-backend. It seems to be caused by concurrently package installing rxjs@6.6.3 while grafana-data requires 6.6.2. When aligning the versions the problem is gone.

@hugohaggmark @torkelo not sure if we want to include this in a patch release. As a temporary workaround until 7.4 is released I can force the toolkit starter template to resolve rxjs@6.6.2. This should not influence the way toolkit(which has dependency conconcurrently) works. WDYT?

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,7 +282,7 @@
     "regenerator-runtime": "0.13.3",
     "reselect": "4.0.0",
     "rst2html": "github:thoward/rst2html#990cb89",
-    "rxjs": "6.6.2",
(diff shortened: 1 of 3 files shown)


torkelo (approved):
Sounds like good plan

PR #28663, opened 2020-10-29
Drone: Don't build on Windows for PRs

aknuds1:
Disable building of PRs on Windows, since it can really slow them down, due to us having few and slow Windows runners.


wbrowne (approved):
LGTM

PR #28670, opened 2020-10-29
Build: changing docs docker image to prevent setting up frontend devenv.

mckn:
To prevent people from needing to setup their environment to be able to generate the frontend docs I have added a new image that will include them within the image. So all you need to do is basically just to run the container and mount the volume as you did prior the the auto generated docs.

This has a dependency on this PR: https://github.com/grafana/website/pull/2744 and that we publish a version of the `grafana/docs-devenv` image to docker hub.


torkelo (approved):
Looks good!

oddlittlebird:
You rock, @mckn !

dprokop (approved):

PR #28682, opened 2020-10-29
PanelHeader: show streaming indicator (and allow unsubscribe) 

ryantxu:
This replaces #27349

We have a good loading indicator, but nothing that indicates data is still streaming.  I suggest a simple dot or something.  The best existing icon i see is circle: 

image


ryantxu:
@Ijin08  https://github.com/grafana/grafana/pull/27349#issuecomment-686275485
>  The dot feels a bit dead and inactive. If blinking pulsating feels a bit too much maybe simple a green dot. As @dprokop pointed out we can have access to if we want.
What happen if one unsubscribes? Can one click to resubscribe?

If you unsubscribe, new events stop coming in, and the spinner/dot go away -- this is the same as the loading state "Loading" but that shows a spinner..

Can you resubscribe?  Yes, with the refresh button (same as for the Loading state)

ryantxu:
@Ijin08 @jessover9000 -- I get your feedback on something more than just a dot and am totally open to suggestions... but want to make sure that it is something subtle since it will likely need to exist in the corner of every panel on a dashboard -- too much spinning/pulsing will be super overwhelming.

torkelo:
Been testing this the tooltip is not working well, keeps hiding on render, investigating

torkelo:
Did this but feels to strong / distracting? 
Screenshot from 2020-10-30 15-50-15
Screenshot from 2020-10-30 15-48-46

ryantxu:
with one panel anything is fine!  I am mostly worried about dashboards where everything it live -- What we really need is:
* consistent way to unsubscribe
* know that should be streaming, but is not  (kinda the opposite of what we have now)

that is why i landed on as subtle as possible ;)

ryantxu:
@torkelo @dprokop -- i think this is an improvement and would like to merge.  Yes the dot is subtle and meaningless, but we also do not work it to be too overwhelming

torkelo (approved):

PR #28585, opened 2020-10-27
Backend: Fix IPv6 address parsing erroneous

taciomcosta:
**What this PR does / why we need it**:

It changes IPv6 parsing to accept ips that are not surround by brackets, it is, that don't include ports.

**Example**:
2001:608:0:a00::102d becomes acceptable

**Which issue(s) this PR fixes**:

28203 - IPv6 adress parsing errornous

Fixes #28203 

**Special notes for your reviewer**:

PR #28688, opened 2020-10-30
WIP: Backend: Fix parsing of client IP address

### LINKED ISSUE #28203: IPv6 adress parsing erroneous
**What happened**:
After turning logging to Debug i noticed that IPv6 is not handled properly:

t=2020-10-12T18:17:38+0200 lvl=dbug msg="Failed to parse client IP address" logger=auth clientAddr=2001:608:0:a00::102d err="failed to split network address \"2001:608:0:a00::102d\" by host and port: net.SplitHostPort failed for '2001:608:0:a00::102d': address 2001:608:0:a00::102d: too many colons in address"

Proper handling of IPv6 Adresses

Turn log level to debug, Login to grafana via IPv6 connection

**Environment**:
- Grafana version: Grafana v7.2.1 (72a6c64532)
- Data source type & version: does not apply to auth section
- User OS & Browser: Fedora 32 / Chrome 86.0.4240.75 + Firefox 81.0.1
- Grafana plugins: core + Grafana Image Renderer
- Others:

aknuds1:
Fix parsing of client address in HTTP requests. There are currently some bugs in our parsing of IPv6 addresses, f.ex. #28203. I think this is the right way to deal with the problem, i.e. convert the string supplied by Macaron into a valid IP address or `nil`. The PR also contains a comprehensive suite of tests for all the cases I could think of.

I propose this as an improvement to #28585, and have asked the author to merge in these changes.

Fixes #28203.


marefr (commented):
Tested login quickly and seems like it stores a null in the database now together with visualizing a null in the UI. Not sure what makes most sense, but earlier I think it defaulted to 127.0.0.1 in this case.

```
INFO[11-17|15:51:27] Successful Login                         logger=http.server User=admin@localhost
INFO[11-17|15:51:27] New token                                logger=context userId=0 orgId=1 uname= unhashed token=8a409e2cf38b4fc224edce657768a41e
INFO[11-17|15:51:27] Request Completed                        logger=context userId=0 orgId=1 uname= method=POST path=/login status=200 remote_addr=[::1] time_ms=41 size=23 referer=http://localhost:3000/login
```

image

image

aknuds1:
@marefr Interesting, I will have a look when time allows.

aknuds1:
> Tested login quickly and seems like it stores a null in the database now together with visualizing a null in the UI. Not sure what makes most sense, but earlier I think it defaulted to 127.0.0.1 in this case.

@marefr Thanks for catching this. On investigating, I found that on my local machine, my address was passed as "[::1]". This, an IPv6 address enclosed in square brackets, wasn't parsed and led to a nil IP address because of the square brackets. I think strictly speaking, an IPv6 address shouldn't be enclosed in square brackets unless it's combined with a port (f.ex. [::1]:1000), but since it happens in practice I've modified the code to support it.

I also realized that the old code would convert an invalid address to an empty string (not 127.0.0.1 from what I can tell, but please point out if you find evidence of the opposite). Therefore I fixed up this branch to convert `nil` IP addresses to empty strings instead of "<nil>".

marefr (approved):
Great, LGTM

Verified both locally and from my mobile phone and for mobile phone it picked up the local network IP address.

image

PR #28690, opened 2020-10-30
TableFilters: Fixes filtering with field overrides

### LINKED ISSUE #28598: Table filter with override value mapping
I have applied a value override with the value mapping range to text in the table plugin. Two value mapping is applied with text "ok" and "Not Ok" and it works fine in the table. But when I select "ok" in the table filter, only one row is listed even if there are serval rows with "ok".

Before filter "ok is selected":
image
image
After "Ok" is selected:
image

**Environment**:
- Grafana version: Version 7.2.2
- Data source type & version: Influxdb
- OS Grafana is installed on: Windows Azure
- User OS & Browser: Chrome, Edge
- Grafana plugins: Table

hugohaggmark:
Fixes issue as described in #28598.

Fixes #28598


hugohaggmark:
Merge backport PR after 7.3.1 release please 🙏

packages/grafana-ui/src/components/Table/utils.ts
```diff
@@ -143,16 +148,26 @@ export function calculateUniqueFieldValues(rows: any[], field?: Field) {
...
 
   return set;
 }
 
+export function rowToFieldValue(row: any, field?: Field): string {
+  if (!field || !row) {
+    return '';
+  }
+
+  const index = parseInt(row.id, 10);
```
torkelo, line 163: there is an index prop on a row here any reason not to use that?

torkelo (approved):
Looks good

packages/grafana-ui/src/components/Table/utils.ts
hugohaggmark: I didn't know that, is index a more correct way to access the __`column index`__?

dprokop: yes, react-table's Row has in index. But, to get the **column index** we usually keep a counter in the fields loop and pass that counter downstream.

dprokop (approved):
I really like the test coverage here Hugo!

packages/grafana-ui/src/components/Table/utils.ts
hugohaggmark: Thank you for clarifying, I've changed it to index instead.

PR #28702, opened 2020-10-30
Dashboard: fix view panel mode for Safari / iOS

### LINKED ISSUE #28695: Grafana v7.3.0 - unable to view single graph on iOS
**What happened**:
Right clicking a graph in a dashboard and selecting "View" results in a blank screen.

A view of the single graph.

Right clicking a graph and selecting "View" results in a blank screen using any apple iOS device.

Still works perfectly fine on a desktop PC running Firefox. Issue only occurred after updating to Grafana v7.3.0

Dashboards still display correctly. It is only the single graph view.

**Environment**:
- Grafana version: 7.3.0
- Data source type & version: InfluxDB 1.8.3
- OS Grafana is installed on: HassOS 4.15 (buildroot)
- User OS & Browser: iOS 13.7, Safari
- Grafana plugins: Carpet Plot.
- Others:

jackw:
iOS / Safari users are reporting when setting dashboard to view panel mode they're left with an empty screen. This PR addresses this bug plus fixes an issue where the dashboard scroll position was affecting the positioning of the panel.

| Before | After |
| --- | --- |
| Empty Panel View |
| before | after |
| Scroll Position | 
| before | after |

Fixes #28695

Tested on Safari on mac and iOS. The issue mentions graphs but from testing it appears to affect all panel types.

### DIFF
--- a/public/app/features/dashboard/containers/DashboardPage.tsx
+++ b/public/app/features/dashboard/containers/DashboardPage.tsx
@@ -145,6 +145,7 @@ export class DashboardPage extends PureComponent<Props, State> {
         this.setState({
           viewPanel: panel,
           rememberScrollTop: this.state.scrollTop,
+          updateScrollTop: 0,
         });
       });
     }
--- a/public/sass/pages/_dashboard.scss
+++ b/public/sass/pages/_dashboard.scss
@@ -18,6 +18,7 @@
 
 .dashboard-content {
   padding: $dashboard-padding;
+  flex-basis: 100%;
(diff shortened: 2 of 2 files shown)


torkelo (approved):
Looks good!

nikobearrr:
Hi, I am also observing the same issue on Windows10 with Chrome 83.0.1403.106 (64 bit). That was not an issue with 7.2.2
Is this fix going to address win/chrome or is it iOS specific?

Issue #2748, opened 2015-09-15
Dashboard Level Datasource


jsabin:
When developing a new product, we use dashboards in a development environment. When the product is put in production (hosting model), we switch our dashboards from a development datasource to a production datasource. It would be nice to be able to do this on a per dashboard level rather than having to change each graph in every dashboard.

Perhaps this request could be fulfilled using this feature request:
Datasource as variable? #816

torkelo:
not ideal, but a temp work around, export the dashboard as json, search replace in editor, then import it :)

jsabin:
That's yucky.

One thought I had is for each graph you could specify a data source or specify "dashboard data source". Then in the dashboard properties you could specify a data source.

jozic:
+1

utkarshcmu:
+1 

I agree its not an ideal case but it can be a good feature if we can set datasource on the dashboard level.

utkarshcmu:
@torkelo I think this can be a really useful feature. Here are the use cases:
1. Testing multiple datasources behind Grafana.
2. If one datasource dies, then we do not have to go to each and every panel and change the name of the datasource. We can just go to dashboard settings and change the datasource for the whole dashboard. [I know we can always change the datasource URLs etc in the datasources].
3. When running a backup datasource and a single switch can point to the other datasource. 
4. May be two different datasources show views of two different systems but have same queries.

In all these cases, this feature will be useful. Are u willing to accept a PR which implements this feature?

feraudet:
+1
Really need this feature

torkelo:
New template variable is now available (Data source)
![image](https://cloud.githubusercontent.com/assets/10999/14895241/f756942e-0d77-11e6-910a-091c0c470caf.png)

You pick the data source type and the variable will be automatically populated with all your data sources of that type. You can also provide a regex filter. When you add a new data source it will be automatically available in the variable values dropdown (if it passes the regex filter that is)

You can pick the variable in the panel data source selection:
![image](https://cloud.githubusercontent.com/assets/10999/14895284/279246ce-0d78-11e6-8cb5-acd25b54a9cc.png)

You can also have other query template variables use the template variable: 
![image](https://cloud.githubusercontent.com/assets/10999/14895305/4e1450c6-0d78-11e6-9d31-8d0fcf702d58.png)

Aakansha-Talati:
[...]

PR #28708, opened 2020-10-30
Same changes as in PR 28679, for Diana's approval

achatterjee-grafana:
Updated "Share a Panel" topic.

Added section "Publish Snapshots" and a new image (image added to website repo: Add new image #2748).

From Ximena, for whom I created the PR:

The doc mentions Embed Panel and Direct Link but really should be a section for the Snapshot option which is the easiest of them all. ...

There's technically 3 ways to share a panel so having all 3 in that doc makes sense. IMHO "Publish Snapshots" should be at the top due to how 1-2-3 easy it is to do.

### DIFF
--- a/docs/sources/sharing/share-panel.md
+++ b/docs/sources/sharing/share-panel.md
@@ -10,7 +10,20 @@ draft = "true"
 
 # Share a panel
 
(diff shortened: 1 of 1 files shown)


oddlittlebird (approved):
LGTM!

PR #28716, opened 2020-10-31
Removed auto-capitalization from the login form

tabarra:
On mobile browsers, by default most keyboards will auto capitalize the first letter of the email/username, which is case sensitive.  
This PR fixes that.

Auto capitalization of the login form in mobile phones.


torkelo (approved):

torkelo:
Thanks for contributing!

PR #28719, opened 2020-11-01
ColorScheme: Add Red to Green color scheme

peterholmberg:
Adds another ColorScheme, red to green for situations where a low value should be red and high value should be green. The use case I have is disk space (see screenshot).

Screenshot from 2020-11-01 14-20-01


torkelo:
I removed the inverted ones thinking that having separate inverted ones would create too many schemes. The solution could be a more custom scheme picker where you can invert any scheme. Some tricky UX for that though

peterholmberg:
Yes, was also thinking about a "invert colors" checkbox or something similar. Can see how adding more and more color schemes won't scale

ryantxu (approved):
I think this is fine, but lets be careful not to add all of them ;) having an invert would be nice

PR #28736, opened 2020-11-02
Dependencies: Update angularjs to 1.8.2

torkelo:
This update removes the feature of pre assigned bindings. In order not to break all old angular plugins I have monkey patched $injector.invoke to handle the common plugin controller cases that makes sure they will still work. 

#### Plugin compatibility 

We have upgraded AngularJS from version 1.6.6 to 1.8.2. Due to this upgrade some old angular plugins might stop working and will require a small update. This is due to the deprecation and removal of pre-assigned bindings. So if your custom angular controllers expect component bindings in the controller constructor you need to move this code to an `$onInit` function. For more details on how to migrate AngularJS code open the [migration guide](https://docs.angularjs.org/guide/migration) and search for **pre-assigning bindings**.

In order not to break all angular panel plugins and data sources we have some custom [angular inject behavior](https://github.com/grafana/grafana/blob/master/public/app/core/injectorMonkeyPatch.ts) that makes sure that bindings for these controllers are still set before constructor is called so many old angular panels and data source plugins will still work.


dprokop (approved):
@torkelo I have tested this and haven't found problems so far in the core Grafana. Great work.

dprokop:
@daniellee maybe you could also have a look with your instance full of plugins?

PR #28739, opened 2020-11-02
Transformations: Add new grouping to matrix transformer

### LINKED ISSUE #7169: [Feature request] Table transform to both rows and columns
- What Grafana version are you using? 4.0.1
- What datasource are you using? influxdb
- What OS are you running grafana on? windows

I'm trying to build a table from an influxdb query. The query groups by 2 tags (with say m and n unique values respectively). 

If I understand correctly, at the moment I can either display the data as columns or rows where there is a column/row for each combination of the two tag values (m x n values). 

It would be amazing if there could something like a pivot option where the user could choose to use some tags on rows and some tags on columns so that rather than a 1 X (m x n) table you could get an m X n table.

Hope this makes sense!

kazeborja:
**What this PR does / why we need it**: It adds a new transformer that allows to structure data in a "matrix" wise. This feature has been requested by an open issue and current implementations always miss some datasources.

Being a transformer it's kind of datasource independant and doesn't require plugin updates.

Fixes #7169 

- This transformer is meant to work with the Grafana table plugin
- Adding support for variables in the transformers framework will make it much more "non editor" friendly as users may select the "columns" and "rows" from the templating.
- No big expertisement in the program language neither Grafana internals have been used for the development, so it may lack some basics.

### DIFF
--- a/packages/grafana-data/src/transformations/transformers/groupingToMatrix.test.ts
+++ b/packages/grafana-data/src/transformations/transformers/groupingToMatrix.test.ts
@@ -5,13 +5,14 @@ import { toDataFrame } from '../../dataframe';
 import { transformDataFrame } from '../transformDataFrame';
 import { ArrayVector } from '../../vector';
 import { groupingToMatrixTransformer, GroupingToMatrixTransformerOptions } from './groupingToMatrix';
+import { observableTester } from '../../utils/tests/observableTester';
 
 describe('Grouping to Matrix', () => {
   beforeAll(() => {
     mockTransformationsRegistry([groupingToMatrixTransformer]);
   });
 
-  it('generate Matrix with default fields', () => {
+  it('generate Matrix with default fields', done => {
     const cfg: DataTransformerConfig<GroupingToMatrixTransformerOptions> = {
       id: DataTransformerID.groupingToMatrix,
       options: {},
@@ -25,18 +26,23 @@ describe('Grouping to Matrix', () => {
       ],
     });
 
-    const result = transformDataFrame([cfg], [seriesA]);
-    const expected: Field[] = [
-      createField('Time\\Time', FieldType.string, [1000, 1001, 1002]),
-      createField('1000', FieldType.number, [1, '', '']),
-      createField('1001', FieldType.number, ['', 2, '']),
-      createField('1002', FieldType.number, ['', '', 3]),
-    ];
+    observableTester().subscribeAndExpectOnNext({
+      observable: transformDataFrame([cfg], [seriesA]),
+      expect: result => {
+        const expected: Field[] = [
+          createField('Time\\Time', FieldType.string, [1000, 1001, 1002]),
+          createField('1000', FieldType.number, [1, '', '']),
+          createField('1001', FieldType.number, ['', 2, '']),
+          createField('1002', FieldType.number, ['', '', 3]),
+        ];
 
-    expect(unwrap(result[0].fields)).toEqual(expected);
+        expect(unwrap(result[0].fields)).toEqual(expected);
+      },
+      done,
+    });
   });
 
-  it('generate Matrix with multiple fields', () => {
+  it('generate Matrix with multiple fields', done => {
     const cfg: DataTransformerConfig<GroupingToMatrixTransformerOptions> = {
       id: DataTransformerID.groupingToMatrix,
       options: {
@@ -55,17 +61,22 @@ describe('Grouping to Matrix', () => {
       ],
     });
 
-    const result = transformDataFrame([cfg], [seriesA]);
-    const expected: Field[] = [
-      createField('Row\\Column', FieldType.string, []),
-      createField('C1', FieldType.number, [1, 4, '']),
-      createField('C2', FieldType.number, [5, '', '']),
-    ];
+    observableTester().subscribeAndExpectOnNext({
+      observable: transformDataFrame([cfg], [seriesA]),
+      expect: result => {
+        const expected: Field[] = [
+          createField('Row\\Column', FieldType.string, ['R1', 'R2']),
+          createField('C1', FieldType.number, [1, 4]),
+          createField('C2', FieldType.number, [5, '']),
+        ];
 
-    expect(unwrap(result[0].fields)).toEqual(expected);
+        expect(unwrap(result[0].fields)).toEqual(expected);
+      },
+      done,
+    });
   });
 
-  it('generate Matrix with multiple fields and value type', () => {
+  it('generate Matrix with multiple fields and value type', done => {
     const cfg: DataTransformerConfig<GroupingToMatrixTransformerOptions> = {
       id: DataTransformerID.groupingToMatrix,
       options: {
@@ -84,14 +95,19 @@ describe('Grouping to Matrix', () => {
       ],
     });
 
-    const result = transformDataFrame([cfg], [seriesA]);
-    const expected: Field[] = [
-      createField('Row\\Column', FieldType.string, []),
-      createField('C1', FieldType.number, [1, 4, ''], { units: 'celsius' }),
-      createField('C2', FieldType.number, [5, '', ''], { units: 'celsius' }),
-    ];
+    observableTester().subscribeAndExpectOnNext({
+      observable: transformDataFrame([cfg], [seriesA]),
+      expect: result => {
+        const expected: Field[] = [
+          createField('Row\\Column', FieldType.string, ['R1', 'R2']),
+          createField('C1', FieldType.number, [1, 4], { units: 'celsius' }),
+          createField('C2', FieldType.number, [5, ''], { units: 'celsius' }),
+        ];
 
-    expect(unwrap(result[0].fields)).toEqual(expected);
+        expect(unwrap(result[0].fields)).toEqual(expected);
+      },
+      done,
+    });
   });
 });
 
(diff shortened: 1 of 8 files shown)


kazeborja:
Current failing bit in the continuous-integration pipeline I'm not super sure what causes it... may need some extra help there.

kazeborja:
pinging for help/review :)

kazeborja:
New ping

zarbis:
New ping.

kazeborja:
New ping

kazeborja:
I'm back to work on this and have started looking into the field selector... closed thing I could find to what I'd like to achieve seems the SortBy behaviour, but it uses a lot of internal structures to handle the combobox value, so I'm not sure it's the way to go... may I get some guidance on how to achieve this properly?

kazeborja:
And now everything seems to be green for me :) (note for myself always rebase with main before any changes)

kazeborja:
Hi, I see things moving but not many updates on this PR itself... I'm afraid it will keep getting outdated with time and it will be harder and harder to make sure it's up to date.

Is there anything missing that I should look into in order to help speeding up the process?

public/app/core/components/TransformersUI/GroupingToMatrixTransformerEditor.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+      <div className="gf-form gf-form--grow">
+        <div className="gf-form-label width-8">Row</div>
+        <Select menuShouldPortal options={fieldNames} value={options.rowField} onChange={onSelectRow} isClearable />
+      </div>
+      <div className="gf-form gf-form--grow">
+        <div className="gf-form-label width-8">Cell Value</div>
+        <Select menuShouldPortal options={fieldNames} value={options.valueField} onChange={onSelectValue} isClearable />
+      </div>
+    </div>
```
ashharrison90, line 70: think specifying the `gf-form` classes is deprecated now, take a look at https://github.com/grafana/grafana/blob/main/public/app/features/transformers/prepareTimeSeries/PrepareTimeSeriesEditor.tsx#L67 for a nicer way 👍

ashharrison90 (commented):
hi @kazeborja 👋 

really sorry we haven't looked at this until now. it kinda fell through the cracks and never ended up on a team's project board. i promise we're not usually this bad at reviewing external contributions 😅 

the code itself looks good! the transformers code has moved slightly (see https://github.com/grafana/grafana/pull/45497 - it looks big but it's really just moved a lot of files from `app/core/components/TransformersUI/` to `app/features/transformers/editors/`) which is why there are now conflicts

i've also added a tiny comment about the jsx in the editor.

kazeborja:
Thanks for the reply, will try to have a look to it and update the code as required :)

kazeborja:
I've changed the code accordingly and it seems to keep working in my local installation.
For the CI I guess someone from Grafana needs to add the needed labels/allow the needed workflows :)

public/app/features/transformers/editors/GroupingToMatrixTransformerEditor.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useCallback } from 'react';
+import {
+  DataTransformerID,
+  SelectableValue,
+  standardTransformers,
+  TransformerRegistryItem,
+  TransformerUIProps,
+} from '@grafana/data';
+import { InlineField, InlineFieldRow, Select } from '@grafana/ui';
+import { GroupingToMatrixTransformerOptions } from '@grafana/data/src/transformations/transformers/groupingToMatrix';
```
ashharrison90, line 10: instead of importing from a subpath, can you expose this method in the `@grafana/data` index the same way as valueMatchers? https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/index.ts#L23

```suggestion
import { GroupingToMatrixTransformerOptions } from '@grafana/data';
```

ashharrison90 (commented):
have approved the workflows and set the correct labels - minor nit on an import otherwise lgtm 👍

public/app/features/transformers/editors/GroupingToMatrixTransformerEditor.tsx
kazeborja: Sure thing, seems a bit incosistent with other transformers tho

ashharrison90: mmmm agreed, there's a bit of inconsistency here. it doesn't really matter in the end as they both work. my gut is that anything we want to export from `@grafana/data` should be exposed on the index, then the public interface is clear. wonder if @ryantxu has an opinion? 😄

ashharrison90:
@mckn @ryantxu any objections to merging this? 🤔

packages/grafana-data/src/transformations/transformers/groupingToMatrix.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        { name: 'Value', type: FieldType.number, values: [1, 2, 3] },
+      ],
+    });
+
+    await expect(transformDataFrame([cfg], [seriesA])).toEmitValuesWith((received) => {
+      const processed = received[0];
+      const expected: Field[] = [
+        {
+          name: 'Time\\Time',
+          type: FieldType.string,
```
mckn, line 37: Are there any reason why we don't want to keep the `FieldType.time` here?

mckn (approved):
I left one question but overall I think it looks good. Great work 🙌🏻

public/app/features/transformers/editors/GroupingToMatrixTransformerEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+            options={fieldNames}
+            value={options.columnField}
+            onChange={onSelectColumn}
+            isClearable
+          />
+        </InlineField>
+        <InlineField label="Row" labelWidth={8}>
+          <Select menuShouldPortal options={fieldNames} value={options.rowField} onChange={onSelectRow} isClearable />
+        </InlineField>
+        <InlineField label="Cell Value" labelWidth={8}>
```
mckn, line 65:
```suggestion
        <InlineField label="Cell Value" labelWidth={10}>
```

mckn: Skärmavbild 2022-02-24 kl  11 35 09

packages/grafana-data/src/transformations/transformers/groupingToMatrix.test.ts
kazeborja: Not really, but since it's used for "Column/Rows" names I thought forcing String made sense

ashharrison90:
@kazeborja think you'll need to sync with latest main to get the CI to pass 👍

packages/grafana-data/src/transformations/transformers/groupingToMatrix.ts
```diff
@@ -0,0 +1,105 @@
...
+        const keyRowField = findKeyField(frame, rowFieldMatch);
+        const valueField = findKeyField(frame, valueFieldMatch);
+        const rowColumnField = `${rowFieldMatch}\\${columnFieldMatch}`;
+
+        if (!keyColumnField || !keyRowField || !valueField) {
+          return data;
+        }
+
+        const columnValues = [...new Set(keyColumnField.values.toArray())];
+        const rowValues = [...new Set(keyRowField.values.toArray())];
```
mckn, line 51: It might be a good idea to change these to to something like the example below because it will skip one copy of the values array from being created.

```suggestion
        const columnValues = uniqueValues(keyColumnField.values);
        const rowValues = uniqueValues(keyRowField.values);
```
And add a helper function like this:

```ts
function uniqueValues(values: Vector): any[] {
  const unique = new Set();

  for (let index = 0; index < values.length; index++) {
    unique.add(values.get(index));
  }

  return Array.from(unique);
}
```

packages/grafana-data/src/transformations/transformers/groupingToMatrix.test.ts
mckn: Ok ok, sounds good!

packages/grafana-data/src/transformations/transformers/groupingToMatrix.ts
mckn: I think we probably could make this transformation even more efficient but this is good enough for now.

kazeborja: I've applied the suggested change and strongly agree with the possibility for improvements :)

ashharrison90:
@kazeborja looks like the rebase might have gone slightly awry 😅

ashharrison90 (approved):
lgtm 🙌

ashharrison90 (approved):
lgtm 🙌

ashharrison90:
thanks for the great first contribution @kazeborja! 🙌

torkelo:
@ashharrison90 this was merged without any docs or examples, how are users to know what this does?

torkelo:
was not added to any changelog or what's new

torkelo:
wait it has add to changelog, but is not in the changelog, strange

ashharrison90:
@torkelo ah, good point about the docs. @kazeborja is that something you'd be willing to contribute? 😄  the docs live here: https://github.com/grafana/grafana/blob/main/docs/sources/panels/transform-data/transformation-functions.md#L417

yeah idk what happened with the changelog as it definitely had the label 😕

kazeborja:
Happy to drop some lines for the docs :)

kazeborja:
https://github.com/grafana/grafana/pull/49316 don't hesitate to comment there with anything that might be needed

PR #28749, opened 2020-11-02
devenv: Upgrade MSSQL Docker image

aknuds1:
Upgrade MSSQL Docker image in devenv block to version 2019. I figure there's no reason to stay on 2017?


papagian (approved):
I've tested and works as expected.

PR #28754, opened 2020-11-02
Docs: Troubleshoot starting docker containers on Mac

jackw:
Introduces a troubleshooting section to the README with instructions to help a developer running Mac OS X spin up docker containers that mount `/var/log` which can fail due to permission issues.


ivanahuckova (approved):
LGTM! This is great and it might save a loooooot of time to developers. Great that you have added it! 🙏 🔝

PR #28758, opened 2020-11-02
Disable selecting enterprise plugins with no license

Clarity-89:

- Disable selecting enterprise plugins with no enterprise license
- Make plugins `Learn more` url customisable

Fixes https://github.com/grafana/grafana-enterprise/issues/827


xlson:
@Clarity-89 right. Perhaps we need to move this completely to the frontend then? We have logic in buildCategories.ts that knows if a plugin is an enterprise plugin or not. It's not quite as clean but it should work if we can break out and re-use that information right?

Clarity-89:
@xlson I guess we could move this logic to the front-end and use the licensing data from the config object 🤔

Clarity-89:
@xlson I moved the edition info that you added and `hasValidLicense`(the existing `hasLicense` doesn't seem to care if the license is valid) to front end and it seems to be working now. Tested with both valid and invalid license.

pkg/services/licensing/oss.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/grafana/grafana/pkg/setting"
 )
 
+const (
+	OpenSource = "Open Source"
```
aknuds1, line 11: Please un-export this one, since it isn't used outside of the package.
```suggestion
	openSource = "Open Source"
```

pkg/services/licensing/oss.go
```diff
@@ -21,7 +25,7 @@ func (*OSSLicensingService) Expiry() int64 {
 }
 
 func (*OSSLicensingService) Edition() string {
-	return "Open Source"
+	return OpenSource
```
aknuds1, line 28:
```suggestion
	return openSource
```

pkg/setting/setting.go
```diff
@@ -273,6 +273,7 @@ type Cfg struct {
 	PluginsAppsSkipVerifyTLS bool
 	PluginSettings           PluginSettings
 	PluginsAllowUnsigned     []string
+	MarketPlaceUrl           string
```
aknuds1, line 276: [Go initialisms should have consistent case](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
```suggestion
	MarketPlaceURL           string
```

pkg/setting/setting.go
```diff
@@ -796,6 +797,7 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 		plug = strings.TrimSpace(plug)
 		cfg.PluginsAllowUnsigned = append(cfg.PluginsAllowUnsigned, plug)
 	}
+	cfg.MarketPlaceUrl = pluginsSection.Key("marketplace_url").MustString("https://grafana.com/grafana/plugins/")
```
aknuds1, line 800:
```suggestion
	cfg.MarketPlaceURL = pluginsSection.Key("marketplace_url").MustString("https://grafana.com/grafana/plugins/")
```

pkg/api/frontendsettings.go
```diff
@@ -230,14 +230,17 @@ func (hs *HTTPServer) getFrontendSettingsMap(c *models.ReqContext) (map[string]i
...
+			"hasValidLicense": hs.License.HasValidLicense(),
+			"expiry":          hs.License.Expiry(),
+			"stateInfo":       hs.License.StateInfo(),
+			"licenseUrl":      hs.License.LicenseURL(c.SignedInUser),
+			"edition":         hs.License.Edition(),
 		},
 		"featureToggles":    hs.Cfg.FeatureToggles,
 		"rendererAvailable": hs.RenderService.IsAvailable(),
 		"http2Enabled":      hs.Cfg.Protocol == setting.HTTP2Scheme,
+		"marketPlaceUrl":    hs.Cfg.MarketPlaceUrl,
```
aknuds1, line 243:
```suggestion
		"marketPlaceUrl":    hs.Cfg.MarketPlaceURL,
```

conf/defaults.ini
```diff
@@ -769,6 +769,7 @@ enable_alpha = false
 app_tls_skip_verify_insecure = false
 # Enter a comma-separated list of plugin identifiers to identify plugins that are allowed to be loaded even if they lack a valid signature.
 allow_loading_unsigned_plugins =
+marketplace_url = https://grafana.com/grafana/plugins/
```
aknuds1, line 772: You need to update conf/sample.ini also. And maybe documentation also (unsure)?

aknuds1 (changes requested):
Mostly LGTM, but see some issues.

conf/defaults.ini
Clarity-89: This is the same value that was hardcoded on front end before, so need to check if it needs to be documented.

aknuds1: It sounds to me as if it should be documented, since it's a new configuration parameter.

aknuds1 (approved):
Code looks good to me, but the new config param should probably be documented.

conf/defaults.ini
xlson: I think we should document it. There's not really anything to lose from documenting it but not doing so might be confusing.

xlson (approved):
LGTM!

pkg/setting/setting.go
```diff
@@ -796,6 +797,7 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 		plug = strings.TrimSpace(plug)
 		cfg.PluginsAllowUnsigned = append(cfg.PluginsAllowUnsigned, plug)
 	}
+	cfg.MarketPlaceURL = pluginsSection.Key("marketplace_url").MustString("https://grafana.com/grafana/plugins/")
```
torkelo, line 800: nit picking: Marketplace is one word so p should not really be capitalized here

torkelo (approved):
LGTM

pkg/setting/setting.go
aknuds1: Agree with @torkelo, didn't notice myself.

packages/grafana-runtime/src/config.ts
```diff
@@ -62,6 +62,7 @@ export class GrafanaBootConfig implements GrafanaConfig {
   rendererAvailable = false;
   http2Enabled = false;
   dateFormats?: SystemDateFormatSettings;
+  marketPlaceUrl?: string;
```
aknuds1, line 65:
```suggestion
  marketplaceUrl?: string;
```

pkg/api/frontendsettings.go
```diff
@@ -230,14 +230,17 @@ func (hs *HTTPServer) getFrontendSettingsMap(c *models.ReqContext) (map[string]i
...
+			"hasValidLicense": hs.License.HasValidLicense(),
+			"expiry":          hs.License.Expiry(),
+			"stateInfo":       hs.License.StateInfo(),
+			"licenseUrl":      hs.License.LicenseURL(c.SignedInUser),
+			"edition":         hs.License.Edition(),
 		},
 		"featureToggles":    hs.Cfg.FeatureToggles,
 		"rendererAvailable": hs.RenderService.IsAvailable(),
 		"http2Enabled":      hs.Cfg.Protocol == setting.HTTP2Scheme,
+		"marketPlaceUrl":    hs.Cfg.MarketPlaceURL,
```
aknuds1, line 243:
```suggestion
		"marketplaceUrl":    hs.Cfg.MarketplaceURL,
```

packages/grafana-runtime/src/config.ts
torkelo, line 65:
```suggestion
  marketplaceUrl?: string;
```

public/app/features/datasources/state/buildCategories.ts
```diff
@@ -197,7 +203,7 @@ function getPhantomPlugin(options: GetPhantomPluginOptions): DataSourcePluginMet
       author: { name: 'Grafana Labs' },
       links: [
         {
-          url: 'https://grafana.com/grafana/plugins/' + options.id,
+          url: config.marketPlaceUrl + options.id,
```
torkelo, line 206:
```suggestion
          url: config.marketplaceUrl + options.id,
```

pkg/setting/setting.go
```diff
@@ -273,6 +273,7 @@ type Cfg struct {
 	PluginsAppsSkipVerifyTLS bool
 	PluginSettings           PluginSettings
 	PluginsAllowUnsigned     []string
+	MarketPlaceURL           string
```
aknuds1, line 276:
```suggestion
	MarketplaceURL           string
```

public/app/features/datasources/state/buildCategories.ts
aknuds1, line 206:
```suggestion
          url: config.marketplaceUrl + options.id,
```

aknuds1 (approved):

papagian (approved):
LGTM

PR #28760, opened 2020-11-02
Reorganization of content

achatterjee-grafana:
Moved examples of setup to correct location. 

After we added information on node_exporter, the setup section was split into two and the examples moved to an incorrect location.


marcusolsson (approved):
LGTM

PR #28761, opened 2020-11-02
Cloudwatch: Fix issue with field calculation transform not working properly with Cloudwatch data

### LINKED ISSUE #28678: Add field from calculation transformation not working on Cloudwatch Datasources since 7.3
**What happened**:
Add field from calculation transformation no longer aggregates series correctly when using cloudwatch as a datasource.

Add field from calculation is expected to work on cloudwatch as it does on other datasources and prior to 7.3.

Add any cloudwatch metrics query with 2 or more stats, or add more than 1 cloudwatch metrics query.
Go to Transform Tab > add a Add field from calculation transformation.

field_from_calculation_grafana_bug

I assume it is related to the data frame changes as it seems to treat everything as unique data frames.

**Environment**:
- Grafana version: Grafana Enterprise v7.3.0 (bf8837f6f8)
- Data source type & version: Cloudwatch built-in
- OS Grafana is installed on: Docker / ECS
- User OS & Browser: Windows 10, Firefox 82.0.2
- Grafana plugins: dozens of other plugins working OK
- Others:

torkelo:
Fixes #28678

Due to the new cloudwatch data frame code the frame is not using the same defined constant names for the time and value field so it's not recognized as time series by the
add field calc transform.

We have some constants defined in the frontend, https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/dataFrame.ts#L170 

Should have the same set in the backend, not sure where is best, probably in the SDK


aknuds1:
Don't you have to change lines 163 and 167 also?

aknuds1:
@torkelo Do you want me to push fixes to the branch?

torkelo:
@aknuds1 I pushed more fixes, need to update the go sdk as well, the backend should not set DisplayName in field config as that will make it impossible for user to change it via Field tab. There is new property DisplayNameFromDS that can be used to control naming from data sources without overwriting user-specified display name

aknuds1:
@torkelo Did you forget to push? I don't see any more commits.

aknuds1 (approved):
LGTM

torkelo:
@aknuds1 can you test and verify that this fixes the problem? And would be nice to get this into 7.3.2

aknuds1:
@torkelo I can test it.

aknuds1:
@torkelo Works from my testing! At least it behaves the same as v7.2.2.

PR #28765, opened 2020-11-03
Live: improve broadcast semantics and avoid double posting

ryantxu:
This PR replaces the not so clear `OnPublish` semantics with a more explicit function that checks if you can publish on a channel.

When writing data on a websocket channel, the server can only block passing the bytes to everyone else.  Our previous implementation would double publish events.

This also reduces some info attached to each user object -- when streaming lots of data we need to be more careful about the packet sizes and avoid unnecessary items


pkg/services/live/live.go
```diff
@@ -199,26 +194,8 @@ func (g *GrafanaLive) Init() error {
...
-		if err != nil {
-			logger.Debug("error reading user", "dto", dto)
-			ctx.Resp.WriteHeader(404)
-			return
-		}
-		logger.Info("Logged in user", "user", user)
-
 		cred := &centrifuge.Credentials{
 			UserID: fmt.Sprintf("%d", user.UserId),
-			Info:   jsonData,
```
ryantxu, line 221: I am removing this info for now -- it is not used and adds extra bytes to messages that do not need it!

pkg/services/live/features/dashboard.go
```diff
@@ -39,10 +39,9 @@ func (g *DashboardHandler) OnSubscribe(c *centrifuge.Client, e centrifuge.Subscr
 	return nil
 }
 
-// OnPublish called when an event is received from the websocket
-func (g *DashboardHandler) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
-	// TODO -- verify and keep track of editors?
-	return e.Data, nil
+// AllowBroadcast checks if a message from the websocket can be broadcast on this channel
+func (g *DashboardHandler) AllowBroadcast(c *centrifuge.Client, e centrifuge.PublishEvent) error {
+	return nil // tell other panels it is editing
```
papagian, line 44: I would suggest moving this comment in the function godoc.

pkg/services/live/features/measurements.go
```diff
@@ -33,9 +33,8 @@ func (m *MeasurementsRunner) OnSubscribe(c *centrifuge.Client, e centrifuge.Subs
...
 }
 
-// OnPublish is called when an event is received from the websocket.
-func (m *MeasurementsRunner) OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error) {
-	// currently generic... but should be stricter
-	// logger.Debug("Measurements runner got event on channel", "channel", e.Channel)
-	return e.Data, nil
+// AllowBroadcast checks if a message from the websocket can be broadcast on this channel
+func (m *MeasurementsRunner) AllowBroadcast(c *centrifuge.Client, e centrifuge.PublishEvent) error {
+	// currently allows anything... should replace with HTTP interface
```
papagian, line 38: Same as above, I would suggest moving this comment in the function godoc.

papagian (approved):
AFAICT looks good. I have a comment about the documentation.

pkg/models/live.go
```diff
@@ -13,8 +13,9 @@ type ChannelHandler interface {
 	// Called when a client wants to subscribe to a channel
 	OnSubscribe(c *centrifuge.Client, e centrifuge.SubscribeEvent) error
 
-	// Called when something writes into the channel.  The returned value will be broadcast if len() > 0
-	OnPublish(c *centrifuge.Client, e centrifuge.PublishEvent) ([]byte, error)
+	// Called when a client writes a message to the channel websocket
+	// returning an error will cancel the broadcast
```
aknuds1, line 17:
```suggestion
	// AllowBroadcast is called when a client writes a message to the channel websocket.
	// Returning an error will cancel the broadcast.
```

aknuds1 (approved):
LGTM with a nit.

PR #28769, opened 2020-11-03
Drone: Upgrade build pipeline tool

aknuds1:
Upgrade build pipeline tool in Drone, to get retries also for pushing of Docker manifests to Docker Hub (this sometimes flakes). Previous fix only implemented retries for Docker image pushing (I overlooked the manifests part).

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,7 +17,7 @@ steps:
   image: grafana/build-container:1.2.28
   commands:
   - mkdir -p bin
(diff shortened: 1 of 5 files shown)


wbrowne (approved):
Every time I see a grabpl version bump, I wonder if there's not an easy way to consolidate the source of truth for the version number. 

Looks good to me though!

aknuds1:
> Every time I see a grabpl version bump, I wonder if there's not an easy way to consolidate the source of truth for the version number.

@wbrowne I guess the solution is to automate the grafana/build-pipeline release process via Drone.

PR #28777, opened 2020-11-03
API: replace SendLoginLogCommand with LoginHook

AgnesToulet:
Replace the `SendLoginLogCommand` sent when a user logs in (successfuly or not) by a `LoginHook`. It's a more generic solution allowing to set multiple services to register to this hook and take actions independently

### DIFF
--- a/pkg/models/user_auth.go
+++ b/pkg/models/user_auth.go
@@ -36,6 +36,15 @@ type ExternalUserInfo struct {
 	IsDisabled     bool
 }
 
+type LoginInfo struct {
+	AuthModule   string
+	User         *User
+	ExternalUser ExternalUserInfo
+	Query        *LoginUserQuery
+	HTTPStatus   int
+	Error        error
+}
+
 // ---------------------
 // COMMANDS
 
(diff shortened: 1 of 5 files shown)


pkg/models/user_auth.go
```diff
@@ -36,6 +36,15 @@ type ExternalUserInfo struct {
 	IsDisabled     bool
 }
 
+type LoginInfo struct {
+	AuthModule   string
+	User         *User
+	ExternalUser ExternalUserInfo
+	Query        *LoginUserQuery
```
sakjur, line 43: Do we need to pass the password here?

Other than the IP address, what additional information do we get from sending the query in the hook?

xlson (approved):
Looks good!

sakjur (commented):
I'm not sure about using the same structure for logging in and the hook. There are advantages, but I'd like to strip out sensitive data from the request before sending it to the hook, could we do that somehow? 🤔

pkg/models/user_auth.go
AgnesToulet: The IP address can easily be retrieved from the request (and it's sent through the hook) so yes, I guess we only need the username used in the login form. I'll change that!

PR #28785, opened 2020-11-03
Adds storybook integrity check to drone config

dprokop:
To be merged after https://github.com/grafana/build-pipeline/pull/150


aknuds1 (changes requested):
You will need to make the changes in the Starlark code (scripts/*.star), since .drone.yml is auto-generated from that.

scripts/lib.star
```diff
@@ -272,6 +272,23 @@ def build_storybook_step(edition, ver_mode):
         ],
     }
 
+def verify_storybook_step(edition, ver_mode):
```
aknuds1, line 275: Could you include the command `./bin/grabpl verify-storybook' in the build-storybook and publish-storybook steps instead? I don't see any benefit to making its own step. Every step introduces some risk/complexity/performance degradation.

aknuds1 (commented):
You will also need to bump grabpl to 0.5.24, since I just uploaded that.

scripts/lib.star
```diff
@@ -298,7 +299,7 @@ def publish_storybook_step(edition, ver_mode):
         'name': 'publish-storybook',
         'image': publish_image,
         'depends_on': [
-            'build-storybook',
+            'verify-storybook',
```
torkelo, line 302: Is this right? it looks like it's part of the build step, not a seperate drone step

aknuds1: @torkelo You're right, that change shouldn't be made. I fixed it.

aknuds1 (approved):

aknuds1:
Great! :)

scripts/lib.star
dprokop: thanks @aknuds1

PR #19118, opened 2019-09-16
Update routing to require sign in on every route

dprokop:
Fixes #19003 

This make all routes require sign in by default. The issue mentioned in #19003 was caused by the fact that backend router was returning 404 on route `/%2f` which path was considered to be `/`. Instead, it should redirect to login.

When investigating this I have also noticed that for instance it is possible to navigate to `/dashboard/new` when not being singed in. That's because this route is not specified as the one that requires sign in. 

Not 100% sure, maybe there is a reason for allowing some routes to be accessible without login when anonymous access is disabled?

PR #28788, opened 2020-11-03
Backend: Adds route for well-known change password URL

kaydelaney:
This PR adds support for the [well-known](https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers) `/.well-known/change-password` URL, which browsers can use to point a user to the change password page if, for example, their password is compromised.

Previously unsupported routes were handled in `api.go` with the following:
```
r.Get("/*", reqSignedIn, hs.Index)
```

meaning that the not found handler established in `http_server.go` was never used, and consequently unsupported routes returned with HTTP status 200 rather than 404. Support for the well-known change-password URL relies on the web server returning status 404 for unknown routes, so I've change the route handling logic slightly in service of this.

More info on this topic: https://web.dev/change-password-url/

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -26,6 +26,7 @@ func (hs *HTTPServer) registerRoutes() {
 	r := hs.RouteRegister
 
 	// not logged in views
+	r.Get("/", reqSignedIn, hs.Index)
 	r.Get("/logout", hs.Logout)
 	r.Post("/login", quota("session"), bind(dtos.LoginCommand{}), Wrap(hs.LoginPost))
 	r.Get("/login/:name", quota("session"), hs.OAuthLogin)
@@ -35,6 +36,7 @@ func (hs *HTTPServer) registerRoutes() {
 	// authed views
 	r.Get("/profile/", reqSignedIn, hs.Index)
 	r.Get("/profile/password", reqSignedIn, hs.Index)
+	r.Get("/.well-known/change-password", RedirectToChangePassword)
 	r.Get("/profile/switch-org/:id", reqSignedIn, hs.ChangeActiveOrgAndRedirectToHome)
 	r.Get("/org/", reqOrgAdmin, hs.Index)
 	r.Get("/org/new", reqGrafanaAdmin, hs.Index)
@@ -442,6 +444,4 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/api/snapshots/:key", Wrap(GetDashboardSnapshot))
 	r.Get("/api/snapshots-delete/:deleteKey", reqSnapshotPublicModeOrSignedIn, Wrap(DeleteDashboardSnapshotByDeleteKey))
 	r.Delete("/api/snapshots/:key", reqEditorRole, Wrap(DeleteDashboardSnapshot))
-
-	r.Get("/*", reqSignedIn, hs.Index)
 }
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -73,6 +73,7 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/d/:uid", reqSignedIn, redirectFromLegacyPanelEditURL, hs.Index)
 	r.Get("/dashboard/db/:slug", reqSignedIn, redirectFromLegacyDashboardURL, hs.Index)
 	r.Get("/dashboard/script/*", reqSignedIn, hs.Index)
+	r.Get("/dashboard/new", reqSignedIn, hs.Index)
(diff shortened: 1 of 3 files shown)


pkg/api/api.go
```diff
@@ -26,6 +26,7 @@ func (hs *HTTPServer) registerRoutes() {
 	r := hs.RouteRegister
 
 	// not logged in views
+	r.Get("/", reqSignedIn, hs.Index)
```
aknuds1, line 29: Doesn't `reqSignedIn` require the client to be signed in? That contradicts "not logged in views", right? Why is this change necessary?

pkg/api/http_server.go
```diff
@@ -288,7 +288,7 @@ func (hs *HTTPServer) applyRoutes() {
 	// then custom app proxy routes
 	hs.initAppPluginRoutes(hs.macaron)
 	// lastly not found route
-	hs.macaron.NotFound(hs.NotFoundHandler)
+	hs.macaron.NotFound(middleware.ReqSignedIn, hs.NotFoundHandler)
```
aknuds1, line 291: Why is this change necessary?

pkg/api/user.go
```diff
@@ -250,6 +250,11 @@ func ChangeUserPassword(c *models.ReqContext, cmd models.ChangeUserPasswordComma
 	return Success("User password changed")
 }
 
+// GET /.well-known/change-password
```
aknuds1, line 253: Declaration doc comments should be sentences describing the thing in question:
```suggestion
// RedirectToChangePassword handles GET /.well-known/change-password.
```

aknuds1 (changes requested):
Requesting changes since I'm unsure about some of the modifications.

pkg/api/user.go
kaydelaney: Was following the pattern established by other functions in this file

aknuds1: Yeah, there's a lot of technical debt in the Grafana code base :) When we get proper doc comment linting enabled (it's a bit tricky) it'll catch this.

pkg/api/http_server.go
kaydelaney: Some time in the past it was discovered that there was a security issue allowing users to bypass the login screen to a degree by pointing the browser to an unsupported route. 
https://github.com/grafana/grafana/pull/19118

pkg/api/api.go
kaydelaney: Was reverting it back to a previous state (see https://github.com/grafana/grafana/blob/7b7b95341e73d0ac173e9ccef25cc129048b8701/pkg/api/api.go#L27) but I agree that it's a bit confusing

marefr: Seems like you're reverting #19118. Maybe need to verify with Dominik this change is intended/works as expected.

aknuds1:
I'll have to get back to this, will be AFK for a while.

pkg/api/api.go
aknuds1: Is the idea that #19118 can be reverted, so long as you make the change on line 291 of pkg/api/http_server.go?

@dprokop What do you think?

aknuds1:
@kaydelaney Do you know why e2e tests are failing?

aknuds1:
Also, has the motivation for this PR been stated, apart from the "well-known" protocol's existence?

pkg/api/api.go
torkelo: I think this looks fine as the not found handler now has req signed in check

torkelo (approved):
Looks good to me

pkg/api/api.go
dprokop: Yeah, think this is ok to revert that change. Let me double-check if we are not introducing the same bug again.

aknuds1: If we are reverting the change, can we move the route handler (`r.Get("/", reqSignedIn, hs.Index)`) outside the "not logged in views" section? It doesn't sound right to me.

dprokop: Checked - all good!

dprokop (approved):

aknuds1 (approved):
LGTM given Torkel and Dominik's comments, but please see my suggestions regarding un-exporting the route handler, since it doesn't need to be exported (every exported symbol makes maintenance harder), and that e2e tests are failing.

And I think we should move the index route handler registration outside of the "not logged in views" section, since it isn't right.

marefr (approved):
LGTM, but agree with Arve's comments

PR #28793, opened 2020-11-03
Dashboard Schemas

trotttrotttrott:

Initial move of schemas from [grafana/dashboard-spec](https://github.com/grafana/dashboard-spec). Instead of copying them in their current state as OpenAPI YAML, I've converted them to [Cue](https://cuelang.org/). Cue is much more concise and can be exported as an OpenAPI document ([more info on that here](https://cuelang.org/docs/usecases/datadef/#json-schema--openapi)). It also has other interesting integrations like with [Go](https://cuelang.org/docs/integrations/go/) which may prove useful for backend validation.

This is not a complete collection of schemas. This is only to propose the use of Cue and how to organize the schemas. The `README` has more info on this.


aknuds1:
Thanks, going to have a look!

aknuds1:
Sorry for the wait, got caught up in the meantime, trying to review again.

dashboard-schemas/Dashboard.cue
```diff
@@ -0,0 +1,68 @@
...
+	style: *"light" | "dark"
+	// Timezone of dashboard,
+	timezone?: *"browser" | "utc"
+	// Whether a dashboard is editable or not.
+	editable: bool | *true
+	// 0 for no shared crosshair or tooltip (default).
+	// 1 for shared crosshair.
+	// 2 for shared crosshair AND shared tooltip.
+	graphTooltip: int & >=0 & <=2 | *0
+	// Time range for dashboard, i.e. last 6 hours, last 7 days, etc
```
aknuds1, line 24: You probably want e.g., since they are example values:
```suggestion
	// Time range for dashboard, e.g. last 6 hours, last 7 days, etc
```

dashboard-schemas/Dashboard.cue
```diff
@@ -0,0 +1,68 @@
...
+		from: string | *"now-6h"
+		to:   string | *"now"
+	}
+	// Timepicker metadata.
+	timepicker?: {
+		// Whether timepicker is collapsed or not.
+		collapse: bool | *false
+		// Whether timepicker is enabled or not.
+		enable: bool | *true
+		// Whether timepcker is visible or not.
```
aknuds1, line 35:
```suggestion
		// Whether timepicker is visible or not.
```

dashboard-schemas/panels/Row.cue
```diff
@@ -0,0 +1,24 @@
+package panels
+
+// A row is a logical divider within a dashboard. It is used
+// to group panels together.
+#Row: {
+	// Whether the row is collapsed or not.
+	collapsed: bool | *true
+	// Name of default datasource.
```
aknuds1, line 8: I think we've standardized on writing data source in two words.
```suggestion
	// Name of default data source.
```

aknuds1 (approved):
Generally looks good as far as I can tell. Are these CUE files used to generate OpenAPI specs? How can I test them?

aknuds1:
@jdbaldry gave me this command for converting the CUE sources to OpenAPI output, but it's taking very long for some reason: `cue eval -o openapi+yaml:-  ./...`.

aknuds1 (changes requested):
I think something must be broken since the following fails: 

```
$ cue export --out openapi -o - ./dashboard-schemas/...
expected label or ':', found 'IDENT' Organize:
    ./dashboard-schemas/transformations/Organize.cue:4:4
```

trotttrotttrott:
@aknuds1 fixed the comments per your suggestions and fixed a syntax error in `Organize.cue`. The command you mention should succeed now.

Exporting these files into an OpenAPI doc would be compatible with the code generator in [grafana/dashboard-spec](https://github.com/grafana/dashboard-spec). However, I suspect there could be a cleaner implementation of the generator [processing Cue in Go](https://cuelang.org/docs/integrations/go/#processing-cue-in-go) which I haven't yet managed to pursue. Cue appears to me to have more potential for versatility within the Grafana codebase, but my rationale is based on theory at this point. More work needs to be done for determining exactly how it will work in practice. Two big use cases being: 1) how can plugin frontends leverage these as models 2) how can the backend leverage these for validation.

aknuds1:
@trotttrotttrott Thanks, now it works! :) I.e., I can export OpenAPI! 

```
$ cue export --out openapi -o - ./dashboard-schemas/..
{
    "openapi": "3.0.0",
    "info": {
        "title": "Generated by cue.",
        "version": "no version"
    },
[...]
}
```

aknuds1 (approved):
LGTM!

PR #28798, opened 2020-11-04
Variables: Fixes URL values for dependent variables

### LINKED ISSUE #28791: Template variables: URL values not used for dependent variables
**What happened**:

We have many dashboards with 2 template variables A and B, where B depends on A. In some cases, there are links like `<grafana_url>?var-A=val`, where B is left empty and should be filled in. After upgrading to `Grafana v7.3.1` we are seeing that the template value query for B is not using the value for A specified in the URL. 

For a concrete example, let's say we have 2 variables `parentCluster` and `cluster`. Given a link like `<grafana_url>?var-parentCluster=val` the behaviors are:

## Grafana 7.2.0 

`values?tag=parentCluster&limit=10000`
`expr=parentCluster%3Dval&tag=cluster&limit=10000`

Which results in the initial values of variables:
`parentCluster=val`
`cluster=cval` (picked the first returned value from the query)

## Grafana 7.3.1

`values?tag=parentCluster&limit=10000`
`expr=parentCluster%3Ddefaultval&tag=cluster&limit=10000`

where `defaultval` is the default value saved to the dashboard for parentCluster instead of the value passed in the URL. 

Which results in the initial values of variables:
`parentCluster=val`
`cluster=cdefaultval` (The saved default cluster)

I would expect that dependent variables would be resolved using the values passed in the URL

**Environment**:
- Grafana version: `Grafana v7.3.1 (6668161a88)`
- Data source type & version: Graphite (Metrictank)
- OS Grafana is installed on: Rhel 7
- User OS & Browser: Mac OS X & Chrome
- Grafana plugins:
- Others:

hugohaggmark:
As described by #28791. When fixing #28215 with #28257 I made a mistake when moving logic further down the update path. We should complete only those variables that are never refreshed in __`variableUpdated`__ .

I also added a integration test that makes sure that the order of actions are correct.

Fixes #28791

### DIFF
--- a/public/app/features/variables/shared/testing/multiVariableBuilder.ts
+++ b/public/app/features/variables/shared/testing/multiVariableBuilder.ts
@@ -6,4 +6,8 @@ export class MultiVariableBuilder<T extends VariableWithMultiSupport> extends Op
     this.variable.multi = multi;
     return this;
   }
+  withIncludeAll(includeAll = true) {
+    this.variable.includeAll = includeAll;
+    return this;
+  }
 }
--- a/public/app/features/variables/state/actions.test.ts
+++ b/public/app/features/variables/state/actions.test.ts
@@ -25,9 +25,16 @@ import {
   removeVariable,
   setCurrentVariableValue,
   variableStateCompleted,
+  variableStateFetching,
   variableStateNotStarted,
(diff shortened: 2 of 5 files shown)


torkelo (approved):
Looks good

PR #28806, opened 2020-11-04
Stat panel: Adds option to always show sparkline

### LINKED ISSUE #21760: Stat panel: No sparkline shown in small graphs
**What happened**: I tried to convert some small graphs (2 cols wide, 3 rows high) from singlestat to stat panels, but I noticed that the sparkline is missing. Only when I expand the image to 3 cols wide the sparklines appear.

**What you expected to happen**: Stat graphs show their sparkline regardless of their size.

**How to reproduce it (as minimally and precisely as possible)**: Create a 2x3 single stat panel, notice the sparkline. Convert to stat, sparkline is missing.

Grafana 6.5 with singlestat:
screenshot1

Grafana 6.6 with one panel converted to stat:
screenshot2

Grafana 6.6 with one panel converted to stat & extended to 3 rows (also the value font is quite small compared to singlestat):
screenshot

**Environment**:
- Grafana version: 6.6.0
- Data source type & version: Prometheus 2.15.2
- OS Grafana is installed on: Official docker container
- User OS & Browser: Fedora 31 & Chrome 79.0.3945.130

dprokop:
Closes https://github.com/grafana/grafana/issues/21760

Adds new Stat panel display option to config whether or not the sparkline should be automatically hidden:
image

When set to **adaptive**(default) it will behave as currently, when set to **always** it will always show the graph regardless of the panel size.

### DIFF
--- a/packages/grafana-ui/src/components/BigValue/BigValue.tsx
+++ b/packages/grafana-ui/src/components/BigValue/BigValue.tsx
@@ -26,6 +26,10 @@ export enum BigValueGraphMode {
   Line = 'line',
   Area = 'area',
 }
+export enum BigValueShowGraphMode {
+  Adaptive = 'adaptive',
+  Always = 'always',
+}
 
 export enum BigValueJustifyMode {
   Auto = 'auto',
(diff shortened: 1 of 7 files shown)


torkelo:
I do not think we have any evidence we need this new option, the stat panel already has so many options, and no upvotes for this: https://github.com/grafana/grafana/issues/21760  

Think if we where to add this I would try to make it part of the graph mode option and not a separate new option

torkelo:
It shows for pretty small panels, maybe we can tweak this logic to show it for even smaller ones
Screenshot from 2020-11-04 10-51-24

sakarikl:
I would gladly have this option. I don't personally understand why this decision must be done in code. This just seems to me that the user knows best what works in his/her use case.

PR #28809, opened 2020-11-04
MSSQL: Support request cancellation properly (Uses new backendSrv.fetch Observable request API)

hugohaggmark:
This PR replaces the usage of datasourceRequest in the Mssql data source with fetch and updates tests accordingly.

Relates #27222

### DIFF
--- a/public/app/plugins/datasource/mssql/specs/datasource.test.ts
+++ b/public/app/plugins/datasource/mssql/specs/datasource.test.ts
@@ -203,7 +203,6 @@ describe('MSSQLDatasource', () => {
   });
 
   describe('When performing metricFindQuery', () => {
-    let results: any;
     const query = 'select * from atable';
     const response = {
       results: {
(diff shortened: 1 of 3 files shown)


public/app/plugins/datasource/mssql/specs/datasource.test.ts
```diff
@@ -335,3 +334,17 @@ describe('MSSQLDatasource', () => {
     });
   });
 });
+
+function createFetchResponse<T>(data: T): FetchResponse<T> {
```
dprokop, line 338: I see we use this function in Postgres datasource too, maybe you can extract it so a helper util?

dprokop (approved):
Looks good, left one nit to consider 👍

public/app/plugins/datasource/mssql/specs/datasource.test.ts
hugohaggmark: Yepp, I think I did that in another PR already. I was thinking about doing that and changing any observableTester usages to the new jest matcher at the same PR.

PR #28810, opened 2020-11-04
grafana/toolkit: Allow builds with lint warnings

### LINKED ISSUE #23232: grafana/toolkit: eslint warnings prevent a plugin build
Noticed this today on canary, upgrading from 6.6.1

I have generated code in my plugin in src/codegen which produces lots of lint errors- preventing the build.

I added a .eslintignore and listed that dir.

`yarn build` now reports them all as warnings
```
/Users/slafferty/sevone-grafana-datasource/src/api/src/models/webhook-webhook-entry.ts
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override
```

However I can't do a prod build because the toolkit build fails on warns 

https://github.com/grafana/grafana/blob/ec743cf9a729236f01a16f81183a29b462a272c5/packages/grafana-toolkit/src/cli/tasks/plugin.build.ts#L102-L108

I thought I might be able to use the `--quiet` flag from eslint that swallows warns, however its only accessible via cli flags.

dprokop:
Closes https://github.com/grafana/grafana/issues/23232

Lint errors will still prevent the plugin from being built.

### DIFF
--- a/packages/grafana-toolkit/src/cli/tasks/plugin.build.ts
+++ b/packages/grafana-toolkit/src/cli/tasks/plugin.build.ts
@@ -103,13 +103,19 @@ export const lintPlugin = ({ fix }: Fixable = {}) =>
     }
 
     const { errorCount, results, warningCount } = report;
+    const formatter = cli.getFormatter();
 
-    if (errorCount > 0 || warningCount > 0) {
-      const formatter = cli.getFormatter();
+    if (warningCount > 0) {
       console.log('\n');
       console.log(formatter(results));
       console.log('\n');
-      throw new Error(`${errorCount + warningCount} linting errors found in ${results.length} files`);
+    }
+
+    if (errorCount > 0) {
+      console.log('\n');
+      console.log(formatter(results));
+      console.log('\n');
+      throw new Error(`${errorCount} linting errors found in ${results.length} files`);
     }
   });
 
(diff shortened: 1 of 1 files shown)


packages/grafana-toolkit/src/cli/tasks/plugin.build.ts
```diff
@@ -103,13 +103,19 @@ export const lintPlugin = ({ fix }: Fixable = {}) =>
...
       console.log('\n');
       console.log(formatter(results));
       console.log('\n');
-      throw new Error(`${errorCount + warningCount} linting errors found in ${results.length} files`);
+    }
+
+    if (errorCount > 0) {
+      console.log('\n');
+      console.log(formatter(results));
+      console.log('\n');
```
stevenvachon, line 117: Could've put this in an `errorCount > 0 && warningCount > 0` to avoid redundancy

stevenvachon (approved):

packages/grafana-toolkit/src/cli/tasks/plugin.build.ts
dprokop: Yeah, we could, and then throw only when there are errors. Good catch

dprokop: See 247160e

stevenvachon (approved):

PR #28818, opened 2020-11-04
CodeEditor: Added support for javascript language.

ae3e:
Plugins that use the code editor component will be able to display Javascript code with syntax highlighting


dprokop (approved):

mckn (approved):
Works as expected! Thanks for contributing.

PR #28820, opened 2020-11-04
VizLayout: Simple viz layout component for legend placement and scaling

torkelo:
Need a simple layout component for new PieChart and found https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph3/VizLayout.tsx  a bit more complex than what we need and the API a bit tricky. 

This is a simple POC to show an alternative approach, not sure this is better? 

This creates a VizLayout component that has one legend that it renders and measures in order then to render the main viz (that get's remaining size). The layout has options to control the max width or height of the legend slot in relation to the container.  

Not sure how flexible we need this (multiple slots, like legend both to the right & bottom), or we will do with only two and if we need more maybe we can nest this.


packages/grafana-ui/src/components/VizLayout/VizLayout.story.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {items.map((_, index) => (
+        <div style={{ height: '30px', width: '100%', background: 'blue', marginBottom: '2px' }}>
+          Legend item {index}
+        </div>
+      ))}
+    </VizLayout.Slot>
+  );
+
+  return (
+    <VizLayout width={400} height={500} legend={legend}>
```
dprokop, line 37: I think the "slots" (legend in this case) might be independent from the props and rendered declaratively instead? Don't think VizLayout component API should know any specifics about what's there to be rendered

packages/grafana-ui/src/components/VizLayout/VizLayout.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  return (
+    <div style={containerStyle}>
+      <div style={vizStyle}>{size && children(size.width, size.height)}</div>
+      {legend && (
+        <div style={legendStyle} ref={legendRef}>
+          <CustomScrollbar>{legend}</CustomScrollbar>
+        </div>
+      )}
+    </div>
```
dprokop, line 71: Think this is fine as long as we assume visualization can utilize only a single additional slot (i.e. legend only).

dprokop: With the approach taken in https://github.com/grafana/grafana/blob/master/public/app/plugins/panel/graph3/VizLayout.tsx the idea was that the visualization can have multiple additional eleents in different slots.

packages/grafana-ui/src/components/VizLayout/VizLayout.story.tsx
torkelo: Not sure what you mean independent? Slots/Legend needs to be part of the API somehow either via props or via children

Having it via children makes the API A LOT more complicated, needed multiple different child components and possibly react context to access the viz size. I started with that but using props for the slots and the main child render prop for the viz made it 10x simpler and more elegant.  And it's still declarative like this, just passed via prop instead of via child 

I can change the name of the prop from legend to spots?

packages/grafana-ui/src/components/VizLayout/VizLayout.tsx
torkelo: yea, that was what I can see we need now in new NG Chart & Pie chart. But  would be nice to have the API extend to more slots in the future without API changes

packages/grafana-ui/src/components/VizLayout/VizLayout.story.tsx
torkelo: >  Don't think VizLayout component API should know any specifics about what's there to be rendered

It does need to know the constraints (max-width / height) & position of the slots and that is all that is done here via the VizLayout.Slot component

packages/grafana-ui/src/components/VizLayout/VizLayout.tsx
dprokop: Ah, and one more thing I forgot with my approach. It allows you to render Graph plugins as children of a graph component, as the actual layou is rendered as such:
```
 <GraphNG data={data.series} timeRange={timeRange} timeZone={timeZone} {...canvasSize}>
            {builder.addSlot('canvas', <Canvas />).render()}
```

But I guess we can **maybe** get rid of this as the Legend does not have to be rendered as uPlot plugin as it unly uses `usePlotData` hook which is not necessary there. It can just accept the data frame pushed to the GraphPanel instead.

packages/grafana-ui/src/components/VizLayout/VizLayout.story.tsx
torkelo: renamed VizLayout.Slot to VizLayout.Legend as it feels more accurate

packages/grafana-ui/src/components/VizLayout/VizLayout.tsx
torkelo: Ok, I understand. 

But yea, if avoiding complexity like that sounds like a good thing.

I really do not understand how that even works though with uPlot and new graph panel, the legend has to render first in order to know the canvas size. But it if the legend is a uplot plugin that renders as part of the uplot then uplot needs to render first with some initial size and then again after the legend size has been determined?

torkelo:
I don't think to need multiple spots at the same time, if we do we can just nest a VizLayout inside another VizLayout. So think this API should be pretty future proof

packages/grafana-ui/src/components/VizLayout/VizLayout.tsx
leeoniya: it probably makes sense to have a distinction between a uPlot plugin and GraphNG/UPlotChart plugin.

uPlot needs an explicit width/height in pixels, so anything you need to fit next to it must be known in advance.

since you guys render the legend (and probably title) externally, `Legend` would be a `GraphNG` plugin that ends up rendering first and then passing the remainder of the container size to init uPlot afterwards.

i'm assuming there are no plans to actually allow authoring and passing raw uPlot plugins through this component stack, but just create purpose-build GraphNG plugins with well-defined APIs?

torkelo: > since you guys render the legend (and probably title) externally, Legend would be a GraphNG plugin that ends up rendering first and then passing the remainder of the container size to uPlot afterwards.

Yea exactly, Grafana needs to render the legend first to get the available size remaining for the chart/viz

leeoniya: probably all slots that must be implicitly sized. this might present some challenges if you plan to use nested `VizLayout`s that contain other graphs, but that seems pretty edge case.

in this case grid-template-areas may come to the rescue in terms of layout not governing component order, since you can render components in implict -> explicit order and accumulate their dimensions before rendering the center, though i'm not sure if they will be measurable until the next event loop tick :/

1. top slot
2. right slot
3. bottom slot
4. left slot
5. center slot (explicit, pass accumulated dims)

PR #28822, opened 2020-11-04
Git: Fix line endings on Windows

aknuds1:
Git is currently configured via .gitattributes such that files are checked out with CRLF line endings on Windows, even if I've disabled autocrlf in Git itself. I managed to solve it by adding `eol=lf` to .gitattributes.

Before the fix, I can't run `yarn start`, since `prettier` reacts to the CRLF line endings.

Would be good if others test on Windows, and give feedback on how it works for them :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text=auto
+* text=auto eol=lf


sakjur (approved):
LGTM

FWIW, this might cause problems for users running an old version of Git on Windows:
https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248

Weighing the two options, I'd rather have this in and simply ask users to upgrade if that's them, but it's something we should be aware of.

AgnesToulet (approved):
LGTM! As mentioned on Slack, it doesn't impact files that were checked out before this fix on Windows (CRLF files will stay CRLF if there is no other change to check out) but it works on clean clone and updated files.

PR #28825, opened 2020-11-04
Units: Changes FLOP/s to FLOPS and some other rates per second units get /s suffix 

Berbe:
Homogoneise units suffixes:
- Use `/` instead of `p` for time rate units suffixes (along with commonly seen units abbreviation, when that was the case)
- Use the generally used `FLOPS` unit instead of `FLOP/s` (as much as I personally prefer the latter, I never saw it being used...)

No extra test introduced in `public/app/plugins/panel/table-old/specs/renderer.test.ts`. Shall other units be tested there?


Berbe:
This PR is threatened with staleness.

@torkelo, @kaydelaney?

torkelo:
Thanks for contributing!

torkelo (approved):

tofurky:
'cpm' appears to be the accepted unit name for radioactivity i.e. https://en.wikipedia.org/wiki/Counts_per_minute

should at least the 'c/m' be changed back to 'cpm' or is this meant to be more generic?

Berbe:
> 'cpm' appears to be the accepted unit name for radioactivity i.e. https://en.wikipedia.org/wiki/Counts_per_minute
That use is accepted, but shall be the norm?
It looks to me `cpm` comes from US use.

> should at least the 'c/m' be changed back to 'cpm' or is this meant to be more generic?
This PR attempts at homogeneising the way units are being displayed to enhance coherence. Cue commit messages.

tofurky:
i understand wanting to unify things, but some of these appear to be consistently written that way (with the 'p') regardless of whether SI is in use. it feels awkward, like writing 'ppb' or 'ppm' as 'p/b' and 'p/m'. but, maybe that's just my skewed view coming from a (the) non-SI country :)

anyways, i've found that there is a way to override this on a per-graph basis, which will handle my graph affected by this PR:
image

PR #28826, opened 2020-11-04
Dashboards: Delete related data (permissions, stars, tags, versions, annotations) when deleting a dashboard or a folder

AgnesToulet:
This PR does the following things:
- Delete dashboard permissions related to a dashboard when it's deleted
- Delete all the following information related to a folder's dashboards when this folder is deleted: stars, tags, versions, annotations, provisioning and permissions. This is already handled when a dashboard is deleted, so it should be handled when a folder causes the deletion of all its children.
- Add migrations to remove dashboard permissions, tags and stars related to deleted dashboards and folders (to clean up existing Grafana instance).

### DIFF
--- a/pkg/services/sqlstore/dashboard.go
+++ b/pkg/services/sqlstore/dashboard.go
@@ -374,23 +374,47 @@ func deleteDashboard(cmd *models.DeleteDashboardCommand, sess *DBSession) error 
 		"DELETE FROM dashboard_version WHERE dashboard_id = ?",
 		"DELETE FROM annotation WHERE dashboard_id = ?",
 		"DELETE FROM dashboard_provisioning WHERE dashboard_id = ?",
+		"DELETE FROM dashboard_acl WHERE dashboard_id = ?",
 	}
 
 	if dashboard.IsFolder {
-		deletes = append(deletes, "DELETE FROM dashboard_provisioning WHERE dashboard_id in (select id from dashboard where folder_id = ?)")
 		deletes = append(deletes, "DELETE FROM dashboard WHERE folder_id = ?")
 
 		dashIds := []struct {
 			Id int64
 		}{}
-		err := sess.SQL("select id from dashboard where folder_id = ?", dashboard.Id).Find(&dashIds)
+		err := sess.SQL("SELECT id FROM dashboard WHERE folder_id = ?", dashboard.Id).Find(&dashIds)
 		if err != nil {
 			return err
 		}
 
-		for _, id := range dashIds {
+		dashIdsArray := make([]interface{}, len(dashIds))
+		queryEnd := "("
+		for i, id := range dashIds {
 			if err := deleteAlertDefinition(id.Id, sess); err != nil {
 				return err
+			}
+			dashIdsArray[i] = id.Id
+			queryEnd += "?,"
+		}
+		queryEnd = queryEnd[:len(queryEnd)-1] + ")"
+
+		if len(dashIds) > 0 {
+			childrenDeletes := []string{
+				"DELETE FROM dashboard_tag WHERE dashboard_id IN ",
+				"DELETE FROM star WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_version WHERE dashboard_id IN ",
+				"DELETE FROM annotation WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_provisioning WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_acl WHERE dashboard_id IN ",
+			}
+			for _, sql := range childrenDeletes {
+				sqlOrArgs := append([]interface{}{sql + queryEnd}, dashIdsArray...)
+				_, err := sess.Exec(sqlOrArgs...)
+
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


pkg/services/sqlstore/dashboard.go
```diff
@@ -374,24 +374,48 @@ func deleteDashboard(cmd *models.DeleteDashboardCommand, sess *DBSession) error
...
+			childrenDeletes := []string{
+				"DELETE FROM dashboard_tag WHERE dashboard_id IN ",
+				"DELETE FROM star WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_version WHERE dashboard_id IN ",
+				"DELETE FROM annotation WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_provisioning WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_acl WHERE dashboard_id IN ",
+			}
+			for _, sql := range childrenDeletes {
+				sqlOrArgs := append([]interface{}{sql + queryEnd}, dashIdsArray...)
```
marefr, line 412: I would suggest to do `delete from <table> where  dashboard_id in(select id from dashboard where folder_id=?)` instead of concatenating strings.

marefr (changes requested):
Nicely spotted :+1: Weird we haven't noticed the lack of this before. Looks good I think. See comment

pkg/services/sqlstore/dashboard.go
AgnesToulet: It's more readable for sure but I was trying to avoid doing the subquery `select id from dashboard where folder_id=?` too many times as the `dashboard` table can be quite big and the `folder_id` column is not indexed. What do you think?

papagian: I was about to suggest using `strings.Join()` for building the query placeholders but after reading @marefr's suggestion I'm in favour of it. You are right that there is no index for folder_id but there is a composite one for org_id, folder_id and title so you could use this one alternatively (and fixing the rest queries using the folder_id). WDYT?

pkg/services/sqlstore/dashboard.go
```diff
@@ -374,24 +374,48 @@ func deleteDashboard(cmd *models.DeleteDashboardCommand, sess *DBSession) error
...
+		dashIdsArray := make([]interface{}, len(dashIds))
+		queryEnd := "("
+		for i, id := range dashIds {
 			if err := deleteAlertDefinition(id.Id, sess); err != nil {
 				return err
 			}
+			dashIdsArray[i] = id.Id
+			queryEnd += "?,"
+		}
+		queryEnd = queryEnd[:len(queryEnd)-1] + ")"
```
xlson, line 400: Could use `strings.Join(...)` here to simplify the code a little.

pkg/services/sqlstore/dashboard.go
xlson: Makes sense to me.

xlson (approved):
Great work!

pkg/services/sqlstore/dashboard.go
```diff
@@ -374,24 +374,48 @@ func deleteDashboard(cmd *models.DeleteDashboardCommand, sess *DBSession) error
...
+				"DELETE FROM dashboard_provisioning WHERE dashboard_id IN ",
+				"DELETE FROM dashboard_acl WHERE dashboard_id IN ",
+			}
+			for _, sql := range childrenDeletes {
+				sqlOrArgs := append([]interface{}{sql + queryEnd}, dashIdsArray...)
+				_, err := sess.Exec(sqlOrArgs...)
+
+				if err != nil {
+					return err
+				}
```
aknuds1, line 417: I suggest removing the blank line between error definition and check, since it makes it harder to see the connection when reading the code:
```suggestion
				_, err := sess.Exec(sqlOrArgs...)
				if err != nil {
					return err
				}
```

aknuds1 (approved):
Just left a nit suggestion.

pkg/services/sqlstore/dashboard.go
AgnesToulet: Did the update to use the composite index, thanks for the tip!

papagian (approved):
Nice! I've tested with MySQL and PostgreSQL and I've verified that it's using the index now.

I leave here for reference a list with the queries that we possibly have to improve to use the index as well:
https://github.com/grafana/grafana/blob/7897c6b7d5a81702eb130064e9e1749010d97d71/pkg/services/sqlstore/dashboard.go#L380-L382

https://github.com/grafana/grafana/blob/7897c6b7d5a81702eb130064e9e1749010d97d71/pkg/services/sqlstore/dashboard.go#L386

https://github.com/grafana/grafana/blob/7897c6b7d5a81702eb130064e9e1749010d97d71/pkg/services/sqlstore/dashboard.go#L445-L455

https://github.com/grafana/grafana/blob/7897c6b7d5a81702eb130064e9e1749010d97d71/pkg/services/sqlstore/dashboard_acl.go#L78-L115

https://github.com/grafana/grafana/blob/7897c6b7d5a81702eb130064e9e1749010d97d71/pkg/services/sqlstore/sqlbuilder.go#L44-L84

marefr (approved):
LGTM

PR #28831, opened 2020-11-04
Snapshots: Increase MySQL Blob column size for encrypted dashboard data

### LINKED ISSUE #28815: Dashboard snapshot cannot be saved since it's too long
A user in our hosted service is having problems storing a dashboard snapshot due to the size of the dashboard. 

```
error: Error 1406: Data too long for column 'dashboard_encrypted' at row 1
msg :  "Failed to create snapshot"
```

Seems like we using `DB_BLOB` instead of `DB_TEXT` as a datatype for encrypted dashboard snapshots. The Blob datatype is limited to 65535 bytes (64 KB). A `MEDIUMBLOB` is 16777215 bytes (16 MB) and a `LONGBLOB` is 4294967295 bytes (4 GB). So I suggest we change it to `MEDIUMBLOG`

wbrowne:

Fixes #28815


aknuds1 (approved):
Assume you've tested this.

wbrowne:
@aknuds1 Yep have tested!

marefr (approved):
LGTM great

PR #28841, opened 2020-11-04
VizLegend: Simple legend proposal (RFC)

torkelo:
The new pie chart needs a legend and I am thinking of using something a bit simpler than the Graph legend. 

Something that can only show one value (or no value). So no table & no multiple stats. 

Also propose all legends switch from the small thin colored line to a bigger square? 
Screenshot from 2020-11-04 20-59-53

Todo:
* [ ] Support showing value 
* [ ] Use DisplayValue type or not on for the ItemProps 

Beyond MVP
* [ ] Data links
* [ ] Hover state/callback (to be able to highlight the corresponding series in the viz as you hover over a legend item) 

But feel unsure, maybe using/tweaking the Graph Legend is better https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Legend/Legend.tsx  ?


packages/grafana-ui/src/components/VizLegend/VizLegend.tsx
```diff
@@ -0,0 +1,51 @@
...
+    styles.flexDirection = 'column';
+  }
+
+  return <div style={styles}>{children}</div>;
+};
+
+export interface VizLegendItemProps {
+  color: string;
+  name: string;
+  value?: string;
```
torkelo, line 24: could just pass in a DisplayValue here, but was thinking of making this component display only and not care much about data types

dprokop:
+1 for the updated UI, think it's way more accessible than the one we use in Graph. Would love to explore the usage of squares instead of lines in the Graph context too.

PR #28842, opened 2020-11-04
Prometheus: fix missing labels from value

### LINKED ISSUE #28833: Prometheus: No labels associated with data returned by data source
**What happened**:
upgraded grafana from 7.2 to 7.3.1 and our dashboards that utilizies tables and the labels to fields transformation stopped being displayed properly and it can't be recreated due to fields not showing up under transformation.

A prometheus and a node_exporter running on the same host should be enough to showcase the problem.
This pormetheus instance needs to be added as a datasource into grafana once it starts polling. Prometheus and node_exporter can be downloaded from https://prometheus.io/download/.
Prometheus config:
```
global:
  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
scrape_configs:
  - job_name: 'prometheus'
    static_configs:
      - targets: [ 'localhost:9100']
```
The query in question is just an example, we actually used this visualization method to create a reachabilitiy/latency matrix, this is just to demonstrate as minmally as possible
```
sum(rate(node_cpu_seconds_total[5m])) by (cpu, mode)
```
Attached are the GIFs about how to reproduce the problem, working as expected under 7.2.1:
labels-to-field-7 2 1
but doesn't show the required fields in 7.3.1
labels-to-field-7 3 1

**Environment**:
- Grafana version: 7.3.1
- Data source type & version: Prometheus 2.22.0
- OS Grafana is installed on: ubuntu/grafana docker image
- User OS & Browser: Windows 10, Firefox and tried Chrome browser as well.
- Grafana plugins: none
- Others:

zoltanbedi:
Fixes #28833

### DIFF
--- a/public/app/plugins/datasource/prometheus/result_transformer.test.ts
+++ b/public/app/plugins/datasource/prometheus/result_transformer.test.ts
@@ -248,6 +248,33 @@ describe('Prometheus Result Transformer', () => {
   });
 
   describe('When the response is a matrix', () => {
+    it('should have labels with the value field', () => {
+      const response = {
+        status: 'success',
+        data: {
+          resultType: 'matrix',
+          result: [
+            {
+              metric: { __name__: 'test', job: 'testjob', instance: 'testinstance' },
+              values: [
+                [0, '10'],
+                [1, '10'],
+                [2, '0'],
+              ],
+            },
+          ],
+        },
+      };
+
+      const result: DataFrame[] = transform({ data: response } as any, {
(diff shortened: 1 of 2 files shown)


ivanahuckova (approved):
LGTM! Works great. Also awesome that you have added test and also typings. Good job! 🤘

aocenas (approved):

PR #28843, opened 2020-11-04
Live: update centrifuge and the ChannelHandler api

ryantxu:
Centrifuge has updated their API so that Channel Options are configured when they are used, not in an independent function.  This requires a few changes to our ChannelHandler API, but I think makes the behavior more clear.

See https://github.com/centrifugal/centrifuge/blob/master/changelog.md#v0130


aknuds1 (approved):
LGTM

PR #28852, opened 2020-11-05
Chore: Remove unused Go code

aknuds1:
Remove unused Go code. I've tested against enterprise to verify nothing breaks.

I can split into more PRs if desirable.


pkg/components/simplejson/simplejson.go
```diff
@@ -445,26 +432,3 @@ func (j *Json) MustInt64(args ...int64) int64 {
...
-		log.Panicf("MustUint64() received too many arguments %d", len(args))
-	}
-
-	i, err := j.Uint64()
-	if err == nil {
-		return i
-	}
-
-	return def
-}
```
torkelo, line 470: this is more like an embedded library, just because we do not use these functions yet seems unnecessary to remove them

aknuds1: @torkelo OK, so to clarify, I should undo all simplejson changes in this PR? And flag all of the simplejson code as in-use (so the linter doesn't flag them again)?

torkelo: yea, think so. Or vendor in https://github.com/bitly/go-simplejson  but not sure if we have made changes and these have diverged, can't remember why I just copied that lib and did not vendor it

aknuds1: OK, thanks!

marefr:
Looks good. Have you tested with Enterprise?

aknuds1:
@marefr I think I did, but let me double check.

marefr (approved):
LGTM when conflicts been resolved

PR #28853, opened 2020-11-05
Variables: Added __user.email to global variable

### LINKED ISSUE #28840: P/R #25275:  user.email was removed from global variables.
user.email was removed from global variables. 

The position taken was "there is no use case" and you wanted to limit the data exposed. 

I submit that the data is exposed in the client regardless (when logged on) and as such this only serves to limit the functionality.

The use case I have is quite simple (and clearly others do as well).   I pull data from multiple systems, many of which do not allow for federated authentication (Some use shortened logons, others require globally unique usernames across a large shared platform and don't permit email addresses as a username) 

The only useful "Key" that allows a user to be uniquely identified across the dozen or so systems is their email address. 
We are using Grafana to create user specific dashboards (ie user logs on - and they see data specific to them).   

The use of "Login Name" is not an acceptable substitute because it cannot be guaranteed to be the same in all system that we are integrating, furthermore, because we are a group of companies, the domain portion for a user may change depending on what their org unit is.   As such we cannot simply use "loginid" and append the domain portion.

Happy to re-resubmit a PR to put this back

A

mckn:
We excluded email from the global template variables to limit the amount of exposed data and most of the scenarios the `__user.id` is the same as the `__user.email`. But since explained in #28840 that might not be the case in all scenarios so I added the email as an alternative to `__user.id`. 

Fixes #28840

Should this be in `7.3.2` or `7.4.0`? I guess it is a new addition to the template variable api so maybe `7.4.0`?


hugohaggmark (approved):
Looks outstanding ⭐  
(nit: e2e tests are failing though)

jackw (approved):
🚀

docs/sources/variables/variable-types/global-variables.md
```diff
@@ -70,6 +70,7 @@ This variable is the ID of the current organization.
 
 `${__user.id}` is the ID of the current user.
 `${__user.login}` is the login handle of the current user.
+`${__user.email}` is the email for the current user.
```
achatterjee-grafana, line 73:
```suggestion
`${__user.email}` is the email of the current user.
```

mckn:
@hugohaggmark what do you think about the milestone? 7.4.0 or 7.3.2?

achatterjee-grafana (approved):
The code needs a review from dev. Added a copy-edit suggestion.

hugohaggmark:
> @hugohaggmark what do you think about the milestone? 7.4.0 or 7.3.2?

I think 7.4.0 would do?

oddlittlebird:
> > @hugohaggmark what do you think about the milestone? 7.4.0 or 7.3.2?
> 
> I think 7.4.0 would do?

Yes, please hold major features like these for the minor releases. It makes managing the docs much easier for @grafana/docs-squad .

PR #28857, opened 2020-11-05
Dashboards: Adds cheat sheet toggle to supported query editors

kaydelaney:
Adds support for viewing data source cheat sheets from the panel editor view.
Nov-05-2020 10-35-03


mckn (approved):
LGTM!

Added some small nits and questions that might be worth considering.

public/app/features/dashboard/panel_editor/HelpToggle.tsx
```diff
@@ -0,0 +1,21 @@
+import { Icon } from '@grafana/ui';
+import { css, cx } from 'emotion';
+import React, { useState } from 'react';
+
+const helpStyle = css`
+  margin-top: 4px;
+`;
+
+export const HelpToggle: React.FunctionComponent = ({ children }) => {
```
mckn, line 9: nit: feels like this pattern exists on more places? Is this something that we could break out into a grafana/ui component?

public/app/features/dashboard/panel_editor/HelpToggle.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const [isHelpVisible, setIsHelpVisible] = useState(false);
+
+  return (
+    <>
+      <label className="gf-form-label query-keyword pointer" onClick={_ => setIsHelpVisible(!isHelpVisible)}>
```
mckn, line 14: nit: would it be possible to skip the dependency on the legacy classNames and use emotion instead? Maybe replace this label with components from grafana/ui that embeds the styling?

kaydelaney: Good idea, will do

kaydelaney: Was thinking about this and I think `<button>` might be the more semantically appropriate tag, though I'd like to leave the styling as-is so it fits in well with query editor styling. Might have to factor out some of the `<Label>` styles.

aocenas:
@kaydelaney sorry for being late with this, but is there a separate PR for Explore with this? Or do you plan to add the PR later?

kaydelaney:
@aocenas Oh, I suppose that would make sense. I can add it real quick

aocenas:
We can also add it in separate PR no problem, what ever makes more sense to you.

aocenas:
Right now in explore we cannot do mixed DS queries so each query will be the same DS. With that having a button per query does not make much sense:
Screenshot from 2020-11-13 13-03-14

Maybe we can just put it in the bottom button rows and have single help button there.

kaydelaney:
@aocenas You're right, will change

ivanahuckova (approved):
> Maybe we can just put it in the bottom button rows and have single help button there.

Yeah, I agree with this as well, but I see that you already mentioned that you are going to change the Help button's position. So other than that LGTM! 🚀

public/app/features/dashboard/panel_editor/HelpToggle.tsx
```diff
@@ -0,0 +1,21 @@
+import { Icon } from '@grafana/ui';
+import { css, cx } from 'emotion';
+import React, { useState } from 'react';
+
+const helpStyle = css`
+  margin-top: 4px;
```
dprokop, line 6: Maybe theme.spacing.xs instead? :)

public/app/features/dashboard/panel_editor/HelpToggle.tsx
dprokop: Hm, I cannot recall any other places we use this pattern tbh. We use something similar which is the InfoBox, but it serves slighlty different purpose I think.

public/app/features/dashboard/panel_editor/HelpToggle.tsx
```diff
@@ -0,0 +1,21 @@
...
+`;
+
+export const HelpToggle: React.FunctionComponent = ({ children }) => {
+  const [isHelpVisible, setIsHelpVisible] = useState(false);
+
+  return (
+    <>
+      <button className="gf-form-label query-keyword pointer" onClick={_ => setIsHelpVisible(!isHelpVisible)}>
```
dprokop, line 14: Could we rewrite the styles to emotion maybe instead of using the SASS class names?

dprokop: Or maybe even another way. Waht about using Button[variant=link] and then when the help is visible using the [InfoBox component](https://developers.grafana.com/ui/latest/index.html?path=/story/layout-infobox--basic) an utilizing its `onDismiss` property?

dprokop (commented):
Looks good @kaydelaney , left few thoughts to consider before merging this work :)

public/app/features/dashboard/panel_editor/QueryEditorRow.tsx
```diff
@@ -284,6 +285,7 @@ export class QueryEditorRow extends PureComponent<Props, State> {
     }
 
     const editor = this.renderPluginEditor();
+    const DatasourceCheatsheet = datasource.components?.ExploreStartPage;
```
torkelo, line 288: This feels awkward, should we rename this to QueryEditorHelp? feels like the way this is implemented can't it just be part of the QueryEditor?  Feels like Grafana does not do much beyond rendering the component below the editor, something the QueryEditor component itself could easily do. Guess the expand / collapse state is nice to reuse. And making this a component like this makes it possible for Grafana to change the design for how to access it. Was thinking to make we make it an icon in the query row bar instead (next to duplicate and the eye icons).

aocenas: No feelings about the naming, but for making this into a datasource API the main thing was that before it was shown outside of the editorRow (in Explore) and I think it may be the case in the future (like showing in modal, slide panel or what ever). There are other ways to solve that (provide something like `openModal` prop for the editor) but in general this makes it easier to provide consistent look and feel for the editor/query help which I think is a good think in contrast of having it implemented/styled differently across editors (especially if talking about 3rd party data sources). At the same time this does not prevent still implementing it in the queryEditor if somebody wants to just gives an easy to opt in way to do it.

torkelo: yea, agree fully. Having this be a standard component that the plugin exposes makes it possible for Grafana to refine and change how this is presented and make it consistent. 

The awkward part was "ExploreStartPage" name for the component, proposing a more generic name like QueryEditorHelp

torkelo:
> Right now in explore we cannot do mixed DS queries so each query will be the same DS. With that having a button per query does not make much sense:

But the cheat has examples, that could effect that query, or could be adapted to what the query currently has (in theory).

The dashboard query editor row has a header with eye, duplicate, remove icons (think we should move re-using this in explore as well (long term) to create more UI consistency, and reduce duplication). A help icon there that triggers this view above or below the editor? it would then not take up so much space for each query .

aocenas:
> But the cheat has examples, that could effect that query, or could be adapted to what the query currently has (in theory).
> The dashboard query editor row has a header with eye, duplicate, remove icons (think we should move re-using this in explore as well (long term) to create more UI consistency, and reduce duplication). A help icon there that triggers this view above or below the editor? it would then not take up so much space for each query.

I agree and considering we want to support multiple data sources in explore for v8 we may redesign the queryRow in explore (and aligning with dashboard makes sense imho). Not sure though if it makes sense to wait until then I think we can do this as a first step.

torkelo:
I assumed this was only for the dashboard, where we already have the header row for query actions / toggles

public/app/features/dashboard/panel_editor/HelpToggle.tsx
kaydelaney: Dec-02-2020 17-34-34

Is this what you mean?

aocenas: Personally I like it better when these kind of toggles are in the same place, so I can open close without moving a mouse or searching for the close button that much.

kaydelaney: I agree, just trying to get an understanding of what @dprokop had in mind

torkelo:
here is where the query row actions are (icon buttons): https://github.com/grafana/grafana/blob/master/public/app/features/query/components/QueryEditorRow.tsx#L250 

would use a new help icon button there to toggle this view

public/app/features/dashboard/panel_editor/HelpToggle.tsx
dprokop: Yups, when I see this now @kaydelaney @aocenas I agree we both of you.

dprokop:
> there is where the query row actions are (icon buttons): https://github.com/grafana/grafana/blob/master/public/app/features/query/components/QueryEditorRow.tsx#L250
would use a new help icon button there to toggle this view

I think this is an excellent idea

torkelo (commented):
I think this looks good now, :smile:  but I think the view should close when you click on an example query because as it is now you do not see anytihing happening without scrolling all the way to the bottom

torkelo (approved):

marcusolsson:
Could you also update the docs for [Adding a start page for Explore](https://github.com/grafana/grafana/blob/master/docs/sources/developers/plugins/add-support-for-explore-queries.md#add-a-start-page-for-explore)?

packages/grafana-data/src/types/datasource.ts
```diff
@@ -75,8 +75,8 @@ export class DataSourcePlugin<
     return this;
   }
 
-  setExploreStartPage(ExploreStartPage: ComponentType<ExploreStartPageProps>) {
-    this.components.ExploreStartPage = ExploreStartPage;
+  setQueryEditorHelp(QueryEditorHelp: ComponentType<QueryEditorHelpProps>) {
```
torkelo, line 78: would be good if we could find a runtime way to support this function still to not break old plugins like x-ray: https://github.com/grafana/x-ray-datasource/blob/master/src/module.ts#L13 

We could either deprecate it, or just a runtime way to bind this.setExploreStartPage = this.setQueryEditorHelp

marcusolsson (changes requested):
This needs docs updates for [Adding a start page for Explore](https://github.com/grafana/grafana/blob/master/docs/sources/developers/plugins/add-support-for-explore-queries.md#add-a-start-page-for-explore)

docs/sources/developers/plugins/add-support-for-explore-queries.md
```diff
@@ -10,7 +10,7 @@ This guide assumes that you're already familiar with how to [Build a data source
 
 With Explore, users can make ad-hoc queries without the use of a dashboard. This is useful when users want to troubleshoot or to learn more about the data.
 
-Your data source already supports Explore by default, and will use the existing query editor for the data source. If you want to offer extended Explore functionality for your data source however, you can define a Explore-specific query editor. Optionally, your plugin can also define a _start page_ for Explore.
+Your data source already supports Explore by default, and will use the existing query editor for the data source. If you want to offer extended Explore functionality for your data source however, you can define a Explore-specific query editor.
```
oddlittlebird, line 13: @marcusolsson , this looks like there should be a link to the previous topic here. Can you confirm?

docs/sources/developers/plugins/add-query-editor-help.md
```diff
@@ -0,0 +1,72 @@
...
+   import { QueryEditorHelpProps } from '@grafana/data';
+
+   export default (props: QueryEditorHelpProps) => {
+     return (
+       <h2>My cheat sheet</h2>
+     );
+   };
+   ```
+
+2. Configure the plugin to use the `QueryEditorHelp`.
```
oddlittlebird, line 18: If possible, use repetitive numbering per the [Grafana markdown guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-markdown-guide.md#numbered-lists).

docs/sources/panels/queries.md
```diff
@@ -118,6 +118,8 @@ You can:
 
 | Icon | Description |
 |:--:|:---|
+| {{< docs-imagebox img="/img/docs/queries/query-editor-help-7-4.png" class="docs-image--no-shadow" max-width="30px" max-height="30px" >}} | Toggle query editor help. If supported by the data source, this will toggle displaying information on how to use its query editor, or provide quick
+access to commonly used queries. |
```
oddlittlebird, line 122:
```suggestion
access to commonly-used queries. |
```

oddlittlebird (changes requested):

kaydelaney:
> No need to escape these.

Ah, I see. I was under the impression that was just the syntax

docs/sources/developers/plugins/add-support-for-explore-queries.md
marcusolsson: I don't think so?

marcusolsson (approved):

torkelo:
@kaydelaney to make 7.4 beta this needs to be merged now

oddlittlebird (approved):

PR #28860, opened 2020-11-05
Chore: Fix staticcheck issues

aknuds1:
Fix issues reported by [staticcheck](https://staticcheck.io) in the backend. Mostly error strings being capitalized.


pkg/api/render.go
```diff
@@ -24,25 +24,25 @@ func (hs *HTTPServer) RenderToPng(c *models.ReqContext) {
 
 	width, err := strconv.Atoi(queryReader.Get("width", "800"))
 	if err != nil {
-		c.Handle(400, "Render parameters error", fmt.Errorf("Cannot parse width as int: %s", err))
+		c.Handle(400, "Render parameters error", fmt.Errorf("cannot parse width as int: %s", err))
```
papagian, line 27: I guess the error is not required to be wrapped here because it's an api response?

pkg/infra/log/file.go
```diff
@@ -193,12 +193,12 @@ func (w *FileLogWriter) DoRotate() error {
 		// close fd before rename
 		// Rename the file to its newfound home
 		if err = os.Rename(w.Filename, fname); err != nil {
-			return fmt.Errorf("Rotate: %s", err)
+			return fmt.Errorf("rotate: %s", err)
```
papagian, line 196: Do we want the error be wrapped here?

pkg/infra/log/file.go
```diff
@@ -193,12 +193,12 @@ func (w *FileLogWriter) DoRotate() error {
...
 		}
 
 		// re-start logger
 		if err = w.StartLogger(); err != nil {
-			return fmt.Errorf("Rotate StartLogger: %s", err)
+			return fmt.Errorf("rotate StartLogger: %s", err)
```
papagian, line 201: Same as above, do we want the error be wrapped here?

papagian (approved):
LGTM; I left some comments about wrapping some errors.

pkg/api/render.go
aknuds1: @papagian Yeah, I can't imagine that any wrapped error will be checked by `c.Handle`.

pkg/infra/log/file.go
aknuds1: @papagian Not sure, I didn't really take it into account as part of this PR. I actually read a commentary in the meantime on how you don't _always_ want to wrap errors, in case you want to avoid leaking them to your callers. It would probably require some more thought to tell if we want to wrap here or not.

pkg/infra/log/file.go
aknuds1: @papagian See my reply above, I'm rather agnostic in the scope of this PR as to whether we want to do it or not.

pkg/infra/log/file.go
papagian: ok, let's leave it as it used to be for now.

PR #28865, opened 2020-11-05
CloudWatch: Support request cancellation properly

hugohaggmark:
This PR replaces the usage of datasourceRequest in the Google Cloud Monitoring data source with fetch and updates tests accordingly. I have tested most cases except for Annotations because I couldn't find a query that worked, so if someone can verify this it would be very appreciated.

Relates #27222

I'm just testing using the dashboards supplied by Grafana Labs for CloudWatch. For ds settings DM me in Slack.

image


public/app/plugins/datasource/cloudwatch/datasource.test.ts
```diff
@@ -1,88 +1,101 @@
...
 
 describe('datasource', () => {
   describe('query', () => {
-    it('should return error if log query and log groups is not specified', async () => {
+    it('should return error if log query and log groups is not specified', done => {
       const { datasource } = setup();
-      const response: DataQueryResponse = (await datasource
-        .query({
+
+      observableTester().subscribeAndExpectOnNext({
```
aocenas, line 13: Sorry don't want to be nitpicky and it's not really relevant for this PR and we already talked about it a bit but I still don't understand the benefits of the `observableTester` vs just doing `toPromise()` or probably better approach than both of them with `toArray().toPromise()`.

Looking at the `observableTester` and if I read the code right it does not allow to check multiple emits or multiple errors so ti basically behaves the same maybe with the distinction that if there are multiple emits the `observableTester` will fail if the second emit does not pass the check, while `toPromise` will just ignore it. In this sense `toArray().toPromise` is probable best because you can check all the values after the observable completed and check whether there are more values than you would expect for example or test them one by one.

Also some of the `observableTester` API like `subscribeAndExpectOnNextAndComplete` has:
```
      next: (value: T) => {
        try {
          expectOnNext(value);
          done();
        } catch (err) {
          done.fail(err);
        }
      },
```
Which if I understand it will call jest `done` to end the test after the first emit but it does not unsubscribe from the observable or does it?

So I am not sure if I read the code right, if not I would like to know how it works.

hugohaggmark: I understand @aocenas but I will remove all occurrences of observableTester and the observableTester and use the new jest matches introduced in #28929 instead. I'll do this in a separate PR. Thanks for pushing me in the right direction.

aocenas: Ah that matchers approach is really nice. Yeah that makes sense definitely no need to do it in this PR.

aocenas (approved):
Nice, everything seems to work fine.

papagian (approved):
LGTM: I have tested metrics and template variable query and deep linking and work as expected.
Unfortunately I didn't find either any annotation query that returns any results (but I see them run successfully at least).

hugohaggmark:
Thank you @aocenas and @papagian for testing this out!

PR #28866, opened 2020-11-05
Chore: Fix issues reported by staticcheck; enable stylecheck linter

aknuds1:

Final staticcheck PR! Fixing remaining backend linting issues, and enabling automatic style checking (with a couple of rules disabled for now).

* Fix last issues reported by staticcheck/stylecheck
* Enable stylecheck linter

### DIFF
--- a/pkg/cmd/grafana-cli/commands/commandstest/fake_ioutil.go
+++ b/pkg/cmd/grafana-cli/commands/commandstest/fake_ioutil.go
@@ -22,7 +22,7 @@ func (util *FakeIoUtil) ReadDir(path string) ([]os.FileInfo, error) {
 	return util.FakeReadDir, nil
 }
 
-func (i *FakeIoUtil) ReadFile(filename string) ([]byte, error) {
+func (util *FakeIoUtil) ReadFile(filename string) ([]byte, error) {
 	return make([]byte, 0), nil
 }
 
--- a/pkg/cmd/grafana-cli/commands/commandstest/fake_ioutil.go
+++ b/pkg/cmd/grafana-cli/commands/commandstest/fake_ioutil.go
@@ -22,7 +22,7 @@ func (util *FakeIoUtil) ReadDir(path string) ([]os.FileInfo, error) {
 	return util.FakeReadDir, nil
 }
 
-func (util *FakeIoUtil) ReadFile(filename string) ([]byte, error) {
+func (*FakeIoUtil) ReadFile(filename string) ([]byte, error) {
 	return make([]byte, 0), nil
 }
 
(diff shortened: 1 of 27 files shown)


pkg/plugins/plugins.go
```diff
@@ -367,7 +367,7 @@ func (s *PluginScanner) loadPlugin(pluginJSONFilePath string) error {
 	return nil
 }
 
-func (scanner *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
+func (s *PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
```
sakjur, line 370:
```suggestion
func (*PluginScanner) IsBackendOnlyPlugin(pluginType string) bool {
```

Since `s` isn't used, do we need to name the receiver?

I don't think we'll need it to be a pointer receiver either, but didn't change that since it shouldn't hurt if the PluginScanner is only ever used as a reference.

scripts/go/configs/.golangci.toml
```diff
@@ -44,6 +44,16 @@ enable = [
...
 
+# Enable when appropriate
+[[issues.exclude-rules]]
+linters = ["stylecheck"]
+text = "ST1003"
+
+# Enable when appropriate
+[[issues.exclude-rules]]
+linters = ["stylecheck"]
+text = "ST1001"
```
sakjur, line 55:
```suggestion
# Enable when appropriate
# Poorly chosen identifier
[[issues.exclude-rules]]
linters = ["stylecheck"]
text = "ST1003"

# Enable when appropriate
# Dot imports that aren't in external test packages are discouraged.
[[issues.exclude-rules]]
linters = ["stylecheck"]
text = "ST1001"
```

pkg/cmd/grafana-cli/commands/commandstest/fake_ioutil.go
```diff
@@ -22,7 +22,7 @@ func (util *FakeIoUtil) ReadDir(path string) ([]os.FileInfo, error) {
 	return util.FakeReadDir, nil
 }
 
-func (i *FakeIoUtil) ReadFile(filename string) ([]byte, error) {
+func (util *FakeIoUtil) ReadFile(filename string) ([]byte, error) {
```
sakjur, line 25:
```suggestion
func (*FakeIoUtil) ReadFile(filename string) ([]byte, error) {
```

sakjur (approved):
LGTM with suggestions for minor improvements

pkg/plugins/plugins.go
aknuds1: Having a look.

aknuds1:
Thanks @sakjur - applied!

PR #28868, opened 2020-11-05
CI: Add lighthouse checks to CI pipeline

kaydelaney:
(Roughly) follows [this guide](https://web.dev/lighthouse-ci/) for adding lighthouse checks to the CI pipeline. First time messing with any CI stuff, so I'm sure I made a mistake somewhere. 😛


aknuds1 (changes requested):
You will need to update the Drone Starlark source code, in scripts/*.star, and then run `drone starlark convert` to generate .drone.yml.

Can you move the lighthouse command to the "test-frontend" step instead, of adding a new step? Every pipeline step adds complexity/risk/overhead, so it's better to keep the number down.

.drone.yml
```diff
@@ -119,6 +119,14 @@ steps:
   - codespell
   - shellcheck
 
+- name: lighthouse
+  image: grafana/build-container:1.2.28
+  commands:
+  - lhci autorun || echo "LHCI failed!"
```
aknuds1, line 125: I think this is best moved to the test-frontend step, don't see any reason to not run it as part of frontend testing?

kaydelaney: Would that mean testing with a dev build? Just asking because lighthouse makes most sense when run against a production build

aknuds1: It should work the same as putting it in another step. Basically, Drone will build Grafana front-end code in the build-frontend step, and the test-frontend step will have the same checkout and build artifacts in its working directory (since Drone shares the working directory between the pipeline steps).

kaydelaney: Maybe I'm reading drone.yml wrong, but doesn't build-frontend come after test-frontend?

aknuds1: @kaydelaney Yes you're correct. Do you mean that lighthouse has to run after building? In that case you can just insert it into the `commands` list of the `build-frontend` step, right?

Does lighthouse take long BTW?

aknuds1, line 125: There must be a bug here, since `lhci` doesn't exist. Is it supposed to be in ./node_modules/.bin?

aknuds1 (changes requested):

scripts/lighthouse/login.js
```diff
@@ -0,0 +1,15 @@
...
+  await page.type('[aria-label="Password input field"]', 'admin');
+  await page.click('[aria-label="Login button"]');
+  await Promise.all([
+    page.waitForNavigation(),
+    page.waitForSelector('[aria-label="Skip change password button"]').then(a => a.click()),
+  ]);
+  await page.close();
+}
+
+module.exports = main;
```
hugohaggmark, line 15: Do we need to use puppeteer?
We moved away from pupeteer in favor of Cypress so I'm wondering if we can't execute a cypress test instead and collect the lighthouse data anyway?

aknuds1: @hugohaggmark The simpler the better :+1:

hugohaggmark (changes requested):
Do we need to use puppeteer to collect lighthouse data? This is my main concern.

kaydelaney:
Tried testing the drone pipeline locally with little success, so I've pushed some work purely to test it here. If anyone has any tips on running the drone pipeline locally I'd love to hear it

aknuds1:
I've tried Drone locally, but not much so can't provide that much insight :/

PR #28872, opened 2020-11-05
Field overrides: Added matcher to match all fields returned by a specific query 

mckn:
It adds a new field matcher that will match all fields returned in a data frame/query. This is really handy when you for example have two queries that returns the same data but with different time range (e.g. on a timeshifted query).

Skärmavbild 2020-11-05 kl  15 19 12

I'm not 100% about the naming. In the code I have tried to use the term frame refId but since we call it queries in the UI i used the query refId.

Note: this is something that have been requested from customers during the ADX work.


dprokop (approved):
Just tested it and it works great. I think the naming in the UI is ok, clear and understandable. Good stuff @mckn

PR #28875, opened 2020-11-05
FieldColor: Handling color changes when switching panel types

dprokop:
Based on https://github.com/grafana/grafana/pull/28220
Closes #28220 

Couple of changes introduced:

1. **[BREAKING]**. New API for standard options config that allows concise settings and default value config(not typed, this is triiiicky):

```ts
  type StandardOptionConfig = {
    defaultValue?: any;
    settings?: any;
  };

  /**
   * Configuration object of the standard field config properites
   *
   * @example
   * ```typescript
   * {
   *   standardOptions: {
   *     [FieldConfigProperty.Decimals]: {
   *       defaultValue: 3
   *     }
   *   }
   * }
   * ```
   */
   standardOptions?: Partial<Record<FieldConfigProperty, StandardOptionConfig>>;

   // @deprecated
   standardOptionsDefaults:  ...
```
2. API to **exclude/disable** certain standard options in `useFieldConfig` methods:
```ts
  /**
   * Array of standard field config properties that should not be available in the panel
   * @example
   * ```typescript
   * {
   *   disableStandardOptions: [FieldConfigProperty.Min, FieldConfigProperty.Max, FieldConfigProperty.Unit]
   * }
   * ```
   */
  disableStandardOptions?: FieldConfigProperty[];
```

3. **Actual field color properties configuration and handling the mode change when changing viz type**:
Field color property accepts the following config:

```ts
export interface FieldColorConfigSettings {
  byValueSupport?: boolean;
  preferThresholdsMode?: boolean;
}
```

Both of the setting properties are **true by default**. 
- When `byValueSupport: false` - only non-value based modes are available to choose from
- When `preferThresholdsMode: false` - the visualization uses is switching to will preserve the currently selected color mode
- When `preferThresholdsMode: true` - the visualization user is switching to will switch its mode to From Thresholds automatically.

TODO: 
- [ ] When switching to Gauge/Stat etc from a panel which has i.e. Static color selected, the correct color mode (from tresholds) is applied, but the visualization is not picking it up until it's refreshed.


torkelo (approved):
The code looks great, great job!

Tested this and works as expected!

PR #28888, opened 2020-11-06
grafana-cli: Fix security issue

aknuds1:
Try to fix security issue in grafana-cli reported by GitHub CodeQL.

Update: I couldn't find any way to make CodeQL detect the fix, so I ended up dismissing it as a false positive.


pkg/cmd/grafana-cli/commands/install_command_test.go
```diff
@@ -132,22 +132,6 @@ func TestInstallPluginCommand(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func TestIsPathSafe(t *testing.T) {
```
aocenas, line 135: Could we somehow keep these tests? I mean right now it's hard to figure out if we actually fixed anything. if there was an issue I would expect the problematic case to be added to the tests not that we remove all the tests.

aknuds1: @aocenas Thanks for the feedback, I will see if I can figure out the testing side of things.

aocenas: It feels like just this needs to be tested `filepath.IsAbs(zf.Name) || strings.HasPrefix(zf.Name, ".."+string(filepath.Separator))` right. So this could just be content of the `isPathSafe` and we could keep the tests. May not be the ideal code structure but at least we will know nothing got broken. Because it seems to me like this is a simpler check and so this `assert.True(t, isPathSafe("../path/dest/one", dest))` won't hold anymore. Not sure if that is ok or not (ie file path has ../ prefix but still points inside the archive) but it seems like that is valid from security perspective.

aocenas: What I remember as a problematic was for example the image renderer plugin as that containes a full chromium and that contains some dlls and symlinks so I think that was a valid use case for having `../` in the file path but not sure if I remember that 100% correctly.

aknuds1: I have to look at how this code overlaps with the current test suite, and reach a decision wrt. eventual extra tests.

Good point that paths starting with .. could potentially point inside the archive although it would be weird.

torkelo (approved):

aknuds1:
Will look this over again today.

pkg/cmd/grafana-cli/commands/install_command_test.go
aknuds1: Added tests, that verify that absolute archive members and ones pointing upwards in the filesystem are caught.

PR #28896, opened 2020-11-06
Variables: Fix backward compatibility in custom variable options that contain colon

### LINKED ISSUE #28762: Custom Dashboard Variable: preview does not show correct values when values are URL's
Screen Shot 2020-11-02 at 12 30 41 PM

**What happened**:
Creating a custom dashboard variable with URL's as values does not properly show under "Preview of Values"

Preview of Values should show correct values

Create a dashboard and go to the dashboard settings.
Select "Variables" in the left nav
Add a Variable
Select Type: Custom
In the "Values separated by comma" enter random URLs including the "http(s)://" (see attached screenshot)
Notice that the preview of values only shows the protocol.

This was not a bug in Grafana 6.6.1

**Environment**:
- Grafana version: 7.3.0. this was working fine in Grafana 6.6.1
- OS Grafana is installed on:  Ubuntu 18.04
- User OS & Browser: Mac OS, Chrome
- Grafana plugins: N/A
- Others: N/A

mckn:
This PR https://github.com/grafana/grafana/pull/27829 introduced a new way of defining key/value pairs in custom variable options. It had some negative side effects e.g. when custom value options is urls. This PR aims to prevent that from happening.

Fixes https://github.com/grafana/grafana/issues/28762

Should we do it this way `key : value` or should we change character from `:` to e.g. `|`?

### DIFF
--- a/public/app/features/variables/custom/reducer.test.ts
+++ b/public/app/features/variables/custom/reducer.test.ts
@@ -10,9 +10,9 @@ import { CustomVariableModel } from '../types';
 describe('customVariableReducer', () => {
   const adapter = createCustomVariableAdapter();
 
-  describe('when createCustomOptionsFromQuery is dispatched', () => {
-    it('then state should be correct', () => {
-      const query = 'a,b,c,d:e';
+  describe('when createCustomOptionsFromQuery is dispatched with key/value syntax', () => {
+    it('should then mutate state correctly', () => {
+      const query = 'a,b,c,d : e';
       const id = '0';
       const { initialState } = getVariableTestContext(adapter, { id, query });
       const payload = toVariablePayload({ id: '0', type: 'custom' });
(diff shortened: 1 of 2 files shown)


hugohaggmark (approved):
Great work fixing this, with this solution the docs are still intact, right?

mckn:
> Great work fixing this, with this solution the docs are still intact, right?

Yes, I would say that it still i correct.

HigH-HawK:
@mckn if I get the code right, the colon `:` will still be the delimiter but you have to put a space before and after the colon for it to be a key/value pair?

mckn:
> @mckn if I get the code right, the colon `:` will still be the delimiter but you have to put a space before and after the colon for it to be a key/value pair?
> 
> Example:
> 
> ```
> key:value (not a pair)
> key : value (a pair)
> ```

Yes, that is how this implementation works!

hugohaggmark:
> > Great work fixing this, with this solution the docs are still intact, right?
> 
> Yes, I would say that it still i correct.

Great work!

torkelo:
Think we missed updating docs with this PR

mckn:
@torkelo hmm, it looked like the previous docs was describing the feature correctly. I will have a look at it and see if I can improve them a bit.

torkelo:
yea I found it later, but was so hidden :)

PR #28898, opened 2020-11-06
Gauge: making sure threshold panel json is correct before render

### LINKED ISSUE #27146: Gauge: can't set thresholds
**What happened**:
I tried to add a threshold to a gauge

To be able to set a new value for the threshold

I don't know. I can't get it to happen when creating new gauges, only when trying to modify existing ones.

Whenever I click "Add threshold", all I get is another "Base" entry. All your base are belong to us...

Screenshot from 2020-08-23 11-57-08
Screenshot from 2020-08-23 11-57-24
Screenshot from 2020-08-23 11-57-40

**Environment**:
- Grafana version: 7.1.4
- OS Grafana is installed on: Ubuntu 20.04 LTS
- User OS & Browser: Linux, Firefox

mckn:
For some reason the author to the following issue: #27146 manage to get an invalid panel json structure saved regarding the thresholds values. It caused the threshold list to stop working. I added a filter to prevent this from happening.

Fixes #27146

Can I add this as a migration script instead? Or is this good enough?

### DIFF
--- a/packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.test.tsx
+++ b/packages/grafana-ui/src/components/ThresholdsEditorNew/ThresholdsEditor.test.tsx
@@ -1,6 +1,6 @@
 import React, { ChangeEvent } from 'react';
 import { mount } from 'enzyme';
-import { Threshold, ThresholdsConfig, ThresholdsMode } from '@grafana/data';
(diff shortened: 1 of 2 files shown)


torkelo (approved):
LGTM

jackw (approved):
LGTM! 🚀

PR #28901, opened 2020-11-06
Plugins: allow override when allowing unsigned plugins

xlson:

Fixes #


aknuds1 (approved):
Is `ReplaceAllowUnsignedPluginCondition` supposed to be called by Enterprise?

xlson:
@aknuds1 yes

pkg/plugins/plugins.go
```diff
@@ -41,26 +41,33 @@ var (
...
+	log                           log.Logger
+	scanningErrors                []error
+	allowUnsignedPluginsCondition func(plugin *PluginBase) bool
 }
 
 func init() {
 	registry.RegisterService(&PluginManager{})
 }
 
+// ReplaceAllowUnsignedPluginCondition has to be called before PluginManager.Init() has run
```
sakjur, line 66: This comment might not age very well as Grafana gets more dynamic during runtime. I've proposed an alternative:

```suggestion
// ReplaceAllowUnsignedPluginCondition changes the policy for allowing unsigned plugins. Signature validation only runs when plugins are starting
// and running plugins will not be terminated if they violate the new policy.
// Run this before PluginManager.Init to be sure policy is applied to all loaded plugins.
```

pkg/plugins/plugins.go
```diff
@@ -41,26 +41,33 @@ var (
...
+func (pm *PluginManager) ReplaceAllowUnsignedPluginCondition(condition func(base *PluginBase) bool) {
+	pm.allowUnsignedPluginsCondition = condition
```
sakjur, line 68: 🤔 Why is this a method rather than just a public instance variable? Do we expect this setter to get validation logic or something like that?

aknuds1: I was also wondering tbh.

xlson: Probably just my java roots showing. I don't have a good reason for keeping this and will remove it.

pkg/plugins/plugins.go
```diff
@@ -40,21 +40,28 @@ var (
...
-	log                  log.Logger
-	plugins              map[string]*PluginBase
+	pluginPath                    string
+	errors                        []error
+	backendPluginManager          backendplugin.Manager
+	cfg                           *setting.Cfg
+	requireSigned                 bool
+	log                           log.Logger
+	plugins                       map[string]*PluginBase
+	allowUnsignedPluginsCondition func(plugin *PluginBase) bool
```
AgnesToulet, line 53:
```suggestion
	allowUnsignedPluginsCondition unsignedPluginConditionFunc
```

PR #28905, opened 2020-11-06
devenv: Add docker load test which authenticates with API key

dafydd-t:

This provides for a use case where we wanted to authenticate with a Grafana instance with an API Key.

Also:
- Edit load tests to assign the orgId in the default function, as the setup function is only called once for all VUs and therefore the change is not persisted to each VU.
- Remove side effect from orgId setup function and explicitly set orgId in setup client

There's no issue for this PR.

Functionality of all existing test cases is unchanged. Verify the new test by generating an Admin-level API key in your grafana instance and providing it as a value for the `-k` flag.

### DIFF
--- a/devenv/docker/loadtest/README.md
+++ b/devenv/docker/loadtest/README.md
@@ -53,6 +53,12 @@ Run auth proxy test:
 $ ./run.sh -c auth_proxy_test
 ```
 
+Run API key test (option `-k` must be a valid `admin` API key)
+
+```bash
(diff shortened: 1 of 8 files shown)


wbrowne (approved):
LGTM

papagian (approved):
LGTM I've tested running auth_key_test without key, with invalid (deleted) key, viewer key and admin key and it works as expected.

PR #28909, opened 2020-11-06
TraceViewer: Make sure it does not break when no trace is passed

davkal:
There is a special case of an empty trace viewer that we did not cover: a wrong jaeger URL will return HTML with a 200 (a missing trace with a valid URL produces a 404). This fix makes sure that the trace viewer does not break even though no trace was passed.

- exit trace conversion early when there is no trace data
- added test to cover missing trace for trace viewer

Future work: we should extend the jaeger datasource to put a proper error message if HTML is detected in the trace query response.

### DIFF
--- a/packages/grafana-data/src/types/trace.ts
+++ b/packages/grafana-data/src/types/trace.ts
@@ -80,6 +80,10 @@ export type TraceData = {
   warnings?: string[] | null;
 };
 
+export type TraceViewData = TraceData & {
+  spans: TraceSpanData[];
+};
+
 export type Trace = TraceData & {
   duration: number;
   endTime: number;
--- a/public/app/features/explore/Explore.tsx
+++ b/public/app/features/explore/Explore.tsx
@@ -22,6 +22,7 @@ import {
   LogsModel,
   EventBusExtended,
   EventBusSrv,
+  TraceViewData,
 } from '@grafana/data';
 
 import store from 'app/core/store';
@@ -410,7 +411,7 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
                             // If there is not data (like 404) we show a separate error so no need to show anything here
                             queryResponse.series[0] && (
                               <TraceView
-                                trace={queryResponse.series[0].fields[0].values.get(0) as any}
+                                trace={queryResponse.series[0].fields[0].values.get(0) as TraceViewData | undefined}
                                 splitOpenFn={splitOpen}
                               />
                             )}
--- a/public/app/features/explore/TraceView/TraceView.test.tsx
+++ b/public/app/features/explore/TraceView/TraceView.test.tsx
@@ -20,6 +20,13 @@ describe('TraceView', () => {
     expect(header).toHaveLength(1);
   });
 
+  it('does not render anything on missing trace', () => {
+    // Simulating Explore's access to empty response data
+    const trace = [][0];
+    const wrapper = shallow(<TraceView trace={trace} splitOpenFn={() => {}} />);
+    expect(wrapper.isEmptyRender()).toBeTruthy();
+  });
+
   it('toggles detailState', () => {
     let { timeline, wrapper } = renderTraceView();
     expect(timeline.props().traceTimeline.detailStates.size).toBe(0);
--- a/public/app/features/explore/TraceView/TraceView.test.tsx
+++ b/public/app/features/explore/TraceView/TraceView.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { shallow } from 'enzyme';
+import { render } from '@testing-library/react';
 import { TraceView } from './TraceView';
 import { TracePageHeader, TraceTimelineViewer } from '@jaegertracing/jaeger-ui-components';
 import { TraceSpanData, TraceData } from '@grafana/data';
@@ -23,8 +24,8 @@ describe('TraceView', () => {
   it('does not render anything on missing trace', () => {
     // Simulating Explore's access to empty response data
     const trace = [][0];
-    const wrapper = shallow(<TraceView trace={trace} splitOpenFn={() => {}} />);
-    expect(wrapper.isEmptyRender()).toBeTruthy();
+    const { container } = render(<TraceView trace={trace} splitOpenFn={() => {}} />);
+    expect(container.hasChildNodes()).toBeFalsy();
   });
 
   it('toggles detailState', () => {
--- a/public/app/features/explore/TraceView/TraceView.tsx
+++ b/public/app/features/explore/TraceView/TraceView.tsx
@@ -16,18 +16,15 @@ import { useChildrenState } from './useChildrenState';
 import { useDetailState } from './useDetailState';
 import { useHoverIndentGuide } from './useHoverIndentGuide';
 import { colors, useTheme } from '@grafana/ui';
-import { TraceData, TraceSpanData, Trace, TraceSpan, TraceKeyValuePair, TraceLink } from '@grafana/data';
+import { TraceViewData, Trace, TraceSpan, TraceKeyValuePair, TraceLink } from '@grafana/data';
 import { createSpanLinkFactory } from './createSpanLink';
 
 type Props = {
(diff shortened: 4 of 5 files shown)


public/app/features/explore/TraceView/TraceView.tsx
```diff
@@ -25,6 +25,9 @@ type Props = {
 };
 
 export function TraceView(props: Props) {
+  if (!props.trace) {
+    return null;
```
aocenas, line 29: We should also change the typing somewhere. Probably we need to change the typings here to make it optional and also in parent because it was probably any or non optional there too.

davkal: Yeah, I was wondering if there is a typing solution to this that would have caught the problem in the first place.

aocenas: Yeah unfortunately we just get it from a dataFrame field so not much type safety there https://github.com/grafana/grafana/blob/a2e6f751c9648ddc2540414cfcf49673e8f7e085/public/app/features/explore/Explore.tsx#L404 but should definitely be something like `Trace | undefined`.

public/app/features/explore/TraceView/TraceView.tsx
```diff
@@ -25,6 +25,9 @@ type Props = {
 };
 
 export function TraceView(props: Props) {
+  if (!props.trace) {
```
zoltanbedi, line 28: I think this is not a proper fix. We already has this check in line 84.
What would make sense is to change the trace type in line 23 to be optional and follow the typescript error path to https://github.com/grafana/grafana/blob/davkal/fix-missing-trace/packages/jaeger-ui-components/src/model/transform-trace-data.tsx#L74 and change it to something like this.

```
export default function transformTraceData(data?: TraceData & { spans: TraceSpanData[] }): Trace | null {
  if (!data?.traceID) {
    return null;
  }
  const traceID = data.traceID.toLowerCase();
```

zoltanbedi (changes requested):

zoltanbedi:
Also we use the new backport v7.3.x label instead of cherry pick. So @grafanabot can pick it up and merge it to v7.3 branch.

public/app/features/explore/Explore.tsx
```diff
@@ -410,7 +411,7 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
                             // If there is not data (like 404) we show a separate error so no need to show anything here
                             queryResponse.series[0] && (
                               <TraceView
-                                trace={queryResponse.series[0].fields[0].values.get(0) as any}
+                                trace={queryResponse.series[0].fields[0].values.get(0) as TraceViewData | undefined}
```
hugohaggmark, line 414: just a small nit: why do we need the typings here? Is this something that could be typed further down the chain?

zoltanbedi (approved):
lgtm 🚀

public/app/features/explore/TraceView/TraceView.test.tsx
```diff
@@ -20,6 +20,13 @@ describe('TraceView', () => {
     expect(header).toHaveLength(1);
   });
 
+  it('does not render anything on missing trace', () => {
+    // Simulating Explore's access to empty response data
+    const trace = [][0];
+    const wrapper = shallow(<TraceView trace={trace} splitOpenFn={() => {}} />);
+    expect(wrapper.isEmptyRender()).toBeTruthy();
+  });
+
```
hugohaggmark, line 29: small nit: we're trying to move away from enzyme in favor of @testing-library/react. Is there some way we could take a first small step towards using @testing-library/react instead of adding to our enzyme tech debt?

packages/grafana-data/src/types/trace.ts
```diff
@@ -80,6 +80,10 @@ export type TraceData = {
   warnings?: string[] | null;
 };
 
+export type TraceViewData = TraceData & {
+  spans: TraceSpanData[];
+};
+
```
hugohaggmark, line 86: Great!

public/app/features/explore/TraceView/TraceView.tsx
```diff
@@ -16,11 +16,11 @@ import { useChildrenState } from './useChildrenState';
 import { useDetailState } from './useDetailState';
 import { useHoverIndentGuide } from './useHoverIndentGuide';
 import { colors, useTheme } from '@grafana/ui';
-import { TraceData, TraceSpanData, Trace, TraceSpan, TraceKeyValuePair, TraceLink } from '@grafana/data';
+import { TraceViewData, Trace, TraceSpan, TraceKeyValuePair, TraceLink } from '@grafana/data';
 import { createSpanLinkFactory } from './createSpanLink';
 
 type Props = {
-  trace: TraceData & { spans: TraceSpanData[] };
+  trace?: TraceViewData;
```
hugohaggmark, line 23: Awesome!

hugohaggmark (approved):
Looks outstanding, but I haven't tested it myself. Left some tiny nits for consideration.

public/app/features/explore/Explore.tsx
davkal: What's "down" for you? The producer or the consumer?

hugohaggmark: If we could get a typed queryResponse, that would be optimal, I think.

public/app/features/explore/TraceView/TraceView.test.tsx
davkal: Tried my best to mimic `isEmptyRender()`

public/app/features/explore/Explore.tsx
aocenas: The problem is that we do not exactly know what is in that dataFrame until we are in this component and we are deciding what to show so it would be more work to type that dataFrame.field properly and do all the checks in between. We should probably do that at some point but a bit too much work for this PR.

public/app/features/explore/TraceView/TraceView.tsx
aocenas: I think it actually makes sense to bail out as quickly as possible. If we do not have a trace there is nothing to render and it does not matter where it actually fails.

aocenas: But we could probably remove the later check because at that point we should be sure that it's not undefinend

aocenas: Or not really the transform do some more check, not sure if it could happen but seems like in addition if the trace does not have traceID then we also don't render anything so the second check should still be kept.

aocenas: But maybe both checks can be lifted up so we check all the requirements as the first step instead of running any hooks when we know we won't render anything anyway.

public/app/features/explore/TraceView/TraceView.tsx
```diff
@@ -16,15 +16,18 @@ import { useChildrenState } from './useChildrenState';
...
   splitOpenFn: (options: { datasourceUid: string; query: any }) => void;
 };
 
 export function TraceView(props: Props) {
+  if (!props.trace?.traceID) {
```
zoltanbedi, line 28: my only concern with this is that it break the [rules of hooks](https://reactjs.org/docs/hooks-rules.html).

davkal: Don't think so, the hooks are still at the toplevel inside that function. We just happen to execute something else before.

zoltanbedi: I'm pretty sure it is breaking the rule by conditionally running the hooks but I'm okay if we merge it like this now.

aocenas: I am not 100% sure but I think this should be a problem if you run different number of hooks during render but if you run none that may be fine?

davkal: I think it's all hooks (and in-order) or no hooks at all in a react view. At least that's what the link says.

PR #28912, opened 2020-11-07
CloudWatch Logs: Change how we measure query progress

### LINKED ISSUE #28902: CloudWatch logs: Query sometimes times out without reason
For checking if the logs query is working we are periodically polling for results and check `recordsMatched` meta attribute to see if there is any progress. This was introduced [here](https://github.com/grafana/grafana/pull/26621) to make the timeout logic more robust. Problem is that `recordsMatched` sometimes isn't progressing so probably using `recordsScanned` makes more sense in this context.

aocenas:
Fixes: https://github.com/grafana/grafana/issues/28902

For timeout we look at query progress and if there are multiple consecutive checks where no progress is made we timed out the query. We used Records matched variable which probably isn't a best measure of progress as query does not have to match any records.

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/datasource.ts
+++ b/public/app/plugins/datasource/cloudwatch/datasource.ts
@@ -336,7 +336,7 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
         ({ failures, prevRecordsMatched }, frames) => {
           failures++;
           for (const frame of frames) {
-            const recordsMatched = frame.meta?.stats?.find(stat => stat.displayName === 'Records matched')?.value!;
+            const recordsMatched = frame.meta?.stats?.find(stat => stat.displayName === 'Records scanned')?.value!;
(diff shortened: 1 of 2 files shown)


aocenas:
@kaydelaney looking at the timeout code there I don't see why it is there. I don't think it gives any assurance of timing the query out at some time cause it can slowly progress indefinitely, while it can still time it out very soon just because it won't do progress at the start. Seems like we could just have simple time based timeout logic and not actually care about progress at all especially as I don't think there are any assurances from cloud watch that the progress will be in any way linear. Was there any other issue that this part of code was solving?

kaydelaney:
@aocenas I'll double check my notes but I think you're probably right and a simpler time-based timeout would be fine. AFAIK on the CloudWatch Logs Insights side of things, it's a flat, non-configurable 15 minute timeout.

aocenas:
@kaydelaney ok lets merge this one as it should help a bit at least we will have something for 7.3.2 and we can figure out the timeout code later.

kaydelaney (approved):

PR #28913, opened 2020-11-07
GraphNG: Using new VizLayout, moving Legend into GraphNG and some other refactorings

torkelo:
* [x] In order to share aligned data I moved Legend & VizLayout to GraphNG, think this makes sense and simplifies and optimizes some things
* [x] Use new simple VizLayout from GraphNG
* [x] Remove old VizLayout & useMeasure
* [x] Made the align transform sync, fixes issue with no graph on the first render & optimizes rendering
* [x] Removed LegendPlugin and compose legend items in GraphNG now we can share the same fields loop as all the other series/lines, reuse the same calculated color etc. 
* [x] Updated uPlot 
* [x] Use new axis size feature to measure axis size 
* [x] Fixes region selection color in dark theme
* [x] Fixes x-axis time formats to use same logic as old graphs 
* [x] Changes placement type from including `under` to `bottom` and use same type everywhere 
* [x] Added storybook story for GraphNG

Noticed bugs
* First render trigger 4 renders  (might be worse in master, think this PR reduced it to only 3), on data updates, it's only 1 render. 
* Not sure how to know when last x-axis label is outside of graph and I need to add more gutter 

Screenshot from 2020-11-08 16-55-46


leeoniya:
> First render trigger 4 renders

might be related to https://github.com/leeoniya/uPlot/issues/359 though that's a 1.3 thing.

couple of issues related to initializing an empty chart and then running setData, which causes more redraw than necessary due to axis size caching. i'll have to review the geometry invalidation logic to make sure it exits deterministically. target is 1.3.1 in next few days.

packages/grafana-ui/src/components/uPlot/hooks.ts
```diff
@@ -94,9 +94,14 @@ export const DEFAULT_PLOT_CONFIG = {
   legend: {
     show: false,
   },
+  gutters: {
+    x: 0,
```
leeoniya, line 98: you probably don't want this.

with gutter:

image

without a gutter:

image

torkelo:
The first render triggers on init and 3 setData issue is not a uPlot problem but a problem with the react code

packages/grafana-ui/src/components/uPlot/hooks.ts
torkelo: Grafana's panel content container already has a 8px padding, so the graph needs no extra. Is there a way to not have label ticks on the edge?

torkelo: But I can remove the panel container gutters instead and use the uplot gutters

torkelo: But even with an 8px gutter uPlot renders x-axis outside the graph, think this needs to be fixed in uPlot as Grafana does not know where the axis ticks are rendered so cannot adjust the gutter. Might be solved in Grafana I could set uPlot right-side gutter only,  but looking at this more closely this won't work either as uplot places quite long x-axis ticks on the edge. 

Screen Shot 2020-11-08 at 09 16 30

Screen Shot 2020-11-08 at 09 07 25

torkelo: uplot issue https://github.com/leeoniya/uPlot/issues/362

packages/grafana-ui/src/components/uPlot/hooks.ts
```diff
@@ -88,15 +88,20 @@ export const DEFAULT_PLOT_CONFIG = {
   },
   cursor: {
     focus: {
-      prox: 30,
+      prox: -1,
```
dprokop, line 91: This change breaks the tooltip in single mode as the returned focused series is always null. Why introducing it?

dprokop (approved):
I think this looks good, let's follow up the things we have discussed (geometries API refactor) in the following PRs

packages/grafana-ui/src/components/uPlot/hooks.ts
torkelo: ah, I added this because hovering caused a bunch execution and renderings of axis, so this was like an optimization

PR #28915, opened 2020-11-08
Plugins: Jaeger datasource uses configured access options when making requests

### LINKED ISSUE #25699: Jaeger datasource always proxies, event when Access: Browser is selected
**What happened**: Jaeger requests are made through Grafana proxy mode, even when the datasource is configured to make requests directly from the Browser:

browser-mode
request

Appears to be the result of: https://github.com/grafana/grafana/blob/5211a23f8d921714d839898a8e503ef8987d0773/public/app/plugins/datasource/jaeger/datasource.ts#L92
**What you expected to happen**: Requests were made to the backend from the Browser directly

**How to reproduce it (as minimally and precisely as possible)**: Configure a Jaeger datasource with Access: Browser. Open Explore mode and see that requests are made to the proxy.

**Anything else we need to know?**: If the Grafana backend does not itself have access to the datasource URL then this prevents use of the Jaeger datasource plugin.

**Environment**:
- Grafana version: 7.0.3
- User OS & Browser: macOS 10.15.5, Safari 13.1.1
- Grafana plugins:
- Others:

timpalpant:
**What this PR does / why we need it**: Previously the Jaeger datasource plugin neglected to apply some of the config options available in the `DataSouceHttpSettings`, including whether to make requests directly from the browser or through the proxy. This change applies those settings to requests if they are configured.

This change allows the Jaeger datasource plugin to be used in environments where the Grafana backend may not have access to the Jaeger data source, but the Grafana user does.

**Which issue(s) this PR fixes**: #25699 

Fixes #25699


zoltanbedi (approved):
Thank you @timpalpant this looks good.

zoltanbedi:
Sorry for holding this but I want a confirmation from @davkal or @aocenas about this comment https://github.com/grafana/grafana/issues/25699#issuecomment-721659176

zoltanbedi:
Hi @timpalpant,
I'm sorry but we decided not to accept this pr. We want to move away supporting the browser access mode.

Issue #4469, opened 2016-03-24
[Feature Request] Option to align/match Y-axis width 


khenderick:
I would love to have the Y-axises' width aligned on multiple graphs. 

For example on the below screenshot, I have 2 graphs with information about a given humidity sensor, and at the bottom a graph that shows the speed of my ventilation. However, the different type of axis data cause the axises to have a different width. This makes it harder to compare the graphs because they are basically shifted and stretched compared to each other.

![grafana_alignment](https://cloud.githubusercontent.com/assets/1453214/14010150/f38ebb74-f192-11e5-8b0b-df49c5196a9c.png)

I'm running versoin 2.6.0

benrubson:
This is what @torkelo in #4349 and #4352, good idea :+1:

timgriffiths:
+1

steverweber:
also experienced this paper-cut.

bbczeuz:
I'd like to have aligned zero lines (For example have Bits/second on the left and Packets/second on the right, but zero lines of both axes should be aligned).
Semi-Workaround: If both axes have the same (manual) min or max, one can align the zero lines; but in my case I split Rx and Tx to positive-Y and negative-Y, so this doesn't help in my application.

sergeybinfinidat:
Hi, we also need this feature... How can I upvote?

simPod:
The main challenge here is IMO that the y-axis widths would have to be computed in grafana and not flot

hsmptg:
It is a pity that more than 2 years have passed since the first "Feature Request" and the problem remains! :(
Note that one reason for having several graphs in the same dashboard is the ability of correlate them, but if the time axes do not match that ability is compromised!

simPod:
@torkelo that doesn't suffice when you want to compare several charts below. When graphs's starts and ends are not aligned, it's difficult to compare whether for example a peak on _chart A_ is somehow related to peak you see on _chart B_ below because those peaks are shifted. And you have to go and try to compare crosshairs manually for each such case.

hsmptg:
The "Graph Tooltip option" even not being the perfect solution for me is an acceptable compromise.
Thanks

nharkins:
+1 for this feature request. The Graph Tooltip option only helps a little. I often screenshot a grafana dashboard (multiple graphs, vertically stacked, and ideally aligned to show correlation) into post-mortem documents.

nicolasStevenin:
+1 for this feature request.
[...]

PR #28917, opened 2020-11-08
GraphNG: update the options config

ryantxu:
Before graph-ng gets baked in, lets reconsider the options structure.  Currently it is a series of nested flags for point/line/bar/axis -- this PR proposes changing that to a flat structure with a series of enumerations.

image


ryantxu:
@dprokop -- I don't think this is the last change to the configs we want to make, but this gets us much closer to something reasonable.

dprokop:
@ryantxu I see, and I think the changes you proposed make a lot of sense. I don't like the current switches, radios are more friendly.

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -0,0 +1,75 @@
...
+  fillAlpha: number; // 0-1
+
+  points: PointMode;
+  pointRadius: number; // pixels
+  symbol: string; // eventually dot,star, etc
+
+  // Axis is actually unique based on the unit... not each field!
+  axisPlacement: AxisPlacement;
+  axisLabel?: string;
+  axisWidth?: number; // pixels ideally auto?
```
dprokop, line 51: yes!

torkelo: we can get rid of axisWidth

leeoniya: i think we will want this if there's any hope of suppprting https://github.com/grafana/grafana/issues/4469.

my suggestion was to have a toggle between auto and explicit size.

we'll want this for gutter as well.

is there a way to retain these settings in the panel json but not expose them in the ui, eg. for advanced users?

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -0,0 +1,75 @@
...
+  Always = 'always',
+  Never = 'never',
+}
+
+export enum GraphMode {
+  Line = 'line', // default
+  Bar = 'bar', // will also have a gap percent
+  Points = 'points', // use the points config
+  Staircase = 'staircase', // https://leeoniya.github.io/uPlot/demos/line-stepped.html
+  Smooth = 'smooth', // https://leeoniya.github.io/uPlot/demos/line-smoothing.html
```
torkelo, line 36: After thinking more on merging the staircase and smooth with the mode here not sure. 

I can see a future problem with this where we want to support more line interpolation modes than just staircase & smooth (for example step before step after, smooth has different algos as well). 

Another thing that feels a bit awkward is having two Points modes. GraphMode.Points and a PointsMode. Think we should hide PointsMode if GraphMode = Points (as PointsMode would not be needed then) 

aving staircase and smooth (interpolated) be part of this is that it can be limited if we want to support multiple different interpolation modes.

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -99,26 +115,26 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
-          scaleKey={scale}
-          stroke={seriesColor}
-          width={customConfig?.line.show ? customConfig?.line.width || 1 : 0}
-        />
+        <Line key={`line-${scale}-${i}`} scaleKey={scale} stroke={seriesColor} width={customConfig?.lineWidth ?? 0} />
       );
     }
 
-    if (customConfig?.points?.show) {
+    if (customConfig?.points !== PointMode.Never) {
```
torkelo, line 124: the Points.Auto does not work, always shows points. Think we need to pass this auto mode to Points and define show function (uPlot calls this) to make this dynamic. Or we can check data point count here and make this part dynamic

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -99,26 +115,26 @@ export const GraphNG: React.FC<GraphNGProps> = ({
     const colorMode = getFieldColorModeForField(field);
     const seriesColor = colorMode.getCalculator(field, theme)(0, 0);
 
-    if (customConfig?.line?.show) {
+    if (customConfig?.mode !== GraphMode.Points) {
```
torkelo, line 118: GraphMode can be bars as well

packages/grafana-ui/src/components/uPlot/config.ts
torkelo: the gutter is set in the code, not sure we need an option for gutter.

But yes to align the y-axis on panels placed on top of each other manually assigning width could be useful. Think we can maybe have an advanced section to hide uncommon options.  

> is there a way to retain these settings in the panel json but not expose them in the ui, eg. for advanced users?

Yes, that is possible :)

leeoniya: > the gutter is set in the code, not sure we need an option for gutter.

well, if the gutter becomes dynamic (auto) as we've been discussing to support edge labels, then the plotting area width becomes unpredictable and you won't be able to guarantee alignment. it's exactly the same problem as not having explicit control of the axis width on the left.

packages/grafana-ui/src/components/uPlot/config.ts
dprokop: Regarding the interpolation, I agree that it probably makes more sense to separate it from graph display mode.

Don't really understand the PointsMode, Always/Never/Auto? What auto behavior would be?

leeoniya: the default uPlot behavior is not to show points when the avg density is too high. this is a major perf benefit, and there's not much use for showing e.g. 10k points in a 600px wide graph.

the current density threshold is 2x point diam.

ryantxu: :+1: I'm moving staircase/smooth to its own setting `lineMode`

@dprokop 'auto' is a great point mode that is default in uPlot that grafana has not had.  It will show points when there are few enough that it makes sense.  Ideally we do that always

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
ryantxu: Getting `PointMode.Auto` to actually work is outside the scope of this PR -- that will take some digging into the react bindings.  these are settings on the uplot config object, but I'm not sure how we expose them via react yet

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
ryantxu: :+1:

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
ryantxu: auto | always should be passed to the configs below

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -0,0 +1,75 @@
...
+  Right = 'right',
+  Bottom = 'bottom',
+  Left = 'left',
+  Hide = 'hide',
+}
+
+export function getUPlotSideFromAxis(axis: AxisPlacement) {
+  switch (axis) {
+    case AxisPlacement.Top:
+      return 0;
```
ryantxu, line 15: Perhaps there is some better way?
https://github.com/leeoniya/uPlot/blob/master/dist/uPlot.d.ts#L461

packages/grafana-ui/src/components/uPlot/geometries/types.ts
```diff
@@ -19,6 +18,13 @@ export interface AreaProps {
   color: string;
 }
 
+export enum AxisSide {
```
ryantxu, line 21: matches:
https://github.com/leeoniya/uPlot/blob/master/dist/uPlot.d.ts#L461

packages/grafana-ui/src/components/uPlot/config.ts
leeoniya: i haz to expoze moar tipes!

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -77,48 +93,54 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
     }
 
     // need to update field state here because we use a transform to merge framesP
     field.state = { ...field.state, seriesIndex: seriesIdx };
 
     const colorMode = getFieldColorModeForField(field);
     const seriesColor = colorMode.getCalculator(field, theme)(0, 0);
-
-    if (customConfig?.line?.show) {
+    if ((customConfig.mode ?? GraphMode.Line) === GraphMode.Points) {
```
torkelo, line 123:
```suggestion
    if ((customConfig.mode ?? GraphMode.Line) === GraphMode.Line) {
```

packages/grafana-ui/src/components/uPlot/geometries/types.ts
```diff
@@ -19,6 +20,13 @@ export interface AreaProps {
...
+  Top, // 0
+  Right, // 1
+  Bottom, // 2
+  Left, // 3
```
mckn, line 27:
```suggestion
  Top = 0,
  Right = 1,
  Bottom = 2,
  Left = 3,
```
Or are there any reasons not to define it this way?

mckn: Maybe also add a code comment about the numeric values are being matched with the ones in uPlot.

mckn (approved):
I think it looks good! I guess it is easier to work with a flattened object compared to an object with lots of nesting.

dprokop:
Ah, this is now merged, but I've only merged master with it and have not tested id 😬  Hope it doesn't break anything :P

Issue #22844, opened 2020-03-17
Write functional Grafana tests


aknuds1:
Spend some time writing and improving functional Grafana tests, also convertting GoConvey tests to standard library.

As part of this, make linting stricter, using Google Sheets plugin as template.

PR #28919, opened 2020-11-08
Chore: Rewrite test helpers from GoConvey to stdlib

ceh:

Rewrite two test helpers, `createUser` and `insertTestDashboard`, in the `sqlstore` package to not require a GoConvey context, to make rewriting GoConvey tests to stdlib and testify easier.

Part of #22844.

What do you think @aknuds1, @papagian?

### DIFF
--- a/pkg/services/sqlstore/dashboard_test.go
+++ b/pkg/services/sqlstore/dashboard_test.go
@@ -25,10 +25,10 @@ func TestDashboardDataAccess(t *testing.T) {
 		InitTestDB(t)
 
 		Convey("Given saved dashboard", func() {
-			savedFolder := insertTestDashboard("1 test dash folder", 1, 0, true, "prod", "webapp")
-			savedDash := insertTestDashboard("test dash 23", 1, savedFolder.Id, false, "prod", "webapp")
-			insertTestDashboard("test dash 45", 1, savedFolder.Id, false, "prod")
-			insertTestDashboard("test dash 67", 1, 0, false, "prod", "webapp")
+			savedFolder := insertTestDashboard(t, "1 test dash folder", 1, 0, true, "prod", "webapp")
+			savedDash := insertTestDashboard(t, "test dash 23", 1, savedFolder.Id, false, "prod", "webapp")
+			insertTestDashboard(t, "test dash 45", 1, savedFolder.Id, false, "prod")
+			insertTestDashboard(t, "test dash 67", 1, 0, false, "prod", "webapp")
 
 			Convey("Should return dashboard model", func() {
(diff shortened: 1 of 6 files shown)


pkg/services/sqlstore/dashboard_test.go
```diff
@@ -442,7 +449,7 @@ func TestDashboard_SortingOptions(t *testing.T) {
 	})
 }
 
-func insertTestDashboard(title string, orgId int64, folderId int64, isFolder bool, tags ...interface{}) *models.Dashboard {
+func insertTestDashboard(title string, orgId int64, folderId int64, isFolder bool, tags ...interface{}) (*models.Dashboard, error) {
```
aknuds1, line 452: I think with test utility functions, it's better to make them take a `*testing.T` first argument and use `require.NoError(t, err)` etc. Also stick a `t.Helper()` call at the beginning of the function. You're in the context of a test, so you might as well use the `testing.T` facilities.

aknuds1 (commented):
I think that, instead of making test utilities return an error, it's better to adjust them to take a `*testing.T` first argument and use `require.NoError(t, err)` etc, as if you were right inside a test function. I don't see any reason not to use `testing.T` facilities within test helpers. WDYT?

ceh:
I updated the PR @aknuds1, please take another look.

pkg/services/sqlstore/dashboard_test.go
```diff
@@ -481,19 +482,21 @@ func insertTestDashboardForPlugin(title string, orgId int64, folderId int64, isF
...
 	err := CreateUser(context.Background(), &currentUserCmd)
-	So(err, ShouldBeNil)
+	require.NoError(t, err)
 
 	q1 := models.GetUserOrgListQuery{UserId: currentUserCmd.Result.Id}
 	err = GetUserOrgList(&q1)
-	So(err, ShouldBeNil)
-	So(q1.Result[0].Role, ShouldEqual, role)
+	require.NoError(t, err)
+	require.Equal(t, q1.Result[0].Role, models.RoleType(role))
```
aknuds1, line 499: The expectation comes before the actual value in `require.Equal`:
```suggestion
	require.Equal(t, models.RoleType(role), q1.Result[0].Role)
```

aknuds1 (approved):
LGTM

aknuds1:
Do you mind applying my suggestion before I merge?

pkg/services/sqlstore/dashboard_test.go
ceh: Ah, yes. Fixed. Thanks!

aknuds1:
Thanks for the contribution!

PR #28926, opened 2020-11-09
Chore: Bump storybook to v6

### LINKED ISSUE #25312: Bump Storybook to 6.x

jackw:
This PR makes the following changes to bump Storybook to V6:

- bumps storybook and some of its dependencies / peer dependencies
- replaces `addon-docs` and `addon-actions` with `addon-essentials`
- uses latest storybook config suggestions
- migrates Button stories to use args and the controls addon 
- provides steps required to migrate other stories
- splits the type checking from the compilation process preventing the `storybook` dev task from bailing if type errors exist
- the above process splitting also allows the `storybook` task to start up approx 50% faster

Fixes #25312

Somewhere in the bump storybook introduced a mismatch in react versions into the app compilation. Having struggled to track it down this PR tells webpack and jest to always resolve react to the project root version.

The theme switch is currently not working because of https://github.com/storybookjs/storybook/pull/11352. Hopefully we can live with this until there's a fix or 6.1 is stable.


Clarity-89:
There seem to be some issues with theme switching. The Intro menu and the Docs tab only have light theme.

peterholmberg:
@Clarity-89 updated the PR description. Maybe we should disable theme for now?

Clarity-89:
> @Clarity-89 updated the PR description. Maybe we should disable theme for now?

I wouldn't disable theme switch as it works in canvas.

package.json
```diff
@@ -35,8 +35,8 @@
     "start:ignoreTheme": "grafana-toolkit core:start --hot",
     "start:noTsCheck": "grafana-toolkit core:start --noTsCheck",
     "stats": "webpack --mode production --config scripts/webpack/webpack.prod.js --profile --json > compilation-stats.json",
-    "storybook": "cd packages/grafana-ui && yarn storybook --ci",
-    "storybook:build": "cd packages/grafana-ui && yarn storybook:build",
+    "storybook": "yarn workspace @grafana/ui storybook --ci",
+    "storybook:build": "yarn workspace @grafana/ui storybook:build",
```
dprokop, line 39: 👌

packages/grafana-ui/rollup.config.ts
```diff
@@ -43,7 +43,7 @@ const buildCjsPackage = ({ env }) => {
         // When 'rollup-plugin-commonjs' fails to properly convert the CommonJS modules to ES6 one has to manually name the exports
         // https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports
         namedExports: {
-          '../../node_modules/lodash/lodash.js': [
+          'node_modules/lodash/lodash.js': [
```
dprokop, line 46: doesn't this break grafana/ui build?

dprokop (approved):
Looks good and works! Awesome work @jackw and @peterholmberg ! Could you share the update and it's impact during next Frontend Weekly?

Clarity-89 (approved):
Works for me, just wish the build would be faster, but I think we can look into that next. Also is it possible to change the order of tabs so controls are before knobs?

jackw:
I can reduce the build to under a minute on my laptop by fiddling with the webpack config but then storybook fails to resolve certain components and the stories break. The order of tabs should be handled by the order the addons are declared [here](https://github.com/grafana/grafana/blob/storybook-bump-v6/packages/grafana-ui/.storybook/main.ts#L15-L20). I'm happy to update that in this PR or maybe we change the order once we've migrated more stories from knobs to controls?

packages/grafana-ui/rollup.config.ts
jackw: Before I made this change the build was failing at the `grafana/ui` build step with `Error: 'isString' is not exported by node_modules/lodash/lodash.js, imported by compiled/components/TimePicker/TimeZonePicker.js`. So I updated the path in rollup config to match the location of lodash mentioned in the error.

Digging a little deeper into the grafana-ui bundle output this branch introduces duplicates of lodash.

| Before | After |
| --- | --- |
| before | after |

If anyone has the time I would appreciate some input on how best to resolve this prior to merging.

dprokop: @jackw I think the duplicate is caused by direct lodash imports used in grafana/ui. So if there is an `import ... from 'lodash'` instead of `import X from 'lodash/X'` then the entire lodash.js will be added to the bundle.

dprokop: And it seems like there's plenty of such imports in grafana-ui:
image

jackw: Yep, tree shaking Lodash requires some effort on our part but if you look at the bundle visualisations you'll see lodash has moved in to the node_modules of grafana/ui in this branch and remains in the root node_modules too (due to it being used by our other deps) which then causes the bundle size to increase. Maybe I create an issue to clean this up after merging this PR? Also what would be the best way to confirm grafana/ui is still functioning as expected?

Clarity-89:
> I'm happy to update that in this PR or maybe we change the order once we've migrated more stories from knobs to controls?

Hmm, maybe display controls first, since the users familiar with knobs will just switch to knobs tab if controls is empty but if an empty knobs tab is shown by default, not everyone might realise to check the controls tab.

packages/grafana-ui/rollup.config.ts
dprokop: > Maybe I create an issue to clean this up after merging this PR?

Yes, this is a good idea to look at it separately.

> Also what would be the best way to confirm grafana/ui is still functioning as expected?

First check if it builds correctly with the updated config. Then you can test it against panel plugin template from toolkit and verify it works. You can follow this guide https://github.com/grafana/grafana/blob/master/packages/README.md#setting-up-grafana-packages-for-local-development to use the built grafana-ui locally.

contribute/style-guides/storybook.md
```diff
@@ -141,39 +141,66 @@ interface MyProps {
...
+5.  Add default props (or args in Storybook language).
+
+    ```tsx
+    Simple.args = {
+      prop1: 'Example text',
+      prop2: 'option 1',
+    };
+    ```
+
+6.  If the component has advanced props type (ie. other than string, number, boolean). You need to
```
achatterjee-grafana, line 192:
```suggestion
6.  If the component has advanced props type (ie. other than string, number, boolean), you need to
```

contribute/style-guides/storybook.md
```diff
@@ -141,39 +141,66 @@ interface MyProps {
...
+1.  Remove the `@storybook/addon-knobs` dependency.
+2.  Import the Story type from `@storybook/react`
+
+    `import { Story } from @storybook/react`
+
+3.  Import the props interface from the component you're working on (these must be exported in the component).
+
+    `import { Props } from './Component'`
+
+4.  Add the Story type to all stories in the file, replace the props sent to the component
```
achatterjee-grafana, line 161:
```suggestion
4.  Add the Story type to all stories in the file, then replace the props sent to the component
```

contribute/style-guides/storybook.md
```diff
@@ -141,39 +141,66 @@ interface MyProps {
 }
 ```
 
-### Knobs
+### Controls
 
-Knobs is an [addon to Storybook](https://github.com/storybookjs/storybook/tree/master/addons/knobs) which can be used to easily switch values in the UI. A good use case for it is to try different props for the component. Using knobs is easy. Grafana is set up so knobs can be used straight out of the box. Here is an example of how you might use it.
+The [controls addon](https://storybook.js.org/docs/react/essentials/controls) provides a way to interact with a component's properties dynamically and requires much less code than knobs. As such we're deprecating knobs in favour of using controls.
```
achatterjee-grafana, line 146:
```suggestion
The [controls addon](https://storybook.js.org/docs/react/essentials/controls) provides a way to interact with a component's properties dynamically and requires much less code than knobs. We're deprecating knobs in favor of using controls.
```

contribute/style-guides/storybook.md
```diff
@@ -141,39 +141,66 @@ interface MyProps {
...
 
-```jsx
-// In MyComponent.story.tsx
+#### Migrating a story from Knobs to Controls
 
-import { number, text } from "@storybook/addon-knobs";
+The [button story](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Button/Button.story.tsx) has been migrated as a test. Here's a basic guide on how to migrate a story to controls.
```
achatterjee-grafana, line 150:
```suggestion
As a test, we migrated the [button story](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Button/Button.story.tsx). Here's the guide on how to migrate a story to controls.
```

achatterjee-grafana (commented):
Added copy-edit suggestions.

packages/grafana-ui/rollup.config.ts
jackw: I finally got round to testing the bundle locally following your instructions @dprokop and then going through the various "build a plugin" tutorials. I also pulled it into an existing metrics plugin and all looks good. I'll create an issue shortly for the lodash cleanup.

ivanahuckova: Yeah, the same happens also with Prism and it is breaking Intellisense (as we are relying on global Prism object that should be same in grafana/ui and core). I have created a fix [here](https://github.com/grafana/grafana/pull/29381) for our local data source, but other data source plugins will be affected by this as well (e.g. X-Ray, Splunk,...)
@zoltanbedi has looked at this as well and when he discovered that when prismjs is removed from grafana/ui's node_module folder it worked again.

PR #28929, opened 2020-11-09
Tests: Adds expects for observables

hugohaggmark:
I finished the RxJs Advanced class course last week (during downtime week) and one of the Advanced class subjects was how to test observables. The course brought up two common patterns:

- The subscribe and assert approach
- Testing observables using marble tests

Here is just [one blog post that explains the details](https://medium.com/angular-in-depth/how-to-test-observables-a00038c7faad) about testing observables.

While I lean towards the second option (Testing observables using marble tests) I also realize that it's quite an enormous leap for anyone new to observables. A side note is that I found it interesting that both ways of testing Observables were in the Advanced, not the Basic RxJs course. Anyway, I thought about how to simplify this for anyone new to observables.

I've already [added a helper called observableTester](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/utils/tests/observableTester.ts) but I found it to be not as simple as I want.

My hypothesis was we could use a custom matcher in jest instead? The answer is yes, and this PR will try to show how.

__The proposed api for testing emitted values__

```ts
const observable = interval(10).pipe(take(3));
await expect(observable).toEmitValues([0, 1, 2]);
```

__The proposed api for testing emitted values but also expect something else__

```ts
const observable = interval(10).pipe(take(3));
await expect(observable).toEmitValuesWith(received => {
  expect(received).toEqual([1, 2, 3]);
  expect(someMock).toHaveBeenCalled();
});
```
One caveat with this approach is that you have to __`await the expect`__ but with observableTester you had to call it with __`the done callback`__ so I think this is cleaner. Let me know what you think.

I also looked into [jest-helpers](https://github.com/rxreact/jest-helpers) but I wasn't a perfect fit for our existing tests.


packages/grafana-data/package.json
```diff
@@ -37,11 +37,11 @@
     "@rollup/plugin-json": "4.0.3",
     "@rollup/plugin-node-resolve": "7.1.1",
     "@types/braintree__sanitize-url": "4.0.0",
-    "@types/jest": "23.3.14",
+    "@types/jest": "26.0.0",
```
hugohaggmark, line 40: I bumped @types/jest because the type definitions for jest have changed.

packages/grafana-runtime/package.json
```diff
@@ -31,7 +31,7 @@
     "@grafana/tsconfig": "^1.0.0-rc1",
     "@rollup/plugin-commonjs": "11.0.2",
     "@rollup/plugin-node-resolve": "7.1.1",
-    "@types/jest": "23.3.14",
+    "@types/jest": "26.0.0",
```
hugohaggmark, line 34: I bumped @types/jest because the type definitions for jest have changed.

packages/grafana-toolkit/package.json
```diff
@@ -111,5 +111,8 @@
   },
   "_moduleAliases": {
     "puppeteer": "node_modules/puppeteer-core"
+  },
+  "devDependencies": {
+    "@types/jest": "26.0.0"
```
hugohaggmark, line 116: I thought this was a dev dependency so I need to change this one :(

packages/grafana-ui/package.json
```diff
@@ -86,7 +86,7 @@
     "@types/classnames": "2.2.7",
     "@types/common-tags": "^1.8.0",
     "@types/d3": "5.7.2",
-    "@types/jest": "23.3.14",
+    "@types/jest": "26.0.0",
```
hugohaggmark, line 89: I bumped @types/jest because the type definitions for jest have changed.

public/app/plugins/datasource/loki/datasource.test.ts
```diff
@@ -210,84 +183,76 @@ describe('LokiDatasource', () => {
...
+        // second result always comes from runRangeQuery
+        const secondResult = received[1];
+        const dataFrame = secondResult.data[0] as DataFrame;
+        const fieldCache = new FieldCache(dataFrame);
 
-      fetchStream.next(testResponse);
-      fetchStream.next(omit(testResponse, 'data.status'));
+        expect(fieldCache.getFieldByName('line')?.values.get(0)).toBe('hello');
+        expect(dataFrame.meta?.limit).toBe(500);
+        expect(dataFrame.meta?.searchWords).toEqual([]);
```
hugohaggmark, line 209: I had to change limit and searchWords values here to get this test to pass. Seems like  __`rate({job="grafana"} |= "foo" [10m])`__ query doesn't work for range queries with streaming result as this test has. 

I think we need to change the second mockImplementation below to Matrix result instead or fix the streaming part. 

```ts
 fetchMock
  .mockImplementationOnce(() => of(testResponse))
  .mockImplementation(() => of(omit(testResponse, 'data.status')));
```

This didn't break when the query was changed in https://github.com/grafana/grafana/pull/28325 because I had messed up the test previously when I added two observableTester() that both called the done callback which meant that only one of them was tested 🙈 
Can you help me with this @ivanahuckova?

public/test/matchers/toEmitValues.test.ts
```diff
@@ -0,0 +1,125 @@
+import { interval, Observable, of, throwError } from 'rxjs';
+import { map, mergeMap, take } from 'rxjs/operators';
+import { OBSERVABLE_TEST_TIMEOUT_IN_MS } from './types';
+
+describe('toEmitValues matcher', () => {
+  describe('failing test (need to skip these for obvious reasons)', () => {
+    describe.skip('passing null in expect', () => {
```
dprokop, line 7: why the `skip` are here, is this by accident?

hugohaggmark: How do I test a failing test without failing the tests?

ivanahuckova:
I replaced me as a reviewer with @zoltanbedi as he is our observables master in Observability squad. 🕺  I am planning to go trough observability lectures/course as well, but he is currently going to be a much better reviewer than I am. 🙂

public/app/plugins/datasource/loki/datasource.test.ts
hugohaggmark: Can you help me with this test then @zoltanbedi?

zoltanbedi: Sure, I'll take a look asap.

zoltanbedi: So this is more like Loki thing than observable @ivanahuckova. Changing the limit to 500 is right the problem that it doesn't find the searchword is an actual bug however because it is a metrics query, we don't use the searchWords there. So my suggestion is to remove the searchWords check. WDYT?

yarn.lock
```diff
@@ -9444,11 +9450,6 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
-caniuse-db@1.0.30000772:
```
zoltanbedi, line 9447: I think this should be not removed.

public/test/matchers/toEmitValues.test.ts
zoltanbedi: I would write it like this
```
        try {
          await expect(observable).toEmitValues([1, 2, 3]);
        } catch (error) {
          expect(error.matcherResult.message()).toBe('');
        }
```

public/app/types/jest/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Observable } from 'rxjs';
+
+export {};
```
zoltanbedi, line 3:
```suggestion
```

Think this is not needed.

packages/grafana-data/src/types/jest/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Observable } from 'rxjs';
```
zoltanbedi, line 1: Would it make sense to not to duplicate this file and just include the typing from public app in the tsconfig.json?

zoltanbedi:
Love this PR! Awesome job Hugo. Left some comments to consider.

public/test/matchers/toEmitValues.test.ts
hugohaggmark: > I would write it like this
> 
> ```
>         try {
>           await expect(observable).toEmitValues([1, 2, 3]);
>         } catch (error) {
>           expect(error.matcherResult.message()).toBe('');
>         }
> ```

Have you tried that @zoltanbedi? I thought you couldn't catch failing excepts?

yarn.lock
hugohaggmark: Hmm thats strange, I didn't remove that?

packages/grafana-data/src/types/jest/index.d.ts
hugohaggmark: I don't think we want dependencies on our main app from grafana-data, but otherwise I agree.

public/app/types/jest/index.d.ts
hugohaggmark: Thanks Zoltan, I'll have a look.

public/app/plugins/datasource/loki/datasource.test.ts
hugohaggmark: @zoltanbedi as I wrote the mocked response is using a __`streaming`__  response not sure that is a good test case for this test as we're using a __`rate`__  query now. Feel free to change the test in this PR to something more appropiate.

public/test/matchers/toEmitValues.test.ts
hugohaggmark: > Love this PR! Awesome job Hugo. Left some comments to consider.

Thank you Zoltan! If you love it a lot maybe you could approve it? ;)

hugohaggmark:
> Love this PR! Awesome job Hugo. Left some comments to consider.

Thank you Zoltan! If you love it a lot maybe you could approve it? ;)

yarn.lock
hugohaggmark: What I mean is this is something yarn must have done as I just did __`yarn add @types/jest@26.0.0`__ in all packages folders. But I'll double-check again.

public/app/plugins/datasource/loki/datasource.test.ts
hugohaggmark: No rush with this BTW :)

public/test/matchers/toEmitValues.test.ts
zoltanbedi: Yes, I tried it

hugohaggmark: This is glorious news to me, thank you Zoltan! I'll look into all your comments after the Gardening week.

packages/grafana-data/src/types/jest/index.d.ts
zoltanbedi: It is the other way around but I agree. However we already do this 😄  if you look at the tsconfig.json in grafana-data it includes types from public app. I think this is not a problem because these are only meant for testing.

public/test/matchers/toEmitValues.test.ts
hugohaggmark: @zoltanbedi I struggled quite a bit with this, as your suggestion wouldn't work if the test was wrong and passed for some reason. Then it would never reach the catch statement, I know a bit overkill perhaps but I want these tests to be solid. Let me know what you think of the update.

public/app/types/jest/index.d.ts
hugohaggmark: Fixed!

packages/grafana-data/src/types/jest/index.d.ts
hugohaggmark: The best solution should be to have these jest.matchers in a completely separate package but instead I made the main app depend on typings in grafana-data. That makes the most sense to me at least.

public/app/types/jest/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import { Observable } from 'rxjs';
```
mckn, line 1: nit: would it be possible to use the same definition file as we use in grafana/data?

hugohaggmark: Oh noes, I forgot to remove that one. Great catch!!

public/test/matchers/toEmitValues.test.ts
```diff
@@ -0,0 +1,139 @@
+import { interval, Observable, of, throwError } from 'rxjs';
+import { map, mergeMap, take } from 'rxjs/operators';
+import { OBSERVABLE_TEST_TIMEOUT_IN_MS } from './types';
+
+describe('toEmitValues matcher', () => {
```
mckn, line 5: mindblown

mckn (approved):
Wow, this i great! Love it!

hugohaggmark:
> Wow, this i great! Love it!

Thank you @mckn

public/test/matchers/toEmitValues.test.ts
zoltanbedi: I like your approach there.

zoltanbedi (approved):
lgtm 🍎
@ivanahuckova is going to confirm the loki test update if it make sense or not

public/app/plugins/datasource/loki/datasource.test.ts
ivanahuckova: Oh sorry, I totally missed this conversation. Looking at it now.

public/test/matchers/toEmitValues.test.ts
hugohaggmark: Did I do something unclear here, could I simplify it? Can you elaborate on the mind blown thing @mckn?

public/app/plugins/datasource/loki/datasource.test.ts
hugohaggmark: @ivanahuckova maybe we shouldn't rush this. You could have a look after we merge the PR, WDYT?

ivanahuckova: Yes, I agree. The problem is (I think) that fact that we have 1 type of response for both - log and metrics queries and we should have 2 types, as they differ. But that is not a regression so this is good to be merged and I can create a PR with needed fix.

ivanahuckova (approved):
LGTM, Regarding Loki test - I can create a PR that creates second mock for test response as we use the logs response also for metrics queries.

hugohaggmark:
> LGTM, Regarding Loki test - I can create a PR that creates second mock for test response as we use the logs response also for metrics queries.

I created this issue (#29203) @ivanahuckova

public/test/matchers/toEmitValues.test.ts
mckn: Love that you are using the extension to test the extension <3

hugohaggmark: ❤️

public/test/matchers/types.ts
```diff
@@ -0,0 +1,13 @@
+import { Observable } from 'rxjs';
+
+export const OBSERVABLE_TEST_TIMEOUT_IN_MS = 1000;
```
aocenas, line 3: I wonder if it would make sense to tie it to jest `testTimeout` config value? I am not sure how to get it easily though, don't see any access for that in jest object :(. Defaults can be accessed with `const {defaults} = require('jest-config');` but you would need to reread the configs to get current value which is not great.

hugohaggmark: The jest default is 5000ms, that is too long, so I skipped it. But good suggestion.

PR #28939, opened 2020-11-09
Grafana Mixin: Reduce severity level to warning

bergquist:
This reduced the severity level to warning since it can trigger very easily on endpoints with low usage.


torkelo (approved):
Looks good

PR #28942, opened 2020-11-09
Configuration: Add an option to hide certain users in the UI

AgnesToulet:
Add an option in the configuration to hide certain users in the UI.
Hidden users are not hidden for themselves and for Grafana admins. 

As Grafana admins will continue to see all users, admin routes are not impacted by this change (except for `/api/orgs/:orgId/users` that uses a helper function shared with non-admin routes - but the behavior will not change)


sakjur (commented):
This looks really nice :+1:

I left some Go specific feedback. I'll have to spend some more time examining the SQL to see if I can find something there.

pkg/api/dtos/models.go
```diff
@@ -79,3 +79,15 @@ func GetGravatarUrlWithDefault(text string, defaultText string) string {
 
 	return GetGravatarUrl(text)
 }
+
+func IsHiddenUser(userLogin string, signedInUser *models.SignedInUser) bool {
+	if signedInUser.IsGrafanaAdmin || userLogin == signedInUser.Login {
+		return false
+	}
+
+	if _, hidden := setting.HiddenUsers[userLogin]; hidden {
```
torkelo, line 88: do not access global settings object, we are trying to get rid of this globals access

pkg/setting/setting.go
```diff
@@ -143,6 +143,7 @@ var (
 	ExternalUserMngInfo     string
 	OAuthAutoLogin          bool
 	ViewersCanEdit          bool
+	HiddenUsers             map[string]struct{}
```
torkelo, line 146: Add to Cfg struct instead, we do not want more global variables

pkg/services/sqlstore/team.go
```diff
@@ -24,26 +25,58 @@ func init() {
...
 }
 
-func getTeamSearchSqlBase() string {
+func getFilteredUsers(signedInUser *models.SignedInUser) ([]int64, error) {
+	filteredUsers := make([]int64, 0)
+	if signedInUser == nil || signedInUser.IsGrafanaAdmin {
+		return filteredUsers, nil
+	}
+
+	ids, err := GetHiddenUsersIDs()
```
torkelo, line 34: this to me feels bad to have to issue a separate DB call here, can the query read user login so we can filter based that same as we do elsewhere why do we need to filter based on id here.

torkelo (changes requested):
Looks good!  

but there are some issues adding more global variables (something we are trying to stop with), and the extra query to get hidden user ids seems unnecessary

pkg/setting/setting.go
AgnesToulet: I updated it! It's a bit hard to access settings in the database layer as functions are not part of the SqlStore service.

pkg/services/sqlstore/team.go
AgnesToulet: I updated this, it adds a INNER JOIN in the query but it shouldn't be too heavy as it's restricted to the team members.

conf/defaults.ini
```diff
@@ -296,6 +296,9 @@ editors_can_admin = false
 # The duration in time a user invitation remains valid before expiring. This setting should be expressed as a duration. Examples: 6h (hours), 2d (days), 1w (week). Default is 24h (24 hours). The minimum supported duration is 15m (15 minutes).
 user_invite_max_lifetime_duration = 24h
 
+# Enter a comma-separated list of users login to hide them in the Grafana UI. These users are shown to Grafana admins and themselves.
```
torkelo, line 299:
```suggestion
# Enter a comma-separated list of usernames to hide them in the Grafana UI. These users are shown to Grafana admins and themselves. 
```

torkelo (approved):
Looks good now, but I will leave @sakjur to leave final ok, as I did not go through the SQL, sounded like he wanted to take a close look at that

pkg/services/sqlstore/team.go
```diff
@@ -24,26 +25,54 @@ func init() {
...
-		team.email as email,
-		(SELECT COUNT(*) from team_member where team_member.team_id = team.id) as member_count,
-		team_member.permission
-		FROM team as team
-		INNER JOIN team_member on team.id = team_member.team_id AND team_member.user_id = ? `
+		team.name AS name,
+		team.email AS email,
+		team_member.permission, ` +
+		getTeamMemberCount(filteredUsers) +
+		`FROM team AS team
```
sakjur, line 64: _very_ much a nitpick, but to avoid problems in the future I think we should have a space here

```suggestion
		` FROM team AS team
```

pkg/services/sqlstore/team.go
```diff
@@ -24,26 +25,54 @@ func init() {
...
 	return `SELECT
 		team.id as id,
 		team.org_id,
 		team.name as name,
-		team.email as email,
-		(SELECT COUNT(*) from team_member where team_member.team_id = team.id) as member_count
-		FROM team as team `
+		team.email as email, ` +
+		getTeamMemberCount(filteredUsers) +
+		`FROM team as team `
```
sakjur, line 75:
```suggestion
		` FROM team as team `
```

pkg/services/sqlstore/team.go
```diff
@@ -157,14 +186,18 @@ func SearchTeams(query *models.SearchTeamsQuery) error {
 	var sql bytes.Buffer
 	params := make([]interface{}, 0)
 
+	filteredUsers := getFilteredUsers(query.SignedInUser, query.HiddenUsers)
 	if query.UserIdFilter > 0 {
-		sql.WriteString(getTeamSearchSqlBase())
+		sql.WriteString(getTeamSearchSqlBase(filteredUsers))
 		params = append(params, query.UserIdFilter)
```
sakjur, line 192: ⚠️ I'm a bit rusty, but it looks to me like you might want to swap the order for adding the user ID filter and the filtered users list? (this is the important comment)

pkg/services/sqlstore/team.go
```diff
@@ -157,14 +186,18 @@ func SearchTeams(query *models.SearchTeamsQuery) error {
...
 	} else {
-		sql.WriteString(getTeamSelectSqlBase())
+		sql.WriteString(getTeamSelectSqlBase(filteredUsers))
+	}
+	for _, user := range filteredUsers {
+		params = append(params, user)
 	}
```
sakjur, line 198:
```suggestion
	params = append(params, filteredUsers...)
```

pkg/services/sqlstore/team.go
```diff
@@ -206,12 +239,19 @@ func SearchTeams(query *models.SearchTeamsQuery) error {
 
 func GetTeamById(query *models.GetTeamByIdQuery) error {
 	var sql bytes.Buffer
+	params := make([]interface{}, 0)
+
+	filteredUsers := getFilteredUsers(query.SignedInUser, query.HiddenUsers)
+	sql.WriteString(getTeamSelectSqlBase(filteredUsers))
+	for _, user := range filteredUsers {
+		params = append(params, user)
+	}
```
sakjur, line 248:
```suggestion
	params = append(params, filteredUsers...)
```

sakjur (changes requested):
I have one question about the order of params in one of the queries, other than that it looks good

pkg/services/sqlstore/team.go
AgnesToulet: This is not compiling: `cannot use filteredUsers (type []string) as type []interface {} in append`

pkg/services/sqlstore/team.go
AgnesToulet: Good catch! Thanks!

sakjur (approved):
LGTM, we should add tests for this (but I get that we want this out asap, so we can take that in a separate PR)

AgnesToulet:
I noticed the following issue related to dashboard / folder permissions:
When a user adds or removes a permission, the whole permissions list is used. So if a user wants to update permissions when there are hidden users in the list, the permission of the hidden users are automatically deleted.

I pushed a fix but this need to be tested a bit more. I already noticed a weird case:
If we set a permission `View` for the user `test` on dashboard A and, then, a permission `Edit` for the user `test` on the folder of the dashboard A: the dashboard permissions are not valid anymore. And if the user `test` is a hidden user, it prevents every non Grafana admin users to update the permissions as it returns an error (`you can only override a permission to be higher`).

pkg/services/guardian/guardian.go
```diff
@@ -213,6 +214,38 @@ func (g *dashboardGuardianImpl) getTeams() ([]*models.TeamDTO, error) {
 	return query.Result, err
 }
 
+func (g *dashboardGuardianImpl) AddHiddenPermissions(updatePermissions []*models.DashboardAcl, cfg *setting.Cfg) ([]*models.DashboardAcl, error) {
```
xlson, line 217: I'm wondering if we want to move this out of dashboard guardian. It's very specific to what we are doing right now and doesn't fit with what the guardian is doing well. What do you think about replacing it with a `GetHiddenAcls`-function? It would do almost the same thing but not add the permissions. It means the guardian gets to stay focused on getting permissions and checking permissions instead of adding a third job.

xlson:
Looks good I think. Let me know when this is ready for testing and I'll sit down and test it as well as re-review the code.

AgnesToulet:
I added automated tests and refactored what you suggested, thanks for the feedback! It's ready for a final review.

xlson (approved):
Great work on adding tests! Looks good to go. I'm going to test it manually now.

PR #28946, opened 2020-11-09
GraphNG: Restore focus option

torkelo:
I set focus proximity to -1 as we did not use the focus highlight, did not know it would impact the tooltip. 

The reason I turned this off was that having this on triggers axis re-rendering while hovering and the axis formats function is called while hovering: 
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/uPlot/geometries/Axis.tsx#L116

Is this really needed for the focus feature @leeoniya ?

### DIFF
--- a/packages/grafana-ui/src/components/uPlot/hooks.test.ts
+++ b/packages/grafana-ui/src/components/uPlot/hooks.test.ts
@@ -10,7 +10,7 @@ describe('usePlotConfig', () => {
         "axes": Array [],
         "cursor": Object {
           "focus": Object {
-            "prox": -1,
+            "prox": 30,
           },
         },
         "focus": Object {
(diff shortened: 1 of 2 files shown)


leeoniya:
this is expected behavior.

uPlot only has a single repaint function that blows away the whole canvas and repaints everything. aside from whatever you need to do externally to uPlot (e.g. measureText), repaint is cheap because all the expensive series paths are cached in Path2D objects and not re-built. the axes & grid are typically cheap to recalc and draw, so there's no cache for them and they get called on every redraw. if this is something that ends up showing up in the perf profiles i can look into caching that stuff as well, but so far i have not found it to be worth the added complexity & extra code.

leeoniya (approved):

PR #28953, opened 2020-11-09
Refactor/Explore: Inline datasource actions into initialisation

aocenas:
Part of https://github.com/grafana/grafana/issues/28535.

What this does:
- inlines datasource related actions into the initialize action. As there is much less needed for init the datasource code there is less complex and this also allowed for deleting some unused datasource actions.
- removed `datasourceLoading` state variable. Seems like between setting this to true and then false there was no async action and so this was done in 2 sync consecutive actions. If there was anything long running in between it would block any interactions so there would be a different problem than that there is no loading indicator.


public/app/features/explore/Wrapper.test.tsx
```diff
@@ -117,7 +118,7 @@ describe('Wrapper', () => {
     );
 
     // Editor renders the new query
-    await screen.findByText(`loki Editor input: other query`);
+    await screen.findByText(`elastic Editor input: other query`);
```
aocenas, line 121: No idea how this worked before. I think because before there were more rerenders the editor did not change to new editor before this was run and so this just tested whether the old editor was still rendered.

public/app/features/explore/Wrapper.test.tsx
```diff
@@ -154,8 +155,10 @@ describe('Wrapper', () => {
     (datasources.elastic.query as Mock).mockReturnValueOnce(makeLogsQueryResponse());
 
     // Make sure we render the logs panel
-    const logsPanels = await screen.findAllByText(/^Logs$/i);
-    expect(logsPanels.length).toBe(2);
+    await waitFor(() => {
+      const logsPanels = screen.getAllByText(/^Logs$/i);
+      expect(logsPanels.length).toBe(2);
+    });
```
aocenas, line 161: `findAllByText` will still return even though it finds only one panel, this makes it a bit more robust so it waits until both panels are rendered.

public/app/features/explore/state/datasource.test.ts
```diff
@@ -1,115 +1,43 @@
...
-          };
-
-          reducerTester<ExploreItemState>()
-            .givenReducer(datasourceReducer, initialState)
-            .whenActionIsDispatched(updateDatasourceInstanceAction({ exploreId: ExploreId.left, datasourceInstance }))
-            .thenStateShouldEqual(expectedState);
-        });
-      });
-    });
+describe('Datasource reducer', () => {
```
aocenas, line 6: The `loadDatasource` was removed as it was used only in init.

public/app/features/explore/state/datasource.ts
```diff
@@ -1,54 +1,26 @@
...
+import { createEmptyQueryResponse, loadAndInitDatasource, makeInitialUpdateState } from './utils';
 
 //
 // Actions and Payloads
 //
 
-/**
- * Display an error when no datasources have been configured
- */
-export interface LoadDatasourceMissingPayload {
```
aocenas, line 23: All these were important only for init code.

public/app/features/explore/state/datasource.ts
```diff
@@ -67,108 +39,35 @@ export function changeDatasource(
...
- * Main action to asynchronously load a datasource. Dispatches lots of smaller actions for feedback.
- */
-export const loadDatasource = (exploreId: ExploreId, instance: DataSourceApi, orgId: number): ThunkResult<void> => {
-  return async (dispatch, getState) => {
-    const datasourceName = instance.name;
-
-    // Keep ID to track selection
-    dispatch(loadDatasourcePendingAction({ exploreId, requestedDatasourceName: datasourceName }));
-
-    if (instance.init) {
```
aocenas, line 155: As the `init` isn't async it probably does not have much value in having loading state. As the init isn't async, if it takes long time, it blocks the js loop so nothing would work, at that moment showing loading state isn't the most important thing. So init will probably be quick and the loading indicator should not be necessary.

public/app/features/explore/state/explorePane.test.ts
```diff
@@ -102,8 +102,8 @@ describe('refreshExplore', () => {
           .givenThunk(refreshExplore)
           .whenThunkIsDispatched(exploreId);
 
-        const initializeExplore = dispatchedActions[1] as PayloadAction<InitializeExplorePayload>;
-        const { type, payload } = initializeExplore;
+        const initializeExplore = dispatchedActions.find(action => action.type === initializeExploreAction.type);
```
aocenas, line 105: The actions tests here are not ideal as they test what is dispatched which is a bit of an internal implementation detail. Ideally at some moment the tests would just dispatch action and check final state after the action.

public/app/features/explore/state/explorePane.ts
```diff
@@ -131,9 +152,16 @@ export function initializeExplore(
...
         range,
         originPanelId,
+        datasourceInstance: instance,
+        history,
       })
     );
     dispatch(updateTime({ exploreId }));
+
+    if (instance) {
+      dispatch(runQueries(exploreId));
```
aocenas, line 162: Before the runQueries was hidden inside the `loadExploreDatasourcesAndSetDatasource` and triggered by `importQueries` switch which did not make much sense.

torkelo:
Nice!

ivanahuckova (commented):
@aocenas It seems like some datasources use the datasource name just in the ExploreURL state and some also add `datasource: name_of_datasource` to the url. In current master, datasource in **ExploreURL state** and **individually-standing** datasource are synced, but in this PR, they becomes un-synced and sometimes you have mixture of previous-current datasource in URL. 
Kapture 2020-11-18 at 13 53 25

Also, nothing new is added to query history.

public/app/features/explore/state/datasource.ts
```diff
@@ -67,108 +39,36 @@ export function changeDatasource(
...
+        datasourceInstance: instance,
+        history,
       })
     );
 
+    const queries = getState().explore[exploreId].queries;
+
     if (options?.importQueries) {
-      await dispatch(importQueries(exploreId, queries, currentDataSourceInstance, newDataSourceInstance));
+      console.log({ currentDataSourceInstance, instance });
```
zoltanbedi, line 57:
```suggestion
```

zoltanbedi (changes requested):
After refresh the query history is erased.
query-history

public/app/features/explore/Wrapper.tsx
```diff
@@ -6,19 +6,26 @@ import { StoreState } from 'app/types';
...
 }
 
 export class Wrapper extends Component<WrapperProps> {
   componentWillUnmount() {
     this.props.resetExploreAction({});
   }
 
+  componentDidMount() {
+    const richHistory = getRichHistory();
+    this.props.richHistoryUpdatedAction({ richHistory });
```
aocenas, line 26: @ivanahuckova moved this init here. I assume it did not need to be initialised for each explore pane right?

aocenas:
@zoltanbedi good catch, should be fixed now.

zoltanbedi (approved):
good job 🌶

Elfo404 (approved):
(less) code looks good! 🚀

PR #28958, opened 2020-11-10
Variables: Fixes loading with a custom all value in url

### LINKED ISSUE #28635: Data link not working with Template Variable "Custom all value"
**What happened**:
After adding a Data Link to a table using field values as well as template variables the dashboard visualization broke since the template variable field "All" is replaced by its custom value ".*" on the dropdown Menu. in order to reload the data you have to manually click on "All" again.

original variable selection:
image

Variable definition:
image

Data link:
/d/${__dashboard.uid}?&var-ProductFamily=${ProductFamily}...

After clicking in data link:
image

Data link will put "All" in template variable when "All" is selected, and then be replaced to custom value on queries

create a template variable, and add a custom value. then use `$varName=${varName}` on the table data link.

**Environment**:
- Grafana version: tested on 7.1.3 and 7.3.0
- Data source type & version: InfluxDB 1.7
- OS Grafana is installed on: linux
- User OS & Browser: windows 10 - Google Chrome

hugohaggmark:
I chose an approach that felt the less obtrusive i.e fix things when loading the dashboard with a custom all value in the url. The other option is to dive into TemplateSrv replace or change the data links logic. Neither of those options had code that I was very familiar with.

Fixes #28635

### DIFF
--- a/public/app/features/variables/shared/testing/multiVariableBuilder.ts
+++ b/public/app/features/variables/shared/testing/multiVariableBuilder.ts
@@ -10,4 +10,8 @@ export class MultiVariableBuilder<T extends VariableWithMultiSupport> extends Op
     this.variable.includeAll = includeAll;
     return this;
   }
+  withAllValue(allValue: string) {
+    this.variable.allValue = allValue;
+    return this;
+  }
 }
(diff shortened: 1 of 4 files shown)


torkelo (approved):
Looks good!

PR #28969, opened 2020-11-10
Elasticsearch: Interpolate variables in Filters Bucket Aggregation

### LINKED ISSUE #5342: Elasticsearch Template Variable formatting in Filters Agg
- **I'm submitting a ...**
- [x] Bug report

Multiselection of Template Variable does not produce correct Elasticsearch query:

I have a Template Variable called "hoststage" with values like p, q, t
I use this template variable in a Filter query like: "hoststage:($hoststage)"

Grafana only show results if I chose only one value of the hostage template variable. With multi selection I get "parse_exception: Cannot parse 'hoststage:({p, t})'
Because correct Elasticsearch syntax should be hoststage:("p";"t")

Not working with multi selected template variable values:
not_working

Manual its working:
working

I can't make a working dashboard on play.grafana.org because there is only 3.0.2 installed.

If I make a lucence query hoststage:($hoststage)  than multi selection is also working!
- Grafana v3.0.4
- Elasticsearch 2.x
- Client Win7 / Backend: RHEL7

Elfo404:
In Elasticsearch, when issuing a query with a Filters Bucket Aggregation, filters queries were are not interpolated with template variables. This PR makes sure that template variables are applied where needed.

Fixes #5342

We are applying variables also to other query fields in other parts of the datasource, this needs further refactoring as it won't currently work when going from a panel to explore

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/datasource.test.ts
+++ b/public/app/plugins/datasource/elasticsearch/datasource.test.ts
@@ -864,6 +864,34 @@ describe('ElasticDatasource', function(this: any) {
       expect(typeof JSON.parse(query.split('\n')[1]).query.bool.filter[0].range['@time'].gte).toBe('number');
     });
   });
+
+  it('should correctly interpolate variables in query', () => {
+    const query = {
+      alias: '',
+      bucketAggs: [{ type: 'filters', settings: { filters: [{ query: '$var', label: '' }] }, id: '1' }],
+      metrics: [{ type: 'count', id: '1' }],
+      query: '$var',
+    };
+
+    const interpolatedQuery = ctx.ds.interpolateVariablesInQueries([query], {})[0];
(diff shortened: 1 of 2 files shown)


ivanahuckova (approved):
Can you please add also a test for multi-select. Because that was the main source of the issue. Otherwise looks and works good! Great jooooob! 🎃

Elfo404:
> Can you please add also a test for multi-select. Because that was the main source of the issue.

@ivanahuckova this is already covered in https://github.com/grafana/grafana/blob/21ed77d7a59e94a9c47b13622f0d70a68aa60b8b/public/app/features/templating/template_srv.test.ts#L312

the problem wasn't that it wasn't working, but that it wasn't being called at all. the tests in this PR mocks out the actual templateSrv implementation.

wdyt?

PR #28972, opened 2020-11-10
LogsPanel: Don't show scroll bars when not needed

### LINKED ISSUE #25571: Disabling line wrap in Explore causes extra 'frame' to appear in Firefox
**What happened**:
Using firefox (77.0.1, macOS) with Explore, toggling the 'wrap lines' causes the rendered logs to pop in/out of some sort of hidden iframe/scroll area. I've reduced the amount of lines shown in this gif just to clearly illustrate the issue, but this is repeatable whether or not you've used text filtering or any other default settings. Does not occur in chrome. A similar issue was previously fixed in #24205 

Kapture 2020-06-12 at 14 07 11

Lines should wrap or not wrap, but this extra element should not be visible

In firefox select a Loki log stream with events. Toggle the line wrap on/off.

**Environment**:
- Grafana version: 7.0.3
- OS Grafana is installed on: Ubuntu 18
- User OS & Browser: MacOS 15.5.5 Firefox 77.0.1 
- Grafana plugins: n/a

aocenas:
Fixes: https://github.com/grafana/grafana/issues/25571

With overflow: scroll the scroll bars were always visible which seem weird when there is nothing to scroll especially if you only have the background without actual scroll bars.

### DIFF
--- a/packages/grafana-ui/src/components/Logs/getLogRowStyles.ts
+++ b/packages/grafana-ui/src/components/Logs/getLogRowStyles.ts
@@ -62,7 +62,7 @@ export const getLogRowStyles = stylesFactory((theme: GrafanaTheme, logLevel?: Lo
     `,
     logsRowsHorizontalScroll: css`
       label: logs-rows__horizontal-scroll;
-      overflow: scroll;
+      overflow: auto;
     `,
     context: context,
     logsRow: css`


ivanahuckova (approved):
LGTM 🎃

Elfo404 (approved):

PR #28990, opened 2020-11-10
Annotations: fixing so when changing annotations query links submenu will be updated.

### LINKED ISSUE #28782: Dashboard: Adding an annotation query or dashboard link does not show up after going back to dashboard
After adding an annotation query dashboard link I expect to see them when I go back to the dashboard but I do not. I have to save and reload the dashboard for them to be shown in the SubMenu. 

Probably due to react props not changing. 

https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/components/SubMenu/SubMenu.tsx#L63

mckn:
This PR will make sure that we trigger an update to the annotations list in the SubMenu when doing changes to the annotation query links.

Fixes #28782

The reason why this is happening is that the following:

1. Changes to the dashboard model will trigger a render of the DashboardPage.
1. It will not trigger a render on the SubMenu component due to the annotations list is being mutated so it will keep the same reference after it has been changed:
     - Since we are connecting the SubMenu component to redux a connected HOC PureComponent will be returned. We can override this behavior by passing an options object to the connect function.
     - The SubMenu component is also a PureComponent so we need to update that one to be a Component.

So instead of changing both the SubMenu and the connected SubMenu to regular components I added a dependency on the AppEvent that is being triggered when we change the annotations list.

**Edit**: refactored this so we don't mutate the annotations list but instead create a new list when doing changes to any of the annotations or adding new ones.


kaydelaney (approved):
LGTM

public/app/features/dashboard/components/SubMenu/Annotations.tsx
```diff
@@ -1,18 +1,29 @@
...
     setVisibleAnnotations(annotations.filter(annotation => annotation.hide !== true));
   }, [annotations]);
 
+  useEffect(() => {
+    const handler = () => {
+      setVisibleAnnotations(annotations.filter(annotation => annotation.hide !== true));
+    };
+    events.on(CoreEvents.submenuVisibilityChanged, handler);
+    return () => events.off(CoreEvents.submenuVisibilityChanged, handler);
+  }, []);
```
torkelo, line 25: would prefer to not use the bus for this, if can make the annotations array become a new instance when it gets updated (or visibility changed), we can just pass this annotations array to SubMenu and use normal react prop change detection to trigger rendering

mckn: OK ok, I will look into that during the day. Great feedback, thanks 🙏

public/app/features/annotations/editor_ctrl.ts
```diff
@@ -116,25 +124,49 @@ export class AnnotationsEditorCtrl {
...
+
+      if (i === index + dir) {
+        target.push(annotation);
+      }
+
+      if (dir < 0) {
+        target.push(current);
+      }
+    }
+
```
torkelo, line 150: this seems like a lot of code for a move, is there no more terse way to do this? something like 

```
const move = (arr, from, to) => {
  const clone = [...arr];
  Array.prototype.splice.call(clone, to, 0,
    Array.prototype.splice.call(clone, from, 1)[0]
  );
  return clone;
};
```

public/app/features/dashboard/state/DashboardModel.ts
```diff
@@ -48,6 +48,25 @@ export interface DashboardLink {
...
 }
 
+type AnnotationType = 'tags' | 'dashboard';
+
+/**
+ * Typings for Annotation is added only for the basic scenarios.
+ * We need to improve the typings to support custom annotations
+ * added by different types of datasources.
+ */
+export interface Annotation {
```
torkelo, line 58: Instead of adding this use this type  https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/annotations.ts#L10

torkelo (approved):
Looks good, found two minor issues

PR #28992, opened 2020-11-10
Dashboards: Hide playlist edit functionality from viewers and snapshots link from unauthenticated users

### LINKED ISSUE #24987: Create new playlist button shouldn't be visible for user with viewer role on playlists page
**What happened**:
We have our **unauthenticated** user set to "Viewer" permission level. Within Grafana, the user is able to navigate into the left-side menu items. Relevant here are the routes `playlists` and `dashboard/snapshots`, where the user may view their respective contents. The issue is that the calls to the server either prevent them from creating a new / manipulating old playlists (for `playlists` route), or present them with an empty list of snapshots (for `dashboards/snapshots` route, even when existing snapshots are purely local snapshots). The `playlists` case is met with a "403 Forbidden" with response data `{message: "Permission denied"}`.
Given that the **unauthenticated** user can view these areas, we expected them to be able to see the list of existing snapshots and/or be able to create/manipulate playlists.
Set **unauthenticated** user in custom.ini or default.ini to have the "Viewer" permission level by setting `org_role = Viewer`. Attempt to create a playlist or view existing snapshots from existing dashboards.
n/a
**Environment**:
- Grafana version: 7.0.0
- Data source type & version: Prometheus 2.13.1
- OS Grafana is installed on: Windows Server 2012
- User OS & Browser: Windows Server 2012 & Google Chrome 81.0.4404.122
- Grafana plugins:
- Others:

EDIT: Used the word "unauthorized" rather than "unauthenticated".

jackw:
Viewing snapshots and creating playlists should not be available for users with "viewer" permissions. This PR:
- hides the snapshot nav link from unauthenticated viewers as the snapshot table is always blank.
- fixes a bug which caused snapshot to multiply in the UI if backend error'd.
- hides the playlists create / edit functionality from viewers
- if no playlists exist shows an info box stating the user doesn't have permission to create playlists

#### Viewer

| Without Playlists | With Playlists |
| --- | --- |
| Screenshot 2020-11-11 at 11 38 33 | Screenshot 2020-11-11 at 11 39 39 |

#### Admin / Editor

| Without Playlists | With Playlists |
| --- | --- |
| Screenshot 2020-11-11 at 11 38 13 | Screenshot 2020-11-11 at 11 39 22 |

Fixes #24987

This is my first attempt at writing Go and I'm not overly familiar with the permissions levels or auth settings so input there would be really helpful. Should there be some sort of authorisation on the snapshots route as well as hiding it in the FE from "viewers"?

I'd also appreciate any feedback on the info message displayed for viewers when no playlists exist.

### DIFF
--- a/pkg/api/index.go
+++ b/pkg/api/index.go
@@ -165,7 +165,7 @@ func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool) ([]*dto
 		{Text: "Playlists", Id: "playlists", Url: setting.AppSubUrl + "/playlists", Icon: "presentation-play"},
 	}
 
-	if c.OrgRole != models.ROLE_VIEWER {
+	if hasEditPerm {
 		dashboardChildNavs = append(dashboardChildNavs, &dtos.NavLink{
 			Text: "Snapshots",
 			Id:   "snapshots",
--- a/pkg/api/index.go
+++ b/pkg/api/index.go
@@ -131,7 +131,7 @@ func getAppLinks(c *models.ReqContext) ([]*dtos.NavLink, error) {
 	return appLinks, nil
 }
 
-func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool) ([]*dtos.NavLink, error) {
+func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool, hasPlaylists bool) ([]*dtos.NavLink, error) {
 	navTree := []*dtos.NavLink{}
 
 	if hasEditPerm {
@@ -162,9 +162,16 @@ func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool) ([]*dto
 		{Text: "Home", Id: "home", Url: setting.AppSubUrl + "/", Icon: "home-alt", HideFromTabs: true},
 		{Text: "Divider", Divider: true, Id: "divider", HideFromTabs: true},
 		{Text: "Manage", Id: "manage-dashboards", Url: setting.AppSubUrl + "/dashboards", Icon: "sitemap"},
-		{Text: "Playlists", Id: "playlists", Url: setting.AppSubUrl + "/playlists", Icon: "presentation-play"},
 	}
 
+	if c.OrgRole != models.ROLE_VIEWER || hasPlaylists {
+		dashboardChildNavs = append(dashboardChildNavs, &dtos.NavLink{
+			Text: "Playlists",
+			Id:   "playlists",
+			Url:  setting.AppSubUrl + "/playlists",
+			Icon: "presentation-play",
+		})
+	}
 	if c.OrgRole != models.ROLE_VIEWER {
 		dashboardChildNavs = append(dashboardChildNavs, &dtos.NavLink{
 			Text: "Snapshots",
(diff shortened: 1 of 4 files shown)


dprokop (approved):
I think this looks good code wise. I would like to confirm with @torkelo or maybe @marefr as they have more knowledge about permissions than I do.

pkg/api/index.go
```diff
@@ -159,7 +159,15 @@ func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool) ([]*dto
 		{Text: "Divider", Divider: true, Id: "divider", HideFromTabs: true},
 		{Text: "Manage", Id: "manage-dashboards", Url: setting.AppSubUrl + "/dashboards", Icon: "sitemap"},
 		{Text: "Playlists", Id: "playlists", Url: setting.AppSubUrl + "/playlists", Icon: "presentation-play"},
-		{Text: "Snapshots", Id: "snapshots", Url: setting.AppSubUrl + "/dashboard/snapshots", Icon: "camera"},
+	}
+
+	if c.OrgRole != models.ROLE_VIEWER {
```
torkelo, line 164: this is better types has if c.HasRole(models.ROLE_EDITOR)   (will be true for admins as well)

torkelo (approved):
Looks good just one small issue

pkg/api/index.go
marefr: I think this check need to consider admin/editor permission in folders/dashboards and should use `hasEditPerm` argument instead. But, needs to be verified.

torkelo: @marefr since viewers cannot create playlist today, I think we can wait with this until we make it possible to create playlists inside a folder

marefr: @torkelo yes that's true but this check is specifically for snapshots and not playlists, right? Platlist navitem is above.

torkelo: Ah , Sam your right . Mistook this item for playlists

marefr: @torkelo @jackw need some changes to move this PR forward I suppose?

marefr (dismissed):
See comment regarding snapshots

pkg/api/index.go
```diff
@@ -159,7 +159,15 @@ func (hs *HTTPServer) getNavTree(c *models.ReqContext, hasEditPerm bool) ([]*dto
 		{Text: "Divider", Divider: true, Id: "divider", HideFromTabs: true},
 		{Text: "Manage", Id: "manage-dashboards", Url: setting.AppSubUrl + "/dashboards", Icon: "sitemap"},
 		{Text: "Playlists", Id: "playlists", Url: setting.AppSubUrl + "/playlists", Icon: "presentation-play"},
-		{Text: "Snapshots", Id: "snapshots", Url: setting.AppSubUrl + "/dashboard/snapshots", Icon: "camera"},
+	}
+
+	if hasEditPerm {
```
marefr, line 164: I just tried with viewer role and I can actually create snapshots so don't think this. Sorry, but think snapshots shouldn't have any permission check here. WDYT?

marefr: When created a snapshot as viewer I can (in master) see the snapshots I've created:
image

jackw: Thanks for highlighting this Marcus. Turns out I misread the original issue which states the snapshot table is always empty for an unauthenticated user.

Based on this I'm guessing we either change the logic here to remove the snapshots menu item if there isn't a logged in user or revert this backend change and place a note in the frontend stating something like "you need to be logged in to see snapshots". What do you think?

marefr: Not authenticated - no snapshot menu item should work I think

wbrowne (commented):
Go code looks good @jackw. Tested it out locally and works AFAICT. One thing I wanted to confirm was that with this change, it's still possible for a viewer to create and view their own snapshots (just not anyone else's)?

jackw:
@wbrowne that's correct. I've not changed any logic around what a viewer can do with snapshots only that an unauthenticated user won't be presented with a nav link that leads to an empty table. The PR is a little misleading - I've updated the title and description so it hopefully makes more sense.

wbrowne (approved):
👍

PR #29004, opened 2020-11-11
BackendSrv: support binary responseType like $http did

ryantxu:
When we switched from the angular $http based service to using fetch, this broke external plugins using the backendSrv for binary data.  See https://github.com/ryantxu/ajax-panel

This PR maps the 4 common settings to the fetch equivalent -- it does not support `document()` since XML is so 2010


public/app/core/services/backend_srv.ts
```diff
@@ -175,15 +175,33 @@ export class BackendSrv implements BackendService {
...
+            case 'json':
+              data = await response.json();
+              break;
+            case 'text':
+              data = (await response.text()) as any;
+              break;
+          }
+        }
+        // @ts-ignore
+        if (!data) {
```
ryantxu, line 197: @hugohaggmark any suggestions for a more clean approach?

hugohaggmark: @ryantxu I would try to use a handler/strategy pattern [like we do for headers](https://github.com/grafana/grafana/blob/b2dcf06b6017aab7719ccf3a8e4de389c9b6c1b7/public/app/core/utils/fetch.ts#L59).

We could also turn this into separate RxJs operators. That would be super neat now that I think about it more, but handler pattern would do perfectly too and we already use it.

ryantxu: @hugohaggmark can you take another look?  Is this what you were thinking?

hugohaggmark: Hmm, not really what I meant, but this is a great start!

public/app/core/utils/fetch.ts
```diff
@@ -91,6 +91,33 @@ export const parseBody = (options: BackendSrvRequest, isAppJson: boolean) => {
...
+    }
+  }
+
+  const textData = await response.text(); // this could be just a string, prometheus requests for instance
+  try {
+    return JSON.parse(textData); // majority of the requests this will be something that can be parsed
+  } catch {}
+  return textData as any;
+}
+
```
hugohaggmark, line 120: Now that this is its own method, can we add some tests for it?

hugohaggmark (approved):
Looks a lot better ⭐  but parseResponseBody just needs some tests.

public/app/core/utils/fetch.ts
ryantxu: :+1:

PR #29006, opened 2020-11-11
DataFrames: add utility function to check if structure has changed

ryantxu:
In the DataFrame model, it is hard to know the difference between having the same shape response with new data vs a whole new shape.  This PR introduces:
* a utility function to check if the structure is the same two responses
* adds a property to PanelData that will change when the structure changes

The function names are terrible! any better ideas


packages/grafana-data/src/dataframe/frameComparisons.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+      if (cfgA[key] !== cfgB[key]) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+export function arrayCompare<T>(a: T[], b: T[], fn: (a: T, b: T) => boolean) {
```
dprokop, line 39: Maybe you could use https://lodash.com/docs/4.17.15#isEqualWith instead?

packages/grafana-data/src/types/panel.ts
```diff
@@ -28,6 +28,14 @@ export interface PanelData {
   /** Contains data frames with field overrides applied */
   series: DataFrame[];
 
+  /**
+   * This flag will change when the structure is different from the previous response
+   * The structure includes frames, names, types, and any config properties
+   *
+   * 0 or undefined indicates that the change status is unknown or not computed
+   */
+  seriesStructureChanged?: number;
```
dprokop, line 37: Name suggests a bool value. Alternatives I would consider:

- `seriesStructureRev` | `seriesStructureRevision`
- `seriesStructureToken`

packages/grafana-data/src/types/panel.ts
```diff
@@ -28,6 +28,14 @@ export interface PanelData {
   /** Contains data frames with field overrides applied */
   series: DataFrame[];
 
+  /**
+   * This flag will change when the structure is different from the previous response
+   * The structure includes frames, names, types, and any config properties
+   *
+   * 0 or undefined indicates that the change status is unknown or not computed
```
dprokop, line 35: Is this true? Looking at the code it seems like the value can be either `undefined` or >1000

packages/grafana-data/src/dataframe/frameComparisons.ts
```diff
@@ -0,0 +1,52 @@
+import { DataFrame } from '../types/dataFrame';
+
+/**
+ * Returns true if both frames have the same list of fields and configs.
+ * Each field can have diferent names, labels and values
+ */
+export function framesHaveSameStructure(a: DataFrame, b: DataFrame): boolean {
```
dprokop, line 7: Maybe `compareDataFramesStructure`?

dprokop (commented):
Think this looks good, left some comments to discuss

packages/grafana-data/src/dataframe/frameComparisons.ts
ryantxu: replaced with `compareDataFrameStructures`

packages/grafana-data/src/dataframe/frameComparisons.ts
ryantxu: I tried... but it was not happy with that -- `isEqualWith` will still pass in the whole array, not each element in the array

packages/grafana-data/src/types/panel.ts
ryantxu: FYI: removing the changes to property for now.  Lets start with a utility function, and perhaps add it as a higher level property if it is more widely useful and efficient

packages/grafana-data/src/dataframe/frameComparisons.ts
```diff
@@ -8,6 +8,9 @@ import { DataFrame } from '../types/dataFrame';
  * ```
  * areArraysEqual(a, b, framesHaveSameStructure);
  * ```
+ * NOTE: this does a shallow check on the FieldConfig properties, when using the query
+ * editor, this should be sufficient, however if applicaitons are mutating properties
+ * deep in the FieldConfig this will not recognize a change
```
leeoniya, line 13: hmm, definite footgun. maybe the default should be deep diff and have a shallow flag if you know what you're doing?

leeoniya: or i guess if you think shallow covers 80% of cases, then a deep flag instead?

ryantxu: It should cover 100% of cases where a datasource or options are changed from the editor.  If we don't catch a change, the function should be updated so it does 😀

torkelo (approved):

PR #29010, opened 2020-11-11
Thresholds: Fixes color assigned to null values

### LINKED ISSUE #29008: Color thresholds applied on null values
**What happened**:

Table with the following values and thresholds (null is value mapped to Bla)

image

image

image

Non-numeric values like null not to be used in the thresholds.

https://gist.github.com/wardbekker/5319d94c27fbe85b79c417a09084856a

**Environment**:
- Grafana version: 7.3.1.
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

torkelo:
Fixes #29008

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.test.ts
+++ b/packages/grafana-data/src/field/displayProcessor.test.ts
@@ -16,10 +16,8 @@ function getDisplayProcessorFromConfig(config: FieldConfig) {
 function assertSame(input: any, processors: DisplayProcessor[], match: DisplayValue) {
   processors.forEach(processor => {
     const value = processor(input);
-    for (const key in match) {
-      if (match.hasOwnProperty(key)) {
-        expect((value as any)[key]).toEqual((match as any)[key]);
-      }
+    for (const key of Object.keys(match)) {
+      expect((value as any)[key]).toEqual((match as any)[key]);
     }
   });
 }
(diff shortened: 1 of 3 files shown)


packages/grafana-data/src/field/displayProcessor.test.ts
```diff
@@ -16,9 +16,10 @@ function getDisplayProcessorFromConfig(config: FieldConfig) {
 function assertSame(input: any, processors: DisplayProcessor[], match: DisplayValue) {
   processors.forEach(processor => {
     const value = processor(input);
-    expect(value.text).toEqual(match.text);
-    if (match.hasOwnProperty('numeric')) {
-      expect(value.numeric).toEqual(match.numeric);
+    for (const key in match) {
+      if (match.hasOwnProperty(key)) {
```
Clarity-89, line 20: NIT: I think this could be simplified with `for ..of` as then `hasOwnProperty` check won't be needed - `for (const key of Object.keys(match))`.

Clarity-89 (approved):

packages/grafana-data/src/field/displayProcessor.test.ts
torkelo: yea, maybe. Just never liked creating a new array just to be able to loop over properties

torkelo: but agree that is probably the preferred/best practice method

PR #29023, opened 2020-11-11
Chore: Bump build pipeline version

sakjur:
This doesn't change anything for the OSS version of Grafana, just a bump since we have a new version of the build pipeline.

@aknuds1 Should we backport this to the v7.3.x branch?

PR #29025, opened 2020-11-11
[v7.3.x] Chore: Bump build pipeline version

sakjur:
Backport of #29023

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,7 +17,7 @@ steps:
   image: grafana/build-container:1.2.28
   commands:
   - mkdir -p bin
-  - curl -fL -o bin/grabpl https://grafana-downloads.storage.googleapis.com/grafana-build-pipeline/v0.5.20/grabpl
+  - curl -fL -o bin/grabpl https://grafana-downloads.storage.googleapis.com/grafana-build-pipeline/v0.5.25/grabpl
   - chmod +x bin/grabpl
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -2376,7 +2376,7 @@ steps:
   - package
 
 - name: end-to-end-tests
-  image: grafana/ci-e2e:12.19.0-1
+  image: grafana/ci-e2e:12.18-1
```
aknuds1, line 2379: Should this be downgraded?

aknuds1 (changes requested):
Should figure out why grafana/ci-e2e is being downgraded.

.drone.yml
sakjur: 👀 That's weird

aknuds1: @sakjur Does it also happen if you re-generate .drone.yaml?

sakjur: Yes, I even deleted `.drone.yml` before generating it to make sure it didn't have any artifacts.

And it makes sense, because `lib.star` points to 12.18, so seems like it could be something like an accidental backport 🤔

https://github.com/grafana/grafana/blob/v7.3.x/scripts/lib.star#L608

We can go up or stay down, but I'm not sure which one we should go for.

aknuds1: I'll check.

aknuds1: @sakjur I think we can safely upgrade. The difference is just a bump of Node minor version and installing of Git into the image.

aknuds1 (approved):

PR #29031, opened 2020-11-11
Snapshots: Fixes so that dashboard snapshots show data when using Stat, Gauge, BarGauge or Table panels

### LINKED ISSUE #29005: 7.2 Upgrade Grafana Snapshot: Panels loading with no plot/no data which are using Table, Stat and Bar gauge
**What happened**: The panels are loading empty or no data when we generate snapshots on the dashboard where we used plugins like Table with transformations, Gauge, Stat and Bar gauge. Also tried increasing timeout which didn't work.

**What you expected to happen**: Dashboards with any type of plugin should be able to load with data properly in Snapshot.

**How to reproduce it (as minimally and precisely as possible)**: Create a snapshot on a dashboard where we used plugins like Table with transformations, Gauge, Stat and Bar gauge. Please refer Snaphot: https://play.grafana.org/dashboard/snapshot/RrFFR3IWErhDpbXo8mjt2djc5NXs35PI?orgId=1 generated on: https://play.grafana.org/d/0KapoFkMk/1-new-features-in-v7-0?orgId=1&refresh=5s also able to reproduce it locally.

**Environment**:
- Grafana version: 7.2
- Data source type & version: Any
- Grafana plugins: Table with transformations, Gauge, Stat and Bar gauge. 
- Others:

torkelo:
Fixes #29005

Not sure if this is the best fix, since this appears to have been broken for some time maybe it's best to change how we store snapshot data.
Right now we store data that has been transformed & overrides applied in the panel JSON. But when loading it we still have to pass the data
through applyFieldOverrides so that the data frames get displayProcessors & etc.

Alternative would be to store raw query data and when showing a snapshot have it go through the normal data paths. Would require a bigger change
and would likely breaking to some extend (ie break some old snapshots that did work).

### DIFF
--- a/public/app/features/dashboard/utils/loadSnapshotData.ts
+++ b/public/app/features/dashboard/utils/loadSnapshotData.ts
@@ -1,0 +1,27 @@
+import { applyFieldOverrides, DefaultTimeRange, LoadingState, PanelData } from '@grafana/data';
+import { config } from 'app/core/config';
+import { getDatasourceSrv } from 'app/features/plugins/datasource_srv';
+import { DashboardModel, PanelModel } from '../state';
+import { getProcessedDataFrames } from '../state/runRequest';
+
+export function loadSnapshotData(panel: PanelModel, dashboard: DashboardModel): PanelData {
+  const data = getProcessedDataFrames(panel.snapshotData);
+
+  return {
+    timeRange: DefaultTimeRange,
+    state: LoadingState.Done,
+    series: applyFieldOverrides({
+      data,
+      fieldConfig: {
+        defaults: {},
+        overrides: [],
+      },
+      autoMinMax: true,
+      replaceVariables: panel.replaceVariables,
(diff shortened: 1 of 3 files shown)


public/app/features/dashboard/utils/loadSnapshotData.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return {
+    timeRange: DefaultTimeRange,
+    state: LoadingState.Done,
+    series: applyFieldOverrides({
+      data,
+      fieldConfig: {
+        defaults: {},
+        overrides: [],
+      },
```
dprokop, line 18: Won't this break data links if they are stored in the snapshot?

dprokop (approved):
I see your considerations @torkelo. Left one question I am not sure about, but for bringing back the snapshots to work I think we should merge this as a temp solition.

public/app/features/dashboard/utils/loadSnapshotData.ts
torkelo: should already be on the data frame stored in the panel as the stores snapshot data already has gone through this

PR #29032, opened 2020-11-11
Dashboards: Exit to dashboard when deleting panel from panel view / edit view

### LINKED ISSUE #11941: [BUG] Deleting a panel from the edit mode, leaves the page blank
Grafana v5.1.2
Graphite
High Sierra 10.13.14
### What did you do?
I deleted a panel from the edit mode of that same panel.
Be taken out of the edit mode, and see my whole dashboard without that panel.
### What happened instead?
The whole dashboard dissapeared, and I had to refresh to see my dashboard again.

kaydelaney:
Previously after deleting a panel while looking at that panel via either the panel view or edit view, you would still be on those views after deleting. This PR makes it so you are brought back to the dashboard after deleting the panel.

Closes #11941


public/app/features/dashboard/utils/panel.ts
```diff
@@ -39,7 +41,14 @@ export const removePanel = (dashboard: DashboardModel, panel: PanelModel, ask: b
     });
     return;
   }
+
   dashboard.removePanel(panel);
+  reduxStore.dispatch(
```
torkelo, line 46: Think best to only do this viewPanel !== null , not possible to delete from edit mode so there should be no need to nullify editPanel and tab

kaydelaney: It's possible to delete if you use the `p r` shortcut

torkelo: Interestingly, seems to not delete the dashboard your editing though :)  Should probably not allow delete in either view or edit mode

kaydelaney: So just disallow use of the shortcut in those modes? Any thoughts from a UX perspective @jessover9000 ?

jessover9000: Hmm good question. I guess deleting during edit mode doesn't make that much sense, the usual UX pattern would be either deleting or editing and not both at the same time. But why should you not be able to delete in view mode?

mckn: Did I understand the suggested behavior correct like this:

1. You should only be able to use the `p r` shortcut when in settings? Not in view or edit mode.
2. Not display or disable the "delete dashboard" button in settings if you currently is in edit mode.

I'm not sure I am a big fan of the solution in step 2. since the settings will behave differently depending on how you navigate to it. But hopefully I missed something.

mckn (commented):
I'm awaiting decision regarding how we want this to behave.

public/app/features/dashboard/utils/panel.ts
torkelo: > You should only be able to use the p r shortcut when in settings? Not in view or edit mode.

in settings? no in the dashboard (non edit/view mode). 

> I'm not sure I am a big fan of the solution in step 2. since the settings will behave differently depending on how you navigate to it. But hopefully I missed something.

not sure where you get settings from :) what settings do you mean? 

All we are talking about is disabling the shortcut while in edit & view mode, and the delete menu option from the view mode (already removed from edit mode)

mckn: Hehe, so then I am totally lost.

lost

Need to re-test this one to give proper feedback.

torkelo (approved):
LGTM

PR #29033, opened 2020-11-11
annotations: fetch all grafana annotations when no tags or dashboard ID is specified

kevingessner:

There's currently no way to have a dashboard/panel display all Grafana-stored annotations -- you can either show them just on the panel where they were created, or by tag. This changes the behavior of the Tags type with no tags specified: it now fetches all annotations, instead of none. This enables showing all annotations, instead of an otherwise unuseful behavior. 

I ran the go and jest tests and they pass.

### DIFF
--- a/public/app/plugins/datasource/grafana/datasource.test.ts
+++ b/public/app/plugins/datasource/grafana/datasource.test.ts
@@ -71,7 +71,52 @@ describe('grafana data source', () => {
         return ds.annotationQuery(options);
       });
 
+      it('should query by dashboard ID', () => {
+        expect(calledBackendSrvParams.dashboardId).toBe(1);
+      });
       it('should remove tags from query options', () => {
+        expect(calledBackendSrvParams.tags).toBe(undefined);
+      });
+    });
+
+    describe('with type tags and tags', () => {
+      const options = setupAnnotationQueryOptions(
+        {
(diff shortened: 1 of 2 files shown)


hugohaggmark (changes requested):
@kevingessner I'm sorry, but we rarely take contributions without a feature request that has a reasonable amount of feedback/upvotes. We'll probably have a look next week.

All the best

kevingessner:
@hugohaggmark Would it be helpful if I created a feature request issue to go along with this?

hugohaggmark:
> @hugohaggmark Would it be helpful if I created a feature request issue to go along with this?

Yes, that would be a great start, still that doesn't mean we'll accept this PR. It depends on popular demand like feedback/upvotes, complexity e.t.c. Thank you

kevingessner:
I've submitted #29105 for this! Thanks for taking a look.

kevingessner:
bump -- issue is in the backlog of feature requests

kevingessner:
(bump)

hugohaggmark:
We've looked at this and having it to default to fetch all annotations is not a great experience for users that have a lot of annotations. There should be an explicit choice in the UI somehow, so users know that making this choice will cause fetching all annotations.

torkelo:
Closing this as I think fetching all annotations without any filter feels like a bad idea

PR #29034, opened 2020-11-11
Grafana UI: Card API refactor

### LINKED ISSUE #28591: Grafana UI Card Component
We did the first pass on a new unified card component that looks very promising. https://github.com/grafana/grafana/pull/28216

But it has a few issues that I think we should explore before start using it more. 

A) Big API

```typescript
export interface Props extends ContainerProps {
  /** Main heading for the Card **/
  heading: ReactNode;
  /** Additional data about the card. If array is supplied, elements will be rendered with vertical line separator */
  metadata?: ReactNode | ReactNode[];
  /** Card description text */
  description?: string;
  /** List of tags to display in the card */
  tags?: string[];
  /** Optional callback for tag onclick event */
  onTagClick?: OnTagClick;
  /** Indicates if the card and all its actions can be interacted with */
  disabled?: boolean;
  /** Image or icon to be displayed on the let side of the card */
  image?: ReactNode;
  /** Main card actions **/
  actions?: ReactElement[];
  /** Right-side actions */
  secondaryActions?: ReactElement[];
  /** Link to redirect to on card click. If provided, the Card inner content will be rendered inside `a` */
  href?: string;
  /** On click handler for the Card */
  onClick?: () => void;
}
```

One thing that stands out to me here is onTagClick feels strange, makes me think tags should be an element like the actions?  

It's kind of messy to use with props that take in element arrays:
```react
 <Card
        heading={<a href={ruleUrl}>{renderText(rule.name)}</a>}
        image={
          <Icon size="xl" name={rule.stateIcon as IconName} className={`alert-rule-item__icon ${rule.stateClass}`} />
        }
        metadata={[
          <span key="state">
            <span key="text" className={`${rule.stateClass}`}>
              {renderText(rule.stateText)}{' '}
            </span>
            for {rule.stateAge}
          </span>,
          rule.info ? renderText(rule.info) : null,
        ]}
        actions={[
          <Button variant="secondary" icon={rule.state === 'paused' ? 'play' : 'pause'} onClick={onTogglePause}>
            {rule.state === 'paused' ? 'Resume' : 'Pause'}
          </Button>,
          <LinkButton variant="secondary" href={ruleUrl} icon="cog">
            Edit alert
          </LinkButton>,
        ]}
      />
````

Originally I thought this design was necessary to fulfill requirements (I made a variant of the card last year that had same API as above). But we are getting more experience using child elements as part of the API instead (in new graph for example). 

Which would open the door for an API like this: 

```jsx
<Card header="Hello">
  <Card.Figure>
    ....
  </Card.Figure>
  <Card.Tags>
   ...
  </Card.Tags>
  <Card.Actions>
    ...
  </Card.Actions>
<Card>
```

The component would have to look for those specific child elements and render them in the correct part of the tree (not use children as you usually do). 

This would make usage of the component read better and clean up API a bit, it would hurt discoverability but if we had good examples in the storybook that should help with that. 

B) Improve responsive layout

Think the new design works great for smaller screens but for wider screens having the actions below makes for really poor use of space. 
Screenshot from 2020-10-27 12-57-31

Clarity-89:
Refactor Card component api to make it simpler and delegate more responsibility to child components. 

Fixes https://github.com/grafana/grafana/issues/28591


Clarity-89:
@torkelo I played around with refactoring Card api and while it does look simpler now, the logic for retrieving specific child components seems a bit hacky as there's no easy way in React to get specific child component it seems. I wonder if you had smth else in mind?

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -243,3 +234,68 @@ export const getCardStyles = stylesFactory((theme: GrafanaTheme, disabled = fals
...
+};
+
+Meta.displayName = 'Meta';
+
+interface ActionsProps extends ChildProps {
+  children: JSX.Element[];
+}
+const Actions: FC<ActionsProps> = ({ children, styles, disabled }) => {
+  return (
+    <div className={styles?.actions}>{React.Children.map(children, child => cloneElement(child, { disabled }))}</div>
```
torkelo, line 281: why React.Children needed here? why do we need to pass this disabled property it does not look like it's a property that is calculated internally but passed from the outside

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -243,3 +234,68 @@ export const getCardStyles = stylesFactory((theme: GrafanaTheme, disabled = fals
...
+            .filter(Boolean)
+            .reduce((prev, curr, i) => [
+              prev,
+              <span key={`separator_${i}`} className={styles?.separator}>
+                |
+              </span>,
+              curr,
+            ])
+        : children,
+    [children, styles?.separator]
```
torkelo, line 269: the readability of this is low for me (using ? operator that stretches many lines), wrapped in inline function that stretches many lines (without curly braces),  Why the filter(Boolean) ? why the reduce? 

The memoization also seems unnecessary and could be put on the component instead to make it a pure component.

torkelo (commented):
Yes this was what I had in mind, yea finding each child is a bit hacky. But from the consumer, it's really nice. 

getting rid of tags on click is nice and overall I tihink it's an improvement, a bit more complex internally. Close. Would love others thoughts on the Pros & Cons of this approach :)

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: We disable actions if the card has `disabled` prop and that needs to be passed to each action.

torkelo: but the actions are rendered outside so the user can set that. Guess it does not hurt (if the component is not a button that takes in this property)

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: Seems like filter isn't necessary indeed. Reduce is needed to join multiple metadata elements by a horizontal line. Agree that memoization could be moved onto component itself.

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: The idea is that the Card itself handles disabled logic. If there actions, they get disabled, otherwise the Card itself is disabled. Delegating part of this logic to the actions will makes things more complicated. Plus there are secondary actions that also get this prop.

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -243,3 +234,64 @@ export const getCardStyles = stylesFactory((theme: GrafanaTheme, disabled = fals
...
+}
+const Actions: FC<ActionsProps> = ({ children, styles, disabled }) => {
+  return (
+    <div className={styles?.actions}>{React.Children.map(children, child => cloneElement(child, { disabled }))}</div>
+  );
+};
+
+Actions.displayName = 'Actions';
+
+const SecondaryActions: FC<ActionsProps> = ({ children, styles, disabled }) => {
```
Clarity-89, line 283: Seems like the only difference between this and `Actions` is the styles, so probably better idea would be to make `Actions` component accept `variant` prop that would toggle primary or secondary actions styles.

torkelo: :+1:

packages/grafana-ui/src/components/Card/Card.tsx
torkelo: makes sense, just found it strange passing a prop from outside to the children when the consumer owns both elements & state & properties. A bit hidden logic that the Card component clones the element you gave it and sets this new property that the consumer did not set. But I can see that it's nice as well

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -49,95 +47,82 @@ const CardInner: FC<CardInnerProps> = ({ children, href }) => {
...
-      className={cx(styles.container, className)}
+      className={cx(containerStyles, className)}
       onClick={onCardClick}
       {...htmlProps}
     >
       <CardInner href={href}>
-        {image && <div className={styles.media}>{image}</div>}
+        {figure}
         <div className={styles.inner}>
           <div className={styles.heading}>{heading}</div>
```
kaydelaney, line 118: Might be a good idea to add the "heading" role here
```suggestion
          <div className={styles.heading} role="heading">
            {heading}
          </div>
```

Clarity-89: Great catch!

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -49,95 +47,84 @@ const CardInner: FC<CardInnerProps> = ({ children, href }) => {
...
-              |
-            </span>,
-            curr,
-          ])
-        : metadata,
-    [metadata, styles.separator]
+  const styles = getCardStyles(theme);
+  const [tags, figure, meta, actions, secondaryActions] = ['Tags', 'Figure', 'Meta', 'Actions', 'SecondaryActions'].map(
+    item => {
+      //@ts-ignore
```
dprokop, line 88: I guess this piece of code can help you getting rid of the ts-ignore here: https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/uPlot/geometries/SeriesGeometry.tsx#L57

packages/grafana-ui/src/components/Card/Card.tsx
```diff
@@ -49,95 +47,84 @@ const CardInner: FC<CardInnerProps> = ({ children, href }) => {
...
+      const found = React.Children.toArray(children).find((child: JSX.Element) => {
+        return child?.type?.displayName === item;
+      }) as JSX.Element | undefined;
+
+      if (found) {
+        return React.cloneElement(found, { ...found.props, styles, disabled });
```
dprokop, line 94: I like this approach with disabled and styles being handled by the card. I see one thing that might be confusing from the POV of the consumer:
```
<Card>
  <Card.Actions disabled .../>
</Card>
```

This will not have any effect, doesn't it, as long as the disabled prop is not defined on the Card component?

dprokop (approved):
The updated API looks really good, I like this approach a lot from readability POV. Great work!

packages/grafana-ui/src/components/Card/Card.tsx
Clarity-89: Good point, and since `Actions` does expect `disabled` prop, this won't raise any TS issues. Maybe it would be a good idea not override child props when cloning, in case consumers want to pass props to the children directly 🤔

PR #29040, opened 2020-11-11
DashboardLinks: will only refresh dashboard search when changing tags for link.

### LINKED ISSUE #28869: Dashboard links loaded thousands of times during template variable load resulting 10x slowdown in dashboard load
Our complex dashboard with ~50 of template variables loads in Grafana 6.4.1 in 7s based on network analyzer.
In Grafana 7.3.1 loading takes 55s. 
Dashboard load result ~300 Prometheus queries (some variables queried tens of times for unknown, I was expecting <100 queries)

**Current behavior:**
We are using Dashboard links to provide links to few other Dashboards.

In 6.41 Dashboards links were loaded once when dashboard was loaded.

In Grafana 7.3.1 during the dashboard load or during time range change the template variable update (we reload dashboard variable in time range change) Grafana seems to load Dashboards links 2x1047 times. Prometheus queried are stalled while Dashboard links are loaded.

Looks that for Dashboards Links are loaded  ~100 times, then few Prometheus template variable queries, then again 100 Dashboard Link queries, then few Prometheus queries.
Each Link download takes up to 100ms and maybe 20 parallel request are done. Thus 100 request will take up to 1s and small batch or real Prometheus queries are done once per second.

Removing another Dashboard link tag reduce the calls to half.

image

image

**Expected behavior:**
Dashboard links loaded only once.

**Repro step:**
Create multiple Dashboards with large number of template variable and link Dashboards together using few tags.

**Environment**:
- Grafana version: 7.3.1, repro in all 7.x versions
- OS Grafana is installed on: Windows 10 and Linux
- User OS & Browser: Chrome
- Grafana plugins:
- Others:

mckn:
The linked issue described that we are doing multiple dashboard searches when having a dashboard with nested template variables and then changing the variable values.

This will prevent that from happening but we will still re-run the link templating when changing variables.

Fixes #28869

I cleaned this code up a little bit but there a couple of things that can be improved further. Maybe something we can do in upcoming PRs.

### DIFF
--- a/e2e/suite1/specs/templating-dashboard-links-and-variables.ts
+++ b/e2e/suite1/specs/templating-dashboard-links-and-variables.ts
@@ -59,8 +59,7 @@ e2e.scenario({
     e2e.components.DashboardLinks.dropDown()
       .should('be.visible')
       .click()
-      .wait('@tagsTemplatingSearch')
-      .wait('@tagsDemoSearch');
+      .wait('@tagsTemplatingSearch');
 
     // verify all links, should have p2 value
     verifyLinks('p2');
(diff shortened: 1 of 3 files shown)


torkelo (approved):
Looks good

PR #29044, opened 2020-11-11
GraphNG: Adding ticks test dashboard and improves tick spacing

torkelo:
Created a copy of a test dashboard we have for the angular graph that tests some tricky y-axis tick formatting.

Angular
Screenshot from 2020-11-11 21-09-42

uPlot (many of the differences are not due to uPlot but due to the display processor which is used for y-axis ticks and is not used in the flot graph) 
Screenshot from 2020-11-11 21-09-55

Can't get uPlot to create a tick for 2500 or 7500 for some reason, but not critical. 

Still don't know how I can fix the problem of x-axis time ticks that go outside graph on the right edge

### DIFF
--- a/devenv/dev-dashboards/panel-graph/graph-ng-yaxis-ticks.json
+++ b/devenv/dev-dashboards/panel-graph/graph-ng-yaxis-ticks.json
@@ -1,0 +1,752 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "links": [],
+  "panels": [
+    {
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axis": {
+              "grid": true,
+              "label": "",
+              "side": 3,
+              "width": 60
+            },
+            "bars": {
+              "show": false
+            },
+            "fill": {
+              "alpha": 0
+            },
+            "line": {
+              "show": true,
+              "width": 1
+            },
+            "nullValues": "null",
+            "points": {
+              "radius": 4,
+              "show": false
+            }
+          },
+          "links": [],
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          },
+          "unit": "short"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 7,
+        "w": 8,
+        "x": 0,
+        "y": 0
+      },
+      "id": 7,
+      "options": {
+        "legend": {
+          "asTable": false,
+          "isVisible": true,
+          "placement": "bottom"
+        },
+        "tooltipOptions": {
+          "mode": "single"
+        }
+      },
+      "pluginVersion": "7.4.0-pre",
+      "targets": [
+        {
+          "refId": "A",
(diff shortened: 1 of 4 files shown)


leeoniya:
> Can't get uPlot to create a tick for 2500 or 7500 for some reason, but not critical.

the top plots have less height than the bottom plots, so the ticks end up being too close. try reducing the y-axis `axis.space`

leeoniya:
> Still don't know how I can fix the problem of x-axis time ticks that go outside graph on the right edge

i can probably help better if you describe how you expect a fix to behave. if you simply want to omit the tick label from rendering at the edge, you can do this by using `axis.filter`:

https://github.com/leeoniya/uPlot/blob/01087cb4487e2bc7e4d544da906bd0344aa1156b/dist/uPlot.d.ts#L494-L495

in the screenshots above, uPlot has visibly less x gutter allotted to it (on the right) than the Flot variant, and definitely looks too tight.

torkelo:
> i can probably help better if you describe how you expect a fix to behave. if you simply want to omit the tick label from rendering at the edge, you can do this by using axis.filter:

I think uplot needs to add to the right gutter if the space for a label extend the inner graph dimensions

torkelo:
> in the screenshots above, uPlot has visibly less x gutter allotted to it (on the right) than the Flot variant, and definitely looks too tight.

Yes this is due to not being able to target the right gutter without affecting the left gutter

leeoniya:
> Yes this is due to not being able to target the right gutter without affecting the left gutter

as i mentioned in https://github.com/leeoniya/uPlot/issues/362#issuecomment-723557200, i'm pretty sure that gutters only get added to sides without any axes.

> I think uplot needs to dynamically add to the right gutter if the space for a label extends the inner graph dimensions. uPlot has the info where it exactly it renders the labels and their space so should be able to do this.

this is pretty much asking for the same thing as axis auto-sizing, so i'll expose the same functionality for gutters, which will allow you to measure and adjust the gutters. but the same caveat applies: modifying the available space for the plotting area will re-evaluate the tick density and may end up changing the ticks. it'll then be up to you to decide how many re-measuring/resize rounds you want to do when this happens. in general i think it'll work okay without too much perf impact.

torkelo:
> as i mentioned in leeoniya/uPlot#362 (comment), i'm pretty sure that gutters only get added to sides without any axes.

ah, did not know this, though it added to both sides always. 

> this is pretty much asking for the same thing as axis auto-sizing, so I'll expose the same functionality for gutters, which will allow you to measure and adjust the gutters.

Ok sounds good. Will need the x-axis ticks & position in that callback so I know how much gutter to add in case the rightmost x-axis tick is on the edge. Or if the gutter function is called after axis size is measured I can take that and divide it by /2 to get a minimum gutter (assuming labels are at most centered on the right edge) 

> but the same caveat applies: modifying the available space for the plotting area will re-evaluate the tick density and may end up changing the ticks.

Why would this trigger new re-renders? I assume space & axis size functions are called before rendering? 

Now to the problem of tick and spacing:

I mentioned I cannot get uPlot to make ticks for  2500 or 7500. It's not a matter of lowering the y-axis space. I can lower to a point where I get tick labels every 2000 but that is too many ticks. Seems to be some logic that prevents 2500 and 7500 steps from being considered. 

Screenshot from 2020-11-12 13-30-50

Lower space bit by bit I get this eventually (but no step in between)
Screenshot from 2020-11-12 13-33-55

leeoniya:
> Why would this trigger new re-renders? I assume space & axis size functions are called before rendering?

i guess i should say rounds-of-remeasuring. the actual re-rendering will be short-circuited until the re-measuring settles.

> Now to the problem of tick and spacing:
>
> I mentioned I cannot get uPlot to make ticks for 2500 or 7500. It's not a matter of lowering the y-axis space. I can lower to a point where I get tick labels every 2000 but that is too many ticks. Seems to be some logic that prevents 2500 and 7500 steps from being considered.

i'll dig into this to see what's up.

leeoniya:
@torkelo 

after battling with float & numeric precision issues (yet again), i've pushed a couple commits that fix the 2500 case. there is some slight collateral behavior changes as a result of the `2.5` multiple is added globally to the default numeric (not time) `axis.incrs`. otherwise i think it works well. before/after:

test

torkelo:
@leeoniya that great! Thank you!

PR #29051, opened 2020-11-12
Graph: Fixes stacking issues like floating bars when data is not aligned 

### LINKED ISSUE #9294: [Bug] Graphs incorrectly stacked when data points are missing
- What Grafana version are you using? 4.4.1
- What datasource are you using? Prometheus
- What OS are you running grafana on? CentOS
- What did you do? Make a stacked graph of four time series
- What was the expected result? The lines stack on top of each other
- What happened instead? When one time series goes missing, the lines no longer stack properly

Query:

```
sum by (instance) (service_instance:stats_requests:irate1m{service="foo"})
```

Screenshot:

image

The red and green lines should be stacked, and their total height should remain the same as the other hosts take over the load.

We worked around it by checking "nulls as zeroes" but this still seems like a bug.

### LINKED ISSUE #27033: Floating stacked values in graph
**What happened**:
I get floating stacked values when a previous item is missing for that period

image

The graph entries for 08/08, 08/10, and 08/16 to start at zero. 

Select a dataset with values missing at midrange intervals, then use stacked bars to present data. 

I've tried null_as_zero, null, and connected on stack values. 
I've tried (un)hiding null or zero values. 
This emulates a line/area graph in "continuous" representation, which is inaccurate. 

**Environment**:
- Grafana version: 7.0.3
- Data source type & version: MariaDB 10.4
- OS Grafana is installed on: Ubuntu 18.04.5 LTS 
- User OS & Browser: Mac OS 10.15.6 (19G73) on Chrome 84.0.4147.125
- Grafana plugins: 
- Others:

torkelo:
Fixes #27033
Fixes #9294

Many are encountering strange stacking issues for years. This removes the interpolation logic that existed for data points that does not align (or missing).

Before:
Screenshot from 2020-11-12 09-21-50

After:
Screenshot from 2020-11-12 09-21-17

Might be some scenarios that now get worse when data points are close but not perfectly aligned, they will now not stack together

### DIFF
--- a/public/app/plugins/panel/graph/graph_tooltip.ts
+++ b/public/app/plugins/panel/graph/graph_tooltip.ts
@@ -102,17 +102,11 @@ export default function GraphTooltip(this: any, elem: any, dashboard: any, scope
         minTime = pointTime;
       }
 
-      if (series.stack) {
-        if (panel.tooltip.value_type === 'individual') {
-          value = series.data[hoverIndex][1];
-        } else if (!series.stack) {
-          value = series.data[hoverIndex][1];
-        } else {
-          lastValue += series.data[hoverIndex][1];
-          value = lastValue;
-        }
-      } else {
-        value = series.data[hoverIndex][1];
+      value = series.data[hoverIndex][1];
+
+      if (series.stack && value !== null && panel.tooltip.value_type !== 'individual') {
+        lastValue += value;
+        value = lastValue;
       }
 
       // Highlighting multiple Points depending on the plot type
--- a/public/vendor/flot/jquery.flot.stack.js
+++ b/public/vendor/flot/jquery.flot.stack.js
@@ -134,7 +134,7 @@ charts or filled areas).
                         // we got past point below, might need to
                         // insert interpolated extra point
                         if (i > 0 && points[i - ps] != null) {
-                            intery = py + (points[i - ps + accumulateOffset] - py) * (qx - px) / (points[i - ps + keyOffset] - px);
+                            intery = 0;
                             newpoints.push(qx);
                             newpoints.push(intery + qy);
                             for (m = 2; m < ps; ++m)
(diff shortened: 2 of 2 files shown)


dprokop (approved):

davkal:
+1 on the simplicity of the solution. I wonder if it makes sense to adjust the bar width to prevent overlap. But perhaps that's not practically since buckets can be infinitesimally close to each other.

alexanderzobnin:
This fix causes issues with Zabbix data source, because Zabbix API returns points "as is" and timestamp of the point is the exact time where it was written in the DB. So here's an example of the same metric for 4 different hosts, metric collection interval is 10 seconds:

Screenshot from 2020-11-17 12-33-02

Screenshot from 2020-11-17 12-32-55

So as you can see, data is not aligned and removing interpolation makes `Stack` mode not working at all. Plugin has a feature that allows to align data (`groupBy()` data processing function), but the main problems are that this changes original data and interval should match metric collection interval, wich can be different for different metrics, so it's not possible to apply it automatically. Here's an example of applying `groupBy(10s, avg)`:

Screenshot from 2020-11-17 12-33-17

torkelo:
group by looks like a great solution and the correct solution for this

alexanderzobnin:
Yes, but it's not possible to use it right now. I can think about adding default alignment, but it needs some time to add this to the plugin. So probably I can add default alignment first in the plugin and then we can remove interpolation.

alexanderzobnin:
Added default points alignment to the Zabbix data source in this PR https://github.com/alexanderzobnin/grafana-zabbix/pull/1099, so with that changes stacking graphs are working correctly:

Screenshot from 2020-11-20 14-21-28

Screenshot from 2020-11-20 14-22-25

CC @torkelo

torkelo:
great @alexanderzobnin , I think this data group by time logic is better done in data sources or in a Grafana transform rather than a flaky interpolation in the graph stacking code. But not sure, hope this wont introduce a ton of issues for people & existing dashboards

alexanderzobnin:
I think aligned data looks better, and also, another benefit from this is ability to fill gaps with `null` values (Zabbix doesn't put nulls in the DB if there's no value collected). The only concern is that some users can be confused if they want to see original data, but I can make it possible to disable alignment in the query editor.

alexanderzobnin:
Also, I'm making conversion to the Data Frames, so alignment is super useful, because I can return multiple series in the same frame.

PrestonTaylor:
Is this bug the same? It seems like it's stacking on the last non-null value in the time series which is incorrect. 
image

PR #29057, opened 2020-11-12
React: Upgrading react to v17, wip

torkelo:
Getting a lot of strange typing errors, have not time to continue. If anyone wants to help investigate these type issues (rc-slider)


package.json
```diff
@@ -257,8 +257,9 @@
     "prop-types": "15.7.2",
     "rc-cascader": "1.0.1",
     "re-resizable": "^6.2.0",
-    "react": "16.12.0",
-    "react-dom": "16.12.0",
+    "react-beautiful-dnd": "13.0.0",
```
torkelo, line 260: was added as got new errors that this could not be resolved

packages/grafana-ui/package.json
```diff
@@ -53,7 +53,7 @@
     "papaparse": "5.3.0",
     "rc-cascader": "1.0.1",
     "rc-drawer": "3.1.3",
-    "rc-slider": "9.3.1",
+    "rc-slider": "9.6.2",
```
torkelo, line 56: Tried updating this to resolve the rc-slider type errors but did not help

package.json
mckn: hmm, I think that one is hoisted from the `grafana/ui` package. Since the Transformations UI was located in the `grafana/ui` in the beginning the `react-beautiful-dnd` was added as a dependency to `grafana/data`. But looks like we did not move it to the root package when moving the transformations UI.

So we can remove it completely from `grafana/ui` and only have it in the root project.

packages/grafana-ui/package.json
```diff
@@ -53,14 +53,14 @@
...
     "rc-time-picker": "^3.7.3",
-    "react": "16.12.0",
+    "react": "17.0.1",
     "react-beautiful-dnd": "13.0.0",
```
mckn, line 59:
```suggestion
```

mckn:
I'm getting this when building storybook but that might be something that will be resolved with the update to 6.1?

Skärmavbild 2020-11-18 kl  07 46 00

mckn (approved):
LGTM!

I think we can remove the `react-beautiful-dnd` in a separate PR for better tracking.

PR #29061, opened 2020-11-12
Update react-testing-library

Clarity-89:

There are a few front end tests randomly failing when trying to query for an element after an async operation. It seems like the issue happens in the long running tests. Upgrading to the latest version of RTL fixes the issue, at least locally.


package.json
```diff
@@ -76,8 +76,8 @@
     "@grafana/api-extractor": "7.10.1",
     "@grafana/eslint-config": "2.0.5",
     "@rtsao/plugin-proposal-class-properties": "7.0.1-patch.1",
-    "@testing-library/jest-dom": "^5.11.3",
-    "@testing-library/react": "^10.4.8",
+    "@testing-library/jest-dom": "^5.11.5",
+    "@testing-library/react": "^11.1.2",
```
dprokop, line 80:
```suggestion
    "@testing-library/jest-dom": "5.11.5",
    "@testing-library/react": "11.1.2",
```

dprokop (approved):

Issue #29017, opened 2020-11-11
CloudWatch access with sts:AssumeRoleWithWebIdentity broken in 7.3.x


uncycler:
**What happened**:

On AWS EKS 1.18, Grafana 7.3.1, deployed with the helm chart, cannot perform sts:AssumeRoleWithWebIdentity authentication anymore using IRSA. Query fails with this error:
failed to call cloudwatch:ListMetrics: WebIdentityErr: failed to retrieve credentials\ncaused by: AccessDenied: Not authorized to perform sts:AssumeRoleWithWebIdentity\n\tstatus code: 403

Cloudwatch datasource provider is set to "AWS SDK Default".

When I change the image back to 7.2.2, using the same chart version, it works as expected.

**What you expected to happen**:

Authenticate succesfully to AWS using IRSA token.

**How to reproduce it (as minimally and precisely as possible)**:

On a EKS 1.18 cluster:
* Create IAM role along with policy that give it access to cloudwatch.
* Deploy Grafana helm chart and set the serviceAccount annotation "eks.amazonaws.com/role-arn" to the IAM role ARN.
* Create a Cloudwatch datasource and test.

**Environment**:
- Grafana version: 7.3.x
- Data source type & version: CloudWatch
- OS Grafana is installed on: Docker
- User OS & Browser: Ubuntu/Chrome

sunker:
@aknuds1 can this be related to the #25548?

Legion2:
We are using AWS EKS IAM Roles for Service Accounts with Grafana 7.3.3. I think the problem here is that you created a new Service Account with a different name and namespace and attached the annotation to it. This does not work, the role can only be attached to a service account with a specific name.

We also deployed grafana with helm and set the name of the service account, after that we created the IAM role with eksctl for that service account.

sunker:
@uncycler would you be able to do what @Legion2 is suggesting and see if the problem still exist?

uncycler:
Sorry for the late reply! @Legion2, right, we had a typo in the trust relationships condition.

Weird that it worked with 7.2.x.

We can close this issue!

PR #29065, opened 2020-11-12
Auth: Enable more complete credential chain for SigV4 default SDK auth option

### LINKED ISSUE #28931: Sigv4 returns 403 when running from EKS with IAM role for service Accounts
We are running Grafana from an AWS EKS cluster (v1.18) and connecting to AWS Elastic Search service version 7.7.

We are using Service accounts and IAM roles in our pod (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)

At the moment we are successfully connecting to ElasticSearch with aws-sigv4-proxy as a sidecar container using the proxy URL as URL:

image

When we change to use the ElasticSearch endpoint and enable Sigv4:

image

We get 403:

image

Not relevant error other than request completed with 403 response:

`info msg="Request Completed" logger=context userId=39 orgId=1 uname=----- method=GET path=------- status=403 remote_addr=------- time_ms=3175 size=141 referer=https://grafana.------`

**What happened**:

Can't use Sigv4 in our Elastic Search Data Source when running Grafana from Eks with IAM roles for Service Accounts

Able to use new Sigv4 feature and remove our aws-sigv4-proxy

Add Grafana to an EKS cluster with a service account and IAM role and try to connect to ES using Sigv4.

For clarity: and in case it helps: https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

wbrowne:

Fixes #28931

**Special notes:** 
`defaults.CredChain` doesn't include the IAM roles for Service Accounts credential provider but the default cred chain in `session.NewSession()` does. See here: https://github.com/aws/aws-sdk-go/blob/d6d63adc1c143626733f1ab3d575b6f777357d04/aws/session/credentials.go#L18

### DIFF
--- a/pkg/models/sigv4.go
+++ b/pkg/models/sigv4.go
@@ -107,6 +107,11 @@ func (m *SigV4Middleware) signer() (*v4.Signer, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		if m.Config.AssumeRoleARN != "" {
+			return v4.NewSigner(stscreds.NewCredentials(s, m.Config.AssumeRoleARN)), nil
+		}
+
 		return v4.NewSigner(s.Config.Credentials), nil
 	}
 
--- a/pkg/models/sigv4.go
+++ b/pkg/models/sigv4.go
@@ -11,7 +11,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
-	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/session"
 	v4 "github.com/aws/aws-sdk-go/aws/signer/v4"
 	"github.com/aws/aws-sdk-go/private/protocol/rest"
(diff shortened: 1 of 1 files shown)


marefr (approved):
LGTM 

Is this something needed for cloudwatch as well?

papagian (approved):
LGTM!

pkg/models/sigv4.go
```diff
@@ -90,9 +89,28 @@ func (m *SigV4Middleware) signRequest(req *http.Request) (http.Header, error) {
...
+		c = credentials.NewStaticCredentials(m.Config.AccessKey, m.Config.SecretKey, "")
+	case Credentials:
+		c = credentials.NewSharedCredentials("", m.Config.Profile)
+	}
+
+	// passing nil credentials will force AWS to allow more complete credential chain vs the explicit default
+	if c == nil {
+		s, err := session.NewSession(&aws.Config{
+			Region:      aws.String(m.Config.Region),
+			Credentials: nil},
```
aknuds1, line 108: There's no need to provide the default value.
```suggestion
		},
```

aknuds1 (approved):
Looks good AFAICT, with some nits.

wbrowne:
@marefr SigV4 and CloudWatch should behave consistently now with this change (so CloudWatch was always working as expected AFAICT).  I've tested it out locally with the default SDK option and  I'm able to reach [here](https://github.com/aws/aws-sdk-go/blob/d6d63adc1c143626733f1ab3d575b6f777357d04/aws/session/credentials.go#L18), which will look for the extended default options like the one mentioned in #29017. I think we would need to investigate further and try it out on the version specified and compare with 7.3.X since the credentials chain might have changed between versions.

marefr (approved):
LGTM

PR #29067, opened 2020-11-12
Legend/GraphNG: Refactoring legend types and options

torkelo:
Legend placement is crashing chrome, need to figure that out :) 

Screen Shot 2020-11-12 at 18 41 30


dprokop:
Wow, the Chrome crash is very weird. No output in the console and basically the entire tab gets broken:

image

Is this the same you are experiencing @torkelo ?

Issue #15696, opened 2019-02-28
[Feature Request] Support mixed type plugin for backend


Gauravshah:
Trying to port https://github.com/GoshPosh/grafana-meta-queries to backend so that alerting is supported.

Mixed plugins behave differently as in they require access to all queries in the dashboard and respective datasources.

Mixed data source gets all the queries & datasource of a given dashboard:
https://github.com/grafana/grafana/commit/b30dfcf28a92d770e02c007dc67b309acb287cef
and is used like : https://github.com/GoshPosh/grafana-meta-queries/blob/master/src/datasource.js#L43

But the same thing is not achievable on backend plugins.
Backend plugins get call to 
```
func (t *DruidDatasource) Query(ctx context.Context, tsdbReq *datasource.DatasourceRequest) (*datasource.DatasourceResponse, error) {
```

which has access to current query only. Tried digging through the code 
https://github.com/grafana/grafana/blob/master/pkg/plugins/datasource/wrapper/datasource_plugin_wrapper.go#L61
sends only the current query and datasource

Digging a little further
the single query is saved within the alert table of grafana database and hence has no information to the dashboard.

Unsure how to proceed

Included information:
### What Grafana version are you using?
grafana 5 & 6
### What datasource are you using?
https://github.com/GoshPosh/grafana-meta-queries
with druid
### What OS are you running grafana on?
mac & linux
 
relates to  https://github.com/grafana/grafana/issues/6841

Gauravshah:
if there is a planned direction we can try creating contributing to a PR. Looking at current code, from our understanding, it requires considerable changes.  
I am not sure if we should make dashboardCache available to the plugin and also unsure if we should store entire dashboard within the alert.
The other issues is getting all the dataSources and issue queries to them from the mixed plugin

bergquist:
I'm not sure this is something we want to support in the backend plugins. I might have to think about this a little bit more...

Gauravshah:
I can share more details about the use case.
Meta-queries-plugin has been super helpful in dealing computations not only with single datasource but across multiple.  One thing that is missing from our existing system is to be able to support alerts on computed metrics for eg when shares per active user goes down is an alert for business health.
Also if the data is across multiple dashboards, then things like if our spends increased by 50% from previous week did our orders increase in a similar fashion.

Gauravshah:
@bergquist or would it make more sense to move this plugin inside grafana source. I am aware that we are trying to move plugins away into its own repo but this one being a special case.

[...]

PR #29072, opened 2020-11-12
Expressions: Move GEL into core as expressions

### LINKED ISSUE #28993: Move Server-Side Expressions (GEL) into Grafana
Take https://github.com/grafana/gel-app/, and move into Grafana, renaming to expressions. (The frontend piece of this is already in core grafana).

Because:
 - Flow with alerting makes more sense (instead of stuff flowing from grafana, to expression plugin, back to gel)
 - Should be easier to iterate on if in core, in particular with any context we want to carry through to alerting. 
 - Don't have to worry about signing + embed plugin for it (other than cleanup)

TODO:
 - [x] Decide where the code lives and rename from gel to expressions 
 - [x] Replace Transform endpoints with calls to expressions

Cleanup TODO:
 - [ ] Remove embedded plugin 
 - [ ] Cleanup gel-app repo
 - [ ] cleanup builds for embedded plugin

Note: Not sure if there is anything I need to think about when cleaning up old GEL that might break backport / older builds which may delay changing the buildings are archiving the repo.

kylebrandt:
Take https://github.com/grafana/gel-app/, and move into Grafana, renaming to expressions. (The frontend piece of this is already in core grafana).

Fixes #28993

 - [x] ~Still needs plugins-bundled figured out~ (Update: removed the entry from external.json)
 - The transform plugin pieces have been removed (removing from SDK would be follow up I think https://github.com/grafana/grafana-plugin-sdk-go/issues/237)
 - This should still be behind the feature flag (so review request is really for all the stuff outside expr)
 - If you have the gel plugin in your plugins dir, Grafana will not start, so maybe we should add removal of the Transform endpoint just in case someone discovered it.
 - Package structure: I do not like the way things are falling under `expr/`. To much stuff is going to end up in mathexpr (e.g. the types for series etc, resample, reduce). This can be for a follow up PR. The correct layout is not apparent to me at the moment. Although Node being an interface (set of methods) probably allows for some flexibility here.


papagian (commented):
I found some references to the `transform` plugin type that I think they are obsolete:

https://github.com/grafana/grafana/blob/1024e652e85a61f8de2de4a2c5f028a36f02ba2b/pkg/plugins/plugins.go#L358-L366

https://github.com/grafana/grafana/blob/1024e652e85a61f8de2de4a2c5f028a36f02ba2b/pkg/plugins/plugins.go#L376-L379

Note also that if `ngalert` is enabled but `expressions` are disabled then one can still evaluate an alert definition using the API.
If we decide that we don't want this we can disabling `ngalert` if `expressions` are disabled [here](https://github.com/grafana/grafana/blob/43f580c299d7ed237ec79c19a205139fbff9c066/pkg/services/ngalert/ngalert.go#L44).

aknuds1 (approved):
That's a whole lotta code! Just skimmed it since most of it is just moving code into pkg/expr.

marefr (approved):
LGTM great :+1: 

Moving forward when expression are a core thing maybe could make sense to remove the hack with data source id -100. I don't have all the details and it's maybe hard, but worth to consider. I guess we had this earlier to be able to send the data GEL needed in the data query request. Now you can change the Grafana HTTP API request model to whatever GEL needs, Maybe having a specific HTTP API for expressions would make sense as well in case we would like request metrics specifically for that endpoint and so on.

papagian (approved):

Gauravshah:
this removes the ability of external plugins to be able to do similar things.  we are looking to add backend support for https://github.com/GoshPosh/grafana-meta-queries.
Had raised an issue as well earlier
https://github.com/grafana/grafana/issues/15696

PR #29077, opened 2020-11-12
StatPanel: Fixes hanging issue when all values are zero

### LINKED ISSUE #29073: Stat panel type locks up browser when constant 0 value is returned in grafana 7.3.2
**What happened**:
Upgraded to 7.3.2 from 7.3.1 and a few dashboards started to lock up the browsers.

The easiest way is to setup a prometheus datasource (doesn't even need to work)
create a new graph panel
to the query just simply put in 0
change the graph type to Stat

**Environment**:
- Grafana version: 7.3.2
- Data source type & version: prometheus (but I suspect this is datasource independent as the original issue was with elasticsearch)
- OS Grafana is installed on: debian, reproduce on ubuntu/docker
- User OS & Browser: Windows10, Firefox and chrome behaves the same way
- Grafana plugins: none
- Others:

0-stat-broken

torkelo:
Fixes #29073

Apperently Number.MIN_VALUE is not what I thought. Stupid rookie mistake :(

This PR https://github.com/grafana/grafana/pull/28982/ that fixed min max calculations when one series had zero made it so that max became Number.MIN_VALUE when all values are zero as zero is actually smaller than Number.MIN_VALUE. This max value made bizcharts change the browser.

This changes the logic so min & max remain null if all values are null.

### DIFF
--- a/packages/grafana-data/src/field/fieldOverrides.test.ts
+++ b/packages/grafana-data/src/field/fieldOverrides.test.ts
@@ -88,6 +88,17 @@ describe('Global MinMax', () => {
     expect(minmax.max).toEqual(1234);
   });
 
+  it('find global min max when all values are zero', () => {
+    const f0 = new ArrayDataFrame<{ title: string; value: number; value2: number | null }>([
+      { title: 'AAA', value: 0, value2: 0 },
+      { title: 'CCC', value: 0, value2: 0 },
+    ]);
+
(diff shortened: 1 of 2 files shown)


ryantxu (approved):
:rofl:  I saw the Number.MIN and wondered if that was really what you were after!

dprokop (approved):

PR #29084, opened 2020-11-13
Provisioning: Fixes application not pinned to the sidebar when it's enabled

### LINKED ISSUE #29058: Provisioning: application is not pinned to the sidebar
**What happened**:
When using provisioning with an application that has already being installed, but was disabled manually, it's going to be enabled, but not pinned to the sidebar.

Application is enabled and pinned to the sidebar.

1. Install application.
1. Go to the app config page in Grafana.
1. Enable it. App should be pinned to the sidebar (if it's a feature of the app).
1. Disable app. 
1. Add app provisioning config, like:
    ```yaml
    apps:
      - type: grafana-synthetic-monitoring-app
        disabled: false
    ```
1. Restart grafana server.
1. Reload app config page. App is enabled, but not pinned to the sidebar.

This happens because of this line:
https://github.com/grafana/grafana/blob/a1e80af800f04d70d28deb9793d73ce55796b7f8/pkg/services/provisioning/plugins/plugin_provisioner.go#L50
If plugin config found in Grafana DB, then `Pinned` property will be always given from that config. But if app is disabled, then `Pinned` is usually `false` (I can say it's false in 99% cases, since if we disable app, then we want to remove it from the sidebar as well, so we set `pinned` to `false` in the query).

I think there're 2 possible solutions:
1. Set `Pinned` to `true` if we enable app via provisioning (`disabled: false`).
1. Add `pinned` option to the provisioning conf (probably, with default set to `false` or using `disabled` style - name it `unpinned`):
    ```yaml
    apps:
      - type: grafana-synthetic-monitoring-app
        disabled: false
        pinned: true
    ```

**Environment**:
- Grafana version: latest master (7.4.0-pre)
- Data source type & version: Synthetic Monitoring app (or any other app that appears on the sidebar)
- OS Grafana is installed on: 
- User OS & Browser:
- Grafana plugins:
- Others:

alexanderzobnin:
This PR makes provisioned application pinned to the sidebar when it's enabled.

Fixes #29058

### DIFF
--- a/pkg/services/provisioning/plugins/plugin_provisioner.go
+++ b/pkg/services/provisioning/plugins/plugin_provisioner.go
@@ -47,7 +47,6 @@ func (ap *PluginProvisioner) apply(cfg *pluginsAsConfig) error {
 			}
 		} else {
 			app.PluginVersion = query.Result.PluginVersion
-			app.Pinned = query.Result.Pinned
 		}
 
 		ap.log.Info("Updating app from configuration ", "type", app.PluginID, "enabled", app.Enabled)


marefr (approved):
Tested it and works as expected :+1: 

LGTM

PR #29097, opened 2020-11-13
Backend: More renaming for standard style conformance

aknuds1:
More renaming of variables etc to conform with standard Go style.


kylebrandt (approved):
Looks nicer to me

PR #29098, opened 2020-11-13
Chore: Bumps types for jest

hugohaggmark:
Moving this out of #28929 to simplify the moving parts of that PR, will update #28929 when this is merged.

Fixes #

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6550,13 +6550,6 @@
   dependencies:
     "@types/istanbul-lib-report" "*"
 
-"@types/jest-diff@*":
-  version "24.3.0"
-  resolved "https://registry.yarnpkg.com/@types/jest-diff/-/jest-diff-24.3.0.tgz#29e237a3d954babfe6e23cc59b57ecd8ca8d858d"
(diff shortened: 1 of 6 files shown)


dprokop (approved):
LGTM

yarn.lock
```diff
@@ -9541,11 +9515,6 @@ caniuse-api@^3.0.0:
     lodash.memoize "^4.1.2"
     lodash.uniq "^4.5.0"
 
-caniuse-db@1.0.30000772:
```
zoltanbedi, line 9544: I still see this not removed when I run yarn. Can you add this back?

zoltanbedi (approved):

yarn.lock
hugohaggmark: You need to explain this more, because I'm just updating packages with yarn add command. What am I doing wrong?

zoltanbedi: I'm not sure what goes wrong here. When I checkout your branch and run yarn without the lock freeze then I see this added back to the lockfile.

hugohaggmark: But drone is not complaining about this, it would if the yarn.lock file needed to update because it runs `yarn install --frozen-lockfile --no-progress`.

PR #29101, opened 2020-11-13
Permissions: Validate against Team/User permission role update

### LINKED ISSUE #29059: Permissions: Inconsistent state inherited for User+Team ACLs after adding Editor/Viewer permissions
**What happened**:
It seems to me the react state isn't being reset correctly in `AddPermission.tsx` for when the type select form field changes - or more concretely, the **role** field is never being set to `undefined` when the form flips back to type `Team` after adding a permission. The backend intercepts the role value as normal (will follow up with this in a separate issue) and eventually ends up creating a weird state for the user (see #28615). 

`role` will be set to `null`/`undefined` when creating a Team permission so that the correct state is represented in the database.

Kapture 2020-11-12 at 13 02 20

**Environment**:
- Grafana version: master
- User OS & Browser: MacOS, Chrome v86

wbrowne:
It's my understanding that team/user permissions cannot have an associated role as it leads to weird permission leaks (see linked issue). This will fix the bad state in the frontend and also validate against the request on the backend in the case the state ever breaks on the client again.

Fixes #29059

### DIFF
--- a/pkg/api/dashboard_permission_test.go
+++ b/pkg/api/dashboard_permission_test.go
@@ -186,9 +186,9 @@ func TestDashboardPermissionAPIEndpoint(t *testing.T) {
 					"/api/dashboards/id/:id/permissions", cmd, func(sc *scenarioContext) {
 						callUpdateDashboardPermissions(sc)
 						assert.Equal(t, 400, sc.resp.Code)
-						respJSON, err := simplejson.NewJson(sc.resp.Body.Bytes())
+						respJSON, err := jsonMap(sc.resp.Body.Bytes())
 						require.NoError(t, err)
-						assert.Equal(t, models.ErrPermissionsWithRoleNotAllowed.Error(), respJSON.Get("error").MustString())
+						assert.Equal(t, models.ErrPermissionsWithRoleNotAllowed.Error(), respJSON["error"])
 					})
 			}
 		})
--- a/pkg/api/dashboard_permission_test.go
+++ b/pkg/api/dashboard_permission_test.go
@@ -166,6 +166,33 @@ func TestDashboardPermissionAPIEndpoint(t *testing.T) {
 				})
 		})
 
+		t.Run("When trying to update team or user permissions with a role", func(t *testing.T) {
+			role := models.ROLE_EDITOR
+			cmds := []dtos.UpdateDashboardAclCommand{
+				{
+					Items: []dtos.DashboardAclUpdateItem{
+						{UserId: 1000, Permission: models.PERMISSION_ADMIN, Role: &role},
+					},
+				},
+				{
+					Items: []dtos.DashboardAclUpdateItem{
+						{TeamId: 1000, Permission: models.PERMISSION_ADMIN, Role: &role},
+					},
+				},
+			}
+
+			for _, cmd := range cmds {
+				updateDashboardPermissionScenario(t, "When calling POST on", "/api/dashboards/id/1/permissions",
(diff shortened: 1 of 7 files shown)


pkg/api/dashboard_permission_test.go
```diff
@@ -166,6 +166,33 @@ func TestDashboardPermissionAPIEndpoint(t *testing.T) {
...
+					},
+				},
+			}
+
+			for _, cmd := range cmds {
+				updateDashboardPermissionScenario(t, "When calling POST on", "/api/dashboards/id/1/permissions",
+					"/api/dashboards/id/:id/permissions", cmd, func(sc *scenarioContext) {
+						callUpdateDashboardPermissions(sc)
+						assert.Equal(t, 400, sc.resp.Code)
+						respJSON, err := simplejson.NewJson(sc.resp.Body.Bytes())
```
aknuds1, line 189: Can we use the standard json package instead of simplejson? I think we want to move away from the latter.

pkg/api/folder_permission_test.go
```diff
@@ -172,6 +172,33 @@ func TestFolderPermissionAPIEndpoint(t *testing.T) {
...
+				},
+			},
+		}
+
+		for _, cmd := range cmds {
+			updateFolderPermissionScenario(t, "When calling POST on", "/api/folders/uid/permissions",
+				"/api/folders/:uid/permissions", cmd, func(sc *scenarioContext) {
+					callUpdateFolderPermissions(sc)
+					assert.Equal(t, 400, sc.resp.Code)
+					respJSON, err := simplejson.NewJson(sc.resp.Body.Bytes())
```
aknuds1, line 195: Also wondering if we can use the standard json package here, instead of simplejson.

pkg/models/dashboard_acl.go
```diff
@@ -24,10 +24,11 @@ func (p PermissionType) String() string {
...
 var (
-	ErrDashboardAclInfoMissing           = errors.New("User id and team id cannot both be empty for a dashboard permission")
-	ErrDashboardPermissionDashboardEmpty = errors.New("Dashboard Id must be greater than zero for a dashboard permission")
-	ErrFolderAclInfoMissing              = errors.New("User id and team id cannot both be empty for a folder permission")
-	ErrFolderPermissionFolderEmpty       = errors.New("Folder Id must be greater than zero for a folder permission")
+	ErrDashboardAclInfoMissing           = errors.New("user id and team id cannot both be empty for a dashboard permission")
+	ErrDashboardPermissionDashboardEmpty = errors.New("dashboard id must be greater than zero for a dashboard permission")
+	ErrFolderAclInfoMissing              = errors.New("user id and team id cannot both be empty for a folder permission")
+	ErrFolderPermissionFolderEmpty       = errors.New("folder id must be greater than zero for a folder permission")
+	ErrPermissionsWithRoleNotAllowed     = errors.New("teams and users permissions cannot have an associated role")
```
aknuds1, line 31: Isn't it more correct to say "team permissions" and "user permissions"?
```suggestion
	ErrPermissionsWithRoleNotAllowed     = errors.New("team and user permissions cannot have an associated role")
```

aknuds1 (approved):
Not really familiar with the functionality/semantics, but generally LGTM. See comments though.

marefr (approved):
LGTM

mckn (approved):
LGTM!

hugohaggmark (approved):
LGTM!

PR #29106, opened 2020-11-13
Refactor declarative series configuration to a config builder

dprokop:
Refactors declarative, React-first API to build series, scales and axes cofig.

TODO:
- [x] Plugins are currently broken - need to update the logic indicating that plot is ready to be initialized, it's overcomplicated
- [ ] Tests are failing (due to above)
- [x] Config tests needs to be updated to reflect the builder, currently commented out


packages/grafana-ui/src/components/uPlot/context.ts
```diff
@@ -74,19 +52,19 @@ export const usePlotPluginContext = (): PlotPluginsContextType => {
...
+//   if (!ctx) {
+//     throwWhenNoContext('usePlotConfigContext');
+//   }
+//
+//   return {
+//     addSeries: ctx!.addSeries,
+//     addAxis: ctx!.addAxis,
+//     addScale: ctx!.addScale,
+//   };
+// };
```
dprokop, line 67:
```suggestion
```

public/app/plugins/panel/graph3/GraphPanel.tsx
```diff
@@ -28,8 +28,8 @@ export const GraphPanel: React.FC<GraphPanelProps> = ({
       <TooltipPlugin mode={options.tooltipOptions.mode as any} timeZone={timeZone} />
       <ZoomPlugin onZoom={onChangeTimeRange} />
       <ContextMenuPlugin />
-      {data.annotations && <ExemplarsPlugin exemplars={data.annotations} timeZone={timeZone} />}
-      {data.annotations && <AnnotationsPlugin annotations={data.annotations} timeZone={timeZone} />}
+      {data.annotations ? <ExemplarsPlugin exemplars={data.annotations} timeZone={timeZone} /> : <></>}
+      {data.annotations ? <AnnotationsPlugin annotations={data.annotations} timeZone={timeZone} /> : <></>}
```
dprokop, line 32: To be fixed

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -46,110 +47,101 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
+      const customConfig = config.custom;
+
+      if (i === timeIndex || field.type !== FieldType.number) {
+        continue;
+      }
+
+      const fmt = field.display ?? defaultFormatter;
+      const scale = config.unit || '__fixed';
+
+      if (!uniqueScales[scale]) {
```
torkelo, line 90: can we move this unique scales dictionary to the builder and just have !builder.hasScale(scale)

torkelo (commented):
:heart: awesome! big improvement I think in terms of less complex and easy to grock config building process, and more performant and easy to memoize the whole thing

leeoniya:
@dprokop 

i rebased my DataFrame alignment work onto this PR.

since the core logic did not change much, i was able to reuse & wire up pretty much everything i had that was previously based on master. generally, the closer things are to vanilla es6, the more i like them, so 👍 for any simplification / flattening.

i don't have much to add beyond this since i'm still getting familiar with a lot of the code organization and only recognize what's going on at the bottom and now how everything flows from the top.

i've run into what i think is a typescript issue that i can't seem to resolve without more help (see out slack convo).

leeoniya (approved):

packages/grafana-ui/src/components/GraphNG/GraphConfigBuilder.tsx
```diff
@@ -0,0 +1,143 @@
+import tinycolor from 'tinycolor2';
```
torkelo, line 1: I think this builder should be part of the Plot component (the uPlot abstraction), not the GraphNG component. 

This builder is about mapping Grafana high-level prop concepts & types to internal uPlot config.

packages/grafana-ui/src/components/GraphNG/GraphConfigBuilder.tsx
```diff
@@ -0,0 +1,143 @@
...
+    config.series = [...config.series, ...this.series.map(s => s.getConfig())];
+    config.scales = this.scales.reduce((acc, s) => {
+      return { ...acc, ...s.getConfig() };
+    }, {});
+
+    return config;
+  }
+}
+
+class GraphAxisBuilder extends UPlotConfigBuilder<AxisProps, uPlot.Axis> {
```
torkelo, line 41: would be nice to move this to a uPlot/AxisConfig file (or something like that), and have that file also contain the calculateSpace & formatTime functions in the same file, found having all that together was helpful

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -159,15 +151,13 @@ export const GraphNG: React.FC<GraphNGProps> = ({
       {(vizWidth: number, vizHeight: number) => (
         <UPlotChart
           data={alignedData}
+          config={currentConfig}
```
torkelo, line 154: Not sure what is best, but I was thinking that the builder (our config object) object was the public interface for UPlotChart and that UPlotChart still did not expose any UPlot types. At least if we plan to expose UPlotChart directly to plugins

packages/grafana-ui/src/components/GraphNG/GraphConfigBuilder.tsx
dprokop: agree

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
dprokop: this is done now

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
dprokop: absolutely

packages/grafana-ui/src/components/uPlot/hooks.test.ts
```diff
@@ -1,9 +1,11 @@
 import { usePlotConfig } from './hooks';
-import { renderHook, act } from '@testing-library/react-hooks';
+import { renderHook } from '@testing-library/react-hooks';
+import { act } from '@testing-library/react';
+import { UPlotConfigBuilder } from './config/UPlotConfigBuilder';
 
 describe('usePlotConfig', () => {
   it('returns default plot config', async () => {
-    const { result } = renderHook(() => usePlotConfig(0, 0, 'browser'));
+    const { result } = renderHook(() => usePlotConfig(0, 0, 'browser', new UPlotConfigBuilder()));
```
dprokop, line 8: @Clarity-89 I'm struggling to get this test to work. Would love to get some help from you. As you can see here https://github.com/grafana/grafana/blob/45b43598a6d6b41af9e8f3974cfd4fcf1e066e57/packages/grafana-ui/src/components/uPlot/hooks.ts the plugins hook uses rAF to make sure that all plugins have registered. The `arePluginsReady` flag is being used by the main `usePlotConfig` hook to indicate that [we are ready to create the final config for the uPlot](https://github.com/grafana/grafana/blob/45b43598a6d6b41af9e8f3974cfd4fcf1e066e57/packages/grafana-ui/src/components/uPlot/hooks.ts#L123). But I cannot get the rAF to work correctly in the test. I've tried couple of approaches with mocking rAF included, but it seems that the hook is being unmounted too soon, meaning the `arePluginsRead` never has a chance to be set to true. 

Any good practices, advice ho to aproach rAF with react-testing-library?

torkelo (approved):

packages/grafana-ui/src/components/uPlot/hooks.test.ts
torkelo: can we comment out this so we can merge this?

Clarity-89: I think the best approach would still be to mock RAF, I'll check why it might not be working as expected.

packages/grafana-ui/src/components/uPlot/hooks.ts
```diff
@@ -121,136 +120,24 @@ export const usePlotConfig = (width: number, height: number, timeZone: TimeZone)
...
-              tmp[key] = config;
-            }
-            return tmp;
-          });
-        },
-      };
-    },
-    [setScaleConfig]
-  );
+  }, [arePluginsReady, plugins, width, height, configBuilder]);
```
Clarity-89, line 138: `tzDate` is missing from dependency list.

packages/grafana-ui/src/components/uPlot/hooks.test.ts
dprokop: @torkelo yup, I'll comment this out for now and try to nail it down independently

Clarity-89: @dprokop so the issue here is not as much with the tests as with the dependencies array in usePlotConfig's `useEffect`. Basically `configBuilder` and `plugins` need to be memoized otherwise they seem to trigger multiple `useEffect` calls and the tests hang. For `configBuilder` it's quite easy, you just need to instantiate it before passing to the hook:
```
const builder = new UPlotConfigBuilder();
const { result } = renderHook(() => usePlotConfig(0, 0, 'browser', builder));
```
It's a bit more tricky with `plugins`, since it's a return value from another hook, haven't looked into that but quite sure it can be fixed. 
For mocking RAF I used this package: https://github.com/alexreardon/raf-stub. Also the hooks testing library has `waitForNextUpdate`, which has to be called in order for state updates to be properly applied. So the test would look like this:
```
import { usePlotConfig } from './hooks';
import { renderHook } from '@testing-library/react-hooks';
import { act } from '@testing-library/react';
import { UPlotConfigBuilder } from './config/UPlotConfigBuilder';

import { replaceRaf } from 'raf-stub';

replaceRaf();

describe('usePlotConfig', () => {
  it('returns default plot config', async () => {
    const builder = new UPlotConfigBuilder();
    const { result, waitForNextUpdate } = renderHook(() => usePlotConfig(0, 0, 'browser', builder));
    requestAnimationFrame.step();
    await waitForNextUpdate();

    expect(result.current.currentConfig).toMatchInlineSnapshot(`
      Object {
        "axes": Array [],
        "cursor": Object {
          "focus": Object {
            "prox": 30,
          },
        },
        "focus": Object {
          "alpha": 1,
        },
        "gutters": Object {
          "x": 8,
          "y": 8,
        },
        "height": 0,
        "hooks": Object {},
        "legend": Object {
          "show": false,
        },
        "plugins": Array [],
        "scales": Object {},
        "series": Array [
          Object {},
        ],
        "tzDate": [Function],
        "width": 0,
      }
    `);
  });
});
```

With this test and removing `plugins` from the dependency array, the test passes. Didn't look into the rest of the tests as they depend on plugins, but I'm sure you can take it from here :) Just need to make sure that `plugins` reference is stable during renders.

dprokop: @Clarity-89 thanks a lot for taking a look. For now I'm gonna merge this with test commented out and fix those in a followup PR

PR #29114, opened 2020-11-13
Plugins: Add support for relative URLs to the News panel

josiahg:

Some RSS feeds, including the [Grafana blog](https://grafana.com/blog/news.xml) use relative URLs in the `<item>` `<link>` field. Without this PR, these relative URLs are rendered as relative to the host where Grafana is running. 

This means installs with the default dashboard include broken links out of the box.

IE: http://localhost:3000/some-grafana-blog-post instead of https://grafana.com/some-grafana-blog-post. 

This PR uses a regex to determine if the URL provided in each `<link>` is absolute or relative. If the URL is determined to be relative, the `link` field is populated by concatenating the `protocol` and `host` of the URL with the relative URL provided. Otherwise, the `link` is rendered as-is from the source.

N/A

Happy to take feedback on the code style or best place for this to be implemented!

### DIFF
--- a/public/app/plugins/panel/news/rss.ts
+++ b/public/app/plugins/panel/news/rss.ts
@@ -9,6 +9,10 @@ export async function loadRSSFeed(url: string): Promise<RssFeed> {
     items: [],
   };
 
+  // regex and url for handling feeds with relative paths in <item> <link>
+  const r = new RegExp('^(?:[a-z]+:)?//', 'i');
+  const base_url = new URL(url).protocol + new URL(url).host;
+
   const getProperty = (node: Element, property: string) => {
     const propNode = node.querySelector(property);
     if (propNode) {
@@ -20,7 +24,7 @@ export async function loadRSSFeed(url: string): Promise<RssFeed> {
   doc.querySelectorAll('item').forEach(node => {
     const item: RssItem = {
       title: getProperty(node, 'title'),
-      link: getProperty(node, 'link'),
+      link: r.test(getProperty(node, 'link')) ? getProperty(node, 'link') : base_url + getProperty(node, 'link'),
       content: getProperty(node, 'description'),
       pubDate: getProperty(node, 'pubDate'),
     };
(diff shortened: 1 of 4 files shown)


dprokop (changes requested):
Thank you @josiahg for the contribution! Do you mind adding a test for the newly added logic before we merge it?

hhoffstaette:
@josiahg Thanks for this patch. I wanted to test it & applied the patch to my installed rss.ts, verfied the patched version and finally restarted Grafana. Unfortunately I still see the broken relative outgoing links in the News panel. Suggestions?
Sorry if this is a dumb question, I assume I can just update the .ts files without rebuilding anything else.

josiahg:
@dprokop glad to add testing! I will push an update with those shortly.

@hhoffstaette Thanks for trying the patch! How are you running Grafana? I am following the steps in the [Developer Guide](https://github.com/grafana/grafana/blob/master/contribute/developer-guide.md#build-grafana) - running both the backend with `make run` and frontend with `yarn start`.

hhoffstaette:
> @hhoffstaette Thanks for trying the patch! How are you running Grafana?

Nevermind, it was a stale JS caching issue. Everything working fine now, rss links go to the outside world as expected. Thanks :)

josiahg:
@dprokop Did a bit of a refactor to better fit with the existing tests, and added a test specific for this patch.

Looks like the Grafana blog RSS feed has fixed the issue that led to this effort in the first place. :)

Let me know if this needs additional work to merge!

torkelo:
We fixed this in the RSS feed itself (was a recent bug that made them relative).

So do not think we need this anymore

dprokop:
@torkelo I think it would be good to include this anyways if someone uses the NewsPanel with feed that does not expose absolute urls on the items.

torkelo:
> @torkelo I think it would be good to include this anyway for cases when someone uses the NewsPanel with feed that does not expose absolute urls on the items.

How common is that? the first time and only time I have herd of this is the bug with the Grafana blog RSS which we fixed quickly

aknuds1:
The branch should be merged with master, to fix the CI failure.

josiahg:
@torkelo per the [RSS spec](https://validator.w3.org/feed/docs/warning/ContainsRelRef.html) you are correct that this should be rare.

Unfortunately, based on my reading while researching this issue it seems to be somewhat common in the wild due to poor defaults or otherwise mis-configured services.

torkelo:
Ok, still would wait with this until we have some clear evidence that this is a problem

PR #28918, opened 2020-11-08
Chore: Rewrite playlist sqlstore test from GoConvey to stdlib and testify

ceh:
**What this PR does / why we need it**:

Convert test from GoConvey to stdlib and testify.

**Special notes**:

Part of #22844.

PR #29117, opened 2020-11-14
Chore: Require OrgId to be specified in delete playlist command

ceh:

Require `OrgId` to be specified in delete playlist command to ensure that the interface for deleting playlists reflects implementation.

A follow-up to https://github.com/grafana/grafana/pull/28918, where I noticed that the "Can remove playlist" subtest didn't specify `OrgId`. If `OrgId` isn't specified, then the playlist isn't deleted per https://github.com/grafana/grafana/blob/fbf0d2c0869bb40acbdc3175f287d4c92f3e095a/pkg/services/sqlstore/playlist.go#L116-L117

Require `OrgId` to be specified and add associated tests for it.
Update the "Can remove playlist" subtest to check that the playlist was deleted.

Also add a test for deleting a playlist that doesn't exist while here.

### DIFF
--- a/pkg/services/sqlstore/playlist.go
+++ b/pkg/services/sqlstore/playlist.go
@@ -108,7 +108,7 @@ func GetPlaylist(query *models.GetPlaylistByIdQuery) error {
 }
 
 func DeletePlaylist(cmd *models.DeletePlaylistCommand) error {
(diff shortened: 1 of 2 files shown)


torkelo (approved):
LGTM

PR #29120, opened 2020-11-14
playlist: Improve test

aknuds1:
Improve playlist backend test, by using a sub-test for each test case and dropping assertion message as it becomes unnecessary (overlaps with sub-test description).


ceh (approved):
👍

papagian (approved):

PR #29134, opened 2020-11-16
Update categories.ts

ngtrthanh:
Add name: 'Normal cubic meter per hour', id: 'Nm3/h', fn: toFixedUnit('Nm³/h')

Fixes #


packages/grafana-data/src/valueFormats/categories.ts
```diff
@@ -251,6 +251,7 @@ export const getCategories = (): ValueFormatCategory[] => [
       { name: 'Litre/min (L/min)', id: 'flowlpm', fn: toFixedUnit('L/min') },
       { name: 'milliLitre/min (mL/min)', id: 'flowmlpm', fn: toFixedUnit('mL/min') },
       { name: 'Lux (lx)', id: 'lux', fn: toFixedUnit('lux') },
+      { name: 'Normal cubic meter per hour', id: 'Nm3/h', fn: toFixedUnit('Nm³/h') },
```
torkelo, line 254: normal?

ryantxu:
FYI, you can just type "Nm³/h" into the unit field and get the same result

I am reluctant to add every possible unit combination

PR #29143, opened 2020-11-16
Provisioning: Build paths in an os independent way

### LINKED ISSUE #29137: provisioning dashboards on windows uses linux path separator - and fails
**What happened**:
Grafana does not provision a dashboards on a fresh windows install of grafana - logfile says the error is directory not found.  The error message in the log also indicates the path searched was constructed incorrectly, using the linux style separator '/' instead of the windows style separator '\' for the final directory of the path.

Example log entry: 

    t=2020-11-16T11:01:27+0000 lvl=eror msg="can't read dashboard provisioning files from directory" logger=provisioning.dashboard path="D:\\Program Files\\Grafana\\provisioning/dashboards" error="open D:\\Program Files\\Grafana\\provisioning/dashboards: The system cannot find the file specified."

Notice the final slash before 'dashboards' is '/', but should have been '\\'.

I expected Grafana to provision the dashboards at the path: "D:\Program Files\Grafana\provisioning\dashboards".  Instead, no dashboards are provisioned.

This should be reproducible on any fresh install on a windows host where the user attempts to provision a dashboard from the local filesystem.

You must enable provisioning in the custom.ini with a like like this:

    provisioning = provisioning

Then create a file dashboards.yml in Grafana's provisioning\dashboards directory with the following text

    apiVersion: 1
    
    providers:
      # <string> an unique provider name. Required
      - name: 'Ansible'
      # <int> Org id. Default to 1
      orgId: 1
      # <string> name of the dashboard folder.
      folder: 'dashboards'
      # <string> folder UID. will be automatically generated if not specified
      folderUid: ''
      # <string> provider type. Default to 'file'
      type: file
      # <bool> disable dashboard deletion
      disableDeletion: true
      # <int> how often Grafana will scan for changed dashboards
      updateIntervalSeconds: 60
      # <bool> allow updating provisioned dashboards from the UI
      allowUiUpdates: false
      options:
        # <string, required> path to dashboard files on disk. Required when using the 'file' type
        path: D:\Program\ Files\Grafana\data\dashboards\
        # <bool> use folder names from filesystem to create folders in Grafana
        foldersFromFilesStructure: true

Start Grafana and inspect the logfile.  You will get a line similar to the line in the "what happened" section above.

The file [provisioning.go](https://github.com/grafana/grafana/blob/master/pkg/services/provisioning/provisioning.go) imports path, and uses path.join() in [line 136](https://github.com/grafana/grafana/blob/master/pkg/services/provisioning/provisioning.go#L136).  The GO documentation for path [here](https://golang.org/pkg/path/) states that "This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package."  This is likely the root cause of this issue.

**Environment**:
- Grafana version: Grafana 7.3.0
- Data source type & version: MSSQL 2016, although this did not likely impact the issue.
- OS Grafana is installed on: Windows server 2016 v1607 used, but the error will likely apply to all versions of windows, both server and desktop.
- User OS & Browser: MacOS, Safari 14.0.1 and Firefox 82.0.2
- Grafana plugins: world-map-plugin (not required to recreate error)
- Others:

amattheisen:
This PR builds provisioning paths in an os independent way.   It is required to use provisioning on windows OS.

The patch implements filepath instead of path to manipulate filename paths in a way compatible with the target operating system-defined file paths.

Fixes #29137

None.

### DIFF
--- a/pkg/services/provisioning/provisioning.go
+++ b/pkg/services/provisioning/provisioning.go
@@ -2,7 +2,7 @@ package provisioning
 
 import (
 	"context"
-	"path"
+	"path/filepath"
 	"sync"
 
 	"github.com/grafana/grafana/pkg/infra/log"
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
LGTM

aknuds1:
Thanks for the fix!

aknuds1:
This is targeted for v7.3.3.

PR #29165, opened 2020-11-17
Table: Fixes issues with phantom extra 0 for zero values 

### LINKED ISSUE #29108: Elasticsearch phantom zero
**What happened**: Running into issues in Grafana 7.2.3 where there is a phantom zero appearing for values in my Elasticsearch documents when the value of that field is `0`.

**What you expected to happen**: Only one `0` to show up.

**How to reproduce it (as minimally and precisely as possible)**: Try displaying a field via `raw data` mode in Elasticsearch in Grafana and you will (hopefully) see an extra `0`.

**Anything else we need to know?**: This also happens with the value mapping as well. The value appears to be outside of the underlying `<div>` as well.

**Environment**:
- Grafana version: 7.2.3
- Data source type & version: Elasticsearch 7.0
- User OS & Browser: Mac/Chrome

Screen Shot 2020-11-13 at 11 11 44 AM
Screen Shot 2020-11-13 at 10 01 56 AM

dprokop:
Fixes https://github.com/grafana/grafana/issues/29108

### DIFF
--- a/packages/grafana-ui/src/components/Table/DefaultCell.tsx
+++ b/packages/grafana-ui/src/components/Table/DefaultCell.tsx
@@ -49,7 +49,7 @@ export const DefaultCell: FC<TableCellProps> = props => {
           {value}
         </a>
       )}
-      {showFilters && cell.value && <FilterActions {...props} />}
+      {showFilters && cell.value !== undefined && <FilterActions {...props} />}
     </div>
   );
 };


torkelo (approved):

packages/grafana-ui/src/components/Table/DefaultCell.tsx
```diff
@@ -49,7 +49,7 @@ export const DefaultCell: FC<TableCellProps> = props => {
           {value}
         </a>
       )}
-      {showFilters && cell.value && <FilterActions {...props} />}
+      {showFilters && cell.value !== undefined && <FilterActions {...props} />}
```
hugohaggmark, line 52: nit: I read somewhere about React best practices that we should try to use ternaries instead.
```suggestion
      {showFilters && cell.value ?  <FilterActions {...props} /> : null}
```

hugohaggmark (approved):
Looks great!

packages/grafana-ui/src/components/Table/DefaultCell.tsx
bertho-zero: It will display 'false' as text if the condition is false, should be:

```jsx
{showFilters && cell.value !== undefined ? <FilterActions {...props} /> : null}
```

torkelo:
This was a really poor title for changelog, explains not the user impact of the bugfix (phantom zeros for zero values). Need to remember PR titles matter :)

PR #29167, opened 2020-11-17
Panels: fix positioning of the header title

### LINKED ISSUE #29112: The Stat panel shows the title incorrectly if there is a description
The Stat panel shows the title incorrectly if there is a description and the panel size is small.

Grafana 7.1.5

Grafana 7.2.x, 7.3.x
7 2 x, 7 3 x

**Environment**:
- 7.2.x ,7.3.x:

krvajal:
the menu toggle icon is now absolute positioned and it no longer  affects the centering of the menu title

Fixes the positioning of the panel title 

Fixes #29112

### DIFF
--- a/public/sass/components/_panel_header.scss
+++ b/public/sass/components/_panel_header.scss
@@ -60,6 +60,8 @@ $panel-header-no-title-zindex: 1;
 }
 
 .panel-menu-toggle {
+  position: absolute;
+  top: calc(50% - 9px);
   color: $text-color-weak;
   cursor: pointer;
   margin: 2px 0 0 2px;


public/sass/components/_panel_header.scss
```diff
@@ -60,6 +60,8 @@ $panel-header-no-title-zindex: 1;
 }
 
 .panel-menu-toggle {
+  position: absolute;
+  top: calc(50% - 9px);
```
krvajal, line 64: Not a fan of this solution, but I don't know a better way of doing it. I noticed we are doing something like this already if you look a few lines up on this file. So I don't expect this to be a problem unless in the future the icon size changes

torkelo: not sure the calc is needed here, since you still need a magic pixel count

krvajal: The magic number `9px` comes from `icon_height/2 + icon_y_margin` where `icon_height = 16px` and `icon_y_margin = 2px`. The container element (`.panel-title`) currently has a height of `28px`. So I could replace the `calc` by just `5px` but in the container changes its height in the future the magic number `5px` will not work anymore.

Let me know how should I proceed

PR #29178, opened 2020-11-18
Variables: New Variables are stored immediately

### LINKED ISSUE #20703: Template Variable save UX
When I add a variable query and it returns data, it pushes the “add” button off the screen and I forget to scroll down and click it.  The scrollbar is very faint.  There is no warning when you navigate away etc.
Could the save be moved up top right, similar to dashboards? or perhaps a modal alerting you that you did not save?

### LINKED ISSUE #24477: Templating: Changing label does not trigger unsaved changes warning when moving away from dashboard without saving
**What happened**:
Two small things, should these be in two seperate issues?

**1) Dashboard variable changes are applied without hitting the `Update` button**
**2) After changing the dashboard variable and leaving the dashboard the change is discarded without confirmation**

@ 1) Edit should only be saved when hitting `Update` otherwise the button is useless. This is already the case in `v6.7.2`
@ UI Team: The `Update` button is way at the buttom of the page and you have to scroll if there are many label values.
If you create a **new variable** instead it **will be lost without warning** if you do not hit the `Add` button at the bottom and go back to the dashboard. I managed to do that on several occasions..

@ 2) When leaving the dashboard there should be a confirmation before discarding changes. Grafana Version `6.7.2` **does** bring up a popup when leaving the dashboard:

grafik

- Create a dashboard with a variable, save the dashboard
- Edit the variable, e.g. change `Label` or enable `Multi-value`
- Go back to dashboard **without clicking `Update` button**
- The changes **are applied**
- Hit `F5` or navigate to another dashboard, the variable **change is discarded without confirmation**

Startcondition, `Label` is "Commit", no `Multi-Value`:
before

## Demo:
CXZOxN3XjK

**Environment**:
- Grafana version: `Grafana v7.0.0-beta3 (2169fc1a3f)`
- OS Grafana is installed on: `Windows 10 1909`
- User OS & Browser: `Windows 10 1909` , Firefox 75

hugohaggmark:
Before when you created a new variable if you forgot to __`Add`__ it all the work with that variable was gone as it wasn't added to the dashboard and stored. With this PR we change the behavior to be that whenever you create a new variable, the variable is stored immediately.

Fixes #20703
Fixes #24477


public/app/features/variables/inspect/VariablesDependenciesButton.tsx
```diff
@@ -21,6 +21,10 @@ export const UnProvidedVariablesDependenciesButton: FC<Props> = ({ variables })
   const nodes = useMemo(() => createDependencyNodes(variables), [variables]);
   const edges = useMemo(() => createDependencyEdges(variables), [variables]);
 
+  if (!edges.length) {
+    return null;
+  }
+
```
hugohaggmark, line 27: I noticed the dependency graph button is visible all the time, so this hides it if there are no dependencies. Not exactly part of this PR though.

mckn (commented):
The code looks really great however I'm not a big fan of the UX I think we probably can figure out a better way of doing this.

Kapture 2020-11-18 at 09 37 47

Feels like you need to do so many clicks to create a variable. Will see if I can think of another way of doing this.

hugohaggmark:
> The code looks really great however I'm not a big fan of the UX I think we probably can figure out a better way of doing this.
> 
> Kapture 2020-11-18 at 09 37 47
> 
> Feels like you need to do so many clicks to create a variable. Will see if I can think of another way of doing this.

I agree. I'll revert the button.

public/app/features/variables/editor/actions.ts
```diff
@@ -131,3 +105,14 @@ export const switchToEditMode = (identifier: VariableIdentifier): ThunkResult<vo
 export const switchToListMode = (): ThunkResult<void> => dispatch => {
   dispatch(clearIdInEditor());
 };
+
+export function getNextAvailableId(type: VariableType, variables: VariableModel[]): string {
+  let counter = 0;
+  let nextId = `${type}${counter}`;
+
+  while (variables.find(variable => variable.id === nextId)) {
```
dprokop, line 113: haven't seen `while` for a while;)

dprokop (approved):
Great work @hugohaggmark, just tested it and it works as expected. This is a huge improvement IMO.

There is only one, super small thing that I have noticed, probably not worth tackling unless it's a trivial fix. Basically, when you add a new variable, the id is assigned with the type of the variable as part of it. Then, if you change the variable type the id is not being updated. I assume the _type in the name_ is irrelevant in the logic behind variables, hence this is minor thing:

Kapture 2020-11-19 at 10 13 33

public/app/features/variables/editor/actions.ts
hugohaggmark: LOL 😆

hugohaggmark:
> Great work @hugohaggmark, just tested it and it works as expected. This is a huge improvement IMO.
> 
> There is only one, super small thing that I have noticed, probably not worth tackling unless it's a trivial fix. Basically, when you add a new variable, the id is assigned with the type of the variable as part of it. Then, if you change the variable type the id is not being updated. I assume the _type in the name_ is irrelevant in the logic behind variables, hence this is minor thing:
> 
> Kapture 2020-11-19 at 10 13 33

Yeah, I thought about that as well. But that probably only applies for new variables. When changing the type for an existing variable, we shouldn't change the name and that gets a bit tricky so I skipped this :)

torkelo:
Great improvement! 

Would love a full UX overview of the template variables form. for example we could add separate page where users select the variable type, with detailed desciptions and use cases for each type (similar to the flow when you create a datasource)

torkelo (approved):

hugohaggmark:
> Great improvement!
> 
> Would love a full UX overview of the template variables form. for example we could add separate page where users select the variable type, with detailed desciptions and use cases for each type (similar to the flow when you create a datasource)

I love this idea ❤️

PR #29223, opened 2020-11-19
Chore: Handle wrapped errors

aknuds1:
Rewrite some (not all in this one PR) error checking code in the backend to handle [wrapped errors](https://blog.golang.org/go1.13-errors). Comparing errors directly is a bug waiting to happen, since it's now typical to wrap them.

**NB:** GitHub shows files as changed that the PR doesn't touch, must be something funny with their 3-dot diff algorithm again :/ If you diff locally, you'll see the actual changeset. You can also filter out all non-Go files in the GitHub UI, and you'll see what's actually changed.

### DIFF
--- a/pkg/api/admin_users.go
+++ b/pkg/api/admin_users.go
@@ -95,7 +95,7 @@ func AdminUpdateUserPermissions(c *models.ReqContext, form dtos.AdminUpdateUserP
 	}
 
 	if err := bus.Dispatch(&cmd); err != nil {
-		if err == models.ErrLastGrafanaAdmin {
+		if errors.Is(err, models.ErrLastGrafanaAdmin) {
 			return Error(400, models.ErrLastGrafanaAdmin.Error(), nil)
 		}
 
@@ -111,7 +111,7 @@ func AdminDeleteUser(c *models.ReqContext) Response {
 	cmd := models.DeleteUserCommand{UserId: userID}
 
 	if err := bus.Dispatch(&cmd); err != nil {
-		if err == models.ErrUserNotFound {
+		if errors.Is(err, models.ErrUserNotFound) {
 			return Error(404, models.ErrUserNotFound.Error(), nil)
 		}
 		return Error(500, "Failed to delete user", err)
@@ -126,13 +126,13 @@ func (hs *HTTPServer) AdminDisableUser(c *models.ReqContext) Response {
 
 	// External users shouldn't be disabled from API
 	authInfoQuery := &models.GetAuthInfoQuery{UserId: userID}
-	if err := bus.Dispatch(authInfoQuery); err != models.ErrUserNotFound {
+	if err := bus.Dispatch(authInfoQuery); !errors.Is(err, models.ErrUserNotFound) {
 		return Error(500, "Could not disable external user", nil)
 	}
 
 	disableCmd := models.DisableUserCommand{UserId: userID, IsDisabled: true}
 	if err := bus.Dispatch(&disableCmd); err != nil {
-		if err == models.ErrUserNotFound {
+		if errors.Is(err, models.ErrUserNotFound) {
 			return Error(404, models.ErrUserNotFound.Error(), nil)
 		}
 		return Error(500, "Failed to disable user", err)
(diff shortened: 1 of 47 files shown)


pkg/api/admin_users.go
```diff
@@ -126,13 +126,13 @@ func (hs *HTTPServer) AdminDisableUser(c *models.ReqContext) Response {
 
 	// External users shouldn't be disabled from API
 	authInfoQuery := &models.GetAuthInfoQuery{UserId: userID}
-	if err := bus.Dispatch(authInfoQuery); err != models.ErrUserNotFound {
+	if err := bus.Dispatch(authInfoQuery); !errors.Is(err, models.ErrUserNotFound) {
```
wbrowne, line 129: Shouldn't we also be checking for `err != nil` here?

pkg/api/admin_users.go
```diff
@@ -152,13 +152,13 @@ func AdminEnableUser(c *models.ReqContext) Response {
...
```
wbrowne, line 155: Shouldn't we also be checking for `err != nil` here?

wbrowne (changes requested):
Looks good but spotted something!

pkg/api/login.go
```diff
@@ -263,7 +264,8 @@ func (hs *HTTPServer) Logout(c *models.ReqContext) {
 		return
 	}
 
-	if err := hs.AuthTokenService.RevokeToken(c.Req.Context(), c.UserToken); err != nil && err != models.ErrUserTokenNotFound {
+	if err := hs.AuthTokenService.RevokeToken(c.Req.Context(), c.UserToken); err != nil && !errors.Is(
+		err, models.ErrUserTokenNotFound) {
```
sakjur, line 268: Proposing a change in formatting that would change the scope of `err` for the sake of arguably slightly cleaner code, but it doesn't look like it's used anywhere else in the method, so it should be fine. 

```suggestion
	err := hs.AuthTokenService.RevokeToken(c.Req.Context(), c.UserToken)
	if err != nil && !errors.Is(err, models.ErrUserTokenNotFound) {
```

pkg/api/ldap_debug.go
```diff
@@ -314,7 +314,7 @@ func (hs *HTTPServer) GetUserFromLDAP(c *models.ReqContext) Response {
 	cmd := &models.GetTeamsForLDAPGroupCommand{Groups: user.Groups}
 	err = bus.Dispatch(cmd)
 
-	if err != bus.ErrHandlerNotFound && err != nil {
+	if err != nil && !errors.Is(err, bus.ErrHandlerNotFound) {
```
sakjur, line 317: :pray: I appreciate this 🔄

pkg/api/dashboard_permission.go
```diff
@@ -89,7 +89,8 @@ func UpdateDashboardPermissions(c *models.ReqContext, apiCmd dtos.UpdateDashboar
 	}
 
 	if err := bus.Dispatch(&cmd); err != nil {
-		if err == models.ErrDashboardAclInfoMissing || err == models.ErrDashboardPermissionDashboardEmpty {
+		if errors.Is(err, models.ErrDashboardAclInfoMissing) || errors.Is(err,
+			models.ErrDashboardPermissionDashboardEmpty) {
```
sakjur, line 93:
```suggestion
		if errors.Is(err, models.ErrDashboardAclInfoMissing) ||
			errors.Is(err, models.ErrDashboardPermissionDashboardEmpty) {
```

sakjur (approved):
LGTM

pkg/api/admin_users.go
aknuds1: @wbrowne I wondered about the same and I'm unsure since it was like that in the old code. You're probably right?

pkg/api/admin_users.go
aknuds1: @wbrowne I wondered about the same and I'm unsure since it was like that in the old code. You're probably right?

pkg/api/login.go
aknuds1: I don't really see why it's cleaner, but :man_shrugging:

pkg/api/admin_users.go
aknuds1: @wbrowne After applying your change, tests fail so I think it's actually supposed to work this way, even if it's confusing.

pkg/api/admin_users.go
aknuds1: @wbrowne After applying your change, tests fail so I think it's actually supposed to work this way, even if it's confusing.

wbrowne (approved):

PR #29225, opened 2020-11-19
Chore: remove compress:release grunt task

zoltanbedi:

After https://github.com/grafana/grafana/pull/28624 is merged it turned out that `grunt-contrib-compress` was used. This pr removes it's usage.

### DIFF
--- a/scripts/grunt/release_task.js
+++ b/scripts/grunt/release_task.js
@@ -4,10 +4,10 @@ module.exports = function(grunt) {
   'use strict';
 
   // build then zip
-  grunt.registerTask('release', ['build', 'build-post-process', 'compress:release']);
+  grunt.registerTask('release', ['build', 'build-post-process']);
 
   // package into archives
-  grunt.registerTask('package', ['clean:temp', 'build-post-process', 'compress:release']);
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):

torkelo:
I think we should be able to remove grunt completely, the only remaining thing I can find is  these to regex file tests that makes sure we do not check in tests with only or focus:
https://github.com/grafana/grafana/blob/master/scripts/grunt/default_task.js#L31

torkelo (approved):

aknuds1:
@torkelo We moved the FocusConvey check to the Drone config, so we don't need Grunt for that any longer.

torkelo:
@aknuds1 nice, where is that defined? can we add one for frontend only tests as well? is the focus check part of the pre-commit check like the only test is?

papagian (approved):

zoltanbedi:
@torkelo we can use [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest/) to check for no only tests and it can check it in the pre commit phase. I'll add it later.

torkelo:
@zoltanbedi thats great!

PR #29227, opened 2020-11-19
Chore: Enable errorlint linter

aknuds1:
Enable the errorlint linter, which verifies that wrapped errors are checked correctly. Also enabling a few other linters that don't flag anything at the moment, but might come in handy.

### DIFF
--- a/pkg/middleware/auth_proxy/auth_proxy.go
+++ b/pkg/middleware/auth_proxy/auth_proxy.go
@@ -181,12 +181,12 @@ func (auth *AuthProxy) Login(logger log.Logger, ignoreCache bool) (int64, *Error
 	}
 
 	if isLDAPEnabled() {
-		id, e := auth.LoginViaLDAP()
-		if e != nil {
-			if errors.Is(e, ldap.ErrInvalidCredentials) {
+		id, err := auth.LoginViaLDAP()
+		if err != nil {
+			if errors.Is(err, ldap.ErrInvalidCredentials) {
 				return 0, newError("proxy authentication required", ldap.ErrInvalidCredentials)
 			}
-			return 0, newError("failed to get the user", e)
+			return 0, newError("failed to get the user", err)
 		}
 
 		return id, nil
--- a/pkg/middleware/auth_proxy/auth_proxy.go
+++ b/pkg/middleware/auth_proxy/auth_proxy.go
@@ -2,6 +2,7 @@ package authproxy
 
 import (
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"net"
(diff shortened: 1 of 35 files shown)


pkg/middleware/auth_proxy/auth_proxy.go
```diff
@@ -182,7 +183,7 @@ func (auth *AuthProxy) Login(logger log.Logger, ignoreCache bool) (int64, *Error
 	if isLDAPEnabled() {
 		id, e := auth.LoginViaLDAP()
 		if e != nil {
-			if e == ldap.ErrInvalidCredentials {
+			if errors.Is(e, ldap.ErrInvalidCredentials) {
```
sakjur, line 186: Can we rename `e` -> `err`?

pkg/services/dashboards/folder_service_test.go
```diff
@@ -194,9 +195,8 @@ func TestFolderService(t *testing.T) {
 
 			for _, tc := range testCases {
 				actualError := toFolderError(tc.ActualError)
-				if actualError != tc.ExpectedError {
-					t.Errorf("For error '%s' expected error '%s', actual '%s'", tc.ActualError, tc.ExpectedError, actualError)
-				}
+				require.EqualErrorf(t, actualError, tc.ExpectedError.Error(),
```
sakjur, line 198:
```suggestion
				assert.EqualErrorf(t, actualError, tc.ExpectedError.Error(),
```

To make sure all failures are listed.

pkg/services/rendering/plugin_mode.go
```diff
@@ -45,11 +46,12 @@ func (rs *RenderingService) renderViaPluginV1(ctx context.Context, renderKey str
...
 	rsp, err := rs.pluginInfo.GrpcPluginV1.Render(ctx, req)
-	if ctx.Err() == context.DeadlineExceeded {
-		rs.log.Info("Rendering timed out")
-		return nil, ErrTimeout
-	}
 	if err != nil {
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			rs.log.Info("Rendering timed out")
+			return nil, ErrTimeout
+		}
```
sakjur, line 53: Should this have been lifted in the `if err != nil`? Is `err` guaranteed to not be `nil` if `ctx.Err()` is set?

sakjur (commented):
Looks good, a couple of questions

pkg/middleware/auth_proxy/auth_proxy.go
aknuds1: @sakjur Renaming it to `err` does work, since it has the `error` interface. I didn't notice first that it does, but yeah.

pkg/services/rendering/plugin_mode.go
aknuds1: Good point @sakjur. Looking again.

aknuds1: @sakjur Good catch, it was a mistake!

pkg/services/dashboards/folder_service_test.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/grafana/grafana/pkg/bus"
 	"github.com/grafana/grafana/pkg/models"
+	"github.com/stretchr/testify/require"
```
sakjur, line 8:
```suggestion
	"github.com/stretchr/testify/assert"
```

sakjur (approved):
LGTM (I think it'll fail compilation though, added another suggestion for that)

PR #29232, opened 2020-11-19
BarGauge: Fix story for BarGauge, caused knobs to show for other stories

torkelo:
Fixes problem where knobs for this story would show on all other stories on first load


mckn (approved):
LGTM

jackw (approved):
Nicely done! 🚀

PR #29236, opened 2020-11-19
sdk: use version without transform (gel) code

kylebrandt:

Follow up to removing GEL (updating to sdk that has transform/gel related code removed).

Depends on https://github.com/grafana/grafana-plugin-sdk-go/pull/239 being merged and tagged, and this updated to that tag.


aknuds1 (approved):

marefr (approved):
LGTM

PR #29240, opened 2020-11-19
Plugins: Add support for signature manifest V2

### LINKED ISSUE #28899: Plugins: Support plugin manifest version 2.0.0
The new version 2.0.0 plugin manifests include several new fields:

- manifestVersion will start at 2.0.0, with manifests without a version assumed to be the legacy format (aka 1.0.0) and have signatureType grafana
- signatureType will be either private, community, commercial, or grafana
- signedByOrg will be the grafana.com org slug of the signing org
- signedByOrgName will be the grafana.com org name of the signing org
- rootUrls will be required for private signatures, and will contain a list of root_urls that the plugin is authorized to run on

In addition, version 2.0.0 manifests require that all plugin files are covered by the manifest to ensure that plugins cannot be altered by adding or replacing an unsigned file.

The updated signature validation needs to be implemented, and the information about the signature type and signing organization exposed via the API for display in the frontend.

wbrowne:

Fixes #28899

Not 100% on the idea of treating the following cases as a "modified" signature, but it seems to be what we use when encountering content mismatch.
- A file that was listed in the manifest was **not** found in the plugin directory
- A file that was **not** listed in the manifest but was found in the plugin directory


pkg/api/dtos/plugins.go
```diff
@@ -19,25 +19,29 @@ type PluginSetting struct {
...
 	DefaultNavUrl string                      `json:"defaultNavUrl"`
 
-	LatestVersion string                  `json:"latestVersion"`
-	HasUpdate     bool                    `json:"hasUpdate"`
-	State         plugins.PluginState     `json:"state"`
-	Signature     plugins.PluginSignature `json:"signature"`
+	LatestVersion string                        `json:"latestVersion"`
+	HasUpdate     bool                          `json:"hasUpdate"`
+	State         plugins.PluginState           `json:"state"`
+	Signature     plugins.PluginSignatureStatus `json:"signature"`
```
ryantxu, line 25: rather than having this as a flat object... what about making it nested?  and then use the same object in both the singe value and the list value?

wbrowne: Yeah I thought about this - would be cleaner alright. I wanted to have frontend weigh-in and get their opinion before changing

pkg/plugins/models.go
```diff
@@ -21,14 +21,29 @@ var (
...
+}
+
+type PluginSignatureStatus string
+
+const (
+	PluginSignatureInternal PluginSignatureStatus = "internal" // core plugin, no signature
+	PluginSignatureValid    PluginSignatureStatus = "valid"    // signed and accurate MANIFEST
+	PluginSignatureInvalid  PluginSignatureStatus = "invalid"  // invalid signature
+	PluginSignatureModified PluginSignatureStatus = "modified" // valid signature, but content mismatch
+	PluginSignatureUnsigned PluginSignatureStatus = "unsigned" // no MANIFEST file
```
aknuds1, line 37: Can you de-export these as they're only used within the package? If they are to be exported, the comments should be written as descriptive sentences, on account of being part of the package API.
```suggestion
	pluginSignatureInternal PluginSignatureStatus = "internal" // core plugin, no signature
	pluginSignatureValid    PluginSignatureStatus = "valid"    // signed and accurate MANIFEST
	pluginSignatureInvalid  PluginSignatureStatus = "invalid"  // invalid signature
	pluginSignatureModified PluginSignatureStatus = "modified" // valid signature, but content mismatch
	pluginSignatureUnsigned PluginSignatureStatus = "unsigned" // no MANIFEST file
```

pkg/plugins/models.go
```diff
@@ -21,14 +21,29 @@ var (
...
 const (
-	PluginSignatureInternal PluginSignature = "internal" // core plugin, no signature
-	PluginSignatureValid    PluginSignature = "valid"    // signed and accurate MANIFEST
-	PluginSignatureInvalid  PluginSignature = "invalid"  // invalid signature
-	PluginSignatureModified PluginSignature = "modified" // valid signature, but content mismatch
-	PluginSignatureUnsigned PluginSignature = "unsigned" // no MANIFEST file
+	Grafana    PluginSignatureType = "grafana"
+	Commercial PluginSignatureType = "commercial"
+	Community  PluginSignatureType = "community"
+	Private    PluginSignatureType = "private"
```
aknuds1, line 46: It's better if you disambiguate these as a variable name such as `grafana` is extremely ambiguous and will easily clash with other variable names.
```suggestion
	GrafanaType    PluginSignatureType = "grafana"
	CommercialType PluginSignatureType = "commercial"
	CommunityType  PluginSignatureType = "community"
	PrivateType    PluginSignatureType = "private"
```

Furthermore, `Community` and `Commercial` are unused. Can we drop them?

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,127 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+	// Track which plugin files are included in the signature
+	pluginFileInSignature := make(map[string]bool, len(plugin.Files))
+	for _, f := range plugin.Files {
+		pluginFileInSignature[f] = false
 	}
 
 	// Verify the manifest contents
 	log.Debug("Verifying contents of plugin manifest", "plugin", plugin.Id)
 	for p, hash := range manifest.Files {
+		// Verify file is listed in the manifest and exists in the plugin dir
```
aknuds1, line 166: This doesn't look right to me, since `p` stems from the manifest (`manifest.Files`).

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,127 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+		if _, exists := pluginFileInSignature[p]; !exists && manifest.isV2() {
+			log.Warn("Plugin file listed in the manifest was not found", "plugin", plugin.Id, "filename", p, "dir", plugin.PluginDir)
+			return &PluginSignatureState{
+				Status: PluginSignatureModified,
+			}, nil
+		}
```
aknuds1, line 172: I think this block should be removed, since testing whether a file is in the manifest also happens on line 176.
```suggestion
```

aknuds1 (changes requested):
Requesting changes for now, since I think new manifest verification logic has a flaw.

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,127 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+		if !foundMatch {
+			log.Warn("Could not find root URL that matches running application URL", "plugin", plugin.Id, "appUrl", appURL, "rootUrls", manifest.RootUrls)
+			return &PluginSignatureState{
+				Status: PluginSignatureInvalid,
+			}, nil
+		}
+	}
+
+	// Track which plugin files are included in the signature
+	pluginFileInSignature := make(map[string]bool, len(plugin.Files))
```
aknuds1, line 158: I think what you really want here is to make a map of files listed in the manifest, and later use that map to tell if there are extra files not listed there.

pkg/plugins/manifest_test.go
```diff
@@ -46,6 +47,11 @@ NR7DnB0CCQHO+4FlSPtXFTzNepoc+CytQyDAeOLMLmf2Tqhk2YShk+G/YlVX
 		require.NoError(t, err)
 		require.NotNil(t, manifest)
 		assert.Equal(t, "grafana-googlesheets-datasource", manifest.Plugin)
+		assert.Equal(t, "1.0.0-dev", manifest.Version)
+		assert.Equal(t, int64(1586817677115), manifest.Time)
+		assert.Equal(t, "7e4d0c6a708866e7", manifest.KeyID)
+		expectedFiles := []string{"LICENSE", "README.md", "gfx_sheets_darwin_amd64", "gfx_sheets_linux_amd64", "gfx_sheets_windows_amd64.exe", "module.js", "module.js.LICENSE.txt", "module.js.map", "plugin.json"}
```
aknuds1, line 53: It's more readable if you write a long array over several lines:
```suggestion
		expectedFiles := []string{
			"LICENSE", "README.md", "gfx_sheets_darwin_amd64", "gfx_sheets_linux_amd64",
			 "gfx_sheets_windows_amd64.exe", "module.js", "module.js.LICENSE.txt", "module.js.map",
			 "plugin.json",
		}
```

pkg/plugins/plugins.go
```diff
@@ -346,19 +346,32 @@ func (s *PluginScanner) loadPlugin(pluginJSONFilePath string) error {
 	defer reader.Close()
 
 	jsonParser := json.NewDecoder(reader)
-	pluginCommon := PluginBase{}
-	if err := jsonParser.Decode(&pluginCommon); err != nil {
+	plugin := PluginBase{}
```
aknuds1, line 349: Why the renaming? `pluginCommon` made sense, since it's just the `PluginBase` part.

pkg/plugins/plugins.go
```diff
@@ -346,19 +346,32 @@ func (s *PluginScanner) loadPlugin(pluginJSONFilePath string) error {
...
+	if plugin.Id == "" || plugin.Type == "" {
 		return errors.New("did not find type or id properties in plugin.json")
 	}
 
-	pluginCommon.PluginDir = filepath.Dir(pluginJSONFilePath)
-	pluginCommon.Signature = getPluginSignatureState(s.log, &pluginCommon)
+	plugin.PluginDir = filepath.Dir(pluginJSONFilePath)
+	plugin.Files, err = collectPluginFilesWithin(plugin.PluginDir)
+	if err != nil {
+		s.log.Debug("Could not collect plugin file information in directory", "pluginID", plugin.Id, "dir", plugin.PluginDir)
```
aknuds1, line 361: Why Debug level, as opposed to Warn considering it's an error?

pkg/plugins/plugins.go
```diff
@@ -346,19 +346,32 @@ func (s *PluginScanner) loadPlugin(pluginJSONFilePath string) error {
...
+		return err
+	}
+
+	signatureState, err := getPluginSignatureState(s.log, &plugin)
+	if err != nil {
+		s.log.Debug("Could not get plugin signature state", "pluginID", plugin.Id, "err", err)
```
aknuds1, line 367: Why debug rather than warn?

pkg/plugins/plugins_test.go
```diff
@@ -158,6 +158,123 @@ func TestPluginManager_Init(t *testing.T) {
...
+		assert.Equal(t, pluginId, Plugins[pluginId].Id)
+		assert.Equal(t, "1.0.0", Plugins[pluginId].Info.Version)
+		assert.Equal(t, PluginSignatureValid, Plugins[pluginId].Signature)
+		assert.Equal(t, Grafana, Plugins[pluginId].SignatureType)
+		assert.Equal(t, "Grafana Labs", Plugins[pluginId].SignatureOrg)
+		assert.False(t, Plugins[pluginId].IsCorePlugin)
+	})
+
+	t.Run("With back-end plugin with invalid v2 private signature (mismatched root URL)", func(t *testing.T) {
+		origAppUrl := setting.AppUrl
```
aknuds1, line 190:
```suggestion
		origAppURL := setting.AppUrl
```

pkg/plugins/plugins_test.go
```diff
@@ -158,6 +158,123 @@ func TestPluginManager_Init(t *testing.T) {
...
+		}
+		err := pm.Init()
+		require.NoError(t, err)
+
+		assert.Equal(t, []error{fmt.Errorf(`plugin "test" has an invalid signature`)}, pm.scanningErrors)
+		assert.Nil(t, Plugins[("test")])
+	})
+
+	t.Run("With back-end plugin with valid v2 private signature", func(t *testing.T) {
+		origAppUrl := setting.AppUrl
```
aknuds1, line 210:
```suggestion
		origAppURL := setting.AppUrl
```

pkg/plugins/plugins_test.go
```diff
@@ -158,6 +158,123 @@ func TestPluginManager_Init(t *testing.T) {
...
+		assert.Equal(t, pluginId, Plugins[pluginId].Id)
+		assert.Equal(t, "1.0.0", Plugins[pluginId].Info.Version)
+		assert.Equal(t, PluginSignatureValid, Plugins[pluginId].Signature)
+		assert.Equal(t, Private, Plugins[pluginId].SignatureType)
+		assert.Equal(t, "Will Browne", Plugins[pluginId].SignatureOrg)
+		assert.False(t, Plugins[pluginId].IsCorePlugin)
+	})
+
+	t.Run("With back-end plugin with modified v2 signature (missing file from plugin dir)", func(t *testing.T) {
+		origAppUrl := setting.AppUrl
```
aknuds1, line 240:
```suggestion
		origAppURL := setting.AppUrl
```

pkg/plugins/plugins_test.go
```diff
@@ -158,6 +158,123 @@ func TestPluginManager_Init(t *testing.T) {
...
+			BackendPluginManager: &fakeBackendPluginManager{},
+		}
+		err := pm.Init()
+		require.NoError(t, err)
+		assert.Equal(t, []error{fmt.Errorf(`plugin "test"'s signature has been modified`)}, pm.scanningErrors)
+		assert.Nil(t, Plugins[("test")])
+	})
+
+	t.Run("With back-end plugin with modified v2 signature (unaccounted file in plugin dir)", func(t *testing.T) {
+		origAppUrl := setting.AppUrl
```
aknuds1, line 260:
```suggestion
		origAppURL := setting.AppUrl
```

aknuds1 (changes requested):

pkg/plugins/plugins.go
wbrowne: Just because the error will get collected and logged as warns anyway (same as below) - but can change this as it can provide more useful context as warn

pkg/plugins/plugins.go
wbrowne: In that case, why not `pluginBase`?

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,127 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
 }
 
 // getPluginSignatureState returns the signature state for a plugin.
-func getPluginSignatureState(log log.Logger, plugin *PluginBase) PluginSignature {
+func getPluginSignatureState(log log.Logger, plugin *PluginBase) (*PluginSignatureState, error) {
```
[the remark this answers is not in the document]
wbrowne: Yeah just because now with private plugins from https://github.com/grafana/grafana/pull/29240/files#diff-798baa53157b806c13e80304f88da52846c1dd1b02cc79497ac50a2dc722873fR128, there are different error cases to handle. Fine for you?

pkg/plugins/manifest.go
```diff
@@ -111,23 +167,50 @@ func getPluginSignatureState(log log.Logger, plugin *PluginBase) PluginSignature
...
+		if inSignature := pluginFileInSignature[f]; !inSignature && manifest.isV2() {
+			unsignedFiles = append(unsignedFiles, f)
+		}
+	}
+
+	if len(unsignedFiles) > 0 {
+		log.Warn("The following files were not included in the signature", "plugin", plugin.Id, "files", unsignedFiles)
+		return &PluginSignatureState{
+			Status: pluginSignatureModified,
+		}, nil
```
aknuds1, line 206: There shouldn't be any need to execute any of this code if it isn't v2.
```suggestion
	if manifest.isV2() {
		var unsignedFiles []string
		for f := range pluginFileInSignature {
			if !pluginFileInSignature[f] {
				unsignedFiles = append(unsignedFiles, f)
			}
		}
		
		if len(unsignedFiles) > 0 {
			log.Warn("The following files were not included in the signature", "plugin", plugin.Id, "files", unsignedFiles)
			return &PluginSignatureState{
				Status: pluginSignatureModified,
			}, nil
```

pkg/plugins/plugins.go
aknuds1: @wbrowne Makes no big difference, I think the idea is that "base" refers to a base class, while here "common" just refers to common fields (which can be loaded at this initial stage).

The main thing is to signal that this isn't the entire plugin. I know I was confused reading this code, until I realized that the whole plugin gets loaded later, so the explicit naming helps.

pkg/plugins/manifest.go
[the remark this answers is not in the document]
aknuds1: @wbrowne You mean so you can return `nil` here in case of an error? I don't really understand if there might be anything else. If that's the case, you can also just return `PluginSignatureState{}`, noone should care about the non-error return value if an error is returned.

aknuds1 (changes requested):
Generally looks much better thanks! There's still some potential for cleaning up though, please see my comments.

pkg/plugins/plugins.go
wbrowne: Yeah I understand the hierarchy - hard to capture that with just a var name. I'll put it back as it was then for now

pkg/plugins/manifest.go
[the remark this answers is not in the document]
wbrowne: Precisely. I personally prefer the `nil`, also due to the fact the pattern is already followed in this file

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,118 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
-			return PluginSignatureModified
+			return &PluginSignatureState{
+				Status: pluginSignatureModified,
+			}, nil
+		}
+		manifestFiles[p] = true
+	}
+
+	if manifest.isV2() {
+		// Track files that were missing from the manifest
```
aknuds1, line 191: "Were" makes it sound as if it's in the past:
```suggestion
		// Track files missing from the manifest
```

pkg/plugins/manifest.go
```diff
@@ -83,51 +97,118 @@ func readPluginManifest(body []byte) (*pluginManifest, error) {
...
+		var unsignedFiles []string
+		for _, f := range plugin.Files {
+			if exists := manifestFiles[f]; !exists {
```
aknuds1, line 194: You probably want to check whether the key exists in the map, where you have to use the two-return-values form, the benefit of this is also that the type of the map value doesn't matter (i.e., you can replace `bool` with `struct{}`):
```suggestion
			if _, exists := manifestFiles[f]; !exists {
```

pkg/plugins/plugins.go
```diff
@@ -486,3 +499,20 @@ func GetPluginMarkdown(pluginId string, name string) ([]byte, error) {
 	}
 	return data, nil
 }
+
+// gets plugin filenames that require verification for plugin signing
+func collectPluginFilesWithin(rootDir string) ([]string, error) {
+	var files []string
+
+	err := filepath.Walk(rootDir, func(path string, info os.FileInfo, err error) error {
```
aknuds1, line 507:
```suggestion
	err := filepath.Walk(rootDir, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}

```

aknuds1 (approved):
Looks pretty good to me, although I don't agree to use a pointer return value depending on style concerns. There's also a missing error check in callback to `filepath.Walk` (see my suggestion).

pkg/plugins/manifest.go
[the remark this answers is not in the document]
wbrowne: Fair points. In that case, I'll revert and bear the same logic in mind for the future 😄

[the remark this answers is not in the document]
aknuds1: Thanks :) It's no big deal in this case, but it's important to bear in mind that pointers and value types have fundamentally different characteristics. For example, a pointer introduces the risk of a nil pointer panic, which the corresponding value type is safe from. I think pointers are mostly useful when you think there's a performance decrease from copying values, or you need to change the underlying value when passing arguments around.

pkg/api/dtos/plugins.go
ryantxu: Any thoughts on this?  I am nervous about adding more and more to the root element.  There is a long list of more metadata we want to support, so breaking it into clear sections will be helpful

aknuds1: It's a fair point @ryantxu, I don't know enough about client side usage though to offer an opinion.

DanCech:
Looking good 👍

pkg/api/dtos/plugins.go
dprokop: I agree with @ryantxu here, question is what properties would we nest? Is this only about Signarture, SignatureType, and SignatureOrg? Nesting other properties touched by this PR, i.e State, are used in few places and nesting those would require some frontend changes. But I assume Ryan's comment refers to signature specific props.

wbrowne: I assume all related signature properties, which in this case would be the pre-existing `Signature` field which represents the signature state, and both new `SignatureType` and `SignatureOrg` fields. Maybe something like:

```json
...
"signature": {
  "status": "valid",
  "type"  : "private",
  "org"   : "wbrowne"
}
```

wbrowne: @dprokop @ryantxu Are you guys happy to merge as is now (so as not to force a frontend change) with the aim to update this when frontend starts to tie the pieces together?

dprokop: @wbrowne I think it's fine, I think next week I could do some work around front-end updates and then we can update the model on the backend if that's OK with you.

marefr:
Great work!

PR #29241, opened 2020-11-19
Chore: Allow reducerTester to work with every data type & payload-less actions

Elfo404:

currently `reducerTester` doesn't work with any other data structure than objects and `whenActionIsDispatched` only works with actions containing a payload. This PR allows `reducerTester` to be used on any data structure and `whenActionIsDispatched` to accept payload-less actions.


hugohaggmark (approved):
Significant work ⭐

peterholmberg (approved):
Sweet! :+1:

PR #29246, opened 2020-11-19
make it possible to hide change password link in profile menu

### LINKED ISSUE #29245: Disable Change password link in the menu when users cannot change password
In our cloud offering, we dont allow users to change their password. User management is managed on our grafana.com portal so it would be nice to be able to remove the `Change password` link in the bottom left menu.

image

bergquist:
closes #29245

### DIFF
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -207,6 +207,12 @@ var (
 	ImageUploadProvider string
 )
 
+// AddChangePasswordLink returns if logini form is disabled or not since
+// the same intention can be used to hide both features.
+func AddChangePasswordLink() bool {
+	return !DisableLoginForm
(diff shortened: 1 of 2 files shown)


dprokop (approved):
LGTM!

pkg/setting/setting.go
```diff
@@ -207,6 +207,12 @@ var (
 	ImageUploadProvider string
 )
 
+// AddChangePasswordLink returns if logini form is disabled or not since
```
wbrowne, line 210:
```suggestion
// AddChangePasswordLink returns if login form is disabled or not since
```

wbrowne (approved):
LGTM, just noticed a tiny typo!

PR #29265, opened 2020-11-20
Add missing color type

besartberisha:
Noticed that type for [`gray97`](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/themes/default.ts#L6)  was missing while using within plugin development.

### DIFF
--- a/packages/grafana-data/src/types/theme.ts
+++ b/packages/grafana-data/src/types/theme.ts
@@ -136,6 +136,7 @@ export interface GrafanaTheme extends GrafanaThemeCommons {
 
     // New greys palette used by next-gen form elements
     gray98: string;
+    gray97: string;
     gray95: string;
     gray90: string;
     gray85: string;


Clarity-89 (approved):

jackw (approved):
LGTM!

torkelo:
interested why you need this color, from a plugin you should almost never use the palette colors directly, only the high level colors should be used https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/themes/dark.ts#L38

besartberisha:
> interested why you need this color, from a plugin you should almost never use the palette colors directly, only the high level colors should be used https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/themes/dark.ts#L38
> 
> https://developers.grafana.com/ui/latest/index.html?path=/story/docs-overview-themecolors--theme-colors-demo

You are right, I didn't need to use this directly but as I was implementing a custom header for `Drawer` I tested dark/light themes and I noticed that color wasn't typed so I pushed this PR just in case. For this case I'm using the `theme.colors.bg2`.

besartberisha:
Looks like I'm not authorized to merge this branch.

Issue #27703, opened 2020-09-22
Alerting: Create a new hidden page for new alerting rule 


torkelo:
For the new edit alert rule page I want it to share the look and feel & structure of the panel edit UI. 

* [ ] Create new full screen alert rule page & route under /alerting/edit/id 
* [ ] Created shared component for the top bar section with the back button & title and buttons 
* [ ] Create the basic page structure similar panel edit page 
  *  With the tab of Queries, and Sidebar with options and data viz area

Try to reuse/extract to shared components as much as possible from the panel edit UI 

![Screenshot from 2020-09-22 14-14-35](https://user-images.githubusercontent.com/10999/93880603-0977b200-fcde-11ea-837f-65f1a4c14b4f.png)

Mockup: https://www.figma.com/file/UN54dZ5tIAEkGoRl4fIDS8/AlertingNG?node-id=55%3A424
![Screenshot from 2020-09-22 14-25-09](https://user-images.githubusercontent.com/10999/93881671-7fc8e400-fcdf-11ea-83e5-ea2c15a39751.png)

But I image we want to have each box heading be in a single tab just to have the same look and feel of panel edit.

PR #29266, opened 2020-11-20
Core: Move SplitPane layout from PanelEdit.

peterholmberg:
This PR moves the SplitPane logic from the PanelEditor to its own component. The main reason for this is to reuse the same layout in the upcoming alert definition page (wireframes in #27703). 

Part of #27703

There's few areas that's very much up for discussion :sweat_smile: 

* Naming  - Not overly attached to `ThreePaneSplit` but not sure what else to go with either? `SplitPaneComponent`, `SplitLayout`?
* Scope - Is the scope too narrow? I've only had the alert definition page and panel edit in mind right now.


public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -383,7 +300,13 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
       <div className={styles.wrapper} aria-label={selectors.components.PanelEditor.General.content}>
         {this.editorToolbar(styles)}
         <div className={styles.verticalSplitPanesWrapper}>
-          {uiState.isPanelOptionsVisible ? this.renderWithOptionsPane(styles) : this.renderHorizontalSplit(styles)}
+          <ThreePaneSplit
+            horizontalSplitComponents={this.renderPanelAndEditor(styles)}
+            rightPaneComponent={this.renderOptionsPane()}
+            uiState={uiState}
+            updateUiState={updatePanelEditorUIState}
+            rightPanelVisible={uiState.isPanelOptionsVisible}
```
torkelo, line 308:
```suggestion
            rightPaneVisible={uiState.isPanelOptionsVisible}
```

property should be rightPaneVisible

public/app/core/components/ThreePaneSplit/ThreePaneSplit.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+enum Pane {
+  Right,
+  Top,
+}
+
+interface Props {
+  horizontalSplitComponents: React.ReactNode[];
+  rightPaneComponent: React.ReactNode;
+  uiState: { topPaneSize: number | string; rightPaneSize: number | string };
```
torkelo, line 16: looks like we always treat it as a number, when is it string?

torkelo (approved):
Think this looks good, will be really nice to reuse all this!

public/app/core/components/ThreePaneSplit/ThreePaneSplit.tsx
peterholmberg: It takes the topPaneSize and rightPaneSize from this model  https://github.com/grafana/grafana/blob/7d8cb6869a2d3cb2bf3dd54dac2b59e60babc975/public/app/features/dashboard/components/PanelEditor/state/reducers.ts#L20

peterholmberg: but then we don't handle that case in the component :thinking:

PR #29273, opened 2020-11-20
Dashboard: Fixes kiosk state after being redirected to login page and back

### LINKED ISSUE #26899: Kiosk mode doesn't work after oauth redirect
**What happened**:
When using oauth and kiosk mode the kiosk mode doesn't work on first login after redirect.
I think the problem is that kiosk mode only works when kiosk parameter is passed to the end of uri (but after redirect the kiosk parameters appears at the begging of url parameters).

1. Open link with kiosk mode (http://192.168.44.26/grafana/explore?&left=[%22now-1h%22,%22now%22,%22Loki%22,{%22expr%22:%22{stream=%22stdout%22}%22},{%22ui%22:[true,true,true,%22none%22]}]&orgId=1&kiosk)
2. Login with external oauth provider
3. Grafana redirects to http://192.168.44.26/grafana/explore?kiosk=&left=[%22now-1h%22,%22now%22,%22Loki%22,{%22expr%22:%22{stream=%22stdout%22}%22},{%22ui%22:[true,true,true,%22none%22]}]&orgId=1. 
If you try to open http://192.168.44.26/grafana/explore?left=[%22now-1h%22,%22now%22,%22Loki%22,{%22expr%22:%22{stream=%22stdout%22}%22},{%22ui%22:[true,true,true,%22none%22]}]&orgId=1&kiosk everything works fine
4. Kiosk mode doesn't work

torkelo:
Fixes #26899

This fixes a bug in the golang query Values Encode function where it write "=" even for keys that have no value. A query parameter "&kiosk" is treated as a boolean true in the frontend but "&kiosk=" is treated as an empty string by the frontend.

A big hacky fix but since the bug is in the golang standard lib not sure what is best here.

### DIFF
--- a/pkg/middleware/auth_test.go
+++ b/pkg/middleware/auth_test.go
@@ -1,11 +1,13 @@
 package middleware
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/grafana/grafana/pkg/bus"
 	"github.com/grafana/grafana/pkg/models"
 	"github.com/grafana/grafana/pkg/setting"
+	"github.com/stretchr/testify/require"
 
 	. "github.com/smartystreets/goconvey/convey"
 )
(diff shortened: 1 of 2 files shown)


papagian (approved):
I have tested it by enabling anonymous access and signing in (by using the Sign in link from the side menu) with:
- basic auth and
- google OAuth

and everything works fine!

marefr:
@papagian care to take another look. Pushed some additional tests for this

pkg/middleware/auth_test.go
```diff
@@ -104,3 +106,20 @@ func TestMiddlewareAuth(t *testing.T) {
...
+
+func TestRemoveForceLoginparams(t *testing.T) {
+	tcs := []struct {
+		inp string
+		exp string
+	}{
+		{inp: "/?forceLogin=true", exp: "/?"},
+		{inp: "/d/dash/dash-title?ordId=1&forceLogin=true", exp: "/d/dash/dash-title?ordId=1"},
+		{inp: "/?kiosk&forceLogin=true", exp: "/?kiosk"},
+		{inp: "/d/dash/dash-title?ordId=1&kiosk&forceLogin=true", exp: "/d/dash/dash-title?ordId=1&kiosk"},
```
papagian, line 118:
```suggestion
		{inp: "/d/dash/dash-title?ordId=1&kiosk&forceLogin=true", exp: "/d/dash/dash-title?ordId=1&kiosk"},
				{inp: "/d/dash/dash-title?ordId=1&forceLogin=true&kiosk", exp: "/d/dash/dash-title?ordId=1&kiosk"},
		{inp: "/d/dash/dash-title?forceLogin=true&kiosk", exp: "/d/dash/dash-title?&kiosk"},
```

papagian (approved):
Nice; I've added two more cases.

PR #29275, opened 2020-11-20
GraphNG: Init refactorings and fixes

dprokop:
You may encounter rought edges, beware!

TODO:
- [ ] ~comparing data frames structure needs an option for ignoring certain properties. I.e currently modifing min/max results in plot initialization, which should not happen as from uPlot config POV these are ignored~
- [ ] Tests tests tests

* [x] merged with master
* [x] Fixed left & right axis auto placement
* [x] Fixed points auto mode 
* [x] Fixed pass frame min & max to uPlot (has one tricky side effect where frame min & max auto are always across all frames & fields) 
* [x] Fixes time axis after uplot upgrade it switched to millisecond time values
* [x] Fixes time region selection issue due to millisecond time resolution in uPlot now


torkelo:
* merged with master
* Fixed left & right axis auto placement
* Fixed points auto mode 
* Fixed pass frame min & max to uPlot (has one tricky side effect where frame min & max auto are always across all frames & fields)

ryantxu:
> comparing data frames structure needs an option for ignoring certain properties. I.e currently modifing min/max results in plot initialization, which should not happen as from uPlot config POV these are ignored

Why?   Do min/max change often?   Also, I would assume that min/max should set the axis min/max (though will need to expand for the min/max of everything that shares the same axis)

torkelo:
> Why? Do min/max change often? Also, I would assume that min/max should set the axis min/max (though will need to expand for the min/max of everything that shares the same axis)

Well yes, for auto min max they will change every data update. And since min/max is part of axis config, it looks like for this case we need to re-init plot every render for dynamic min/max. 

I have updated this branch so that dataframe min/max does set axis min/max. It's not accurate for different y-axis as Grafana auto min & max are for all fields

dprokop:
> Well yes, for auto min max they will change every data update. And since min/max is part of axis config, it looks like for this case we need to re-init plot every render for dynamic min/max.

This is not necessarily true. We do not set min/max neither for Axis nor for Series and delegate the calculation responsibility to uPlot - which seems not to be correct taking min/max being configurable via field config.

leeoniya:
scale min/max in uPlot is a bit unusual - it is the *current* min/max. these values change as you call setData(), zoom, etc. setting these during init only sets the initial min/max but does not make them persistent.

all persistent min/max configs should be done via `scale.range`.

torkelo:
> like many uPlot opts, scale min/max are a bit unusual - they are the current min/max. these values change as you call setData(), zoom, etc. setting these during init only sets the initial min/max but does not make them persistent.
all persistent min/max configs should be done via scale.range.

Ok, that is going to be a bit tricky, then we need to ignore them in the data frame structure change detection and handle updating them before we call setData

leeoniya:
> Ok, that is going to be a bit tricky, then we need to ignore them in the data frame structure change detection and handle updating them before we call setData

i *think* what you're trying to do just needs to adjust different uPlot opts:

https://github.com/leeoniya/uPlot/blob/406dd367cc6c04f262d3bab66a56c9b30248ba60/demos/soft-minmax.html#L65-L76

dprokop:
> Ok, that is going to be a bit tricky, then we need to ignore them in the data frame structure change detection and handle updating them before we call setData

@torkelo actually it was fairly easy to do, see 7b410bd

torkelo:
Nice

leeoniya:
@dprokop and i talked through the scale min/max adjustment & init. we'll need to revisit this, since the min/max from the dataframe fields will not be too useful for softMin/softMax. we basically need to understand when the min/max config changes due to panel setup vs being computed for the data in the dataframe by grafana at runtime.

ryantxu (approved):
Looks good -- sounds like we will do tests in a new PR

PR #29276, opened 2020-11-20
AlertingNG: (wip/exploritory) Save alert instances

kylebrandt:
Exploring Saving instances to SQL.

Notes:
 - Each instance is uniquely identified by its AlertDefintionID + labels (within an orgID).
 - Instance table will mostly be updates, so need some sort of upsert pattern
 - With multi-instance Grafana, we need to ensure we don't create duplicate entries for alert definition. A state machine will exist on top of this persistence (eventually), so if there are known fundamental issues with this piece, troubleshooting/debuging alerting issues will be very problematic.

To Figure out:

   - [x] Can the string key (current alertId+json searlized labels) key become a primary key in the db to make sure we can't ever create two? **A:** Yes (I was conflating our use primary with fks). Probably better is just to has the labels for the key hashed + the alert def id as a two column unique key.
   - [X] At what level will alert instance updates happen. The software pattern will group updates be alertId, but perhaps those updating those states in larger transactions will result in deadlock issues? **A**: For now keep it simple and do update->insert per instance
   - [X] Can the issues described with upserts in https://sqlperformance.com/2020/09/locking/upsert-anti-pattern be avoided in a way that will work with SQLite, MySQL, and Postgres. **A:** Yes, within a query update, and if not updated, then insert
   - [X] Seems like it will be handcrafted sql? **A:** Probably, SQLite: https://www.sqlite.org/lang_UPSERT.html , MySQL: https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html, Postgres: https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT


pkg/services/ngalert/database.go
```diff
@@ -119,3 +122,65 @@ func (ng *AlertNG) getAlertDefinitions(cmd *listAlertDefinitionsCommand) error {
 		return nil
 	})
 }
+
+// saveAlertDefinition is a handler for saving a new alert definition.
+func (ng *AlertNG) saveAlertInstance(cmd *saveAlertInstanceCommand) error {
```
torkelo, line 127: I know this is early wip, but long term I would want more isolation of database access (ie in its own repository abstraction), to help separate concerns, aid testability etc. 

The main AlertNG should be just the top level service with very knowledge of implementation details.

kylebrandt: @torkelo  Yes, I think once we have the initial version of this and the scheduler merged, we should take a beat and figure out a good pkg layout under ngalert/.

As an aside, the structure under pkg/expr is also not making a sense, so that needs similar thought / attention.

torkelo: ok sounds good. 

I would not wait too long to think about the abstractions we need and different responsibilities & interfaces and dependencies between components

kylebrandt: > I would not wait too long to think about the abstractions we need and different responsibilities & interfaces and dependencies between components

Made an issue https://github.com/grafana/grafana/issues/29386 so we can start to figure this out and write down ideas so we are ready when there is a good point to rename, move stuff around, create interfaces.

pkg/services/ngalert/models.go
```diff
@@ -90,3 +90,45 @@ type listAlertDefinitionsCommand struct {
...
+
+// AlertInstance represent a single alert instance.
+type AlertInstance struct {
+	OrgID             int64 `xorm:"org_id"`
+	AlertDefinitionID int64 `xorm:"alert_definition_id"`
+	Labels            InstanceLabels
+	LabelsHash        string
+	CurrentState      InstanceStateType
+	CurrentStateSince time.Time
+	LastEvalTime      time.Time
```
papagian, line 102: If we use this field for date comparison, maybe it's preferable [to be saved as epoch](https://github.com/grafana/grafana/tree/master/pkg#date-comparison) in the database

kylebrandt:
close in favor of #30223

PR #29279, opened 2020-11-20
DataSourceSettings: Add servername field to DataSource TLS config

iplay88keys:
DataSourceSettings: Add servername field to DataSource TLS config

A DNS lookup URL can be provided in the DataSource URL field in order
to dynamically load balance between multiple instances of a DataSource.
When using mutual TLS, Golang's TLS config implementation checks that
the certificate's common name (< 1.15) or subject alternative name
(>= 1.15) has the same value as the domain being accessed. If the DNS
entry is dynamically generated for a specific environment, the
certificate cannot be generated with a name matching the dynamic
DNS URL. As such, Golang offers a servername field that can be set to
overwrite what value is used when checking against the certificate's
common name (or subject alternative name). Without this change, Skip
TLS Verify must be set to true in order for the DataSource to work,
removing some of the benefits gained by using mutual TLS.

This commit adds the ability to set Grafana's internal Golang TLS config
servername field from the UI or a provisioned DataSource.

The servername field is optional and the existing behavior is retained if
the field is not set.

Co-authored-by: Dana Pruitt <dpruitt@vmware.com>
Co-authored-by: Jeremy Alvis <jalvis@pivotal.io>


iplay88keys:
It seems that Golang 1.15 would change the check to be against the "Subject Alternative Name" on the certificate, but this change still applies in that case. https://golang.org/doc/go1.15#commonname

iplay88keys:
This should be good to go at this point. I needed to update some files and the commit message based on the checks as well as spelling errors, hence the multiple updates.

docs/sources/administration/provisioning.md
```diff
@@ -146,6 +146,7 @@ Since not all datasources have the same configuration settings we only have the
 | tlsAuth                 | boolean | _All_                                                            | Enable TLS authentication using client cert configured in secure json data                  |
 | tlsAuthWithCACert       | boolean | _All_                                                            | Enable TLS authentication using CA cert                                                     |
 | tlsSkipVerify           | boolean | _All_                                                            | Controls whether a client verifies the server's certificate chain and host name.            |
+| servername              | string  | _All_                                                            | Optional. Controls the server name used for certificate common name/subject alternative name verification. Empty defaults to using the datasource url. |
```
oddlittlebird, line 149:
```suggestion
| servername              | string  | _All_                                                            | Optional. Controls the server name used for certificate common name/subject alternative name verification. Empty defaults to using the data source URL. |
```

docs/sources/administration/provisioning.md
```diff
@@ -146,6 +146,7 @@ Since not all datasources have the same configuration settings we only have the
...
 | graphiteVersion         | string  | Graphite                                                         | Graphite version                                                                            |
 | timeInterval            | string  | Prometheus, Elasticsearch, InfluxDB, MySQL, PostgreSQL and MSSQL  | Lowest interval/step value that should be used for this data source                         |
```
oddlittlebird, line 151:
```suggestion
| timeInterval            | string  | Prometheus, Elasticsearch, InfluxDB, MySQL, PostgreSQL, and MSSQL  | Lowest interval/step value that should be used for this data source                         |
```

iplay88keys:
I made both changes from above as well as fixed the `servername` to correctly be `serverName` in the config file.

pkg/models/datasource_cache_test.go
```diff
@@ -129,16 +129,21 @@ func TestDataSourceProxyCache(t *testing.T) {
 
 		json := simplejson.New()
 		json.Set("tlsAuthWithCACert", true)
+		json.Set("serverName", "server-name")
 
 		tlsCaCert, err := util.Encrypt([]byte(caCert), "password")
 		So(err, ShouldBeNil)
 
+		So(err, ShouldBeNil)
```
aknuds1, line 137: Looks to me as if this error check is left over?

```suggestion
```

docs/sources/administration/provisioning.md
```diff
@@ -146,8 +146,10 @@ Since not all datasources have the same configuration settings we only have the
 | tlsAuth                 | boolean | _All_                                                            | Enable TLS authentication using client cert configured in secure json data                  |
 | tlsAuthWithCACert       | boolean | _All_                                                            | Enable TLS authentication using CA cert                                                     |
 | tlsSkipVerify           | boolean | _All_                                                            | Controls whether a client verifies the server's certificate chain and host name.            |
+| serverName              | string  | _All_                                                            | Optional. Controls the server name used for certificate common name/subject alternative name verification. Empty defaults to using the data
+source URL. |
```
aknuds1, line 150: Could it be written "Defaults to using the data source URL" instead? Would make more sense to me than how it's currently written at least.

aknuds1 (changes requested):
Backend part looks OK to me, except there's what's looks like a stray error check in the test code.

pkg/models/datasource_cache_test.go
iplay88keys: Not sure how that happened. I'll fix and push!

aknuds1 (approved):
LGTM from a backend perspective, but we have to hear from the frontend people.

aknuds1:
There are lots of conflicts, is the branch perhaps wrongly based?

aknuds1:
I think you'll have to rebase your branch, something's happened with your latest push AFACT, GitHub showing tons of changes.

iplay88keys:
Hmm, ya, I see that. I just ran `git rebase origin/master` followed by `git pull`. I haven't done many long-running PRs, so maybe I did something wrong there.

aknuds1:
I would just normally merge with our master, instead of rebasing, since the latter can cause havoc (due to rewriting history), and it's also bad to force push to PRs since f.ex. you can lose review comments. You might also overwrite other people's changes, so as a rule of thumb don't force push to PRs (not in our repos at least).

iplay88keys:
Ah, I see. Ok, would you like me to revert back to the prior commit I made and then merge master before force-pushing this time?

aknuds1:
I suspect a rebase (followed by a force-push) is the only way out of this state. I can't think of any other way to resolve it at least.

aknuds1:
Can you fix the merge conflict?

iplay88keys:
That should be good to go now. Sorry for the difficulties. Thanks for bearing with me as I figure this out.

aknuds1:
No worries!

aknuds1 (approved):

achatterjee-grafana (approved):
Doc content looks good.

peterholmberg (changes requested):
Just some minor things. You need to add a serverName property [here](https://github.com/grafana/grafana/blob/56e7ce6f16bca06ea0db1fb521a6f37250c10fb6/packages/grafana-ui/src/components/DataSourceSettings/DataSourceHttpSettings.story.tsx#L27).

packages/grafana-ui/src/components/DataSourceSettings/TLSAuthSettings.tsx
```diff
@@ -31,6 +33,16 @@ export const TLSAuthSettings: React.FC<HttpSettingsBaseProps> = ({ dataSourceCon
     });
   };
 
+  const onServerNameLabelChange = (event: React.SyntheticEvent<HTMLInputElement>) => {
+    const newJsonData = { ...dataSourceConfig.jsonData };
+    newJsonData['serverName'] = event.currentTarget.value;
```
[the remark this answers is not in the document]
iplay88keys: Will do!

peterholmberg (approved):

iplay88keys:
Thanks for merging! 👍

Issue #29133, opened 2020-11-16
Panel Override: use capture groupe results from regex matcher in Display Name field.


tarag:
**What would you like to be added**:

When using a regex matcher in Panel Overrides, I would like to use the result of capture groups inside the regex to provide the value for replacement of the field.

E.g:

<img width="397" alt="Screenshot 2020-11-16 at 08 34 49" src="https://user-images.githubusercontent.com/11686092/99224519-c53cf580-27e6-11eb-8c4e-ee767e690300.png">

**Why is this needed**:

When graphing numerous series that follow the same naming pattern, this allows displaying human friendly name with a single override instead of having to define one per series.

It seems that there are some open issues to be able to do that using aliasing for various datasource, but this could profited a unified way to do it: #5212

torkelo:
This is probably better done in transform

tarag:
I can't use transform for the queries where this feature would be useful because I get this message: _"Organize fields only works with a single frame. Consider applying a join transformation first."_

If I perform a join, then I get one line for each series so I am back to specify the name for each one by one.

This is what the query looks like (InfluxDB): `SELECT mean("value") FROM /^F(1|2|3|4|5)_Appt_($Zone)_Zone_(Principale|Secondaire)_TH_TEMP/ WHERE $timeFilter GROUP BY time($__interval) fill(previous)`

torkelo:
I mean a new regex rename transform

PR #29281, opened 2020-11-20
Transformations: Add "Rename By Regex" transformer

### LINKED ISSUE #5212: Regex support in elasticsearch metrics alias pattern
Metrics field are often longer than wished in a legend.
Example: topbeat ==> "cpu.idle", "cpu.system", "cpu.user", "cpu.nice" 

a regex in alias pattern (like {{regex /cpu.(.*)/}} ) would allow more readible series names (here, only "idle", "user", "system"). This is especially powerful if we can chain it with {{term fieldname}} so as to transform terms values !!!

simianhacker:

This PR allows users to rename metrics by using a regular expression and replacement pattern via a transformer. This is useful for when you're doing a terms aggregation in something like Elasticsearch and you want to modify the values of the terms. For example, let's say you have a terms agg on `host.name` your team is using a pattern like `web-{number}.example.com` and you only want to display `web-{number}` in the results. You would set the `Match` field to `([^.]+).example.com` and the `Replace` field. to `$1`; `web-01.example.com` would be come `web-01`.

In the screenshot below, this transformer is removing the first part of the `event.dataset` and only displaying the suffix, `system.cpu` to `cpu`.

image

This was originally requested in #5212 and subsequently requested in #29133. @torkelo mentioned that this should be done as a transformer in #29133.

Fixes #5212

Let me know if there is a better name then `Rename by regex` or if we need to change the form layout.


simianhacker:
Is there a way to get access to the video output of the E2E tests on CI? These tests pass locally when I run `yarn e2e` but I'm unable to get the same tests to pass in CI. If you look at the commit history, I had to add `scrollIntoView()` for the transform panel to get it to pass locally but if I remove that, it fails.

hugohaggmark:
> Is there a way to get access to the video output of the E2E tests on CI? These tests pass locally when I run `yarn e2e` but I'm unable to get the same tests to pass in CI. If you look at the commit history, I had to add `scrollIntoView()` for the transform panel to get it to pass locally but if I remove that, it fails.

Hi @simianhacker we used to, but we've recently switched to Drone, so right now it isn't possible. I'll checkout your branch and have a look to see if I can figure out what is going on.
(ping @aknuds1)

hugohaggmark:
@simianhacker I pushed some changes to the e2e tests that should fix them. I hope that is Ok.

simianhacker:
Totally OK! Thank you for the help!

hugohaggmark (commented):
I think this PR is off to a brilliant start, but I hesitate to approve it because this transformer only operates on the name of frames. I think we need to incorporate renaming of field names and labels too in order to make it more intuitive for all use cases.

simianhacker:
@hugohaggmark Thanks for the feedback! I will apply it to fields and labels as well similar to `Filter by name`.

simianhacker:
@hugohaggmark I think the changes I just added should work as you suggested. I read through the `getDisplayName()` code and it appears to handle the labels as well, let me know if I've misinterpreted how that works.

mckn (changes requested):
I think this looks great! Well done.

However I am missing some documentation for this. It would be awesome if you can add some information about how to use this with some examples here: https://github.com/grafana/grafana/blob/master/docs/sources/panels/transformations/types-options.md

Also - I think adding support for selecting at what level of the data frame you want to replace the naming e.g. so you can select either field, frame or labels. But I think we can merge this PR and add support for the other types in follow up PRs.

simianhacker:
> Also - I think adding support for selecting at what level of the data frame you want to replace the naming e.g. so you can select either field, frame or labels. But I think we can merge this PR and add support for the other types in follow up PRs.

@mckn In most of the other transformers, they seem to limit the functions to just the field, frame and labels without specifying which one to apply it to; are there plans to add that functionality to those transformers? 

@hugohaggmark What's the harm in applying this to the values as well? For example, one might want to modify the values returned from a term's aggregation in a data table, like below:

image

to something like this

image

Or should that be a seperate transformer? "Replace values by regex" maybe? I put together a quick function that could easily accomplish this as well (for the screenshot above).

```typescript
const replaceValuesByRegex = (options: RenameByRegexTransformerOptions) => (frame: DataFrame) => {
  const regex = new RegExp(options.regex);
  const fields = frame.fields.map(field => {
    if (field.type !== 'string') {
      return field;
    }
    const values = [];
    for (let index = 0; index < field.values.length; index++) {
      const value = field.values.get(index);
      if (regex.test(value)) {
        values.push(value.replace(regex, options.renamePattern));
      } else {
        values.push(value);
      }
    }
    return { ...field, values: new ArrayVector(values) };
  });
  return { ...frame, fields };
};
```

@mckn I started working on the documentation today but I wanted to resolve these questions before I finalize it.

hugohaggmark:
@simianhacker we've recently added a transformer that filters by value, so I think adding a new one that renames by value should be ok. But I would wait a bit with that one, maybe create a feature request to gather more feedback from the community. WDYT?

simianhacker:
@hugohaggmark Sounds good, I will get the docs done and work on wrapping this one up.

simianhacker:
@hugohaggmark I added some documentation, writing is not my forté so please feel free to suggest some edits. I also created some images for the documentation but I couldn't figure out where those images are stored in the repo.

hugohaggmark:
> @hugohaggmark I added some documentation, writing is not my forté so please feel free to suggest some edits. I also created some images for the documentation but I couldn't figure out where those images are stored in the repo.

Maybe @oddlittlebird can help with the images as I really don't know how to either?

achatterjee-grafana:
@hugohaggmark, images are kept in the website repo. Diana, how will we add the images in? Thanks.

oddlittlebird:
The images are stored in a private repo. If you or Hugo can take some screenshots and then attach them here (please label them with what they are and where they should go, then a Grafana maintainer can add them to the private website repo in a separate PR.

oddlittlebird:
The addition to the docs looks good to me. Thanks for including examples!

simianhacker:
@oddlittlebird Here are the images

#### rename-by-regex-before.png

rename-by-regex-before

#### rename-by-regex-after.png

rename-by-regex-after

osg-grafana:
@oddlittlebird What are the guidelines for image size, width, height, and file type? I expect that we can address such questions outside of this particular PR, and this is the context in which I thought to ask this question.

docs/sources/panels/transformations/types-options.md
```diff
@@ -336,3 +337,18 @@ When you have more than one condition, you can choose if you want the action (in
...
+
+You can specify a regular expression, which is only applied to matches, along with a replacement pattern that support back references. For example, let's imagine you're visualizing CPU usage per host and you want to remove the domain name. You could set the regex to `([^\.]+)\..+` and the replacement pattern to `$1`, `web-01.example.com` would become `web-01`.
+
+Here is an example where we are stripping the prefix from event types, in the before you can see everything is prefixed with `system.`
+
+{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-before.png" class="docs-image--no-shadow" max-width= "1100px" >}}
+
+With the transformation applied, you can see we are left with just the remainder of the string.
+
+{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-after.png" class="docs-image--no-shadow" max-width= "1100px" >}}
```
oddlittlebird, line 353:
```suggestion
{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-after-7-3.png" class="docs-image--no-shadow" max-width= "1100px" >}}
```

docs/sources/panels/transformations/types-options.md
```diff
@@ -336,3 +337,18 @@ When you have more than one condition, you can choose if you want the action (in
...
+
+## Rename by regex
+
+Use this transformation to rename parts of the query results using a regular expression and replacement pattern.
+
+You can specify a regular expression, which is only applied to matches, along with a replacement pattern that support back references. For example, let's imagine you're visualizing CPU usage per host and you want to remove the domain name. You could set the regex to `([^\.]+)\..+` and the replacement pattern to `$1`, `web-01.example.com` would become `web-01`.
+
+Here is an example where we are stripping the prefix from event types, in the before you can see everything is prefixed with `system.`
+
+{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-before.png" class="docs-image--no-shadow" max-width= "1100px" >}}
```
oddlittlebird, line 349:
```suggestion
{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-before-7-3.png" class="docs-image--no-shadow" max-width= "1100px" >}}
```

oddlittlebird (commented):
rename-by-regex-before-7-3.png and rename-by-regex-after-7.3.png have been merged to the website repo.

oddlittlebird:
> @oddlittlebird What are the guidelines for image size, width, height, and file type? I expect that we can address such questions outside of this particular PR, and this is the context in which I thought to ask this question.

Image guidelines are in the [Grafana documentation style guide](https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-style-guide.md#images).

docs/sources/panels/transformations/types-options.md
```diff
@@ -336,3 +337,17 @@ When you have more than one condition, you can choose if you want the action (in
...
 
 Conditions that are invalid or incompletely configured are ignored.
+
+## Rename by regex
+
+Use this transformation to rename parts of the query results using a regular expression and replacement pattern.
+
+You can specify a regular expression, which is only applied to matches, along with a replacement pattern that support back references. For example, let's imagine you're visualizing CPU usage per host and you want to remove the domain name. You could set the regex to `([^\.]+)\..+` and the replacement pattern to `$1`, `web-01.example.com` would become `web-01`.
+
+Here is an example where we are stripping the prefix from event types, in the before you can see everything is prefixed with `system.`
```
achatterjee-grafana, line 347:
```suggestion
In the following example, we are stripping the prefix from event types. In the before image, you can see everything is prefixed with `system.`
```

docs/sources/panels/transformations/types-options.md
```diff
@@ -336,3 +337,17 @@ When you have more than one condition, you can choose if you want the action (in
...
+
+{{< docs-imagebox img="/img/docs/transformations/rename-by-regex-before-7-3.png" class="docs-image--no-shadow" max-width= "1100px" >}}
```
achatterjee-grafana, line 349: Why are the images named before-7-3? If we are adding new images, should we capture them from the latest build?

achatterjee-grafana (commented):
Added a few comments.

mckn (approved):
Awesome job on this!

charlescva:
Woot! 4.5 years after asking for this in issue #5212 and it's here! Great job!

oddlittlebird:
@mckn , is this going to be in 7.4? If so, then we need to add a note to the docs indicating that immediately. It will cause user confusion otherwise.

mckn:
> @mckn , is this going to be in 7.4? If so, then we need to add a note to the docs indicating that immediately. It will cause user confusion otherwise.

Yes, added in this PR: #29735. Great catch!

dicer:
> @simianhacker we've recently added a transformer that filters by value, so I think adding a new one that renames by value should be ok. But I would wait a bit with that one, maybe create a feature request to gather more feedback from the community. WDYT?

Is there a feature request or PR for this yet? Would this allow me to rename all kinds of values? I have some queries that return strings like "closed" and I would like to rename them to "0" (and "open" to "1"), so I can graph them.

simianhacker:
> Is there a feature request or PR for this yet? Would this allow me to rename all kinds of values? I have some queries that return strings like "closed" and I would like to rename them to "0" (and "open" to "1"), so I can graph them.

@dicer There isn't a feature request or PR, I had added it as part of this change but then pulled it out per the conversation above. If we implemented this for values, it would give you all the power of a find/replace regular expression. You could create one transformer that replaced all the zeros with `closed` and a second one that replaces all the ones with `open`.

oddlittlebird:
Folks, please open a new issue rather than having discussions on merged PRs.

PR #29288, opened 2020-11-22
GraphNG: support x != time from the UI

ryantxu:
This explores how we can support x != time, and starts to explore a more generic dimension mapping approach

image

Test with "random walk table" and pick min vs max field


leeoniya:
probably hold off on getting too deep into this until proper scatter support lands in uPlot: https://github.com/leeoniya/uPlot/issues/107

torkelo:
nice! exciting to get to this! 

would have been good to base it on https://github.com/grafana/grafana/pull/29275, going to be messy to merge. Been fixing a couple of graph issues in that PR as well (time axis, points mode, pass min & max)

ryantxu:
The interesting bits are all outside GraphNG/react, so I am not too worried about the merge.

leeoniya:
thinking a bit beyond basic x/y scatter...would something like below (multi-series bubble chart) be possible to extract from a single DataFrame? you'd probably need to do something like a field-chunker (take every 4 fields).

```js
[
  [  // series 1
     [1,2,3],  // x
     [10,20,30],  // y
     [2.2,1.5,6.5], // v
     ["a","b","c"],  // l
  ],
  [  // series 2
     [1,2,3],  // x
     [10,20,30],  // y
     [2.2,1.5,6.5], // v
     ["a","b","c"],  // l
  ]
]
```

ryantxu:
> possible to extract from a single DataFrame?

um... we start with `DataFrame[]` that has everything you need :)

packages/grafana-data/src/transformations/matchers.ts
```diff
@@ -30,16 +32,10 @@ export const frameMatchers = new Registry<FrameMatcherInfo>(() => {
 
 export function getFieldMatcher(config: MatcherConfig): FieldMatcher {
   const info = fieldMatchers.get(config.id);
-  if (!info) {
```
ryantxu, line 33: not required since the `.get` above will throw an error

ryantxu:
closed by #30096

PR #29299, opened 2020-11-23
Login: Fixes redirect url encoding issues of # %23 being unencoded after login

### LINKED ISSUE #29237: Login page redirect loses URL encoding with # character in them
**What happened**:
If a template variable value has a hash character in it, and the URL is used to login via a bookmark or sharing with some, the variable gets truncated after the login, so the link no longer works. 

the redirect not to change the variable value.

- test instance, new dashboard, random walk
- add a variable, test, with customer values item1,item_hash#posthash, item3
- set the variable to the hash item.
- copy the URL
- http://localhost:3000/d/rhEFS-TGz/hash-test-copy?orgId=1&var-test=item_hash%23posthash
- paste it into a new private browsing session.
- after login, the URL is
- http://localhost:3000/d/rhEFS-TGz/hash-test-copy?orgId=1&var-test=item_hash#

**Environment**:
- Grafana version: 7.3.3
- OS Grafana is installed on: n/a 
- User OS & Browser: Windows & firefox, but probably n/a
- Grafana plugins: n/a
- Others:

torkelo:
Fixes #29237

This was also a backend issue caused by double url decoding the cookie value. The maracon function c.GetCookie does Url decoding internally so we double decoded the value

### DIFF
--- a/pkg/api/login.go
+++ b/pkg/api/login.go
@@ -121,7 +121,7 @@ func (hs *HTTPServer) LoginView(c *models.ReqContext) {
 			}
 		}
 
-		if redirectTo, _ := url.QueryUnescape(c.GetCookie("redirect_to")); len(redirectTo) > 0 {
+		if redirectTo := c.GetCookie("redirect_to"); len(redirectTo) > 0 {
 			if err := hs.ValidateRedirectTo(redirectTo); err != nil {
 				// the user is already logged so instead of rendering the login page with error
 				// it should be redirected to the home page.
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):

aknuds1:
Shouldn't the milestone be 7.4 on this, and 7.3.4 on the backport PR? Or did I misunderstand the scheme?

torkelo:
@aknuds1 no, we decided that both backport PR and main PR should be in patch release milestone, think we have been over this a couple of times :) 

I was not sure this is correct but most felt this makes other things easier and clearer, and the bot can just sync milestone and makes the intent clear that this is aimed to be backported

aknuds1:
> @aknuds1 no, we decided that both backport PR and main PR should be in patch release milestone, think we have been over this a couple of times :)

In that case I must've misunderstood, or was this decision reached after? Not sure.

papagian (approved):

PR #29305, opened 2020-11-23
AlertingNG: Create a scheduler to evaluate alert definitions

### LINKED ISSUE #28994: AlertingNG: Create a scheduler to evaluate alert definitions
https://github.com/grafana/grafana/pull/28377 will add alert definitions to be saved.

A logical next piece is to create the scheduler to run alert definitions. For now, it can just output "firing" alerts to the log or something, as we do not have an instance state tracker for saving yet.

Some examples:
 - https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/sched/alertRunner.go#L13
 - https://github.com/grafana/grafana/blob/master/pkg/services/alerting/engine.go#L48

Spec a vauge in this area in https://docs.google.com/document/d/1kJvE5hMvrs7kWvpxH7lglSI6cjRR5z8LBuaMNpCL9Zk/edit#heading=h.wozk1kxoe2eo and https://docs.google.com/document/d/1kJvE5hMvrs7kWvpxH7lglSI6cjRR5z8LBuaMNpCL9Zk/edit#heading=h.rdfynijg6veu - but can start development of a simpler scheduler for the evaluator in tandem with this.

papagian:
The main difference with the old scheduler is that we start a routine for each alert definition that runs all the evaluations rather than having a routine for each evaluation.
It also differs from the bosun scheduler in that alert definitions can be updated/deleted between intervals so the scheduler has to take care of it.

The schedule supports:

- [X] retry on evaluation failures
- [X] ability to [pause/unpause](https://github.com/grafana/grafana/pull/29305#issuecomment-738188020) the scheduler via the API using:
```
 curl -XPOST  http://admin:admin@localhost:3000/api/ngalert/pause
 curl -XPOST  http://admin:admin@localhost:3000/api/ngalert/unpause
```
- [X] alert definition versioning and optimisation of the scheduler to fetch alert definitions only after changes

Fixes #28994

This PR includes the following modifications:
- ~`DatasourceCache.GetDatasource()` expects organisation identifier instead of `SignedInUser` (that was used for extracting the organisation identifier from the signed in user)~ (this change was reverted because enterprise excepts the `SignedInUser`)
- parameterise `alerting.Ticker` to run on custom interval and add support for pausing/unpausing

For running all the tests locally:
`GRAFANA_TEST_DB=<database_type>  go test -tags=integration -v ./pkg/services/ngalert/...`

**Updated description**

Database tables `alert_definition` and `alert_definition_version` **are dropped** and recreated.

`name` column is renamed to `title` and `interval` column is renamed to `interval_seconds`. The respected JSON fields should change to `title` and `interval_seconds`.
Also `updated` and `created` columns are `migrator.DB_DateTime` and the respective model fields `AlertDefintion.Updated` and `AlertDefinitionVersion.Created` have changed to `time.Time`.


pkg/services/ngalert/schedule.go
```diff
@@ -0,0 +1,215 @@
...
+			if id == definitionID {
+				// TODO what if it's running
+				return
+			}
+		}
+	}
+}
+
+func (ng *AlertNG) fetchAlertDefinitions(since time.Time) []*AlertDefinition {
+	cmd := listAlertDefinitionsCommand{}
```
torkelo, line 50: A command? list feels like a query, or does it modify data?

pkg/services/ngalert/schedule.go
```diff
@@ -0,0 +1,215 @@
...
+
+			step := int(ng.schedule.baseInterval.Nanoseconds()) / len(readyToRun)
+
+			// send loop is only required for distribute evaluations across time within an interval
+			for _, item := range readyToRun {
+				item.definitionCh.ch <- &evalContext{now: tick}
+
+				delete(registeredDefinitions, item.id)
+
+				time.Sleep(time.Duration(step))
```
torkelo, line 126: Using thread sleep feels like not the optimal way to distribute evaluations. 

the old scheduled seperated the scheduler tick, from actually scheduling jobs. Maybe that would help?
https://github.com/grafana/grafana/blob/master/pkg/services/alerting/scheduler.go

pkg/services/ngalert/schedule.go
```diff
@@ -0,0 +1,215 @@
+package ngalert
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+func (ng *AlertNG) definitionRoutine(definitionID int64, evalCh <-chan *evalContext) {
```
torkelo, line 9: I would prefer of the Scheduler was a separate component/service and not part of the main AlertNG service.

pkg/services/ngalert/database.go
```diff
@@ -119,3 +119,16 @@ func (ng *AlertNG) getAlertDefinitions(cmd *listAlertDefinitionsCommand) error {
 		return nil
 	})
 }
+
+func (ng *AlertNG) getAlertDefinitions(cmd *listAlertDefinitionsCommand) error {
```
torkelo, line 123: This is a query not a command, the idea behind the Query and Command pattern is to really make clear what actions change data want are just reads

pkg/services/ngalert/ngalert.go
```diff
@@ -32,6 +36,12 @@ func (ng *AlertNG) Init() error {
 
 	ng.registerAPIEndpoints()
 
+	ng.schedule = schedule{
+		baseInterval: time.Second,
+		channelMap:   channelMap{definionCh: make(map[int64]definitionCh)},
+		stop:         make(chan int64),
+	}
+	ng.schedule.ctx, ng.schedule.cancelFunc = context.WithCancel(context.Background())
 	return nil
```
torkelo, line 45: In the old alert engine we create an errgroup based on the main server Context 

https://github.com/grafana/grafana/blob/master/pkg/services/alerting/engine.go#L61 

So that once the server shuts down nested groups/routines get's shut down, not sure if that is better

torkelo (commented):
Nice progress! 

What's the main difference between this scheduler a the old scheduler? any way to reuse or modify the old scheduler?

Would prefer more structure (seperate services/components & interfaces).

papagian:
> Nice progress!
> 
> What's the main difference between this scheduler a the old scheduler? any way to reuse or modify the old scheduler?
> 
> Would prefer more structure (seperate services/components & interfaces).

thank you for the useful feedback!
the main difference with the old scheduler is that we start a routine for each alert definition that runs all the evaluations rather having a routine for each evaluation. From the spec:
> This is so problems with checks, such as a long running queries, are less likely to impact other the behavior of alert definitions as they are executed. It should also ensure the sequential checks for the same alert definition to run on top of each other if they run longer than the interval between checks. 
Therefore I can't reuse the same scheduler but I can try to based the new one as close to the old one as possible.

torkelo:
> the main difference with the old scheduler is that we start a routine for each alert definition that runs all the evaluations rather having a routine for each evaluation. From the spec:

Ah I see that's what the map of channels are for.

pkg/services/ngalert/schedule.go
kylebrandt: I've seen some variations around on the idea of "Clock", to make sure ticker can be mocked. For example https://godoc.org/github.com/facebookgo/clock

Some discussion around something like this in the time package https://github.com/golang/go/issues/8869, blog post at https://dmitryfrank.com/articles/mocking_time_in_go , using something like this might help us with the testing here.

kylebrandt: Oh actually, or current alerting is using a clock https://github.com/grafana/grafana/blob/master/pkg/services/alerting/ticker.go "github.com/benbjohnson/clock"

kylebrandt:
I think we will need a way to pause / shut-off the scheduler early. I expect people playing playing with the prototype will want to:
 - Play with creating alert definitions
 - See what happens when those alert definitions are running in the background
 - Be able to play with alert definitions, but without have queries running against their data sources in the background. In short - like current "pause all alerts".

Without this they will have to delete the definitions to stop them. Could also do this on a per definition level and call it on all definitions (which I imagine we will eventually need), but for now maybe simpler to start and stop it all together.

pkg/services/ngalert/models.go
```diff
@@ -15,6 +14,9 @@ type AlertDefinition struct {
 	Name      string
 	Condition string
 	Data      []eval.AlertQuery
+	Updated   int64
```
kylebrandt, line 17: So it seems this can be time.Time (if you keep it). Did some xorm digging/debug tracing. If the db type is a BIGINT, and the go type is time.Time, xorm calls `t.Unix()`.

pkg/services/ngalert/schedule.go
```diff
@@ -0,0 +1,292 @@
...
+				// remove the alert definition from the registered alert definitions
+				delete(registeredDefinitions, itemID)
+			}
+
+			var step int64 = 0
+			if len(readyToRun) > 0 {
+				step = ng.schedule.baseInterval.Nanoseconds() / int64(len(readyToRun))
+			}
+
+			fmt.Println(">>>> step:", step)
```
kylebrandt, line 204: line to cleanup eventually

pkg/services/ngalert/api.go
```diff
@@ -148,36 +111,48 @@ func (ng *AlertNG) deleteAlertDefinitionEndpoint(c *models.ReqContext) api.Respo
 // updateAlertDefinitionEndpoint handles PUT /api/alert-definitions/:alertDefinitionId.
 func (ng *AlertNG) updateAlertDefinitionEndpoint(c *models.ReqContext, cmd updateAlertDefinitionCommand) api.Response {
 	cmd.ID = c.ParamsInt64(":alertDefinitionId")
```
kylebrandt, line 113: Need to set the OrgID from signed in user here. Currently going through as zero.

pkg/services/datasources/cache.go
```diff
@@ -33,22 +33,22 @@ func (dc *CacheServiceImpl) Init() error {
...
+			if ds.OrgId == orgID {
 				return ds, nil
 			}
 		}
 	}
 
-	plog.Debug("Querying for data source via SQL store", "id", datasourceID, "orgId", user.OrgId)
-	ds, err := dc.SQLStore.GetDataSourceByID(datasourceID, user.OrgId)
+	plog.Debug("Querying for data source via SQL store", "id", datasourceID, "orgId", orgID)
+	ds, err := dc.SQLStore.GetDataSourceByID(datasourceID, orgID)
```
kylebrandt, line 51: Best to error if orgID is zero, because then there will be no orgid filter on the query.

pkg/services/ngalert/api.go
kylebrandt: Also needed on 84 above I think in getAlertDefinitionEndpoint

papagian: I think it's better to fix this inside `updateAlertDefinition()` by setting OrgID similar to what we do for the rest fields.
https://github.com/grafana/grafana/blob/ec7943eed942bc888c737b45c487ffc0c850b586/pkg/services/ngalert/database.go#L112-L117

If the `cmd.OrgID` is zero then[ the specific field is not updated](https://godoc.org/github.com/go-xorm/xorm#Engine.Update).  WDYT?

pkg/services/datasources/cache.go
papagian: I think that in that case it will fail anyway since it falls to [this](https://github.com/grafana/grafana/blob/ab078133bfe9771f06bd21712f7ad6f9113c8a61/pkg/services/sqlstore/datasource.go#L38) case.

kylebrandt: It will pass. If the `OrgID` is zero, since the Type of the property is an int64, and the zero value 0, then xorm's engine.Get will not add the 'WHERE org_id=` filter to the query, so it will fetch based on the data source id alone.

pkg/services/ngalert/api.go
kylebrandt: It makes sense that a user can not change the OrgID of an alert def (I think). But we do want to make sure that they can't update or get an alert def that is not in their organization.

kylebrandt (approved):
:tada:

pkg/services/ngalert/eval/eval.go
```diff
@@ -207,3 +207,22 @@ func (evalResults Results) AsDataFrame() data.Frame {
 	f := data.NewFrame("", fields...)
 	return *f
 }
+
+// ConditionEval executes conditions and evaluates the result.
+func ConditionEval(condition *Condition, now time.Time) (Results, error) {
+	alertCtx, cancelFn := context.WithTimeout(context.Background(), setting.AlertingEvaluationTimeout)
```
marefr, line 213: Should use non-global configuration. Something to keep in mind moving forward, but fine for now I guess

pkg/services/ngalert/database_mig.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
+)
+
+func addAlertDefinitionMigrations(mg *migrator.Migrator) {
+	alertDefinition := migrator.Table{
+		Name: "alert_definition",
+		Columns: []*migrator.Column{
+			{Name: "id", Type: migrator.DB_BigInt, IsPrimaryKey: true, IsAutoIncrement: true},
+			{Name: "org_id", Type: migrator.DB_BigInt, Nullable: false},
+			{Name: "name", Type: migrator.DB_NVarchar, Length: 255, Nullable: false},
```
marefr, line 15: Don't see this being unique. Have you considered a __UID__? Will be hard to handle alert rules in multiple environments/provisioning otherwise

pkg/services/datasources/cache.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type CacheService interface {
-	GetDatasource(datasourceID int64, user *models.SignedInUser, skipCache bool) (*models.DataSource, error)
+	GetDatasource(datasourceID int64, orgId int64, skipCache bool) (*models.DataSource, error)
```
marefr, line 14: This change will break Enterprise

pkg/services/ngalert/api.go
```diff
@@ -148,36 +111,49 @@ func (ng *AlertNG) deleteAlertDefinitionEndpoint(c *models.ReqContext) api.Respo
...
-	cmd.SignedInUser = c.SignedInUser
-	cmd.SkipCache = c.SkipCache
+	cmd.OrgID = c.SignedInUser.OrgId
 
 	if err := ng.updateAlertDefinition(&cmd); err != nil {
 		return api.Error(500, "Failed to update alert definition", err)
 	}
 
-	return api.JSON(200, util.DynMap{"affectedRows": cmd.RowsAffected, "id": cmd.Result.Id})
+	return api.JSON(200, util.DynMap{"affectedRows": cmd.RowsAffected, "id": cmd.Result.ID})
```
marefr, line 120: Nit nothing new. A bit weird to return `affectedRows`. Same applies for delete alert definition. At least if it's zero or greater than 1 something is wrong and should be treated as an error or something

marefr (commented):
Added some comments. Biggest thing is change of data source cache contract that will break enterprise, but already talked to Sofia regarding that. 

Ticker shouldn't introduce any breaking changes as far as I'm concerned.

pkg/services/ngalert/eval/eval.go
papagian: good point; I'm not even sure if we want to use the same configuration with the existing alerting. We use some other [constants](https://github.com/grafana/grafana/blob/f45449fae6b68e3ab5351ccf79c3c1ef8c5b9890/pkg/services/ngalert/ngalert.go#L20) that probably we would also like to customise in the future.

pkg/services/ngalert/database_mig.go
papagian: We have a separate [issue](https://github.com/grafana/grafana/issues/29339) for adding UIDs but it makes sense also to change the name to be unique as well. @kylebrandt WDYT?

pkg/services/datasources/cache.go
papagian: I think I can revert it, thanks for noticing!

pkg/services/ngalert/api.go
papagian: Zero is not exactly an error: it means that the alert definition does not exists, so in that case I could handled it and return 404 status. Do you think that this would be better?

papagian: I realised that the zero case is already covered by the `validateOrgAlertDefinition` middleware [check](https://github.com/grafana/grafana/blob/a01ee33da70c71f7c32ea56da8a6526d9f6abf3c/pkg%2Fservices%2Fngalert%2Fmiddleware.go#L12) so it should be impossible the `RowsAffected ` to be something other than one. Therefore I have only added a [warning](https://github.com/grafana/grafana/blob/2c2a0479704fbc9ac43e6d4523690a262daef36c/pkg/services/ngalert/api.go#L137) for those cases.

marefr: Definitely think 404 should be returned if it doesn't exist. In some places in the code base we let the sqlstore check rowsaffected and return `NotFoundError`, not sure if that's the proper way to go. With the changes returning the id of the entity affected is still a good thing to have I think instead of just a string. 

HTTP API design is tricky, but when implemented the folder API back in the days I was quite happy with returning the full DTO in the response ([code](https://github.com/grafana/grafana/blob/aa8fb1ae98bf25158d4d3431a8506d59718c35a5/pkg/api/folder.go#L83)), but highly personal.

These are nits and can be fixed later if you would like to get it merged.

papagian: Just to clarify: `validateOrgAlertDefinition` returns 404 if the alert definition is not found.

marefr: Thanks didn't notice that :+1:

marefr (approved):
LGTM

papagian:
There is a [prefix limitation](https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes/1814594#1814594) of 765 bytes in MySQL < 5.7. Therefore for adding the unique index on org_id, name columns
 I have added a MySQL database migration to resize the name column.

However, we had to impose this limit on the other dialects as well for preventing failures when migrating the database from these dialects to MySQL.
Unfortunately, because of SQLite [does not enforce the length of a VARCHAR](https://www.sqlite.org/faq.html#:~:text=SQLite%20does%20not%20enforce%20the,all%20500%2Dmillion%20characters%20intact.&text=SQLite%20understands%20the%20column%20type,of%20the%20value%20of%20N.) and the ALTER TABLE [limited support to alter and drop columns](https://stackoverflow.com/questions/4007014/alter-column-in-sqlite), there was no easy way to apply this in a database migration.

Hence, I have added an alert definition validation preventing having names greater than 190 characters:
https://github.com/grafana/grafana/blob/672a09323a46e77d3d27240e4c9e01f3fbf9fd75/pkg/services/ngalert/validator.go#L25-L27

Also, naming a column "name" has been proved to be a poor choice since it's reserved word for some dialects and requires quoting.

PR #29310, opened 2020-11-23
InfluxDB: Update flux library to fix support for boolean label values

ryantxu:
TimeSeries queries with boolean values can return errors.  This PR aims to fix that.


pkg/tsdb/influxdb/flux/executor_test.go
```diff
@@ -204,6 +204,12 @@ func TestBuckets(t *testing.T) {
 	})
 }
 
+func TestBooleanGrouping(t *testing.T) {
+	t.Run("Boolean", func(t *testing.T) {
```
aknuds1, line 208: Why do you need a sub-test? Doesn't seem to add anything from the base test case.

aknuds1 (approved):
Generally LGTM.

pkg/tsdb/influxdb/flux/executor_test.go
ryantxu: The test setup could maybe use some other style, but in general, this checks that:
[boolean.csv](https://github.com/grafana/grafana/pull/29310/files#diff-6fa002d068edadc27b2d0c767d5af644f57ddfb7b69c70e332552d8eb7c5bbc8)

becomes:
[boolean.golden.txt](https://github.com/grafana/grafana/pull/29310/files#diff-82729d376d3264b309c15f3ffea879cedf042dd39aff2c4b3271ab6941537a78)

aknuds1: Yeah I think the style of the test is fine, I just don't see why it needs to be wrapped in a sub-test :)

ryantxu: ahh -- cool.. I just remove all the redundant `t.Run` calls -- so now it is a big PR on the tests, but no real changes

torkelo:
Anything we want to include in a patch release?

ryantxu:
yes -- this could go in the patch release if possible

torkelo:
Should this close #28976 ?

torkelo:
was a conflict so did a manual backport https://github.com/grafana/grafana/pull/29333

PR #29313, opened 2020-11-23
Bug: Update yarn.lock to use latest rc-util

zoltanbedi:
This is a regression after https://github.com/grafana/grafana/pull/29057
Fixes this issue that caused by rc-util@5.0.4.
Also removed types/rc-slider dep as rc-slider comes with own typing.

screenshot_from_2020-11-23_11-44-18


mckn (approved):
Great work, as always @zoltanbedi!

PR #29315, opened 2020-11-23
Chore: bump grafana-ui rollup dependencies

jackw:
Updates all rollup dependencies in grafana-ui. Removes `namedExports` and thus the need for build tooling to know about grafana-ui dependencies!

Dependency stats with the updated rollup dependencies looks like:

Screenshot 2020-11-23 at 19 24 39

Fixes #

Labelling it as draft until I find the time to check the output properly.


torkelo:
ran into rollup issue in https://github.com/grafana/grafana/pull/29450 so helped out testing this, published a canary build and tested from plugin, seems to work as it should. Ran build and unit tests from the plugin and could load and render components from the built pacakge

torkelo (approved):

PR #29326, opened 2020-11-24
Docs: fixing link issues in auto generated frontend docs.

### LINKED ISSUE #26260: Docs feedback: /packages_api/data/datasourcesettings.md
it would be great if all the fields in 'options' were documented - it's not entirely clear (from the names) what should be used for what, and the whole jsonData vs secureJsonData semantic should be explained.

### LINKED ISSUE #26459: Docs feedback: /packages_api/runtime/backendsrv.md
There is a link on the documentation page for BackendSrv that links to BackendSrvRequest.md but it's a 404.  There is a similar issue in [/packages_api/runtime/datasourcesrv/](https://grafana.com/docs/grafana/latest/packages_api/runtime/datasourcesrv/) where it tries to link to [DataSourceApi.md](https://grafana.com/docs/grafana/latest/packages_api/runtime/datasourcesrv/data/datasourceapi.md) which is also a 404.  If I find any others I will add them here.

### LINKED ISSUE #26487: Docs feedback: /packages_api/data/dataqueryrequest.md
[DataQueryRequest](https://grafana.com/docs/grafana/latest/packages_api/data/dataqueryrequest/ ) When you have time, please add field descriptions for this, object as well as the associated response.  I realize that with the 7.x datasource api most people may not need to work with these objects any longer but if you try to access the datasource directly (for example from your query editor) it's not always clear what to do with the required fields.

mckn:
This PR will resolve wrongly generated documentation links when we reference other classes, interfaces, enum etc. in code comments by using the `@link` tag.

It also adds the `disable_edit_link = true` so we don't link users to the github edit page since this documentation is autogenerated and needs to be updated in the code.

We are also changing the link type from regular anchor tags to `relref` so we will get Hugo warnings when the link can't be resolved (https://gohugo.io/content-management/cross-references/#use-ref-and-relref).

Fixes #26459
Fixes #26487
Fixes #26260

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@babel/preset-react": "7.8.3",
     "@babel/preset-typescript": "7.8.3",
     "@emotion/core": "10.0.27",
-    "@grafana/api-documenter": "0.9.3",
+    "@grafana/api-documenter": "7.11.1",
     "@grafana/api-extractor": "7.10.1",
     "@grafana/eslint-config": "2.0.6",
     "@rtsao/plugin-proposal-class-properties": "7.0.1-patch.1",
--- a/packages/grafana-data/src/transformations/matchers.ts
+++ b/packages/grafana-data/src/transformations/matchers.ts
@@ -38,7 +38,7 @@ export const frameMatchers = new Registry<FrameMatcherInfo>(() => {
 
 /**
  * Resolves a field matcher from the registry for given config.
- * If no matcher can be found an error will be thrown.
+ * Will throw an error if matcher can not be resolved.
  * @public
  */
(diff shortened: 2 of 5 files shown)


peterholmberg:
are most of the fixes because of the bump from version 0.x to 7.x ?

mckn:
> are most of the fixes because of the bump from version 0.x to 7.x ?

Yeah, but the reason why it is bumped from 0.x to 7.x is because I started to follow the microsoft/api-documenter package versioning. Since this is just a fork with the Hugo markdown support added.

Sorry that I didn't add that to the description above.

dprokop (approved):

PR #29327, opened 2020-11-24
PanelEditor: allow access to the eventBus from panel options

ryantxu:
This PR adds the dashboard eventBus to the PanelOptionsEditor -- this makes it possible to pass events between the panel and the options editor.


torkelo:
Can you describe a scenario where you need to pass events between the two?

ryantxu:
work in progress:
events

torkelo:
Ah, see. Yea for this updating options and setting a selected state feels less ideal than sending an event

torkelo (approved):

PR #29328, opened 2020-11-24
ReleaseNotes: Update link in package.json

torkelo:
Updates package.json to link to new release notes page in docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "grafana": {
     "whatsNewUrl": "https://grafana.com/docs/grafana/latest/guides/whats-new-in-v7-3/",
-    "releaseNotesUrl": "https://community.grafana.com/t/release-notes-v7-3-x/37993"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
   },
   "grafana": {
     "whatsNewUrl": "https://grafana.com/docs/grafana/latest/guides/whats-new-in-v7-3/",
-    "releaseNotesUrl": "https://community.grafana.com/t/release-notes-v7-3-x/37993"
+    "releaseNotesUrl": "https://grafana.com/docs/grafana/latest/release-notes/"
```
mckn, line 49: Shouldn't this be something like: `https://grafana.com/docs/grafana/latest/release-notes/release-notes-7-3-4/`?

torkelo: not sure, was thinking it could go to the index page so we do not have to update this every version bump. But we could change this in the publish code as well now that the URL is deterministic and only includes the version

PR #29330, opened 2020-11-24
Security: Fixes minor security issue with alert notification webhooks that allowed GET & DELETE requests  

wbrowne:


bergquist (approved):
LGTM

PR #29332, opened 2020-11-24
Variables: Adds description field 

hugohaggmark:
**What this PR does / why we need it**:
This PR adds a description field to all variable types that can be used as described in #13138. I refactored all the components so they use the new InlineFormElements too, not sure I'm using them as intended though.

**Which issue(s) this PR fixes**:
Fixes #13138

**Special notes for your reviewer**:
I'm going to branch off this branch and fix proper e2e tests.

PR #29341, opened 2020-11-24
Chore: Adds e2e tests for Variables

hugohaggmark:
During the refactor in #29332 I realized we had a bunch of tests but they were outdated and had __`skip`__ so they were not exercised at all. This PR takes some learnings from those tests and introduces smaller, more targeted tests.

Relates #29332

This is a branch of #29332 so when that merges with master the review should be easier :)


torkelo (approved):
Looks good! Think a good level of detail tested here

hugohaggmark:
Let's hope it doesn't introduce flakiness 🤞

PR #29345, opened 2020-11-24
Logging: log frontend errors caught by ErrorBoundary, including component stack

domasx2:
* `<ErrorBoundary/>` pushes captured errors to sentry, including react component stack as "context"
* Backend log endpoint will log error context

log component stack

Related to #27993


pkg/api/frontend_logging.go
```diff
@@ -56,6 +68,9 @@ func (event *frontendSentryEvent) ToLogContext() log15.Ctx {
 	if event.Exception != nil {
 		ctx["stacktrace"] = event.Exception.FmtStacktraces()
 	}
+	if event.Contexts != nil {
+		addEventContextToLogContext("context", &ctx, &event.Contexts)
+	}
```
aknuds1, line 73: There's no need to check if `event.Contexts` is nil because iterating over a nil map isn't an error, it behaves the same as if it were empty (you can read about this [here](https://blog.golang.org/maps)):
```suggestion
	addEventContextToLogContext("context", ctx, event.Contexts)
```

pkg/api/frontend_logging.go
```diff
@@ -47,6 +47,18 @@ func (exception *frontendSentryException) FmtStacktraces() string {
...
+	for key, element := range *eventCtx {
+		prefix := fmt.Sprintf("%s_%s", rootPrefix, key)
+		switch v := element.(type) {
+		case map[string]interface{}:
+			addEventContextToLogContext(prefix, logCtx, &v)
+		default:
+			(*logCtx)[prefix] = fmt.Sprintf("%v", v)
+		}
+	}
+}
```
aknuds1, line 60: Why should `eventCtx` and `logCtx` be pointers? I don't see why there would be any need, maps are reference types:
```suggestion
func addEventContextToLogContext(rootPrefix string, logCtx log15.Ctx, eventCtx map[string]interface{}) {
	for key, element := range eventCtx {
		prefix := fmt.Sprintf("%s_%s", rootPrefix, key)
		switch v := element.(type) {
		case map[string]interface{}:
			addEventContextToLogContext(prefix, logCtx, &v)
		default:
			logCtx[prefix] = fmt.Sprintf("%v", v)
		}
	}
}
```

aknuds1 (changes requested):

dprokop (approved):
Frontend-wise this looks good

pkg/api/frontend_logging.go
domasx2: Oh boy, go has some unconventional design decisions :) though does makes sense. Fixed-

pkg/api/frontend_logging.go
domasx2: Fixed! Didn't realize that maps are ref types and are not copied when passed as params. I started reading a book on golang, will get myself up to speed

aknuds1 (approved):

domasx2:
@aknuds1 @dprokop  could any of you guys merge this? I'm not authorized, thanks :)

PR #29334, opened 2020-11-24
Drone: Check CUE dashboard schemas

aknuds1:
**What this PR does / why we need it**:
Extend Drone config to check CUE dashboard schemas. Also install `cue` and `shellcheck` (to avoid downloading in Drone builds) tools in build image.

Since we install `shellcheck` now in the build image, I'm removing the downloading of it from the Drone config

PR #29352, opened 2020-11-24
DashboardSchema: Compatible with Cue 3.0.0-alpha5

trotttrotttrott:

Lead all top-level properties that we do not intend to export with an underscore.

In 2.2.0, only definitions starting with a capital letter were exported. As well, properties with a leading underscore were not visible to an entire package - just within the same file. 3.0.0 still considers them hidden, however, they are now visible to an entire package.

Unblocks https://github.com/grafana/grafana/pull/29334#issuecomment-732805078.

I had hoped to also rename files of hidden properties with a leading underscore, but `cue export` ignores them. Instead just the properties themselves are renamed. My POC in [grafana/dashboard-spec](https://github.com/grafana/dashboard-spec/tree/0a756ea8a6aa4c76b24add2057beb70c37d6c65a/specs/7.0/panels) used a leading underscore naming convention which I think helped distinguish files with hidden properties. Not a big deal, but I wanted to mention I made the attempt.

### DIFF
--- a/dashboard-schemas/panels/Gauge.cue
+++ b/dashboard-schemas/panels/Gauge.cue
@@ -2,7 +2,7 @@ package panels
 
 // Gauge is a single value panel that can repeat a gauge for every series,
 // column or row.
-#Gauge: panel & {
+#Gauge: _panel & {
 	// Field config.
 	fieldConfig: {
 		// Defaults.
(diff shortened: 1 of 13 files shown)


aknuds1 (approved):

aknuds1:
Thanks Chris, works!

PR #29353, opened 2020-11-24
GraphNG: support x != time in library

ryantxu:
This is a simplification of #29288 that removes any changes to the UI and mapping fields to dimensions.  There are enough UX issues that I think we should tackle that independently.

This PR focus on making sure the library/component does not assume x===time, #29288 will look at how to expose that in the UI


torkelo (approved):
Sorry forgot to approve this yesterday. 

Seems like alignDataFrames still return non time series data. Get null back when I tested with a frame with two non time series fields

ryantxu:
> Get null back when I tested with a frame with two non time series fields

Yes current implementation has the exact same expectations as existing panel but is structured so that we can change X/Y in #29288

Better default "Auto" behavior should be:
* if time exists use that as X vs all other numbers
* otherwise first number vs all other numbers

PR #29357, opened 2020-11-24
Instrumentation: Add examplars for request histograms

bergquist:
This will help us identify traces for slow requests and build RED method style monitoring which can be aggregated beyond one Grafana instance.

An active user has about 85 unique combinations of handlers + code + method. This histogram introduces 12 new series which means it adds about 1020 series in total. However, since we can remove both the summaries and counters we get rid of 6 series which means we end up with ~510 new series per instance. 


wbrowne (approved):
Awesome 👏

PR #29358, opened 2020-11-25
SQLStore: customise the limit of retrieved datasources per organisation

### LINKED ISSUE #29262: SQLStore: customise the limit of retrieved datasources per organisation
Change this limit and make it configurable:
https://github.com/grafana/grafana/blob/ab078133bfe9771f06bd21712f7ad6f9113c8a61/pkg/services/sqlstore/datasource.go#L70

**Why is this needed**:
This is short-term solution until we get [pagination](https://github.com/grafana/grafana/issues/29260).

ying-jeanne:
**What this PR does / why we need it**: the PR is to enable customization of the number of data sources returned per organization, the origin ticket is here: https://github.com/grafana/grafana/issues/29262

Fixes #29262

**Special notes for your reviewer**: The Getdatasourcequery is also used in frontendsettings.go, Functionally speaking I am not sure whether we need to use the default limit of 5000 or we should just remove the limit as we do for alerting.


aknuds1:
Looks like there's a [failure](https://drone.grafana.net/grafana/grafana/7238/1/6) to compile the Go code.

conf/defaults.ini
```diff
@@ -255,6 +255,11 @@ min_refresh_interval = 5s
 # Path to the default home dashboard. If this value is empty, then Grafana uses StaticRootPath + "dashboards/home.json"
 default_home_dashboard_path =
 
+################################### Datasources #########################
```
aknuds1, line 258: As far as I recall, we're standardizing on writing it as "data sources".
```suggestion
################################### Data sources #########################
```

conf/sample.ini
```diff
@@ -110,6 +110,11 @@
 # For "sqlite3" only. cache mode setting used for connecting to the database. (private, shared)
 ;cache_mode = private
 
+################################### Datasources #########################
```
aknuds1, line 113:
```suggestion
################################### Data sources #########################
```

conf/defaults.ini
```diff
@@ -255,6 +255,11 @@ min_refresh_interval = 5s
...
+
```
aknuds1, line 259: Can you remove this blank line, just to be consistent?
```suggestion
```

conf/sample.ini
```diff
@@ -110,6 +110,11 @@
...
+[datasources]
+# Max data sources return in the result default is 5000
```
aknuds1, line 115: I wouldn't document the default in a comment here, since it overlaps with the default in defaults.ini, and you risk them going out of sync. I would just use the same documentation comment for both this file and defaults.ini.
```suggestion
# Upper limit of data sources that Grafana will return
```

conf/defaults.ini
```diff
@@ -255,6 +255,11 @@ min_refresh_interval = 5s
...
+[datasources]
+datasource_limit = 5000
```
aknuds1, line 261:
```suggestion
# Upper limit of data sources that Grafana will return
datasource_limit = 5000
```

docs/sources/administration/configuration.md
```diff
@@ -343,6 +343,12 @@ Defaults to `private`.
 
 <hr />
 
+## [database]
+
+### datasource_limit
+
+The limit of data source in the response, Set only if customer wants to overwrite the default value 5000
```
aknuds1, line 350:
```suggestion
The upper limit of number of data sources returned in the response, set this if you need to override the default.
```

pkg/api/datasources.go
```diff
@@ -19,8 +19,8 @@ import (
 
 var datasourcesLogger = log.New("datasources")
 
-func GetDataSources(c *models.ReqContext) Response {
-	query := models.GetDataSourcesQuery{OrgId: c.OrgId}
+func (hs *HTTPServer) GetDataSources(c *models.ReqContext) Response {
+	query := models.GetDataSourcesQuery{OrgId: c.OrgId, DataSourceLimite: &hs.Cfg.DataSourceLimit}
```
aknuds1, line 23:
```suggestion
	query := models.GetDataSourcesQuery{OrgId: c.OrgId, DataSourceLimit: &hs.Cfg.DataSourceLimit}
```

pkg/api/frontendsettings.go
```diff
@@ -15,11 +15,11 @@ import (
 	"github.com/grafana/grafana/pkg/setting"
 )
 
-func getFSDataSources(c *models.ReqContext, enabledPlugins *plugins.EnabledPlugins) (map[string]interface{}, error) {
+func (hs *HTTPServer) getFSDataSources(c *models.ReqContext, enabledPlugins *plugins.EnabledPlugins) (map[string]interface{}, error) {
 	orgDataSources := make([]*models.DataSource, 0)
 
 	if c.OrgId != 0 {
-		query := models.GetDataSourcesQuery{OrgId: c.OrgId}
+		query := models.GetDataSourcesQuery{OrgId: c.OrgId, DataSourceLimite: &hs.Cfg.DataSourceLimit}
```
aknuds1, line 22:
```suggestion
		query := models.GetDataSourcesQuery{OrgId: c.OrgId, DataSourceLimit: &hs.Cfg.DataSourceLimit}
```

pkg/models/datasource.go
```diff
@@ -202,9 +202,10 @@ type DeleteDataSourceByNameCommand struct {
 // QUERIES
 
 type GetDataSourcesQuery struct {
-	OrgId  int64
-	User   *SignedInUser
-	Result []*DataSource
+	OrgId            int64
+	DataSourceLimite *int
```
aknuds1, line 206:
```suggestion
	DataSourceLimit *int
```

pkg/services/sqlstore/datasource.go
```diff
@@ -67,8 +67,12 @@ func GetDataSourceByName(query *models.GetDataSourceByNameQuery) error {
 }
 
 func GetDataSources(query *models.GetDataSourcesQuery) error {
-	sess := x.Limit(5000, 0).Where("org_id=?", query.OrgId).Asc("name")
-
+	var sess *xorm.Session
+	if query.DataSourceLimite == nil {
```
aknuds1, line 71:
```suggestion
	if query.DataSourceLimit == nil {
```

pkg/services/sqlstore/datasource.go
```diff
@@ -67,8 +67,12 @@ func GetDataSourceByName(query *models.GetDataSourceByNameQuery) error {
...
+		sess = x.Where("org_id=?", query.OrgId).Asc("name")
+	} else {
+		sess = x.Limit(*query.DataSourceLimite, 0).Where("org_id=?", query.OrgId).Asc("name")
```
aknuds1, line 74:
```suggestion
		sess = x.Limit(*query.DataSourceLimit, 0).Where("org_id=?", query.OrgId).Asc("name")
```

pkg/services/sqlstore/datasource_test.go
```diff
@@ -214,4 +215,27 @@ func TestDataAccess(t *testing.T) {
 
 		require.Equal(t, 0, len(query.Result))
 	})
+
+	t.Run("GetDataSource", func(t *testing.T) {
+		t.Run("Datasource limit to 6 per organization", func(t *testing.T) {
```
aknuds1, line 220:
```suggestion
		t.Run("Number of data sources returned limited to 6 per organization", func(t *testing.T) {
```

pkg/services/sqlstore/datasource_test.go
```diff
@@ -214,4 +215,27 @@ func TestDataAccess(t *testing.T) {
...
+					Name:     "laban" + strconv.Itoa(i),
+					Type:     models.DS_GRAPHITE,
+					Access:   models.DS_ACCESS_DIRECT,
+					Url:      "http://test",
+					Database: "site",
+					ReadOnly: true,
+				})
+				require.NoError(t, err)
+			}
+			query := models.GetDataSourcesQuery{OrgId: 10, DataSourceLimite: &datasourceLimit}
```
aknuds1, line 235:
```suggestion
			query := models.GetDataSourcesQuery{OrgId: 10, DataSourceLimit: &datasourceLimit}
```

pkg/services/sqlstore/sqlstore.go
```diff
@@ -333,7 +333,7 @@ func (ss *SQLStore) readConfig() {
 	ss.dbCfg.CacheMode = sec.Key("cache_mode").MustString("private")
 }
 
-// Interface of arguments for testing db
+// ITestDB Interface of arguments for testing db
```
aknuds1, line 336: [Comments documenting declarations should be full sentences](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences)
```suggestion
// ITestDB is an interface of arguments for testing DB
```

pkg/setting/setting.go
```diff
@@ -316,6 +316,9 @@ type Cfg struct {
 
 	// Sentry config
 	Sentry Sentry
+
+	// Data Sources
```
aknuds1, line 320:
```suggestion
	// Data sources

	// DataSourceLimit configures the upper limit of number of data sources to return.
```

pkg/setting/setting.go
```diff
@@ -1225,3 +1232,9 @@ func readServerSettings(iniFile *ini.File, cfg *Cfg) error {
 
 	return nil
 }
+
+func readDataSourcesSettings(iniFile *ini.File, cfg *Cfg) error {
```
aknuds1, line 1236: See previous comment, also there's no error to return:
```suggestion
func (cfg *Cfg) readDataSourceSettings() {
```

pkg/setting/setting.go
```diff
@@ -1225,3 +1232,9 @@ func readServerSettings(iniFile *ini.File, cfg *Cfg) error {
...
+	datasources := iniFile.Section("datasources")
```
aknuds1, line 1237: Shouldn't this be done via cfg.Raw?

```suggestion
	datasources := cfg.Raw.Section("datasources")
```

pkg/setting/setting.go
```diff
@@ -1225,3 +1232,9 @@ func readServerSettings(iniFile *ini.File, cfg *Cfg) error {
...
+	cfg.DataSourceLimit = datasources.Key("datasource_limit").MustInt(5000)
+	return nil
```
aknuds1, line 1239:
```suggestion
```

pkg/setting/setting.go
```diff
@@ -806,6 +809,10 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 		return err
 	}
 
+	if err := readDataSourcesSettings(iniFile, cfg); err != nil {
+		return err
+	}
```
aknuds1, line 814: As the pattern otherwise is to use utility methods on Cfg instead of free functions, I think it's best to be consistent here and repeat the pattern:
```suggestion
	cfg.readDataSourceSettings()
```

aknuds1 (commented):
I noticed some typos, and I think there are some things that should be done differently in settings parsing f.ex.

papagian:
> Looks like there's a [failure](https://drone.grafana.net/grafana/grafana/7238/1/6) to compile the Go code.

@ying-jeanne the reason of the failure is the the `GetDatasSources()` now gets a receiver.
Therefore in the test you have to create an HTTPServer:
`
		hs := &HTTPServer{
			Bus: bus.GetBus(),
			Cfg: setting.NewCfg(),
		}
`
and then change the `handlerFunc` to: `hs.GetDataSources` 

Also in the `Cfg` you have passed to the HTTPServer, you can set also the limit like:
`hs.Cfg.DataSourceLimit=3`
and then you can add a similar test for testing the limit when is called from the API.

aknuds1:
@ying-jeanne It's best if you don't force-push to the branch, since it makes GitHub lose review comments. We squash merge in any case, so keeping a clean branch history is little use :)

pkg/services/sqlstore/datasource.go
```diff
@@ -67,8 +67,12 @@ func GetDataSourceByName(query *models.GetDataSourceByNameQuery) error {
 }
 
 func GetDataSources(query *models.GetDataSourcesQuery) error {
-	sess := x.Limit(5000, 0).Where("org_id=?", query.OrgId).Asc("name")
-
+	var sess *xorm.Session
+	if query.DataSourceLimit == nil {
+		sess = x.Where("org_id=?", query.OrgId).Asc("name")
```
papagian, line 72: I wonder if this would be too costly if there is a huge amount of data sources.
Off course the table has an [index](https://github.com/grafana/grafana/blob/aebe8985c57f7b24b7245066d85e1bf3b5b58313/pkg/services/sqlstore/migrations/datasource_mig.go#L67) on the `org_id` column but anyway since the previous behaviour was to fetch only the first 5000
I would suggest to fetch all the datasources in batches of 5000 (using a loop).
@aknuds1 WDYT?

papagian (changes requested):
Great start! I have requested some changes.

Also, in addition to the sqlstore test can you add an API test (similar to [this](https://github.com/grafana/grafana/blob/64f2a11200c00e3e3aee3015dbd402e8f6843c6f/pkg/api/datasources_test.go#L21) one) for testing the limit enforcement in the API?
For such a test you would only need to set the limit in the HTTPServer `Cfg` to something lower to the number of the datasources in the fake handler (4 in the specific case) and check that the number of the datasources in the response equals this `limit.`

pkg/models/datasource.go
```diff
@@ -202,9 +202,10 @@ type DeleteDataSourceByNameCommand struct {
 // QUERIES
 
 type GetDataSourcesQuery struct {
-	OrgId  int64
-	User   *SignedInUser
-	Result []*DataSource
+	OrgId           int64
+	DataSourceLimit *int
```
aknuds1, line 206: I don't think this should be a pointer, instead we should use 0 to indicate that there's no limit. It's more idiomatic Go, and less error prone.

pkg/services/sqlstore/datasource.go
aknuds1: @papagian To tell you the truth I'm unsure. Why is it less costly to fetch in batches? Does it not impose the same load on the whole (we still populate a full array with all the results)? @kylebrandt Do you have any insight to share in this regard?

pkg/models/datasource.go
papagian: I have suggested to use a pointer to imply that there is not limit set intentionally.
But I'm open to discuss it.

pkg/services/sqlstore/datasource.go
aknuds1: @papagian BTW thinking more about this, doesn't this also cause a change in behaviour? I.e., if the caller hasn't set DataSourceLimit, the query will be unlimited. But those callers that uses the configuration will always have set a non-nil `DataSourceLimit`, so those cases will never be unlimited.

As such, I think that we need to retain the old behaviour of 5000 as default limit.

aknuds1 (dismissed):
It looks to me as if the PR changes the default to be not 5000, but unlimited if `GetDataSourcesQuery.DataSourceLimit` is nil.

pkg/models/datasource.go
aknuds1: @papagian AFAICT it changes behaviour though, so calls that would previously be limited to 5000 are unlimited.

I think it shouldn't be a pointer, but probably 5000 by default if `DataSourceLimit` is 0 (in case callers doesn't set it). Maybe when reading the configuration, `DataSourceLimit` should be set to -1 if the user passes zero to indicate unlimited? WDYT?

pkg/services/sqlstore/datasource.go
papagian: yes this is called by the api and the alerting service. For alerting I think that there should no limit applied for the reasons I have described [here](https://github.com/grafana/grafana/issues/29262#issuecomment-732972406).

papagian:
> It looks to me as if the PR changes the default to be not 5000, but unlimited if `GetDataSourcesQuery.DataSourceLimit` is nil.

yes, that was intentional; see [my comment ](https://github.com/grafana/grafana/pull/29358#discussion_r530557968)in the original issue

pkg/services/sqlstore/datasource.go
aknuds1: @papagian If we shouldn't retain the old limit, we should use the value of 0 to represent no limit. Then we would obtain homogeneous behaviour if the user configures 0 for the limit also, i.e. that it's unlimited, rather than actually limited at 0, which would happen now.

pkg/models/datasource.go
aknuds1: This is discussed in a previous comment thread 🙂

papagian: I wrote some additional tests and I've verified that if the [limit](https://godoc.org/github.com/go-xorm/xorm#Engine.Limit) is zero or negative it will be ignored.
Therefore I agree that it will simplify the code not to be a pointer.
@ying-jeanne can you change it?

aknuds1: @papagian @ying-jeanne Can we explicitly not set the limit if it's zero, since [Engine.Limit](https://godoc.org/github.com/go-xorm/xorm#Engine.Limit) doesn't define zero limit as equivalent to unlimited? I.e., we shouldn't depend on undefined behaviour and it would also make the intent obvious to readers of our code.

pkg/services/sqlstore/datasource.go
papagian: It is used by the json package for distinguishing between missing/null and zero values and I had got the impression that it was generally used but it seems that it's not that common.
Anyway given that 0 value is no limit I agree that it is better not to use a pointer and in that case we don't need even this conditional.

papagian: I've updated my comment above in order to align with [this](https://github.com/grafana/grafana/pull/29358/files#r530834486) comment.

docs/sources/administration/configuration.md
```diff
@@ -343,6 +343,12 @@ Defaults to `private`.
 
 <hr />
 
+## [database]
+
+### datasource_limit
+
+The upper limit of number of data sources returned in the response, set this if you need to override the default.
```
oddlittlebird, line 350:
```suggestion
The upper limit of number of data sources returned in the response. Set this if you need to override the default.
```

oddlittlebird: What is the limit? I do not see it stated anywhere.

pkg/services/sqlstore/datasource.go
bergquist: > I would suggest to fetch all the datasources in batches of 5000 (using a loop). was to fetch only the first 5000

I think it's fine to fetch all in this case since 5k will remain the default limit. Anyone who increases that limits is operating outside our recommendation and doing something weird. Paging will eventually be the long-term solution while this is just a temporary workaround IMO. Perhaps we should add a note about this in the config file/settings 

> yes, this is called by the api and the alerting service. For alerting I think that there should no limit applied for the reasons I have described here.

Seems reasonable. Perhaps we could cache the query instead of running it once for every alert rule if dsname is missing? The best solution would ofc be to migrate all those alert rules that don't have a datasource name. But I have a feeling an alert rule created the last three years have data source name or id configured :)

> I'm pretty certain that the Go idiom is to generally use zero values instead of nil pointers to represent defaults, but I'm unable to google it up.

https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

In general, I dont think we have to think too much about performance optimization for instances more than 5000k datasources.

docs/sources/administration/configuration.md
marefr: 5000 is the default limit

conf/sample.ini
```diff
@@ -110,6 +110,11 @@
 # For "sqlite3" only. cache mode setting used for connecting to the database. (private, shared)
 ;cache_mode = private
 
+################################### Data sources #########################
+[datasources]
+# Upper limit of data sources that Grafana will return
```
sakjur, line 115: I think this should mirror defaults.ini

pkg/services/sqlstore/datasource_test.go
```diff
@@ -214,4 +215,27 @@ func TestDataAccess(t *testing.T) {
...
+					Url:      "http://test",
+					Database: "site",
+					ReadOnly: true,
+				})
+				require.NoError(t, err)
+			}
+			query := models.GetDataSourcesQuery{OrgId: 10, DataSourceLimit: &datasourceLimit}
+			err := GetDataSources(&query)
+			require.NoError(t, err)
+			require.Equal(t, datasourceLimit, len(query.Result))
```
sakjur, line 238: Would it make sense to do a validation that the result is larger than the data source limit if no data source limit is set?

i.e. adding this to the bottom of the test

```go
query := models.GetDataSourcesQuery{OrgId: 10}
err := getDataSources(&query)
require.NoError(t, err)
assert.Len(t, query.Result, datasourceLimit+1)
```

sakjur (commented):
Looks good to me, just need to update the comment in `sample.ini`/`defaults.ini` so that they match each other 🙂

conf/defaults.ini
```diff
@@ -255,6 +255,11 @@ min_refresh_interval = 5s
 # Path to the default home dashboard. If this value is empty, then Grafana uses StaticRootPath + "dashboards/home.json"
 default_home_dashboard_path =
 
+################################### Data sources #########################
+[datasources]
+# Upper limit of data sources that Grafana will return and It is a temporary configuration, there is not planned to be support in the future.
```
sakjur, line 260:
```suggestion
# Upper limit of data sources that Grafana will return. Data source limits is a temporary configuration and will disappear when pagination is introduced on the list data sources API.
```

Tried to clarify a little?

ying-jeanne:
The default behavior of XORM is checked, when the limit is negative, no limit is set for the search query. Just to make sure that whenever we integrate a new version of XORM, the behavior should not change on our side. I put the behavior as XORM in our code. When the limit is <= 0, no limit is set for the getDatasource result.

pkg/services/sqlstore/datasource_test.go
```diff
@@ -214,4 +215,73 @@ func TestDataAccess(t *testing.T) {
...
+				require.NoError(t, err)
+			}
+			query := models.GetDataSourcesQuery{OrgId: 10}
+
+			err := GetDataSources(&query)
+
+			require.NoError(t, err)
+			require.Equal(t, numberOfDatasource, len(query.Result))
+		})
+		t.Run("Number of data sources returned limited to -1 per organization", func(t *testing.T) {
```
papagian, line 264: I would suggest adding a new line after the previous test. Also, I would change the description to something like: `No limit should be applied on the returned data sources if the limit is negative`

papagian (approved):
I have tested it and it works as expected! I have some minor suggestions for improving some comments and test descriptions.

sakjur (approved):
LGTM based on reading through the code and the tests, haven't tried it locally (since Sofia did that 👆).

PR #29359, opened 2020-11-25
GraphNG: Bars, Staircase, Smooth modes

leeoniya:
image

still plenty to do:

- [x] fix fills (to 0) in staircase and smooth modes
- [x] Points-only mode
- [x] omit creating intermediate `{x,y}` objects for smoothed mode
- [x] split up into a few files (less monolithic, better org)
- [x] ~~adjustable max bar width or min gap size (maybe)~~
- [x] ~~better stroke alignment for bars~~ **strokes are all center-aligned to the y values, same as line paths**
- [x] ~~optional left or center alignment for both bars and staircase~~ **probably not necessary, tho could be if histograms are done using bar mode instead of staircase mode**
- [x] ~~better data density heuristics for falling back to line mode from bars/smooth/staircase~~ **hard to get right without caveats (e.g. sparse data), and perf benefit not huge for sub-absurd datasets**
- [ ] ~~match point hover style to current Flot panel~~ **a future PR**
- [ ] ~~gaps support in staircase and smooth modes~~ **a future PR**


torkelo:
Nice!!❤️

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -1,31 +1,267 @@
...
-  lineWidth?: number;
-  points?: PointMode;
-  pointSize?: number;
-  pointColor?: string;
-  fill?: boolean;
-  fillOpacity?: number;
-  fillColor?: string;
+}
+
+function buildBarsPaths(u: uPlot, seriesIdx: number, idx0: number, idx1: number): Series.Paths {
```
torkelo, line 11: would be nice to move this part to a separate file (bars.ts or something). Preferably a unit test or two if possible.

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -1,31 +1,267 @@
...
+
+  let fill = series.fill != null ? new Path2D(stroke) : undefined;
+
+  return {
+    stroke,
+    fill,
+  };
+}
+
+function buildStaircasePaths(u: uPlot, seriesIdx: number, idx0: number, idx1: number): Series.Paths {
```
torkelo, line 57: would move this and the catmullRomFitting func to a lineInterpolation file

torkelo (commented):
really excited about getting line interpolation modes!

leeoniya:
i'd like to land this today or tomorrow so i'm gonna adjust push a few work items to a future PR.

- left/center alignment for staircase & bars - this may never be necessary. for now, staircase will be left aligned and bars will be center-aligned on the datapoints.
- gaps - this could be a decent chunk of  work and i'd rather not delay this PR for it. it may also help if i expose uPlot's internal gap accumulator and clip builder as utility functions, too: https://github.com/leeoniya/uPlot/blob/9650f0e9608d2a32960b49e6a1cf50ac6d0c7165/src/uPlot.js#L1121-L1155. needs additional thought/testing.
- bars stroke alignment - initially i thought i'd size the bars with the stroke aligned to the inside, there there would be no drawn pixels that exceed the actual y values, but then i realized that the stroe is always center-aligned even in the line paths and does the exact same thing. so i'm leaving this alone, it's not clear to me there is a huge amount of benefit for inside-aligning the stroke, since large stroke widths when this become problematic are mostly done for stylistic purposes rather than data accuracy. i will still land the part of this which reduces the bar width based on stroke size to keep the bar spacing from shrinking as the stroke width increases.

i found a small oversight in uPlot, where fills were drawn on top of strokes (in reverse order). this caused half the stroke width to be occluded behind the fill (if fill alpha = 1). if fill alpha < 1, half the stroke width's color would be "tainted" by the fill color. this bug is fixed, https://github.com/leeoniya/uPlot/issues/385 and i'll cut another uplot patch release today to bump it.

right now, there is a hard-coded bar width that fills 90% of the hz space, and no maximum allowable width. does anyone think these two controls should be exposed to give more flexibility over the bar spacing & feel?

i should be able to finish up the rest of this PR today.

torkelo:
> right now, there is a hard-coded bar width that fills 90% of the hz space, and no maximum allowable width. does anyone think these two controls should be exposed to give more flexibility over the bar spacing & feel? 

Don't think we need an option. The core panel has used the smallest distance between points as a heuristic but has a problem when there is only 2 data points, so a max of like 30% of hz space or something could be good

leeoniya:
> Don't think we need an option. The core panel has used the smallest distance between points as a heuristic but has a problem when there is only 2 data points, so a max of like 30% of hz space or something could be good

i tweaked it for the gaps to take up 25% of the available space, which was pretty close to existing behavior. we can refine this later.

leeoniya:
i don't see null mode exposed, which we'll need to do for me to test that gaps are handled properly. i'll do this as part of the gaps PR. for now it works essentially asZero:

image

torkelo:
Nice looks great!

leeoniya:
update on this.

i looked into the Flot code for how it chooses the size and opacity of its hover points:

image

the only way that uPlot currently allows customizing its hover point styles is to make a function that crafts the proper HTML element. the internal logic is terribly trivial and and i'm not eager to duplicate it externally (with its utility funcs) and then tweak it further grafana.

the current plan is to have uPlot expose a more declarative interface for hover point customization, similar to what it does for `series.points`. i also like the flot hover point style, so i will re-create it as the new native uPlot behavior, so the amount of tweaking in grafana should be zero.

i will also land the bit of prep work needed for supporting gap detection in the followup PR.

leeoniya:
i'm pushing the hover point work to the next PR. let's try to merge what's done (pending tests passing and review approval).

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
leeoniya: moved out to `paths.ts`

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
leeoniya: moved out to `paths.ts`

packages/grafana-ui/src/components/uPlot/paths.ts
```diff
@@ -0,0 +1,257 @@
+import uPlot, { Series } from 'uplot';
+
+export function Bars(u: uPlot, seriesIdx: number, idx0: number, idx1: number): Series.Paths {
```
ryantxu, line 3: In general classes/components get exported as uppercase and functions/objects are lowercase.

Since this has to match a known type, should we export it as:
```suggestion
export const barsBuilder: Series.PathBuilder = (u: uPlot, seriesIdx: number, idx0: number, idx1: number) => {```

ryantxu (approved):
Looks good!  lets merge this and fix the rest in new PRs

leeoniya:
:shipit:

PR #29377, opened 2020-11-25
Tracing: Fix issue showing more than 300 spans

### LINKED ISSUE #29244: Tempo datasource cuts off large traces
**What happened**:
Large traces get cut off in the Tempo datasource.  This trace (with 939 spans) is cut off about 1/4 of the way down:

image

See the entire trace.

1) Generate and query for a trace of a couple hundred spans
2) Scroll down til the trace gets cutoff

**Environment**:
- Grafana version:  Grafana v7.4.0-6940pre (993adb72e0)
- Data source type & version:  Tempo (unsure on version)
- OS Grafana is installed on:  Docker/Kubernetes
- User OS & Browser: Ubuntu/Chrome
- Grafana plugins:
- Others:

zoltanbedi:
We have to switch from window scrolling because we use a scroll canvas for scrolling.

Fixes #29244


aocenas:
@zoltanbedi I am not sure this is good fix for this. As far as I can tell this disables virtualisation altogether and just shows all the spans which make any interaction horribly slow. I think easy fix may be some internal wrapper with defined height so that we have 2 different scrolls page + trace spans. Not ideal but we can make the height big enough so most traces will fit without need to scroll while big ones will be usable.

zoltanbedi:
@aocenas good catch. You are right it didn't work as I expected. I've changed the viewbuffer to a lower limit because I felt like 300 was too slow when I clicked on trace.

packages/jaeger-ui-components/src/TraceTimelineViewer/ListView/index.tsx
```diff
@@ -202,6 +202,10 @@ export default class ListView extends React.Component<TListViewProps> {
       }
       window.addEventListener('scroll', this._onScroll);
       this._windowScrollListenerAdded = true;
+    } else {
+      // The wrapper element should be the one that handles the scrolling. Once we are not using scroll-canvas we can remove this.
+      this._wrapperElm = document.getElementsByClassName('scroll-canvas')[0];
```
aocenas, line 207: Can we instead push this as a prop like `scrollElement?: DOMElement` or something like that from inside the explore/TraceView? Just to decouple this from the explore class names.

aocenas (approved):
Have one comment but after that is addressed this can be merged.

packages/jaeger-ui-components/src/TraceTimelineViewer/ListView/index.tsx
zoltanbedi: Absolutely. 👌

PR #29390, opened 2020-11-25
Alerting: Improve Prometheus Alert Rule error message

### LINKED ISSUE #29100: Alerting module should provide more detailed information in case of an error

I would like the alerting UI to be transparent why a certain error has happened when querying the underlying data source. While the same query will show its fully detailed error in the dashboard/explore view it is hidden as part of alerting evaluation.

**Why is this needed**:

Just displaying the status code is often not enough. Most data sources return better error conditions in the response body.

**Steps to reproduce**:

I have created this quick mockup URL which can be setup as Prometheus data source: https://curly-bird-331e.swine-registry.workers.dev/502/

Now when I query it with explore, I get this error:

scrn-2020-11-13-13-53-23

But as part of the alert this information is hidden:

scrn-2020-11-13-14-00-27

wbrowne:

Fixes #29100

Screenshot 2020-11-25 at 15 32 07

### DIFF
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/grafana/grafana/pkg/tsdb/prometheus"
+
 	gocontext "context"
 
 	"github.com/grafana/grafana-plugin-sdk-go/data"
@@ -159,11 +161,7 @@ func (c *QueryCondition) executeQuery(context *alerting.EvalContext, timeRange *
 
 	resp, err := c.HandleRequest(context.Ctx, getDsInfo.Result, req)
 	if err != nil {
-		if errors.Is(err, gocontext.DeadlineExceeded) {
-			return nil, fmt.Errorf("alert execution exceeded the timeout")
-		}
-
-		return nil, fmt.Errorf("tsdb.HandleRequest() error %v", err)
+		return nil, toCustomError(err)
 	}
 
 	for _, v := range resp.Results {
@@ -359,4 +357,20 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 	}
 
 	return seriesSlice, nil
+}
+
+func toCustomError(err error) error {
+	// is context timeout
+	if errors.Is(err, gocontext.DeadlineExceeded) {
+		return fmt.Errorf("alert execution exceeded the timeout")
+	}
+
+	// is Prometheus error
+	var e prometheus.PrometheusError
+	if errors.As(err, &e) {
+		return err
+	}
+
+	// generic fallback
+	return fmt.Errorf("tsdb.HandleRequest() error %v", err)
 }
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -366,9 +366,8 @@ func toCustomError(err error) error {
 	}
 
 	// is Prometheus error
-	var e prometheus.PrometheusError
-	if errors.As(err, &e) {
-		return err
+	if prometheus.IsAPIError(err) {
+		return prometheus.WrapAPIError(err)
 	}
 
 	// generic fallback
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -367,7 +367,7 @@ func toCustomError(err error) error {
 
 	// is Prometheus error
 	if prometheus.IsAPIError(err) {
-		return prometheus.WrapAPIError(err)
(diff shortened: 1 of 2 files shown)


pkg/services/alerting/conditions/query.go
```diff
@@ -360,3 +358,19 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
...
+
+func toCustomError(err error) error {
+	// is context timeout
+	if errors.Is(err, gocontext.DeadlineExceeded) {
+		return fmt.Errorf("alert execution exceeded the timeout")
+	}
+
+	// is Prometheus error
+	var e prometheus.PrometheusError
+	if errors.As(err, &e) {
```
aknuds1, line 370: You're not doing anything with the converted error, so should be `errors.Is`
```suggestion
	if errors.Is(err, prometheus.PrometheusError) {
```

aknuds1 (approved):
LGTM, but please see my suggestion.

pkg/services/alerting/conditions/query.go
wbrowne: I tried this pattern but it didn't work as I expected. It's not defined with `errors.New`, just given the `error` type so I'm not sure the suggestion applies?

simonswine: I think you need to create a struct rather than an alias type: https://play.golang.org/p/Ak4SLvQ2APB

wbrowne: Thanks for the tip @simonswine - made the wrong assumption that alias represented a distinct type. Have reworked it a different way without introducing an error type

pkg/tsdb/prometheus/prometheus.go
```diff
@@ -220,7 +218,12 @@ func parseResponse(value model.Value, query *PrometheusQuery) (*tsdb.QueryResult
 	return queryRes, nil
 }
 
-func wrapAsPrometheusError(err error) PrometheusError {
+func IsAPIError(err error) bool {
+	var e *apiv1.Error
+	return errors.As(err, &e)
```
aknuds1, line 223: `errors.Is` is the function that should be used to tell if an error is of certain type (or wraps it).
```suggestion
	return errors.Is(err, &apiv1.Error{})
```

pkg/tsdb/prometheus/prometheus.go
```diff
@@ -216,3 +217,16 @@ func parseResponse(value model.Value, query *PrometheusQuery) (*tsdb.QueryResult
...
+
+func IsAPIError(err error) bool {
+	var e *apiv1.Error
+	return errors.As(err, &e)
+}
+
+func WrapAPIError(err error) error {
+	var e *apiv1.Error
+	if errors.As(err, &e) {
+		return fmt.Errorf("%s: %s", e.Msg, e.Detail)
```
aknuds1, line 229: What do you actually gain by this? `apiv1.Error` already implements the `error` interface, and `e.Error()` returns exactly the same string as here.

aknuds1 (changes requested):
Unsure about `WrapAPIError`.

pkg/tsdb/prometheus/prometheus.go
```diff
@@ -216,3 +217,16 @@ func parseResponse(value model.Value, query *PrometheusQuery) (*tsdb.QueryResult
 
 	return queryRes, nil
 }
+
+func IsAPIError(err error) bool {
+	var e *apiv1.Error
+	return errors.As(err, &e)
+}
+
+func WrapAPIError(err error) error {
```
aknuds1, line 226: Better naming since the error isn't wrapped
```suggestion
func ConvertAPIError(err error) error {
```

aknuds1 (changes requested):

aknuds1 (approved):
LGTM!

pkg/services/alerting/conditions/query.go
```diff
@@ -360,3 +358,18 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
...
+
+func toCustomError(err error) error {
+	// is context timeout
+	if errors.Is(err, gocontext.DeadlineExceeded) {
+		return fmt.Errorf("alert execution exceeded the timeout")
+	}
+
+	// is Prometheus error
+	if prometheus.IsAPIError(err) {
+		return prometheus.WrapAPIError(err)
```
papagian, line 370:
```suggestion
		return prometheus.ConvertAPIError(err)
```

papagian (changes requested):
there is a reference to `WrapAPIError` left but otherwise works fine

papagian (approved):

PR #29396, opened 2020-11-25
GraphNG: Add test coverage for uPlot initialization and data updates

dprokop:
Fixed few minor bugs with initialization and data updates that surfaced during testing. Needs another iteration, I'm still not 100% happy with the logic. It's quite complex and easy to break due to dependencies between the data model being updated and the fields config possibly changing independently(from the checks point of view).


packages/grafana-ui/src/components/GraphNG/GraphNG.test.tsx
```diff
@@ -33,135 +51,140 @@ const mockData = () => {
...
+          height={100}
+          legend={{ displayMode: LegendDisplayMode.Hidden, placement: 'bottom' }}
+        />
+      );
+
+      // expect(uPlot).toBeCalledTimes(1);
+      // we wait 1 frame for plugins initialisation logic to finish
+      act(() => {
+        mockRaf.step({ count: 1 });
+      });
```
dprokop, line 124: Implementation detail leaking :(

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -44,13 +55,14 @@ export const UPlotChart: React.FC<PlotProps> = props => {
   }, [plotInstance]);
 
   useLayoutEffect(() => {
+    const data = props.data.frame.fields.map(f => f.values.toArray()) as AlignedData;
+    const shouldUpdate = !isEqual(plotData?.current?.data, data);
```
dprokop, line 59: This is a pottential bottleneck

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -30,7 +28,20 @@ export const UPlotChart: React.FC<PlotProps> = props => {
...
 
     pluginLog('UPlotChart: init uPlot', false, 'initialized with', plotData.current, currentConfig);
-    const instance = new uPlot(currentConfig, plotData.current, canvasRef.current);
+    const instance = new uPlot(
+      {
+        ...currentConfig,
+        hooks: {
+          init: [
+            u => {
+              u.root.setAttribute('data-testid', 'uplot-root');
```
torkelo, line 37: is this used anywhere?

dprokop: Nope, forgot to remove.

packages/grafana-ui/src/components/uPlot/Plot.tsx
torkelo: not sure I understand this, if we are in this useLayoutEffect data has always changed, should be no need to do this and this could be very expensive no? 

of should be super easy to just check if data is different from previous? Ah I see this function changed from last I edited, it used to only depend on data, now on data and config. Then we need to track previous data as well so we can just do a instance difference check, should be no need for isEqual here, just normal !== check from previous

torkelo (commented):
This looks more like a test of Plot.tsx internals than the logic in GraphNG

For a GraphNG unit test would we not just want to test that the correct props are sent to the Plot component?

dprokop:
Closing in favor of https://github.com/grafana/grafana/pull/29496

Issue #26383, opened 2020-07-16
Table: Allow table cells to calculate min and max based on column's values


jutley:
**What would you like to be added**:

When using table cells with gauge displays, I would like the automatically calculated "max" to be based on the values in the cells column, not the table as a whole.

**Why is this needed**:

In any table, it is reasonable to expect the values in each column to represent completely different concepts. For example, we may want each row to represent a Kubernetes Pod, and have columns to show CPU usage and memory usage. These two values are drastically different in magnitude. CPU is measured in cores, and may be in the range of 0-4. Memory is measured in bytes, and may the range of 0-4*1024^3. 

This means that the automatically calculated "Max" may be 4*1024^3. A gauge representing CPU usage will look identical for both very low CPU (0.01) and very high CPU (4). This makes the bar gauge functionally useless in this context. 

Right now, the best we can do is to define the max manually, but this is very finicky. If set too low, the gauges will look identical for any value above the max, which is misleading. If it is too high, it tends to waste a lot of horizontal space. Plus, the max in one context may be completely inappropriate for another context.

**Example**:

Here is an example that illustrated what is described above. You can see that while the CPU Usage values differ by orders of magnitude, they all appear exactly the same. This is because the very large values for "Memory Usage" are determining the max value for the "CPU Usage" bar gauges.

![image](https://user-images.githubusercontent.com/5660346/91223363-5ebba600-e6d5-11ea-82b0-86d3f6eb958f.png)

weibeld:
This feature would be extremely valuable. Currently, we're effectively restricted to a single column per table with numerical values when using the gauge display mode (unless the values in all columns are in a similar range, which is often not the case). So, I hope this feature will actually make it to the 7.4.0 release.

torkelo:
@ryantxu currently the BarGaugeCell calls getFieldConfigWithMinMax with local: false parameter, we can fix this issue and get local (field) level min & max for bar gauge in table. 

Feels like the we can just change that , but the question is should there be an option for this and is this an ok breaking change? My thinking is yes.

MAFLO321:
Any updates on this?

MAFLO321:
@torkelo @ivanahuckova Any updates on this?

MAFLO321:
@torkelo @ivanahuckova Any updates on this? Maybe move to a different milestone.

imReker:
😢  Still a problem on Grafana v9.0, maybe add to v100.0 milestone?
[...]

PR #29404, opened 2020-11-26
Min/Max: move auto out of default apply overrides

ryantxu:
I'm not sure about this one... :smirk:  it looks like calculating min/max was recently moved from fieldDisplay to fieldOverrides, this PR moves things back.  I suspect there was a good reason to move it, so this may be a bad idea.  My motivation for the change is because we want the auto min/max to be tied to unique units.  

I know understand why we needed the special handling to avoid updates on changes to min/max! 

Below are three series, two share share the same (default) units and a max 40, the third has its own units and auto range.  This was not possible when applying the min/max across all numeric fields

image

image

If we do need to keep auto min/max in `applyFieldOverrides` that should be modified so that it:
* is calculated per unit
* is not optional -- it seems weird to have an optional setting that is required in all execution paths

### DIFF
--- a/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
+++ b/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
@@ -65,7 +65,7 @@ export const GraphNG: React.FC<GraphNGProps> = ({
   const hasLegend = useRef(legend && legend.displayMode !== LegendDisplayMode.Hidden);
   const alignedFrame = alignedFrameWithGapTest.frame;
   const compareFrames = useCallback(
-    (a: DataFrame, b: DataFrame) => compareDataFrameStructures(a, b, ['min', 'max']),
+    (a: DataFrame, b: DataFrame) => compareDataFrameStructures(a, b, []), //  ['min', 'max']),
     []
   );
   const configRev = useRevision(alignedFrame, compareFrames);
--- a/packages/grafana-ui/src/components/uPlot/config/UPlotScaleBuilder.ts
+++ b/packages/grafana-ui/src/components/uPlot/config/UPlotScaleBuilder.ts
@@ -1,5 +1,5 @@
 import isNumber from 'lodash/isNumber';
-import { Scale } from 'uplot';
+import uPlot, { Scale, Range } from 'uplot';
 import { PlotConfigBuilder } from '../types';
 
 export interface ScaleProps {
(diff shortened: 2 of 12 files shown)


packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -65,7 +65,7 @@ export const GraphNG: React.FC<GraphNGProps> = ({
   const hasLegend = useRef(legend && legend.displayMode !== LegendDisplayMode.Hidden);
   const alignedFrame = alignedFrameWithGapTest.frame;
   const compareFrames = useCallback(
-    (a: DataFrame, b: DataFrame) => compareDataFrameStructures(a, b, ['min', 'max']),
+    (a: DataFrame, b: DataFrame) => compareDataFrameStructures(a, b, []), //  ['min', 'max']),
```
ryantxu, line 68: ignoring min/max should not be necessary since they should not update with every query

packages/grafana-ui/src/components/uPlot/config/UPlotScaleBuilder.ts
```diff
@@ -11,12 +10,21 @@ export interface ScaleProps {
 
 export class UPlotScaleBuilder extends PlotConfigBuilder<ScaleProps, Scale> {
```
ryantxu, line 11: @leeoniya -- can you sanity check this function?

packages/grafana-ui/src/components/uPlot/config/UPlotScaleBuilder.ts
```diff
@@ -11,12 +10,21 @@ export interface ScaleProps {
...
+        },
+      };
+    }
     return {
       [scaleKey]: {
-        time: !!isTime,
-        range,
+        range: (u: uPlot, initMin: number, initMax: number) => {
+          const { min, max } = this.props;
+          const [smin, smax] = uPlot.rangeNum(initMin, initMax, 0.05 as any, true);
```
leeoniya: if your know that you have an explicitly-set min or max, it probably makes sense to pass these to the rangeNum instead of the dataMin/Max, this way it expands the padding relative to the final range rather than the data range.

```
uPlot.rangeNum(min ?? initMin, max ?? initMax, 0.05 as any, true);
```

also 0.05 padding factor is pretty tight, the default is 0.1, but your call.

torkelo:
I was thinking of also linking it to unit , but that way it won't always solve this problem right? https://github.com/grafana/grafana/issues/26383 (unless table fields have different units of course)

torkelo:
It sure why it was moved out of fieldOverrides? Field display is only used by she stat panels , so won't this break things for say the table panel?

ryantxu:
this approach makes it too hard for panels to get what they need consistently

Issue #4623, opened 2016-04-08
[Feature Request] Add option to hide 0-valued series from tooltip when Tooltip Mode = All series


toddwh50:
When graph is comprised of "sparser" data where some times do not have values for all series (see image), it would be useful to be able to check a "Hide 0-Valued Series in Tooltip" option to cause these to be excluded from the tooltip. Then, only series having values would remain.

![image](https://cloud.githubusercontent.com/assets/17885456/14398537/89c207d6-fdaa-11e5-85f3-58e843cf3fce.png)

scraly:
Hi,
I still have the issue with grafana 3.1.0 version for "All series" tooltip mode.
![grafana_tooltip_0_issue](https://cloud.githubusercontent.com/assets/2543381/17176145/9bf3e07a-540b-11e6-8a79-51f82b4017b6.png)

mrares:
+1

ldktta:
+1

scraly:
+1

scraly:
![grafana_0_value_402](https://cloud.githubusercontent.com/assets/2543381/21257083/3abe33d4-c376-11e6-8da2-d8794c5cd7b8.png)
Problem exists again and again :-).

scraly:
Hi, is it possible to have a look at this? Thanks

mrares:
I "solved" it by sorting the tooltip, any >0 values appear on top... still the UI is bad.

ivanpricewaycom:
sorting desc by tooltip is kinda ok, but very large tooltips can run above the window. it seems like this should be fairly easy to 'fix' around line ~130 of this file:

https://github.com/grafana/grafana/blob/8b25ba6a7c8c26191704ddfe77dec524072cf13c/public/app/plugins/panel/graph/graph_tooltip.js

but i can't work out how to rebuild just the JS, any tips / pointers ?

james-johnston-thumbtack:
+1

scraly:
Any news? :-)

jaune-rouge:
+1

amiralibazazi:
+1

jdefaver:
+1

starcraft66:
+1

machytkafitanalytics:
+1

ebolstad:
+1

oleg-glushak:
+1

lambdaq:
same here. I try not to type +1 in plain, but still, +1.

sunilkukde:
[...]

Issue #3094, opened 2015-10-30
Feature Request: Option to hide null / zeroe values in  tooltip


linar-jether:
Related to this: https://github.com/grafana/grafana/issues/1381

An option to hide null / zero values from the tool tip would be great, both for stacking (when we must use null as zero) and for intermittent graphs, such as points only or partial graphs like the ones below,  which still show in the tool tip even when there is no data in that position

![image](https://cloud.githubusercontent.com/assets/11908736/10840619/c640949c-7ee9-11e5-8db5-b579ffc0ea82.png)

![image](https://cloud.githubusercontent.com/assets/11908736/10840702/8e757e0a-7eea-11e5-8732-a2098bb03db4.png)

torkelo:
there already is an option for hide null, (called hide empty)

torkelo:
will close as there already is an issue for this :)

https://github.com/grafana/grafana/issues/1381

linar-jether:
Hide empty only hides the series if it's all nulls, but if only a part of the series is not null then it stay on the tooltip for all points (even null points, as in the first picture)

Issue #1381, opened 2015-01-23
Feature Request: Option to hide zero values in legend and tooltip


jmcle:
The new options to hide null values in the legend and 'all series' tooltip are excellent but I run into a problem when using them with stacked bar graphs.

With stacked graphs we need to set 'null as zero' in order to get the stacking to work properly but this negatively impacts the ability to hide null values in the legend and tooltip as null values will show up as zero instead of null.

I'm not sure if there's a better way to address this but one option would be to add an option to 'hide zero values' in addition to the existing 'hide null values'. Another option would be to make the 'hide value' option a parameter which you could set to any numeric value. This would be similar to what the 'removeBelowValue' function does but in this case it would affect the display of legends and tooltips.

fzaire:
I have the same issue and I end up with 80% of the metrics in my legend with 0 value that i need to hide.
If more people are facing the same issue, we could try to commit this feature.

torkelo:
cant you filter out the zero metrics/series in your query?

linar-jether:
+1 for this issue, as we can't filter out zeroes without breaking stacks, i.e 
![image](https://cloud.githubusercontent.com/assets/11908736/10841685/57bd3704-7ef4-11e5-803c-e71854815769.png)

agourlay:
+1 for this issue.

apakulov:
+1

utkarshcmu:
+1

digrich:
+1

george-vieira:
I have the same issue, my metrics shows traffic by IP and they spike in and out so at any point in time there could be values or 0. The tooltip shows 100 IPs which is overall but I only care of the point in time I'm hovering over.

utkarshcmu:
The above PR only hides the zero values if all datapoints are 0. Even if a single datapoint is non-zero on the graph for a given time range, it will not hide the metric.

Pryz:
+1

It could be nice to specify a pattern. Like "hide if < 0" or "hide if = 0" etc

utkarshcmu:
As an example, Opentsdb cannot remove zeros in query itself, so it would be great if Grafana can eliminate them from Legend and Tooltip. And looks like a lot of people wants this feature to be there in Grafana.

@torkelo Are these +1s enough for #3336 to get merged? Please let me know if any refinement is needed in the PR #3336.

dacr:
+1

with a pattern such as "hide when no value" (in current time range)

torkelo:
[...]

PR #29405, opened 2020-11-26
Graph: Nonnull tooltip hover mode

kvc0:
## Add a Nonnull tooltip hover mode

This adds a Nonnull option to hover tooltip modes.
Nonnull hides series in a tooltip when they would be displayed as null.
This is useful when you have sparsely reported series and only care which values were reported at a given time column.

All tests pass.

image

This new option hides series with null value at the currently hovered time column from the data tooltip in graph panels.

When a time series has sparse data - e.g., reporting infrequent errors by `device_id`, and potentially _many_ distinct series across a dashboard's time range - e.g., a `device_id`, a tooltip becomes quickly unreadable if you don't sort by value descending, or unresponsive if you sort 10,000 mostly-null series by value descending for every pixel the mouse moves.

This gives dashboard authors the ability to choose whether null valued series show up in a hover tooltip.

There are many "closed" issues from people who appear to be asking for this but they are either misunderstood or fall into the trap of also mentioning the existing feature of "hide a series from the tooltip when ALL of its values happen to be 0 or null" which is different from this PR which focuses on sparse data.

Relates to:
https://github.com/grafana/grafana/issues/3094
https://github.com/grafana/grafana/issues/1381
https://github.com/grafana/grafana/issues/4623

This PR does not include a "Nonzero" option and it does not affect what is in the legend, so it does not seem to satisfy every aspect of each of these (and myriad other loosely related) issues.

### DIFF
--- a/dashboard-schemas/panels/Graph.cue
+++ b/dashboard-schemas/panels/Graph.cue
@@ -120,10 +120,12 @@ package panels
 	timeShift: string
 	// Tooltip settings.
 	tooltip: {
-		// * true - The hover tooltip shows all series in the graph.  Grafana
+		// * 1 - The hover tooltip shows all series in the graph.  Grafana
 		// highlights the series that you are hovering over in bold in the series
 		// list in the tooltip.
-		// * false - The hover tooltip shows only a single series, the one that you
+		// * 2 - The hover tooltip shows all series in the graph but filters out
+		// null values.
+		// * 0 - The hover tooltip shows only a single series, the one that you
 		// are hovering over on the graph.
 		shared: bool | *true
 		// * 0 (none) - The order of the series in the tooltip is determined by the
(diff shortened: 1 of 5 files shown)


oddlittlebird:
Doc change looks fine to me.

A dev will review the code when time allows.

WarriorOfWire:
Per stalebot, I’m pinging for review with a status update:  Ready for review.

WarriorOfWire:
This PR from November remains current and awaiting review.  Does stalebot need its thresholds updated to meet the expectation of review latency in this repository?

torkelo:
Closing this as we do not want to modify the old graph panel any more, 

Here is the new tooltip for the new time series panel:
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/uPlot/plugins/TooltipPlugin.tsx

anacrolix:
The new time series panel doesn't seem to have any tooltip options for the proposed mode.

WarriorOfWire:
@anacrolix yup. Grafana has ensured that I won’t port this feature to the new panel, but I’d be happy if someone else were to conjure the arcane incantations necessary to merge a minor feature pr into it!

PR #29406, opened 2020-11-26
MinMax: keep global min/main in field state

ryantxu:
Alternative to https://github.com/grafana/grafana/pull/29404

Below are three series, two share share the same (default) units and a max 40, the third has its own units and auto range.  This was not possible when applying the min/max across all numeric fields

image

image


packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -117,7 +117,6 @@ export interface ApplyFieldOverrideOptions {
   getDataSourceSettingsByUid: (uid: string) => DataSourceInstanceSettings | undefined;
   theme: GrafanaTheme;
   timeZone?: TimeZone;
-  autoMinMax?: boolean;
```
ryantxu, line 120: Lets make this always save the state rather than explicitly enabled

packages/grafana-data/src/field/fieldDisplay.ts
```diff
@@ -77,7 +77,6 @@ export interface GetFieldDisplayValuesOptions {
   replaceVariables: InterpolateFunction;
   sparkline?: boolean; // Calculate the sparkline
   theme: GrafanaTheme;
-  autoMinMax?: boolean;
```
ryantxu, line 80: this was no longer used

packages/grafana-ui/src/components/Table/BarGaugeCell.tsx
```diff
@@ -20,7 +21,7 @@ const defaultScale: ThresholdsConfig = {
 export const BarGaugeCell: FC<TableCellProps> = props => {
   const { field, column, tableStyles, cell, cellProps } = props;
 
-  let { config } = field;
+  let config = getFieldConfigWithMinMax(field, false);
```
ryantxu, line 24: 'false' will use the global min/max -- that matches the current behavior.  However, we likely want a panel option that will toggle this from the whole table vs the single column

packages/grafana-ui/src/components/uPlot/config/UPlotScaleBuilder.ts
```diff
@@ -11,12 +10,21 @@ export interface ScaleProps {
...
+        },
+      };
+    }
     return {
       [scaleKey]: {
-        time: !!isTime,
-        range,
+        range: (u: uPlot, dataMin: number, dataMax: number) => {
+          const { min, max } = this.props;
+          const [smin, smax] = uPlot.rangeNum(dataMin, dataMax, 0.1 as any, true);
```
leeoniya, line 25: @ryantxu 

`uPlot.rangeNum(min ?? dataMin, max ?? dataMax, 0.1 as any, true);`

torkelo (approved):
Tested this and I think this looks good

PR #29407, opened 2020-11-26
Variables: Fixes Constant variable persistence confusion

### LINKED ISSUE #18916: Constant Variable Should Never Be in Current
**What happened**:
If you create a constant variable its value is still saved in current variable values (or hower are those called), so when you change the constant value, that value isn't changed in current var values after a save if `Save current variables` isn't selected. It's not intuitive that constants are saved in current and just complicates things. The "fix" of selecting `Save current variables` when saving a dashboard is far from intuitive as well.
Constants shouldn't have current variable values and they should always use the value that is specified in the variable itself.

When I change the value of a constant it should instantly change and work after saving a dashboard.

Create a constant, set some value and save the dashboard selecting `Save current variables`. Use that variable somewhere that you can see it on the dashboard, e.g. in a title of a graph. Now change the value of that constant and save the dashboard again without selecting `Save current variables`. Refresh the dashboard from a browser (F5) and the variable value which is saved in the var won't be used, but the value saved in current.

**Environment**:
- Grafana version: 6.2.1
- User OS & Browser: Linux/Firefox
- Grafana plugins:
- Others:

### LINKED ISSUE #28428: [Feature Request]: Minimize the confusion regarding saving/updating Variables
When a user forgets to check the __Save current variables__ toggle/checkbox the resulting state of the dashboard when refreshed is unexpected. Grafana needs to make this flow less error prone and easier to understand for the user.

Fixes: #18916

Related issues:
#20703
#20600
#27158
#28089
#26874
#21870
#16600
#15006
#13320
#10748

**Save Dashboard Modal in different versions**
__6.4.3__
image

__Grafana v7.4.0-pre (8032b43838)__
image

hugohaggmark:
As described in #18916 and in #28428. 

Fixes #18916
Fixes #28428

#### Constant variables
 
In order to minimize the confusion with Constant variable usage, we've removed the ability to make Constant variables visible. This change will also migrate __`all`__ existing __`visible`__ Constant variables to Textbox variables because which we think this is a more appropriate type of variable for this use case.


public/app/features/variables/state/sharedReducer.test.ts
```diff
@@ -16,38 +15,78 @@ import {
...
+      };
+
+      const payload = toVariablePayload({ id: 'name from model', type: 'query' }, { global: true, index: 0, model });
 
       reducerTester<VariablesState>()
         .givenReducer(sharedReducer, { ...initialVariablesState })
         .whenActionIsDispatched(addVariable(payload))
         .thenStateShouldEqual({
-          [0]: {
-            ...lodashDefaults({}, model, initialQueryVariableModelState),
```
hugohaggmark, line 46: I changed this test because the expect is actually implementing the reducer logic.

mckn (approved):
LGTM!

PR #29411, opened 2020-11-26
Build: move canary packages to github

### LINKED ISSUE #27953: Build: move canary channel/tag from npm to github registry
Currently we publish a new package to the canary tag/channel in the npm registry on every commit to master. This is really useful when developing plugins and you want to stay in sync with the latest changes in master. Especially close to shipping a new major/minor release.

To make this possible we have a script that is kind of error prone. The reason is that we need to deprecate/unpublish npm packages in the canary tag/channel before publishing a new package to prevent hitting the npmjs registry limit. At multiple time this logic has failed and as a result the whole build in master will fail.

**Why is this needed**:
Earlier this year @dprokop looked into moving the canary tag/channel from npmjs to the github npm registry but that feature of github was not ready to be used since it was in beta. Now it is released as a stable version and what I can see free for public projects. So the suggestion is to move the canary tag/channel to github npm registry to prevent hitting any upper limit and there for simplify the build pipline in master.

https://github.com/features/packages

jackw:
We need to migrate the canary channel over to use github packages to prevent hitting the npmjs registry limit. This PR updates the publish canary tasks / scripts to publish the `@grafana/<package>` to github packages and adds some documentation to help consumers understand how to install canary packages from github.

Once this PR is approved we can add a [custom deprecation message](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions#deprecating-a-single-version-of-a-package) to all canary packages currently on npmjs explaining we have migrated to github.

Fixes #27953


.drone.yml
```diff
@@ -215,6 +215,19 @@ steps:
   - test-backend
   - test-frontend
 
+- name: release-next-npm-packages
+  image: grafana/build-container:1.2.30
+  commands:
+  - ./node_modules/.bin/lerna bootstrap
```
mckn, line 221: nit: Not sure we need to run this again. I think the node_modules should already be setup in place.

jackw: Yeah yarn workspaces should have sorted all that out for us. Will delete it. Thanks @mckn! 🙏

torkelo (approved):
Think this looks good

packages/README.md
```diff
@@ -27,7 +27,29 @@ Every commit to master that has changes within the `packages` directory is a sub
...
+
+1. You must use a personal access token to install packages from Github. To create an access token [click here](https://github.com/settings/tokens) and create a token with the `read:packages` scope. Make a copy of the token.
+2. Create / modify your `~/.npmrc` file with the following:
+
+```
+@grafana:registry=https://npm.pkg.github.com
+//npm.pkg.github.com/:_authToken={INSERT_GH_TOKEN_HERE}
+```
+
+3. Update the package.json of your project to use either the `canary` channel or a version of the `canary channel
```
dprokop, line 44:
```suggestion
3. Update the package.json of your project to use either the `canary` channel or a version of the `canary` channel
```

scripts/pr.star
```diff
@@ -50,6 +51,7 @@ def pr_pipelines(edition):
         build_docker_images_step(edition=edition, ver_mode=ver_mode, archs=['amd64',]),
         postgres_integration_tests_step(),
         mysql_integration_tests_step(),
+        release_canary_npm_packages_step(edition),
```
dprokop, line 54: Don't forget to revert changes in this file before merging to master :)

dprokop (approved):
LGTM, great progress @jackw

scripts/pr.star
jackw: Thanks for the reminder. Reverted it! 👍

PR #29414, opened 2020-11-26
PanelEvents: Refactors and removes unnecessary events, fixes panel editor update issue when panel options change

### LINKED ISSUE #27464: PanelEdit: PanelProps onOptionsChange doesn't immediately update panel settings UI
**What happened**:
When reflecting changes back to a panel's options using `onOptionsChange`, the UI component in the settings side pane for the changed option will not update immediately. If you refresh the dashboard (or presumably any other action that re-renders the UI), the option will then render correctly. 
Calling `onOptionsChange` within a panel should immediately refresh the settings UI to reflect the correct state. 

Minimal example using the simple-react-panel template
https://github.com/mure/grafana-options-bug 

**Environment**:
- Grafana version: 7.1.3
- Data source type & version:
- OS Grafana is installed on: macOS
- User OS & Browser: macOS Chrome
- Grafana plugins:
- Others:

torkelo:
Fixes #27464

This also removes and unifies some events that were only used to notify DashboardGrid that the panels array changed. A lot of this could be fixed by moving more state to redux or immutable react state. 

Had to add one new event for when panels change options so that the PanelEditor will know. 

Updated some events to the new event model.

### DIFF
--- a/public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
+++ b/public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
@@ -2,7 +2,7 @@ import React, { PureComponent } from 'react';
 import { connect, MapDispatchToProps, MapStateToProps } from 'react-redux';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import { css, cx } from 'emotion';
-import { Unsubscribable } from 'rxjs';
+import { Subscription } from 'rxjs';
 
 import { FieldConfigSource, GrafanaTheme, PanelPlugin } from '@grafana/data';
 import { selectors } from '@grafana/e2e-selectors';
@@ -36,6 +36,7 @@ import { CoreEvents, LocationState, StoreState } from 'app/types';
 import { DisplayMode, displayModes, PanelEditorTab } from './types';
 import { VariableModel } from 'app/features/variables/types';
 import { DashboardModel, PanelModel } from '../../state';
+import { PanelOptionsChangedEvent } from 'app/types/events';
 
 interface OwnProps {
   dashboard: DashboardModel;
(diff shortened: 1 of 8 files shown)


public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -113,8 +129,9 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
   };
 
   onPanelOptionsChanged = (options: any) => {
+    // we do not need to trigger force update here as the function call below
+    // fires PanelOptionsChangedEvent which we subscribe to above
     this.props.panel.updateOptions(options);
```
dprokop, line 134: 👍

public/app/features/dashboard/dashgrid/DashboardGrid.tsx
```diff
@@ -102,26 +102,17 @@ export interface Props {
...
 
   componentDidMount() {
     const { dashboard } = this.props;
-
-    dashboard.on(panelAdded, this.triggerForceUpdate);
-    dashboard.on(panelRemoved, this.triggerForceUpdate);
-    dashboard.on(CoreEvents.repeatsProcessed, this.triggerForceUpdate);
-    dashboard.on(CoreEvents.rowCollapsed, this.triggerForceUpdate);
-    dashboard.on(CoreEvents.rowExpanded, this.triggerForceUpdate);
+    this.eventSubs.add(dashboard.events.subscribe(DashboardPanelsChangedEvent, this.triggerForceUpdate));
```
dprokop, line 111: Lovely!

dprokop (approved):
I think this looks great code-wise. Didn't have a chance to test this yet, adding @ryantxu  as a reviewer to confirm it addresses the issues he has encountered.

PR #29416, opened 2020-11-26
Plugins: Add support for includes' icon

joanlopez:

I was developing a Grafana application plugin when I realized that there's no support for includes' icon.

However, the [`PluginInclude`](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/plugin.ts#L91) type already has the `icon` attribute and the [`DropDownChild`](https://github.com/grafana/grafana/blob/master/public/app/core/components/sidemenu/DropDownChild.tsx) component already has support for displaying an icon. See on the image below that `NavLink`s other than plugin includes are already drawing the icons.

screenshot1 

So, I introduced the minor changes needed to add support for includes' icon.

Note that I only added support for `page` includes, but it could make sense to add it for other types of includes.
Also, I added the `icon` field into the [`plugin.json`](https://grafana.com/docs/grafana/latest/developers/plugins/metadata/#includes) reference, but more changes might be required.

Thanks!!

### DIFF
--- a/docs/sources/developers/plugins/metadata.md
+++ b/docs/sources/developers/plugins/metadata.md
@@ -68,6 +68,7 @@ Plugin dependencies.
 | `name`       | string  | No       |             |
 | `role`       | string  | No       |             |
 | `type`       | string  | No       |             |
+| `icon`       | string  | No       |             |
 
 ## info
 
--- a/pkg/api/index.go
+++ b/pkg/api/index.go
@@ -104,6 +104,7 @@ func getAppLinks(c *models.ReqContext) ([]*dtos.NavLink, error) {
 						Text: include.Name,
 					}
 				}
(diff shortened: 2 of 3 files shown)


aknuds1 (approved):
Could a test be added for this? Asking since I'm not familiar with the touched code :)

joanlopez:
> Could a test be added for this? Asking since I'm not familiar with the touched code :)

I figured out the complete `getAppLinks` function has not tests at all (in fact it's a private one) neither the `index.go` file. So, I cannot see an easy way to add tests specifically for these changes other than working on coverage for the whole code (multiple structures and functions involved).

However, I'm completely open to any proposal 😇  Thanks!!!

torkelo (approved):
LGTM

PR #29419, opened 2020-11-26
QueryEditors: Refactoring & rewriting out dependency on PanelModel

torkelo:
* [x] Fix query inspector
* [x] Fix query options (max data points etc)
* [x] Testing 
* [x] Explore data source support 
* [x] Fix dashboard (shared) query editor 

TestStuffPage (example page to test it out): 122 lines total incl state, query execution & data viz:

Screenshot from 2020-11-27 17-11-00


public/app/features/query/components/QueriesTab.tsx
```diff
@@ -96,85 +101,62 @@ export class QueriesTab extends PureComponent<Props, State> {
...
-      panel.targets.forEach(target => {
-        if (target.datasource !== ExpressionDatasourceID) {
-          target.datasource = panel.datasource;
-          if (!target.datasource) {
-            target.datasource = config.defaultDatasource;
+      for (const query of queries) {
+        if (query.datasource !== ExpressionDatasourceID) {
+          query.datasource = query.datasource;
+          if (!query.datasource) {
+            query.datasource = config.defaultDatasource;
```
dprokop, line 118: Is it OK to mutate query here? Shouldn't we first make a copy of the `queries` object?

public/app/features/query/components/QueriesTab.tsx
```diff
@@ -273,22 +257,21 @@ export class QueriesTab extends PureComponent<Props, State> {
...
-    const { panel, dashboard } = this.props;
+    const { onQueriesChange, queries, onRunQueries } = this.props;
     const { dataSourceItem, data } = this.state;
 
-    if (isSharedDashboardQuery(dataSourceItem.name)) {
-      return <DashboardQueryEditor panel={panel} panelData={data} onChange={query => this.onUpdateQueries([query])} />;
-    }
+    // if (isSharedDashboardQuery(dataSourceItem.name)) {
+    //   return <DashboardQueryEditor panel={panel} panelData={data} onChange={query => this.onUpdateQueries([query])} />;
+    // }
```
dprokop, line 265:
```suggestion
```

dprokop (changes requested):
@torkelo amazing effort on this one. Code wise I have only one question that I've left in the review.

When testing this I have encountered an error though that's easy to reproduce:
1. Go to Elastic data source tests dev dashboard (i've tested v6 and v7) 
2. Try to edit any of the panels
3. The following error occurs:
image

FYI - I do not have Elastic data source running when performing the test above. I have also checked this flow against the master and it's working fine.

public/app/features/query/components/QueriesTab.tsx
torkelo: this was from the old code, so no change here. But yes this should be done immutably

public/app/features/query/components/QueriesTab.tsx
torkelo: ah, forgotten to fix this part. will add to the todo

peterholmberg (commented):
Works for me! Only one question.

public/app/features/query/components/QueriesTab.tsx
```diff
@@ -96,85 +101,62 @@ export class QueriesTab extends PureComponent<Props, State> {
...
     if (newDsItem.meta.mixed) {
-      // Set the datasource on all targets
-      panel.targets.forEach(target => {
-        if (target.datasource !== ExpressionDatasourceID) {
-          target.datasource = panel.datasource;
-          if (!target.datasource) {
-            target.datasource = config.defaultDatasource;
+      for (const query of queries) {
+        if (query.datasource !== ExpressionDatasourceID) {
+          query.datasource = query.datasource;
```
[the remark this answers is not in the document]
torkelo: not sure, this was copied from the old one.

mckn (commented):
I really like this refactoring! Makes it so much easier to follow what parts of the PanelModel is being used by leafs in the tree. Great work!

I saw that there was some commented code that might need to be address but @dprokop already have added a comment about that one.

public/app/features/query/components/QueriesTab.tsx
```diff
@@ -24,14 +21,23 @@ import {
...
-  dashboard?: DashboardModel;
+  queryRunner: PanelQueryRunner;
+  queries: DataQuery[];
+  dataSourceName: string | null;
+  options: QueryGroupOptions;
+  onOpenQueryInspector?: () => void;
+  onRunQueries: () => void;
+  onQueriesChange: (queries: DataQuery[]) => void;
+  onDataSourceChange: (ds: DataSourceSelectItem, queries: DataQuery[]) => void;
+  onOptionsChange: (options: QueryGroupOptions) => void;
```
torkelo, line 40: thinking of moving queries & dataSourceName into the QueryGroupOptions to simplify this interface a bit,

we could then get rid of onDataSourceChange and onQueriesChange as we can cover that with onOptionsChange. Thoughts?

torkelo: ok, going to merge this and do this in a follow-up PR

PR #29422, opened 2020-11-26
Add PPL support configuration setting

### LINKED ISSUE #28674: Feature Request: PPL Query Syntax support for Elasticsearch Plugin
**What would you like to be added:**

This feature request proposes to extend Grafana’s Elasticsearch data source plugin with Piped Processing Language (PPL) support. [Pipe Processing Language](https://opendistro.github.io/for-elasticsearch-docs/docs/ppl/) is a semi-structured query language using a set of commands delimited by a vertical bar (|), and it can be used to query Elasticsearch when PPL plugin is installed for Elasticsearch. Adding PPL support will provide users with an alternative query syntax to Lucene, the only query syntax currently supported in the Elasticsearch data source plugin.
 
**Use Cases:**

The Elasticsearch plugin currently only supports Apache Lucene query syntax. PPL has an arguably more intuitive and flexible syntax than Lucene, and many users are already familiar with the [Unix pipe-style syntax](https://en.wikipedia.org/wiki/Pipeline_(Unix)). Particularly, PPL makes it easier to construct ad hoc queries, which would improve the user experience in Grafana Explore. 

Example of PPL query syntax:

```
source=accounts
| where age > 18
| fields firstname, lastname
```

**Proposal:**

PPL support could be optionally turned on as an Elasticsearch data source configuration option. When enabled, users will be able to choose between Lucene and PPL query syntax in the query editor. Different editor interfaces could be shown depending on the selected query syntax option. The end behavior would be the same for both query syntax choices—two functionally equivalent queries in different syntax should produce the same output.

Note that PPL queries are sent to a different endpoint exposed by the Elasticsearch plugin framework instead of the standard _search endpoint. Furthermore, the response structure is in [JDBC format](https://opendistro.github.io/for-elasticsearch-docs/docs/sql/protocol/#jdbc-format) and will need to be mapped to the required data frame structure. These would need to be addressed when implementing this feature.

More information regarding PPL query syntax and its commands can be found here: [PPL Reference Manual](https://github.com/opendistro-for-elasticsearch/sql/blob/develop/docs/experiment/ppl/index.rst)

cc @alolita @ahlaw

ahlaw:

This PR is one of several PRs targeting the `elasticsearch-ppl-support` branch, which should ultimately contain all the changes required for basic PPL support from the Elasticsearch plugin on the client-side. Basic PPL support entails being able to write PPL queries in the query editor and visualize responses from Elasticsearch instances with the ODFE SQL plugin installed. Expected functionality such as variable interpolation and ad hoc filtering should also work with PPL queries. Features that are out of scope for the `elasticsearch-ppl-support-frontend` branch include alerting on PPL queries and the option to use PPL queries in the dashboard settings menu.

This PR adds the option to enable PPL support under the Elasticsearch Details section of the Elasticsearch data source configuration menu.

image

This setting is only available when the Elasticsearch version is set 7.0+ as PPL is only available in the opendistro_sql plugin version that goes with Elasticsearch version 7.9.1+ (see [plugin compatibility](https://opendistro.github.io/for-elasticsearch-docs/docs/install/plugins/#plugin-compatibility) for details). This setting is then made available to the `ElasticDatasource` class, which can then be used to control whether the user has the option of writing PPL queries.

Partially resolves #28674

cc: @alolita

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/configuration/ConfigEditor.test.tsx
+++ b/public/app/plugins/datasource/elasticsearch/configuration/ConfigEditor.test.tsx
@@ -25,8 +25,9 @@ describe('ConfigEditor', () => {
     //@ts-ignore
     delete options.jsonData.timeField;
     delete options.jsonData.maxConcurrentShardRequests;
+    delete options.jsonData.pplSupportEnabled;
 
-    expect.assertions(3);
+    expect.assertions(4);
 
     mount(
       <ConfigEditor
@@ -34,6 +35,7 @@ describe('ConfigEditor', () => {
           expect(options.jsonData.esVersion).toBe(5);
           expect(options.jsonData.timeField).toBe('@timestamp');
           expect(options.jsonData.maxConcurrentShardRequests).toBe(256);
+          expect(options.jsonData.pplSupportEnabled).toBe(false);
         }}
         options={options}
       />
@@ -41,7 +43,7 @@ describe('ConfigEditor', () => {
   });
 
   it('should not apply default if values are set', () => {
-    expect.assertions(3);
+    expect.assertions(4);
 
     mount(
       <ConfigEditor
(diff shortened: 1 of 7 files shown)


ahlaw:
Closing this as configuration is no longer a requirement

Issue #29201, opened 2020-11-18
Chore: enable eslint-plugin-react rules 


dprokop:
Rules to be enabled in Grafana core (.eslintrc):

- [x] "react/display-name": "off", https://github.com/grafana/grafana/pull/29428
- [x] "react/no-deprecated": "off", https://github.com/grafana/grafana/pull/29428
- [x] "react/no-find-dom-node": "off", https://github.com/grafana/grafana/pull/29428
- [x] "react/no-render-return-value": "off", https://github.com/grafana/grafana/pull/29519
- [x] "react/no-unknown-property": "off", https://github.com/grafana/grafana/pull/29519
- [x] "react/no-children-prop": "off", https://github.com/grafana/grafana/pull/29062
- [x] "react/no-unescaped-entities": "off", https://github.com/grafana/grafana/pull/29062
- [x] "react/jsx-no-target-blank": "off", https://github.com/grafana/grafana/pull/29062

PR #29428, opened 2020-11-27
Chore: Enable eslint-plugin-react partial rules

chihsuan:

This PR enables the following eslint rules in Grafana core (.eslintrc) and fixed the corresponding lint issues:

- react/display-name
- react/no-deprecated
- react/no-find-dom-node

Partially for https://github.com/grafana/grafana/issues/29201

- I still use inline `// eslint-disable-next-line react/display-name` for some arrow functions since I think rewriting them by function declaration might not be a good idea for coding convention. Correct me, if I'm wrong.

### DIFF
--- a/packages/grafana-ui/src/components/ColorPicker/SeriesColorPickerPopover.tsx
+++ b/packages/grafana-ui/src/components/ColorPicker/SeriesColorPickerPopover.tsx
@@ -19,20 +19,22 @@ export const SeriesColorPickerPopover: FunctionComponent<SeriesColorPickerPopove
       customPickers={{
         yaxis: {
           name: 'Y-Axis',
-          tabComponent: () => (
-            <Switch
-              key="yaxisSwitch"
-              label="Use right y-axis"
-              className="ColorPicker__axisSwitch"
-              labelClass="ColorPicker__axisSwitchLabel"
-              checked={yaxis === 2}
-              onChange={() => {
-                if (onToggleAxis) {
-                  onToggleAxis();
-                }
-              }}
-            />
-          ),
+          tabComponent() {
+            return (
+              <Switch
+                key="yaxisSwitch"
+                label="Use right y-axis"
+                className="ColorPicker__axisSwitch"
+                labelClass="ColorPicker__axisSwitchLabel"
+                checked={yaxis === 2}
+                onChange={() => {
+                  if (onToggleAxis) {
+                    onToggleAxis();
+                  }
+                }}
+              />
+            );
+          },
         },
       }}
     />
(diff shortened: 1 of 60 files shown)


packages/grafana-ui/src/components/ColorPicker/SeriesColorPickerPopover.tsx
```diff
@@ -19,20 +19,22 @@ export const SeriesColorPickerPopover: FunctionComponent<SeriesColorPickerPopove
...
-              labelClass="ColorPicker__axisSwitchLabel"
-              checked={yaxis === 2}
-              onChange={() => {
-                if (onToggleAxis) {
-                  onToggleAxis();
-                }
-              }}
-            />
-          ),
+          tabComponent() {
```
dprokop, line 22: Interesting, is this related to the rules enabled in this PR?

packages/grafana-ui/src/components/OptionsUI/fieldColor.tsx
```diff
@@ -36,7 +36,9 @@ export const FieldColorEditor: React.FC<FieldConfigEditorProps<FieldColor | unde
       description: mode.description,
       isContinuous: mode.isContinuous,
       isByValue: mode.isByValue,
-      component: () => <FieldColorModeViz mode={mode} theme={theme} />,
+      component: function render() {
+        return <FieldColorModeViz mode={mode} theme={theme} />;
+      },
```
dprokop, line 41: Similar comment to the previous one- how does this approach help the fixes?

dprokop (approved):
@chihsuan thank you so much for this contribution. I went through the code and it seems to be OK! Left two questions I'm interested in getting answer for. Other than that it's awesome!

packages/grafana-ui/src/components/ColorPicker/SeriesColorPickerPopover.tsx
chihsuan: Yes, this is related to the `react/display-name` rule.

Since arrow function is  anonymous, the declared component will not have the displayName. So I changed this to using shorthand method function definition to creates a named function.

packages/grafana-ui/src/components/OptionsUI/fieldColor.tsx
chihsuan: This is same like the previous one. I used regular function declaration to have the displayName.
But I just found this could be also written in the shorthand method style. I've changed this to make it consistent.  Thanks!

packages/grafana-ui/src/components/ColorPicker/SeriesColorPickerPopover.tsx
dprokop: Perfect, thanks for the explanation:)

dprokop:
@chihsuan seems like the PR now have a conflict with the master. Can you please sync this with master and I think I can merge it if the build gets green :) Thank you!

chihsuan:
@dprokop done. please review, thanks!

PR #29432, opened 2020-11-27
prometheus: Improve IsAPIError's documentation

aknuds1:
Improve `prometheus.IsAPIError`s documentation, since we realized that `errors.As` is the correct function to use.


wbrowne (approved):
👍

PR #29433, opened 2020-11-27
grafana/ui: Migrate Field knobs to controls

peterholmberg:
Migrate the Field story to use controls instead of knobs.

For some reason Storybook doesn't infer the boolean props to the correct control. Can't figure out why?


packages/grafana-ui/src/components/Button/Button.story.tsx
```diff
@@ -20,15 +20,14 @@ export default {
...
   },
 };
 
-export const Simple: Story<ButtonProps> = ({ disabled, icon, children, size, variant }) => {
-  return (
-    <Button variant={variant} size={size} icon={icon} disabled={disabled}>
-      {children}
-    </Button>
-  );
+export const Simple: Story<ButtonProps> = ({ children, ...args }) => {
```
jackw, line 29: nit: could use arrow function return shorthand.

packages/grafana-ui/src/components/Forms/Field.story.tsx
```diff
@@ -1,62 +1,63 @@
...
-  const loading = boolean('Loading', false, BEHAVIOUR_GROUP);
-  const error = text('Error message', '', BEHAVIOUR_GROUP);
-
-  return { containerWidth, disabled, invalid, loading, error };
+  },
 };
 
-export const Simple = () => {
-  const { containerWidth, ...otherProps } = getKnobs();
+export const Simple: Story<FieldProps> = args => {
```
jackw, line 25: nit: could use arrow function return shorthand.

jackw (approved):
Nice one! 🚀 

Left a question about a missing prop and a couple of nitpicks.

PR #29434, opened 2020-11-27
Instrumentation: Trace all bus.DispatchCtx calls

bergquist:


wbrowne (approved):

torkelo:
An argument to keep the bus architecture?

bergquist:
@torkelo Nah. But minor improvements to what we currently have. I think we should add generic tracing to all database calls instead. That would achieve almost the same thing.

PR #29436, opened 2020-11-27
Support svg embedded favicons in whitelabeling

undef1nd:
Fixes templating when favicon is specified as an embedded resource.

### DIFF
--- a/pkg/api/dtos/index.go
+++ b/pkg/api/dtos/index.go
@@ -2,6 +2,7 @@ package dtos
 
 import (
 	"github.com/grafana/grafana/pkg/setting"
+
 	"html/template"
 )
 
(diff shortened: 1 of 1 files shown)


sakjur (approved):
For other reviewers: template.URL is trusted by the Go templating system. In OSS, we don't allow overriding the favicon, so it's statically set to a relative and trusted path — so this change is in practice a noop. For the Enterprise whitelabeling feature this is overridable from grafana.ini, but doesn't support all URL schemas we'd like unless we use `template.URL`. We do our own validation of the validity of the configuration.

torkelo (approved):

papagian (approved):

xlson (approved):

PR #29440, opened 2020-11-27
Datasources: fixed long error message overflowing container

### LINKED ISSUE #29196: Long error messages are incorrectly displayed
Not so much about the message but how it looks.  new datasource -> elastic
Screenshot 2020-11-18 at 11 24 35ce

Screenshot 2020-11-18 at 11 24 42

thisisobate:
Signed-off-by: Uchechukwu Obasi <obasiuche62@gmail.com>

This PR fixes long error messages overflowing its container. 

Fixes #29196 

Long error messages now fit into its parent container
Screenshot 2020-11-27 at 17 00 55

### DIFF
--- a/public/sass/components/_alerts.scss
+++ b/public/sass/components/_alerts.scss
@@ -17,7 +17,6 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  overflow-wrap: break-word;
 }
 
 // Alternate styles
@@ -78,8 +77,9 @@
 }
 
 .alert-body {
+  overflow-wrap: break-word;
+  word-break: break-word;
   flex-grow: 1;
-  min-width: 0;
 }
 
 .alert-icon-on-top {
--- a/public/sass/components/_alerts.scss
+++ b/public/sass/components/_alerts.scss
@@ -79,7 +79,7 @@
 
 .alert-body {
   flex-grow: 1;
-  width: 50%;
+  min-width: 0;
 }
 
 .alert-icon-on-top {
--- a/public/sass/components/_alerts.scss
+++ b/public/sass/components/_alerts.scss
@@ -17,6 +17,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
+  overflow-wrap: break-word;
 }
 
 // Alternate styles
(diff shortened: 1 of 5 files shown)


public/sass/components/_alerts.scss
```diff
@@ -78,6 +79,7 @@
 
 .alert-body {
   flex-grow: 1;
+  width: 50%;
```
torkelo, line 82: why is the alert body 50%?

thisisobate: Creating a child container- in this case, `Alert-body` -without explicitly setting its width causes the content to overflow.
Because the data is dynamically generated, it's best to set constraints so as to avoid issues like this.

thisisobate: I replaced `width` with `min-width` which seems to work pretty fine.
wdyt? @torkelo

torkelo (commented):
I updated the page to use our new Alert component and fixed the issue in that one as well 

seems like with 

```css
overflow-wrap: break-word; // (for chrome and safari)
word-break: break-word; // (for firefox)
```

That seems to work for me without any min-width

thisisobate:
> I updated the page to use our new Alert component and fixed the issue in that one as well
> 
> seems like with
> 
> ```css
> overflow-wrap: break-word; // (for chrome and safari)
> word-break: break-word; // (for firefox)
> ```
> 
> That seems to work for me without any min-width

Wow...that's quite different for me.
I just removed `min-width: 0` and the error is still persistent
Screenshot 2020-12-01 at 13 56 53

torkelo:
Did you add

```
overflow-wrap: break-word; 
word-break: break-word;
```
to the body?

thisisobate:
yes, I did.
Screenshot 2020-12-01 at 14 02 25

thisisobate:
@torkelo If you've got some free time, maybe we could jump on a quick pair programming session.

torkelo:
@thisisobate be sure to pull latest from this branch, have pushed two changes. I cannot replicate the issue anymore in chrome or firefox

thisisobate:
Alright.
Thanks!

torkelo (approved):

PR #29441, opened 2020-11-27
Prometheus: Fix problem where changing display name in Field tab has no effect

### LINKED ISSUE #29437: Prometheus: Field display name has no effect
image

I haven't found a way to change this ...

Can you tell me how to do this?

grafana 7.3.3

zoltanbedi:

Fixes #29437 

### DIFF
--- a/public/app/plugins/datasource/prometheus/result_transformer.test.ts
+++ b/public/app/plugins/datasource/prometheus/result_transformer.test.ts
@@ -1,6 +1,22 @@
 import { DataFrame } from '@grafana/data';
 import { transform } from './result_transformer';
 
+const matrixResponse = {
+  status: 'success',
+  data: {
+    resultType: 'matrix',
+    result: [
+      {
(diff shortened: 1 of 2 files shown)


torkelo:
I can still see two places where displayName is set 
Small/large aren't necessarily for handling resolution or file size, but more for different visuals. SignalFx, for example have both "SignalFx" and "SFx" logos; each with different applications.

zoltanbedi:
That is just the name of the object that is passed to getValueField function. I've changed the property in getValueField. Should I rename these properties?

torkelo (commented):
I think that is fine, maybe rename it to just name for that function to avoid confusion

torkelo (approved):

ivanahuckova (approved):
LGTM! 🍧 I also like that you have renamed `displayNameFromDS` in `getValueField` so it is more clear what is happening. 👍 for test.

Issue #26769, opened 2020-08-03
Document the X-Grafana-Org-Id HTTP header


adeverteuil:
Document the X-Grafana-Org-Id HTTP header in the API docs.

Added by PR https://github.com/grafana/grafana/pull/8122
Referenced here: https://community.grafana.com/t/organization-state-in-the-api/7954

torkelo:
This is an internal header the frontend uses

torkelo:
But could be added to api auth docs https://grafana.com/docs/grafana/latest/http_api/auth/

marefr:
> Also needs to be added in the data source settings, because you can update it in the UI as well.

@oddlittlebird what do you mean? The only way in the UI is to use the org switcher or what am I missing here?

oddlittlebird:
Loki and Prometheus at least allow you to add custom HTTP headers

marefr:
@oddlittlebird that's a completely different feature than this. Should probably be its own issue, something like "Document custom data source HTTP headers"

PR #8122, opened 2017-04-13
use X-Grafana-Org-Id header to ensure backend uses correct org

DanCech:
This patch adds support for passing an X-Grafana-Org-Id header to override the org for a particular request, and updated the frontend to send that header with every request.

This allows the use of multiple tabs with different active orgs, and ensures that requests are always handled using the correct org.

PR #29442, opened 2020-11-27
Admin API: Add admin API for managing auth keys

### LINKED ISSUE #29435: Ability to create an API key for a different org without switching user context
Ability to create an API key for a different org without switching user context.

**Why is this needed**:
I have a service that manages orgs in Grafana using a single Grafana admin service account. Currently, to perform some action in an org, I have to create a token using this admin service account and then use this token to call the APIs. Because of the fact that org context is server side in Grafana, there is a race condition inbetween changing the user context and creating a token, if I'm running multiple instances of my service.

For example, if `instance 1` tries to create a token for org "A" and `instance 2` is creating the token for org "B" at the same time, we can end up creating tokens for org "A" in both instances of the service because `instance 1` switched to org context A before `instance 2` sent the "Create API key" request.

One solution is to allow specifying `orgId` in `POST /api/auth/keys` request. The request would be fulfilled if the current user is an admin of the org, otherwise it must be forbidden.

I could also use some form of distributed locking in my service to control concurrent access to the admin account but it would be way more convenient to have this use case supported on Grafana side.

semyon2105:

Fixes #29435

This API allows Grafana admins to create auth keys for any org without switching the org context.

So far I've only implemented `POST /api/admin/auth/keys` endpoint. I can implement the rest of this API (GET, DELETE) and add tests/docs if this feature gets accepted.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -232,7 +232,7 @@ func (hs *HTTPServer) registerRoutes() {
 		// auth api keys
 		apiRoute.Group("/auth/keys", func(keysRoute routing.RouteRegister) {
 			keysRoute.Get("/", Wrap(GetAPIKeys))
-			keysRoute.Post("/", quota("api_key"), bind(models.AddApiKeyCommand{}), Wrap(hs.AddAPIKey))
+			keysRoute.Post("/", quota("api_key"), bind(dtos.AddApiKeyCommand{}), Wrap(hs.AddAPIKey))
 			keysRoute.Delete("/:id", Wrap(DeleteAPIKey))
 		}, reqOrgAdmin)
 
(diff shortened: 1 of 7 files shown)


wbrowne:
Thanks for the contribution @semyon2105! It should be possible to achieve this using the feature that was added in https://github.com/grafana/grafana/pull/8122, therefore I will close this issue. There's a follow-up issue (https://github.com/grafana/grafana/issues/26769) already create documentation for this, which we'll hopefully get to soon. Thanks again!

daniellee:
Also see comment: https://github.com/grafana/grafana/issues/29435#issuecomment-744408060

semyon2105:
Thanks, I think this solution works out for me!

PR #29446, opened 2020-11-28
GraphNG: use uPlot timestamp config format & pre-compile templates

leeoniya:
this is a proposal to use uPlot's datetime config format & pre-compiler. it also simplifies not having to have timezone knowledge at the axis level since this is a global setting in uPlot that handles `time: true` scales (and linked axes).


torkelo:
You can change date formats in Grafana configuration (ini file), so we need to continue to support that

leeoniya:
for the record, this is what uPlot can compile out of the box in 150 LOC:

https://github.com/leeoniya/uPlot/blob/master/src/fmtDate.js

with easy customization available for words:

https://leeoniya.github.io/uPlot/demos/months-ru.html

Issue #29417, opened 2020-11-26
InternalLinking: Better architecture around linking datasources


aocenas:
At this moment there are multiple features that are basically a data link to a query in a specific datasource (usually opened in Explore) like Derived Fields in Loki, Data Links in Elastic, Exemplars in Prometheus or Trace to Logs functionality.

Right now they are implemented by leveraging parts of DataLink which was aimed only at creating url links and so the code is messy because compared the normal Datalink, InternalLink needs:
1. Know the target datasource and how to create query.
2. Know how to actually open the link as if we see the link in Explore we open it as a second Explore split pane while in Dashboards we navigate to Explore. So we need both proper URL (also to open in separate tab) but also an optional function in Explore to run onClick.

Even though both things are implemented there are lots of issue with the current implementation:
1. We hard code lots of the features to target only specific data source types. For example Derived fields in Loki can only target specific tracing datasources which accept single trace ID as a query. The main issue here is that for user to be able to create proper query they basically need a proper queryEditor otherwise they would need to create a proper query as text/json which is not supper great UX. This would be similar to providing editor for template variables or annotation.
2. We need to pass a lot more information to data link related code and create 2 separate code paths for internal links and normal data links. Things like `getDataSourceSettingsByUid` is passed deep into `fieldOverrides` code because it's needed to create proper links. Similar with the onClick. This probably could be avoided with better separation of concerns and properly define links as either external/internal.

gelicia:
Duplicate of #55236

PR #29447, opened 2020-11-28
DataLinks: Removes getDataSourceSettingsByUid from applyFieldOverrides 

torkelo:
https://github.com/grafana/grafana/pull/25613 added getDataSourceSettingsByUid as a temporary measure that we were to remove in follow-up PR. 

Was about to start doing some work on DataSourceSrv to make the get function work on name & uid transparently and discovered that all getDataSourceSettingsByUid was used for was to just lookup the name. So why not get the name when we build the link model instead? And that looked easy, gave it a shot. Tested with loki + jaeger.  

Todo:
* [x] Update unit tests (could need help with this) 
* [ ] Update X-ray datasource

So I added datasourceName to internal link type, but think we can get rid of that soon as we well. Just need to make data source lookup possible with name or uid, so then getInternalHref can use uid for data source instead of name. And we can just set title instead (in data source code) when the internal link model is built. 

Related to #29417

After #29449 is merged we should be able to remove datasourceName and just set title on the link as normal, as with #29449 we should be able to target the data source with the uid instead of the name


aocenas:
Nice yeah did not realize this before but it makes sense. Maybe we could/should switch Explore links to use UIDs instead of names?

public/app/plugins/datasource/loki/result_transformer.ts
```diff
@@ -387,9 +387,13 @@ export const enhanceDataFrame = (dataFrame: DataFrame, config: LokiOptions | nul
  * Transform derivedField config into dataframe field with config that contains link.
  */
 function fieldFromDerivedFieldConfig(derivedFieldConfigs: DerivedFieldConfig[]): Field<any, ArrayVector> {
+  const dataSourceSrv = getDataSourceSrv();
+
   const dataLinks = derivedFieldConfigs.reduce((acc, derivedFieldConfig) => {
     // Having field.datasourceUid means it is an internal link.
     if (derivedFieldConfig.datasourceUid) {
+      const dsSettings = dataSourceSrv.getDataSourceSettingsByUid(derivedFieldConfig.datasourceUid);
```
dprokop, line 395: Guess this won't be necessary at some point?

dprokop (approved):
LGTM!

public/app/plugins/datasource/loki/result_transformer.ts
torkelo: I think we will likely always have to lookup the data source to set a link title at least. Or we let the user set the link title when they define the derived field

dprokop: OK I see thanks!

torkelo:
> Nice yeah did not realize this before but it makes sense. Maybe we could/should switch Explore links to use UIDs instead of names?

Yes, will try to support that in https://github.com/grafana/grafana/pull/29449 or in follow-up PR to that one. But we still need to set the link name somewhere, either in the data source result transformer or when the derived field is configured

torkelo:
Merging this, created an issue to followup with x-ray fix: https://github.com/grafana/grafana/issues/29506 

But want to wait with that, because depending on if we can get explore to only use data source uid, we can set the title and skip this datasourceName fully

Issue #1664, opened 2015-03-30
Grafana 2.0: Renaming data source does not update graphs


KuchenKerze:
If you are renaming data source the new name is not propagated to existing graphs.
Error message: "Datasource named <old name> was not found"

mobarre:
+1. I'm hitting this also. I'm guessing each data source should be stored with a unique ID that serves as a reference. using the name isn't very good for the UX.

lpic10:
I'm getting this same issue using latest nightly build

utkarshcmu:
+1 to this. This is an important feature for a Grafana administrator.

Josh-West-Bose:
+1

thomasbiddle:
Also getting this.

piotrkochan:
+1

imghasemi:
Also have same problem!

dschade2016:
+1

jedd:
Is there a manual workaround that I can do to effect such a change?

That is specifically to change the name of a datasource name only, and have it propagate to all dashboard variables, panels, and elsewhere.   I don't need to effect the 'default' datasource definition.

bigeast:
I have many pannel using the same datasource, and after changing the name, all of them failed when showing the graph. Now I have to edit each pannel and set the data source to the new one.

Is there a better way?

azhiltsov:
The problem is still exists in 5.3.2 
Renaming a datasource without parsing and updating all dashboards is impossible.
Can dashboard store datasource ID instead of it name ?

netroby:
+1, please rename graph also
Change data source name, will make all graph not working

netroby:
ping @torkelo , Any idea to resolve this problem?

netroby:
Thanks for reply. Can grafana auto rename relation dashboard for us?

johntdyer:
sucks that this isnt working

azhiltsov:
> Not without scanning whole db and fetching each every dashboard and then updating every dashboard that contains the old name

This can be done once as DB migration. Generally dashboards should reference any resource by ID, not by a name.

jorgelbg:
I think that we've hit this issue as well. What happened at least for us when renaming a Prometheus datasource is that not only the panel lose the connection with the datasource but, the entire query is gone. The panel defaults to the default datasource with no query and when a new datasource is selected the query appears as empty.

torkelo:
Sorry, stale bot was wrongly configured (should only have acted on PRs)

[...]

PR #29449, opened 2020-11-28
DataSourceSrv: Look up data source by uid and name transparently

torkelo:
This is related to work to make query editors and query execution a little more future-proof and less stuck in how things have been for ever in dashboards where data source always was referred to by name in the persisted models. 

This should make it easier to start using data source uid as the primary way to lookup data sources. The biggest problem with the current approach is that if you change a data source name dashboards and panels do not update (and become broken).  This PR does not fix that but lays some ground work (to keep scope smaller) by making DataSourceSrv a bit more robust (with more unit tests) and support for uid & name looks in the same get function. Thinking is we first check if name matches a uid then assume you're fetching a data source by uid. This way we could potentially start changing panel.datasource to be the uid while being backward compatible with name.

Question is when we change from data source name to uid, it will make the json harder to reason about for users who use dashboard json & provisioning. They could continue using name but after a change in data source picker in Grafana it would switch it to uid. Think that will be acceptable, but can be discussed when we do that change. 

This PR also tries to clean up some access to config.datasources (think none should use this and long term this should go away) but some remain to be removed in follow-up PRs

Related #1664


packages/grafana-runtime/src/utils/DataSourceWithBackend.ts
```diff
@@ -57,34 +57,37 @@ export class DataSourceWithBackend<
...
     }
+
     const queries = targets.map(q => {
       let datasourceId = this.id;
+
       if (q.datasource === ExpressionDatasourceID) {
         return {
           ...q,
           datasourceId,
-          orgId,
```
torkelo, line 71: @ryantxu I did not get this, why is orgId part of each query? it was not used in https://github.com/grafana/grafana/blob/master/pkg/api/metrics.go#L61. And if used it needs to be added in the backend, the frontend value cannot be trusted, or is it just for tracking/logging?

ryantxu: Looks like it can be safely removed.

ryantxu (approved):
Looks good -- can/should we also deprecate:

```
  getDataSourceSettingsByUid(uid: string): DataSourceInstanceSettings | undefined;
```

and recommend `getSettingsFor`?  The function seems to mostly be used in tests

packages/grafana-runtime/src/services/dataSourceSrv.ts
```diff
@@ -29,6 +29,11 @@ export interface DataSourceSrv {
    * Get all data sources except for internal ones that usually should not be listed like mixed data source.
    */
   getExternal(): DataSourceInstanceSettings[];
+
+  /**
+   * Get settings by name or uid
+   */
+  getSettingsFor(nameOrUid: string | null | undefined): DataSourceInstanceSettings | undefined;
```
hugohaggmark, line 36: I'm a bit confused about the api for `DataSourceSrv `:

Should we deprecate getDataSourceSettingsByUid or have 2 distinct functions? One for uid and one for name? I understand that name is going away, but I think having both uid and name in the same argument is confusing for consumers.

torkelo: I have another PR that touches some stuff related to this function (for a feature it was created for). Was hoping to remove it but might have to deprecate it for a version first
https://github.com/grafana/grafana/pull/29447

PR #29456, opened 2020-11-29
PanelChrome: New logic-less emotion based component with no dependency on PanelModel or DashboardModel

torkelo:
This early work trying to extract the styles and visualization aspect of the panel chrome to a new component with emotion styles that has no logic or dependency on PanelModel & DashboardModel.  (Plan is to rename PanelChrome and PanelChromeAnguar to something else). 

Plan is to then experiment with new layout and styles for panel header & icons & panel links. 

super early state 
Screenshot from 2020-12-04 14-59-09


mckn:
panel_chrome

packages/grafana-ui/src/components/PanelChrome/PanelChrome.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const PanelChrome: React.FC<PanelChromeProps> = ({
+  title = '',
+  children,
+  width,
+  height,
+  padding = 'md',
+  leftItems = [],
+}) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
```
torkelo, line 26:
```suggestion
  const styles = useStyles(getStyles);
```

torkelo (commented):
Looks good!

mckn (approved):
Adding a review for the "Looks good" from @torkelo

peterholmberg (changes requested):
Very nice @mckn ! It would be nice with a story with controls too so I as a user can change behavior without changing the code.

torkelo:
I think all the behaviors are already exposed in the different examples

peterholmberg (approved):

PR #29458, opened 2020-11-29
Chore: Enable exhaustive linter

aknuds1:
Enable "exhaustive" linter, to verify that `switch` statements have branches for all cases. This is useful since there can be bugs due to not handling cases, i.e. cases might have been forgotten about or they might have been introduced in e.g. 3rd party dependencies.


pkg/models/sigv4.go
```diff
@@ -97,6 +97,8 @@ func (m *SigV4Middleware) signer() (*v4.Signer, error) {
 		c = credentials.NewStaticCredentials(m.Config.AccessKey, m.Config.SecretKey, "")
 	case Credentials:
 		c = credentials.NewSharedCredentials("", m.Config.Profile)
+	default:
+		panic(fmt.Sprintf("Unexpected auth type %q", authType))
```
papagian, line 101: I think it should not panic here in case the [default](https://github.com/grafana/grafana/blob/bdbe127a304ee5f0f4298a2a8ba08b09e97515b8/pkg/models/sigv4.go#L21) authentication is selected.
This case used to be handled [here](https://github.com/grafana/grafana/blob/bdbe127a304ee5f0f4298a2a8ba08b09e97515b8/pkg/models/sigv4.go#L105) (no credentials are set).

papagian (changes requested):
Generally LGTM except from the sigv4 modification.

pkg/models/sigv4.go
aknuds1: @papagian Thanks, I will double check.

aknuds1: @papagian Ah, I didn't notice that check. I guess it goes to show that explicit is better than implicit :)

pkg/models/sigv4.go
```diff
@@ -97,10 +97,8 @@ func (m *SigV4Middleware) signer() (*v4.Signer, error) {
 		c = credentials.NewStaticCredentials(m.Config.AccessKey, m.Config.SecretKey, "")
 	case Credentials:
 		c = credentials.NewSharedCredentials("", m.Config.Profile)
-	}
-
-	// passing nil credentials will force AWS to allow a more complete credential chain vs the explicit default
-	if c == nil {
+	case Default:
+		// passing nil credentials will force AWS to allow a more complete credential chain vs the explicit default
```
papagian, line 101: 👍 
I guess you have to add also a default case for any other authentication type, otherwise the exhaustive linter would complain?

aknuds1: @papagian Actually it doesn't, since it can tell there are no more cases.

aknuds1: The default of the linter is actually to require that all switch branches are covered, you have to configure it to accept `default` as fallback.

papagian: GTK, thanks!

papagian (approved):

PR #29461, opened 2020-11-29
Grunt: Removes grunt dependency and replaces some of its usage

torkelo:
* [x] Replace only test check with eslint plugin 
* [x] Call sass:lint directly from package.json script
* [x] Actually run lint steps in CI as this was missing (this must have been missed in drone or some CI work) 
* [x] Remove sasslint from precommit as it's slow

I hope this removes some old legacy npm packages from yarn.lock


package.json
```diff
@@ -146,18 +149,6 @@
...
-    "grunt-cli": "1.3.2",
-    "grunt-contrib-clean": "2.0.0",
-    "grunt-contrib-copy": "1.0.0",
-    "grunt-exec": "3.0.0",
-    "grunt-newer": "1.3.0",
-    "grunt-notify": "0.4.5",
-    "grunt-postcss": "0.9.0",
-    "grunt-sass-lint": "0.2.4",
-    "grunt-usemin": "3.1.1",
-    "grunt-webpack": "3.1.3",
```
torkelo, line 160: 🎉

packages/grafana-ui/src/components/Tooltip/_Tooltip.scss
```diff
@@ -27,6 +27,9 @@ $popper-margin-from-ref: 5px;
   max-width: 400px;
 }
 
+.empty {
+}
```
torkelo, line 31: testing that lint checks run on CI now

build.go
```diff
@@ -233,202 +206,8 @@ func readVersionFromPackageJson() {
 	}
 }
 
-type linuxPackageOptions struct {
```
torkelo, line 236: @aknuds1 removed a bunch of stuff here that I think are not used anymore (especially the grunt package stuff as that was removed in PR a while back), and this linux package code is no longer used as far as I can tell (checked local Makefile)

aknuds1: Thanks @torkelo :) It had to get cleaned up at some point.

aknuds1 (approved):
LGTM!

build.go
```diff
@@ -125,34 +125,7 @@ func main() {
 			}
 
 		case "build-frontend":
-			grunt(gruntBuildArg("build")...)
-
-		case "test":
-			test("./pkg/...")
-			grunt("test")
```
yuri-tceretian, line 132: This command is referenced in the [documentation](https://github.com/grafana/grafana/blob/main/contribute/developer-guide.md#on-windows) for those who use Windows. I wonder if the removal of the entire command was intentional or it was supposed to be just one line at the bottom. If it was, I will then delete it in the docs.

PR #29473, opened 2020-11-30
WIP/POC: support complex math in calculate fields transformer

### LINKED ISSUE #19654: Math value depends on input data
A way to do math depending on data/query result.
**Why is this needed**:
Need to convert data from a unit to other unit, each data have different factor.

SO2 in ug/m3 -> SO2 in ppb have factor 2.62
NO2 in ug/m3 -> SO2 in ppb have factor 1.88

ryantxu:

This PR allows entering an equation to calculate a field with the "Calculate field" transformer.  It uses https://mathjs.org/ for the heavy lifting

Fixes #19654

**TODO**:

- [ ] tests
- [ ] audit bundling math.js -- async loading?
- [ ] variable mapping UI/UX


ryantxu:
replaced with #37758

PR #29474, opened 2020-11-30
Live: publish all dashboard changes to a single channel

ryantxu:

This changes the broadcast logic to publish both an scoped event for the single dashboard, and a channel that collects all changes

This request came from @malcolmholmes work on https://github.com/grafana/grizzly


dprokop (approved):
LGTM

PR #29476, opened 2020-11-30
Stat/Gauge: expose explicit font sizing

### LINKED ISSUE #23305: Enable font size option for Stat panel

ryantxu:

This exposes a explicit font size settings for 'title' and 'value' in the stats and gauge panels

Rather than trying to be tricky with settings (like #28091), this just exposes them directly:

image

Fixes #23305, helps #22209  

alternative to #28091


public/app/plugins/panel/stat/types.ts
```diff
@@ -108,4 +109,32 @@ export function addStandardDataReduceOptions(
       defaultValue: 'auto',
     });
   }
+
+  if (includeFontSizes) {
+    builder.addNumberInput({
+      path: 'fontSize.title',
+      category: ['Font Size'],
```
torkelo, line 116:
```suggestion
      category: ['Font size'],
```

Alternative (to be less technical) `Text size` ?

public/app/plugins/panel/stat/types.ts
```diff
@@ -108,4 +109,32 @@ export function addStandardDataReduceOptions(
...
+        integer: false,
+        min: 1,
+        max: 200,
+      },
+      defaultValue: undefined,
+    });
+
+    builder.addNumberInput({
+      path: 'fontSize.value',
+      category: ['Font Size'],
```
torkelo, line 129:
```suggestion
      category: ['Font size'],
```

(senstance case)

ryantxu:
Bar gauge still needs some work:
image
image

torkelo (approved):
Think this looks good

torkelo (commented):
Think we have a bug (not related to this PR) but when two panels share the same panel options, the values stay in the UI but are not taking effect, I need to update the value again for it to take effect. 

For example have stat panel, set value size 100, switch to Gauge, the options in the side bar stay at 100 in the input but the viz is not afffected, but if update to 99 or something else it will take effect

Issue #5880, opened 2016-08-22
[Elasticsearch]: Enforce ES index setting, limit requests to specific index


wleese:
Bug report
- What Grafana version are you using?
  3.1
- What datasource are you using?
  Graphite with Elasticsearch for annotations
- What did you do?
  Created an Elasticsearch datasource for annotations, specifying a specific index to retrieve the data from.
- What was the expected result?
  Grafana users could only see data in that index
- What happened instead?
  Grafana creates an endpoint the user can manipulate in order to access all data on the elasticsearch cluster. This is a security issue in environments that restrict access to parts of their ES dataset.

Endpoint in our case:
https://grafana/api/datasources/proxy/8/_msearch

Post data:

```
{"search_type":"query_then_fetch","ignore_unavailable":true,"index":["changelog-2016.08.22"]}
{"fields":["@timestamp","_source"],"query":{"filtered":{"query":{"bool":{"should":[{"query_string":{"query":"@environment:test"}}]}},"filter":{"bool":{"must":[{"range":{"@timestamp":{"from":1471831950322,"to":1471853550322}}}]}}}},"size":10000}
```

From our perspective this makes the annotations ES datasource useless, because it introduces a security issue.

wleese:
Just a start. Although it doesn't restrict users from accessing other indices, it does prevent other malicious behavior such as changing settings, mappings and deleting data.

https://github.com/grafana/grafana/pull/6031

wleese:
New PR with a more extensive approach https://github.com/grafana/grafana/pull/7806

marefr:
#12757 will solve this

matyax:
Can anyone confirm is this one still valid?

conet:
I still see this in 8.5.4 and consider it a security issue, I know it's an older branch but I can't upgrade until https://github.com/grafana/grafana/issues/25469 is fixed.

Steps to reproduce:
1. Copy an elasticsearch datasource frontend request as curl (the on like `/api/datasources/proxy/99/_msearch`)
2. Change the index and/or the query
3. It will run the query

Expected result:
The query on other index patterns that those defined in the datasources should not be allowed.

PR #29477, opened 2020-12-01
Elasticsearch: Allow users to override index pattern per query

### LINKED ISSUE #20674: Selectable Index Patterns for Elasticsearch DataSources
Allow elasticsearch index name selection when creating a query for a panel (similar to namespace selection on a CloudWatch datasource.

**Why is this needed**:
Presently, when setting up an elasticsearch datasource, you must enter the index name as part of the datasource configuration.  This means that even though accessing the same elasticsearch cluster, I need to manually define every single index pattern as its own datasource.

If instead there was only necessity to define the timeseries pattern format (EX `(.*)-YYYY.MM.DD`), it would be very easy to provide the user with a selectable list of index patterns that match per query using the `GET /_aliases` endpoint on an elasticsearch cluster, or perhaps one of the `/_cat/` endpoints depending on how the information is parsed.

simianhacker:

This PR adds the ability to override the index pattern per Elasticsearch query. This would allow users to override the index pattern per query and have the added benefit of supporting multiple index patterns on the same visualization. 

image

Fixes #20674 

I put this PR together as a possible solution for #20674, I'm open to other ideas for solving this problem. I've chosen this as my preferred solution because of the low surface area (code wise) and it provides the maximum benefit because it allows users to mix index patterns in the same visualization.


dagwieers:
This is exactly what we need. Currently we have about 10 different Elastic datasources with the only difference being the index pattern. And we are adding more each few months.

Using the "Outer join" transformation we can enrich our query data with information from another "source".

halfa:
That would be a welcome addition, opening the possibility to mix indices in a single dashboard

dagwieers:
@simianhacker Just a heads-up that this branch now has conflicts

simianhacker:
@Elfo404 This has been refactored to the React PR and is ready for review. I also fixed an issue where the Alias input field wasn't being populated correctly.

dagwieers:
@simianhacker It is obvious, but important to realize that with this change access to indices is no longer controlled from the data source. So it is important to have Grafana's access properly fine-tuned in Elastic itself.

Which you should be doing anyhow, but may not be the case for everyone running both together today. I think we have to mention this specifically in the release notes when this is merged.

Also, we should probably improve the Elasticsearch data source documentation with guidance on what privileges would work for Grafana. Which is just **read** on the indices listed:
```yaml
indices:
- names:
  - your-index-*
  - your-other-index-*
  privileges:
  - read
``

Elfo404:
@dagwieers good point. There's also an existing issue regarding this: https://github.com/grafana/grafana/issues/5880.

> with this change access to indices is no longer controlled from the data source. 

That's not completely true as currently there are no checks on which indices can be accessed by using the proxy endpoint, that's actually why this PR works.

IMHO it should not be Grafana's (or this plugin's) responsibility to enforce this kind of "safety" measures since it will only provide a false sense of security. What do you think?

Anyway, I agree that the documentation will benefit from having a bit more information on that matter.

dagwieers:
Please do not forget about this one :wink:

henrikno:
This is going to reduce the number of datasources we have to create by at least 50x.

simianhacker:
@henrikno That's awesome to hear.

hikerspath:
This was marked for v8, any updates on when this could go live?  Would be SUPER useful to be sure.  I could reduce about 35 datasources to 1

danielmotaleite:
This is really needed!
any hint when this will be merged

hikerspath:
Likewise, very curious...

Elfo404:
Sorry for the late update, this one slipped through a bit :)

we have some refactoring planned for the ES data source, moving most of the logic to the BE which will hopefully solve quite a few pending issues (especially when it comes to alerting).

That said, this PR is actually conflicting with some of the decisions and concerns that arose when discussing the feature request, so (also together with @simianhacker ) we decided to pause the work being done here and pick it up as soon as possible once the migration is completed.

I understand it's not the answer you were hoping for, but I want to confirm this has PR has not been forgotten, we just need to do some prep work before moving forward.

cjabrantes:
Hi,

Can someone confirm this one? i was needed this, but seems its not available yet.

Thanks,

PR #29485, opened 2020-12-01
typos

flopp999:
found some typo and also some suggestions


aknuds1:
I don't see any benefit to the changes here, which are largely just opinionated (f.ex. adding punctuation). Also, you would have to update conf/defaults.ini to keep them in sync.

PR #29490, opened 2020-12-01
Docs: InfluxDB_V2 datasource: adding an example on how to add InfluxQL as a datasource

rumsrami:

This is to document how to add an InfluxQL InfluxDB v2 datasource. This is a tricky part as some users of influxdb migrated to version 2 while still not used to flux and would like to use the old influxQL

### DIFF
--- a/docs/sources/datasources/influxdb.md
+++ b/docs/sources/datasources/influxdb.md
@@ -286,12 +286,12 @@ datasources:
       httpMode: GET
 ```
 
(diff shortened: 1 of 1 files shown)


achatterjee-grafana (approved):
LGTM, thank you for your contribution! Adding backend-platform to review the code before merging.

ryantxu (approved):
LGTM -- thanks!

PR #29491, opened 2020-12-01
Middleware: Simplifications

aknuds1:
Chore PR to simplify backend middleware code a bit, in preparation for a deeper PR to remove dependency on global state.


pkg/registry/registry.go
```diff
@@ -25,14 +25,25 @@ func RegisterServiceWithPriority(instance Service, priority Priority) {
 }
 
 func RegisterService(instance Service) {
-	services = append(services, &Descriptor{
+	Register(&Descriptor{
```
wbrowne, line 28: Just wanted to confirm you meant to exclude the `RegisterServiceWithPriority` func from using `Register`?

wbrowne (approved):
LGTM! Just a small remark

pkg/registry/registry.go
aknuds1: @wbrowne I will check.

aknuds1: @wbrowne Looking at the code, I simply didn't think of it

aknuds1: Thanks, made the change!

PR #29493, opened 2020-12-01
Chore: Upgrade typescript to 4.1

kaydelaney:
Bumps typescript to version 4.1


packages/grafana-data/src/field/fieldDisplay.ts
```diff
@@ -160,7 +160,7 @@ export const getFieldDisplayValues = (options: GetFieldDisplayValuesOptions): Fi
             });
 
             values.push({
-              name,
+              name: '',
```
kaydelaney, line 163: As far as I understand, previously this code incorrectly referenced the (now deprecated) global [name](https://developer.mozilla.org/en-US/docs/Web/API/Window/name) property. Here I've replaced it with an empty string to satisfy the interface, but there might be a more appropriate value.

packages/grafana-ui/src/components/ThemeColors/ThemeColors.tsx
```diff
@@ -31,7 +31,7 @@ const DemoBox: FC<DemoBoxProps> = ({ bg, border, children }) => {
           padding-bottom: 16px;
         `}
       >
-        {name}
+        {window.name}
```
dprokop, line 34: This looks funky to me. I guess this should be a `name` property maybe, but I cannot see anything like that in the props interface.

dprokop (approved):
Looks good @kaydelaney !

packages/grafana-ui/src/components/ThemeColors/ThemeColors.tsx
kaydelaney: Yeah, it's just another instance of what I mentioned in my other comment. `{name}` was actually referencing the global `name` property (which is the same as `window.name`). I assume this wasn't the intended behavior, but I changed it to `window.name` to satisfy the type checker as well as at least not _changing_ the behavior, even if it wasn't actually meant. Maybe someone else can clarify what the original intention was. @torkelo ?

packages/grafana-data/src/field/fieldDisplay.ts
kaydelaney: Yep

packages/grafana-ui/src/components/ThemeColors/ThemeColors.tsx
torkelo: You can remove this whole div, not used, just results in an empty div https://developers.grafana.com/ui/latest/index.html?path=/story/docs-overview-themecolors--theme-colors-demo

torkelo (approved):

packages/grafana-ui/src/components/ThemeColors/ThemeColors.tsx
kaydelaney: Do we want to keep the `16px` padding?

PR #29496, opened 2020-12-01
GraphNG: simplify effects responsible for plot updates/initialization 

dprokop:
Simplifies UPlotChart component to utilize single effect for handling data updates vs initialization. Moves tests closer to the actual component being tested


packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -1,83 +1,79 @@
...
-    }
-
-    pluginLog('UPlotChart: init uPlot', false, 'initialized with', plotData.current, currentConfig);
-    const instance = new uPlot(currentConfig, plotData.current, canvasRef.current);
-
-    setPlotInstance(instance);
-  }, [setPlotInstance, currentConfig]);
-
+  const prevProps = usePrevious(props);
+  const prevBuilder = usePrevious(props.config);
```
torkelo, line 20: this seems unnecessary, we can just use prevProps.config   (usePrevious is a bit expensive, it includes its own useEffect)

dprokop: great catch!

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -1,83 +1,79 @@
...
 
-  useEffect(() => {
-    const currentInstance = plotInstance;
-    return () => {
-      currentInstance?.destroy();
-    };
-  }, [plotInstance]);
+    // 1. When config is ready and there is no uPlot instance, create new uPlot and return
+    if (isConfigReady && !plotInstance.current) {
+      plotData.current = prepareData(props.data);
```
torkelo, line 44: looking at the code it looks like we can remove plotData as useRef, it is only written two, only time I can see we actually use it is  on line 65, when config or timezone change, and we could easily just call prepareData there as that should be infrequent

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -1,83 +1,79 @@
...
       });
+      return;
     }
-  }, [plotInstance, props.width, props.height]);
+
+    // 3. When config or timezone has changed, re-initialize plot
+    if (isConfigReady && (props.config !== prevBuilder || prevProps.timeZone !== props.timeZone)) {
+      if (plotInstance.current) {
+        pluginLog('uPlot core', false, 'destroying instance');
+        plotInstance.current.destroy();
```
torkelo, line 63: I can see this is the only place we destroy, looks like we do not destroy on unmount

torkelo (approved):
Nice improvement!

packages/grafana-ui/src/components/uPlot/Plot.tsx
dprokop: Another great catch, thank you!

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -64,10 +64,11 @@ export const GraphNG: React.FC<GraphNGProps> = ({
   const legendItemsRef = useRef<LegendItem[]>([]);
   const hasLegend = useRef(legend && legend.displayMode !== LegendDisplayMode.Hidden);
   const alignedFrame = alignedFrameWithGapTest.frame;
-  const compareFrames = useCallback(
-    (a: DataFrame, b: DataFrame) => compareDataFrameStructures(a, b, ['min', 'max']),
-    []
-  );
+
+  const compareFrames = useCallback((a: DataFrame, b: DataFrame) => {
+    return compareDataFrameStructures(a, b, ['min', 'max']);
```
torkelo, line 69: I do not think you need to ignore min max anymore as they set dynamically by applyFieldOverrides anymore

PR #29502, opened 2020-12-01
Auth proxy: Return standard error type

aknuds1:
Change auth proxy methods to return standard `error` type, instead of the custom `Error` type, to follow standard Go style. Also rewrite tests to use standard library (the two are somewhat intertwined).

Probably easiest reviewed in a graphical diff tool, since the test changes look worse in GitHub's diff UI than they actually are.

Part of a series of PRs to enable removing global state from middleware.


papagian (approved):
LGTM

PR #29507, opened 2020-12-01
Instrumentation: fix handler labels for HTTP requests metrics

### LINKED ISSUE #29487: Instrumentation: Handler label is incorrect for some HTTP requests metrics
Some HTTP metrics are not registred correctly. 

Ex: Annotation Post requests should be registered something like this.
`{handler="/annotations", method="post"}`

But instead they are registered under `{handler="/", method="post"}`

This is very problematic when troubleshooting incidents since we dont know what kind of requests are sent to the Grafana server. 

To reproduce this
1. Start Grafana
2. Create annotations over the API.
```bash 
#!/bin/bash

COUNTER=0
MAX=1600000
while [  $COUNTER -lt $MAX ]; do
    now=$(date +%s000)
    timestamp=$(($now - 6000))
    timestamp_end=$(($timestamp + 6000))
    msg='{"dashboardId":1,"panelId":2,"time":'$timestamp',"isRegion":false,"tags":["bar"],"text":"foo","timeEnd":'$timestamp_end'}'
    curl -H "Content-Type: application/json" http://admin:admin@localhost:3000/api/annotations -d "$msg"
    let COUNTER=COUNTER+1
done
```
3. Check `http://localhost:3000/metrics` and you will see that `http_request_duration_milliseconds_count{handler="/",method="post",statuscode="200"}` is increasing.

papagian:
Not sure if it's the preferable solution but it fixes the handler label issue described [here](https://github.com/grafana/grafana/issues/29487).
With this fix the metric is:
`http_request_duration_milliseconds_count{handler="/api/annotations",method="post",statuscode="200"} 409`

Fixes #29487

### DIFF
--- a/pkg/middleware/request_metrics.go
+++ b/pkg/middleware/request_metrics.go
@@ -63,7 +63,7 @@ func RequestMetrics(cfg *setting.Cfg) func(handler string) macaron.Handler {
 				// avoiding the sanitize functions for in the new instrumentation
 				// since they dont make much sense. We should remove them later.
 				histogram := httpRequestDurationHistogram.
-					WithLabelValues(handler, strconv.Itoa(rw.Status()), req.Method)
+					WithLabelValues(req.URL.String(), strconv.Itoa(rw.Status()), req.Method)
 				if traceID, ok := cw.ExtractSampledTraceID(c.Req.Context()); ok {
 					// Need to type-convert the Observer to an
 					// ExemplarObserver. This will always work for a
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
Looks good AFAICT, but maybe @bergquist should chime in?

bergquist (changes requested):
This will generate too many new time series. Ex if an ID is part of the path or query strings with different values.

I think this has to be solved in https://github.com/grafana/grafana/blob/master/pkg/api/routing/route_register.go . To somehow flatten the combinations of groups into one long handler route.

PR #29512, opened 2020-12-01
GraphNG: add spanNulls config option

ryantxu:
FieldConfig has a nullValuesMode that we do not (yet) expose in the UI -- this addes it as a standard option in the UI.

| Flot  | React |
|---|----|
| nulls |  image |


packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -190,6 +193,28 @@ export const getStandardFieldConfigs = () => {
     category,
   };
 
+  const nullValue: FieldConfigPropertyItem<any, NullValueMode, SelectFieldConfigSettings<NullValueMode>> = {
+    id: 'nullValueMode',
+    path: 'nullValueMode',
+    name: 'Null values',
+    description: 'Behavior when data includes null values',
```
leeoniya, line 200: 'Behavior for null values in the data'?

packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -122,4 +122,5 @@ export enum FieldConfigProperty {
   Mappings = 'mappings',
   Links = 'links',
   Color = 'color',
+  NullValuesMode = 'nullValueMode',
 }
```
leeoniya, line 126: one is plural and the other isn't?

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -190,6 +193,28 @@ export const getStandardFieldConfigs = () => {
...
+
+    editor: standardEditorsRegistry.get('radio').editor as any,
+    override: standardEditorsRegistry.get('radio').editor as any,
+    process: selectOverrideProcessor,
+
+    settings: {
+      options: [
+        { label: 'Gaps', value: NullValueMode.Null },
+        { label: 'Connected', value: NullValueMode.Ignore, description: 'Ignore the values from any calculation' },
+        { label: 'Zero', value: NullValueMode.AsZero },
```
leeoniya, line 210: "Zeros" (plural to match "Gaps")

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -190,6 +193,28 @@ export const getStandardFieldConfigs = () => {
...
+    description: 'Behavior when data includes null values',
+
+    editor: standardEditorsRegistry.get('radio').editor as any,
+    override: standardEditorsRegistry.get('radio').editor as any,
+    process: selectOverrideProcessor,
+
+    settings: {
+      options: [
+        { label: 'Gaps', value: NullValueMode.Null },
+        { label: 'Connected', value: NullValueMode.Ignore, description: 'Ignore the values from any calculation' },
```
leeoniya, line 209: "Connected" always seemed odd to me, the nulls themselves are not connected, they're "Spanned" or Ignored

torkelo: Yea agree, ignored / skipped seems better

packages/grafana-data/src/types/fieldOverrides.ts
ryantxu: now matching the actual property value

leeoniya:
IMO, having `Gaps` as part of a 3-state toggle at the shared config level feels too graph/sparkline specific. on the other hand it's one place to choose the behavior for `null` values.

i would find it more intuitive perhaps if the Ignored/asZeros was general and an extra boolean toggle was added to the graph level for spanning nulls (false by default). if the shared config is set to "as zeros" then the span setting would have no visible effect (because all the nulls would have been "converted" to zeros).

dprokop:
>i would find it more intuitive perhaps if the Ignored/asZeros was general and an extra boolean toggle was added to the graph level for spanning nulls (false by default). if the shared config is set to "as zeros" then the span setting would have no visible effect (because all the nulls would have been "converted" to zeros).

+1 for this. I think the null value mode describes the data behaviour rather than the aesthetics of the graph. I would separate these two things and treat them as standard field option(null value mode) and custom, graph specific toggle(gaps vs spans rendering)

torkelo:
But then the question is should the null value mode (ignore / as zero) be a transform or standard field option? The ignore can be solved with the new filter transform

ryantxu:
>  (ignore / as zero) be a transform or standard field option?

It should be a transform, and we should figure out how to remove/deprecate it from reducers.  BUT i think that is out of scope for this PR ;)

packages/grafana-ui/src/utils/standardEditors.tsx
```diff
@@ -190,6 +193,30 @@ export const getStandardFieldConfigs = () => {
...
+    editor: standardEditorsRegistry.get('radio').editor as any,
+    override: standardEditorsRegistry.get('radio').editor as any,
+    process: selectOverrideProcessor,
+
+    settings: {
+      options: [
+        { label: 'Null', value: NullValueMode.Null },
+        // The "connectd" of "span gaps" behavior is moved to graph settings.
+        // This option should either be a boolean or removed an just "AzZero" added as a transformer
+        // { label: 'Ignore', value: NullValueMode.Ignore, description: 'Ignore the values from any calculation' },
```
ryantxu, line 211: I'm hiding the "ignore" option since that has a less clear migration path.   For "AzZero" we can use a transformation

torkelo:
But if we hide as zero then I almost lobby skipping the null value standard field mode and only have a graph gaps mode? 

The ignore null option could in some cases be useful as an option for the reducers when calc mean etc, hmm. But is it really a standard option?? Tricky

ryantxu:
I just updated the title of this PR and removed the standard options config.  In a new PR I will look at how to deprecated and remove the standard `nullValueMode` -- I think that should become a transformer that would replace nulls with either 0 or the previous value

torkelo (approved):

PR #29515, opened 2020-12-01
docker: fix Dockerfile after Gruntfile.js removed

bsnyder788:

Fixes the Dockerfiles after the Gruntfile.js file was removed here: https://github.com/grafana/grafana/pull/29461


aknuds1 (approved):
LGTM

aknuds1:
Thanks for the fix!

PR #29520, opened 2020-12-02
Arrow: toArray() on nullable values should include null values

ryantxu:
When calling `toArray()` on an arrow vector it returns the underlying native ArrayBuffer -- this is great... unless there are null values.

The new uPlot panel uses toArray... so this becomes an issue.

Interestingly we had solved this before explicitly for the DTO object.


packages/grafana-data/src/dataframe/__snapshots__/ArrowDataFrame.test.ts.snap
```diff
@@ -265,7 +265,7 @@ Object {
       "labels": undefined,
       "name": "float32_values",
       "type": "number",
-      "values": Array [
+      "values": Float32Array [
```
ryantxu, line 268: NOTE:  the real test is that the rest of the null values in this file that are still supported :)

torkelo:
Sounds very strange, the native Array supports null, what native array buffer is this that does not support nulls?

leeoniya:
> Sounds very strange, the native Array supports null, what native array buffer is this that does not support nulls?

he means typed arrays don't support nulls. and decoded Arrow arrays are typed.

```js
let x = new Float64Array(10);
x[0] = null;
console.log(x);
```

ryantxu:
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer

Arrow holds an additional bitmask to support null values, but it seems you only get that when you access via `.get(index)`

leeoniya (approved):

PR #29529, opened 2020-12-02
Instrumentation: Fix bug with invalid handler label value for HTTP request metrics

### LINKED ISSUE #29487: Instrumentation: Handler label is incorrect for some HTTP requests metrics

bergquist:
closes https://github.com/grafana/grafana/issues/29487

I'll submit another PR later to refactor the tests a bit.But I want to keep this PR as small as possible. 

### DIFF
--- a/pkg/api/routing/route_register.go
+++ b/pkg/api/routing/route_register.go
@@ -118,22 +118,24 @@ func (rr *routeRegister) Register(router Router) {
 
 func (rr *routeRegister) route(pattern, method string, handlers ...macaron.Handler) {
 	h := make([]macaron.Handler, 0)
+	fullPattern := rr.prefix + pattern
+
 	for _, fn := range rr.namedMiddleware {
-		h = append(h, fn(pattern))
+		h = append(h, fn(fullPattern))
 	}
 
 	h = append(h, rr.subfixHandlers...)
 	h = append(h, handlers...)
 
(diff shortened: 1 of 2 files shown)


papagian (approved):
LGTM and works as expected!

PR #29530, opened 2020-12-02
Icon: changed the icon for signing in

### LINKED ISSUE #29494: Change the icon for signing in
I find it very confusing having the sign in button icon pointing to the left:

image

See https://play.grafana.org/
I think we should use an icon pointing to the right and we should align the icons used both in the sing in menu as well as in the menu that pops out when user is logged in (see screenshot in the comment below)

**Why is this needed**:
I think the left direction is misleading and suggests signing out rather than signing in

thisisobate:
Signed-off-by: Uchechukwu Obasi <obasiuche62@gmail.com>

This PR gives the user the right suggestion for signing in by using an icon that points to the right.

Fixes #29494 

This is how it looks after I made changes:
Screenshot 2020-12-02 at 12 57 20

### DIFF
--- a/packages/grafana-ui/src/types/icon.ts
+++ b/packages/grafana-ui/src/types/icon.ts
@@ -34,6 +34,7 @@ export type IconName =
   | 'exclamation-triangle'
   | 'times'
   | 'signin'
+  | 'signout'
   | 'cloud-upload'
   | 'step-backward'
   | 'square-shape'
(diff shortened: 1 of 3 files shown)


hugohaggmark (commented):
Looks good, just a quick question, should we also change the SignOut icon?

thisisobate:
Well, I think we shouldn’t change it for now. As @torkelo mentioned on slack, the `sign out` button pointing outwards the shape has been the conventional way of representing a `sign out` button. 

Maybe @dprokop and @torkelo might want to share their thoughts on this.

torkelo (approved):

hugohaggmark:
> Well, I think we shouldn’t change it for now. As @torkelo mentioned on slack, the `sign out` button pointing outwards the shape has been the conventional way of representing a `sign out` button.
> 
> Maybe @dprokop and @torkelo might want to share their thoughts on this.

Aha that makes sense and I wasn't aware of those details. Great work 🎉 !

PR #29531, opened 2020-12-02
Graph NG: Invalidate uPlot config on timezone changes

dprokop:
Since the config builder was only invalidated on data frame structure changes, the time zone change did not have any effect on the axis config being returned from the builder. This PR brings back the timezone being respected on the time axis when the user changes it.


torkelo (approved):

mckn (approved):
LGTM!

PR #22255, opened 2020-02-17
SQL: Define primary key for tables without it

azhiltsov:
The `annotation_tag` and `alert_rule_tag` tables did not have
PRIMARY KEY defined what cause problems with migration to
MariaDB with Galera setup with
 `innodb_force_primary_key=1`
Or MySQL > 8.0.13 with
 `sql_require_primary_key=ON`

Which can manifest as follows:

```
MariaDB
Error 1173: This table type requires a primary key
```

```
MySQL
ERROR 3750 (HY000): Unable to create or change a table
without a primary key, when the system variable 'sql_require_primary_key' is set.
```

Extra reading for curious:
https://jfg-mysql.blogspot.com/2017/08/danger-no-pk-with-RBR-and-mariadb-protection.html

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #12971

**Special notes for your reviewer**:
I screwed https://github.com/grafana/grafana/pull/21387 and github have closed it. 
Resubmitting as new.

PR #29534, opened 2020-12-02
Annotations: Remove annotation_tag entries as part of annotations cleanup

### LINKED ISSUE #29484: Annotations clean up job should delete annotations tags
Currently, the annotation clean up job in Grafana only deletes rows from the `annotation` table. It should also delete all rows in  `annotation_tag` that refers to the delete annotations. Otherwise, the annotation_tag table will grow endlessly. 

https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/annotation_cleanup.go

dafydd-t:
We have an annotation cleanup job which deletes annotations based on age or max count. When we write annotations, we also might create a tag and we definitely create a record in the `annotation_tag` table. Therefore when we delete annotations, we also need to delete the records associated with that annotation from the `annotation_tag` table.

We need this because we have run into issues with a customer who uses annotations very heavily. The size of `annotation_tag` grows so large that the queries to insert and read annotations are very slow. 

#29484 

Fixes #29484 

- I've not used xorm before. If there is a better way to perform the count query, I'd be happy to use it. I'm not use how the struct->table mapping for annotation_tag would work, there are no examples of it in the codebase.
- I've opted for cleaning up orphaned records from `annotation_tag` as a separate query, executed after the original cleanup queries. I felt this was cleaner that something where we extract the ID of the annotation we're deleting and delete records from `annotation_tag` which have that id in the `annotation_id` column
- The tests validate purely with counts, rather than checking specific records have been deleted. This is consistent with what is already there.
- I've only tested this against SQLite3 manually. Are other databases tested as part of the tests or do I need to verify those manually?

### DIFF
--- a/pkg/services/sqlstore/annotation_cleanup.go
+++ b/pkg/services/sqlstore/annotation_cleanup.go
@@ -43,6 +43,8 @@ func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *
 		return totalCleanedAnnotations, 0, err
 	}
 
+	acs.log.Debug("Finished cleaning annotations. Now cleaning orphaned records from annotation_tag")
+
 	affected, err = acs.cleanOrphanedAnnotationTags(ctx)
 	return totalCleanedAnnotations, affected, err
 }
@@ -98,7 +100,7 @@ func (acs *AnnotationCleanupService) executeUntilDoneOrCancelled(ctx context.Con
 				return err
 			})
 			if err != nil {
-				return 0, err
+				return totalAffected, err
 			}
 
 			if affected == 0 {
--- a/pkg/services/sqlstore/annotation_cleanup.go
+++ b/pkg/services/sqlstore/annotation_cleanup.go
@@ -21,8 +21,11 @@ const (
 	apiAnnotationType       = "alert_id = 0 AND dashboard_id = 0"
 )
 
-// CleanAnnotations deletes old annotations created by
-// alert rules, API requests and human made in the UI.
+// CleanAnnotations deletes old annotations created by alert rules, API
+// requests and human made in the UI. It subsequently deletes orphaned rows
+// from the annotation_tag table.
+// Returns the number of annotation and annotation_tag rows deleted. If an
+// error occurs, it returns the number of rows affected so far.
 func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *setting.Cfg) (int64, int64, error) {
 	var totalCleanedAnnotations int64
 	affected, err := acs.cleanAnnotations(ctx, cfg.AlertingAnnotationCleanupSetting, alertAnnotationType)
--- a/pkg/services/sqlstore/annotation_cleanup.go
+++ b/pkg/services/sqlstore/annotation_cleanup.go
@@ -9,7 +9,7 @@ import (
 	"github.com/grafana/grafana/pkg/setting"
 )
 
-// AnnotationCleanupService is responseible for cleaning old annotations.
+// AnnotationCleanupService is responsible for cleaning old annotations.
 type AnnotationCleanupService struct {
 	batchSize int64
 	log       log.Logger
@@ -23,7 +23,9 @@ const (
 
 // CleanAnnotations deletes old annotations created by alert rules, API
 // requests and human made in the UI. It subsequently deletes orphaned rows
-// from the annotation_tag table.
+// from the annotation_tag table. Cleanup actions are performed in batches
+// so that no query takes too long to complete.
+//
 // Returns the number of annotation and annotation_tag rows deleted. If an
 // error occurs, it returns the number of rows affected so far.
 func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *setting.Cfg) (int64, int64, error) {
@@ -45,8 +47,6 @@ func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *
 	if err != nil {
 		return totalCleanedAnnotations, 0, err
 	}
-
-	acs.log.Debug("Finished cleaning annotations. Now cleaning orphaned records from annotation_tag")
 
 	affected, err = acs.cleanOrphanedAnnotationTags(ctx)
 	return totalCleanedAnnotations, affected, err
--- a/pkg/services/sqlstore/annotation_cleanup.go
+++ b/pkg/services/sqlstore/annotation_cleanup.go
@@ -34,7 +34,12 @@ func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *
 		return err
 	}
 
-	return acs.cleanAnnotations(ctx, cfg.DashboardAnnotationCleanupSettings, dashboardAnnotationType)
+	err = acs.cleanAnnotations(ctx, cfg.DashboardAnnotationCleanupSettings, dashboardAnnotationType)
+	if err != nil {
+		return err
+	}
+
+	return acs.cleanOrphanedAnnotationTags(ctx)
 }
 
 func (acs *AnnotationCleanupService) cleanAnnotations(ctx context.Context, cfg setting.AnnotationCleanupSettings, annotationType string) error {
@@ -56,6 +61,11 @@ func (acs *AnnotationCleanupService) cleanAnnotations(ctx context.Context, cfg s
 	}
 
 	return nil
+}
+
+func (acs *AnnotationCleanupService) cleanOrphanedAnnotationTags(ctx context.Context) error {
+	sql := "DELETE FROM annotation_tag WHERE NOT EXISTS (SELECT null FROM annotation WHERE annotation_tag.annotation_id = annotation.id)"
+	return acs.executeUntilDoneOrCancelled(ctx, sql)
 }
 
 func (acs *AnnotationCleanupService) executeUntilDoneOrCancelled(ctx context.Context, sql string) error {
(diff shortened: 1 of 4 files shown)


dafydd-t:
@bergquist did you want this behind a feature flag?

papagian:
> I've only tested this against SQLite3 manually. Are other databases tested as part of the tests or do I need to verify those manually?

Integrations tests are run against [PostgreSQL](https://drone.grafana.net/grafana/grafana/7718/1/23) and [MySQL](https://drone.grafana.net/grafana/grafana/7718/1/24) as a part of the pull request CI.
However, the specific [test](https://github.com/grafana/grafana/blob/3f8ca2c99f9e54ce1cd2d4780ec7f4983809d92f/pkg/services/sqlstore/annotation_cleanup_test.go) is missing the integration build tag so it does not run. I think that we have to added it the [same way](https://github.com/grafana/grafana/blob/3f8ca2c99f9e54ce1cd2d4780ec7f4983809d92f/pkg/services/sqlstore/apikey_test.go#L1) as it happens for all the tests under sqlstore package.

For testing locally one can use the devenv docker blocks (`postgres_tests` and `mysql_tests`) we have for starting a PostgreSQL and MySQL server respectively and creating the user and database that the tests except.
`make devenv sources=postgres_tests,mysql_tests`

Then, one can run the integration tests against PostgreSQL using:
`GRAFANA_TEST_DB=postgres go test -tags=integration -v ./pkg/...`

and for running them agains MySQL:
`GRAFANA_TEST_DB=mysql go test -tags=integration -v ./pkg/...`

dafydd-t:
@papagian Thank you - I ran the int tests locally with the two other DBs. I've added that build constraint to the test file so the CI should capture it as well.

pkg/services/sqlstore/annotation_cleanup.go
```diff
@@ -58,6 +63,11 @@ func (acs *AnnotationCleanupService) cleanAnnotations(ctx context.Context, cfg s
 	return nil
 }
 
+func (acs *AnnotationCleanupService) cleanOrphanedAnnotationTags(ctx context.Context) error {
+	sql := "DELETE FROM annotation_tag WHERE NOT EXISTS (SELECT null FROM annotation WHERE annotation_tag.annotation_id = annotation.id)"
+	return acs.executeUntilDoneOrCancelled(ctx, sql)
+}
```
papagian, line 69: I've noticed that `annotation_tag` is only an intermediate table that relates `annotation` with `tag` entries (instead of using foreign keys). So I suspect that after this we will end up with a lot orphaned entries in the `tag` table. I think that they should be cleaned as well, WDYT?

pkg/services/sqlstore/annotation_cleanup_test.go
```diff
@@ -25,13 +27,15 @@ func TestAnnotationCleanUp(t *testing.T) {
 
 	createTestAnnotations(t, fakeSQL, 21, 6)
 	assertAnnotationCount(t, fakeSQL, "", 21)
+	assertAnnotationTagCount(t, fakeSQL, 42)
```
papagian, line 30: I agree that it hard to test has a test has the expected behaviour by checking the count.
The best alternative I could think would be to check that each time the tags should equal the summary of the remaining annotations multiplied by two (since each created annotation has two tags)

papagian (changes requested):
Generally works fine but I think that the cleanup should remove also the relative entries from the `tag` table.

pkg/services/sqlstore/annotation_cleanup.go
dafydd-t: A valid point. We found that the number of records in the `tag` table does not increase at the same rate as the records in the `annotation_tag` table. Tags are not unique, and the same tag is used in several annotations. We only insert a new tag if it doesn't exist already ([link](https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/annotation.go#L52)). 

If we delete from the `tag` table, we will need to ensure that the record is not in use by any other component. Dashboards and alert rules also use tags. I think this means the query needed to clean up orphaned tags would become considerably more complex, as we would need to check that there are no records in `database_tag` or `annotation_tag` that use a given tag as well. Very happy to be told this is not as hard as I think it is.

With that in mind, we hoped to only need to clean records from the `annotation_tag` table, because this is the part currently causing one particular hosted grafana instance to slow down significantly without manual intervention. We would like to address this issue of cleaning up records from the `tag` table in a separate PR.



@bergquist may want to add some details if I've missed something/.

pkg/services/sqlstore/annotation_cleanup_test.go
dafydd-t: OK, I will include some logic so assert against the sum of the annotation counts * 2, instead of having it explicitly written  as part of the test case.

pkg/services/sqlstore/annotation_cleanup.go
papagian: The fact that the tag entries are not unique is enough for making me change my mind and think that we should not remove them as well.

papagian (approved):

dafydd-t:
I noticed that I had not used the batch size in the query that I used to delete the annotation_tag records, so I have updated the query so it respects this limit. It happens to follow a similar pattern to the other delete queries, which is reassuring.

@bergquist

dafydd-t:
I also noticed when debugging this locally that the annotation cleanup job doesn't produce any debug logs, whereas the other cleanup tasks do. I'm aware we want to get this PR in pretty quickly, but [this diff](https://github.com/grafana/grafana/compare/bugfix-29848-annotations-cleanup...bugfix-294848-annotations-cleanup-with-logging?expand=1) shows the extra changes required to count and log rows affected in the `annotation` and `annotation_tag` cleanup tasks. If the changes look good, we should add them both. If they're contentious, we can just add this PR and then work on the logging after. What do you think?

@bergquist

dafydd-t:
Some info on time: the annotation and annotation_tag cleanup took 18 seconds to clean 3.5K annotations & 32.3K annotation_tags.

I haven't measured this against the master build, which just deletes annotations.

bergquist:
> I also noticed when debugging this locally that the annotation cleanup job doesn't produce any debug logs, whereas the other cleanup tasks do. I'm aware we want to get this PR in pretty quickly, but this diff shows the extra changes required to count and log rows affected in the annotation and annotation_tag cleanup tasks. If the changes look good, we should add them both. If they're contentious, we can just add this PR and then work on the logging after. What do you think?

Let's add those to this PR. If this wasn't targeting a release branch it would have made sense to open a new PR for those but let's avoid backporting multiple changes in one file.

> took 17 seconds to clean 3.5K annotations & 32.3K annotation_tags

Not super fast but good enough. I any idea about where we spend the most time?

dafydd-t:
I've added one more debug statement so we can get a timestamp between annotation and annotation_row cleanup.

Deleting 13.6K annotations took 11 seconds, deleting the associated 122.4K annotation_tags took 17 seconds.

I've merged the debug logging into this branch now, so we should have everything on here.

@papagian Please could you take another look to make sure you're happy, this PR has changed a fair bit since you approved it. Thank you!

papagian (approved):
LGTM I have some minor suggestions.

dafydd-t:
I tried this with ~1M annotation_tags. Here are the times:

```
t=2020-12-07T12:07:06+0000 lvl=dbug msg="Deleted old/expired dashboard versions" logger=cleanup rows affected=0
t=2020-12-07T12:08:29+0000 lvl=dbug msg="Finished cleaning annotations. Now cleaning orphaned records from annotation_tag" logger=annotationcleaner
t=2020-12-07T12:12:21+0000 lvl=dbug msg="Deleted excess annotations" logger=cleanup annotations affected=100680 annotation tags affected=906120
```

so it took
1m23s to delete 100680 annotations
3m51s to delete 906120 annotation_tags

pkg/services/sqlstore/annotation_cleanup.go
```diff
@@ -21,48 +21,74 @@ const (
...
 
-// CleanAnnotations deletes old annotations created by
-// alert rules, API requests and human made in the UI.
-func (acs *AnnotationCleanupService) CleanAnnotations(ctx context.Context, cfg *setting.Cfg) error {
-	err := acs.cleanAnnotations(ctx, cfg.AlertingAnnotationCleanupSetting, alertAnnotationType)
+// CleanAnnotations deletes old annotations created by alert rules, API
+// requests and human made in the UI. It subsequently deletes orphaned rows
+// from the annotation_tag table.
+// Returns the number of annotation and annotation_tag rows deleted. If an
+// error occurs, it returns the number of rows affected so far.
```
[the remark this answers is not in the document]
dafydd-t: I've added in a sentence about the batch sizes.

pkg/services/sqlstore/annotation_cleanup_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
bergquist, line 1: We usually dont add this build flag even if the tests writes/reads from the database. The unit tests use an inmemory sqlite which is fast enough for tests.

bergquist (approved):

pkg/services/sqlstore/annotation_cleanup_test.go
dafydd-t: OK, I can remove it. It was useful for testing against different backends locally.

bergquist:
@dafydd-t anything else you want to do in this or? Otherwise, I think you can merge it :)

dafydd-t:
@bergquist  I'm not authorized to merge 😅

bergquist:
@dafydd-t to backport this PR to 7.3.5 we would also have to backport https://github.com/grafana/grafana/pull/22255
I guess we should evaluate that.

PR #29535, opened 2020-12-02
Middleware: Rewrite tests to use standard library

aknuds1:
Rewrite middleware tests to use standard library instead of goconvey. Part of a series of PRs to remove global state from middleware.

I would recommend diffing with a graphical diff tool, since it will make it much easier to see how things actually change.

Sorry for a large PR! Couldn't be helped unfortunately, since many of these tests are intertwined through common dependencies, and when you change one you tend to have to change all.

### DIFF
--- a/pkg/middleware/middleware_test.go
+++ b/pkg/middleware/middleware_test.go
@@ -346,7 +346,7 @@ func TestMiddlewareContext(t *testing.T) {
 
 		assert.False(t, sc.context.IsSignedIn)
 		assert.Equal(t, int64(0), sc.context.UserId)
-		assert.Empty(t, sc.context.UserToken)
+		assert.Nil(t, sc.context.UserToken)
 	})
 
 	middlewareScenario(t, "When anonymous access is enabled", func(sc *scenarioContext) {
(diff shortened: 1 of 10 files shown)


pkg/middleware/middleware_test.go
```diff
@@ -276,283 +294,334 @@ func TestMiddlewareContext(t *testing.T) {
...
+		sc.fakeReq("GET", "/").exec()
 
-			Convey("Should not init context with user info", func() {
-				So(sc.context.IsSignedIn, ShouldBeFalse)
-				So(sc.context.UserId, ShouldEqual, 0)
-				So(sc.context.UserToken, ShouldBeNil)
-			})
+		assert.False(t, sc.context.IsSignedIn)
+		assert.Equal(t, int64(0), sc.context.UserId)
+		assert.Empty(t, sc.context.UserToken)
```
wbrowne, line 349: `assert.Nil` would be more correct

aknuds1: @wbrowne Checking

wbrowne (approved):
Looking good 🥇

pkg/middleware/middleware_test.go
aknuds1: @wbrowne Correct! Mistake on my end.

PR #29544, opened 2020-12-02
Panel: making sure we support all versions of chrome when detecting position of click event.

### LINKED ISSUE #25018: Panel context dropdown menu does not open from title in chrome
**What happened**:
I created a new panel, clicked the title and the caret next to it, but the menu did not appear

I expected the menu to appear

click on the title of a panel in chrome

bug does not appear in firefox 76.0.1 (64-bit)

**Environment**:
- Grafana version: 7.0 (installed via homebrew)
- Data source type & version: prometheus/latest
- OS Grafana is installed on: macOS 10.15.4
- User OS & Browser: macOS 10.15.4, Chrome Version 83.0.4103.61 (Official Build) (64-bit)
- Grafana plugins: none
- Others:

### LINKED ISSUE #29138: PanelMenu and click event with fractional coordinates
**What happened**:
As per [crbug.com/989958](crbug.com/989958), "click" event will become a PointerEvent in Chrome (instead of a MouseEvent) as per [Pointer Event specification](https://github.com/w3c/pointerevents/pull/317), [HTML specification](https://github.com/whatwg/html/pull/5324) and [UI events specification](https://github.com/w3c/uievents/pull/259). When the ClickPointerEvent flag is enabled (now it is disabled by default, you can enable it with Chrome command line flag --enable-blink-features=ClickPointerEvent) the "click" event's coordinates will become fractional (as per Pointer Event specification) instead of integer.

There was a report of PanelMenu stop opening when clicked (see [Comment 18 on crbug.com/989958](https://bugs.chromium.org/p/chromium/issues/detail?id=989958#c18)). 

The root cause is with [this technique](https://github.com/grafana/grafana/commit/1afc590c703731cd2026fb8286388e1db6afc09f) of comparing "click" and "mousedown" events coordinates.

One way of fixing the issue is to round "click" event's clientX, clientY coordinates [when reading "click" event's clientX and clientY coordinates](https://github.com/grafana/grafana/blob/8c7e7e6d221338756380340a68b3f21187995ece/public/app/features/dashboard/dashgrid/PanelHeader/PanelHeader.tsx#L55).

PanelMenu opens when clicked.

Open https://play.grafana.org/d/000000012/grafana-play-home?orgId=1 when running Chrome with --enable-blink-features=ClickPointerEvent.
Click on Feature Showcase PanelMenu (as per [Comment 18 on crbug.com/989958](https://bugs.chromium.org/p/chromium/issues/detail?id=989958#c18)). 

**Environment**:
- Grafana version: Grafana v7.3.2 (b8ad4eaab3)
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

mckn:
Seems like some version of chrome starts to emit events where coordinates contains float values instead of integer values. This change make sure we are always using integer values.

Fixes #25018
Fixes https://github.com/grafana/grafana/issues/29138

### DIFF
--- a/public/app/features/dashboard/dashgrid/PanelHeader/PanelHeader.tsx
+++ b/public/app/features/dashboard/dashgrid/PanelHeader/PanelHeader.tsx
@@ -1,6 +1,14 @@
 import React, { PureComponent } from 'react';
 import classNames from 'classnames';
-import { DataLink, LoadingState, PanelData, PanelMenuItem, QueryResultMetaNotice, ScopedVars } from '@grafana/data';
+import {
+  DataLink,
+  LoadingState,
+  numberUtil,
+  PanelData,
+  PanelMenuItem,
+  QueryResultMetaNotice,
+  ScopedVars,
+} from '@grafana/data';
 import { AngularComponent, config, getTemplateSrv } from '@grafana/runtime';
 import { ClickOutsideWrapper, Icon, IconName, Tooltip, stylesFactory } from '@grafana/ui';
 import { selectors } from '@grafana/e2e-selectors';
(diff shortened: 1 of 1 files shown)


dprokop:
@mckn seems like this one might be related: https://github.com/grafana/grafana/issues/29138 ?

mckn:
> @mckn seems like this one might be related: #29138 ?

Yes!

jackw (approved):
Nice one! 🚀

PR #29550, opened 2020-12-02
Cloudwatch: use shared library for aws auth

### LINKED ISSUE #31890: CloudWatch: Use grafana-aws-sdk go package
At the moment we have 6 AWS data sources for grafana, and all of them have the same need for handling authentication. The latest changes related to auth in the CloudWatch data (https://github.com/grafana/grafana/pull/31805 & https://github.com/grafana/grafana/pull/31804) is something that all data sources can benefit from, so the changes have been replicated in the [grafana-aws-sdk package](https://github.com/grafana/grafana-aws-sdk/pull/8). Therefore the CloudWatch data source should consume that SDK rather than having its own, equal (from a feature standpoint) implementation.

### LINKED ISSUE #31892: AWS: AllowedAuthProviders should default to the existing auth providers
When reading the `allowed_auth_providers` variables in settings.go we should default to the providers that we've had for some time - `default,keys,credentials`

https://github.com/grafana/grafana/blob/9f92eb685982f92e1ad98c39012970d75d3a8840/pkg/setting/setting.go#L946

ryantxu:
For the Timestream and iot-sitewise plugins, we needed to match the same auth workflow as cloudwatch; I pulled the the logic from core into its own repository: https://github.com/grafana/grafana-aws-sdk/tree/main/pkg/awsds

This PR updates core grafana to use the same SDK.  In the plugins, we load settings using: https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/settings.go#L84, since this uses simplejson, we still have to load the settings manually

TODO:
- [ ] If this looks good, we should make a real release on grafana-aws-sdk 0.1.0?  (since 1.0 triggers behavior I don't think we want!)

Fixes #31890
Fixes #31892


aknuds1:
Having a look.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -89,109 +75,13 @@ type cloudWatchExecutor struct {
...
-
-	sessCacheLock.Lock()
-	sessCache[cacheKey] = envelope{
-		session:    sess,
-		expiration: expiration,
-	}
-	sessCacheLock.Unlock()
-
-	return sess, nil
+	return e.sessions.GetSession(region, *dsInfo)
```
aknuds1, line 84: It doesn't look right to me that `e.sessions.GetSession` should take `region` in addition to the region field on `dsInfo` (which should be renamed to reflect the new type). It should instead just take the region via `dsInfo`, which is basically how the old code works (although it uses the raw `region` parameter to caclculate the cache key, I'm not sure if that's correct).

aknuds1 (commented):
GetSession's use of a separate `region` parameter doesn't look right to me, it also changes behaviour versus what we currently have.

pkg/tsdb/cloudwatch/cloudwatch.go
sunker: I agree! 

Besides from that it looks great! Will test once this is refactored.

sunker (changes requested):

ryantxu:
@aknuds1 any suggestions on how to improve?   The SDK replaces the global session cache with an explicit SessionCache objcet.  Do you think global makes more sense?  (if feels a bit odd to have multiple datasources access the same session cache -- in practice likely not dangerous, but perhaps not great)

aknuds1:
@ryantxu will have to get back to it tomorrow, got up too early today and had to concentrate on more burning matters.

aknuds1:
About caching, let's keep the current strategy, we shouldn't change the behaviour just to switch to using this SDK.

We know the current caching implementation isn't good, but it needs separate consideration.

ryantxu:
Lets pause work on this until someone has the time/energy to look at auth across all all the AWS datasources.  I'm not sure the best API, but there will need to be some minor updates so that the structures make sense for cloudwatch and external plugins.  The value in the end is a single authentication path and shared setup documentation.

sunker:
> About caching, let's keep the current strategy in CloudWatch, we shouldn't change the behaviour just to switch to using this SDK.
> 
> We know the current caching implementation isn't good, but it needs separate consideration.

I'm not very familiar with the auth part of the cloudwatch code, but replacing to global cache sounds like the right thing to do to me. What other problems are there with the existing caching implementation @aknuds1?

aknuds1:
> > About caching, let's keep the current strategy in CloudWatch, we shouldn't change the behaviour just to switch to using this SDK.
> > We know the current caching implementation isn't good, but it needs separate consideration.
> 
> I'm not very familiar with the auth part of the cloudwatch code, but replacing to global cache sounds like the right thing to do to me. What other problems are there with the existing caching implementation @aknuds1?

@sunker I don't remember the details now, but @patstrom was supposed to come up with a PR. It's a complicated topic, and it should be solved separately from refactoring to use the SDK.

sunker:
> > > About caching, let's keep the current strategy in CloudWatch, we shouldn't change the behaviour just to switch to using this SDK.
> > > We know the current caching implementation isn't good, but it needs separate consideration.
> > 
> > 
> > I'm not very familiar with the auth part of the cloudwatch code, but replacing to global cache sounds like the right thing to do to me. What other problems are there with the existing caching implementation @aknuds1?
> 
> @sunker I don't remember the details now, but @patstrom was supposed to come up with a PR. It's a complicated topic, and should be solved separately from refactoring to use the SDK.

Okay, let's catch up later.

patstrom:
Unfortunately I've had a lot to do for the past few months so I haven't really had time to sit down with that PR. But from what I remember my objection to the current implementation was just that the cache-key is built in a very convoluted manner. My PR would essentially just be to replace the current way of building a cache key with the data source id and the data source version.

Last time I sat down with this I thought it kind of confusing whether or not a `CloudwatchExecutor` instance was actually localized to a single data source + data source version or not (it seemed to be in some cases and some cases not) so I am unsure about how reasonable it is to store the cache as a struct field.

Caching is always complicated but from my research into the source code during the refactor I determined (and tested) that simply using a cache key that combined a data source id and a data source version worked well to reduce session creation and fixed bugs where the configuration of one data source could interfere with another one.

What I did in the original PR was to have a global cache of something like `map[dataSourceVersion]CloudwatchExecutor` that made sure there was one instance of CloudwatchExecutor for each data source id (and that it didn't have to be re-initialized at every request). The `getExecutor(dataSourceId)` function (or whatever it was called) invalidated the cache if the version had been updated. Each `CloudwatchExecutor` then had a cache of `region -> aws.Session` that it managed for the entire process lifetime. This did seem to work very well and while I did test it thoroughly I only did manual testing with no actual concrete coverage.

I think if I were to do it today in the current implementation and I only considered the authentication I would probably keep the global cache but change the type to something like `map[dataSourceId+dataSourceVersion]map[region]aws.Session` (in some kind of psuedo-golang) and just go with that. You would, of course, have to actually clean up that cache explicitly since the GC doesn't know that old dataSourceVersions are unlikely to be used.

However, there is also a separate global cache of things like custom metrics and whatnot (I cannot remember exactly) so if I were to do a "proper" PR I would create a construct that can hold information about a specific (data source id, data source version) tuple (`CloudwatchExecutor`?) and implement all required caches as fields/methods in that struct (similar to my original PR). This, of course, leads to more cache constructs but at least they are isolated from each other. The main issue would again be to actually purge the cache (in particular the global `(dataSourceId, dataSourceVersion) -> SomeConstructThatHoldsConfigAndCaches`).

aknuds1:
Thanks for chiming in @patstrom.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -384,28 +274,7 @@ func (e *cloudWatchExecutor) executeLogAlertQuery(ctx context.Context, queryCont
...
-		return "sharedCreds"
-	case authTypeKeys:
-		return "keys"
-	default:
-		panic(fmt.Sprintf("Unrecognized auth type %d", at))
-	}
-}
-
-func (e *cloudWatchExecutor) getDSInfo(region string) *datasourceInfo {
+func (e *cloudWatchExecutor) getDSInfo(region string) *awsds.AWSDatasourceSettings {
```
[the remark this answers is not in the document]
ryantxu: +1

pkg/tsdb/cloudwatch/cloudwatch.go
ryantxu: Perhaps we should add an additional method signature that does not take "region" since you obviously created a dsInfo with the same region in the line above.

In the plugin flavor the settings and session cache is created once for each instance.  See https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/sitewise/datasource.go#L35

ryantxu:
If we are putting effort into adding (and testing!!!!!) a new authentication method, it would be a shame to do it in a way that requires us to backport and test it twice (at least).

I just updated this PR so it builds and I think is worth evaluating -- any concerns we have are currently duplicated in all plugins, so keep that in mind.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -415,31 +276,7 @@ func (e *cloudWatchExecutor) executeLogAlertQuery(ctx context.Context, queryCont
...
-func (at authType) String() string {
-	switch at {
-	case authTypeDefault:
-		return "default"
-	case authTypeSharedCreds:
-		return "credentials"
-	case authTypeKeys:
-		return "keys"
-	case authTypeEC2IAMRole:
-		return "ec2_iam_role"
```
ryantxu, line 436: If this is the new auth method... it will need to be added to the SDK!  along with anything else that may have been added recently

pkg/setting/setting.go
```diff
@@ -940,17 +941,19 @@ func (cfg *Cfg) readLDAPConfig() {
...
+	allowedAuthProviders := awsPluginSec.Key("allowed_auth_providers").MustString("default,keys,credentials")
 	for _, authProvider := range strings.Split(allowedAuthProviders, ",") {
 		authProvider = strings.TrimSpace(authProvider)
 		if authProvider != "" {
 			cfg.AWSAllowedAuthProviders = append(cfg.AWSAllowedAuthProviders, authProvider)
 		}
 	}
 	cfg.AWSListMetricsPageLimit = awsPluginSec.Key("list_metrics_page_limit").MustInt(500)
+	os.Setenv("ASSUME_ROLE_ENABLED", strconv.FormatBool(cfg.AWSAssumeRoleEnabled))
+	os.Setenv("ALLOWED_AUTH_PROVIDERS", allowedAuthProviders)
```
ryantxu, line 956: @sunker these do not get passed to plugins unless explicitly added by the backend plugin manager.  I replaced them with constants defined in the grafana-aws-sdk

go.mod
```diff
@@ -10,6 +10,8 @@ replace github.com/denisenkom/go-mssqldb => github.com/denisenkom/go-mssqldb v0.
 // It's also present on grafana/loki's go.mod so we'll need till it gets updated.
 replace k8s.io/client-go => k8s.io/client-go v0.18.8
 
+replace github.com/grafana/grafana-aws-sdk => ../more/grafana-aws-sdk
```
ryantxu, line 13: this is referencing the WIP/Draft code in:
https://github.com/grafana/grafana-aws-sdk/pull/8

pkg/setting/setting.go
sunker: great!

go.mod
sunker: :+1:

pkg/setting/setting.go
```diff
@@ -940,17 +942,20 @@ func (cfg *Cfg) readLDAPConfig() {
...
 	for _, authProvider := range strings.Split(allowedAuthProviders, ",") {
 		authProvider = strings.TrimSpace(authProvider)
 		if authProvider != "" {
 			cfg.AWSAllowedAuthProviders = append(cfg.AWSAllowedAuthProviders, authProvider)
 		}
 	}
 	cfg.AWSListMetricsPageLimit = awsPluginSec.Key("list_metrics_page_limit").MustInt(500)
+	// Also set environment variables that can be used by core plugins
+	os.Setenv(awsds.ENV_VAR_AssumeRoleEnabled, strconv.FormatBool(cfg.AWSAssumeRoleEnabled))
+	os.Setenv(awsds.ENV_VAR_AllowedAuthProviders, allowedAuthProviders)
```
sunker, line 958: The sdk will retrieve these settings through environment variables, so all we need to do is to set them.

pkg/setting/setting.go
```diff
@@ -940,17 +942,20 @@ func (cfg *Cfg) readLDAPConfig() {
 	cfg.LDAPAllowSignup = LDAPAllowSignup
 }
 
-func (cfg *Cfg) readAWSConfig() {
+func (cfg *Cfg) handleAWSConfig() {
 	awsPluginSec := cfg.Raw.Section("aws")
 	cfg.AWSAssumeRoleEnabled = awsPluginSec.Key("assume_role_enabled").MustBool(true)
-	allowedAuthProviders := awsPluginSec.Key("allowed_auth_providers").String()
+	allowedAuthProviders := awsPluginSec.Key("allowed_auth_providers").MustString("default,keys,credentials")
```
sunker, line 948: We default to the auth provider that we've had for some time
See issue: https://github.com/grafana/grafana/issues/31892

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -65,20 +49,23 @@ func init() {
...
 }
 
 func (s *CloudWatchService) Init() error {
+	s.sessions = awsds.NewSessionCache()
 	return nil
 }
 
 func (s *CloudWatchService) NewExecutor(*models.DataSource) (plugins.DataPlugin, error) {
-	return newExecutor(s.LogsService, s.Cfg), nil
+	return newExecutor(s.LogsService, s.Cfg, s.sessions), nil
```
sunker, line 61: A new executor is created for each new query, basically making the cache useless since it would be re-created every time. Instead I initialize the cache in CloudWatchService.Init. That way the cache will become global the way it has always been for the CloudWatch data source. We should consider making it non global, but that is something that should be done in a separate pr.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -91,135 +78,13 @@ type cloudWatchExecutor struct {
...
-
-	sessCacheLock.Lock()
-	sessCache[cacheKey] = envelope{
-		session:    sess,
-		expiration: expiration,
-	}
-	sessCacheLock.Unlock()
-
-	return sess, nil
+	return e.sessions.GetSession(region, *awsDatasourceSettings)
```
sunker, line 87: I think it's correct that we use the free region variable throughout the whole GetSession method. The awsDatasourceSettings.Region (previously dsInfo.Region) is supposed to correspond to the region that is defined in the data source configuration page. So this variable should only be changed in case it's changed through config page ui or provisioning. So in the GetSession method, we should never overwrite this variable, but use its value in case the free region has a value of 'default'.

sunker:
I've picked up work on this PR again. Please see reasoning behind that in [this](https://github.com/grafana/grafana/issues/31890) issue. I've made a few changes to the code, and most of them should be commented in this PR.

@patstrom thanks for feedback! I think the changes you're suggestion sounds great and it's something we should look into. From what I understand you're not suggesting any specific changes to the way authentication is implemented, so there's nothing stopping us from getting this merged and then refactor so that we have one session cache (and all the other metric find caches) per executor in a separate PR in the near future.

package.json
```diff
@@ -202,7 +202,7 @@
   },
   "dependencies": {
     "@emotion/core": "10.0.27",
-    "@grafana/aws-sdk": "0.0.2",
+    "@grafana/aws-sdk": "0.0.22",
```
sunker, line 205: using the latest version of the frontend package

pkg/plugins/backendplugin/manager/manager.go
```diff
@@ -86,6 +88,13 @@ func (m *manager) Register(pluginID string, factory backendplugin.PluginFactoryF
 		}
 	}
 
+	if m.Cfg.AWSAssumeRoleEnabled {
+		hostEnv = append(hostEnv, awsds.ENV_VAR_AssumeRoleEnabled+"=true")
+	}
+	if len(m.Cfg.AWSAllowedAuthProviders) > 0 {
+		hostEnv = append(hostEnv, awsds.ENV_VAR_AllowedAuthProviders+"="+strings.Join(m.Cfg.AWSAllowedAuthProviders, ","))
+	}
```
aknuds1, line 96: It's confusing that these AWS specifics show up here in otherwise generic plugin management code, could you stick them in an AWS specific function (called from here)? It would make the code easier to follow at least for me.

aknuds1 (approved):
AFAICT it looks good, but assuming you have tested. Does the caching strategy change with this PR?

pkg/plugins/backendplugin/manager/manager.go
sunker: Sounds good, will fix!

sunker:
> AFAICT it looks good, but assuming you have tested. Does the caching strategy change with this PR?

No the caching strategy will be the same - it will still be one global cache for all cloudwatch instances on the server. 

Regarding testing - have tested this locally and on an ec2 instance and it seems to work fine. I plan to do more testing throughout the day, and if that goes well I aim to publish the changes in the SDK and merge this PR tomorrow.

sunker (approved):

patstrom:
@sunker That's correct. The authentication changes (auth providers and whatnot) themselves I proposed way back has been merged already by Arve. My only remaining concern is the caching strategy as I described it in my last post.

Nice work with this PR!

PR #29553, opened 2020-12-02
GraphNG: support auto and explicit axis width

### LINKED ISSUE #4469: [Feature Request] Option to align/match Y-axis width

ryantxu:
Fixes #4469

Explicit:
image

Auto:
image

### DIFF
--- a/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
+++ b/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
@@ -114,13 +114,13 @@ export const GraphNG: React.FC<GraphNGProps> = ({
       }
 
       const fmt = field.display ?? defaultFormatter;
-      const scale = config.unit || '__fixed';
-      const isNewScale = !builder.hasScale(scale);
+      const scaleKey = config.unit || '__fixed';
 
-      if (isNewScale && customConfig.axisPlacement !== AxisPlacement.Hidden) {
-        builder.addScale({ scaleKey: scale, min: field.config.min, max: field.config.max });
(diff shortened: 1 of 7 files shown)


leeoniya:
a proper fix for the linked issue has to address:

- explicit sizing of all y axes (secondary, tertiary, etc)
- explicit sizing of x gutters (the y side which contains no axes)

without these, the requirement shown in the image with the initial report cannot be satisfied.

ryantxu:
I'm pretty sure we support:
> explicit sizing of all y axes (secondary, tertiary, etc)

The caveat is that the scale&axis settings are derived from the union of all the fields that share the same unit -- it is a little weird, but in practice I think that makes the most sense and fits with existing infrastructure pretty clearly

dprokop:
> > I'm pretty sure we support:

> explicit sizing of all y axes (secondary, tertiary, etc)

Yeah, with the changes in this PR that makes field config's axis width being used back again, we can size axes using the overrides.

dprokop (approved):
Just tested this, and it works as expected!

torkelo (approved):
LGTM

leeoniya:
this is not a complete fix for #4469 without explicit x gutter size control.

PR #29556, opened 2020-12-03
Overrides: show category for override properties

ryantxu:
The current UI does not show overides category -- that makes picking things under "Axis" awkward

| before | after |
|---|---|
| image | image |


hugohaggmark (approved):
Great improvment ⭐

PR #29558, opened 2020-12-03
Grafana/ui: Fix typing non-numerical characters into max data points input

### LINKED ISSUE #29207: Grafana/ui: Input type="number" should consider various browser implementations
**What happened**:
Issue #28487 has highlighted that Firefox and Safari allow a user to type any character into an `input[type="number"]` field rather than Chromes approach of limiting numeric character input. If a browser doesn't enforce numeric input the user can be left confused as currently there is no visual feedback or implementation alignment.

If a user inputs invalid characters into a numeric field they should either be blocked from entering such values or the input should reflect an invalid state to indicate something is wrong.

In Firefox open the panel editor, then in Query Options type alphanumeric characters into Max Data Points.

**Environment**:
- Grafana version: Grafana v7.3.3 (2489dc4d3a)
- User OS & Browser: Mac OS X Firefox and Mac OS X Safari
- Grafana plugins: N/A
- Others:

chihsuan:

This PR fixed the issue that Firefox and Safari allow a user to type any non-numerical characters into an **input[type="number"]** of the max data points field. This changes will reset the UI when onblur event is triggered if the value is invalid as below.

ezgif-7-4705b1b5201c

Fixes #29207

I found that the Input UI won't be changed in the Firefox when the value is non-numerical so I use the ref to force it to change. There is also an old discussion (https://github.com/facebook/react/issues/1549).

### DIFF
--- a/packages/grafana-ui/src/components/Input/Input.tsx
+++ b/packages/grafana-ui/src/components/Input/Input.tsx
@@ -1,4 +1,4 @@
-import React, { HTMLProps, ReactNode } from 'react';
+import React, { HTMLProps, ReactNode, useRef, useCallback, RefObject } from 'react';
 import { GrafanaTheme } from '@grafana/data';
 import { css, cx } from 'emotion';
 import { getFocusStyle, sharedInputStyle } from '../Forms/commonStyles';
@@ -214,7 +214,19 @@ export const getInputStyles = stylesFactory(({ theme, invalid = false, width }: 
 });
 
 export const Input = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
-  const { className, addonAfter, addonBefore, prefix, suffix, invalid, loading, width = 0, ...restProps } = props;
+  const {
+    className,
+    type,
+    addonAfter,
+    addonBefore,
+    prefix,
+    suffix,
+    invalid,
+    loading,
+    onBlur,
+    width = 0,
+    ...restProps
+  } = props;
   /**
    * Prefix & suffix are positioned absolutely within inputWrapper. We use client rects below to apply correct padding to the input
    * when prefix/suffix is larger than default (28px = 16px(icon) + 12px(left/right paddings)).
@@ -225,6 +237,21 @@ export const Input = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
 
   const theme = useTheme();
   const styles = getInputStyles({ theme, invalid: !!invalid, width });
+  const innerRef = useRef<HTMLInputElement>(null);
+
+  const onNumberInputBlur = useCallback(
+    (e: React.FocusEvent<HTMLInputElement>) => {
+      if (isNaN(parseInt(e.target.value, 10))) {
+        // Force UI change for Firefox and Safari issue #29207
+        ((ref as RefObject<HTMLInputElement>) || innerRef).current!.value = '';
+      }
+
+      if (onBlur) {
+        onBlur(e);
+      }
+    },
+    [onBlur]
+  );
 
   return (
     <div className={cx(styles.wrapper, className)}>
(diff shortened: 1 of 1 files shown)


mckn (commented):
I think this looks really good. Great work! However, I am thinking that we might do this within the input field to get this behavior on all inputs with type number.

I will add some more reviewers for second opinion.

dprokop (changes requested):
@chihsuan thanks for the contribution. However, I agree with @mckn comment that this behaviour should be integrated directly into Input[type=number] component rather than a specific usage of it. Let me know if you need any guidance around this and we are happy to help!

Ref:
> I think this looks really good. Great work! However, I am thinking that we might do this within the input field to get this behavior on all inputs with type number.

chihsuan:
@mckn @dprokop thanks so much for the review and comments. I just made the changes to integrate it into Input component. Could you please review again? thanks.

packages/grafana-ui/src/components/Input/Input.tsx
```diff
@@ -225,6 +237,21 @@ export const Input = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
 
   const theme = useTheme();
   const styles = getInputStyles({ theme, invalid: !!invalid, width });
+  const innerRef = useRef<HTMLInputElement>(null);
+
+  const onNumberInputBlur = useCallback(
+    (e: React.FocusEvent<HTMLInputElement>) => {
+      if (isNaN(parseInt(e.target.value, 10))) {
```
jackw, line 244: It might an idea to use `e.target.validity.valid` here?

packages/grafana-ui/src/components/Input/Input.tsx
```diff
@@ -225,6 +237,21 @@ export const Input = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
...
+        // Force UI change for Firefox and Safari issue #29207
+        ((ref as RefObject<HTMLInputElement>) || innerRef).current!.value = '';
```
jackw, line 246: Do we need the ref here if we already have access to the element value via `event.target.value`?

jackw (commented):
Thanks so much for looking into this @chihsuan. I've left a couple of comments / questions for your consideration.

I wonder if we should be giving the user visual feedback so they realise what they typed is incorrect as well as or instead of clearing the value. 🤔

packages/grafana-ui/src/components/Input/Input.tsx
chihsuan: sure!

packages/grafana-ui/src/components/Input/Input.tsx
chihsuan: Indeed. we don't need ref, thanks.

chihsuan:
@jackw Thanks for leaving the comments. 

Agree that giving the user visual feedback might be better. Should we use the similar approach like [LoginForm](https://github.com/grafana/grafana/blob/6a97c456ba93434d9d02ca62d5e4346cc292bcbf/public/app/core/components/Login/LoginForm.tsx#L37) that wrapping with the Field component inside the Input component?

The UI effect might like below
Screen Shot 2020-12-17 at 1 35 36 PM

jackw:
@chihsuan apologies for not replying sooner. We discussed this and think it might be easier if we align this behaviour with the way chrome handles `input[type="number"]` so a user can only ever input numerical values. That should be possible by making use of `onChange` (or a similar event) rather than `onBlur` to do the checking before updating the value. That should circumvent the need to display errors or warnings. What do you think?

chihsuan:
@jackw agree It would be easier for implementation. I think the disadvantage is that blocking user input is not user-friendly but if the consistency between different browsers is important, then, It's true that we could just make it the same. I'll update this PR later. thanks!

chihsuan:
@jackw, it seems that Safari and Firefox won't fire the `onChange` event if the input is non-numerical characters. 
Example: https://jsfiddle.net/chihsuan/2q51ck9g/10/

hugohaggmark:
I don't think there is a simple way to fix this even if you can check `e.target.validity.valid` there are issues with special characters like `.`, `,`, `-`, `e` in FireFox and Safari when I tried this with `onBeforeInput|onInput` in #32944. To solve this, I think we need to build a proper parser for this and I'm not sure that is something we should do and maintain in the long run. 

I purpose we close this PR, what do you think @mckn @jackw @dprokop?

Issue #29376, opened 2020-11-25
Investigate gosec warnings G304/G307


aknuds1:
**What happened**:
Investigate gosec warnings in our code:

- [x] [G304](https://securego.io/docs/rules/g304.html)
- [x] G307: Deferring a method which returns an error

Reproduce them by disabling default golangci-lint filter, putting the following in scripts/go/configs/.golangci.toml:

```toml
[issues]
exclude-use-default = false
```

You might also want to add filters for other warnings than the two in question (G304/G307).

aknuds1:
This is done.

PR #29562, opened 2020-12-03
Security: Fix gosec G307

aknuds1:
Fix gosec warning G307.

See #29376.

### DIFF
--- a/pkg/services/alerting/notifiers/discord.go
+++ b/pkg/services/alerting/notifiers/discord.go
@@ -2,6 +2,7 @@ package notifiers
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"mime/multipart"
 	"os"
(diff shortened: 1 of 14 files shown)


pkg/services/alerting/notifiers/slack.go
```diff
@@ -346,37 +346,47 @@ func slackFileUpload(evalContext *alerting.EvalContext, log log.Logger, url stri
...
 		return nil, b, err
 	}
 	// Add the channel(s) to POST to
-	err = w.WriteField("channels", recipient)
-	if err != nil {
+	if err := w.WriteField("channels", recipient); err != nil {
 		return nil, b, err
 	}
-	w.Close()
+	if err := w.Close(); err != nil {
```
bergquist, line 387: Any reason for closing it twice? Shouldn't the deferred call on the top be good enough?

pkg/services/alerting/notifiers/discord.go
```diff
@@ -192,7 +201,9 @@ func (dn *DiscordNotifier) embedImage(cmd *models.SendWebhookSync, imagePath str
 		return err
 	}
 
-	w.Close()
+	if err := w.Close(); err != nil {
```
bergquist, line 204: Any reason for closing it twice? Shouldn't the deferred call on the top be good enough?

bergquist (approved):
LGTM

two questions but not blockers.

pkg/services/alerting/notifiers/slack.go
aknuds1: @bergquist Yes there is a reason :) It's actually a known pattern in Go, because returning an error from `defer` can be tricky. Either you have to change the function to use named errors, so `defer` can return an error, or you do as I've done here, i.e. close explicitly and return any error, and close in a `defer` just so you make sure it's closed also on error paths.

The second close shouldn't matter from what I've read, it's safe to close a resource several times.

pkg/services/alerting/notifiers/discord.go
aknuds1: @bergquist See my comment above.

PR #29563, opened 2020-12-03
CloudWatch: Namespace in search expression should be quoted if match exact is enabled

### LINKED ISSUE #29109: CloudWatch: namespace in search expression should be quoted if match exact is enabled
**What happened**:
When I try to create a chart with one dimension set, and another with a wildcard, on a custom namespace, I get a 400 error from the cloudwatch api.

Data is clearly in cloudwatch
image

This illustrates the problem, you can see the filtered list of dimension values, but selecting `*` does not show the chart as you would expect.  However taking off exact match does work, but this is not suitable for our use case.

You would expect a series to be shown for each of the unique values of the second dimension.

Can't reproduce, but the critical configurations are: -
* Custom namespace on AWS
* Containing a metric with 2 dimensions that may not have data in the last 24 hours
* This metric has multiple combinations of dimensions for the same metric name/namespace, to enable drilling into the data
* Exact match enabled

**Environment**:
- Grafana version: 7.3.1
- Data source type & version: Cloudwatch
- OS Grafana is installed on: CentOs
- User OS & Browser: Ubuntu FireFox
- Grafana plugins: All standard + PagerDuty
- Others:

ying-jeanne:
…exact is enabled #29109

**What this PR does / why we need it**: The pr is to fix problem of issue #29109 

Fixes #29109 

### DIFF
--- a/pkg/tsdb/cloudwatch/metric_data_query_builder.go
+++ b/pkg/tsdb/cloudwatch/metric_data_query_builder.go
@@ -88,12 +88,11 @@ func buildSearchExpression(query *cloudWatchQuery, stat string) string {
 	}
 
 	if query.MatchExact {
-		schema := query.Namespace
+		schema := fmt.Sprintf("%q", query.Namespace)
 		if len(dimensionNames) > 0 {
 			sort.Strings(dimensionNames)
 			schema += fmt.Sprintf(",%s", join(dimensionNames, ",", `"`, `"`))
 		}
-
 		return fmt.Sprintf("REMOVE_EMPTY(SEARCH('{%s} %s', '%s', %s))", schema, searchTerm, stat, strconv.Itoa(query.Period))
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/tsdb/cloudwatch/metric_data_query_builder.go
```diff
@@ -88,12 +88,11 @@ func buildSearchExpression(query *cloudWatchQuery, stat string) string {
 	}
 
 	if query.MatchExact {
-		schema := query.Namespace
+		schema := fmt.Sprintf(`"%s"`, query.Namespace)
```
papagian, line 91:
```suggestion
		schema := fmt.Sprintf("%q", query.Namespace)
```

I would just using `%q` instead that prints a double-quoted string safely escaped with Go syntax.

papagian (changes requested):
It works fine but I have a small suggestion.

papagian (approved):
LGTM

ying-jeanne:
beforechange
afterchange

the search with "" or not seems having the same meaning, compared the graph with 2 different ways, seems the same, we could also compare the response from backend regarding the data points to make sure.

ying-jeanne:
image
image

the input is configured with matrix = * and match exact turned on, in the screen copy, we can see the search expression generated are different (after change with double-quote on the namespace, before none), the search result dataframe are exactly the same, so no regression introduced.

sunker (approved):
LGTM

PR #29564, opened 2020-12-03
middleware: Make scenario test functions take a testing.T argument

aknuds1:
Fix a bug in middleware tests, where they use the wrong `testing.T` instance. Basically, the issue is that utility functions like `middlewareScenario` will create a new sub-test, but don't pass the sub-test's `testing.T` instance on to the callbacks they receive. As a result, the callbacks end up using the `testing.T` instance belonging to the parent test, which means for example that you can't run sub-tests in isolation.


wbrowne (approved):
🕺

PR #29565, opened 2020-12-03
PanelLibrary: Adds library_panel table

hugohaggmark:
Adds the panellib table to the db if the feature toggle `panelLibrary` is enabled. Not sure what the exact table design and indexes should be at the moment so this is just a first POC.

Fixes #


pkg/services/panellib/panellib.go
```diff
@@ -0,0 +1,61 @@
...
+
+// AddMigration defines database migrations.
+// If Panel Library is not enabled does nothing.
+func (pl *PanelLib) AddMigration(mg *migrator.Migrator) {
+	if !pl.IsEnabled() {
+		return
+	}
+
+	panellibV1 := migrator.Table{
+		Name: "panellib",
```
torkelo, line 46: panel_lib ? the entity/model name feels like it will be LibraryPanel, or PanelLibraryItem ? 

Hm.. we can also think about this with global variables in mind. maybe the tables are just panel and (variable for future global variables)?  

So 
* panellib  (hard to read with the double lowercase l)
* panel_lib
* library_panel
* library_variable
* library_dashboard_template
* panel

naming is hard :)

torkelo (approved):
added some naming options :)

pkg/services/panellib/panellib.go
hugohaggmark: I think you make a valid point, my first thought was `panel` as well but I dropped it. I think the `library_xyz`naming is better so I'll change. Thank you

pkg/services/librarypanel/librarypanel.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"github.com/grafana/grafana/pkg/infra/log"
+	"github.com/grafana/grafana/pkg/registry"
+	"github.com/grafana/grafana/pkg/services/sqlstore"
+	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
+	"github.com/grafana/grafana/pkg/setting"
+)
+
+// LibraryPanel is the service the Panel Library feature.
+type LibraryPanel struct {
```
torkelo: LibraryPanelService? (to not confuse it with the entity)

torkelo: or PanelLibraryService

hugohaggmark: I just pushed that change :)

torkelo: lol, was thinking dam that was fast!

PR #29575, opened 2020-12-03
GraphNG: adding possibility to toggle tooltip, graph and legend for series

mckn:
Because it can be awesome.

This is very much work in progress.


ryantxu:
> Because it can be awesome.
:heart: now I know what to write in that line!!!!

mckn:
So I have something working okay. There are still some work to clean this upp by adding tests, refactoring some code etc. Before I start doing that I would like to get some feedback on this solution.

I have introduced something I called `SystemConfigOverrideRule` that basically is a special kind of override rule that the system adds. Also, I have added a helper function that helps me map from an `AlignedDataField` to a `FieldIndexRef` that basically just points to the indexes in the original set of data.

toggle_legend

Some things that can get tricky with this solution:

- Since we are adding an override then you toggle/change visibility of a series you can end up in a situation where the user previously have added an override that will result in when trying to toggle/change a graph nothing happens. See example below.

Skärmavbild 2020-12-09 kl  21 38 09

- The solution of having a `__systemRef` will not work if we would like to have multiple system overrides with the same system reference.

- The regex being created can potentially be really big. This can be resolved by introducing a smartness that change from excluding everything that matches to includes.

Apart from that I think this could work as a solution. The scenario where the user ends up in a situation where toggling/changing a series in the legend doesn't affect the visualization can't be resolved if we want to use overrides. The only way, that I can think of, to resolve that is by introducing some new kind of hidden (for the user) override but then I feels like this should not be an override but something else.

@ryantxu @torkelo @dprokop @leeoniya would love to get your input on this before I continue. Thanks 🙏

dprokop:
Lovely, will dig deeper into the guts tomorrow mornign!

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -20,24 +20,30 @@ import { LegendDisplayMode, LegendItem, LegendOptions } from '../Legend/Legend';
...
   legend?: LegendOptions;
   fields?: XYFieldMatchers; // default will assume timeseries data
+  onLegendClick?: (event: GraphNGLegendEvent) => void;
 }
 
 const defaultConfig: GraphFieldConfig = {
   drawStyle: DrawStyle.Line,
   showPoints: PointVisibility.Auto,
   axisPlacement: AxisPlacement.Auto,
+  seriesConfig: {
```
ryantxu, line 42: maybe `fieldVisibility`?  or `showFieldIn` or `showField`

avoid "DisplayField" since that is too close to [FieldDisplay](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/field/fieldDisplay.ts#L65)

ryantxu: Since we want the default to be visible in everything, could we have the options be:
```
hideFrom: {
   graph: false,
   legend: false,
   tooltip: false,
}
```

That will make the UI more complicated, but then checking for `config.hideFrom?.graph` would default more smoothly

public/app/plugins/panel/graph3/configFactory.ts
```diff
@@ -0,0 +1,126 @@
...
+    existing.push(displayName);
+  } else {
+    existing.splice(index, 1);
+  }
+
+  return {
+    __systemRef: displayOverrideRef,
+    matcher: {
+      id: FieldMatcherID.byRegexp,
+      options: `^(?!${existing.join('|')}$).*$`,
```
ryantxu, line 104: rather than use a standard regex matcher that exposes an editor in the overrides, what about a custom one that is not end-user editable?  ("remove" would still be there)

This *may* also be a clean way to support:  
image

although a transformation may make more sense

ryantxu: regardless a custom matcher would let you more clearly control "only field X" vs "everything except X"

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -63,11 +63,18 @@ export interface AxisConfig {
   axisWidth?: number; // pixels ideally auto?
 }
 
+export interface SeriesConfig extends Record<string, any> {
```
ryantxu, line 66: why Record?

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
mckn: Great suggestions!

public/app/plugins/panel/graph3/configFactory.ts
mckn: We actually started with a component looking like the one you posted but changed to the multiselect (which is far from optimal). But I agree with you that the example you posted looks better! I will ask for some feedback from the UX team on this.

mckn: Another approach could also be that if the user change the override we make sure the exclude the (__systemRef) so it is no longer considered as an system override but an user override instead.

torkelo: yea, I like the idea of a custom hidden matcher that the user can't edit directly, only delete.

packages/grafana-ui/src/components/uPlot/config.ts
mckn: https://github.com/grafana/grafana/pull/29575/files#diff-9dc503bbea5a0385b2a5675eb77afd7ebbc71cd4c294d7cbcd744e77c457ee81R15

We will get a type error about not being able to iterate over it if we don't have it.

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
dprokop: @ryantxu yups, we discussed this with @mckn yesterday and it feels like showIn* option in the UI is super confusing, especially if you look at the overrides config.

public/app/plugins/panel/graph3/configFactory.ts
dprokop: @torkelo @ryantxu @mckn I am considering two options here regarding the "system" overrides - either they are completely hidden and controlled by the user from other parts of the UI (panel in particular) or readonly. Rather than having a readonly matcher, we would consider an entire override not editable directly by the user. What do you think?

packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -15,6 +15,26 @@ export interface ConfigOverrideRule {
   properties: DynamicConfigValue[];
 }
 
+/**
+ * Describes config override rules created when interacting with Grafana.
+ *
+ * @public
```
dprokop, line 21: Maybe we should consider this an internal thingy?

packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -30,7 +30,13 @@ export abstract class PlotConfigBuilder<P, T> {
...
 
+export interface FieldIndexRef {
+  frameIndex: number;
+  fieldIndex: number;
+}
+
 export interface AlignedFrameWithGapTest {
   frame: DataFrame;
   isGap: Series.isGap;
+  getFieldIndexRef: (alignedFieldIndex: number) => FieldIndexRef | undefined;
```
dprokop, line 41: Wonder if we shouldn't throw for not existent indexes rather than returning `undefined`? I guess getFieldIndex range will always return something for a valid index and accessing anything out of range could be potentially just a bug in the code.

packages/grafana-ui/src/components/GraphNG/types.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export enum GraphNGLegendEventMode {
+  select = 'select',
+  append = 'append',
+}
+
+export interface GraphNGLegendEvent {
+  field: Field;
+  frame: DataFrame;
+  data: DataFrame[];
```
dprokop, line 11: As we discussed I think we could use indexes instead of passing the entire shebang here:)

packages/grafana-ui/src/components/Legend/Legend.tsx
```diff
@@ -36,12 +36,18 @@ export type LegendPlacement = 'bottom' | 'right';
 
 export interface LegendOptions extends LegendBasicOptions, LegendRenderOptions {}
 
+export interface LegendSeriesRef {
+  fieldIndex: number;
+  frameIndex: number;
+}
```
dprokop, line 42: Maybe we could call this `FieldRef` or `DataFrameFieldRef` and move this to grafana/data?

packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -30,7 +30,13 @@ export abstract class PlotConfigBuilder<P, T> {
   abstract getConfig(): T;
 }
 
+export interface FieldIndexRef {
+  frameIndex: number;
+  fieldIndex: number;
+}
```
dprokop, line 36: See my other comment about `LegendSeriesRef` type that's exactly the same :)

dprokop (commented):
Great progress on this one @mckn !

packages/grafana-ui/src/components/uPlot/config.ts
torkelo: but this removes a lot of type checking from the type. Might be better to just typecast it to Record in the place you want to iterate over it

packages/grafana-data/src/types/fieldOverrides.ts
mckn: Ok ok, I will add that.

torkelo:
Can we merge sync this to master? Eager to test the latest changes :)

public/app/plugins/panel/graph3/configFactory.ts
```diff
@@ -0,0 +1,140 @@
...
+const createFreshOverride = (displayName: string): SystemConfigOverrideRule => {
+  return {
+    __systemRef: displayOverrideRef,
+    matcher: {
+      id: FieldMatcherID.readOnly,
+      options: {
+        innerId: FieldMatcherID.byRegexp,
+        innerOptions: `^(?!${displayName}$).*$`,
+        formattedValue: `All series except: ${displayName}`,
+      },
```
dprokop, line 75: I really like the idea of readonly matcher that actually takes other matcher as an option, very clever.

1. What I'm not sure about tho is whether or not we should have the formattedValue being displayed or rather the matcher editor with read-only property. Comparing the UI of the two alternatives I find the editor working better with the read-only input making the entire UI more consistent. 
image
vs
image
However, the big advantage of your proposal that the matcher rule is human readable (I'm not saying that only robots can read regex ;) )

2. Another thing regarding the _system_ overrides is whether or not we should allow multiple properties to be overridden for such an override. I mean, I think the `Add override property`  and property removal buttons should be disabled IMO:
Screen Shot 2020-12-15 at 14 24 11

3. For the read-only / system override I think we should add some kind of indication or maybe a tooltip why people can not edit those. Maybe @jessover9000 or @hemdrup could help us with this? 

In general I feel like the ui of the system/readonly overrides is something worth bringing to the UX feedback session tomorrow.

packages/grafana-data/src/transformations/matchers/readOnlyMatcher.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface ReadOnlyFieldMatcherOptions<T = any> {
+  innerId: FieldMatcherID;
+  innerOptions: T;
+  formattedValue: string;
+}
+
+const defaultMatcher: FieldMatcher = () => false;
+
+fieldMatchers.register({
+  id: FieldMatcherID.readOnly,
```
dprokop, line 14: We can exclude the readonly matchers from the matchers select as these are not usable for people 🙈 
image

packages/grafana-data/src/transformations/matchers/readOnlyMatcher.ts
```diff
@@ -0,0 +1,33 @@
...
+  name: 'Read-only field matcher',
+  description: 'Field name by inner matcher',
+  get: (options: ReadOnlyFieldMatcherOptions): FieldMatcher => {
+    if (options.innerId === FieldMatcherID.readOnly) {
+      return defaultMatcher;
+    }
```
dprokop, line 20: In such case there is something definitely wrong - what about throwing an error rather than returning default matcher?

dprokop (commented):
Wow, loads of progress here, awesome work @mckn! Code wise I think all looks good. Great idea with the read-only matcher.  

My major concerns resonate around the UI and the general experience of the readonly matcher. I have also found a regression in the keyboard behaviour when compared with Angular graph. 

Let me know if any of the comments is not clear.

torkelo:
Great progress! 

1) Took this for a spin, but when I tested this with a graph with 3 series it did not show the one I clicked on. The series names where `A-series`,  `A-series1`, `A-series2`, seemed like it always showed the series named `A-series` no matter which I clicked on. 

2) Regarding dominiks comment regarding the editor, not sure I think I do prefer the plain text over say a complicated regex. 

3) I do not like the use of inline styles here, feels very inconsistent. But not sure what the other option is. Also not sure about the hide in tooltip option. Is that not the same as hide in Graph? When do you want to show something in tooltip but not in Graph or vice versa? If we remove that one we can make it into a radio button. 

4) Need to hide this option from the Field tab as it only makes sense as an override. 

A super ticky future problem: How can we ignore these changes when made from the dashboard and not have them trigger unsaved changes warning?

dprokop:
> I do not like the use of inline styles here, feels very inconsistent. But not sure what the other option is. Also not sure about the hide in tooltip option. Is that not the same as hide in Graph? When do you want to show something in tooltip but not in Graph or vice versa? If we remove that one we can make it into a radio button.

I think we have discussed that and the case is when one has long series name that makes te tooltip giant.
 
> Need to hide this option from the Field tab as it only makes sense as an override.

Yes, I'm the API to enable this

@mckn I have found one weird behaviour regarding series color, mind what happens with 2 series scenario when you clikck the 2nd series. The first one gets the color of the second one.
Kapture 2020-12-15 at 15 14 05

packages/grafana-ui/src/components/GraphNG/types.ts
```diff
@@ -0,0 +1,11 @@
+import { DataFrameFieldIndex } from '@grafana/data';
+
+export enum GraphNGLegendEventMode {
+  select = 'select',
+  append = 'append',
```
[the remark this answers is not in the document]
mckn: Have updated the names so they are much more clear.

mckn:
> Great progress!
> 
> 1. Took this for a spin, but when I tested this with a graph with 3 series it did not show the one I clicked on. The series names where `A-series`,  `A-series1`, `A-series2`, seemed like it always showed the series named `A-series` no matter which I clicked on.
> 2. Regarding dominiks comment regarding the editor, not sure I think I do prefer the plain text over say a complicated regex.
> 3. I do not like the use of inline styles here, feels very inconsistent. But not sure what the other option is. Also not sure about the hide in tooltip option. Is that not the same as hide in Graph? When do you want to show something in tooltip but not in Graph or vice versa? If we remove that one we can make it into a radio button.
> 4. Need to hide this option from the Field tab as it only makes sense as an override.
> 
> A super ticky future problem: How can we ignore these changes when made from the dashboard and not have them trigger unsaved changes warning?

Great feedback @torkelo! Me and @dprokop are talking about it now.

mckn:
> Great progress!
> 
> 1. Took this for a spin, but when I tested this with a graph with 3 series it did not show the one I clicked on. The series names where `A-series`,  `A-series1`, `A-series2`, seemed like it always showed the series named `A-series` no matter which I clicked on.

Great catch! This one should be resolved.

> 2. Regarding dominiks comment regarding the editor, not sure I think I do prefer the plain text over say a complicated regex.

I'm planning on sticking with the plain text.

> 3. I do not like the use of inline styles here, feels very inconsistent. But not sure what the other option is. Also not sure about the hide in tooltip option. Is that not the same as hide in Graph? When do you want to show something in tooltip but not in Graph or vice versa? If we remove that one we can make it into a radio button.

Looking into simplifying this as we speak.

> 4. Need to hide this option from the Field tab as it only makes sense as an override.

Will try to get this in place tomorrow.

> A super ticky future problem: How can we ignore these changes when made from the dashboard and not have them trigger unsaved changes warning?

I think this is a good side effect. However it would be nice to be able to see what have changed from the "unsaved changes warning."

dprokop:
> >A super ticky future problem: How can we ignore these changes when made from the dashboard and not have them trigger unsaved changes warning?
>I think this is a good side effect. However it would be nice to be able to see what have changed from the "unsaved changes warning."

I agree with @mckn that this side effect is actually good, as we are fundamentally changing the behaviour of the legend compared to the Angular graph. I would rather be warned about unsaved changes than surprised that some of my series are not on the graph when I go back to the dashboard

leeoniya:
does show/hide series currently re-init the whole plot?

would be good to optimize this the same as `.setSize()` and use the .setSeries(`{show}`) API.

mckn:
> does show/hide series currently re-init the whole plot?
> 
> would be good to optimize this the same as `.setSize()` and use the `.setSeries(`show`)` API.

hmm, awesome feedback! Will have a look at that tomorrow.

leeoniya:
i pushed a few uPlot commits today (unreleased) that fixed some regressions in the legend marker styling. there is now more explicit control over the legend markers, and they can be styled pretty easily as the current Flot chart (i think).

since there is already a bunch of legend-related code in uPlot, as well as things like hover-focus, toggle sync, etc. it's a shame that none of that is being used and has to be re-implemented on the Grafana side. the legend appears to be pretty simple, so i was curious what additional config options you guys feel would be necessary to use uplot's native legend. this may also eliminate the need to optimize series toggling and remove some possible perf-footguns.

what i see missing right now is a legend "solo" mode, which inverts uplot's native click-toggle behavior, so that it isolates the series rather than toggling it. also the tooltip on hover, which i can expose a callback for (there are already mouseenter/mouseleave listeners on the legend items to handle hover-focus).

torkelo:
> what i see missing right now is a legend "solo" mode, which inverts uplot's native click-toggle behavior, so that it isolates the series rather than toggling it.

think this is the default, when you click on a series Grafana will only show that series. CMD/CTRL + click will hide it 

Yea the legend is complex and re-used in other places. Supports showing reduced values like Last, min, max, supports table mode with sort (that integrates with stacking order) etc. It needs somewhat consistent and reused between panels/visualizations. You can change the series color by opening the color picker from it, and assign y-axis from it. 

It is far from great though, missing a few things (like show live what values you hover over), and it's a bit too complex for simple visualizations that do not need all it supports. Started a simple legend POC https://github.com/grafana/grafana/pull/28841  But not sure, might be better to improve and make the Graph legend more composable and usable for simple scenarios.

dprokop:
> would be good to optimize this the same as .setSize() and use the .setSeries({show}) API. i think this is important, since this is an interaction event, rather than a one-time config/setup event.

I think we can wait with this optimization. Legend interaction isn't sth that would happen at a high rate or often, so i don't conside this worth solving in this PR

torkelo:
Agree with not optimizing this, it finds if this causes a re-init, graph init + rendering is so fast anyway.  

Discovered 2 issues
1) The legend items do not show their disabled state (weak text color). 
2) The overrides in panel edit are not updating, think we need an event similar to how we did for panel options: https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx#L86 ,  We can fix this in another PR, I might be able to fix it, looking at ways to fix related issue in https://github.com/grafana/grafana/issues/29815

dprokop:
First and foremost: excellent work on this @mckn!

Some remarks from the testing

### 1. Keyboard behaviour
When clicking with CMD or Shift key, the behaviour is opposite to the one that Angular graph represents, see the GIF below:
cmdclick
Personally, I think this is the correct and more natural behaviour than the Angular one, as it mimics the behaviour in the OS when you cmd/shift click - basically selecting multiple items in the Finder for instance. Yet, this might be confusing for someone switching to new graph, as there is no "easy way to hide one particular series from multiple ones. This feels like a thing we should bring up with the UX team.

  The difference, in the simples words, can be described as: 
  When clicking cmd/shift click on the series name:
  1) Angular graph **hides**  series
  2) Graph NG **isolates** series

This has a consequences in the single series scenario, see below.

### 2. Click behaviour with single series
Both interactions below were performed with cmd/shift +click . Note the opposite behaviours
cmdclicksingle

### 3. Hide in Tooltip
Hide in tooltip has no effect on the tooltip's content:
tooltip

### 4. Multiple Field by name matchers available
In the matcher selection there are two **Field by name** matchers available. Selecting the second one results in a runtime error. See the attached log file for more details [localhost-1608648330191.log](https://github.com/grafana/grafana/files/5730458/localhost-1608648330191.log)

matcher

### 5. The system override looks great!
image

I think this UI is really nice. The only thing I'm missing here is the information that this is the **system** override and some explanation what it is about. I think we can add that as a follow up :)


So that would be it when it comes to testing, now time for the code review :)

mckn:
> First and foremost: excellent work on this @mckn!
> 
> Some remarks from the testing
> 
> ### 1. Keyboard behaviour
> When clicking with CMD or Shift key, the behaviour is opposite to the one that Angular graph represents, see the GIF below:
> cmdclick
> Personally, I think this is the correct and more natural behaviour than the Angular one, as it mimics the behaviour in the OS when you cmd/shift click - basically selecting multiple items in the Finder for instance. Yet, this might be confusing for someone switching to new graph, as there is no "easy way to hide one particular series from multiple ones. This feels like a thing we should bring up with the UX team.
> 
> The difference, in the simples words, can be described as:
> When clicking cmd/shift click on the series name:
> 
> 1. Angular graph **hides**  series
> 2. Graph NG **isolates** series
> 
> This has a consequences in the single series scenario, see below.
> 
> ### 2. Click behaviour with single series
> Both interactions below were performed with cmd/shift +click . Note the opposite behaviours
> cmdclicksingle
> 
> ### 3. Hide in Tooltip
> Hide in tooltip has no effect on the tooltip's content:
> tooltip
> 
> ### 4. Multiple Field by name matchers available
> In the matcher selection there are two **Field by name** matchers available. Selecting the second one results in a runtime error. See the attached log file for more details [localhost-1608648330191.log](https://github.com/grafana/grafana/files/5730458/localhost-1608648330191.log)
> 
> matcher
> 
> ### 5. The system override looks great!
> image
> 
> I think this UI is really nice. The only thing I'm missing here is the information that this is the **system** override and some explanation what it is about. I think we can add that as a follow up :)
> 
> So that would be it when it comes to testing, now time for the code review :)

Love the feedback! Will fix it now.

dprokop:
Found one more UI thingy. There is no visual clue about series being hidden in the legend if the legend mode is set to Table:
tablelehend

public/app/plugins/panel/graph3/module.tsx
```diff
@@ -154,6 +155,22 @@ export const plugin = new PanelPlugin<Options, GraphFieldConfig>(GraphPanel)
...
+          path: 'hideFrom',
+          defaultValue: {
+            tooltip: false,
+            graph: false,
+            legend: false,
+          },
+          editor: SeriesConfigEditor,
+          override: SeriesConfigEditor,
+          shouldApply: () => true,
+          showIf: () => false,
```
dprokop, line 172: No need for this "hack". You can use the legitimate API introduced in https://github.com/grafana/grafana/pull/29879

dprokop (commented):
Plenty of code to digest, but it feels great! Left few nits only

packages/grafana-data/src/transformations/matchers.ts
```diff
@@ -19,17 +19,19 @@ import { getEqualValueMatchers } from './matchers/valueMatchers/equalMatchers';
...
 import { getSimpleFieldMatchers } from './matchers/simpleFieldMatcher';
 import { getRegexValueMatcher } from './matchers/valueMatchers/regexMatchers';
+import { getReadOnlyFieldMatchers } from './matchers/readOnlyMatcher';
 
 /**
  * Registry that contains all of the built in field matchers.
  * @public
  */
-export const fieldMatchers = new Registry<FieldMatcherInfo>(() => {
+export const fieldMatchers: Registry<FieldMatcherInfo> = new Registry<FieldMatcherInfo>(() => {
```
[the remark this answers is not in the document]
mckn: Because I am referencing the `fieldMatchers` in the init function.

packages/grafana-data/src/transformations/matchers/nameMatcher.test.ts
```diff
@@ -113,7 +114,29 @@ describe('Field Multiple Names Matcher', () => {
     });
     const config = {
       id: FieldMatcherID.byNames,
-      options: ['C'],
+      options: {
+        mode: ByNamesMatcherMode.allIn,
```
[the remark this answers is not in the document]
mckn: hahah

torkelo:
Took this for a new spin! Looking great now, the UI for the option and matcher is looking good! 

And now the hidden series are gray in the legend! 

Discovered 2 issues

1) Could not toggle series by clicking again and go back to default mode where all series are visible 
2) Holding down shift to show more than one series in exclusive mode did not work

packages/grafana-ui/src/components/GraphNG/types.ts
```diff
@@ -0,0 +1,20 @@
+import { DataFrameFieldIndex } from '@grafana/data';
+
+/**
+ * Mode to describe if a legend is isolated/selected or being appended to an existing
+ * series selection.
+ * @public
+ */
+export enum GraphNGLegendEventMode {
+  toggleSelection = 'select',
+  appendToSelection = 'append',
```
torkelo, line 10: For enums we usually use PascalCase

public/app/plugins/panel/graph3/hideSeriesConfigFactory.ts
```diff
@@ -0,0 +1,147 @@
...
+    matcher: {
+      id: FieldMatcherID.readOnly,
+      options: {
+        prefix: 'All except:',
+        matcherId: FieldMatcherID.byNames,
+        matcherOptions: {
+          mode: ByNamesMatcherMode.exclude,
+          names: names,
+        },
+      },
```
torkelo, line 134: Don't really understand this, the id of the matcher is readOnly? should this not be the "all series except" matcher? Is readonly not a property on the override instead of being a nested override? that feels more complex, but not sure if that is what is done here nesting overrides? 

Having to specify all names to hide feels like it's going to be a problem. You maybe have 50+ series (dynamic names) but you only want to show one series by default, then the persisted json will include all 49 series that are hidden?  (which could be dynamic and change for each refresh, for a dynamic top 50 query).

torkelo: Ah, on the second point I misunderstood the exclude matcher mechanics

packages/grafana-ui/src/components/GraphNG/types.ts
mckn: Looks like we mix a lot in the code base. But I will change this one. Thanks for the feedback.

public/app/plugins/panel/graph3/hideSeriesConfigFactory.ts
mckn: Yeah, I am thinking that I might simplify this a bit.

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -15,6 +15,7 @@ import { PlotConfigBuilder } from '../types';
 export interface SeriesProps extends LineConfig, AreaConfig, PointsConfig {
   drawStyle: DrawStyle;
   scaleKey: string;
+  show?: boolean;
```
ryantxu, line 18: for the uplot config... rather than requiring an affermative "show", can we do `hide?: boolean`?  this will have better default semantics automatically

ryantxu (commented):
@mckn this is looking great!

There are a few issues I see, but I'm not sure if they are caused by this PR or existing issues with the overrides UI system.

1. When the UI is visible, changes do not seem to update consistently:
update-behavior

2. Removing an override just collapses it
remove-override

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
leeoniya: `!false`

ryantxu (approved):
I just merged master, and the issues I was concerned with are fixed :tada: 

I vote we merge this soon to avoid conflicts with other PRs,

torkelo (approved):
Looks this for a final spin as well and read through the code. 

Works great now!  

Awesome job!

PR #29583, opened 2020-12-03
CloudWatch: Add support for AWS DirectConnect ConnectionErrorCount metric

### LINKED ISSUE #29582: CloudWatch: Add support for AWS DirectConnect new connection error metric

New metric `AWS/DX` > `ConnectionErrorCount`, which replaces `ConnectionCRCErrorCount`, as described in the [documentation](https://docs.aws.amazon.com/directconnect/latest/UserGuide/monitoring-cloudwatch.html).

Also, from the AWS support :
> Please note the ConnectionCRCErrorCount metric is no longer in use although it still shows data points, this metric has been replaced by the ConnectionErrorCount metric which is the error count that occurred since the last reported datapoint. When there are errors on the interface, the metric reports non-zero values.

As the old metric `ConnectionCRCErrorCount` is still present in the AWS console and still shows data, I suggest to keep it in Grafana as well for now. This would also prevent current dashboards from breaking.

**Why is this needed**:

The new metric is needed to properly monitor AWS Direct Connect connections for all types of MAC level errors on the AWS device.

haeringer:
Add support for AWS DirectConnect new metric `ConnectionErrorCount`, as described in the issue.

Fixes #29582


papagian (approved):
> As the old metric ConnectionCRCErrorCount is still present in the AWS console and still shows data, I suggest to keep it in Grafana as well for now. This would also prevent current dashboards from breaking.

👍  I totally agree.

thank you for your contribution!

PR #29589, opened 2020-12-03
Chore: Rewrite sqlstore migration test to use standard library

ceh:

Rewrite sqlstore migration test to use standard library.

Part of #22844.

Simplified the test while here, by removing the table-like driven test for test DBs since it's only testing the SQLite driver (using others, like `MySQLTestDB`, would require running an instance so not for this PR).

What do you think @aknuds1, @papagian?


aknuds1 (approved):
Nice cleanup!

aknuds1:
Thanks!!

PR #29594, opened 2020-12-03
Alerting: Add Rundeck Alert Notifier

gmr:
This PR adds a new notifier type that makes an API request to [Rundeck](https://www.rundeck.com/) to trigger the execution of a Rundeck job on an alert.

While this satisfies the needs I have, I figured I'd make a draft PR to get feedback to see what you all would like to see from it, in order to get it merged.

Right now it only triggers the API call on Alert, which was intentional as to not over complicate it. However, I could see a case being made to allow for jobs to be specified for each state change.


aknuds1:
Thank you for submitting this PR, but we will have to reject it since we're generally not accepting new notifiers into Grafana core.

We are hoping to have a plugin model for notifiers in place in the future, and that will be the solution going forward. Unfortunately it isn't implemented yet.

PR #29596, opened 2020-12-03
GraphNG: time range should match the panel timeRange

ryantxu:
This adds a flag to use the timeRange for x axis, not based on the data

zooms


packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -95,38 +107,15 @@ function initializePlot(data: AlignedDataWithGapTest, config: Options, el: HTMLD
...
+) {
   if (!plotInstance || !data) {
     return;
   }
   pluginLog('uPlot core', false, 'updating plot data(throttled log!)', data);
-  updateScales(frame, config, plotInstance);
   plotInstance.setData(data);
 }
-
-function updateScales(frame: DataFrame, config: UPlotConfigBuilder, plotInstance: uPlot) {
```
ryantxu, line 107: this is not needed since the scale ranges are all defined as functions

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -95,38 +107,15 @@ function initializePlot(data: AlignedDataWithGapTest, config: Options, el: HTMLD
   return new uPlot(config, data, el);
 }
 
-function updateData(frame: DataFrame, config: UPlotConfigBuilder, plotInstance?: uPlot, data?: AlignedData | null) {
+function updateData(
+  frame: DataFrame,
+  config: UPlotConfigBuilder,
+  plotInstance?: uPlot,
+  data?: AlignedDataWithGapTest | null
```
ryantxu, line 114: unrelated change, but needs fixed -- the gaps should be passed in

leeoniya:
my futile attempt: https://github.com/grafana/grafana/tree/graph-ng-extent-leon

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -75,6 +75,7 @@ export const GraphNG: React.FC<GraphNGProps> = ({
     // X is the first field in the alligned frame
     const xField = alignedFrame.fields[0];
     if (xField.type === FieldType.time) {
+      builder.xScaleIsTimeRange = true;
       builder.addScale({
         scaleKey: 'x',
         isTime: true,
```
dprokop, line 81: Maybe instead of setting this flag explicitly we could set it when adding scale that `isTime`? Not sure we will fallback to data set time range in time scenario. WDYT?

packages/grafana-ui/src/components/uPlot/Plot.tsx
dprokop: Great!

dprokop (approved):
Looks good, left one thing to consider @ryantxu

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
torkelo: yea, having both seem strange

packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -40,6 +49,7 @@ export const UPlotChart: React.FC<PlotProps> = props => {
     // 1. When config is ready and there is no uPlot instance, create new uPlot and return
     if (isConfigReady && !plotInstance.current) {
       plotInstance.current = initializePlot(prepareData(props.data), currentConfig.current, canvasRef.current);
+      updateTimeScale();
```
torkelo, line 52: This feels messy to have to set this directly of init, can't we initialize the time scale with the correct (initial) min/max values when we defined the scale?
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx#L80

dprokop: I think the problem will be the invalidation of the config in such case, which is sub optimal as it will reinitialize the entire plot. 

But ref https://github.com/leeoniya/uPlot/blob/005b3e0903444521f592c14cf33ab922d6e364ca/dist/uPlot.d.ts#L156 range of a scale can be defined as a function, then we could extend the UPlotConfigBuilder.addScale interface to accept the range (if it isn't accepting it already), and using references retrieved it in the memoized builder. 

A draft of this approach would be something like this:

```ts
const GraphNG = () => {
  // reference change will not triger re-render
  const currentTimeRange = useRef();
  // keep reference value in sync with the props
  useEffect(() => {
    currentTimeRange.current = props.timeRange;
  }, [props.timeRange]);

  const configBuilder = useMemo(() => {
    const builder = new UPlotConfigBuilder();

    builder.addScale({
      // lazily access reference value without adding dependency to props.timeRange to the memoization
      range: () => [timeRange.current.from.valueOf(), timeRange.current.to.valueOf()],
    });
    return builder;
  }, [configRev, timeZone]);
}
```

leeoniya: @dprokop :100: what i was trying to do.

torkelo: > I think the problem will be the invalidation of the config in such case 

But this case is for when the plot is re-initialized, which means the config is new so not sure that is a problem

leeoniya: defining `range` as a function makes this distinction i irrelevant, as long as it's always able to return the correct current values.

ryantxu:
@dprokop -- I updated with your suggestion.  Now it looks like the range available in the ref is one previous to the current value?

dprokop:
>@dprokop -- I updated with your suggestion. Now it looks like the range available in the ref is one previous to the current value?

@ryantxu yups, I've made a mistake suggesting useEffect instead of useLayoutEffect. It's updated now and seems to be working fine :)

PR #29599, opened 2020-12-04
GraphNG: simple settings migration from flot panel

ryantxu:
This aims to migrate the most basic settings.  

We will need to revist this migration script for better feature parity, but this is a good first step

Top panel auto migration from the bottom one
image


leeoniya:
note to self:

probably change `mode` to `drawStyle`

ryantxu:
Migrations working... but some issues with updates:
migrate-update

I assume that needs fixed in a different issue

public/app/plugins/panel/graph3/migrations.ts
```diff
@@ -0,0 +1,266 @@
...
+    graph.showPoints = PointVisibility.Always;
+  } else if (graph.drawStyle !== DrawStyle.Points) {
+    graph.showPoints = PointVisibility.Never;
+  }
+  if (graph.drawStyle === DrawStyle.Bars) {
+    graph.fillOpacity = 1.0; // bars were always
+  }
+
+  graph.lineWidth = angular.lineWidth;
+  graph.pointSize = angular.pointradius;
```
leeoniya, line 146: experimentally, this appears to need:

```js
graph.pointSize = 2 + angular.pointradius * 2;
```

but this falls outside the 1-10 range of the GraphNG `pointSize` slider. nevertheless, this mapping works and looks close. @torkelo should we detach `pointSize` from actual pixel diameter and find an appropriate scaling factor for a useful 1-10 range? e.g. 1px dia is pretty much useless and invisible over the thinnest available lineWidth.

public/app/plugins/panel/graph3/migrations.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  const graph = y1.custom ?? ({} as GraphFieldConfig);
+  graph.drawStyle = angular.bars ? DrawStyle.Bars : angular.lines ? DrawStyle.Line : DrawStyle.Points;
+  if (angular.points) {
+    graph.showPoints = PointVisibility.Always;
+  } else if (graph.drawStyle !== DrawStyle.Points) {
+    graph.showPoints = PointVisibility.Never;
+  }
+  if (graph.drawStyle === DrawStyle.Bars) {
+    graph.fillOpacity = 1.0; // bars were always
```
leeoniya, line 142: when i do a single-series bars style in Flot and disable lines, then switch to uPlot, it seems to reach here, but the bars opacity does not end up with 1.0, but the default 0.1. additionally, the bars have a non-zero lineWidth (i would expect this to be set to zero when migrating from flot).

public/app/plugins/panel/graph3/migrations.ts
torkelo: yea I think we should do something as the increments are too small to be noticable now.

torkelo (approved):
:100:

PR #29600, opened 2020-12-04
API: Add ID to snapshot API responses

AgnesToulet:
Enrich snapshot API to add snapshot IDs in the response (allow to have more information in audit logs).


aknuds1 (approved):
LGTM, might it make sense to include a test or two though?

aknuds1 (approved):
Thanks for tests!

PR #29603, opened 2020-12-04
Docs: Update broken aliases

marcusolsson:

Fixes a few broken aliases in docs.

These are the ones I found using the following regex (occurrences of `/docs/` that's not followed by a `g`).

```
ag "\"/docs/[^g]"
```


oddlittlebird (approved):
Nice catch!

PR #29608, opened 2020-12-04
Elastic: Fixes config UI issues

torkelo:
Fixes problem with version select and internal link toggle alignment and spacing for Add link button 

Before
Screenshot from 2020-12-04 11-33-43

After
Screenshot from 2020-12-04 11-31-43


ivanahuckova (approved):
LGTM. Just one thing, as we are fixing Elastic config UI issues, we could add a fix for `Index names` and `Pattern` going over each other even though there is enough space. There is `gf-form max-width-25` className (which is not enough for `Index names`) and `Pattern` has `width-10` and that could be less. 
image

PR #29585, opened 2020-12-03
GraphNG: Fix issues with plugins not retrieving plot instance

dprokop:
Fixes issue when plugins did not initialize because `isPlotReady` flag was removed from the context.

PR #29445, opened 2020-11-27
GraphNG: use uPlot's native ms support

leeoniya:
this updates uPlot and fixes the x-axis splits generation when feeding raw `ms`-resolution timestamps to uPlot.

it also reduces the split minimum spacing to get a similar density to Flot's.

before:

image

after:

image

PR #29613, opened 2020-12-04
GraphNG: Fix annotations and exemplars plugins

dprokop:
Followup to https://github.com/grafana/grafana/pull/29585 and https://github.com/grafana/grafana/pull/29445

Problem was related to plugins performing effects before uPlot insance was ready, resulting in early returns. Addidtionally, exemplar and annotation marks were rendered off canvas due to timestamps not being passed as ms (regression after  https://github.com/grafana/grafana/pull/29445).


packages/grafana-ui/src/components/uPlot/Plot.tsx
```diff
@@ -68,8 +70,8 @@ export const UPlotChart: React.FC<PlotProps> = props => {
 
   // Memoize plot context
   const plotCtx = useMemo(() => {
-    return buildPlotContext(canvasRef, props.data, registerPlugin, getPlotInstance);
-  }, [plotInstance, canvasRef, props.data, registerPlugin, getPlotInstance]);
+    return buildPlotContext(isPlotReady, canvasRef, props.data, registerPlugin, getPlotInstance);
```
dprokop, line 73: Brought the flag back, it makes life so much easier

PR #29615, opened 2020-12-04
Build: Publish next and latest npm channels to Github

### LINKED ISSUE #27953: Build: move canary channel/tag from npm to github registry

jackw:
As a follow up to the canary channel github package publishing to make life easier for devs we should also publish next and latest channels to github so there's no need to edit local npmrc config.

Fixes #27953

We'll continue to publish these channels to the NPM registry.

### DIFF
--- a/scripts/release.star
+++ b/scripts/release.star
@@ -37,6 +37,8 @@ def release_npm_packages_step(edition, ver_mode):
 
     if ver_mode == 'release':
         commands = ['./scripts/build/release-packages.sh ${DRONE_TAG}']
+    else:
+        commands = []
 
     return {
         'name': 'release-npm-packages',
--- a/scripts/release.star
+++ b/scripts/release.star
@@ -35,10 +35,8 @@ def release_npm_packages_step(edition, ver_mode):
     if edition == 'enterprise':
         return None
 
-    commands = [
-        './node_modules/.bin/lerna bootstrap',
-        'echo "//registry.npmjs.org/:_authToken=$${NPM_TOKEN}" >> ~/.npmrc',
-    ]
+    commands = []
+
     if ver_mode == 'release':
         commands.append('./scripts/build/release-packages.sh ${DRONE_TAG}')
 
--- a/scripts/release.star
+++ b/scripts/release.star
@@ -35,10 +35,8 @@ def release_npm_packages_step(edition, ver_mode):
     if edition == 'enterprise':
         return None
 
-    commands = []
-
     if ver_mode == 'release':
-        commands.append('./scripts/build/release-packages.sh ${DRONE_TAG}')
+        commands = ['./scripts/build/release-packages.sh ${DRONE_TAG}']
 
     return {
         'name': 'release-npm-packages',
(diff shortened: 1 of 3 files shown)


scripts/release.star
```diff
@@ -35,10 +35,8 @@ def release_npm_packages_step(edition, ver_mode):
     if edition == 'enterprise':
         return None
 
-    commands = [
-        './node_modules/.bin/lerna bootstrap',
-        'echo "//registry.npmjs.org/:_authToken=$${NPM_TOKEN}" >> ~/.npmrc',
-    ]
+    commands = []
```
jackw, line 38: I'm not sure about this. It feels strange that we have a step that releases the packages but only adds the release-packages.sh to the list of commands if `ver_mode = 'release'`. I'm likely missing something related to releases.

scripts/release.star
```diff
@@ -35,10 +35,8 @@ def release_npm_packages_step(edition, ver_mode):
...
+
     if ver_mode == 'release':
         commands.append('./scripts/build/release-packages.sh ${DRONE_TAG}')
```
dprokop, line 41:
```suggestion
    if ver_mode == 'release':
        commands = ['./scripts/build/release-packages.sh ${DRONE_TAG}']
```

dprokop (approved):
Looks great. Let's wait for @aknuds1 to respond to your comment @jack about `commands`

scripts/release.star
```diff
@@ -35,12 +35,8 @@ def release_npm_packages_step(edition, ver_mode):
     if edition == 'enterprise':
         return None
 
-    commands = [
-        './node_modules/.bin/lerna bootstrap',
-        'echo "//registry.npmjs.org/:_authToken=$${NPM_TOKEN}" >> ~/.npmrc',
-    ]
     if ver_mode == 'release':
-        commands.append('./scripts/build/release-packages.sh ${DRONE_TAG}')
+        commands = ['./scripts/build/release-packages.sh ${DRONE_TAG}']
```
aknuds1, line 39: You should define `commands`. Not sure how it currently works.
```suggestion
        commands = ['./scripts/build/release-packages.sh ${DRONE_TAG}']
    else:
        commands = []
```

aknuds1 (changes requested):
See a problem with the Starlark code.

aknuds1 (approved):

Issue #29239, opened 2020-11-19
Panel Description box does not expand to fit markdown


Mario-Hofstaetter:
**What happened**:
Panel Descriptions support Markdown formatting. If one uses monospaced **INLINE code** format to show e.g. the prometheus metric name, or _anything_ really, the panel does not adjust it's size to fit the content.
The code section overlaps the tooltip in an ugly fashion.

**What you expected to happen**:
- a) The panel should expand in width to contain its content if there is enough space
- b) If there is not enough space (Browser Windows size or whatever) or a reasonable max width is reached, it should line wrap the text.

**How to reproduce it (as minimally and precisely as possible)**:
- Edit any panel, e.g. <https://play.grafana.org/d/3SWXxreWk/grafana-dashboard?orgId=1&editPanel=4>
- Add Description text containing monospaced markdown code, e.g.:
```
This is a Description text of metric:
`windows_mssql_genstats_mars_deadlocks{instance=~"$server.*", exported_instance=~"$dbinstance.*"}`
This is text that follows the inline code. If many words are used in normal text, the lines will be wrapped.
```
- Result:
![grafik](https://user-images.githubusercontent.com/33002073/99694888-2f76c400-2a8d-11eb-9c25-33608b7082e5.png)
(1) Inline code escapes box
(2) Words are line wrapped

If the panel happens to be on the right side of the dashboard, the description may be cut off entirely:

![grafik](https://user-images.githubusercontent.com/33002073/99699009-c5145280-2a91-11eb-9c77-8c01e41381e3.png)

As a workaround I discovered using a code block with three backticks and newlines toggles line break, which looks better but is not ideal either. It would be nice to have a wider tooltip box.

![grafik](https://user-images.githubusercontent.com/33002073/99696808-651cac80-2a8f-11eb-9c84-15929995969c.png)

Result:

![grafik](https://user-images.githubusercontent.com/33002073/99697043-9f864980-2a8f-11eb-8945-56ff9742f432.png)

**Anything else we need to know?**:

- A Markdown table will also kill the layout ❌
- Embedding an URL to an image will scale the image to fit nicely ✔

**Environment**:
- Grafana version: `Grafana v7.3.2 (b8ad4eaab3)`
- Data source type & version: 
- OS Grafana is installed on:
- User OS & Browser: Windows 10 `20H2`, Firefox `82.0.3`
- Grafana plugins:
- Others:

dprokop:
[...]

PR #29628, opened 2020-12-04
fix panel header description inline code wrapping

shahzad31:
Resolves:https://github.com/grafana/grafana/issues/29239
 fixes panel header description inline code wrapping

https://github.com/grafana/grafana/issues/29239

After:
Inline code in description properly wraps into next line.
image


thisisobate (approved):
LGTM!

PR #29631, opened 2020-12-05
put the default template for AWS EC2 target region us-east-2

ying-jeanne:
**What this PR does / why we need it**: today the template EC2 for cloudwatch doesn't have a default setting for region us-east-2, the PR is to set the region for template

Fixes #

### DIFF
--- a/public/app/plugins/datasource/cloudwatch/dashboards/ec2.json
+++ b/public/app/plugins/datasource/cloudwatch/dashboards/ec2.json
@@ -1524,7 +1524,10 @@
       },
       {
         "allValue": null,
-        "current": {},
+        "current": {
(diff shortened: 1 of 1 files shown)


sunker (changes requested):
I'm afraid we can't make the assumption that all users wants us-east-2 to be the default region. However, the default region can be defined for each account in the data source config page. Can you instead investigate if it is possible to use that default region instead? 

I'm not sure, but it might also be possible to set 
`"current": {
          "text": "default",
          "value": "default"
        },` and that might be transformed into the real default value for the account.

ying-jeanne:
the solution is not correct, we want the getregion function to work instead of any hardcoded value

PR #29632, opened 2020-12-05
GraphNG: replace bizcharts with uPlot for sparklines

ryantxu:
This sketches out how we may like to support sparklines more generally

image
image
image

This adds a simple `<Sparkline .../>` component that takes two fields -- this is a breaking API change as it gets rid of `GraphSeriesValue[][];` but I think that is OK.

TODO:

- [x] fix layout issues

### DIFF
--- a/packages/grafana-data/src/dataframe/frameComparisons.ts
+++ b/packages/grafana-data/src/dataframe/frameComparisons.ts
@@ -6,7 +6,7 @@ import { DataFrame } from '../types/dataFrame';
  *
  * To compare multiple frames use:
  * ```
- * areArraysEqual(a, b, framesHaveSameStructure);
+ * compareArrayValues(a, b, framesHaveSameStructure);
  * ```
  * NOTE: this does a shallow check on the FieldConfig properties, when using the query
  * editor, this should be sufficient, however if applicaitons are mutating properties
--- a/packages/grafana-data/src/field/fieldDisplay.ts
+++ b/packages/grafana-data/src/field/fieldDisplay.ts
@@ -2,7 +2,6 @@ import toString from 'lodash/toString';
 import isEmpty from 'lodash/isEmpty';
 
 import { getDisplayProcessor } from './displayProcessor';
-import { getFlotPairs } from '../utils/flotPairs';
 import {
   DataFrame,
   DisplayValue,
@@ -13,10 +12,10 @@ import {
   FieldType,
   InterpolateFunction,
   LinkModel,
+  TimeRange,
   TimeZone,
 } from '../types';
 import { DataFrameView } from '../dataframe/DataFrameView';
-import { GraphSeriesValue } from '../types/graph';
 import { GrafanaTheme } from '../types/theme';
 import { reduceField, ReducerID } from '../transformations/fieldReducer';
 import { ScopedVars } from '../types/ScopedVars';
@@ -56,11 +55,18 @@ function getTitleTemplate(stats: string[]): string {
   return parts.join(' ');
 }
 
+export interface FieldSparkline {
+  y: Field; // Y values
+  x?: Field; // if this does not exist, use the index
+  timeRange?: TimeRange; // Optionally force an absolute time
+  highlightIndex?: number;
+}
+
 export interface FieldDisplay {
   name: string; // The field name (title is in display)
   field: FieldConfig;
   display: DisplayValue;
-  sparkline?: GraphSeriesValue[][];
+  sparkline?: FieldSparkline;
 
   // Expose to the original values for delayed inspection (DataLinks etc)
   view?: DataFrameView;
@@ -185,15 +191,6 @@ export const getFieldDisplayValues = (options: GetFieldDisplayValuesOptions): Fi
             field,
             reducers: calcs, // The stats to calculate
           });
-          let sparkline: GraphSeriesValue[][] | undefined = undefined;
-
-          // Single sparkline for every reducer
-          if (options.sparkline && timeField) {
-            sparkline = getFlotPairs({
-              xField: timeField,
-              yField: series.fields[i],
-            });
-          }
 
           for (const calc of calcs) {
             scopedVars[VAR_CALC] = { value: calc, text: calc };
@@ -202,6 +199,19 @@ export const getFieldDisplayValues = (options: GetFieldDisplayValuesOptions): Fi
               ...field.state?.scopedVars, // series and field scoped vars
               ...scopedVars,
             });
+
+            let sparkline: FieldSparkline | undefined = undefined;
+            if (options.sparkline) {
+              sparkline = {
+                y: series.fields[i],
+                x: timeField,
+              };
+              if (calc === ReducerID.last) {
+                sparkline.highlightIndex = sparkline.y.values.length - 1;
+              } else if (calc === ReducerID.first) {
+                sparkline.highlightIndex = 0;
+              }
+            }
 
             values.push({
               name: calc,
--- a/packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
+++ b/packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
@@ -77,6 +77,11 @@ export class Sparkline extends PureComponent<Props, State> {
   prepareConfig = (data: DataFrame, props: Props) => {
     const { theme } = this.props;
     const builder = new UPlotConfigBuilder();
+    builder.setCursor({
+      show: true,
+      x: false, // no crosshairs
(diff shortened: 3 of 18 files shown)


leeoniya:
and just like that, 1.4MB of the frontend bundle disappear :wastebasket:

https://bundlephobia.com/result?p=bizcharts@4.0.16

:feelsgood: :tada:

torkelo:
yea think for this we need a separate, simple sparkline component, with limited options. Main open question is if this Sparkline component uses Plot.tsx or uPlot directly

leeoniya:
> Main open question is if this Sparkline component uses Plot.tsx or uPlot directly

it's not uncommon for people to want quite a lot of the same features from a sparkline/graph-stat component as from a full-blown graph.

some time ago i pitched replacing Home-Assistant's graph card with uPlot (https://github.com/home-assistant/frontend/issues/5587). the card they currently use is the svg-based https://github.com/kalkih/mini-graph-card. if you look through their feature requests sorted by reactions, and even already-supported use-cases, you'll see that people frequently want quite a bit of graph control in these stat cards.

https://github.com/kalkih/mini-graph-card/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-heart-desc
https://github.com/kalkih/mini-graph-card/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

log scales, min/max control, softMin/softMax, multiple series with differing drawStyles and lineInterpolations.

i think it would ease maintenance significantly if the only distinction between Graph and Sparkline components was the amount of control exposed via the UI. but those who wanted the full power of the Graph in a sparkline can still dig through the JSON and set all the right options, or start with a Graph panel and then just switch the type to Sparkline in the json.

if the main difference is the current-value overlay, or some kind of a stat bar with multiple values (aggregate, avg, min/max), then just adding this as an option to the GraphNG should be pretty minimal overhead and would keep everything in one place, allowing for expansion of the Sparkline UI [as needed] to the full power of GraphNG.

/ my $0.02

leeoniya:
also relevant to this PR, is inset labels can be acheived by setting some negative `size` values and adjusting `labelAlign`:

https://github.com/leeoniya/uPlot/issues/345
https://github.com/leeoniya/uPlot/pull/347

image

ryantxu:
@leeoniya -- re full featured sparklines -- yes, but we still want to be cautious on how it is exposed. After we have feature parity with the existing chart, we could look at adding a `bars` option here:  
image

packages/grafana-data/src/dataframe/frameComparisons.ts
```diff
@@ -6,7 +6,7 @@ import { DataFrame } from '../types/dataFrame';
  *
  * To compare multiple frames use:
  * ```
- * areArraysEqual(a, b, framesHaveSameStructure);
+ * compareArrayValues(a, b, framesHaveSameStructure);
```
ryantxu, line 9: unrelated... but I noticed it while working on this PR

packages/grafana-data/src/field/fieldDisplay.ts
```diff
@@ -56,11 +55,18 @@ function getTitleTemplate(stats: string[]): string {
   return parts.join(' ');
 }
 
+export interface FieldSparkline {
+  y: Field; // Y values
```
ryantxu, line 59: or maybe "value" -- but that leads to to having `value.value`!

packages/grafana-ui/src/components/uPlot/types.ts
```diff
@@ -3,7 +3,7 @@ import uPlot, { Options, Series, Hooks } from 'uplot';
 import { DataFrame, TimeRange, TimeZone } from '@grafana/data';
 import { UPlotConfigBuilder } from './config/UPlotConfigBuilder';
 
-export type PlotSeriesConfig = Pick<Options, 'series' | 'scales' | 'axes'>;
+export type PlotSeriesConfig = Pick<Options, 'series' | 'scales' | 'axes' | 'cursor'>;
```
ryantxu, line 6: @dprokop -- should this be renamed `PlotConfig` -- it will eventually be everything the `UPlotConfigBuilder` knows how to build

packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
```diff
@@ -0,0 +1,166 @@
...
+  configBuilder: UPlotConfigBuilder;
+}
+
+const defaultConfig: GraphFieldConfig = {
+  drawStyle: DrawStyle.Line,
+  showPoints: PointVisibility.Auto,
+  axisPlacement: AxisPlacement.Hidden,
+};
+
+export class Sparkline extends PureComponent<Props, State> {
```
ryantxu, line 35: This class is similar to GraphNG -- but it knows that the input is only 1 or 2 fields

dprokop: Great

packages/grafana-ui/src/components/uPlot/types.ts
dprokop: Yes, we can rename it!

packages/grafana-data/src/dataframe/frameComparisons.ts
dprokop: 👍

packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
```diff
@@ -0,0 +1,171 @@
...
+  render() {
+    const { data, configBuilder } = this.state;
+    const { width, height, sparkline } = this.props;
+
+    return (
+      <UPlotChart
+        data={{
+          frame: data,
+          isGap: () => true, // any null is a gap
+        }}
```
torkelo, line 162: this is not ideal, creating a new data object here will make the Plot think the data always has updated. But think it's fine also as this will only be in-efficient when resizing the sparkline (only other reason for the sparkline to re-render in cases when data not changed is resize as far as I can see).  So this is not a problem really.

torkelo (approved):
Awesome job! this works great. Expected to have to spend a lot of time fixing/tweaking/refactoring this but everything just works and looks good! 

Would be nice with storybook for the new Sparkline component but that can be added later. 

I did find one minor issue and that is a random point, in the beginning, is highlighted for me. And setting reducer has no effect on what point is highlighted.  Looking at plot config I cannot see how we can control what point is highlighted from outside.  

Screenshot from 2020-12-08 10-12-55

But think we can fix that later. This looks mergeable to mee. Did fix a minor bug in the code where prepareData function in sparkline took in props but did not use it (used this.props instead).

packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
```diff
@@ -0,0 +1,166 @@
...
+        {
+          ...sparkline.y,
+          config: yFieldConfig,
+        },
+      ],
+      length,
+    };
+  };
+
+  prepareConfig = (data: DataFrame, props: Props) => {
```
dprokop: I have slight doubts about Sparkline implementing the entire config preparation. What about constructing the entire dataframe in the Sparkline, and only passing it to GraphNG component so that it can handle the config creation?

packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const { theme } = this.props;
+    const builder = new UPlotConfigBuilder();
+    builder.setCursor({
+      show: true,
```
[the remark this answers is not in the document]
torkelo: think it's nice for the hover effect, but it shows even when not hovering. Think the plan is to (later) get the value to show the hovered value as well

packages/grafana-ui/src/components/Sparkline/Sparkline.tsx
torkelo: but this is not using GraphNG (and I do not think it should)

torkelo: hm.. now I look again, see your point a bit more. There is a lot of code here to build the config that is similar to what is in GraphNG

ryantxu: I agree we we should avoid duplicated code -- but in this case everything could/should be in the UPlotConfigBuilder.  I think the distinction here should be:

* GraphNG
  * DataFrame[] >>> uplot configs, uplot data, legend
* GraphNG
  * Sparkline( x, y field ) >>> uplot configs, uplot data

PR #29635, opened 2020-12-05
GraphNG: rename "points" to "showPoints"

leeoniya:
i think this nomenclature is more clear (similar to the "display" -> "drawStyle" rename)

### DIFF
--- a/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
+++ b/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
@@ -12,7 +12,7 @@ import {
 import { alignDataFrames } from './utils';
 import { UPlotChart } from '../uPlot/Plot';
 import { PlotProps } from '../uPlot/types';
-import { AxisPlacement, GraphFieldConfig, DrawStyle, ShowPoints } from '../uPlot/config';
+import { AxisPlacement, GraphFieldConfig, DrawStyle, PointVisibility } from '../uPlot/config';
 import { useTheme } from '../../themes';
 import { VizLayout } from '../VizLayout/VizLayout';
 import { LegendDisplayMode, LegendItem, LegendOptions } from '../Legend/Legend';
@@ -35,7 +35,7 @@ export interface GraphNGProps extends Omit<PlotProps, 'data' | 'config'> {
 
 const defaultConfig: GraphFieldConfig = {
   drawStyle: DrawStyle.Line,
-  showPoints: ShowPoints.Auto,
+  showPoints: PointVisibility.Auto,
   axisPlacement: AxisPlacement.Auto,
 };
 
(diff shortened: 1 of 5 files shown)


packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -35,7 +35,7 @@ export interface GraphNGProps extends Omit<PlotProps, 'data' | 'config'> {
 
 const defaultConfig: GraphFieldConfig = {
   drawStyle: DrawStyle.Line,
-  points: PointMode.Auto,
+  showPoints: ShowPoints.Auto,
```
torkelo, line 38: points might be bad, but feel `showPoints` is not much better, as it signifies a verb/action, not an option. 

* linePoints  (LinePointsMode for enum) 
* linePointMode (LinePointsMode for enum)

leeoniya: the problem with a "line" prefix is that there's not a line rendered. points can also be shown on bars (or lizards). i was thinking about using a single "pointDensity" which would map to uPlot's `points.space`. this gives the most control, where:

```js
null | "" -> "Auto"
0 -> "Always"
-1 -> "Never"
100 -> 100px
```

the UI can still hide the density aspect unless we ever want to expose it.

leeoniya: i the current UI setup, "when/if points are shown" is exactly what this setting semantically does. "mode" is ambiguous, and "line" is orthogonal. i see you point, though. the naming makes more sense when viewed only through the JSON output, and is not as straightforward if you have to use it in the code.

leeoniya: "pointVisibility" - a bit wordy, but might work better.

leeoniya: the old graph uses `showPoints`, which i think is pretty clear. 

image

would you be okay with changing the enum name to `PointVisibility` but leaving the option as `showPoints`?

torkelo: Yea, it works ok. Reason I suggested linePoints is that it is only enabled/relevant when line drawStyle is set

leeoniya: ok, cool. i think it's more flexible to leave it unprefixed and omit "line". especially since we have a dedicated "Points" mode where all the line options are disabled.

torkelo: But is not an argument for the line prefix as it's only relevant for line? But I guess it be relevant in other future non point drawStyle options :)

leeoniya: i see what you did there; touché, good sir!

leeoniya:
i think this is good to go now.

dprokop (approved):
Looks good, great discussion:)

PR #29637, opened 2020-12-06
Explore: Fix a bug where Typeahead crashes when a large amount of ite…

### LINKED ISSUE #27631: Large amount of time series with a common prefixes cause "RangeError: Maximum call stack size exceeded"
The Gif

**What happened**:

As part of the performance testing of my Thanos install, I am using the [avalanche](https://github.com/open-fresh/avalanche) utility to stress test labels and time series. While running queries against the dataset, the grafana instance fails very ungracefully when doing the type ahead for the "avalanche..." metrics. In this case all the time series generated have the "avalanche" prefix.

Grafana should not crash, and/or disable type ahead for large amounts of metrics.

1. Deploy prometheus & grafana, scrape metrics from the avalanche container, or something else that generates large amounts of time series (was observed for over 10,000 time series). 
2. Type the letter 'a' in the explore view

**Environment**:
- Grafana version: 7.1.5
- Data source type & version: prometheus
- OS Grafana is installed on: k8s, grafana/grafana:7.1.5 docker image
- User OS & Browser: Chrome & OSX
- Grafana plugins: N/A
- Others: N/A

sutaka:
Because calling push API with array spreading will push all arguments to the stack,  it could cause a stack overflow if the input size is too large.
Reference: 
https://github.com/nodejs/node/issues/27732
`Array.prototype.push is not supposed to be called with a large number of arguments.`

This PR changes to use concat API instead of push API  in order to handle a large number of items.

This will resolve #27631

#27631

Before fix
before
After
after

### DIFF
--- a/packages/grafana-ui/src/components/Typeahead/Typeahead.test.tsx
+++ b/packages/grafana-ui/src/components/Typeahead/Typeahead.test.tsx
@@ -3,7 +3,7 @@ import { mount } from 'enzyme';
 
 import { Typeahead, State } from './Typeahead';
 import { TypeaheadItem } from './TypeaheadItem';
-import { CompletionItemKind } from '../../types';
+import {CompletionItemGroup, CompletionItemKind} from '../../types';
 
 describe('Typeahead', () => {
   const completionItemGroups = [{ label: 'my group', items: [{ label: 'first item' }] }];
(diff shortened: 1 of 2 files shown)


mckn (commented):
Really awesome that you added tests for this and open a PR with a solution. However, if we are working with large arrays I think it would be better to change the `flattenGroupItems` to something like this:

```ts
export const flattenGroupItems = (groupedItems: CompletionItemGroup[]): CompletionItem[] => {
  return groupedItems.reduce((all: CompletionItem[], { items, label }) => {
    all.push({
      label,
      kind: CompletionItemKind.GroupTitle,
    });

    return items.reduce((all, item) => {
      all.push(item);
      return all;
    }, all);
  }, []);
};
```

The difference between this one and the `.concat` is that this won't copy the array but instead just only append the items to the existing one which should be beneficial for large arrays.

sutaka:
Thank you for reviewing and the suggestion! 
+1 to your idea and I'll make a change🙂

mckn (approved):
LGTM!

great work

dprokop (approved):

PR #29640, opened 2020-12-07
Rendering: Allow overriding callback URL for direct image rendering

### LINKED ISSUE #20725: Direct image rendering doesn't work when using UNIX socket protocol
**What happened**:
When grafana is configured to use UNIX socket protocols instead of HTTP the direct image rendering doesn't work.

Direct image rendering works.

Reproduce steps:
- configure grafana to use UNIX socket instead of HTTP protocol
```[server]
# Protocol (http, https, h2, socket)
protocol = socket
socket = /run/grafana/socket
```
- use socat to listen on TCP socket and relay your request to the UNIX socket
```
socat  TCP4-LISTEN:3000,fork UNIX-CONNECT:/run/grafana/socket
```
- request a direct image for any graph
- if you change the protocol to HTTP the images are rendered

grafana debug log:
```
t=2019-11-28T07:38:29+0000 lvl=info msg=Rendering logger=rendering path="d-solo/GZNijPbZk/new-dashboard-copy?orgId=1&from=1574905107588&to=1574926707588&panelId=2&width=1000&height=500&tz=Europe%2FStockholm"
t=2019-11-28T07:39:31+0000 lvl=info msg="Rendering timed out" logger=rendering
t=2019-11-28T07:39:31+0000 lvl=eror msg="Timeout error. You can set timeout in seconds with &timeout url parameter" logger=context userId=1 orgId=1 uname=admin error="Timeout error. You can set timeout in seconds with &timeout url parameter"
t=2019-11-28T07:39:31+0000 lvl=eror msg="Request Completed" logger=context userId=1 orgId=1 uname=admin method=GET path=/render/d-solo/GZNijPbZk/new-dashboard-copy status=500 remote_addr=@ time_ms=62010 size=1782 referer="http://localhost:3000/d/GZNijPbZk/new-dashboard-copy?orgId=1"
```

**Environment**:
- Grafana version: 6.5.0
- Data source type & version:
- OS Grafana is installed on: Ubuntu 18.04.3
- User OS & Browser: Linux, Chrome
- Grafana plugins: none
- Others:

whitequark:

This is necessary for direct image rendering to work when Grafana is configured to use UNIX socket protocol. A workaround suggested in https://github.com/grafana/grafana/issues/20725#issuecomment-614059117 was to use `[rendering].callback_url`, but it would not actually work because `callback_url` is ignored for direct image rendering.

Fixes #20725 (by making a workaround possible).

### DIFF
--- a/pkg/services/rendering/rendering.go
+++ b/pkg/services/rendering/rendering.go
@@ -217,7 +217,7 @@ func (rs *RenderingService) getFilePathForNewImage() (string, error) {
 }
 
 func (rs *RenderingService) getURL(path string) string {
-	if rs.Cfg.RendererUrl != "" {
+	if rs.Cfg.RendererCallbackUrl != "" {
 		// The backend rendering service can potentially be remote.
 		// So we need to use the root_url to ensure the rendering service
 		// can reach this Grafana instance.


whitequark:
For some reason I can't agree to the CLA, it just shows me this:
Screenshot_20201207_003005

whitequark:
Ping for review.

whitequark:
Ping for review.

pkg/services/rendering/rendering.go
```diff
@@ -217,7 +217,7 @@ func (rs *RenderingService) getFilePathForNewImage() (string, error) {
 }
 
 func (rs *RenderingService) getURL(path string) string {
-	if rs.Cfg.RendererUrl != "" {
+	if rs.Cfg.RendererCallbackUrl != "" {
```
torkelo, line 220: I do not think this change looks correct. RendererCallbackUrl is always set to something so this will break the existing behavior

torkelo (changes requested):
Think this will break other behaviors

pkg/services/rendering/rendering.go
whitequark: Thanks, I didn't realize that. I'll have to reconsider my approach.

PR #29642, opened 2020-12-07
PanelLibrary: Adds api and db to create Library/Shared/Reusable Panel

hugohaggmark:
Initial backend structure for Library Panels, before adding more functions/endpoints I'd like to get some early feedback if this has the correct structure & tests.

Relates #29610


torkelo (commented):
think this is fine and matches how other services/features are being implemented. Especially in Enterprise. 

But I am a bit uncomfortable mixing so many different details/concerns in a single service/abstractions. It somehow feels too simple :) The alternative that I would like to explore for the new alerting service and other services like this, is to separate the database into it's own Repository abstraction, and the HTTP api to its own abstraction (that depends on the repository). Not sure what the main service does in this case though. Maybe the HTTP API abstraction is not needed and can be done from main service? Not sure if these sub services needs to be normal backend services that self register but maybe just sub services that the main LibraryPanelService create from it's injected dependencies (up for debate). 

Think I will invite to a backend architecture meeting or create a google doc to discuss alternatives.

hugohaggmark:
> think this is fine and matches how other services/features are being implemented. Especially in Enterprise.
> 
> But I am a bit uncomfortable mixing so many different details/concerns in a single service/abstractions. It somehow feels too simple :) The alternative that I would like to explore for the new alerting service and other services like this, is to separate the database into it's own Repository abstraction, and the HTTP api to its own abstraction (that depends on the repository). Not sure what the main service does in this case though. Maybe the HTTP API abstraction is not needed and can be done from main service? Not sure if these sub services needs to be normal backend services that self register but maybe just sub services that the main LibraryPanelService create from it's injected dependencies (up for debate).
> 
> Think I will invite to a backend architecture meeting or create a google doc to discuss alternatives.

Sounds great, I also found that having a repository abstraction could be helpful. Didn't think about the api abstraction though, would need more information about that before I can say anything about it.  That meeting would be great, maybe on the next backend weekly?

papagian:
> Sounds great, I also found that having a repository abstraction could be helpful. Didn't think about the api abstraction though, would need more information about that before I can say anything about it. That meeting would be great, maybe on the next backend weekly?

👍  backend weekly is today!

pkg/services/librarypanels/database.go
```diff
@@ -0,0 +1,42 @@
+package librarypanels
+
+import (
+	"context"
+	"time"
+
+	"github.com/grafana/grafana/pkg/services/sqlstore"
+)
+
+// AddLibraryPanel function adds a LibraryPanel
```
aknuds1, line 10:
```suggestion
// addLibraryPanel function adds a LibraryPanel
```

pkg/services/librarypanels/database_test.go
```diff
@@ -0,0 +1,82 @@
...
+      "type": "text"
+    }
+`),
+		}
+
+		noErr := lps.addLibraryPanel(&command)
+		require.NoError(t, noErr)
+
+		err := lps.addLibraryPanel(&command)
+		require.Error(t, err)
```
aknuds1, line 36: Probably want `require.EqualError` here, to be precise about the expected error.

aknuds1 (commented):
Just a super quick review.

Can you loop me in on how you foresee using the repository pattern? To me this looks relatively clean so far (keep in mind though that I only skimmed very quickly!), and I'm wary of potential over-engineering.

I don't know how you intend to use the repository pattern, or the benefits you foresee, but my own experiences with that pattern have been negative. Repositories have typically been leaky abstractions, and I don't personally see any point in using them instead of simply the service pattern (like SQLStore). All depends on the concrete design you have in mind though.

pkg/services/librarypanels/database_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/database.go

hugohaggmark:
> Just a super quick review.
> 
> Can you loop me in on how you foresee using the repository pattern? To me this looks relatively clean so far (keep in mind though that I only skimmed very quickly!), and I'm wary of potential over-engineering.
> 
> I don't know how you intend to use the repository pattern, or the benefits you foresee, but my own experiences with that pattern have been negative. Repositories have typically been leaky abstractions, and I don't personally see any point in using them instead of simply the service pattern (like SQLStore). All depends on the concrete design you have in mind though.

I just commited a first variant using repository pattern instead. Let me know what you think @torkelo @aknuds1

aknuds1 (commented):
I actually think this is overengineering. 

I don't see any reason to introduce an interface and no responsibility is actually removed from LibraryPanelService, its implementation just has more structure. In my opinion it's over-structured, and you don't achieve a reduction of the number of the service's responsibilities.

I would instead take a top-down approach to the design and concentrate on making LibraryPanelService as simple as possible. If it has too many responsibilities, the solution should be to divide it into several services so it (so the actual number of responsibilities per service diminishes).

torkelo:
> I don't know how you intend to use the repository pattern, or the benefits you foresee, but my own experiences with that pattern have been negative. Repositories have typically been leaky abstractions, and I don't personally see any point in using them instead of simply the service pattern (like SQLStore). All depends on the concrete design you have in mind though.

Well, we should not mix the database logic inside the HTTP API layer in my view. And since we have decided to remove the Command / Query bus abstraction we need a replacement abstraction for database access. The repository pattern is the only one I know of that would fit and is something we use in other places instead of the bus/sqlite store (See Annotation repository for example). But I agree repository pattern has it's pros/cons. Mainly that they can grow and cover too many use cases, and all use cases must share a single interface. This is why I prefer the command/query pattern where different operations do not need to share an interface, (so mocking you only need to mock the operation you are using etc). 

Started a design doc for Grafana Backend service patterns https://docs.google.com/document/d/15m7oHeUkMkTTiM6v336JTWLcEMM2Q1K879vNhyNykJc/edit# 

want to get some more thoughts down and then discuss

aknuds1:
@torkelo The SQLStore is already a database abstraction though? Why do we need another abstraction on top?

aknuds1:
Maybe we have different design philosophies. I believe more in componentized than layered design, i.e. "ravioli" versus "lasagne". I don't think layered designs work very well in practice.

torkelo:
@aknuds1 if used via the bus then it's an abstraction (well the bus is in this case). But if used directly, there is no interface for SQLStore, nothing domain specific there that can be seen as a domain abstraction, it is database library abstraction, not a data access abstraction (the way it's written now)

aknuds1:
@torkelo It is an abstraction when not being used via the bus, I don't understand your argument to be honest.

I have to admit I don't care so much about domain abstractions either, I think in purely technical terms and find that works the best.

aknuds1:
Instead of introducing the repository pattern, I think we should identify eventual problems with SQLStore first, and then figure out how to solve them.

torkelo:
>It is an abstraction when not being used via the bus, I don't understand your argument to be honest.

Can you explain how SQLStore is an abstraction for data access? I do not see how it, it's a SQL libray abstraction at most but it leaks that as it exposes the xorm library details/interfaces. I do not understand what you mean. 

  > I have to admit I don't care so much about domain abstractions either.

You do not care about domain abstractions? that is all code is to me, thinking about the domain and how to structure it, name it, what parts are responsible for what etc. Not layers per say but areas of responsibility and detail

aknuds1:
> Can you explain how SQLStore is an abstraction for data access? I do not see how it, it's a SQL libray abstraction at most but it leaks that as it exposes the xorm library details/interfaces. I do not understand what you mean.k

It's a code abstraction, in that it hides the actual code as opposed to inlining.

I see now what you want, you want a repository so you can hide the xorm details. That is actually the philosophy I don't agree with. From my experience, the leakiness and pain of the repository pattern doesn't make hiding database implementation details worth it. I think it's better to rewrite so we get rid of xorm, which is something we're already talking about. The temporary pain of the rewrite is better than the pain of living with the leaky repository pattern.

> > I have to admit I don't care so much about domain abstractions either.
> 
> You do not care about domain abstractions? that is all code is to me, thinking about the domain and how to structure it, name it, what parts are responsible for what etc. Not layers per say but areas of responsibility and detail

Yeah, I have experience with DDD in projects, and I don't think it's useful. I think what actually works is thinking in terms of functionality/features and having a solid test suite to enforce them.

torkelo:
> Instead of introducing the repository pattern, I think we should identify eventual problems with SQLStore first, and then figure out how to solve them. It's probably not that difficult.

Many enterprise services and new Grafana core services that do not use the bus and just issue SQL queries from anywhere directly from the api/domain logic. Meaning they do not use SQLStore as abstraction, just as a way to get hold of the database lib to make db operations. Making that part of the code impossible to unit test. 
For example: https://github.com/grafana/grafana/blob/master/pkg/services/ngalert/ngalert.go#L17 

Mixes so many things in one service (api, db migrations, db operations, etc)

aknuds1:
> > Instead of introducing the repository pattern, I think we should identify eventual problems with SQLStore first, and then figure out how to solve them. It's probably not that difficult.
> 
> Many enterprise services and new Grafana core services that do not use the bus and just issue SQL queries from anywhere directly from the api/domain logic. Meaning they do not use SQLStore as abstraction, just as a way to get hold of the database lib to make db operations. Making that part of the code impossible to unit test.
> For example: https://github.com/grafana/grafana/blob/master/pkg/services/ngalert/ngalert.go#L17
> 
> Mixes so many things in one service (api, db migrations, db operations, etc)

I don't agree it means that SQLStore isn't an abstraction. I think there are two alternativesavenues when it comes to unit testing `AlertNG` f.ex.:

* Have it depend on a higher level abstraction than SQLStore (an interface that can be faked?)
* Fake the operations carried out by SQLStore

torkelo:
@aknuds1  actually I do not like the repository pattern that much either, prefer command / query seperation. solves all the problems of the repository pattern for me. 

> I see now what you want, you want a repository so you can hide the xorm details. 

Well I want to hide it from the parts of the codebase that should not need to know the details of it, so I can test those parts in isolation. Same reason we create any form of abstraction. To help make a code base more maintainable, same we reason we split a code base up into different services, files, etc.

aknuds1:
> Well I want to hide it from the parts of the codebase that should not need to know the details of it, so I can test those parts in isolation. Same reason we create any form of abstraction. To help make a code base more maintainable, same we reason we split a code base up into different services, files, etc.

You have to weigh pros against cons. I have such bad experience with the repository pattern that I would warn strongly against it. That's why I said we should consider the concrete problems, and think of how to solve them. For example, I already listed the following potential avenues for testing `AlertNG`, both of them make use of component based design instead of layers:

* Have it depend on a higher level abstraction than SQLStore (an interface that can be faked?)
* Fake the operations carried out by SQLStore

torkelo:
> both of them make use of component based design instead of layers:

What are the components you propose? I find calling it components instead of layers 

Right now everything is in one component doing everything

torkelo:
> Have it depend on a higher level abstraction than SQLStore (an interface that can be faked?)

Would this not be the same as a repostitory?

torkelo:
> Fake the operations carried out by SQLStore

How would you fake this?

``
return ng.SQLStore.WithTransactionalDbSession(context.Background(), func(sess *sqlstore.DBSession) error {
		res, err := sess.Exec("DELETE FROM alert_definition WHERE id = ?", query.ID)
		if err != nil {
			return err
		}

		rowsAffected, err := res.RowsAffected()
		if err != nil {
			return err
		}
		query.RowsAffected = rowsAffected
		return nil
	})
```

**Comment by aknuds1** [comment]
> > both of them make use of component based design instead of layers:
> 
> What are the components you propose? I am fine calling it components instead of layers

I would have to flesh out the functionality in this PR before identifying useful components. The one useful component I see so far is LibraryPanelService. After fleshing out all the functionality, we could identify if it's too fat (i.e. too many responsibilities).

> Right now everything is in one component doing everything

Adding sub-components to it won't reduce its surface area though, i.e. it will still have the same responsibilities.

**Comment by aknuds1** [comment]
> > Have it depend on a higher level abstraction than SQLStore (an interface that can be faked?)
> 
> Would this not be the same as a repostitory?

It would follow general component/service based thinking, rather than the more specialized repository pattern which tries to abstract away the database in a general/layered manner.

**Comment by aknuds1** [comment]
> > Fake the operations carried out by SQLStore
> 
> How would you fake this?

I'd have to try in practice, but it looks as if we could use our typical pattern of stubbing the xorm client, and have it send a (generated) mocked session.

**Comment by torkelo** [comment]
> It would follow general component/service based thinking, rather than the more specialized repository pattern which tries to abstract away the database in a general/layered manner. 

Not sure I follow the difference here, sounds like your proposing adding every DB operation to one big interface? that sounds terrible. But that can't be what your proposing. So interested to know what you mean by component/servcie based thinking. This is what I am trying to propose here that we split up components to have different responsibilities (DB , API, background processing, etc). And I am fine with small services maybe won't need this. But for large service like alerting and others I feel having everything in one big messy service is back to the old Transaction script architecture days.

**Comment by aknuds1** [comment]
> > It would follow general component/service based thinking, rather than the more specialized repository pattern which tries to abstract away the database in a general/layered manner.
> 
> Not sure I follow the difference here, sounds like your proposing adding every DB operation to one big interface? that sounds terrible. But that can't be what your proposing. So interested to know what you mean by component/servcie based thinking. This is what I am trying to propose here that we split up components to have different responsibilities (DB , API, background processing, etc). And I am fine with small services maybe won't need this. But for large service like alerting and others I feel having everything in one big messy service is back to the old Transaction script architecture days.

I'm not suggesting everything to one interface, no. I was actually mentioning to maybe use smaller interfaces, depending on which problems you're trying to solve.

File: pkg/services/librarypanels/database_test.go [hunk 3]
```diff
@@ -0,0 +1,82 @@
...
+			InitPriority: 0,
+		}
+
+		return &descriptor, true
+	}
+
+	registry.RegisterOverride(overrideServiceFunc)
+
+	sqlStore := sqlstore.InitTestDB(t)
+	lps.SQLStore = sqlStore
```
aknuds1, line 62: I would set this on initialization of `lps`, I don't see any benefit to delay setting the field. You just risk problems in between while the field is nil.

pkg/services/librarypanels/database_test.go
aknuds1: It's still `Error` and not `EqualError`.

pkg/services/librarypanels/database_test.go
```diff
@@ -0,0 +1,82 @@
...
+		descriptor := registry.Descriptor{
+			Name:         "LibraryPanelService",
+			Instance:     &lps,
+			InitPriority: 0,
+		}
+
+		return &descriptor, true
+	}
+
+	registry.RegisterOverride(overrideServiceFunc)
```
aknuds1, line 59: Why do you need to register the service? Don't see the need, and it's also not cleaned up after the test.

pkg/services/librarypanels/models.go
```diff
@@ -0,0 +1,42 @@
...
+	// errLibraryPanelAlreadyAdded is an error when you add a library panel that already exists.
+	errLibraryPanelAlreadyAdded = fmt.Errorf("library panel with that title already exists")
+)
+
+// Commands
+
+// addLibraryPanelCommand is the command for adding a LibraryPanel
+type addLibraryPanelCommand struct {
+	OrgId        int64                `json:"-"`
+	FolderId     int64                `json:"folderId"`
```
aknuds1, line 36: [Initialisms are written with consistent case in Go](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
```suggestion
	OrgID        int64                `json:"-"`
	FolderID     int64                `json:"folderId"`
```

papagian: That way (if it's `OrgID` instead of `OrgId`) xorm by default will assume that the respective column is `org_i_d` instead of `org_id` therefore xorm tags should be added for alerting the default behaviour:
```suggestion
	OrgId        int64                `json:"-"` `xorm:"org_id"`
	FolderId     int64                `json:"folderId"` `xorm:"folder_id"`
```

pkg/services/librarypanels/database_test.go
hugohaggmark: Otherwise the feature toggle will not be found and the library_panel will not exist and the test will fail.

hugohaggmark: Good point on cleaning up, any suggestions on how?

pkg/services/librarypanels/models.go
hugohaggmark: Thank you I'll look into this.

aknuds1: @papagian This isn't a database model type though, right?

aknuds1: Could you comment on the `LibraryPanel` type instead? It should apply there.

hugohaggmark: Yes it is

papagian: @aknuds1 is right is not used directly by xorm (`LibraryPanel` [is passed](https://github.com/grafana/grafana/blob/a16134544962de0cb7924185bad5fdd29c36efd7/pkg/services/librarypanels/database.go#L35) to Insert) but it's general good to keep this in mind since the xorm docs are not that clear about it.

hugohaggmark: So when do you know if something is used directly by xorm? Newb question 🙈

pkg/services/librarypanels/database_test.go
aknuds1: I see now why it's needed, because the service adds a migration. I don't know why we haven't seen this pattern in tests before. The `registry` package currently lacks a way to de-register a service, I think you'll have to add a function that removes the override, and call it via `t.Cleanup`.

pkg/services/librarypanels/models.go
papagian: I'm not good at definitions but I'll risk saying if it's a struct mapping a database table.

pkg/services/librarypanels/models.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"time"
+
+	"github.com/grafana/grafana/pkg/models"
+)
+
+// LibraryPanel is the model for library panel definitions.
+type LibraryPanel struct {
+	Id       int64
+	OrgId    int64
```
papagian, line 14: Following [this](https://github.com/grafana/grafana/pull/29642/files#r538102765) discussion if we want to conform with GO's case for initialisms this can become:
```suggestion
	ID        int64 `xorm:"pk autoincr 'id'"`
	OrgID     int64 `xorm:"org_id"`
```

pkg/services/librarypanels/models.go
aknuds1: @hugohaggmark You can only tell by seeing if the type is used for xorm queries or insertions.

@papagian There's no point in adding xorm annotations to a type that isn't used as a xorm model (i.e., used for queries or insertions). I don't think xorm docs are needed to clarify here, since there's just no way for those fields to be interpreted by xorm.

papagian: @aknuds1 yes, my bad I had just confused `LibraryPanel ` and `addLibraryPanelCommand `.

aknuds1: @papagian Aha thanks for clarifying! I agree that xorm's docs aren't very clear about field naming directives.

pkg/services/librarypanels/models.go
hugohaggmark: Thank you Sofia, I'll look into this

pkg/services/librarypanels/api.go
```diff
@@ -0,0 +1,31 @@
...
+// createHandler handles POST /api/library-panels.
+func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd AddLibraryPanelCommand) api.Response {
+	panel, err := lps.Create(c, cmd)
+
+	if err != nil {
+		return api.Error(500, "Failed to create library panel", err)
+	}
+
+	return api.JSON(200, util.DynMap{"panel": panel})
+}
```
hugohaggmark, line 31: I added __`createHandler`__ because I wanted to separate the Create to a service level interface without the api response and I couldn't find a good way to test the resulting response status. Any suggestions?

pkg/services/librarypanels/database_test.go
hugohaggmark: Fixed :)

pkg/services/librarypanels/database_test.go
hugohaggmark: I added ClearOverrides function to registry and cleanup to test.

pkg/services/librarypanels/models.go
hugohaggmark: I've fixed this in my latest commit (I hope)

pkg/services/librarypanels/models.go
hugohaggmark: I've fixed this in my latest commit (I hope)

pkg/services/librarypanels/database_test.go
hugohaggmark: Not sure I understand how that is going to work. I need to override the LibraryPanelService in the registry because I want the AddMigration to be called when initializing the SqlStore. Do you mean I should call `sqlstore.InitTestDB(t)` twice?

pkg/services/librarypanels/librarypanels.go
```diff
@@ -63,3 +64,15 @@ func (lps *LibraryPanelService) AddMigration(mg *migrator.Migrator) {
 
 	mg.AddMigration("create library_panel table v1", migrator.NewAddTableMigration(libraryPanelV1))
 }
+
+// Create function stores a Library Panel
```
aknuds1, line 68:
```suggestion
// Create stores a Library Panel.
```

pkg/services/librarypanels/api.go
```diff
@@ -15,18 +15,17 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
...
 	})
 }
 
-// addLibraryPanelEndpoint handles POST /api/library-panels.
-func (lps *LibraryPanelService) addLibraryPanelEndpoint(c *models.ReqContext, cmd addLibraryPanelCommand) api.Response {
-	cmd.OrgId = c.SignedInUser.OrgId
-	cmd.SignedInUser = c.SignedInUser
+// createHandler handles POST /api/library-panels.
+func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd AddLibraryPanelCommand) api.Response {
+	panel, err := lps.Create(c, cmd)
```
aknuds1, line 24: I don't see the point of this change. Why add another indirection with more code? Is there any upside? The downside is that it's more difficult to review and verify the code. The benefits of an abstraction should outweigh its cost (every abstraction has a cost).

I think the `Result` field should be removed from the command though. IMO the command should only contain parameters from the HTTP client, and a normal return value should be used to pass the result back.

pkg/services/librarypanels/api.go
aknuds1: What's the actual problem you're trying to solve @hugohaggmark, that you can't test the response in the test?

hugohaggmark: I thought the LibraryPanelService interface should be separate from the Api, so if other services consume this service it would become clearer, but maybe there is a lot I don't understand with the backend architecture (yet). 

How do services interact with each other? This service will need to interact with both dashboards, tags and folders moving forward.

pkg/services/librarypanels/librarypanels.go
hugohaggmark: Will fix

pkg/services/librarypanels/api.go
hugohaggmark: I totally agree about the command, that was something I copied from `ngalert` as I thought that was a good example, I will change that too. As for the point of the change, I wrote in the other comment.

hugohaggmark: https://github.com/grafana/grafana/pull/29642#discussion_r538283212

torkelo: @aknuds1 we are going to have a provisioning workflow that will also create panels, so we need a function that create a library panels that is shared between api and internal usage

pkg/services/librarypanels/api.go
torkelo: Think separating DB create action into a separate service function seems like a very good pattern to me.

pkg/services/librarypanels/api.go
hugohaggmark: Torkel I've had a chat with Arve and I'll add it when we need it instead. (YAGNI)

aknuds1: Yeah, it's no bother to add a method when it's needed, it's more difficult to envision the right design beforehand.

pkg/services/librarypanels/database.go
```diff
@@ -0,0 +1,45 @@
...
+			return err
+		}
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return libraryPanel, nil
```
aknuds1, line 44:
```suggestion

	return libraryPanel, err
```

pkg/services/librarypanels/database.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	err := lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		if res, err := session.Query("SELECT 1 from library_panel WHERE org_id=? and folder_id=? and title=?", c.SignedInUser.OrgId, cmd.FolderID, cmd.Title); err != nil {
+			return err
+		} else if len(res) == 1 {
+			return errLibraryPanelAlreadyAdded
+		}
+
+		if _, err := session.Insert(libraryPanel); err != nil {
```
aknuds1, line 34:
```suggestion
		if _, err := session.Insert(&libraryPanel); err != nil {
```

pkg/services/librarypanels/librarypanels.go
```diff
@@ -47,7 +53,7 @@ func (pl *LibraryPanelService) AddMigration(mg *migrator.Migrator) {
 			{Name: "org_id", Type: migrator.DB_BigInt, Nullable: false},
 			{Name: "folder_id", Type: migrator.DB_BigInt, Nullable: false},
 			{Name: "title", Type: migrator.DB_NVarchar, Length: 255, Nullable: false},
-			{Name: "data", Type: migrator.DB_Text, Nullable: false},
+			{Name: "model", Type: migrator.DB_Text, Nullable: false},
```
aknuds1, line 56: Won't renaming this column in an existing migration cause a bug?

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -0,0 +1,100 @@
...
+      "type": "text"
+    }
+`),
+		}
+
+		response := lps.createHandler(&context, command)
+		require.Equal(t, 200, response.Status())
+
+		response = lps.createHandler(&context, command)
+		require.Equal(t, 500, response.Status())
```
aknuds1, line 35:
```suggestion
		require.Equal(t, 400, response.Status())
```

aknuds1 (changes requested):
I spotted an API bug (400 vs 500), and also wondering about the change to an existing migration?

pkg/services/librarypanels/librarypanels.go
aknuds1: @papagian Do you know if changing this migration's column will cause a bug?

pkg/services/librarypanels/database.go
```diff
@@ -4,39 +4,42 @@ import (
...
-
-			Created: time.Now(),
-			Updated: time.Now(),
-
-			CreatedBy: cmd.SignedInUser.UserId,
-			UpdatedBy: cmd.SignedInUser.UserId,
-		}
-
-		if res, err := session.Query("SELECT 1 from library_panel WHERE org_id=? and folder_id=? and title=?", cmd.OrgID, cmd.FolderID, cmd.Title); err != nil {
+func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd *addLibraryPanelCommand) (*LibraryPanel, error) {
```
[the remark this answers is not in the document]
torkelo: yea, think this pattern of using pointers for command/queries is only needed for bus messages/handlers.

pkg/services/librarypanels/librarypanels.go
torkelo: this is a new table

torkelo (approved):
Looks great! 

Arve has some good small edit suggestions

pkg/services/librarypanels/librarypanels.go
papagian: It's not big problem here because this is a new feature not yet released but if it can be a problem in development installations where this migration has already been executed because for those installations `model` column wouldn't exist.
Therefore is better avoiding doing that. Alternatively a new `NewAddColumnMigration` migration for `model` column can be introduced. More detailed guidelines for database migrations you can refer [here](https://github.com/grafana/grafana/blob/master/contribute/architecture/database.md#migrations).

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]
aknuds1: I guess in the case @torkelo mentions, it's necessary for mutating the argument. That doesn't happen here.

pkg/services/librarypanels/librarypanels.go
aknuds1: @torkelo It's not really a new table though, since the migration is already in master. I guess it's behind a feature flag OTOH, and can cause the problems @papagian mentions. So yeah I would agree with @papagian and say it's best to not take this risk.

torkelo: no dev installations have this feature enabled yet

torkelo: it's behind a new feature toggle that only people working on this feature (hugo) has enabled

aknuds1: OK, if we can be sure it doesn't affect any deployments, it should be safe.

torkelo: Since it's early days for this feature we decided to put the db migration behind the feature toggle as well so we can make db changes without migrations during this initial phase, of course as more developers start testing/using it we need to start using db migrations

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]
hugohaggmark: Ok, thank you. I'll review all my pointer & value usage.

pkg/services/librarypanels/api.go
```diff
@@ -0,0 +1,31 @@
...
+	})
+}
+
+// createHandler handles POST /api/library-panels.
+func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibraryPanelCommand) api.Response {
+	panel, err := lps.createLibraryPanel(c, &cmd)
+
+	if err != nil {
+		return api.Error(500, "Failed to create library panel", err)
+	}
```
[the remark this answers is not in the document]
hugohaggmark: Yes, that makes sense. I'll change that.

pkg/services/librarypanels/database.go
hugohaggmark: Wow, very nice. I love this suggestion!

pkg/services/librarypanels/librarypanels.go
hugohaggmark: This was a deliberate change from my side and I figured that it wouldn't impact anyone (behind feature toggle and new table) but thank you for clarifying the process moving forward.

pkg/services/librarypanels/database.go
hugohaggmark: A newb question, you're passing a pointer here because __`Insert`__ mutates the __`libraryPanel`__ variable right? 🙈

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Yes, will fix this.

pkg/services/librarypanels/database.go
aknuds1: Yes, I'm pretty certain session.Insert should mutate `libraryPanel.ID`, even though xorm doesn't document it.

hugohaggmark: Right, thank you

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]

pkg/services/librarypanels/api.go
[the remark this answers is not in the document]
hugohaggmark: Done

pkg/services/librarypanels/database.go
```diff
@@ -0,0 +1,45 @@
...
+	"time"
+
+	"github.com/grafana/grafana/pkg/models"
+
+	"github.com/grafana/grafana/pkg/services/sqlstore"
+)
+
+// createLibraryPanel function adds a LibraryPanel
+func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd *addLibraryPanelCommand) (*LibraryPanel, error) {
+	libraryPanel := &LibraryPanel{
```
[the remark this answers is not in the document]
hugohaggmark: Done

pkg/services/librarypanels/database.go
hugohaggmark: Done

pkg/services/librarypanels/database.go
hugohaggmark: Done

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Done

pkg/services/librarypanels/api.go
```diff
@@ -0,0 +1,31 @@
...
+// createHandler handles POST /api/library-panels.
+func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibraryPanelCommand) api.Response {
+	panel, err := lps.createLibraryPanel(c, cmd)
+
+	if err != nil {
+		return api.Error(400, "Failed to create library panel", err)
+	}
+
+	return api.JSON(200, util.DynMap{"panel": panel})
+}
```
hugohaggmark, line 31: I've noticed that we usually add api handlers as methods on our entities and not as functions. Is there any specific reasons for this? I'm guessing convenience @papagian @aknuds1 but I'm curious to why?

pkg/services/librarypanels/api.go
```diff
@@ -21,10 +21,10 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
 
 // createHandler handles POST /api/library-panels.
 func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibraryPanelCommand) api.Response {
-	panel, err := lps.createLibraryPanel(c, &cmd)
+	panel, err := lps.createLibraryPanel(c, cmd)
 
 	if err != nil {
-		return api.Error(500, "Failed to create library panel", err)
+		return api.Error(400, "Failed to create library panel", err)
```
aknuds1, line 27:
```suggestion
		if errors.Is(err, errLibraryPanelAlreadyAdded) {
			return api.Error(400, errLibraryPanelAlreadyAdded.Error(), err)
		}
		return api.Error(500, "Failed to create library panel", err)
```

aknuds1 (changes requested):
Looks good except the check for error type is still missing.

pkg/services/librarypanels/api.go
hugohaggmark: So instead of 

```go
func (lps *LibraryPanelService) registerAPIEndpoints()
```

Why not

```go
func registerAPIEndpoints(lps *LibraryPanelService)
```

aknuds1: How would the handler have access to `lps.createLibraryPanel` without the method receiver, i.e. `lps`?

hugohaggmark: You would have to pass lps *LibraryPanelService to the handler I guess?

aknuds1: Ah, I see you're asking if it should be a free function receiving `lps`. I guess you could say that about any method, it could receive the object as a normal argument instead? I would say it's just about what makes sense logically. If it's working on `lps`, why not make it a method?

pkg/services/librarypanels/api.go
hugohaggmark: Great catch, than you 🙈

aknuds1 (approved):
LGTM

pkg/services/librarypanels/api.go
hugohaggmark: I'm just trying to understand how we define things around here :) I mean `createLibraryPanel` could also be a "free" function and then there is no reason to pass lps at all. I'm just trying to figure out when to define methods (on/for structs) and when to define "free" functions. 

I got the question from https://www.youtube.com/watch?v=29LLRKIL_TI&feature=youtu.be&t=891 when I was digging deeper into pointers & values. 

So to sum it up:
- use what makes most sense logically
- prefer methods over functions to minimize pollution of namespaces

Thank you again for all the help @aknuds1 @papagian @torkelo

PR #29644, opened 2020-12-07
Chore: Contexthandler service improvements

aknuds1:
Make improvements to ContextHandler service, removing even more globals.

I recommend diffing using a graphical tool.


pkg/setting/setting.go
```diff
@@ -724,7 +718,7 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 	cfg.IsEnterprise = IsEnterprise
 	cfg.Packaging = Packaging
 
-	cfg.ErrTemplateName = ErrTemplateName
+	cfg.ErrTemplateName = "error"
```
marefr, line 721: Nit. Converting this from a constant is a bit weird when it cannot be changed, but no biggie.

marefr (approved):
LGTM

Think Enterprise will need changes to adhere to these changes (maybe)
Added a nit comment

pkg/setting/setting.go
aknuds1: @marefr It's originally not a constant, but a [package global variable](https://github.com/grafana/grafana/pull/29644/files/1bbdf27d70fa22a79b005daf4ae4ed66f4eae202#diff-dc899ea32f268c53e94f76ae5ff45a239dde212d815e715fc3804ea4dfe87a75L44).

aknuds1:
> LGTM
> 
> Think Enterprise will need changes to adhere to these changes (maybe)
> Added a nit comment

@marefr Thanks! Let me see if I can find any problems with enterprise.

PR #29650, opened 2020-12-07
Auth: Add SigV4 header allowlist to reduce chances of verification issues

### LINKED ISSUE #29566: Enhance datasource SigV4 auth to allow excluding some header to be signed
This is a followup issue for https://github.com/grafana/grafana/pull/27552

It would be nice if we can allow user to configure the SigV4 auth to exclude certain headers to be included in the signing process. 

**Why is this needed**:

Some headers gets modified hop-by-hop. One such example I discovered is `x-request-id`, some gateway might drop rthe client passed in value for `x-request-id` and generate a new ID. 

This is a problem because consider the following case: 
1. Grafana SigV4 signed a `x-request-id` with value 'ID-1234`
1. Grafana sends the request to remote data source. 
1. The gateways between Grafana and the remote data source ignores the `x-request-id` from the caller, and generate a new ID. 
1. Now, the auth will fail because what Grafana signed is different from what remote data source sees.  

In fact, `x-request-id` is just an example; there may be other headers suffers same problem; hence, it would be nice if we can allow customer to configure what headers to be excluded from sigV4 signing process.

wbrowne:
HTTP Headers can be problematic with AWS request signing verification - therefore in order to minimize the risk of auth problems, Grafana will only sign a determined list of headers as part of the request and strip any others. 

Fixes #29566 

I also made a frontend fix to make sure the default for `sigV4AuthType` is set for jsonData


pkg/models/sigv4.go
```diff
@@ -24,6 +24,13 @@ const (
 	Credentials AuthType = "credentials"
 )
 
+var permittedHeaders = map[string]struct{}{
```
bergquist, line 27: Shouldn't `Host` be in here?

wbrowne: Should have mentioned - it turns out Go uses the `req.Host` field as the source of truth for that and Go takes care of that https://github.com/golang/go/blob/cad6d1fef5147d31e94ee83934c8609d3ad150b7/src/net/http/request.go#L92. I think it makes sense to add it here still though so thanks for the reminder!

wbrowne: It could be that it won't ever be in the header map (AFAICT at least) - but might be confusing to omit it. At the very least, probably should have a comment but I think putting the entry in the map is fine

bergquist (approved):
The backend change looks good.

torkelo (approved):
Looks good

PR #29653, opened 2020-12-07
Expr: fix failure to execute due to OrgID

kylebrandt:
For server side expressions, gets orgID from the plugin context instead of json body of query, which makes more sense anyways.

Needed to makes expressions work again after https://github.com/grafana/grafana/pull/29449 changes.

### DIFF
--- a/pkg/expr/graph.go
+++ b/pkg/expr/graph.go
@@ -52,8 +52,8 @@ func (dp *DataPipeline) execute(c context.Context) (mathexp.Vars, error) {
 
 // BuildPipeline builds a graph of the nodes, and returns the nodes in an
 // executable order.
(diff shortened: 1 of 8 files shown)


kylebrandt:
@papagian With this I imagine we could also change

https://github.com/grafana/grafana/blob/c869ae5791be44834b8862562e8b4674c9a80f1c/pkg/services/ngalert/eval/eval.go#L101-L106

to set the OrgID on the plugin context and and then drop:

https://github.com/grafana/grafana/blob/c869ae5791be44834b8862562e8b4674c9a80f1c/pkg/services/ngalert/eval/alert_query.go#L275-L276

Not sure if should be in this PR, or if that is just going to make for annoying conflicts for you.

wbrowne (approved):
👍

Issue #28440, opened 2020-10-21
Database: Enabled database metrics feature flag corrupts MySQL database tables on restart


papagian:
**What happened**:
Enabled the feature flag `database_metrics` and restarted Grafana. Then restarted Grafana again and the MySQL database migrations failed and .

**What you expected to happen**:
Grafana to start and MySQL database migrations to run successfully and don't corrupt any database tables.

**How to reproduce it (as minimally and precisely as possible)**:
1. start mysql:
`make devenv sources=mysql`
2. run grafana (7.2.2) using:
`GF_LOG_LEVEL=debug  GF_DATABASE_TYPE=mysql GF_DATABASE_HOST=127.0.0.1:3306 GF_DATABASE_USER=grafana GF_DATABASE_PASSWORD=<password> GF_FEATURE_TOGGLES_ENABLE=database_metrics GF_LOG_FILTERS=DBMetrics:debug make run`

2. wait for the server to start
3. restart the server
4. the server fails to start because tries to rerun migrations and reports database errors

**Anything else we need to know?**:
I have tried also running with postgreSQL and sqlite and I'm unable to reproduce it.
Also I have tried the above scenario without the feature flag enabled and I can't reproduce it.

**Environment**:
- Grafana version: 7.2.2 (and master before #28438 was merged)

papagian:
Digging a bit deeper into it; I suspect that the cause is the `"driver: skip fast-path; continue as if unimplemented"` error when restarting grafana with MySQL database and with `GF_FEATURE_TOGGLES_ENABLE=database_metrics`

the log extract when starting grafana with empty database with the feature toggle enabled:
```
t=2020-10-22T13:22:05+0300 lvl=info msg="Starting DB migrations" logger=migrator
t=2020-10-22T13:22:05+0300 lvl=dbug msg="query finished" logger=sqlstore.metrics status=success elapsed time=6.752µs sql="SELECT `TABLE_NAME` from `INFORMATION_SCHEMA`.`TABLES` WHERE `TABLE_SCHEMA`=? and `TABLE_NAME`=?" error="driver: skip fast-path; continue as if unimplemented"
t=2020-10-22T13:22:05+0300 lvl=dbug msg="query finished" logger=sqlstore.metrics status=success elapsed time=1.236763ms sql="SELECT `TABLE_NAME` from `INFORMATION_SCHEMA`.`TABLES` WHERE `TABLE_SCHEMA`=? and `TABLE_NAME`=?"
[...]

Issue #29489, opened 2020-12-01
Instrumentation: All database requests should be instrumented, logged and traced. 


bergquist:
I'd like to see all database requests being instrumented using a histogram and a gauge of current flight requests to be able to see if the database is the bottleneck and how many DB requests are sent from a Grafana instance. 

We should also trace all database requests so we can identify which queries are slow.

https://github.com/grafana/grafana/pull/28236 was an attempt at solving this but didn't make it since it had problems with the MySQL driver.

PR #29662, opened 2020-12-07
Instrumentation: Add histograms for database queries 

dafydd-t:
This PR intends to re-enable a "hidden" feature we added in [this PR](https://github.com/grafana/grafana/pull/28236/files), plus fixing the bug that caused us to revoke it.

The changes in this original PR caused grafana to corrupt its database when using MySQL. We think this happened because we were not passing enough information into the URI constructor in the wrapped database's `Parse` method.

This contributes to #29489 by adding metrics, but we still need to add traces.

The signifiant change from the original feature is to use the "unwrapped" driver's `Parse` function, instead of creating a new one.

Verifying the fix: In the original PR, we encountered a bug when starting a Grafana instance with an existing backend (a previously populated db). The Grafana migration script ran again and corrupted the existing data. Therefore we need to test that this new instance can start with a clean and existing database. Check this is possible for:
- [x] SQLite3
- [x] MySQL 
- [x] Postgres 

Snapshot of metrics when feature flag enabled:
```
# HELP grafana_database_queries_duration_seconds Database query histogram
# TYPE grafana_database_queries_duration_seconds histogram
grafana_database_queries_duration_seconds_bucket{le="0.0001"} 0
grafana_database_queries_duration_seconds_bucket{le="0.0004"} 0
grafana_database_queries_duration_seconds_bucket{le="0.0016"} 1026
grafana_database_queries_duration_seconds_bucket{le="0.0064"} 17491
grafana_database_queries_duration_seconds_bucket{le="0.0256"} 17751
grafana_database_queries_duration_seconds_bucket{le="0.1024"} 17752
grafana_database_queries_duration_seconds_bucket{le="0.4096"} 17752
grafana_database_queries_duration_seconds_bucket{le="1.6384"} 17752
grafana_database_queries_duration_seconds_bucket{le="6.5536"} 17752
grafana_database_queries_duration_seconds_bucket{le="+Inf"} 17752
grafana_database_queries_duration_seconds_sum 41.19716338399985
grafana_database_queries_duration_seconds_count 17752
# HELP grafana_database_queries_total The total amount of Database queries
# TYPE grafana_database_queries_total counter
grafana_database_queries_total{status="success"} 17752
```

Note: When flag is disabled, the metrics still get registered, but they are never incremented. @bergquist is this the desired behaviour? If not, could move it into `WrapDatabaseDriverWithHooks` I think.

### DIFF
--- a/pkg/services/sqlstore/database_wrapper.go
+++ b/pkg/services/sqlstore/database_wrapper.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/gchaincl/sqlhooks"
@@ -18,30 +19,23 @@ import (
 )
 
 var (
-	databaseQueryCounter   *prometheus.CounterVec
-	databaseQueryHistogram prometheus.Histogram
+	databaseQueryHistogram *prometheus.HistogramVec
 )
 
 func init() {
-	databaseQueryCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Namespace: "grafana",
-		Name:      "database_queries_total",
-		Help:      "The total amount of Database queries",
-	}, []string{"status"})
-
-	databaseQueryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
+	databaseQueryHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "grafana",
 		Name:      "database_queries_duration_seconds",
 		Help:      "Database query histogram",
 		Buckets:   prometheus.ExponentialBuckets(0.0001, 4, 9),
-	})
+	}, []string{"status"})
 
-	prometheus.MustRegister(databaseQueryCounter, databaseQueryHistogram)
+	prometheus.MustRegister(databaseQueryHistogram)
 }
 
 // WrapDatabaseDriverWithHooks creates a fake database driver that
 // executes pre and post functions which we use to gather metrics about
-// database queries.
+// database queries. It also registers the metrics.
 func WrapDatabaseDriverWithHooks(dbType string) string {
 	drivers := map[string]driver.Driver{
 		migrator.SQLite:   &sqlite3.SQLiteDriver{},
@@ -56,7 +50,7 @@ func WrapDatabaseDriverWithHooks(dbType string) string {
 
 	driverWithHooks := dbType + "WithHooks"
 	sql.Register(driverWithHooks, sqlhooks.Wrap(d, &databaseQueryWrapper{log: log.New("sqlstore.metrics")}))
-	core.RegisterDriver(driverWithHooks, &databaseQueryWrapperParser{dbType: dbType})
+	core.RegisterDriver(driverWithHooks, &databaseQueryWrapperDriver{dbType: dbType})
 	return driverWithHooks
 }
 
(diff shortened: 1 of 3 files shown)


bergquist:
> Note: When flag is disabled, the metrics still get registered, but they are never incremented.

Awesome! :rocket: 

I think it would be cleaner to only register them if the feature toggle is enabled.

bergquist:
Thinking a bit more about those metrics I think it would make sense to add a success label to the histogram and get rid of the counter. The histogram includes a total counter anyway and separating success and error latency could be very useful. An increase in latency for error queries would be very bad.

dafydd-t:
> I think it would be cleaner to only register them if the feature toggle is enabled.

I've moved the metrics registration into the `WrapDatabaseDriverWithHooks ` function. The change in behaviour here is that, if the registration fails, it will panic at some later time rather than in `init`. This will happen anywhere we call `SQLStore.Init()`. I can see this is used in the CLI, for example - every time we call an admin command.

Ah, this may not work. We're registering the same two metrics multiple times. I think Prometheus will panic at this.

dafydd-t:
I'm going to put this back into draft whilst I verify/rectify the registration issue and edit the metrics to put the success label into the histogram metric.

bergquist:
I think you can use the feature toggle to decide if those should be registered or not. Once we are happy with this and it's proven stable we should enable this by default.

pkg/services/sqlstore/database_wrapper.go
```diff
@@ -100,12 +99,15 @@ func (h *databaseQueryWrapper) OnError(ctx context.Context, err error, query str
...
-func (hp *databaseQueryWrapperParser) Parse(string, string) (*core.Uri, error) {
-	return &core.Uri{
-		DbType: core.DbType(hp.dbType),
-	}, nil
+func (hp *databaseQueryWrapperDriver) Parse(driverName, dataSourceName string) (*core.Uri, error) {
+	driver := core.QueryDriver(hp.dbType)
+	if driver == nil {
+		return nil, fmt.Errorf("could not find driver with name %s", hp.dbType)
+	}
+	return core.QueryDriver(hp.dbType).Parse(driverName, dataSourceName)
```
aknuds1, line 112: Isn't this equivalent to the following? I think being explicit about that would help in reading the code (looks as if the method can work with a `nil` receiver).

```suggestion
	return driver.Parse(driverName, dataSourceName)
```

aknuds1 (approved):
Looks good AFAICT.

pkg/services/sqlstore/database_wrapper.go
dafydd-t: Yes, you're right. Thank you!

dafydd-t:
I've removed the Counter metric and I've put the metric registation back in `init`.

The metrics exported now look like this when the toggle is enabled:

```
# HELP grafana_database_queries_duration_seconds Database query histogram
# TYPE grafana_database_queries_duration_seconds histogram
grafana_database_queries_duration_seconds_bucket{status="success",le="0.0001"} 10227
grafana_database_queries_duration_seconds_bucket{status="success",le="0.0004"} 10227
grafana_database_queries_duration_seconds_bucket{status="success",le="0.0016"} 20079
grafana_database_queries_duration_seconds_bucket{status="success",le="0.0064"} 20503
grafana_database_queries_duration_seconds_bucket{status="success",le="0.0256"} 20509
grafana_database_queries_duration_seconds_bucket{status="success",le="0.1024"} 20509
grafana_database_queries_duration_seconds_bucket{status="success",le="0.4096"} 20509
grafana_database_queries_duration_seconds_bucket{status="success",le="1.6384"} 20509
grafana_database_queries_duration_seconds_bucket{status="success",le="6.5536"} 20509
grafana_database_queries_duration_seconds_bucket{status="success",le="+Inf"} 20509
grafana_database_queries_duration_seconds_sum{status="success"} 10.623043503000003
grafana_database_queries_duration_seconds_count{status="success"} 20509
```

Disabled: nothing. I didn't expect this. Is the behaviour of registering a `HistogramVec` different to a `Histogram`?

bergquist:
Makes sense since it cannot create the histogram without a label value. But it's also weird that they behave differently.

papagian:
awesome! I plan to review and test it this afternoon.

papagian (approved):
I have tested restarting grafana with MySQL and PostgreSQL and I can get the histograms 🎉 
After merging this we can close #28440.

dafydd-t:
Thanks for reviewing @papagian. I am not able to merge this, so I'll just leave it here.

bergquist (approved):
LGTM!

Tested on mysql and postgres :)

bergquist:
:fireworks:

PR #29665, opened 2020-12-07
Expr: make reduction nan/null more consistent

kylebrandt:
More consistent behavior with returning NaN from reduction operations. Noticed while starting to draft some docs.

The behavior is aggressive in returning NaNs, but plan to add options in the future to change behavior.


papagian (approved):
AFAICT LGTM

PR #29679, opened 2020-12-08
Typescript: Fixing typescript strict error, and separate check from publishing

torkelo:
* Fixes a few strict errors
* Separate limit check from publishing
* Add check to PR frontend test action
* Lower limit so it's too low to see if PR fail


hugohaggmark (approved):
LGTM!

Issue #27848, opened 2020-09-28
Graph NG


dprokop:
This issue tracks work to be continued after https://github.com/grafana/grafana/pull/27572:

- [x] Data retrieval in plugins (access to data frame and correct field within it)
- [x] Hiding series makes panel to break  https://github.com/grafana/grafana/pull/28290
- [x] update uPlot to 1.2.x and update x-axis time formatting (ref https://github.com/leeoniya/uPlot/issues/303) (https://github.com/grafana/grafana/pull/28569) @dprokop 
- [x] switch to native `ms` mode in uPlot 1.4.5 (#29445)
- [x] Refactor plot initialization logic to use https://github.com/grafana/grafana/pull/29006 @dprokop 
- [ ] Bring back GraphNG and hooks tests
- [x] Bars, Staircase, Smooth, points-only rendering modes (https://github.com/grafana/grafana/pull/29359)
- [x] Explicit axis sizing for graph alignment (https://github.com/grafana/grafana/pull/29553)
- [x] Enable scale distribution config: 🏃 @dprokop https://github.com/grafana/grafana/pull/29684
- [x] Match hover point styles with Flot (https://github.com/grafana/grafana/pull/29866) @leeoniya 
- [x] Support for null gaps in smooth and staircase path builders (https://github.com/grafana/grafana/pull/29593) @leeoniya 
- [x] Band charts (fillDownTo) (https://github.com/grafana/grafana/pull/30268) @ryantxu
- [x] Series config - null as 0/connected (https://github.com/grafana/grafana/pull/28118, https://github.com/grafana/grafana/pull/29512)
- [ ] Annotations
  - [ ] Display annotations on the visualization
    - [x] Marks (#27972) 
    - [ ] Regions
  - [ ] Editor. Currently blocked by the lack of possibility to correctly attach mouse events to uPlot, but there is an open [proposal](https://github.com/leeoniya/uPlot/issues/298) that will enable this
    - [ ] https://github.com/grafana/grafana/pull/30419
- [ ] Legend plugin
  - [x] MVP(2e71401)
  - [x] Legend sizing
  - [x] Toggling series (#29575)
  - [x] Table legend stats (https://github.com/grafana/grafana/pull/30251)
    - [ ] Sorting
  - [x] Series color control via legend icon (currently series color is configured via field settings) (#30256)
- [x] Dark theme colors adjustments (text, axis grid etc)
- [x] Dark theme zoom region not visible 
- [x] Data links support via context menu
[...]

PR #29684, opened 2020-12-08
GraphNG: Enable scale distribution configuration

dprokop:
Ref: https://github.com/grafana/grafana/issues/27848
This adds a custom editor of scale distribution and updates the UPlotScaleBuilder to use provided config.

The custom editor looks like this: 
Kapture 2020-12-08 at 14 18 28

Added placeholder for currently supported log32 and log1024 modes. These are currently not supported by uPlot([feature request](https://github.com/leeoniya/uPlot/issues/393)), as it only supports log2 and log10 scales.

@leeoniya if you could take a look at this that would be awesome. I haven't tracked down the issue, but apparently enabling log scales via Scale config described in https://github.com/leeoniya/uPlot/blob/master/dist/uPlot.d.ts#L393 does not work. See the comparison of angular vs graph ng.  GraphNG is rendered without y axis and no data is shown. This test is performed on the simplest CSV input for test data source:
image

Scale configuration is as bellow:

```
temp2.areaMI2
{time: undefined, distr: 3, log: 2, range: ƒ}
```


leeoniya:
i'll take a look later today. but this demo works okay for 10 and 2: https://leeoniya.github.io/uPlot/demos/log-scales2.html

dprokop:
> i'll take a look later today. but this demo works okay for 10 and 2: https://leeoniya.github.io/uPlot/demos/log-scales2.html

Yups, it does, that's where I took inspiration from :) I wonder if I'm missing anything.

ryantxu (approved):
:fire:

leeoniya:
@dprokop this now works correctly for +Y values, but crashes and burns for 0 and negative values, because `Math.log()` is undefined for these values. i've added a few commits to uPlot that result in similar behavior to Flot, effectively excluding `<= 0` from scale ranging, and rendering their values an order of magnitude below the chart edge.

i plan to do a uPlot bump tomorrow or Thurs, so let's wait till that happens before landing this.

mckn (approved):
LGTM! I like the UI or this.

public/app/plugins/panel/graph3/ScaleDistributionEditor.tsx
```diff
@@ -0,0 +1,71 @@
...
+          });
+        }}
+      />
+      {value.type === ScaleDistribution.Logarithmic && (
+        <Select
+          allowCustomValue={false}
+          autoFocus
+          options={LOG_DISTRIBUTION_OPTIONS}
+          value={value.log || 2}
+          prefix={'log'}
```
ryantxu, line 59: Maybe prefix with base instead of log

ryantxu (approved):
Looks good

ryantxu:
merged with master and changed "log" to "base" in the UI:
image

PR #29691, opened 2020-12-08
Chore: Add CloudWatch HTTP API tests

aknuds1:
Add suite of HTTP API tests for CloudWatch. These are basically integration tests, that only stub out external services. Really good for testing the CloudWatch functionality, and catching bugs IMO.

Let me know if I should try to divide into smaller PRs.

**Update:**
I've moved the tests to pkg/tests/api/metrics, which I think is a logical location for them. I tried putting them in pkg/tsdb/cloudwatch, but this is a bad idea since it results in an import cycle.


sunker:
Though I have to say I don't have enough knowledge when it comes to some parts of Grafana core, I think this looks like a great start. The ability to run http tests is going to be a really useful for datasource development. 

It feels a bit off to me to have Cloudwatch specific tests in the `pkg/server` dir. Wouldn't it be possible to have some sort of "HttpTestUtil"(or something) that is completely datasource agnostic that exposes grafanaDirectories, sqlStore and the GrafanaTestServer, basically these lines of code:
```
grafDir, cfgPath := createGrafDir(t)
sqlStore := setUpDatabase(t, grafDir)
addr := startGrafana(t, grafDir, cfgPath, sqlStore)  
```
Then the CloudWatch tests in pkg/server/api_metrics_test.go could be moved to some api test file placed in pgk/tsdb/cloudwatch.

aknuds1:
> Though I have to say I don't have enough knowledge when it comes to some parts of Grafana core, I think this looks like a great start. The ability to run http tests is going to be a really useful for datasource development.
> 
> It feels a bit off to me to have Cloudwatch specific tests in the `pkg/server` dir. Wouldn't it be possible to have some sort of "HttpTestUtil"(or something) that is completely datasource agnostic that exposes grafanaDirectories, sqlStore and the GrafanaTestServer, basically these lines of code:
> 
> ```
> grafDir, cfgPath := createGrafDir(t)
> sqlStore := setUpDatabase(t, grafDir)
> addr := startGrafana(t, grafDir, cfgPath, sqlStore)  
> ```
> 
> Then the CloudWatch tests in pkg/server/api_metrics_test.go could be moved to some api test file placed in pgk/tsdb/cloudwatch.

@sunker Oh yeah, good call. The placement of the tests is actually the one thing I was the most unsure about. It was the easiest to place them in pkg/server/ as I did, but I can understand that it's not ideal in the long run.

aknuds1:
@sunker I will try to implement your suggestion.

aknuds1:
@sunker I tried moving the tests to pkg/tsdb/cloudwatch, but realized it simply doesn't work since it causes an import cycle. Thinking about it some more, I realized what makes sense is to structure the HTTP API integration tests in the repo according to what API they're hitting. With that in mind, I placed them now in pkg/tests/api/metrics/. I also put some READMEs in that directory tree to document the structure.

sunker (approved):
@aknuds1 I see your point about circular deps. 

I think breaking out `createGrafDir` and `startGrafana` into a separate file could make it easier for the next person to get started with api testing, but that can be done when the need arrises. 

Great work on this!

PR #29693, opened 2020-12-08
Alerting: Use correct time series name override from frame fields

### LINKED ISSUE #29213: Metric name in alerts always displayed as "Value {}"
**What happened**: All metric names in alerts are being displayed as: `Value {}` (apparently for `CloudWatch` datasource only)

**What you expected to happen**: The label to be showed as it's visible, e.g. in the graph's legend.

**How to reproduce it (as minimally and precisely as possible)**: When an alarm enters pending/error state, the "Metric name" value is always `Value {}`

**Anything else we need to know?**: Data from `CloudWatch` datasource where `Alias` field is set to `{{label}} (XXX)` (_XXX_ is just some text).
Issue observed both on `Email` alerts and `Alert list` panel. Broken on `7.3.2` and `7.3.3` (seemed to be working fine in versions `7.3.1` and earlier as the update from `7.3.1` to `7.3.2` was performed in the afternoon of Nov 12).
**UPDATE**: confirming the issue appeared in `7.3.2` as after downgrading to `7.3.1` it's working as expected.
The screenshot below shows the "broken" PENDING states metric name happening from the 13th of November.

grafana-alerts

**Environment**:
- Grafana version: 7.3.2 and 7.3.3
- OS Grafana is installed on: linux (docker)
- User OS & Browser: windows 10 / chrome 86
- Grafana plugins: n/a
- Others: official grafana docker image

wbrowne:
- Will make sure to only use field label information if it truly exists
- Will consider using `DisplayNameFromDS` if exists (low priority)

Not 100% on the usage of these fields in this context as the API docs are as follow:

```
// DisplayName overrides Grafana default naming, should not be used from a data source
DisplayName string `json:"displayName,omitempty"`

// DisplayNameFromDS overrides Grafana default naming in a better way that allows users to override it easily.
DisplayNameFromDS string `json:"displayNameFromDS,omitempty"`
```

Would be good to have your eyes @torkelo based on your introduction of the `DisplayNameFromDS` field, and whether or not `DisplayName` still makes sense here

Fixes #29213

[Some historical context](https://github.com/grafana/grafana/pull/27186/files) for `DisplayNameFromDS`

Can add more/better tests if necessary - for example, maybe would be good to verify the correct order/priority for each option

### DIFF
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -333,7 +333,7 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 		switch {
 		case field.Config != nil && field.Config.DisplayName != "":
 			ts.Name = field.Config.DisplayName
-		case field.Labels != nil && len(field.Labels) > 0:
+		case len(field.Labels) > 0:
 			ts.Tags = field.Labels.Copy()
 			// Tags are appended to the name so they are eventually included in EvalMatch's Metric property
 			// for display in notifications.
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -333,13 +333,13 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 		switch {
 		case field.Config != nil && field.Config.DisplayName != "":
 			ts.Name = field.Config.DisplayName
+		case field.Config != nil && field.Config.DisplayNameFromDS != "":
+			ts.Name = field.Config.DisplayNameFromDS
 		case field.Labels != nil && len(field.Labels) > 0:
 			ts.Tags = field.Labels.Copy()
 			// Tags are appended to the name so they are eventually included in EvalMatch's Metric property
 			// for display in notifications.
 			ts.Name = fmt.Sprintf("%v {%v}", field.Name, field.Labels.String())
-		case field.Config != nil && field.Config.DisplayNameFromDS != "":
-			ts.Name = field.Config.DisplayNameFromDS
 		default:
 			ts.Name = field.Name
 		}
--- a/pkg/services/alerting/conditions/query.go
+++ b/pkg/services/alerting/conditions/query.go
@@ -333,11 +333,13 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 		switch {
 		case field.Config != nil && field.Config.DisplayName != "":
 			ts.Name = field.Config.DisplayName
-		case field.Labels != nil:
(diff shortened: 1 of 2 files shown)


pkg/services/alerting/conditions/query.go
```diff
@@ -333,11 +333,13 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 		switch {
 		case field.Config != nil && field.Config.DisplayName != "":
 			ts.Name = field.Config.DisplayName
-		case field.Labels != nil:
+		case field.Labels != nil && len(field.Labels) > 0:
```
aknuds1, line 336: You can get the length of nil slices in Go, so the idiom is just the following:
```suggestion
		case len(field.Labels) > 0:
```

aknuds1 (approved):
LGTM

pkg/services/alerting/conditions/query_test.go
```diff
@@ -306,6 +306,53 @@ func TestFrameToSeriesSlice(t *testing.T) {
...
+				},
+			},
+			Err: require.NoError,
+		},
+		{
+			name: "display name from data source",
+			frame: data.NewFrame("",
+				data.NewField("Time", data.Labels{}, []time.Time{}),
+				data.NewField(`Values`, data.Labels{}, []*int64{}).SetConfig(&data.FieldConfig{
+					DisplayNameFromDS: "sloth",
```
kylebrandt, line 328: 🦥

kylebrandt (approved):
LGTM (agree with Arve's un-needed nil check). Do want to make sure torkel agrees this is the right preference behavior though.

pkg/services/alerting/conditions/query.go
wbrowne: Nice tip 🙆‍♂️

ryantxu (approved):

pkg/services/alerting/conditions/query.go
```diff
@@ -333,11 +333,13 @@ func FrameToSeriesSlice(frame *data.Frame) (tsdb.TimeSeriesSlice, error) {
 		switch {
 		case field.Config != nil && field.Config.DisplayName != "":
 			ts.Name = field.Config.DisplayName
-		case field.Labels != nil:
+		case len(field.Labels) > 0:
 			ts.Tags = field.Labels.Copy()
 			// Tags are appended to the name so they are eventually included in EvalMatch's Metric property
 			// for display in notifications.
 			ts.Name = fmt.Sprintf("%v {%v}", field.Name, field.Labels.String())
+		case field.Config != nil && field.Config.DisplayNameFromDS != "":
```
torkelo, line 341: it needs to be above the labels case (otherwise having labels make DisplayNameFromDS have no effect

torkelo (approved):

PR #29695, opened 2020-12-08
DataFrame: add path and description metadata

ryantxu:
When working in many systems we need a consistent way to identify field origins so that we can consistently access and possibly update them.

There are a few larger projects in the pipeline that will use this heavily, however managing multiple branches with distinct metadata models is challenging.  I propose adding these metadata options soon and the implementations that use them in larger projects down the road.

Obviously the same fields need to be added to https://github.com/grafana/grafana-plugin-sdk-go/blob/master/data/field_config.go

See also: https://github.com/grafana/grafana/tree/path-poc

### DIFF
--- a/packages/grafana-data/src/types/data.ts
+++ b/packages/grafana-data/src/types/data.ts
@@ -54,6 +54,16 @@ export interface QueryResultMeta {
   executedQueryString?: string;
 
   /**
+   * A browsable path on the datasource
+   */
+  path?: string;
+
+  /**
+   * defautls to '/'
+   */
+  pathSeperator?: string;
+
+  /**
    * Legacy data source specific, should be moved to custom
    * */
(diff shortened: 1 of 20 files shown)


packages/grafana-data/src/types/data.ts
```diff
@@ -53,6 +53,16 @@ export interface QueryResultMeta {
...
 
+  /**
+   * A browsable path on the datasource
+   */
+  path?: string;
+
+  /**
+   * defautls to '/'
+   */
+  pathSeperator?: string;
```
leeoniya, line 64: "separator"

torkelo, line 64:
```suggestion
  /**
   * defaults to '/'
   */
  pathSeparator?: string;
```

torkelo (approved):
Looks good, just minor spelling issue

PR #29375, opened 2020-11-25
Variables: Fixes issue with upgrading legacy queries

hugohaggmark:
**What this PR does / why we need it**:
When we refactored the variable system in #28217 we added a function for migrating/upgrading the query property for variables that use the StandardVariableSupport. This worked for most cases like importing and for variables with the refresh setting set to `OnDashboardLoad` or `OnTimeRangeChanged` but I forgot about those with refresh setting set to `Never`. This PR moves the logic into the initialization of variables instead.

**Which issue(s) this PR fixes**:
Fixes #29373

**Special notes for your reviewer**:
Instructions on how to test this is in the repro steps for #29373

PR #29704, opened 2020-12-09
Variables: Fixes upgrade of legacy Prometheus queries

### LINKED ISSUE #29703: Variables: Query input is empty when you edit a QueryVariable

hugohaggmark:
When initializing variables we tried to upgrade legacy queries for Prometheus because Prometheus is now using the Standard Variable Support. 

When I tried to fix this in #29375 I missed the case when a query variable depended on a datasource variable so #29375 upgrades legacy queries to early in the initialization process of variables.

This PR moves the upgrade legacy query action to places later on in the initialization process of variables. I also refactored the actions for datasource variables as they were using the now deprecated function __`getMetricSources`__. 

Enjoy your review

Fixes #29703

### DIFF
--- a/public/app/features/variables/datasource/actions.ts
+++ b/public/app/features/variables/datasource/actions.ts
@@ -2,7 +2,7 @@ import { toVariablePayload, VariableIdentifier } from '../state/types';
 import { ThunkResult } from '../../../types';
 import { createDataSourceOptions } from './reducer';
 import { validateVariableSelectionState } from '../state/actions';
-import { DataSourceInstanceSettings, DataSourceSelectItem, stringToJsRegex } from '@grafana/data';
+import { DataSourceInstanceSettings, stringToJsRegex } from '@grafana/data';
 import { getDatasourceSrv } from '../../plugins/datasource_srv';
 import { getVariable } from '../state/selectors';
 import { DataSourceVariableModel } from '../types';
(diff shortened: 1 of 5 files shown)


public/app/features/variables/datasource/actions.ts
```diff
@@ -27,16 +30,21 @@ export const updateDataSourceVariableOptions = (
...
-): ThunkResult<void> => async dispatch => {
-  const dataSources: DataSourceSelectItem[] = await dependencies.getDatasourceSrv().getMetricSources();
-  const filtered = dataSources.filter(ds => !ds.meta.mixed && ds.value !== null);
-  const dataSourceTypes = _(filtered)
+): ThunkResult<void> => dispatch => {
+  const dataSources: DataSourceSelectItem[] = dependencies
+    .getDatasourceSrv()
+    .getList({ metrics: true, variables: true })
+    .filter(setting => {
+      return !setting.meta.mixed && !setting.isDefault;
```
torkelo, line 44: no need for this, mixed is opt-in, and isDefault is just a property on the option, so should filter out the default data source, as then you only have one data source and that is the default then this dropdown will be empty. (the old getMetricSources returned two entries, one named default with value null, getList does not do this, just returns the data sources, and if it happens to be the default isDefault is true)

torkelo (approved):
Looks good, just one minor thing with the use of getList

public/app/features/variables/datasource/actions.ts
hugohaggmark: Great explanation, I'll remove the filter and update the tests.

PR #29706, opened 2020-12-09
NodeGraph: Add node graph visualization

aocenas:
Add Node graph panel. First data source using it will be X-ray but later can be used for any other data source. Can be used as a panel in dashboard, in explore and also as part of trace view.

Screenshot from 2021-01-18 18-26-28


ryantxu (commented):
super interesting!  What does a service map DataFrame look like?  I see xray types in here, I assume that is just for getting started.

Could we add a scenario to the test data that returns the expected structure?

https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/testdata/datasource.ts#L59

torkelo:
Nice, can I ask for a screenshot?

aocenas:
@ryantxu Yes I will put it into test data source, right now I added a test query in the x-ray data source itself. I will have to think a bit about the structure. I don't won't those x-ray types in core grafana so want to move the processing into the x-ray DS. At the same time it needs to be more flexible because when we will use it for Jaeger or other DS they actually have different stats. Like instead of `transaction per minute` we will have to show `self time` because they show conceptually a bit different thing.

But I worry that would be too tied to the visualisation itself. Like if I have `stat1 | stat2 | title | subtitle` it feels a bit weird but don't see a way around that.

aocenas:
@torkelo sorry for the delay was doing some UI fixes still, added screenshots to the description. The layout is still something that's not ideal and we plan to do something better.

packages/grafana-ui/src/components/ServiceMap/GraphViewCy.tsx
```diff
@@ -0,0 +1,214 @@
+import { fromPairs } from 'lodash';
+import React, { useEffect, useState } from 'react';
+import { SimulationNodeDatum, SimulationLinkDatum } from 'd3-force';
+import useMeasure from 'react-use/lib/useMeasure';
+import { response } from './x-ray-response';
+import cytoscape, { ElementDefinition } from 'cytoscape';
```
torkelo, line 6: how big a framework is this? We already use another network graph library (visjs-network), is there a way to use the same? (or update that to use this) 

Used here
https://github.com/grafana/grafana/blob/master/public/app/features/variables/inspect/NetworkGraph.tsx

aocenas: Right now there is a bit more of these as I was testing different libraries and layouts. In the end this will be removed and we will probably go just with the d3-force layout.

ryantxu:
@aocenas -- can you link to an example what the data looks like... and in particular how we stuff it into dataframes?  It would be great to have a clear mapping that fits into a table structure -- 

Maybe:
| nodeId | name | ... | Edges |
|---|---|---|---|
| A | Service A | ... | B,C |
| B | Service B | ... |  |
| C | Service C | ... |  |

Not great if we want/need many edge properties

Or maybe better with two tables, one for nodes and one for edges?


Maybe:
| nodeId | name | ...  |
|---|---|---|
| A | Service A | ... |
| B | Service B | ... |
| C | Service C | ... |


| from | to | name | ...  |
|---|---|---|---|
| A | B | ? | ... |
| A | C | ? | ... |

The two frame approach is nice because then we have clear place for properties associated with the node vs edge

aocenas:
@ryantxu 
We definitely need 2 tables one for edges one for nodes so it will be something like your second example. One thing I am changing right is to make the mapping more generic and not x-ray specific. Ie we will probably need something like:
|id|name|statMain1|statMain2|percentage1|subName|
|--|--------|-------------|-------------|----------------|-------------|

Reason for that is that we may need to display different values later on. With x-ray service map we now show average response time and number of transactions per minute, but with jaeger trace graph we may want to show self time and total time for a span. So the semantics of those values will be different and I do not want to hard code it, because jaeger just don't have the same stats.

aocenas:
@ryantxu @torkelo 
So I think actually it makes more sense to use labels on fields to figure out how to map the data. For example with the colored circle around the node I think it make sense also for the colors to not be hard coded so having a field with labels: `{nodeGraphValueType: 'arc', color: 'purple'}` makes more sense to me than having predefined column names.

One downside could be that right now you could rename columns in transformation but you cannot add labels, but I assume adding such transform would not be hard.

ryantxu:
How would labels work?  Aren't they constant across a whole column?

packages/grafana-ui/src/components/NodeGraph/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  const newy1 = line.y1 + yDiff / 2;
+  return {
+    x1: newx1,
+    y1: newy1,
+    x2: newx1 + vx2,
+    y2: newy1 + vy2,
+  };
+}
+
+export function getNodeFields(nodes: DataFrame) {
```
aocenas, line 27: @torkelo @ryantxu @dprokop this is the main logic how the dataFrame is parsed and fields assigned to particular placeholders. I use constant label values to mark any column and then it's used later on. This means it is possible to send columns with what ever names/values so it would make sense for example in a table view, for example to get Node with highest percentage of errors and then just label some of them to be used for service map. It's also possible to `NodeGraphDetailField` label to add any fields into the context menu header. This should be flexible enough for jaeger/zipkin/tempo later on.

Main downside is that data source will have to add dependency for this (maybe the constants should be in grafana/data but considering this is UI concern I am not sure).

Another thing is that maybe formatting is bit lacking, not sure how to easily add thing like prefix (in x-ray there is `avg. 150ms` as one of the stat inside the node. With this we can just add string type column and have that `avg.` prefix there but seems a bit awkward)

aocenas: Here is a relevant code on the x-ray side of things https://github.com/grafana/x-ray-datasource/pull/60/files#diff-58036290d63f45dcebfa80ddeeb5877187a613f7cdb28e9a5482324f632c14adR225

packages/grafana-ui/src/components/NodeGraph/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    details: fieldsCache.getFieldsByLabel(DataFrameLabels.type, DataFrameDetailValues.true),
+  };
+}
+
+export const DataFrameLabels = {
+  type: 'NodeGraphValueType',
+  detail: 'NodeGraphDetailField',
+};
+
+export const DataFrameTypeValues = {
```
torkelo, line 57: maybe an enum is a better option here?

packages/grafana-ui/src/components/NodeGraph/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    id: fieldsCache.getFieldByName('id'),
+    source: fieldsCache.getFieldByName('source'),
+    target: fieldsCache.getFieldByName('target'),
+    mainStat: fieldsCache.getFieldsByLabel(DataFrameLabels.type, DataFrameTypeValues.mainStat)[0],
+    secondaryStat: fieldsCache.getFieldsByLabel(DataFrameLabels.type, DataFrameTypeValues.secondaryStat)[0],
+    details: fieldsCache.getFieldsByLabel(DataFrameLabels.type, DataFrameDetailValues.true),
+  };
+}
+
+export const DataFrameLabels = {
```
torkelo, line 52: maybe an enum is a better option here?

public/app/features/explore/Explore.tsx
```diff
@@ -285,26 +288,99 @@ export class Explore extends React.PureComponent<ExploreProps, ExploreState> {
...
+        showPanel={true}
+        timeZone={timeZone}
+        onUpdateTimeRange={this.onUpdateTimeRange}
+        showBars={false}
+        showLines={true}
+      />
+    );
+  };
+
+  renderTablePanel = (width: number) => {
```
torkelo, line 310: please use prototype methods for all these renderXX functions. Arrow (instance) methods should only be used if they are required for a callback. Both more readable and requires less memory

public/app/features/explore/NodeGraphContainer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return {
+    range: state.explore[exploreId].range,
+  };
+}
+
+const mapDispatchToProps = {
+  splitOpen,
+};
+
+export const NodeGraphContainer = connect(mapStateToProps, mapDispatchToProps)(UnconnectedNodeGraphContainer);
```
torkelo, line 41: This is not fully / correctly typed (the exported type does not have correct props for example), See how to do type this:
https://github.com/grafana/grafana/blob/master/contribute/style-guides/redux.md#typing-of-connected-props

torkelo (commented):
Looks good found some minor issues. 

Is there any test data scenarios to test this panel? That would help, that and a devenv test dashboard + e2e test (need that for all panels). 

On the dataframe data model the panel expects: Initial thought is that the heavy use of labels to mark different fields will not make this visualization a possible option for almost all data sources that does not have a special code to return this specific structure. Is there a way we can use the field names only and have conventions around the naming of fields. That should open up other data sources (csv) + transforms to use this visualization

public/app/plugins/panel/nodeGraph/plugin.json
```diff
@@ -0,0 +1,16 @@
...
+  "id": "serviceMap",
+
+  "info": {
+    "author": {
+      "name": "Grafana Labs",
+      "url": "https://grafana.com"
+    },
+    "logos": {
+      "small": "img/icn-logs-panel.svg",
+      "large": "img/icn-logs-panel.svg"
```
torkelo, line 13: Need to talk to Song and get a new panel icon designed. Need to mark this panel as alpha for now if you intend to merge it before that,

aocenas:
> Is there any test data scenarios to test this panel?
This is still in TODO also with tests. Wanted to get some feedback about the code to make sure there is nothing major that would need rework.

Right now there is a test response when using latest x-ray data source code, but will move it to test data source.

aocenas:
@torkelo Added test data scenario for this:
Screenshot from 2021-01-14 15-09-05

public/app/plugins/datasource/testdata/nodeGraphUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+    },
+    source: {
+      values: new ArrayVector(),
+      type: FieldType.string,
+    },
+    target: {
+      values: new ArrayVector(),
+      type: FieldType.string,
+    },
+  };
```
torkelo, line 107: This data structure with 2 data frames and a bunch of hard code labels makes this visualization very hard for any other data source to use with out a bunch of specific code. 

Is this dual data frame structure needed ? have you explored a single data frame structure like this? 

| Source | Target | Title                       | SubTitle       | MainStat | SecStat | ErrorRate |
| ------ | ------ | --------------------------- | -------------- | -------- | ------- | --------- |
| Node1  | Node2  | NetworkA                    | Some sub title | 10       | 20      | 0.2       |
| Node2  | Node3  | NetworkB                    | Another sub    | 10       | 20      | 0.2       |
| Node3  |        | ServerA (edge node)         | sub title      | 20       | 10      | 0.1       |
| Node2  | Node4  |                             |                | 25       | 1       | 0.5       |
| Node4  |        | ServerB (another edge node) |                | 50       | 20      | 0.3       |

Where each node is a row and connections are defined with the target field. To create one to many connections the same Source property would be duplicate (but with different targets).

torkelo (commented):
Did not understand that this was going to be a new panel until this week :) 

The data structure feels overly complicated, very few data sources will support this visualisation and those that want to have to create a very special dual data frame response. 

I am sure there is a good reason for this dual data frame + label structure but proposed a simpler single data frame model. But this has probably been evaluated.

aocenas:
@torkelo changed the data frame API to use column names here https://github.com/grafana/grafana/pull/29706/commits/b2ed2608f9ba70e28932f3f5a7d617639df589f6 and also marked the panel as beta. The PR should be final now.

torkelo:
@aocenas great, hope this will make it easier for others to adopt

ivanahuckova (approved):
Let's get this merge! 🙂

yarn.lock
```diff
@@ -11133,6 +11135,15 @@ d3-force@1:
     d3-quadtree "1"
     d3-timer "1"
 
+d3-force@^2.1.1:
```
zoltanbedi, line 11138: It is kind of weird that you only added d3-force and yet there are many more changes here. Probably not going to do harm.

zoltanbedi (approved):
Good job! ⛄️

yarn.lock
aocenas: hmm true, bit weird, I had some other libs here when testing then removed, afterwards got some conflicts with master. I think I will try to reset this to master and install d3-force on clean lock file.

zoltanbedi: looks better 👌

packages/grafana-ui/src/components/NodeGraph/NodeGraph.tsx
```diff
@@ -0,0 +1,243 @@
...
+          config={config}
+          onConfigChange={setConfig}
+          onMinus={onStepDown}
+          onPlus={onStepUp}
+          scale={scale}
+        />
+      </div>
+
+      {hiddenNodesCount > 0 && (
+        <div className={styles.alert} aria-label={'Nodes hidden warning'}>
```
Elfo404, line 151: I see that you are doing this because we "miss" a warning variant for the alert component (meaning we have the variant but it's identical to the error one), I think there was some discussion on reintroducing it (with the orange background), so maybe 

```ts
<div className={styles.alertContainer}>
  <Alert title="" severity="warning">
    {hiddenNodesCount} nodes are hidden for performance reasons.
  </Alert>
</div>
```

and, instead of https://github.com/grafana/grafana/pull/29706/files#diff-ffd3d08d366e741396b93d7037a7145086a2ccfa69e4c86f502e8dcde02ed879R51 :

```ts
  alertContainer: css`
    position: absolute;
    top: 0;
    right: 0;
  `,
```

what do you think?

also, the reason is that the text is way too small on high-res screens:
Screenshot 2021-01-19 at 09 55 40
Screenshot 2021-01-19 at 10 10 28

packages/grafana-ui/src/components/NodeGraph/ViewControls.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [showConfig, setShowConfig] = useState(false);
+  const styles = getStyles(useTheme());
+
+  // For debugging the layout, should be removed here and maybe moved to panel config later on
+  const allowConfiguration = false;
+
+  return (
+    <>
+      <Button icon={'plus-circle'} onClick={onPlus} size={'sm'} title={'Zoom in'} />
+      <Button className={styles.minusButton} icon={'minus-circle'} onClick={onMinus} size={'sm'} title={'Zoom out'} />
```
Elfo404, line 45: according to https://developers.grafana.com/ui/latest/index.html?path=/docs/buttons-button--simple#secondary:

> When there is no main important action on a given page, all `Button` components should use the secondary style.

```suggestion
      <Button icon="plus-circle" onClick={onPlus} size="sm" title="Zoom in" variant="secondary" />
      <Button
        className={styles.minusButton}
        icon="minus-circle"
        onClick={onMinus}
        size="sm"
        title="Zoom out"
        variant="secondary"
      />```

File: packages/grafana-ui/src/components/NodeGraph/ViewControls.tsx [hunk 12]
```diff
@@ -0,0 +1,76 @@
...
+
+  scale: css`
+    font-size: 10px;
+    color: ${theme.colors.textFaint};
+  `,
+
+  scrollHelp: css`
+    font-size: 8px;
+    color: ${theme.colors.textFaint};
+  `,
```
Elfo404, line 20: maybe use theme values for font-size?
```suggestion
  scale: css`
    font-size: ${theme.typography.size.sm};
    color: ${theme.colors.textFaint};
  `,

  scrollHelp: css`
    font-size: ${theme.typography.size.xs};
    color: ${theme.colors.textFaint};
  `,
```
Before:
Screenshot 2021-01-19 at 10 22 17
After:
Screenshot 2021-01-19 at 10 21 31

packages/grafana-ui/src/components/NodeGraph/ViewControls.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState } from 'react';
+import { Button } from '../Button';
+import { stylesFactory, useTheme } from '../../themes';
+import { GrafanaTheme } from '@grafana/data';
+import { css } from 'emotion';
+
+const getStyles = stylesFactory((theme: GrafanaTheme) => ({
+  minusButton: css`
+    margin-left: 2px;
+  `,
```
Elfo404, line 10: maybe use spacing value from the theme (it's 2px)
```suggestion
  minusButton: css`
    margin-left: ${theme.spacing.xxs};
  `,
```

packages/grafana-ui/src/components/NodeGraph/ViewControls.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { config, onConfigChange, onPlus, onMinus, scale } = props;
+  const [showConfig, setShowConfig] = useState(false);
+  const styles = getStyles(useTheme());
+
+  // For debugging the layout, should be removed here and maybe moved to panel config later on
+  const allowConfiguration = false;
+
+  return (
+    <>
+      <Button icon={'plus-circle'} onClick={onPlus} size={'sm'} title={'Zoom in'} />
```
Elfo404, line 44: One thing that confused me a bit was that when reaching the maximum zoom in/out, clicking the buttons won't do anything. Definitely not blocking, but would be nice to disable them when reaching max/min scale value

packages/grafana-ui/src/components/NodeGraph/ViewControls.tsx
Elfo404: also, maybe we could use `HorizontalGroup` which will give elements a predefined (and consistent) spacing without the need of the custom spacing for the minus button:
```ts
      <HorizontalGroup spacing="xs">
        <Button icon="plus-circle" onClick={onPlus} size="sm" title="Zoom in" variant="secondary" />
        <Button icon="minus-circle" onClick={onMinus} size="sm" title="Zoom out" variant="secondary" />
      </HorizontalGroup>
```

Screenshot 2021-01-19 at 10 34 02

packages/grafana-ui/src/components/NodeGraph/NodeGraph.tsx
aocenas: So do you propose to use the red warning for the time being? I wanted to make this not too flashy, it's our problem and I don't wan't people to confuse it with any other error. Also made it smaller because of that don't wan't it to take more space than needed.

Elfo404: I think we should make the `severity="warning"` different from the `severity="error"` alert, but i'd rather be consistent with Grafana general look and feel than implementing custom UI components/patterns that are different from what is found everywhere else. So maybe I would push forward for a change in @grafana/ui (since we actually need it) than implementing a custom alert notification.

aocenas: Yeah I agree but for the sake of this particular PR let's keep it as it is and we can change once the alert styles change

PR #29722, opened 2020-12-09
API: add Status() to RedirectResponse

AgnesToulet:
This is an update to `RedirectResponse` to meet the updated `Response` interface (updated in this PR: https://github.com/grafana/grafana/pull/29642).

The status code returned by the new `Status` function is the same as the default one used by the `Redirect` function of macaron as the `RedirectResponse` struct does not allow to change it.

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -145,6 +145,11 @@ func (r *RedirectResponse) WriteTo(ctx *models.ReqContext) {
 	ctx.Redirect(r.location)
 }
 
+// Status gets the response's status.
+func (*RedirectResponse) Status() int {
+	return http.StatusFound
+}
+
 func Redirect(location string) *RedirectResponse {
 	return &RedirectResponse{location: location}
 }


pkg/api/common.go
```diff
@@ -145,6 +145,11 @@ func (r *RedirectResponse) WriteTo(ctx *models.ReqContext) {
 	ctx.Redirect(r.location)
 }
 
+// Status gets the response's status.
+func (*RedirectResponse) Status() int {
```
aknuds1, line 149: Will this method ever be called? Just wondering, since any build failure can also be solved by adding `Response` on line 141.

aknuds1 (approved):

pkg/api/common.go
AgnesToulet: It seems that this function was added in the interface only for testing purposes and is only used in the `librarypanels` service so no it won't be called rn. But it could be useful for future tests. 

I don't see how adding `Response` on line 141 will avoid us to add this function.

aknuds1: @AgnesToulet If you add `Response` on line 141, it will make `RedirectResponse` have default implementations of all the `Response` methods. The only technical difference would be that calling `Status()` on a `RedirectResponse` would fail. I've used this technique to implement large interfaces, where you don't want to write all the methods.

PR #29735, opened 2020-12-09
Docs: added version note for rename by regex transformation.

mckn:
I was a bit quick in merging the following #29281. Adding missing version note in docs.

### DIFF
--- a/docs/sources/panels/transformations/types-options.md
+++ b/docs/sources/panels/transformations/types-options.md
@@ -402,6 +402,8 @@ Conditions that are invalid or incompletely configured are ignored.
 
 ## Rename by regex
 
+> **Note:** This transformation is only available in Grafana 7.4+.
+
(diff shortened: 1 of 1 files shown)


oddlittlebird (approved):
Perfecto

docs/sources/panels/transformations/types-options.md
```diff
@@ -402,6 +402,8 @@ Conditions that are invalid or incompletely configured are ignored.
 
 ## Rename by regex
 
+> **Note:** This transformation is only available in Grafana 7.4+.
```
osg-grafana, line 405: Small nit: please remove "only".

mckn: Great feedback!

But I have already merged the PR and that line is repeated on multiple places in the docs to I guess you should open an issue regarding this so it can be updated on all places.

osg-grafana: +1

Issue #7794, opened 2017-03-09
Elasticsearch: Support template variables as the Size of Group by Terms


alexandrul:
Currently the template variable is parsed as an int and I'm getting ```illegal_argument_exception```: ```[terms] size doesn't support values of type: VALUE_NULL```

![group_by__size](https://cloud.githubusercontent.com/assets/663809/23757650/197d552c-04f0-11e7-95b7-6f02907f4a9a.png)

Grafana v4.3.0-pre1 (commit: 2f5814d51)
ES 5.x
Windows 7 x64

alexandrul:
This is the int parsing part:
```b.terms.size = 0 === parseInt(a.settings.size, 10) ? 500 : parseInt(a.settings.size, 10)```

This enhancement seems to be working with an extra step: if ```a.settings.size``` is not a number then ```b.terms.size = a.settings.size```

Of course, there should be some sanity checks (it makes no sense to use a multi-value template var here).

alexandrul:
up

apetzo:
I ran into the same issue just now with elastic and Grafana 7.0

simianhacker:
@Elfo404 Should every setting be overridable by variables? The only variables that seem to show up in the `options.scopedVars` is `$_interval` and `$_intervalMS`. I setup a 'constant' variable which only seems to be available via the `templateSrv.getVariables()` on the data source.

Elfo404:
@simianhacker ideally any user input should support variables when it makes sense. I started working on a draft some time ago in #29739 which seems to work quite okay-ish, but it needs a bit more investigation.

PR #29739, opened 2020-12-10
Elasticsearch: Add generic support for template variables

### LINKED ISSUE #8339: ElasticSearch: Permit "field" to be parsed to a variable
Grafana version: 4.3.0-pre1
Datasource: Elasticsearch
OS: CentOS 7
If is tried to use "field": "$variable" in query of Variables, the variable is not resolved and the query for the elasticsearch.

Tryied:
Create a variable named $test, type Custom with value "AAA,BBB"
Select $test ='AAA' in a dashboard.
Set another variable named named $test1 with query = {"find": "terms", "field": "$test"}

Result: 
{"find": "terms", "field": "$test"}

Espected result:
{"find": "terms", "field": "AAA"}

Elfo404:
Adds generic variable interpolation logic to elasticsearch query model, allowing every string field to contain template variables

Fixes #8339 #7794

Not really ready for review as doesn't cover all use-cases, opening this as a starting point to tackle as much variable related issues as possible.

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/datasource.test.ts
+++ b/public/app/plugins/datasource/elasticsearch/datasource.test.ts
@@ -865,7 +865,7 @@ describe('ElasticDatasource', function(this: any) {
     expect((interpolatedQuery.bucketAggs![0] as Filters).settings!.filters![0].query).toBe('resolvedVariable');
   });
 
-  it('should correctly handle empty query strings', () => {
+  it.only('should correctly handle empty query strings', () => {
     const query: ElasticsearchQuery = {
       refId: 'A',
       bucketAggs: [{ type: 'filters', settings: { filters: [{ query: '', label: '' }] }, id: '1' }],
--- a/public/app/plugins/datasource/elasticsearch/datasource.ts
+++ b/public/app/plugins/datasource/elasticsearch/datasource.ts
@@ -320,27 +320,54 @@ export class ElasticDatasource extends DataSourceApi<ElasticsearchQuery, Elastic
     return this.templateSrv.replace(queryString, scopedVars, 'lucene') || '*';
   }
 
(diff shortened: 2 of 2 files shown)


public/app/plugins/datasource/elasticsearch/datasource.test.ts
```diff
@@ -865,7 +865,7 @@ describe('ElasticDatasource', function(this: any) {
     expect((interpolatedQuery.bucketAggs![0] as Filters).settings!.filters![0].query).toBe('resolvedVariable');
   });
 
-  it('should correctly handle empty query strings', () => {
+  it.only('should correctly handle empty query strings', () => {
```
torkelo, line 868: the pre-commit hook should have caught this one

torkelo (commented):
wonder what is more efficient here, render all queries to a json string, then a single call to templateSrv.replace + parse back to object? the query need some special logic (default to lucene escaping)

public/app/plugins/datasource/elasticsearch/datasource.test.ts
Elfo404: yeah, I `--no-verify`'ed it just to push it and pick this up later, there are a few other things I want to add to this commit.

Elfo404:
> wonder what is more efficient here, render all queries to a json string, then a single call to templateSrv.replace + parse back to object? the query need some special logic (default to lucene escaping)

@torkelo That's a very good point. I'll also give that approach a shot, I guess the special use-cases (like the lucene queries) might be handled with some regex.

simianhacker:
@Elfo404 I started looking at #7794 and came across the Lucene escaping. It looks like that's done early enough that the wholesale approach @torkelo would work. We sort of do that now but only with the `options.scopedVars` which seems to only include the `$__interval` variables. We would need to merge that with the list of variables from `templateSrv.getVariables()`. The only caveat is, for the most part, Elasticsearch isn't too picky about the variable types (string vs number) with a few exceptions.

PR #29740, opened 2020-12-10
Middleware: Add CSP support

### LINKED ISSUE #16655: CSP Compliance

aknuds1:
PR to add support to middleware for adding [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) header.

Extracted partially from https://github.com/grafana/grafana/pull/25837.

Fixes #16655.

TODO: 

- [x] Review the CSP template string, to see if it makes sense.
- [x] Check what remains wrt. #6820 also, after merging this.

As per @torkelo comment:

- [x] Check if changes to webpack chunk loading and plugins loading are required


torkelo:
I know this is draft, tries to test the plugin loading but was unable to get the app to load. Let me know when this is ready and I can check the plugin loading

aknuds1:
@torkelo I will. I am currently investigating a strange bug, which causes the nonce not to be transmitted.

torkelo (approved):
Looks good. Tested with many external plugins and seems to work.

aknuds1:
That's awesome @torkelo - we should probably review the default CSP template string though, to make sure it makes sense.

aknuds1:
@sakjur Do you want to review the CSP header template here?

pkg/api/dtos/index.go
```diff
@@ -25,6 +25,8 @@ type IndexViewData struct {
 	AppleTouchIcon          template.URL
 	AppTitle                string
 	Sentry                  *setting.Sentry
+	// Nonce is a cryptographic identifier for use with Content Security Policy.
+	Nonce string
```
wbrowne, line 29: Weird formatting here, no?

aknuds1: @wbrowne How is it weird? The file gets auto-formatted by goimports.

aknuds1: Ah, you're referring to the field alignment. Probably unaligned because of the comment above.

pkg/api/login_test.go
```diff
@@ -131,10 +140,10 @@ func TestLoginErrorCookieAPIEndpoint(t *testing.T) {
 	}
 	sc.m.Get(sc.url, sc.defaultHandler)
 	sc.fakeReqNoAssertionsWithCookie("GET", sc.url, cookie).exec()
-	assert.Equal(t, sc.resp.Code, 200)
+	require.Equal(t, 200, sc.resp.Code)
```
ying-jeanne, line 143: does this mean we always want to use require?
`Package require implements the same assertions as to the `assert` package but stops test execution when a test fails.`
Or we just want to use require when checking the null pointer then stop testing when the pointer is null so no point to continue?

aknuds1: @ying-jeanne It doesn't mean we always want to use `require`, but my thinking here is that the test should stop if an error response is received, because the remaining checks don't make any sense.

conf/defaults.ini
```diff
@@ -227,6 +227,13 @@ x_content_type_options = true
 # when they detect reflected cross-site scripting (XSS) attacks.
 x_xss_protection = true
 
+# Enable adding the Content-Security-Policy header to your requests.
+# CSP allows to control resources the user agent is allowed to load and helps prevent XSS attacks.
+content_security_policy = false
+
+# Set Content Security Policy template used when adding the Content-Security-Policy header to your requests.
+# $NONCE in the template includes a random nonce.
+content_security_policy_template = """script-src 'unsafe-eval' 'strict-dynamic' $NONCE;object-src 'none';font-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data:;base-uri 'self';connect-src 'self' grafana.com;manifest-src 'self';media-src 'none';block-all-mixed-content;"""
```
stevenvachon, line 236: * `img-src 'self' data:` looks like data URIs only, which will block google cloud images within plugin readmes
* maybe add `form-action 'self'` to prevent successful XSS hacks from working

aknuds1: @torkelo What do you think of @stevenvachon's feedback?

torkelo: what does base-uri mean? and will it also block if you add data source with browser access (ie browser will directly access the data source)

aknuds1:
Looking into the backend test failure.

conf/defaults.ini
aknuds1: I will have to look up base-uri @torkelo, I'm not really familiar with the Content Security Policy directives. They were copied from the original PR.

pkg/api/login_test.go
aknuds1: @ying-jeanne If `require.Equal` fails here, the execution of that particular test will end, yes. The rest of the test won't make any sense anyhow.

conf/defaults.ini
aknuds1: @torkelo [Here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri)'s the base-uri documentation.

The `<base>` element seems to control the base URI for all relative URIs in the document, so `base-uri 'self'` probably means that all relative URIs have to have the document origin as their base. It probably makes sense.

aknuds1: > * `img-src 'self' data:` looks like data URIs only, which will block google cloud images within plugin readmes

@stevenvachon I think `img-src 'self' data:` means both URIs with same origin as the document and data URIs. Regarding Google Cloud, yeah does look that way.

aknuds1: @stevenvachon I added `form-action 'self'`, thanks!

aknuds1:
I removed the `block-all-mixed-content` directive since it's marked as obsolete on MSDN.

aknuds1:
@Sytten Does the default CSP header template here make sense to you?

Sytten:
@aknuds1 It's a good start for sure and I really appreciate the work being done, but we also have to acknowledge that `script-src 'unsafe-eval'` reduces the significantly the protection offered by the CSP against XSS.
I think it would also make sense to put the default to 'self' in case we missed one rule (I didn't check if they were all there).

aknuds1:
> @aknuds1 It's a good start for sure and I really appreciate the work being done, but we also have to acknowledge that `script-src 'unsafe-eval'` reduces the significantly the protection offered by the CSP against XSS.
> I think it would also make sense to put the default to 'self' in case we missed one rule (I didn't check if they were all there).

@Sytten According to the [original PR](https://github.com/grafana/grafana/pull/25837), `script-src 'unsafe-eval'` is necessary to keep AngularJS working. Where are you saying we should add 'self'? I'm not that familiar with CSP.

iOrcohen:
> > @aknuds1 It's a good start for sure and I really appreciate the work being done, but we also have to acknowledge that `script-src 'unsafe-eval'` reduces the significantly the protection offered by the CSP against XSS.
> > I think it would also make sense to put the default to 'self' in case we missed one rule (I didn't check if they were all there).
> 
> @Sytten According to the [original PR](https://github.com/grafana/grafana/pull/25837), `script-src 'unsafe-eval'` is necessary to keep AngularJS working. Where are you saying we should add 'self'? I'm not that familiar with CSP.
> 
> @torkelo @iOrcohen Can you confirm that `script-src 'unsafe-eval'` is necessary for AngularJS to work?

Yes, AngularJS using eval function so it's required.

aknuds1:
@Sytten If you see @iOrcohen, it does look as if we need to keep `script-src 'unsafe-eval'` until AngularJS is replaced.

aknuds1:
Merging this. If you have a suggestion for a tweak of the default template @Sytten, please make a PR :)

PR #29741, opened 2020-12-10
PanelLibrary: Adds delete Api

hugohaggmark:
This PR adds `DELETE /api/library-panels/:panelId` api for Library/Shared/Reusable Panels

Relates #29610

### DIFF
--- a/pkg/services/librarypanels/api.go
+++ b/pkg/services/librarypanels/api.go
@@ -18,6 +18,7 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
 
 	lps.RouteRegister.Group("/api/library-panels", func(libraryPanels routing.RouteRegister) {
 		libraryPanels.Post("/", middleware.ReqSignedIn, binding.Bind(addLibraryPanelCommand{}), api.Wrap(lps.createHandler))
+		libraryPanels.Delete("/:panelId", middleware.ReqSignedIn, api.Wrap(lps.deleteHandler))
 	})
 }
 
@@ -33,4 +34,18 @@ func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibra
 	}
 
 	return api.JSON(200, util.DynMap{"panel": panel})
+}
+
+// deleteHandler handles DELETE /api/library-panels/:panelId.
+func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response {
+	err := lps.deleteLibraryPanel(c.ParamsInt64(":panelId"))
+
+	if err != nil {
+		if errors.Is(err, errLibraryPanelNotFound) {
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to delete library panel", err)
+	}
+
+	return api.Empty(204)
 }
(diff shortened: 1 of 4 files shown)


pkg/services/librarypanels/api.go
```diff
@@ -34,3 +35,17 @@ func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibra
...
+	err := lps.deleteLibraryPanel(c.ParamsInt64(":panelId"))
+
+	if err != nil {
+		if errors.Is(err, errLibraryPanelNotFound) {
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to delete library panel", err)
+	}
+
+	return api.Empty(204)
```
hugohaggmark, line 50: Wasn't sure if we usually return 200 or 204.

pkg/services/librarypanels/models.go
```diff
@@ -22,8 +22,10 @@ type LibraryPanel struct {
 }
 
 var (
-	// errLibraryPanelAlreadyAdded is an error when you add a library panel that already exists.
-	errLibraryPanelAlreadyAdded = fmt.Errorf("library panel with that title already exists")
+	// errLibraryPanelAlreadyAdded is an error for when the user tries to add a library panel that already exists.
+	errLibraryPanelAlreadyAdded = errors.New("library panel with that title already exists")
+	// errLibraryPanelNotFound is an error for when a library panel can't be found.
+	errLibraryPanelNotFound = errors.New("library panel could not be found")
```
hugohaggmark, line 28: Changed to errors.New but not sure that is more correct.

papagian: Since you don't need any special format, I think `errors.New()` it's fine.
`fmt.Errorf()` allows to format the error message using variables or wrap another error.

papagian (commented):
LGTM but I would change the response status to 200.

pkg/services/librarypanels/database.go
```diff
@@ -39,3 +39,21 @@ func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd add
...
+
+func (lps *LibraryPanelService) deleteLibraryPanel(panelID int64) error {
+	err := lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		if res, err := session.Query("SELECT 1 from library_panel WHERE id=?", panelID); err != nil {
+			return err
+		} else if len(res) != 1 {
+			return errLibraryPanelNotFound
+		}
+
+		if _, err := session.Exec("DELETE FROM library_panel WHERE id = ?", panelID); err != nil {
```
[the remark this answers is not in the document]
hugohaggmark: Thank you, that sounds like a better approach. Would you still return `404` if  `RowsAffected` is zero?

pkg/services/librarypanels/api.go
hugohaggmark: That's a great suggestion, I'll change and use that instead.

pkg/services/librarypanels/models.go
hugohaggmark: Aha, so that's the use case. Then this is better right now, thank you for clarifying it.

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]
papagian: yes, it would be clearer that way.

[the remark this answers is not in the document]
hugohaggmark: Perfect, then I'll redo this as you suggested.

hugohaggmark:
I just realized I missed OrgID, will update soon.

kylebrandt (approved):

PR #29751, opened 2020-12-10
Chore: Disable default golangci-lint filter

### LINKED ISSUE #29717: Disable golangci-lint default filter
**What happened**:
The golangci-lint default filter should be disabled, in order to f.ex. receive all gosec warnings. Then we'll need to selectively disable the warnings that don't make sense.

aknuds1:
Disable the default golangci-lint filter, in order to get all gosec warnings. As requested by @bergquist. I disable stylistic warnings, we can deal with them later (or continue ignoring them).

Most of the things fixed are lacking closes and also checking errors from closes.

Fixes #29717.

**Reviewer notes**:
The PR got huge due to all the warnings resulting; I could split into several ones if you want.


aknuds1:
Ugh, something breaks when testing manually. Might have to split into more PRs.

pkg/services/contexthandler/authproxy/authproxy.go
```diff
@@ -318,18 +328,20 @@ func (auth *AuthProxy) GetSignedInUser(userID int64) (*models.SignedInUser, erro
...
-	key := auth.getKey()
+	key, err := auth.getKey()
+	if err != nil {
+		return err
+	}
 
 	// Check if user already in cache
-	userID, _ := auth.remoteCache.Get(key)
-	if userID != nil {
+	userID, err := auth.remoteCache.Get(key)
```
wbrowne, line 337: Shouldn't we return if `err != nil`?

aknuds1: @wbrowne In this particular case, we want to return if the user is already in the cache, i.e. if no error is returned and `userID` is non-nil. The old code didn't take into account the error.

wbrowne: But in the case where there is an error?

aknuds1: @wbrowne As I think the original logic is, it's fine to ignore the cache if an error occurs looking it up.

wbrowne (approved):

PR #29754, opened 2020-12-10
Dashboards: Allow string manipulation as advanced variable format option

### LINKED ISSUE #8259: Feature Request: template variable string manipulations
It would be nice if Grafana template variables could have some string manipulations applied. This could be useful to be able to refer to data about the same thing from various datasources from a variable, where that thing is named slightly differently for technical reasons. For instance, if you have a website that you refer to by the name "some_site" in your systems, but you have an AWS S3 bucket for it where you can't use underscores, you could transform that name to 'some-site', or 'somesite' instead.

The immediately obvious ideas for what kind of manipulations to support would be the various variable substitutions that bash supports. In bash I would do the above example by using `${var//_/-}` or `${var//_}`.

rscott231:
**What this PR does / why we need it**: This PR implements regex-based template variable manipulation. There are a variety of use cases that make this valuable, as noted in #8259.

Fixes #8259 

### DIFF
--- a/docs/sources/variables/advanced-variable-format-options.md
+++ b/docs/sources/variables/advanced-variable-format-options.md
@@ -149,3 +149,21 @@ servers = ["test1", "test2"]
 String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
 ```
+
+## Replace
+
+Formats single- and multi-valued variables by replacing text. Accepts one to three arguments and allows escaping of colons using backslashes. When called with two or more arguments, the first 2 arguments follow the semantics of `String.prototype.replace()`. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
+
+```bash
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:upper}'
+Interpolation result: "VAL1,VAL2"
+
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:values?:val}'
+Interpolation result: "val1,val2"
+
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:values?:val:upper}'
+Interpolation result: "VAL1,VAL2"
+```
--- a/docs/sources/variables/advanced-variable-format-options.md
+++ b/docs/sources/variables/advanced-variable-format-options.md
@@ -152,7 +152,7 @@ Interpolation result: "test1 + test2"
 
 ## Replace
 
-Formats single- and multi-valued variables by replacing text. Accepts one to three arguments and allows escaping of colons using backslashes. When called with two or more arguments, the first 2 arguments follow the semantics of `String.prototype.replace()`. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
+Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
 
 ```bash
 servers = ["value1", "values2"]
--- a/docs/sources/variables/advanced-variable-format-options.md
+++ b/docs/sources/variables/advanced-variable-format-options.md
@@ -155,9 +155,9 @@ Interpolation result: "test1 + test2"
 Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
 
 ```bash
-servers = ["value1", "values2"]
+servers = ["value1", "value2"]
 String to interpolate: '${servers:replace:upper}'
-Interpolation result: "VAL1,VAL2"
+Interpolation result: "VALUE1,VALUE2"
 
 servers = ["value1", "values2"]
 String to interpolate: '${servers:replace:values?:val}'
--- a/public/app/features/templating/formatRegistry.ts
+++ b/public/app/features/templating/formatRegistry.ts
@@ -1,6 +1,6 @@
 import kbn from 'app/core/utils/kbn';
-import { Registry, RegistryItem, VariableModel, textUtil, dateTime } from '@grafana/data';
-import { map, isArray, replace } from 'lodash';
+import { dateTime, Registry, RegistryItem, textUtil, VariableModel } from '@grafana/data';
+import { isArray, map, replace } from 'lodash';
 import { formatVariableLabel } from '../variables/shared/formatVariable';
 
 export interface FormatOptions {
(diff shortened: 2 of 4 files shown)


achatterjee-grafana:
@oddlittlebird, looking at this now.

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -148,4 +148,22 @@ Formats single- and multi-valued variables into their text representation. For a
 servers = ["test1", "test2"]
 String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
-```
\ No newline at end of file
+```
+
+## Replace
+
+Formats single- and multi-valued variables by replacing text. Accepts one to three arguments and allows escaping of colons using backslashes. When called with two or more arguments, the first 2 arguments follow the semantics of `String.prototype.replace()`. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
```
achatterjee-grafana, line 155:
```suggestion
Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
```

achatterjee-grafana (commented):
Added copy-edit suggestions.

achatterjee-grafana (approved):
Doc looks OK, needs Dev review & feedback.

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -148,4 +148,22 @@ Formats single- and multi-valued variables into their text representation. For a
...
+```
+
+## Replace
+
+Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
+
+```bash
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:upper}'
+Interpolation result: "VAL1,VAL2"
```
mckn, line 160: Shouldn't this be as described below? Or am I understanding this incorrectly?

```suggestion
Interpolation result: "VALUE1,VALUE2"
```

public/app/features/templating/template_srv.ts
```diff
@@ -129,8 +129,8 @@ export class TemplateSrv implements BaseTemplateSrv {
       format = 'glob';
     }
 
-    // some formats have arguments that come after ':' character
-    let args = format.split(':');
+    // some formats have arguments that come after ':' character but allow escaping of the : with a backslash for regex
+    let args = format.split(/(?<!\\):/);
```
mckn, line 133: hmm, is this breaking backward compability?

public/app/features/templating/formatRegistry.ts
```diff
@@ -241,3 +248,36 @@ function encodeURIComponentStrict(str: string) {
     );
   });
 }
+
+function replaceText({ value, args }: FormatOptions) {
```
mckn, line 252: I would like to refactor this code a bit to make it easier to reason about. I'm not sure how to do that the best way but I needed to walk through this code a couple of times before getting all the different scenarios.

mckn:
Thanks for the contribution and I can see that this would be a great addition to Grafana. I am a bit worried about potentially breaking backward compatibility and how this extension of the format can affect variables in existing dashboards.

Also, I think we can simplify the `replaceText` functionality to be a bit easier to reason about. But overall great work!

docs/sources/variables/advanced-variable-format-options.md
rscot231: I pushed a commit to resolve this.

public/app/features/templating/template_srv.ts
rscot231: I don't _think_ it is, but I'm not 100% sure. I didn't see any cases in the previously existing format functions where preceding a colon with a backslash seemed likely or like a valid value. There are definitely no test cases covering it. I'm happy to consider other suggestions though.

The main reason for doing this is to support colons within the regex and/or the replacement. Without the escaping, the split function would treat them as additional args.

public/app/features/templating/formatRegistry.ts
rscot231: @mckn I pushed one more commit to try to improve the readability. It's quite a bit longer, but things have explicit names now and the various length permutations are more explicit cases. Let me know what you think.

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -148,4 +148,22 @@ Formats single- and multi-valued variables into their text representation. For a
...
+servers = ["value1", "value2"]
+String to interpolate: '${servers:replace:upper}'
+Interpolation result: "VALUE1,VALUE2"
+
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:values?:val}'
+Interpolation result: "val1,val2"
+
+servers = ["value1", "values2"]
+String to interpolate: '${servers:replace:values?:val:upper}'
```
ocardia, line 167: I just run upon this PR coming from an Issue. If my understanding is correct, we should change the documentation as follows: 🙏 

```suggestion
servers = ["value1", "value2"]
String to interpolate: '${servers:replace:value?:val}'
Interpolation result: "val1,val2"

servers = ["value1", "value2"]
String to interpolate: '${servers:replace:value?:val:upper}'
```

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -148,4 +148,22 @@ Formats single- and multi-valued variables into their text representation. For a
 servers = ["test1", "test2"]
 String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
-```
\ No newline at end of file
+```
+
+## Replace
+
+Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the rightmost argument should be either `upper` or `lower`, which will change the case of the result string.
```
achatterjee-grafana, line 155:
```suggestion
Replaces text with single and multi-valued variables. It accepts one to three arguments and allows the escaping of colons using backslashes. When called with two or more arguments, the first two arguments follow the semantics of `String.prototype.replace(. ) `. When called with an odd number of arguments, the argument on the right should be either `upper` or `lower`, which will change the case of the result string.
```

achatterjee-grafana (commented):
Added one copy-edit suggestion.

rscot231:
Any additional thoughts on this? We are currently running on a fork that includes it and it would be great to use a supported version instead.

torkelo:
We are just unsure, scary thing to add & support for all eternity. Is this future-proof etc, is this the syntax & mechanics we want. Very hard to know, which is why this is taking time...

rscot231:
Understood. Is anything I can do to make it less scary. I'm not thrilled with the syntax myself but I was trying to play within the syntax that already exists for variable formatting. I'm be happy to take a different approach if you think that would be better.

dprokop:
@rscot231 thanks again for your proposal. We will think about an aternative approach, and get back to you when we came up with idea. But for now I'm gonna close this one.

ningyougang:
@dprokop , have any new idea about this issue?

tsuna:
Found this PR with excitement looking for a way to lowercase a variable during string interpolation in a PromQL query but unfortunately this didn't go anywhere. :(

PR #29757, opened 2020-12-10
Link suppliers: getLinks API update

dprokop:
This PR aims to replace the link supplier's `getLinks` function `getLinks(scopedVars?: any): Array<LinkModel<T>>;` with `getLinks(replaceVariables?: InterpolateFunction): Array<LinkModel<T>>`. Has plenty of ripple effect but hope it's fairly digestable:)

### DIFF
--- a/packages/grafana-data/src/types/ScopedVars.ts
+++ b/packages/grafana-data/src/types/ScopedVars.ts
@@ -4,6 +4,4 @@ export interface ScopedVar<T = any> {
   [key: string]: any;
 }
 
-export interface ScopedVars {
-  [key: string]: ScopedVar;
-}
+export interface ScopedVars extends Record<string, ScopedVar> {}
--- a/packages/grafana-data/src/types/dataLink.ts
+++ b/packages/grafana-data/src/types/dataLink.ts
@@ -1,12 +1,12 @@
-import { ScopedVars } from './ScopedVars';
 import { DataQuery } from './datasource';
+import { InterpolateFunction } from './panel';
 
 /**
  * Callback info for DataLink click events
  */
 export interface DataLinkClickEvent<T = any> {
   origin: T;
-  scopedVars?: ScopedVars;
+  replaceVariables: InterpolateFunction | undefined;
   e?: any; // mouse|react event
 }
 
@@ -67,7 +67,7 @@ export interface LinkModel<T = any> {
  * TODO: ScopedVars in in GrafanaUI package!
  */
 export interface LinkModelSupplier<T extends object> {
-  getLinks(scopedVars?: any): Array<LinkModel<T>>;
+  getLinks(replaceVariables?: InterpolateFunction): Array<LinkModel<T>>;
 }
 
 export enum VariableOrigin {
--- a/public/app/features/dashboard/state/PanelModel.ts
+++ b/public/app/features/dashboard/state/PanelModel.ts
@@ -30,7 +30,7 @@ import config from 'app/core/config';
 import { PanelQueryRunner } from '../../query/state/PanelQueryRunner';
 import { PanelOptionsChangedEvent, PanelQueriesChangedEvent, PanelTransformationsChangedEvent } from 'app/types/events';
 import { getTimeSrv } from '../services/TimeSrv';
-import { getAllVariableValuesForUrl } from '../../variables/utils';
+import { getAllVariableValuesForUrl } from '../../variables/getAllVariableValuesForUrl';
 
 export interface GridPos {
   x: number;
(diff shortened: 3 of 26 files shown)


packages/grafana-data/src/types/ScopedVars.ts
```diff
@@ -4,6 +4,4 @@ export interface ScopedVar<T = any> {
   [key: string]: any;
 }
 
-export interface ScopedVars {
-  [key: string]: ScopedVar;
-}
+export interface ScopedVars extends Record<string, ScopedVar> {}
```
hugohaggmark, line 7: Nice!

packages/grafana-data/src/types/dataLink.ts
```diff
@@ -1,12 +1,12 @@
...
 import { DataQuery } from './datasource';
+import { InterpolateFunction } from './panel';
 
 /**
  * Callback info for DataLink click events
  */
 export interface DataLinkClickEvent<T = any> {
   origin: T;
-  scopedVars?: ScopedVars;
+  replaceVariables: InterpolateFunction | undefined;
```
hugohaggmark, line 9: nit: Out of curiosity why not `replaceVariables?` instead of adding  `| undefined`

public/app/features/panel/panellinks/link_srv.ts
```diff
@@ -253,7 +255,7 @@ export const getPanelOptionsVariableSuggestions = (plugin: PanelPlugin, data?: D
 };
 
 export interface LinkService {
-  getDataLinkUIModel: <T>(link: DataLink, scopedVars: ScopedVars | undefined, origin: T) => LinkModel<T>;
+  getDataLinkUIModel: <T>(link: DataLink, scopedVars: InterpolateFunction | undefined, origin: T) => LinkModel<T>;
```
hugohaggmark, line 258: nit: small naming change?
```suggestion
  getDataLinkUIModel: <T>(link: DataLink, replaceVariables: InterpolateFunction | undefined, origin: T) => LinkModel<T>;
```

public/app/features/templating/template_srv.ts
```diff
@@ -309,22 +308,6 @@ export class TemplateSrv implements BaseTemplateSrv {
...
-        params['var-' + variable.name] = scopedVars[variable.name].value;
-      } else {
-        if (variable.skipUrlSync) {
-          return;
-        }
-        params['var-' + variable.name] = variableAdapters.get(variable.type).getValueForUrl(variable);
-      }
-    });
-  };
-
```
hugohaggmark, line 327: nit: perhaps not a biggie as this is probably not used by anyone. But maybe we should deprecate this function and call the new util function `getAllVariableValuesForUrl` here instead?

hugohaggmark (commented):
This looks very promising but I couldn't take it for a spin because of hard crash when loading dashboards or adding panel to dashboard.
image

public/app/features/dashboard/state/PanelModel.ts
```diff
@@ -498,9 +502,27 @@ export class PanelModel implements DataConfigSource {
...
+      ...vars,
+      [DataLinkBuiltInVars.keepTime]: {
+        text: timeRangeUrl,
+        value: timeRangeUrl,
+      },
+      [DataLinkBuiltInVars.includeVars]: {
+        text: variablesQuery,
+        value: variablesQuery,
+      },
+    };
```
torkelo, line 524: replace variables is called potentially hundreds of times, would be nice to find ways to memoize and optimize parts of this

public/app/features/templating/template_srv.ts
torkelo: don't think we need that, this was not part of any interface we expose to plugins, and do not see any plugins using this .

torkelo (approved):
Think this looks like a good refactor, some concert over performance where we build expensive urls values for every call to interpolate. Is there a way to do that lazily / on demans? Maybe if we made scopedVars we pass to templateSrv.replace a proxy ?

public/app/features/dashboard/state/PanelModel.ts
dprokop: yups, will do a followup PR

public/app/features/templating/template_srv.ts
dprokop: Removed

dprokop:
> This looks very promising but I couldn't take it for a spin because of hard crash when loading dashboards or adding panel to dashboard.

@hugohaggmark should be OK now. Circular dependencies FTW

aocenas:
Nice yeah I think this makes sense, shouldn't affect much right now but should be easier to simplify some code in explore later on.

public/app/features/dashboard/state/PanelModel.ts
```diff
@@ -498,9 +502,26 @@ export class PanelModel implements DataConfigSource {
...
```
torkelo, line 523: We should explore a way to do this lazily, but could be done in a separate PR to keep scope down for this

hugohaggmark (approved):
Looks great, great work on this ⭐

PR #29761, opened 2020-12-10
Elasticsearch: Sort results by index order as well as @timestamp

### LINKED ISSUE #29700: Logs sometimes show in wrong order with Log panel + Elasticsearch datasource because we cannot customize the sort :(
**What happened**:

I get my logs in the wrong order for log lines that occur in the same millisecond because Grafana assumes logs are ordered by timestamp.

I should be able to define a sort field to sort my logs where it's possible to define an order for documents that have the same millisecond timestamp.

Currently Grafana assumes all log files are ordered by time - if you look at the code:

[query_builder.ts](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/elasticsearch/query_builder.ts#L132)
``` javascript
documentQuery(query: any, size: number) {
    query.size = size;
    query.sort = {};
    query.sort[this.timeField] = { order: 'desc', unmapped_type: 'boolean' };

    // fields field not supported on ES 5.x
    if (this.esVersion < 5) {
      query.fields = ['*', '_source'];
    }

    query.script_fields = {};
    return query;
  }
```

Imagine this log file:

``` bash
2020-12-08 23:12:00:00 INFO Hello World 1
2020-12-08 23:12:00:00 INFO Hello World 2
2020-12-08 23:12:00:00 INFO Hello World 3
```

Because in Grafana we hardcode the sort to be by timestamp I end up seeing this in the log panel:

``` bash
2020-12-08 23:12:00:00 INFO Hello World 1
2020-12-08 23:12:00:00 INFO Hello World 3
2020-12-08 23:12:00:00 INFO Hello World 2
```

Now the problem is all of those log lines happened in the same millisecond. 

This isn't a problem when I ingest the data into Elastic because I'm able to add a "lineNumber" field to each Document (where each Document==log line).

So if I was querying my Elasticsearch database I would be able to give a time range and then sort by line number which would give me this:

``` bash
2020-12-08 23:12:00:00 INFO Hello World 1 (lineNumber=1)
2020-12-08 23:12:00:00 INFO Hello World 2 (lineNumber=2)
2020-12-08 23:12:00:00 INFO Hello World 3 (lineNumber=3)
```

This is quite bad because it's critical for us to be able to view the logs in the order things actually happened so this isn't a "nice to have" for us it's a must without this it makes this very unusable/untrustworthy.

Appreciate this might be an unusual issue I am  happy to help resolve this with a PR if someone can give me guiance on what we should be fixing here 😄.

_**If there's  a way I could fix this in my query/elastic then that would be helpful too.**_

* Elasticsearch 7.0
* Grafana 7.2

STEELBADGE:

Logs sometimes show up in the wrong order because Grafana is not respecting the index order of the documents. This was fixed as part of https://github.com/grafana/grafana/pull/28764 but I am picking some of those changes on here because I would like it to be part of a patch release rather than 7.4.

Fixes #29700

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/query_builder.ts
+++ b/public/app/plugins/datasource/elasticsearch/query_builder.ts
@@ -125,7 +125,14 @@ export class ElasticQueryBuilder {
   documentQuery(query: any, size: number) {
     query.size = size;
     query.sort = {};
-    query.sort[this.timeField] = { order: 'desc', unmapped_type: 'boolean' };
+    query.sort = [
+      {
+        [this.timeField]: { order: 'desc', unmapped_type: 'boolean' },
+      },
+      {
+        _doc: { order: 'desc' },
+      },
+    ];
 
     // fields field not supported on ES 5.x
     if (this.esVersion < 5) {
--- a/public/app/plugins/datasource/elasticsearch/specs/query_builder.test.ts
+++ b/public/app/plugins/datasource/elasticsearch/specs/query_builder.test.ts
@@ -224,12 +224,7 @@ describe('ElasticQueryBuilder', () => {
               ],
             },
           },
-          sort: {
-            '@timestamp': {
-              order: 'desc',
-              unmapped_type: 'boolean',
-            },
-          },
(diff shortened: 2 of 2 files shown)


Elfo404 (approved):
LGTM, thanks for the PR!

PR #29762, opened 2020-12-10
QueryEditor: do not auto refresh on every update

ryantxu:
The recent changes to query editors now issue a query with every update -- previously you needed to call `onRunQueries()` explicitly.

This change means that query editors now need to keep state and only call update when the query is ready -- this is a kinda big change for some editors.

That said, many editors have bad auto-refresh behavior and require hitting the refresh link when it should be automatic.  Perhaps we should add a parameter to `onOptionsChange` that would also trigger refresh?  then it would at least be more clear that we are not triggering refesh

### DIFF
--- a/public/app/features/dashboard/components/PanelEditor/PanelEditorQueries.tsx
+++ b/public/app/features/dashboard/components/PanelEditor/PanelEditorQueries.tsx
@@ -56,7 +56,6 @@ export class PanelEditorQueries extends PureComponent<Props, State> {
     panel.hideTimeOverride = options.timeRange?.hide;
     panel.interval = options.minInterval;
     panel.maxDataPoints = options.maxDataPoints;
-    panel.refresh();
 
     this.setState({ options: options });
   };


torkelo:
> This change means that query editors now need to keep state and only call update when the query is ready -- this is a kinda big change for some editors.

What state do they need to keep? This only removes the auto query on query & option changes. Should not affect internal query editor state? Or will it make query editors "not controlled" (ie only set default value). That would be a big change. 

 Think with a change like this there won't be any immediate effects of query or option changes. Might need a more clear "run query" action like in explore.

Screenshot from 2020-12-11 10-14-10

ryantxu:
Currently many query editors do not keep state -- and will call update for each `onChange` event (like a key press) then call `onRunQueries` when enter is pressed or an input loses focus.  

Consistency across query editors varies widely though -- so yes a more clear "run query" button would be nice.  Some editors like azure even add their own `run query` blue button

torkelo (approved):

PR #29763, opened 2020-12-11
GraphNG: update uPlot v1.5.0

leeoniya:
a lot of goodies here:

- `spline`, `stepped`, and `bars` path builders moved into uPlot core for easier maintenance and tighter integration. all have been refactored to use a new lower-level, more explicit, and faster API. the linear builder was extracted from the core to be equally as modular and consume the same api. all paths now run at peak performance. :rocket: 
- `bars` paths got some fixes with width calculations, you can see how broken it was previously by zooming into only a few bars. besides `widthFactor`, `bars` now have a `maxWidth` parameter, should we choose to expose it.
- the `outerJoinValues()` function was moved into a `uPlot.join()` utility function.
- log scales now handle values <= 0 without crashing.
- gradients supplied to `series.fill` have been fixed (was a bug in the opts cloning algorithm)
- `opts.gutters: {x,y}` is gone, replaced by a more flexible and explicit `opts.padding: [top, rgt, btm, lft]`. this is important for edge-to-edge sparklines without axes that still need padding to prevent cutting off of points and strokes at the edges. (see https://github.com/leeoniya/uPlot/issues/392 for visuals)
- a new option called `drawOrder: ["axes", "series"]` allows for drawing the grid over the series. might be needed for native-izing @ryantxu's https://github.com/NatelEnergy/grafana-discrete-panel

the second commit is unrelated to the update, but fixes a superfluous client-zoom after a selection end by setting `cursor.drag.setScale = false` and manually hiding the selection region in the `setSelect` hook.

i think that about covers it.


leeoniya:
> TypeError: Cannot read property 'bars' of undefined

not sure what's going on with the CI, is it not picking up the new uPlot version? :thinking: 

(feel free to push commits to this branch to fix the tests here)

packages/grafana-ui/src/components/uPlot/config/UPlotConfigBuilder.ts
```diff
@@ -68,9 +69,12 @@ export class UPlotConfigBuilder {
...
       return { ...acc, ...s.getConfig() };
     }, {});
-    if (this.cursor) {
-      config.cursor = this.cursor;
-    }
+
+    config.cursor = this.cursor || {};
+
+    // prevent client-side zoom from triggering at the end of a selection
+    defaultsDeep(config.cursor, { drag: { setScale: false } });
```
ryantxu, line 76:
```suggestion
    config.cursor.drag = { setScale: false };
```

The `defaultsDeep` returns a value -- so I don't think this does what you think

ryantxu (approved):
looks good!

packages/grafana-ui/src/components/uPlot/config/UPlotConfigBuilder.ts
ryantxu: nevermind -- you are right!

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -1,9 +1,19 @@
 import tinycolor from 'tinycolor2';
 import uPlot, { Series } from 'uplot';
 import { DrawStyle, LineConfig, AreaConfig, PointsConfig, PointVisibility, LineInterpolation } from '../config';
-import { barsBuilder, smoothBuilder, stepBeforeBuilder, stepAfterBuilder } from '../paths';
 import { PlotConfigBuilder } from '../types';
 
+const pathBuilders = uPlot.paths;
```
dprokop, line 6: @leeoniya seems like the Plot tests are failing because of this. Are `paths` exposed from uPlot for sure?

dprokop: Is it possible this line is causing troubles https://github.com/leeoniya/uPlot/blob/a568fb12b6567567e9f5ce74a31d3628b56def5c/src/uPlot.js#L2346

leeoniya: the only thing in the npm package is the `/dist` directory, and "module" in `package.json` points to the esm file in there, so:

https://github.com/leeoniya/uPlot/blob/530a46c53213bb63593451ae48d61e566185cb9e/dist/uPlot.esm.js#L3971-L3981

ryantxu:
Since other PRs depend on this, and sorting out the jest import is non-trivial -- i removed the test (renamed the file).  I could not do `.skip` since the failure is on the initial import not how the object gets mocked

image

PR #29764, opened 2020-12-11
AlertListPanel: Add options to sort by Time(asc) and Time(desc)

### LINKED ISSUE #29156: Alert list panel: Option to invert sort order in recent changes mode
Add an option to invert the alert list panel sort order when set to "Recent state changes" mode.

**Why is this needed**:
The current order is confusing, new entries are added at the bottom, nobody looks at the bottom for new entries.
New and important alerts are scrolling out of the box and you will never notice them.

alarm_panel

dboslee:
This also makes all the sortOrderOptions available when using
the "changes" showOption.

Closes: #29156


hugohaggmark (approved):
Thank you for contributing @dboslee! This looks good and works as expected ⭐

PR #29765, opened 2020-12-11
GraphNG: support fill gradient

ryantxu:
This depends on #29763

first pass at supporting gradient -- @torkelo feel free to take this over.  I'm not quite sure how/if the `fillAlpha` should be used


torkelo:
Nice, yea can take this over after #29763 is merged

packages/grafana-ui/src/utils/measureText.ts
```diff
@@ -1,6 +1,21 @@
...
+
 let canvas: HTMLCanvasElement | null = null;
 const cache: Record<string, TextMetrics> = {};
 
+function getCanvasContext() {
+  if (canvas === null) {
+    canvas = document.createElement('canvas');
+  }
+
+  const context = canvas.getContext('2d');
```
leeoniya, line 11: no cache?

packages/grafana-ui/src/utils/measureText.ts
```diff
@@ -43,3 +51,19 @@ export function calculateFontSize(text: string, width: number, height: number, l
   const optimalSize = Math.min(fontSizeBasedOnHeight, fontSizeBasedOnWidth);
   return Math.min(optimalSize, maxSize ?? optimalSize);
 }
+
+/**
+ * @beta
+ */
+export function getCanvasGradient(color: string, amount: number, plotHeight: number): CanvasGradient {
+  const ctx = getCanvasContext();
+  var gradient = ctx.createLinearGradient(0, 0, 0, plotHeight);
```
leeoniya, line 60: hmm, what happens to this when `plotHeight` changes due to resize? i guess just live with it or kill perf and have some form of throttle/debounce that updates config and rehinits the plot?

technically, you can just mutate `u.series[i].fill` and invoke a cheap cached `u.redraw(false)`.

tricky.

torkelo: think we need to pass in the current graph height here, so graph height change will need to invalidate/update the config.

leeoniya: right, but chart resize happens every dragged pixel? and changes in config cause expensive plot recreation. so this might need to be throttled till the resize ends?

torkelo:
Changed the fill gradient option from being a number like opacity to a enum mode. One complaint with the old area gradient mode option was that it made the opacity option unusable. I think using the opacity option to control the gradient amount is a possible option. 

The modes are: 
None (no gradient)
Opacity  (Opacity gradient, controlled by opacity slider) 
Hue (Automagic hue gradient, opacity slider controls just opacity) 

Screenshot from 2020-12-12 22-20-20

Checked in test dashboard:
Screenshot from 2020-12-12 22-21-25
Screenshot from 2020-12-12 22-16-11

leeoniya:
very fancy! :tophat: :wine_glass: :smoking:

ryantxu:
Nice -- as a challenge I'll suggest one more mode...  Thresholds!  Or maybe the configured color scale.

leeoniya:
@ryantxu i was actually gonna ask if the gradients were just stylistic or had data implications that had to follow the rendrred scales - that would be super nice, you'd just need to do the math based on uplot.bbox.height, gradient height and map it to the correct offset relative to scale.min/max. might be a rabbit hole with people wanting to micro-tune gradient stops to exact colors at specific scale values.

ryantxu:
That is why I said it is a challenge!

packages/grafana-ui/src/utils/measureText.ts
torkelo: reading docs for getContext it looks like this is cached internally in the canvas instance, all subsequent calls to getContext all get the same instance, so do not think there is anything gained by caching this call

leeoniya:
challenge accepted!

here's a dynamic gradient fill that's tied to scale values and configured independently per-series:

one tank is red for under-pressure, the other is red for over-pressure. if you toggle one tank off, the other gradient rescales to follow the new scale range.

https://jsfiddle.net/nkswhjx3/

it's a bit hack-ish that it uses the `drawClear` hook and replaces `s.fill`, but adding a function signature to `series.fill` will be super easy, and the re-calc logic would be moved to there. this is a super valid use-case, so it'll be in the next uPlot version :+1:

torkelo:
yea, making fill a callback (function), that way we can Grafana implement that function and check the current canvas height. Then we can use the DataFrame field thresholds for the gradient (if gradient mode is thresholds)

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -100,3 +118,47 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
...
+
+function getCanvasGradient(
+  color: string,
+  gradientMode: AreaGradientMode,
+  opacity: number,
+  plotHeight: number
+): CanvasGradient {
+  const ctx = getCanvasContext();
+  const pxRatio = window.devicePixelRatio ?? 1;
+  const pixelHeight = (plotHeight - 30) * pxRatio;
```
leeoniya, line 130: is 30 a rough fudge factor to exclude the axis & padding? feels like this should be done externally already, so getCanvasGradient doesn't need to care.

torkelo: Yes, but figured this is just temporary as the height is very wrong in other ways as well (does not account for legend below graph for example).

But if fill is changed to a function (callback) then should be able to get canvas height (excluding axes height) from the uplot instance I hope

leeoniya: > But if fill is changed to a function (callback) then should be able to get canvas height (excluding axes height) from the uplot instance I hope

:100:

leeoniya:
> yea, making fill a callback (function), that way we can Grafana implement that function and check the current canvas height.

here we go: https://jsfiddle.net/c5wzo9y6/

leeoniya:
as with anything that sounds simple at first, it ended up being somewhat more involved because the legend series indicators "inherit" the fill & stroke specified for each series, but since the indicators are html/css, that's not always possible. the solution actually ended up closing out another feature request that wanted explicit control over the legend marker's fill and stroke styles (https://github.com/leeoniya/uPlot/issues/383). none of this matters for Grafana, since we use a custom legend, but it had to be solved for the uPlot's native legend to prevent the series markers from disappearing completely.

new demo with both, scale-aligned and height-aligned gradients: https://leeoniya.github.io/uPlot/demos/gradient-fill.html

torkelo:
Think this is ready. For scale / scheme based gradients I want to do that in another PR. 
Here is a design doc for how we could do it with minimal options : https://docs.google.com/document/d/15G523kih8qj_d-kp-7AetLFRP7Q1D-aHvj4eDAXGVo0/edit#

ryantxu:
Looks great!

torkelo (approved):

PR #29766, opened 2020-12-11
Chore: Expose TimeSrv from grafana/runtime package

alexanderzobnin:
This PR makes it possible to use `TimeSrv` from `@grafna/runtime` package. This change required for extracting Elasticsearch data source (and potentially for other core data sources) into standalone plugin.

The only function used by elasticsearch data source is a `timeRange()`. `timeRangeForUrl()` required for initializing `locationUtil` in `GrafanaCtrl`. We can export more methods if we need it, but I only found these two so far.

### DIFF
--- a/packages/grafana-runtime/src/services/index.ts
+++ b/packages/grafana-runtime/src/services/index.ts
@@ -6,3 +6,4 @@ export * from './EchoSrv';
 export * from './templateSrv';
 export * from './legacyAngularInjector';
 export * from './live';
+export * from './timeSrv';
--- a/packages/grafana-runtime/src/services/timeSrv.ts
+++ b/packages/grafana-runtime/src/services/timeSrv.ts
@@ -1,0 +1,37 @@
+import { RawTimeRange, TimeRange } from '@grafana/data';
+
+/**
+ * TimeSrv allows to get and control dashboard time range.
+ *
+ * @public
+ */
+export interface TimeSrv {
+  /**
+   * Get a current time range or default one.
+   */
+  timeRange(): TimeRange;
+
+  /**
+   * Get a raw time range for current URL.
+   */
+  timeRangeForUrl(): RawTimeRange;
+}
+
+let singletonInstance: TimeSrv;
+
+/**
+ * Used during startup by Grafana to set the DataSourceSrv so it is available
+ * via the {@link getTimeSrv} to the rest of the application.
+ *
+ * @internal
+ */
+export const setTimeSrv = (instance: TimeSrv) => {
+  singletonInstance = instance;
+};
+
+/**
+ * Used to retrieve the {@link TimeSrv} that allows to get and control dashboard time range.
+ *
+ * @public
+ */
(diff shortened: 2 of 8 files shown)


torkelo:
For evaluating index patterns it looks like https://github.com/grafana/es-open-distro-datasource/blob/483c6023d209aa91d942bdcb4d59ef1a4f97c35e/src/datasource.ts#L141

alexanderzobnin:
@ryantxu unfortunately, there's no easy way to do it. Data source needs time range to calculate proper index name and this may happen not only inside `query()` method, but in places where time range in not passed to the plugin. So probably it's possible, but we need to investigate it. We don't have much time for that, so this looks like a fastest solution.

torkelo:
@alexanderzobnin I think technically it can, all the methods are all used from  metricFindQuery(query: string) , which does take an options argument as the second option which does have a time range property. (But only for variables that have refresh on time range) .

https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/datasource.ts#L229

https://github.com/grafana/grafana/blob/d0f52d5334d1e3623c7360d0291d29deb5d3f6eb/public/app/features/variables/utils.ts#L127

So the question is what to default to when we do not have a range here (like for testDatasource)

alexanderzobnin:
I guess default is `now-6h`, but I'll check it out. Probably we can get rid of TimeSrv completely in plugin.

ryantxu:
As a quick work-around that does not commit to having timeSrv in the future... what about importing it directly with SystemJS or the angular injector?  like:  https://github.com/ryantxu/image-viewer-panel/blob/master/src/ImageViewer.tsx#L41

alexanderzobnin:
Hm, that might be an option for the transition period. So I'll let this PR open until I find a solution for the plugin. Then we can discuss, do we need this or not.

alexanderzobnin:
@ryantxu unfortunately, that hack doesn't work, I guess, it works in your case, because of 
https://github.com/grafana/grafana/blob/d0f52d5334d1e3623c7360d0291d29deb5d3f6eb/public/app/features/plugins/plugin_loader.ts#L145

But timeSrv is not exposed.

alexanderzobnin:
Looks like it's possible to remove `timeSrv` completely: https://github.com/grafana/es-open-distro-datasource/commit/06901d3075e3e2d84178d7211ce175de68666e9b

torkelo:
@alexanderzobnin can you do those changes to the core plugin and then copy them over? Important to not have the core & external plugins diverge

alexanderzobnin:
@torkelo ok, sure, makes sense.

mckn:
I looks like this PR isn't needed any more, correct? Great that you managed to solve it by using the `timeRange` being passed to `metricFindQuery`. 👍

mckn (commented):
I looks like this PR isn't needed any more, correct? Great that you managed to solve it by using the `timeRange` being passed to `metricFindQuery`. 👍

alexanderzobnin:
Yes, we've found a better way, so closing this one.

PR #29767, opened 2020-12-11
Panel/inspect:  Enable to export csv file with filtered data

### LINKED ISSUE #29113: Download CSV with returned rows when Column filter applied
Not sure if this is a enhancement request or a bug. 

When downloading CSV as Excel format with Column filter applied all data is saved to file and not as filtered list is showing.

It would be nice if the column filters applied would influence the download data.

image

chihsuan:

This PR makes it possible to export csv file with filtered data rather than origin all data.

Fixes #29113

### DIFF
--- a/packages/grafana-ui/src/components/Table/Table.tsx
+++ b/packages/grafana-ui/src/components/Table/Table.tsx
@@ -1,9 +1,10 @@
-import React, { FC, memo, useCallback, useMemo } from 'react';
+import React, { FC, memo, useCallback, useEffect, useMemo } from 'react';
 import { DataFrame, Field, getFieldDisplayName } from '@grafana/data';
 import {
   Cell,
   Column,
   HeaderGroup,
+  Row,
   useAbsoluteLayout,
   useFilters,
   UseFiltersState,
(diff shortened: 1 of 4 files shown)


torkelo:
Thanks for contributing! We are very worried about feature creep, and adding features for edge cases that very few will actually use. 

I am not sure we need this, the CSV is for raw data exports. And excel can easily handle the same filtering. 

would want to see more demand for https://github.com/grafana/grafana/issues/29113

chihsuan:
> Thanks for contributing! We are very worried about feature creep, and adding features for edge cases that very few will actually use.
> 
> I am not sure we need this, the CSV is for raw data exports. And excel can easily handle the same filtering.
> 
> would want to see more demand for #29113

Got it, thanks for explaining about that.

nicka-redwood:
@torkelo is there any chance you'd be willing to reconsider this feature?  Exporting data out of grafana is useful for lots of reasons, and grafana already provides a great mechanism for searching and selecting those items you want to see.  

It's fairly annoying for a user to apply all the correct filters in grafana to see if their data has appeared yet, only to then have to download the entire (possibly very large) data set and re-do all the filtering in excel. 

I think the interest on the issue that prompted this PR shows that others are pretty interested in this as well. 
https://github.com/grafana/grafana/issues/29113

mstepien96:
I would also like to request reconsideration of this feature. Based on the #29113, the interest of the Grafana users seems to be non-negligible.

As @nicka-redwood mentioned, downloading the whole csv file with no filters is excessive in many use cases. Also, the need to reapply filters on downloaded data renders Grafana useless in terms of report-creation, which me and other Enterprise customers are very sad to see.

nokiche:
I clearly confirmed it's something that is asked by our users. 
What's the point to filter data in column filter if you need to redo it after in Excel ? it's clearly a loss of time

RonitGandhi:
I agree with @nokiche and I would love this as a feature its very useful and time saving.

PR #29769, opened 2020-12-11
GraphNG: Add test dashboard for null & and gaps rendering

torkelo:
Discovered a bug in the gaps logic and figured be good to have a test dashboard for this scenario. 

Screenshot from 2020-12-11 12-35-37

The thing that is not working is when we have unaligned data & series with real null values. 
In the graph below Series B have null values between every data point, so there should be no line drawn between any data point. But when I add series A (and depending on how many values series A has) it adds this additional line. There is also a gap for Series A before the last data point that is not there in the data. 
 
Screenshot from 2020-12-11 12-36-21

### DIFF
--- a/devenv/dev-dashboards/panel-graph/graph-ng-nulls.json
+++ b/devenv/dev-dashboards/panel-graph/graph-ng-nulls.json
@@ -1,0 +1,669 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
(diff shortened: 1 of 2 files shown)


leeoniya:
thanks for this, super useful. i'm gonna try to extract the data & settings here and add it to uPlot's gap "test suite" :D. hopefully something simple.

https://leeoniya.github.io/uPlot/demos/path-gap-clip.html

PR #29771, opened 2020-12-11
Explore: Fix remounting of query row

aocenas:
Because in init it used data source from state, which wasn't inited yet, it created temp row key which got changed later, which resulted in remounting of whole query row. Usually it did not have any visual effect but if there was some initialisation in the QueryEditor with some loading state this would be annoying.


ivanahuckova (approved):
Nice! Good job! This is also going to help with on of the issues I observed with Prometheus query editor. We are passing queryKey as a key for queryRow and ~as it changes~ as it was changing, we were mounting and unmounting each row twice. In Prometheus it meant, that we've run metadata query on componentDidMount twice.

PR #29772, opened 2020-12-11
PanelLibrary: Adds get and getAll to the api

hugohaggmark:
This PR adds `GET /api/library-panels/:panelId` and `GET /api/library-panels/api` for Library/Shared/Reusable Panels.

I also added `Body()` method to the `NormalResponse` so I could verify the actual responses.

Finally but not least enjoy the review 🎉 

Relates #29610

I did some refactoring of the tests because it was hard to debug a single test with the previous `libraryPanelScenario`. Please let me know if I went to far with that refactor :)


pkg/services/librarypanels/api.go
```diff
@@ -49,3 +51,28 @@ func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response
...
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to get library panel", err)
+	}
+
+	return api.JSON(200, util.DynMap{"result": libraryPanel})
+}
+
+// getAllHandler handles GET /api/library-panels/
+func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response {
```
torkelo, line 70: think this will need some search, page  & page size at some point, but good to start simple :)

torkelo (approved):
Looks good.

One big thing that I forgot to think about for the db model & api is database numeric id vs uid (string guid). 

To support provisioning and referring to panel libraries in dashboard json where you do not know the DB id we need a uid for library panels. And think this is what panels in json (and in the HTTP api) should use. 

Ideas for persisted dashboard json (ie not runtime in DashboardModel) 
```typescript
panels: [
 {
    id: 10,    
    libraryPanel: {
      uid: 'asd12eadd12',
      name: 'Awesome graph',
    }
]
```

Something like that could be useful (to have uid & name), so when viewing a persisted json (saved to disk / git), you see the uid and the name (the name would be there just for readability). 

Switch to uid can be done in a separate PR. 

@aknuds1 do you have any feedback?

pkg/services/librarypanels/api.go
hugohaggmark: Yes, keeping it simple for now :)

hugohaggmark:
> Looks good.
> 
> One big thing that I forgot to think about for the db model & api is database numeric id vs uid (string guid).
> 
> To support provisioning and referring to panel libraries in dashboard json where you do not know the DB id we need a uid for library panels. And think this is what panels in json (and in the HTTP api) should use.
> 
> Ideas for persisted dashboard json (ie not runtime in DashboardModel)
> 
> ```ts
> panels: [
>  {
>     id: 10,    
>     libraryPanel: {
>       uid: 'asd12eadd12',
>       name: 'Awesome graph',
>     }
> ]
> ```
> 
> Something like that could be useful (to have uid & name), so when viewing a persisted json (saved to disk / git), you see the uid and the name (the name would be there just for readability).
> 
> Switch to uid can be done in a separate PR.
> 
> @aknuds1 do you have any feedback?

Good catch on the uid & name, will introduce that change after this PR.

hugohaggmark:
@aknuds1 any feedback before I merge this?

hugohaggmark:
@wbrowne any feedback before I merge this?

aknuds1:
Checking.

pkg/services/librarypanels/api.go
```diff
@@ -17,13 +17,15 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
...
+		libraryPanels.Post("/", middleware.ReqSignedIn, binding.Bind(createLibraryPanelCommand{}), api.Wrap(lps.createHandler))
 		libraryPanels.Delete("/:panelId", middleware.ReqSignedIn, api.Wrap(lps.deleteHandler))
+		libraryPanels.Get("/:panelId", middleware.ReqSignedIn, api.Wrap(lps.getHandler))
+		libraryPanels.Get("/", middleware.ReqSignedIn, api.Wrap(lps.getAllHandler))
 	})
 }
 
-// createHandler handles POST /api/library-panels.
-func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd addLibraryPanelCommand) api.Response {
+// createHandler handles POST /api/library-panels
```
aknuds1, line 27: Massive nit, but I would keep the dot, since it's [standard](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) for Go documentation comments :)
```suggestion
// createHandler handles POST /api/library-panels.
```

pkg/services/librarypanels/api.go
```diff
@@ -49,3 +51,28 @@ func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response
 
 	return api.Success("Library panel deleted")
 }
+
+// getHandler handles GET /api/library-panels/:panelId
+func (lps *LibraryPanelService) getHandler(c *models.ReqContext) api.Response {
+	libraryPanel, err := lps.getLibraryPanel(c, c.ParamsInt64(":panelId"))
+
+	if err != nil {
```
aknuds1, line 59: I would recommend not having a blank line between an error check and its origin, since it makes verifying harder for someone reading the code (as the connection is no longer obvious):
```suggestion
	libraryPanel, err := lps.getLibraryPanel(c, c.ParamsInt64(":panelId"))
	if err != nil {
```

pkg/services/librarypanels/api.go
```diff
@@ -49,3 +51,28 @@ func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response
...
+	}
+
+	return api.JSON(200, util.DynMap{"result": libraryPanel})
+}
+
+// getAllHandler handles GET /api/library-panels/
+func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response {
+	libraryPanels, err := lps.getAllLibraryPanels(c)
+
+	if err != nil {
```
aknuds1, line 73: I would recommend not having a blank line between an error check and its origin, since it makes verifying harder for someone reading the code (as the connection is no longer obvious):
```suggestion
	libraryPanels, err := lps.getAllLibraryPanels(c)
	if err != nil {
```

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
 
 	return err
 }
+
+// getLibraryPanels gets Library Panels based on statement and args
```
aknuds1, line 66: Why "Library Panels", if it's not referring to the LibraryPanel type? I'd think that if you're not referring to the specific type, you'd just write library panels?

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+func getLibraryPanels(session *sqlstore.DBSession, statement interface{}, args ...interface{}) ([]LibraryPanel, error) {
+	libraryPanels := make([]LibraryPanel, 0)
+
+	err := session.SQL(statement, args...).Find(&libraryPanels)
```
aknuds1, line 70: I don't see the reason for so much whitespace :)
```suggestion
	libraryPanels := make([]LibraryPanel, 0)
	err := session.SQL(statement, args...).Find(&libraryPanels)
```

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+
+	return libraryPanels, err
+}
+
+// getLibraryPanel gets a Library Panel
+func (lps *LibraryPanelService) getLibraryPanel(c *models.ReqContext, panelID int64) (LibraryPanel, error) {
+	orgID := c.SignedInUser.OrgId
+	libraryPanel := LibraryPanel{}
+
```
aknuds1, line 79: I personally prefer removing this blank line, since these variables are specifically set by the closure.
```suggestion
	orgID := c.SignedInUser.OrgId
	libraryPanel := LibraryPanel{}
```

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		libraryPanels, err := getLibraryPanels(session, "SELECT * FROM library_panel WHERE id=? and org_id=?", panelID, orgID)
+
+		if err != nil {
```
aknuds1, line 83: Can we just inline getLibraryPanels? That abstraction is so thin, I think it only makes it more difficult to read the code and see what actually happens. I'd find the following much easier to read:
```suggestion
		var libraryPanels []LibraryPanel
		err := session.SQL("SELECT * FROM library_panel WHERE id=? and org_id=?", panelID, orgID).Find(&libraryPanels)
		if err != nil {
```

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+			return err
+		} else if len(libraryPanels) != 1 {
+			return errLibraryPanelNotFound
```
aknuds1, line 86: Shouldn't this be treated as a bug instead?? It should be logged as an error and returned as 500 to the client, or maybe even ` panic` would be warranted since it would be a hard error on our side.

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+
+// getAllLibraryPanels gets all Library Panels
+func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]LibraryPanel, error) {
+	orgID := c.SignedInUser.OrgId
+	libraryPanels := make([]LibraryPanel, 0)
+
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		panels, err := getLibraryPanels(session, "SELECT * FROM library_panel WHERE org_id=?", orgID)
+
+		if err != nil {
```
aknuds1, line 105: Same concern as before, I don't think the `getLibraryPanels` abstraction should be used, inlining is more readable:
```suggestion
	var libraryPanels []LibraryPanel
	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
		err := session.SQL("SELECT * FROM library_panel WHERE org_id=?", orgID).Find(&libraryPanels)
		if err != nil {
```

pkg/services/librarypanels/database.go
```diff
@@ -62,3 +62,54 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, panelID
...
+			return err
+		}
+
+		libraryPanels = panels
+
+		return nil
```
aknuds1, line 111:
```suggestion
		return nil
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -15,78 +16,251 @@ import (
 )
 
 func TestCreateLibraryPanel(t *testing.T) {
-	libraryPanelScenario(t, "When an admin tries to create a library panel", "then it should fail if library panel already exists", func(t *testing.T) {
-		lps, context := setupTestEnv(t, models.ROLE_ADMIN, map[string]string{})
-		command := getCreateCommand(1, "Text - Library Panel")
+	t.Run("When an admin tries to create a library panel that already exists", func(t *testing.T) {
```
aknuds1, line 19: Why don't you use the `libraryPanelScenario` abstraction any longer, to define test cases? I think it makes sense, and is more ergonomic than the new pattern here? 
```suggestion
	t.Run("When an admin tries to create a library panel that already exists, it should fail", func(t *testing.T) {
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -15,78 +16,251 @@ import (
...
+		t.Run("then it should fail", func(t *testing.T) {
```
aknuds1, line 20: You should only use `t.Run` to define (sub) test cases, not to group assertions.
```suggestion
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -15,78 +16,251 @@ import (
...
+			testScenario(t, getTestContext(), func(t *testing.T, ctx *testContext) {
```
aknuds1, line 21: I think that instead of this `getTextContext()` abstraction, it would be more Go style to provide a config object, with eventual parameters diverging from the defaults.

aknuds1 (changes requested):
Please see my comments. I think mainly the tests should not be written with `t.Run` to group assertions - `t.Run` is meant for defining (sub) test cases, not grouping assertions. It's only confusing when running tests, if t.Run is used like this, since every invocation creates a sub-test that you can run in isolation and it becomes really noisy if every sub-test isn't actually a case.

pkg/services/librarypanels/database.go
torkelo: why would it be a bug? the panel is just not found with the specified id, would be a simple 404 in the API

aknuds1: @torkelo Sorry I didn't think about that case, I thought about the case where `len(libraryPanels) > 1`. I always take that case into account when I write these queries.

pkg/services/librarypanels/api.go
hugohaggmark: Sure I agree, will fix this in my next commit.

pkg/services/librarypanels/api.go
hugohaggmark: Sure I agree, will fix this in my next commit.

pkg/services/librarypanels/database.go
hugohaggmark: Sure I agree, will fix this in my next commit.

pkg/services/librarypanels/database.go
hugohaggmark: Sure, I can see your point. Will fix this in my next commit.

pkg/services/librarypanels/database.go
hugohaggmark: That doesn't seem so YAGNI to me @aknuds1 ;) But sure I can add that as well in my next commit.

pkg/services/librarypanels/database.go
hugohaggmark: Sure, I can see your point. Will fix this in my next commit.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: The simple answer is that I couldn't debug the individual t.Run within the libraryPanelScenario, but as you mentioned that I shouldn't have t.Run within t.Run I'll see what I can come up with.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Okidoki

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Will look into this in my next commit as well.

pkg/services/librarypanels/api.go

pkg/services/librarypanels/database.go

pkg/services/librarypanels/database.go
hugohaggmark: Removed the whole function

pkg/services/librarypanels/database.go
hugohaggmark: Removed

pkg/services/librarypanels/database.go
hugohaggmark: Not YAGNI but added this as suggested.

pkg/services/librarypanels/database.go
hugohaggmark: Removed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/database.go
hugohaggmark: Don't understand this comment, so did not change.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Not sure I understand this comment. Testscenario receives a pointer to a struct not a function. I invoked the function because I didn't see the point of declaring a context variable. Now I have declared an extra variable for each test instead.

hugohaggmark:
@aknuds1 I fixed most of your comments, do you have anything else that needs to be done? (Changes requested)

aknuds1:
@hugohaggmark I'm in the middle of something more pressing, but I am coming up with a suggestion for changes.

aknuds1 (changes requested):
Thanks for applying my suggestions, I think the tests should be written according to a good pattern we have in the middleware tests, where devise wrappers for `t.Run` to have common setup logic. I've implemented this pattern in my branch [arve/libpanel-get](https://github.com/grafana/grafana/tree/arve/libpanel-get). Can you please merge that branch into your PR branch?

Eventually I can just push those changes to your branch if you want.

pkg/services/librarypanels/database.go
aknuds1: That's the first time I hear of any connection between YAGNI and handling possible error cases. Handling possible error cases is only good code hygiene.

hugohaggmark:
> [arve/libpanel-get](https://github.com/grafana/grafana/tree/arve/libpanel-get).

Sure that looks good, thank you!

hugohaggmark:
@aknuds1 is this ready for approval?

aknuds1 (approved):

PR #29773, opened 2020-12-11
Expressions: Add placeholders to hint on input

peterholmberg:
Changing some labels and adding placeholders to input fields to make it easier to understand what you can type.


kylebrandt (approved):
Thanks!

PR #29774, opened 2020-12-11
Usage Stats: Allow to add additional metrics to the stats

vtorosyan:
### What
The change allows to add additional metrics to the usage stats, enabling instrumentation from the outside modules/services.
Also converts existing tests from GoConvey to use standard library.

### Why
We would like to send additional metrics for the enterprise features. With the current implementation,  the service builds metrics within itself in hardcoded fashion. The change will give a chance to have a generic way of adding metrics to the stats, without a need to hardcode anything.

A client would need to call `UsageStats.AddMetric` method, passing the name of the metric and a `func`  evaluating the value.

### Which issue this PR fixes
Enables to instrument enterprise features, as described in the [#915](https://github.com/grafana/grafana-enterprise/issues/915) issue.

### To Do 
- [x] Rewrite tests to use standard lib instead of GoConvey


pkg/infra/usagestats/usage_stats.go
```diff
@@ -186,6 +188,21 @@ func (uss *UsageStatsService) GetUsageReport() (UsageReport, error) {
...
+		result, err := fn()
+		if err != nil {
+			metricsLogger.Error("Failed to fetch external metric", "name", name, "error", err)
+			continue
+		}
+		metrics[name] = result
+	}
+}
+
+func (uss *UsageStatsService) AddMetric(name string, fn MetricFunc) {
```
undef1nd, line 202: Maybe worth naming it `AddExternalMetric`? 🤔

vtorosyan: My thinking was that the fact that those metrics are "external" is an implementation detail, which is relevant only to the usage_stats service/package. From a client perspective, you just add a "metric". This also assumes future compatibility, when we decide to not have hardcoded metrics here anymore, thus the "external" concept goes away. wdyt ?

undef1nd: Makes sense. Thanks for explaining.

marefr (approved):
Great LGTM

Some nits:
- Would maybe rename `AddMetric`  to `RegisterMetric` instead because you're basically registering a callback function that will be called at any point later

pkg/infra/usagestats/service.go
```diff
@@ -30,12 +32,14 @@ type UsageStatsService struct {
...
 
-	oauthProviders map[string]bool
+	oauthProviders  map[string]bool
+	externalMetrics map[string]MetricFunc
 }
 
 func (uss *UsageStatsService) Init() error {
 	uss.log = log.New("infra.usagestats")
 	uss.oauthProviders = social.GetOAuthProviders(uss.Cfg)
+	uss.externalMetrics = make(map[string]MetricFunc)
```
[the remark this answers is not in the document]
marefr: Probably a good idea to initialize it when registering the service above

PR #29775, opened 2020-12-11
Chore: Remove internal import for QueryEditorRow

Elfo404:
Removes the internal import from `@grafana/ui/src/components/Forms/InlineLabel` of `getInlineLabelStyles`.

Not an ideal solution as I'm basically just copying the resulting styles, I added a FIXME comment to tackle this in a better way in a follow-up PR.
We need this as a part of work on extracting Elasticsearch data source into a separate plugin.

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/components/QueryEditor/QueryEditorRow.tsx
+++ b/public/app/plugins/datasource/elasticsearch/components/QueryEditor/QueryEditorRow.tsx
@@ -1,6 +1,5 @@
 import { GrafanaTheme } from '@grafana/data';
 import { IconButton, stylesFactory, useTheme } from '@grafana/ui';
-import { getInlineLabelStyles } from '@grafana/ui/src/components/Forms/InlineLabel';
 import { css } from 'emotion';
 import { noop } from 'lodash';
 import React, { FunctionComponent } from 'react';
@@ -24,7 +23,7 @@ export const QueryEditorRow: FunctionComponent<Props> = ({
 
   return (
     <fieldset className={styles.root}>
-      <div className={getInlineLabelStyles(theme, 17).label}>
+      <div className={styles.wrapper}>
         <legend className={styles.label}>{label}</legend>
         {onHideClick && (
           <IconButton
@@ -57,6 +56,25 @@ const getStyles = stylesFactory((theme: GrafanaTheme) => {
     root: css`
       display: flex;
       margin-bottom: ${theme.spacing.xs};
+    `,
+    // FIXME: this is taken from  `getInlineLabelStyles` in '@grafana/ui/src/components/Forms/InlineLabel' with width = 17
+    // We should have a better way to access / use these styles.
+    wrapper: css`
+      display: flex;
+      align-items: center;
+      justify-content: space-between;
+      flex-shrink: 0;
+      padding: 0 ${theme.spacing.sm};
+      font-weight: ${theme.typography.weight.semibold};
+      font-size: ${theme.typography.size.sm};
+      background-color: ${theme.colors.bg2};
+      height: ${theme.height.md}px;
+      line-height: ${theme.height.md}px;
+      margin-right: ${theme.spacing.xs};
+      border-radius: ${theme.border.radius.md};
+      border: none;
+      width: 136px;
+      color: ${theme.colors.textHeading};
     `,
     label: css`
       font-size: ${theme.typography.size.sm};


public/app/plugins/datasource/elasticsearch/components/QueryEditor/QueryEditorRow.tsx
```diff
@@ -24,7 +23,7 @@ export const QueryEditorRow: FunctionComponent<Props> = ({
 
   return (
     <fieldset className={styles.root}>
-      <div className={getInlineLabelStyles(theme, 17).label}>
+      <div className={styles.wrapper}>
```
torkelo, line 26: not sure why we need these custom styles, can't we use InlineFieldRow & InlineFieldLabel here?

Elfo404: They don't really fit this use case here.

this will render as `fieldset` for the entire row (which uses other `InlineField`s & `Segment`s internally), described by (using what would be the `InlineLabel`) a `legend` element.

semantically this should make more sense than having an `InlineLabel` (inside which also we can't use the 2 icon buttons for remove & hide because the first one would be labeled with, for example `Metric (2)` which is incorrect). 

The label element should be used to describe a single input element, while here we need to describe a group of fields.

what do you think? maybe it makes sense to have this in `@grafana/ui` as well?

torkelo: Not sure I understand the difference, maybe @Clarity-89 knows more, InlineFieldLabel is exactly like a segment from the old form styles (gf-form-label). 

So this should be replacable with 

```jsx
<InlineFieldRow>
  <InlineFieldLabel>
     {label}
{onHideClick && (
          <IconButton
            name={hidden ? 'eye-slash' : 'eye'}
            onClick={onHideClick}
            surface="header"
            size="sm"
            aria-pressed={hidden}
            aria-label="hide metric"
            className={styles.icon}
          />
        )}
        <IconButton
          name="trash-alt"
          surface="header"
          size="sm"
          className={styles.icon}
          onClick={onRemoveClick || noop}
          disabled={!onRemoveClick}
          aria-label="remove metric"
        />
  </InlineFieldLabel>
   {children}
</InlineFieldRow>
```

torkelo: But am unable to get this to work either, there seem to be some important missing components in the new inline forms styles. For example cannot find a component that is similar to gf-form that can wrap many gf-form-labels, InlineField requires a label and single input (child).

torkelo: Close to getting it to work here, https://github.com/grafana/grafana/pull/29782 

but needs more tweaking and updated the inline form styles that could require a lot of changes everywhere the inline forms are used

Clarity-89: Yeah, inline form components do not support fieldset use, we have `FieldSet` [component](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Forms/FieldSet.tsx#L13) for non-inline components, but not for the inline ones.

Clarity-89: @torkelo @Elfo404  After playing around with it a bit, I think I found a solution. Using `InlineField` and passing the label component via its `label` prop seems to work well. `InlineField` will take care of aligning the components inside of it and applying proper margins/spacing. 
```jsx
  <InlineFieldRow>
      <InlineField
        label={
          <InlineLabel width={17}>
            <legend className={styles.label}>{label}</legend>
            {onHideClick && (
              <IconButton
                name={hidden ? 'eye-slash' : 'eye'}
                onClick={onHideClick}
                surface="header"
                size="sm"
                aria-pressed={hidden}
                aria-label="hide metric"
                className={styles.icon}
              />
            )}
            <IconButton
              name="trash-alt"
              surface="header"
              size="sm"
              className={styles.icon}
              onClick={onRemoveClick || noop}
              disabled={!onRemoveClick}
              aria-label="remove metric"
            />
          </InlineLabel>
        }
      >
        <>{children}</>
      </InlineField>
    </InlineFieldRow>
```

Note that `children` needs to be wrapped into a fragment or smth else as `InlineField` expects one child.
Result:
Screenshot 2020-12-12 at 12 11 31

torkelo: That does not work as everything under children will be seen as a single input and will not "line break" (wrap to next row) on smaller screens. 

For query editors, I think we need to support a list of InlineLabel that can wrap and have the correct margins, but tricky. Not sure what the correct building blocks for this is. 

Tried changing the InlineField to have an optional children, the only problem with that is that InlineField has right margin (which gf-form did not have). This right margin creates a double rigth margin when the InlineField only has a label and no child (input)

torkelo: think the best way is probably to create a new component that is similar to InlineField named InlineSegmentGroup or InlineLabelGroup that only has a bottom margin (so similar to gf-form), that represents a pair or a group of labels that will wrap together.

Clarity-89: It seems like it doesn't wrap the inputs on smaller screens for the old inline styles either, so that seems like a new behavior that needs to be added to the inline components, but yeah seems tricky.

torkelo: Updated https://github.com/grafana/grafana/pull/29782 with another approach

torkelo: It does wrap gf-form elements inside an inline wrapper

Elfo404:
closing this in favor of https://github.com/grafana/grafana/pull/29782

PR #29779, opened 2020-12-11
QueryEditorRow: Ability to change query name

### LINKED ISSUE #28999: Expressions: Support Renaming RefIds
Backend expressions refer to other queries by RefId. lf we support making the refIds editable, it makes for more understandable expressions.

Also, it seems to make the "Filter by RefId" transform nicer as well.

I experimented with this by editing the Panel JSON, and I haven't found any issues yet having the RefID be strings other than a single letter:

image

torkelo:
Hovering over query name shows pen icon and input like background. Clicking changes query name into an input with autofocus. 

* [x] Validate name (remove white space). (Any hard requirement in expression to no allow white space in query names?) 
* [x] Validate against duplicate names
* [x] Finalize UX 
* [x] Add keyboard navigation
 
Screenshot from 2020-12-11 17-27-01
Screenshot from 2020-12-11 17-26-37


kylebrandt:
Works for me :-)

image

Currently lets you create a duplicate name, which we probably should not allow:

image

peterholmberg (changes requested):
If you add a space in the refId name, expressions stop working. I think refId needs to be a `{ id: string, displayName: string} ` or changing the name will always break expressions. 

Another case for this is when we have the refId picker, then changing the name will break those dropdowns as well.

Screenshot from 2020-12-14 09-26-44

kylebrandt:
> If you add a space in the refId name, expressions stop working.

The lexer/parser doesn't support spaces for variables currently. Aware of this, thinking of adding `${...}` syntax.

peterholmberg:
> > If you add a space in the refId name, expressions stop working.
> 
> The lexer/parser doesn't support spaces for variables currently. Aware of this, thinking of adding `${...}` syntax.

Yeah, I figured everything with spaces in it would break a parser.

oddlittlebird:
Where are the doc updates for this feature?

torkelo:
@oddlittlebird early stage v8 feature behind a feature flag, will wait with docs until we know how this will look and behave.

oddlittlebird:
> @oddlittlebird early stage v8 feature behind a feature flag, will wait with docs until we know how this will look and behave.

Thanks for clarifying. I saw the 7.4 tag and thought it was for the next release.

torkelo:
@kylebrandt think this is ready for some more testing 

Added validation (duplicate and empty), trimming / removing white space.

kylebrandt (approved):
Working well for me. One thing that would be nice (but could be in a different PR) would be if with expressions and the reduce and resample operations, if the selected "Input" changed when a corresponding refId is renamed.

So for example, when renaming my first query from "Random Walk Query" to "RW Query" the options for Input under my second query, "avg Walk" needs to be manually reselected:

image

(The input options do update with the new name however).

torkelo:
Yea having some special logic to update references in expressions could be done in another PR

peterholmberg (approved):
Looks good! The only thing I'm missing is something to tell the user that they need to wrap the query name in `${..}` if the query name has a space. Maybe we could add it in the placeholder.

PR #29781, opened 2020-12-11
Chore: Enable more go-ruleguard rules

aknuds1:
Enable more go-ruleguard rules and fix ensuing warnings.


pkg/util/proxyutil/proxyutil_test.go
```diff
@@ -11,9 +11,9 @@ func TestPrepareProxyRequest(t *testing.T) {
 	t.Run("Prepare proxy request should clear X-Forwarded headers", func(t *testing.T) {
```
kylebrandt, line 11: I am curious about this. Generally proxies append to X-Forwarded-For behavior is to append. Maybe for security reasons we want to clear them? 

So this is changing the behavior (to match what the test says, but I wonder if it says the right thing)?

kylebrandt (approved):

aknuds1:
I made a smaller PR just extracting the conversion to Header.Set, #29804, to make sure we review those changes properly.

PR #29782, opened 2020-12-11
InlineForms: Changes to make inline forms more flexible for query editors 

torkelo:
Introduced InlineGroupSegment to group inline elements together, only has bottom margin. 

Screenshot from 2020-12-12 12-03-40

Screenshot from 2020-12-12 12-03-48

Screenshot from 2020-12-12 12-05-13

### DIFF
--- a/packages/grafana-ui/src/components/Forms/InlineField.tsx
+++ b/packages/grafana-ui/src/components/Forms/InlineField.tsx
@@ -59,19 +59,11 @@ const getStyles = (theme: GrafanaTheme, grow?: boolean) => {
     container: css`
       display: flex;
       flex-direction: row;
-      align-items: center;
+      align-items: flex-start;
       text-align: left;
       position: relative;
       flex: ${grow ? 1 : 0} 0 auto;
-      margin: 0 ${theme.spacing.xs};
-    `,
-    wrapper: css`
-      display: flex;
-      width: 100%;
-    `,
-
-    fillContainer: css`
-      flex-grow: 1;
+      margin: 0 ${theme.spacing.xs} ${theme.spacing.xs} 0;
     `,
   };
 };
--- a/packages/grafana-ui/src/components/Forms/InlineField.tsx
+++ b/packages/grafana-ui/src/components/Forms/InlineField.tsx
@@ -63,7 +63,7 @@ const getStyles = (theme: GrafanaTheme, grow?: boolean) => {
       text-align: left;
       position: relative;
       flex: ${grow ? 1 : 0} 0 auto;
-      margin: 0 ${theme.spacing.xs} ${theme.spacing.xs} 0;
+      margin: 0 ${theme.spacing.xs};
     `,
     wrapper: css`
       display: flex;
--- a/packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
+++ b/packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
@@ -1,0 +1,36 @@
+import React, { FC } from 'react';
+import { cx, css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { useTheme } from '../../themes';
+
+export interface Props {
+  grow?: boolean;
+  className?: string;
+}
+
+export const InlineSegmentGroup: FC<Props> = ({ children, className, grow, ...htmlProps }) => {
+  const theme = useTheme();
+  const styles = getStyles(theme, grow);
+
+  return (
+    <div className={cx(styles.container, className)} {...htmlProps}>
+      {children}
+    </div>
+  );
+};
+
+InlineSegmentGroup.displayName = 'InlineSegmentGroup';
+
+const getStyles = (theme: GrafanaTheme, grow?: boolean) => {
+  return {
+    container: css`
+      display: flex;
+      flex-direction: row;
+      align-items: flex-start;
+      text-align: left;
+      position: relative;
+      flex: ${grow ? 1 : 0} 0 auto;
+      margin-bottom: ${theme.spacing.xs};
+    `,
+  };
+};
--- a/packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
+++ b/packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
@@ -8,6 +8,7 @@ export interface Props {
   className?: string;
 }
 
+/** @beta */
 export const InlineSegmentGroup: FC<Props> = ({ children, className, grow, ...htmlProps }) => {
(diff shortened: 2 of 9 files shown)


public/app/plugins/datasource/elasticsearch/components/QueryEditor/QueryEditorRow.tsx
```diff
@@ -23,9 +22,9 @@ export const QueryEditorRow: FunctionComponent<Props> = ({
   const styles = getStyles(theme);
 
   return (
-    <fieldset className={styles.root}>
-      <div className={getInlineLabelStyles(theme, 17).label}>
-        <legend className={styles.label}>{label}</legend>
+    <InlineFieldRow>
```
Clarity-89, line 25: `FieldRow` was designed to be used with `InlineFields`, so it might not play well with using Labels directly.

torkelo:
@Clarity-89 pushed an update that adds a new component InlineSegmentGroup that is more similar to g-form to allow grouping label elements together. Only has bottom margin. 

Using this I can group elements to gether and have them wrap together and have correct bottom margins even when wrapped. 

Screenshot from 2020-12-12 12-03-40

Screenshot from 2020-12-12 12-03-48

packages/grafana-ui/src/components/Forms/InlineField.tsx
```diff
@@ -59,19 +59,11 @@ const getStyles = (theme: GrafanaTheme, grow?: boolean) => {
...
       margin: 0 ${theme.spacing.xs} ${theme.spacing.xs} 0;
     `,
-    wrapper: css`
-      display: flex;
-      width: 100%;
-    `,
-
-    fillContainer: css`
-      flex-grow: 1;
-    `,
```
torkelo, line 75: these seemed to be unused

packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { cx, css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { useTheme } from '../../themes';
+
+export interface Props {
+  grow?: boolean;
+  className?: string;
+}
+
+export const InlineSegmentGroup: FC<Props> = ({ children, className, grow, ...htmlProps }) => {
```
Clarity-89, line 11: The component looks like `InlineFieldRow` but with a bit different styling. Could we maybe use `InlineFieldRow` here and pass it the styles instead, so `InlineSegmentGroup` would be just a light wrapper around `InlineFieldRow`?

Clarity-89 (commented):
Looks good in general, but have a few questions/concerns: 
1. The name `InlineSegmentGroup` implies that this component should be used only for grouping segment elements, what about other input-related components? 
2. Looking at the screenshot I realised that labels and segments look exactly the same, so it's hard to distinguish which one is which, this becomes even more confusing when they wrap on multiple lines. Should we somehow make them look a bit different?

packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
torkelo: > The name InlineSegmentGroup implies that this component should be used only for grouping segment elements, what about other input-related components?

Yea, not sure what the best name is. 
* InlineElementGroup?

> Looking at the screenshot I realized that labels and segments look exactly the same, so it's hard to distinguish which one is which, this becomes even more confusing when they wrap on multiple lines. Should we somehow make them look a bit different? 

Segments and labels are the same, all used the gf-form-label class in the old styles. 

InlineFieldRow has flex-wrap: wrap so we cannot use that, and it has no margin-bottom. Using InlineFieldRow and overriding styles seems like a bad option as this is a very common pattern (to have list of labels and no inputs, some needs to be grouped together to wrap together and that is where we need this component) 

It is more similar to InlineField than InlineFieldRow, but InlineField has margin-right and requires a label + child which makes it impossible to use for label groups that might be only one label or sometimes 2-3 labels and no input and with labels have the right margin built-in makes it impossible to use InlineField with no child or with InlineLabel as a child (it will create double margin right)

Clarity-89: > Yea, not sure what the best name is.
InlineElementGroup?

Maybe just `InlineGroup`?

aocenas:
@Elfo404 you may have the most feedback right now considering the elastic refactor.

packages/grafana-ui/src/components/Forms/InlineSegmentGroup.tsx
torkelo: wonder if we should rename InlineFieldRow to InlineFormRow ?

torkelo: Think InlineGroup is maybe too unclear, InlineSegmentGroup feels more clear that it's a group of inline form segment elements (labels or inputs).

Clarity-89: > wonder if we should rename InlineFieldRow to InlineFormRow ?

To me it suggests that we're using it inside a form, which we don't do for inline elements.

Elfo404:
@torkelo Aside from the placement of the icons which is easily fixable, I think the main problem remains the usage of buttons inside a `label` element. you can see it here:
- when hovering on a label, the first inner iconbutton gets `:hover` as well (this also happens when hovering the second button as it will still be a child of the label)
- clicking on the label will trigger `onClick` on the first child iconbutton as the label is effectively labeling the first child button`


Kapture 2020-12-15 at 15 38 54

I'm trying to figure out if this is something "fixable" specifying a combination of `pointer-events: none` and `auto`

torkelo:
that is super strange, what is causing the hovering of the other button? and the onClick?

Elfo404:
@torkelo that's because `button` is a children of `label`, as per the HTML spec the first "labelable element" (such our button) becomes controlled by the label itself.

torkelo:
@Elfo404 ah, dam. Wonder how to solve that. We could create a new component that is identical to InlineLabel but not a label tag, the other option is maybe to have an optional prop to change InlineLabel to use a div tag . That might be best?

Elfo404:
@torkelo created https://github.com/grafana/grafana/pull/29876 starting from your suggestion and I thought it was an interesting approach, but as mentioned there we can scope `as` down to:
```ts
as?: "div" | "span";
```
or just as boolean prop. wdyt?

Elfo404 (approved):
Should be good to go now:
latest

PR #29794, opened 2020-12-13
Variables: Make TextBox variable width adjustable.

### LINKED ISSUE #29672: The text box is too small! (variables)

Please make the text input variable box a configurable size. Either an explicit size when setting it up or allow us to chose between small/medium/large.

**Why is this needed**:

The current text input variable is set to 192px which is far too small, for example:

I feel like this was a deliberate decision decision so we don't create a generic query box for our dashboards and users only input a targeted value which is then used in the underlying query.

However for data sources like Elastic & especially when building log search dashboards you  want to give the user a bit more flexibility in what they input and far better to leave it up to the designer of the dashboard to decide how large the text input should be.

STEELBADGE:

This PR will make the TextBox variable width adjustable.

When creating dashboards especially ones to view logs it is often desirable to have a TextBox variable to allow a user to do complex queries to the data source and so the width of the TextBox needs to be wide to accommodate this.

However sometimes you don't want them to do this and the TextBox should be small so they only enter a small variable.

But ultimately what we need is the choice to decide how wide the TextBox should be. It should be up to the designer of the dashboard to decide this as they will know best.

Fixes #29672

The default width of the TextBox is 192px so this is why I've set 192 as the default value in the reducer.

### DIFF
--- a/docs/sources/variables/variable-types/add-text-box-variable.md
+++ b/docs/sources/variables/variable-types/add-text-box-variable.md
@@ -23,5 +23,6 @@ _Text box_ variables display a free text input field with an optional default va
 ## Enter Text options
 
 1. (optional) In the **Default value** field, select the default value for the variable. If you do not enter anything in this field, then Grafana displays an empty text box for users to type text into.
+1. In the **Width** field, enter the width the Text Box should be if you need slightly more or less space for your users to enter their text. 
 1. In **Preview of values**, Grafana displays a list of the current variable values. Review them to ensure they match what you expect.
 1. Click **Add** to add the variable to the dashboard.
--- a/public/app/features/variables/editor/VariableTextField.tsx
+++ b/public/app/features/variables/editor/VariableTextField.tsx
@@ -2,6 +2,7 @@ import React, { FormEvent, PropsWithChildren, ReactElement } from 'react';
 import { InlineField, Input, PopoverContent } from '@grafana/ui';
 
 interface VariableTextFieldProps {
+  type?: string;
   value: string;
   name: string;
   placeholder: string;
@@ -16,6 +17,7 @@ interface VariableTextFieldProps {
 }
 
 export function VariableTextField({
+  type,
   value,
   name,
   placeholder,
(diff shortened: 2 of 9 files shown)


STEELBADGE:
The license check seems to be a bit flaky...passed on previous commits but doesn't seem to have run on the latest one - what should I do?

docs/sources/variables/variable-types/add-text-box-variable.md
```diff
@@ -23,5 +23,6 @@ _Text box_ variables display a free text input field with an optional default va
 ## Enter Text options
 
 1. (optional) In the **Default value** field, select the default value for the variable. If you do not enter anything in this field, then Grafana displays an empty text box for users to type text into.
+1. In the **Width** field, enter the width the Text Box should be if you need slightly more or less space for your users to enter their text.
```
oddlittlebird, line 26:
```suggestion
1. In the **Width** field, enter the text box width if you need slightly more or less space for your users to enter their text. 
```

oddlittlebird: I recommend adding "The default width is XX." to the end of this step.

oddlittlebird:
Note to reviewers, we will need to update imag

> The license check seems to be a bit flaky...passed on previous commits but doesn't seem to have run on the latest one - what should I do?

Everything seems to be flaky atm. The license check looks fine now. If it still is problematic after this PR is reviewed, then the devs will help you troubleshoot it.

torkelo:
Is another option to make the width dynamic as you type?

STEELBADGE:
That works too - do we just need to set the grow flag on the input component to be true for that to work?

torkelo:
No there is no such support built into the Input component

hugohaggmark (changes requested):
I think the approach where the input grows is preferable to adding more properties to the variable system.

anthonyalayo:
Would it be possible to revive this? I would love to see this feature merged!

BigUV:
@hugohaggmark Can you explain why this PR haven't been merge ?
The super small filter field is so annoying..

freesinger:
Really confused why this simple feature hadn't been taken.
And it really frustrates me even in v9.2.0

juanicabanas:
@torkelo this FR was requested [here](https://github.com/grafana/grafana/issues/29672)
what about setting this properties in <Input /> component?
minWidth={21}
width={updatedValue?.length || 21}

I wonder if this would be correct, since the Input component multiplies the width by 8

torkelo:
@juanicabanas the input should adjust to the text entered (with some max width hardcoded as percentage of screen width )

PR #29795, opened 2020-12-13
Sliders: Update behavior and style tweak

torkelo:
* [x] Adds cursor to slider drag handle so you know you can drag it (Was missing)
* [x] Adds hover state, a small transparent colored area that extends the handle dot, looks like a common design pattern for sliders
* [x] Makes the slider trigger onChange as you drag so you see the effect while you drag and not after let go, this makes the sliders actually have an advantage over other input alternatives. As it was it was it had no real benefit to dropdown or radio. 
* [x] GraphNG: Adjust the graph opacity steps to be much smaller. To have the control that feels good we need 0.025 steps, the values for this input are then really small. Wonder if we should make the option a percent (0-100) instead of 0-1 to make the values easier to interpret. so instead of having 0.075 it would  be 7.5%

Screenshot show when I hover over the dot (or drag), or focused via keyboard (all the same right now). Before there was no visual difference for active, focus, click/drag etc. 
Screenshot from 2020-12-13 10-31-19
Screenshot from 2020-12-13 10-29-57


packages/grafana-ui/src/components/Slider/styles.ts
```diff
@@ -33,28 +38,31 @@ export const getStyles = stylesFactory((theme: GrafanaTheme, isHorizontal: boole
...
       .rc-slider-handle {
-        border: solid 2px ${theme.palette.blue77};
-        background-color: ${theme.palette.blue77};
+        border: none;
+        background-color: ${blueColor};
+        cursor: pointer;
       }
       .rc-slider-handle:hover {
-        border-color: ${theme.palette.blue77};
+        box-shadow: 0px 0px 0px 6px ${blueOpacity};
```
Clarity-89, line 46: Should this be abstracted somewhere as it used many times?

torkelo: yes, should be moved into a variable

leeoniya:
> Wonder if we should make the option a percent (0-100) instead of 0-1 to make the values easier to interpret. so instead of having 0.075 it would be 7.5%

similar to the situation with the new `pointSize` vs the old `pointRadius` -- the required size range is odd that maps properly to the old radius range. probably worth tackling as part of this PR, too?

i think in both cases, the raw values are not too useful (and can be ugly). maybe just decide on some multiple-of-10 qty of steps for each slider and figure out the raw mapping in the code for each.

dprokop (approved):
These are great improvements @torkelo!

Issue #29796, opened 2020-12-13
Provisioning: dashboards are not reloaded when dashboard update information is not available by the underline operating system


cristi-:
**What happened**:

We have our grafana configured with provisioned dashboards and a mysql database backend for annotation storage. Grafana is built as a docker container using bazel, which sets the unix time 0 on provisioned dashboards. When changes are made to the dashboards they are not updated in the database. It looks like the decision on wether to consider the dashboards as up to date or not doesn't take into consideration all the possibilities for the file checksum: https://github.com/grafana/grafana/blob/master/pkg/services/provisioning/dashboards/file_reader.go#L202-L212

**What you expected to happen**:

Existing dashboards are updated when changes are made.

**How to reproduce it (as minimally and precisely as possible)**:

To reproduce this i ran the following:
- cloned grafana from master
- built the grafana/grafana:dev container
 ```
cd grafana
make build-docker-full
```
- using the existing `ha_test` docker compose files, we'll set the provisioned dashboard timestamp on disk to 0 (this is how the files would look like when bazel builds containers)
```
cd devenv/docker/ha_test/
find grafana/provisioning/dashboards -print | while read filename; do touch -t 197001010000 "$filename"; done
```
- add `/etc/hosts` entries and start grafana:
```
cat <<EOF >> /etc/hosts
127.0.0.1       grafana.loc
127.0.0.1       prometheus.loc
EOF
```

```
docker-compose-up -d
``` 
- at this point loading http://grafana.loc and looking at the mysql overview dashoboard in http://grafana.loc/dashboards , everything looks good:

![image](https://user-images.githubusercontent.com/11157105/102690275-e725ff80-420c-11eb-9562-e6b119672c96.png)

- stop  grafana
```
docker-compose stop
```
- make a change to the provisioned mysql dashboard and set the time to 0 again:

```
sed -i 's/MySQL Overview/Updated MySQL Overview/g' grafana/provisioning/dashboards/mysql/overview.json
find grafana/provisioning/dashboards -print | while read filename; do touch -t 197001010000 "$filename"; done
```

```
grep "MySQL Overview" grafana/provisioning/dashboards/mysql/overview.json
  "title": "Updated MySQL Overview",
stat grafana/provisioning/dashboards/mysql/overview.json
Device: 801h/2049d	Inode: 1015306598  Links: 1
Access: 1970-01-01 00:00:00.000000000 +0000
Modify: 1970-01-01 00:00:00.000000000 +0000
[...]

PR #29797, opened 2020-12-13
Provisioning: Use dashboard checksum field as change indicator

cristi-:

This PR helps to ensure provisioned dashboards are reloaded even if the file mtime hasn't changed but it's checksum has.

https://github.com/grafana/grafana/issues/29796

Fixes #


wbrowne (commented):
Thanks for the contribution @cristi-! 😄  It would be great if you could add a test for this and then it should be good to go

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -148,6 +148,47 @@ func TestDashboardFileReader(t *testing.T) {
...
+					},
+				}
+
+				reader, err := NewDashboardFileReader(cfg, logger)
+				So(err, ShouldBeNil)
+
+				err = reader.startWalkingDisk()
+				So(err, ShouldBeNil)
+				So(len(fakeService.inserted), ShouldEqual, 1)
+			})
```
aknuds1, line 191: Should be a blank line between tests.
```suggestion
			})

```

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -148,6 +148,47 @@ func TestDashboardFileReader(t *testing.T) {
...
 			})
 
+			Convey("Older dashboard file with the same checksum will not replace old dashboard", func() {
+				cfg.Options["path"] = oneDashboard
+				absPath, _ := filepath.Abs(oneDashboard + "/dashboard1.json")
+				stat, _ := os.Stat(oneDashboard + "/dashboard1.json")
+				file, _ := os.Open(absPath)
+				defer file.Close()
+
+				checksum, _ := util.Md5Sum(file)
```
aknuds1, line 158:
```suggestion
				absPath, err := filepath.Abs(oneDashboard + "/dashboard1.json")
				So(err, ShouldBeNil)
				stat, err := os.Stat(oneDashboard + "/dashboard1.json")
				So(err, ShouldBeNil)
				file, err := os.Open(absPath)
				So(err, ShouldBeNil)
				defer file.Close()

				checksum, _ := util.Md5Sum(file)
```

pkg/services/provisioning/dashboards/file_reader_test.go
```diff
@@ -148,6 +148,47 @@ func TestDashboardFileReader(t *testing.T) {
...
+
+				err = reader.startWalkingDisk()
+				So(err, ShouldBeNil)
+				So(len(fakeService.inserted), ShouldEqual, 0)
+			})
+
+			Convey("Older dashboard file with different checksum will replace old dashboard", func() {
+				cfg.Options["path"] = oneDashboard
+				absPath, _ := filepath.Abs(oneDashboard + "/dashboard1.json")
+				stat, _ := os.Stat(oneDashboard + "/dashboard1.json")
```
aknuds1, line 177:
```suggestion
				absPath, err := filepath.Abs(oneDashboard + "/dashboard1.json")
				So(err, ShouldBeNil)
				stat, err := os.Stat(oneDashboard + "/dashboard1.json")
				So(err, ShouldBeNil)
```

aknuds1 (changes requested):

aknuds1:
Please also merge with master, since it's out of date and there will be build errors.

pkg/services/provisioning/dashboards/file_reader.go
```diff
@@ -207,8 +207,8 @@ func (fr *FileReader) saveDashboard(path string, folderID int64, fileInfo os.Fil
 		return provisioningMetadata, nil
 	}
 
-	if provisionedData != nil && jsonFile.checkSum == provisionedData.CheckSum {
-		upToDate = true
+	if provisionedData != nil {
+		upToDate = jsonFile.checkSum == provisionedData.CheckSum
```
aknuds1, line 211: I think if we should ignore mtime (which I'm unsure about), line 202 should be changed instead, since this change makes the usage of `resolvedFileInfo.ModTime()` ineffective.

aknuds1 (changes requested):

pkg/services/provisioning/dashboards/file_reader.go
wbrowne: Or we update it to

```
upToDate = upToDate || jsonFile.checkSum == provisionedData.CheckSum

cristi-: Won't this change cause the same issue if the file has an older timestamp but a different checksum? The initial time check for `upToDate` will be `true` but the checksum will be different, making the final `upToDate` being `false` again, causing the provisioned dashboard not to be saved in the database.

cristi-: I removed the `mtime` check since it did feel redundant and added a few more tests to catch all 4 possible cases (older/newer file on disk and the same/different checksum).

pkg/services/provisioning/dashboards/file_reader_test.go
cristi-: Thank you!

aknuds1:
The referenced issue needs more info, as I indicated there before we can consider approving this PR, since we need to understand the motivation.

aknuds1:
I cleaned up the code a bit, for the sake of easy understanding of the changes, and cleanup in the tests (there was a typo among other things).

aknuds1 (approved):
LGTM! All it changes AFAICT is that only checksum is used to tell if the provisioned dashboard has changed. In my book that's the right thing to do. Thanks for adding tests!

wbrowne (approved):
Yep looks good to me 👍

PR #29804, opened 2020-12-14
Chore: Use Header.Set method instead of Header.Add

aknuds1:
Smaller PR extracted from #29781, converting calls to [Header.Add](https://golang.org/pkg/net/http/#Header.Add) to [Header.Set](https://golang.org/pkg/net/http/#Header.Set) since generally we want to set a header value and not append to it.

Please look it over making sure I'm right in my thinking that all these cases can be converted to Header.Set, i.e. that we want to set the header value rather than appending to it.

### DIFF
--- a/pkg/api/avatar/avatar.go
+++ b/pkg/api/avatar/avatar.go
@@ -107,13 +107,13 @@ func (a *CacheServer) Handler(ctx *models.ReqContext) {
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


marefr (approved):
LGTM

Regarding the comment, add make more sense but if you want to remove the lint annotation go ahead and use set instead.

PR #29805, opened 2020-12-14
Grafana-UI: Support optgroup for MultiSelect

Clarity-89:
Currently `MultiSelect` doesn't show values selected from an optgoup as it only maps options one level deep in the options array (optgroups have a nested `options` array). This PR adds recursive mapping for the options to traverse nested options as well.

### DIFF
--- a/packages/grafana-ui/src/components/Select/Select.story.tsx
+++ b/packages/grafana-ui/src/components/Select/Select.story.tsx
@@ -175,7 +175,7 @@ export const MultiPlainValue = () => {
   );
 };
 
-export const MultiOptgroupSelect = () => {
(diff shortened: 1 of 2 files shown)


packages/grafana-ui/src/components/Select/Select.story.tsx
```diff
@@ -175,6 +175,26 @@ export const MultiPlainValue = () => {
   );
 };
 
+export const MultiOptgroupSelect = () => {
```
dprokop, line 178:
```suggestion
export const MultiSelectWithOptionGroups = () => {
```

dprokop (approved):
Looks and works good, thanks! Left one suggestion

PR #29807, opened 2020-12-14
Elasticsearch: add datasource's configured timeField to query model

### LINKED ISSUE #29788: Elasticsearch: Query requests no longer handled in backend
**What happened**:

After the latest refactoring of the Elasticsearch plugin, queries seems to no longer run on the backend. 
image

Outputted error on console: 
`logger=alerting.evalContext ruleId=9 name="Panel Title alert" error="tsdb.HandleRequest() error type assertion to string failed" changing state to=alerting`

This is a screenshot of what the expected output is to look like before the refactoring. It is expected to output the correct object with the query request and response.
image

Using the devenv, set up the Elasticsearch 7 sample datas and create a dashboard panel with Alerts. 
When viewing the `Test rule`, it outputs the error shown in the screenshot above.

**Environment**:

- Grafana version: Grafana 7.3.x
- Data source type & version: Elasticsearch 7.9.1
- User OS & Browser: MacOS & Google chrome

cc. @ahlaw @alolita

Elfo404:
After the React migration, `timeField` was not being persisted in the query model. Queries were working on the frontend because the query builder sets it every time before running a query, but it breaks alerting as the backend only look for it in the query model.

With this we ensure that `timeField` is always persisted in the query model with the one configured in datasource settings.

Fixes #29788

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/components/QueryEditor/ElasticsearchQueryContext.test.tsx
+++ b/public/app/plugins/datasource/elasticsearch/components/QueryEditor/ElasticsearchQueryContext.test.tsx
@@ -1,5 +1,6 @@
 import React, { FunctionComponent } from 'react';
 import { renderHook } from '@testing-library/react-hooks';
+import { render } from '@testing-library/react';
 import { ElasticsearchProvider, useDatasource, useQuery } from './ElasticsearchQueryContext';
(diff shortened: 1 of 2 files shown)


ivanahuckova (approved):
LGTM. Thanks for adding helpful comments!

PR #29808, opened 2020-12-14
DashboardDS: Fixes display of long queries

### LINKED ISSUE #25126: SharedQuery: Dashboard data source query editor UI does not handle large query models
**What happened**:
When constructing a panel using the `-- Dashboard --` data source, the panel configuration is mangled, and unreadable.

The panel configuration should be intelligible.

Create one panel with multiple queries (the more complex, the better) such as:
image
Create another panel using the `-- Dashboard --` data source and the first panel:
image
Notice how the text between each row is overlapping? The more text, the messier it becomes.

**Environment**:
- Grafana version: 7.0.0
- Data source type & version: PostgreSQL 12
- OS Grafana is installed on: CentOS/8
- User OS & Browser: Linux Chrome
- Grafana plugins:
- Others:

hugohaggmark:
Refactored out the rows to their own component and using emotion instead of inline styles.

**Before, short queries**
image

**After, short queries** 
image

**Before, long queries**
image

**After, long queries** 
image

Fixes #25126

### DIFF
--- a/public/app/plugins/datasource/dashboard/DashboardQueryEditor.tsx
+++ b/public/app/plugins/datasource/dashboard/DashboardQueryEditor.tsx
@@ -1,27 +1,18 @@
-// Libraries
 import React, { PureComponent } from 'react';
-
-// Types
-import { LegacyForms, Icon } from '@grafana/ui';
-import { DataQuery, DataQueryError, PanelData, DataFrame, SelectableValue } from '@grafana/data';
-import { DashboardQuery } from './types';
-import config from 'app/core/config';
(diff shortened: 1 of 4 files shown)


torkelo (approved):
Looks good but discovered a bug (unrelated to this PR). 

The queries do not show the second time. For example load this:
http://localhost:3000/d/ZqZnVvFZz/datasource-tests-shared-queries?tab=query&editPanel=4&orgId=1 

Then go Transform tab and back, the queries are gone, same if you exit and go back to edit mode. But this looks like it's caused by my recent changes to Query tab and data source loading as I get the same bug in master. Will open an issue for it and assign myself

PR #29813, opened 2020-12-14
GraphNG: Load uPlot path builders lazily

dprokop:
Ref https://github.com/grafana/grafana/pull/29763#issuecomment-743358129

Fixes uPlot.paths breaking the entire shebang


torkelo (approved):
Looks good!

leeoniya:
this PR caused the path builders to be initialized per plot, per series, per every redraw; there was a reason they were initialized only once, globally for the lifecycle of the whole app.

please try to keep the original intent intact rather than rushing through Jest-related fixes :)

PR #29816, opened 2020-12-14
Expressions: Replace query input fields with select.

### LINKED ISSUE #29164: Expressions: Change RefId input fields to dropdowns
The reduce and resample operations take a RefId as their input.

Currently these are text fields, it would clear what to put in them, and a more fluid UI if these were dropdowns of existing refIds.

I do not know the frontend well enough to know if this as a task should be grouped with renaming refids https://github.com/grafana/grafana/issues/28999 as updates there would also need to update the dropdowns.

peterholmberg:
Replaces inputs with a select

Fixes #29164

- Is this a sensible way of getting queries?
- Should we make it a component, `<RefIdPicker />` ?
- Memoize the selectable values probably?

### DIFF
--- a/packages/grafana-data/src/types/datasource.ts
+++ b/packages/grafana-data/src/types/datasource.ts
@@ -339,6 +339,7 @@ export interface QueryEditorProps<
   range?: TimeRange;
   exploreId?: any;
   history?: HistoryItem[];
+  queries?: DataQuery[];
 }
 
 export enum DataSourceStatus {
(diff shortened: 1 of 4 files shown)


kylebrandt:
Seems to work :-)

- I think "Input" is better than "Query", since the input can be other expressions. 
- With renaming, will want the fields to be a bit wider I think:

image

torkelo (approved):
Looks goot!

PR #29819, opened 2020-12-14
Expressions: support ${my var} syntax

kylebrandt:
Allow variables with spaces if in curly braces, e.g. `${my var}`. This is so expressions are less limited with what refId names can be when that (https://github.com/grafana/grafana/pull/29779) is merged.


papagian (approved):
LGTM

Issue #29324, opened 2020-11-24
Allow to configure cookie domain


n0emis:
**What would you like to be added**:
A config option to specify the cookie_domain

**Why is this needed**:
When having a grafana-installation with multiple domains configured, logging in with oauth on a domain, unequal to the root_url, logging in fails, because the state-cookie is only set for the first  domain.

PR #29823, opened 2020-12-14
Login: OAuth: Allow logging in from all domains

n0emis:
When having multiple domains configured for grafana (e.g. in a loadbalancing-setup or internal/external domains), it is currently not possible to log in via OAuth on a secondary domain, since the oauth-state cookie is stored on the first url but the oauth-redirect always points to the `root_url`. Therefor the state-cookie cannot be read here. This PR set't the redirect_url to the url on which the user first visited grafana, to allow logging in on all domains.

Being unable to log in via OAuth when visiting grafana on a url != the configured `root_url`.

### DIFF
--- a/pkg/api/login_oauth.go
+++ b/pkg/api/login_oauth.go
@@ -68,6 +68,26 @@ func (hs *HTTPServer) OAuthLogin(ctx *models.ReqContext) {
 		return
 	}
 
+	// Get the URL of the current request
+	scheme := ctx.Req.Header.Get("X-Forwarded-Proto")
+	if scheme == "" {
+		scheme = string(hs.Cfg.Protocol)
+	}
+
+	path := ""
+	uri, err := url.ParseRequestURI(ctx.Req.RequestURI)
+	if err != nil {
+		path = "login/" + name
+	} else {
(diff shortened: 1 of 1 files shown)


n0emis:
This would be (in my opinion better) alternative to #29324, because the other fix would only work if both installations are under the same domain.

SimonDreher:
Is there anything I could help towards getting this merged?

We're also stuck with the same problem of two domains and editing the redirect-url manually works perfectly (so the same as this PR would do automatically)

maxkondr:
@n0emis 
Could you please extend sign out action as well in order to send to OAuth server sign out action and return back to the grafana domain (I am referring to the following config option in grafana.ini):
```
[auth]
signout_redirect_url = https://<auth server>/login/signout?fromURI=https://<grafana domain that originally user used in browser>
```

vincentnonim:
Hi,

This is exactly the feature I'm looking for. I have a similar setup with internal/external domain for the same Grafana instance and the OAuth is only working for 1 domain.
It could be great to have an OAuth redirect url based on Host header or even proxy header like `X-Forwarded-Host`.

PR #29825, opened 2020-12-14
Sparkline: Fixes issue with sparkline that sent in custom fillColor instead of fillOpacity

torkelo:
Introduced by PR merged earlier today where I changed the logic a bit and accidentally removed the feature to send in fillColor but not fillOpacity


dprokop (approved):
Looks good! There is a snapshot that needs an update tho.

PR #29827, opened 2020-12-14
Chore: remove unused datasource status enum

ryantxu:
I stumbled into this enum...  it is not used, so we should remove it


torkelo (approved):
think it was used in explore before, but no longer

Issue #29667, opened 2020-12-07
Cloud monitoring: request sent to the backend with undefined queryType


papagian:
**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
1. create new dashboard
2. add new panel
3. select a cloud monitoring (stackdriver) datasource
4. select a service (eg `Compute`) without selecting anything else eg project, metric
5. request fails because `queryType` is undefined
6. in the query editor select `Metrics` as `QueryType` (now the request has a valid queryType and succeeds)

**Anything else we need to know?**:

**Environment**:
- Grafana version: master (commit 874f2fafb2fa4c56423773902af419335390f140)
- Data source type & version: Google cloud monitoring
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

sunker:
I think this has to do with this line: 
https://github.com/grafana/grafana/blob/922f36b386e21393fddffa22031c5b645e8369ca/public/app/plugins/datasource/cloud-monitoring/components/QueryEditor.tsx#L21

Seems like a typo was introduced in [this](https://github.com/grafana/grafana/commit/546f569e0c467d4ba748b4f390ad289b9c74bec6#diff-71862e079d75143b30ed8b1ed31035e4c988a541e201d9d72cdba5e4af478037) commit

sunker:
Actually I can fix this - I'm already making changes to this file in another PR that will be merged soon

PR #28847, opened 2020-11-05
CloudMonitoring: Support request cancellation properly

hugohaggmark:
**What this PR does / why we need it**:
This PR replaces the usage of datasourceRequest in the Google Cloud Monitoring data source with fetch and updates tests accordingly. I just focused on cancellation of TimeSeries requests, more work needs to be done as soon as #28217 is merged to master.

**Which issue(s) this PR fixes**:
Relates #27222

**Special notes for your reviewer**:
Ping me on Slack for details on how to test this using plugin provisioning.

PR #29830, opened 2020-12-14
Cloud Monitoring: Convert datasource to use Dataframes

ying-jeanne:
**What this PR does / why we need it**: convert google cloud monitoring data sources to use data frames 

Fixes #


sunker:
Looks like a great start! I don't get any timeseries plotted in my graph when testing with [this](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/cloud-monitoring/dashboards/gce.json) dashboard, but I guess frontend changes are necessary for it to work?

ying-jeanne:
> Looks like a great start! I don't get any timeseries plotted in my graph when testing with [this](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/cloud-monitoring/dashboards/gce.json) dashboard, but I guess frontend changes are necessary for it to work?

yes, it is coming with the following commits

pkg/tsdb/cloudmonitoring/cloudmonitoring.go
```diff
@@ -622,35 +633,76 @@ func (e *CloudMonitoringExecutor) parseResponse(queryRes *tsdb.QueryResult, data
...
+			for j, field := range frames[i].Fields {
+				if field.Config == nil {
+					frames[i].Fields[j].Config = &data.FieldConfig{}
+				}
+				deepLink := data.DataLink{
+					Title:       "View in Metrics Explorer",
+					TargetBlank: true,
+					URL:         dl,
+				}
+				frames[i].Fields[j].Config.Links = append(frames[i].Fields[j].Config.Links, deepLink)
```
ying-jeanne, line 694: here is where i put the deeplink and unit which were the logics on the frontend part into backend dataframe.fields.config. I am not sure that we need to put it into all fields.config or just fields[1].config (just for the data not the time serie)

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -31,62 +34,33 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
+    this.intervalMs = request.intervalMs;
+    request.targets = request.targets.map(this.migrateQuery);
+    return super.query(request).pipe(
+      mergeMap((res: DataQueryResponse) => {
+        return from(this.processResponse(request, res));
       })
     );
   }
 
+  async processResponse(
```
ying-jeanne, line 54: keep for now, just in case we need to process anything in this pr, otherwise, i would remove in the end

pkg/tsdb/cloudmonitoring/cloudmonitoring.go
```diff
@@ -34,11 +34,26 @@ var (
...
-	legendKeyFormat   = regexp.MustCompile(`\{\{\s*(.+?)\s*\}\}`)
-	metricNameFormat  = regexp.MustCompile(`([\w\d_]+)\.(googleapis\.com|io)/(.+)`)
-	wildcardRegexRe   = regexp.MustCompile(`[-\/^$+?.()|[\]{}]`)
-	alignmentPeriodRe = regexp.MustCompile("[0-9]+")
+	matchAllCap                 = regexp.MustCompile("(.)([A-Z][a-z]*)")
+	legendKeyFormat             = regexp.MustCompile(`\{\{\s*(.+?)\s*\}\}`)
+	metricNameFormat            = regexp.MustCompile(`([\w\d_]+)\.(googleapis\.com|io)/(.+)`)
+	wildcardRegexRe             = regexp.MustCompile(`[-\/^$+?.()|[\]{}]`)
+	alignmentPeriodRe           = regexp.MustCompile("[0-9]+")
+	cloudMonitoringUnitMappings = map[string]string{
```
ying-jeanne, line 42: after changing the mapping, I am not able to get the same result as frontend mapping. The difference is visible on the UI.
before change i got Mb/s
before_change

After change I got MiB
after_change

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -134,10 +106,19 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
+  applyTemplateVariables(target: CloudMonitoringQuery, scopedVars: ScopedVars): Record<string, any> {
+    target = this.prepareTimeSeriesQuery(target, scopedVars);
+    return {
+      ...target,
+      intervalMs: this.intervalMs,
+      type: 'timeSeriesQuery',
+    };
+  }
+
   getTimeSeries(options: DataQueryRequest<CloudMonitoringQuery>): Observable<PostResponse> {
```
sunker, line 118: Looks like function is not needed anymore since it's only used by getLabels, so move the logic to getLabels.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -134,10 +106,19 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
       .toPromise();
   }
 
+  applyTemplateVariables(target: CloudMonitoringQuery, scopedVars: ScopedVars): Record<string, any> {
```
sunker, line 109: This function is a really nice addition to the api. However, I don't think it does what its name stipulates - the template variables seem to be applied in the prepareTimeSeriesQuery function. Is it possible to remove that function and put all that logic in applyTemplateVariables instead?

pkg/tsdb/cloudmonitoring/cloudmonitoring.go
sunker: That's a good question. I think it needs to be put on all fields.config just like you've done, but it would certainly be neater to put it fields[1].config. Can you try that and see if it's sufficient?

public/app/plugins/datasource/cloud-monitoring/datasource.ts
sunker: :+1:

pkg/tsdb/cloudmonitoring/cloudmonitoring.go
ying-jeanne: I just updated the pr, after testing, it seems we need only to set the frames[i].Fields[1], it is enough.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
ying-jeanne: this is a bit difficult, since the function is also used in interpolateVariablesInQueries, i don't know whether it would break something if i changed the call to preparetimeseriesquery to the call to applyTemplateVariables

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -2,26 +2,27 @@ import _ from 'lodash';
...
-  toDataFrame,
+  DataQueryResponse,
 } from '@grafana/data';
 import { getTemplateSrv, TemplateSrv } from 'app/features/templating/template_srv';
 import { getTimeSrv, TimeSrv } from 'app/features/dashboard/services/TimeSrv';
 
 import { CloudMonitoringOptions, CloudMonitoringQuery, Filter, MetricDescriptor, QueryType } from './types';
-import { cloudMonitoringUnitMappings } from './constants';
-import API, { PostResponse } from './api';
+import API from './api';
```
papagian, line 14: `PostResponse` was introduced [here](https://github.com/grafana/grafana/pull/28847) so it would good if @hugohaggmark could acknowledge that request cancellation is not affected by this PR's modifications.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -31,62 +32,33 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
       })
     );
   }
 
+  async processResponse(
+    request: DataQueryRequest<CloudMonitoringQuery>,
+    res: DataQueryResponse
+  ): Promise<DataQueryResponse> {
+    console.log();
+    // const unit = this.resolvePanelUnitFromTargets(request.targets);
```
papagian, line 57: I think that these lines have been committed accidentally; please remove them.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
hugohaggmark: That change is Ok, and the cancellation seems to work as expected.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -2,26 +2,27 @@ import _ from 'lodash';
...
-import { cloudMonitoringUnitMappings } from './constants';
-import API, { PostResponse } from './api';
+import API from './api';
+import { DataSourceWithBackend } from '@grafana/runtime';
 import { CloudMonitoringVariableSupport } from './variables';
 import { catchError, map, mergeMap } from 'rxjs/operators';
 import { from, Observable, of, throwError } from 'rxjs';
-
-export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonitoringQuery, CloudMonitoringOptions> {
+export default class CloudMonitoringDatasource extends DataSourceWithBackend<
```
hugohaggmark, line 19: Please add a new line between imports and class definition.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -31,62 +32,33 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
-          }
-        }
-
-        return { data: result };
+  query(request: DataQueryRequest<CloudMonitoringQuery>): Observable<DataQueryResponse> {
+    this.intervalMs = request.intervalMs;
+    request.targets = request.targets.map(this.migrateQuery);
+    return super.query(request).pipe(
+      mergeMap((res: DataQueryResponse) => {
+        return from(this.processResponse(request, res));
```
hugohaggmark, line 47: This should be simplified to:
```suggestion
    return super.query(request);
```

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -31,62 +32,33 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
 
+  async processResponse(
+    request: DataQueryRequest<CloudMonitoringQuery>,
+    res: DataQueryResponse
+  ): Promise<DataQueryResponse> {
+    console.log();
+    // const unit = this.resolvePanelUnitFromTargets(request.targets);
+
+    return res;
+  }
```
hugohaggmark, line 60: Please remove this function as it's not needed if you simplify the query function as I suggested previously in the review.

public/app/plugins/datasource/cloud-monitoring/specs/datasource.test.ts
```diff
@@ -213,51 +213,6 @@ describe('CloudMonitoringDataSource', () => {
...
-      });
-      describe('and all target units are not the same', () => {
-        it('should return the default value of undefined', () => {
-          const res = ds.resolvePanelUnitFromTargets([{ unit: 'bit' }, { unit: 'min' }]);
-
-          expect(res).toBeUndefined();
-        });
-      });
-    });
-  });
```
hugohaggmark, line 260: Have these tests moved to the backend now?

hugohaggmark (commented):
This is looking pretty good, but I left some comments that you might want to consider.

public/app/plugins/datasource/cloud-monitoring/specs/datasource.test.ts
ying-jeanne: I moved them here cloudmonitoring_test.go, but since now we have one unit per target(dataframe), i just check whether there is a map gcp unit -> grafana unit, if there is i put the unit in dataframe.config, otherwise it is empty.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
ying-jeanne: removed

public/app/plugins/datasource/cloud-monitoring/specs/datasource.test.ts
hugohaggmark: Ok, thank you.

pkg/tsdb/cloudmonitoring/cloudmonitoring_test.go
```diff
@@ -756,34 +755,34 @@ func TestCloudMonitoring(t *testing.T) {
...
+						So(frames[i].Fields[1].Name, ShouldEqual, "0")
 					}
-					So(len(res.Series[i].Points), ShouldEqual, 2)
+					So(frames[i].Fields[1].Len(), ShouldEqual, 2)
 				}
 
 				Convey("timestamps should be in ascending order", func() {
-					So(res.Series[0].Points[0][1].Float64, ShouldEqual, int64(1550859086000))
-					So(res.Series[0].Points[1][1].Float64, ShouldEqual, int64(1550859146000))
+					// So(frames[0].Fields[0].At(0), ShouldEqual, time.Unix(int64(1550859086000/1000), 0))
```
papagian, line 768: This should not be commented out.

public/app/plugins/datasource/cloud-monitoring/specs/datasource.test.ts
ying-jeanne: hi hugo, I just applied all logics into backend with all the test removed above.

papagian (changes requested):
I have checked metrics, SLO and legacy (created before 6.1) queries, cloud monitoring annotations, alerting cloud monitoring queries as well as some transformations and so far all work fine.
I have some minor comments.

papagian (approved):
Great work!

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -176,8 +114,29 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
         },
       ],
       range: this.timeSrv.timeRange(),
-    } as DataQueryRequest<CloudMonitoringQuery>)
+    } as DataQueryRequest<CloudMonitoringQuery>;
+
+    const queries = options.targets
+      .map(this.migrateQuery)
+      .filter(this.filterQuery)
+      .map(q => this.applyTemplateVariables(q, options.scopedVars));
```
sunker, line 122: I know getLabels used to call getTimeSeries and that function used to call  migrateQuery, filterQUery and applyTemplateVariables before. But that might not be necessary now. The getLabels query is already prepared on line 101, so I think you should be able to remove line 119-122.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -134,33 +93,12 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
-          from: options.range.from.valueOf().toString(),
-          to: options.range.to.valueOf().toString(),
-          queries,
-        });
-      }),
-      map(({ data }) => {
-        return data;
-      })
-    );
+  applyTemplateVariables(target: CloudMonitoringQuery, scopedVars: ScopedVars): Record<string, any> {
```
sunker, line 96: Looks like you can move all the code from prepareTimeSeriesQuery and put it into this function.

public/app/plugins/datasource/cloud-monitoring/datasource.ts
```diff
@@ -31,60 +33,17 @@ export default class CloudMonitoringDatasource extends DataSourceApi<CloudMonito
...
-            }
-            result.push(df);
-          }
-        }
-
-        return { data: result };
-      })
-    );
+  query(request: DataQueryRequest<CloudMonitoringQuery>): Observable<DataQueryResponse> {
+    this.intervalMs = request.intervalMs;
```
sunker, line 44: Not sure I like the way this class variable is being set and then used in prepareTimeSeriesQuery. I would prefer to remove the variable and do something like this instead:
```javascript 
query(request: DataQueryRequest<CloudMonitoringQuery>): Observable<DataQueryResponse> {
    request.targets = request.targets.map(t => ({
      ...this.migrateQuery(t),
      intervalMs: request.intervalMs,
    }));
    return super.query(request);
  }

**Review by sunker — CHANGES_REQUESTED** [review]

File: pkg/tsdb/cloudmonitoring/cloudmonitoring.go [hunk 16]
```diff
@@ -221,16 +232,48 @@ func (e *CloudMonitoringExecutor) executeTimeSeriesQuery(ctx context.Context, ts
...
-			dl = query.buildDeepLink()
+
+		err = e.parseResponse(queryRes, resp, query)
+		if err != nil {
+			queryRes.Error = err
 		}
-		queryRes.Meta.Set("deepLink", dl)
+
+		if len(unit) > 0 {
+			frames, _ := queryRes.Dataframes.Decoded()
```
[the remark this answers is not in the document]
ying-jeanne: the only problem of the parseresponse, it is that it doesn't have access to all request. If we check the logics on frontend, we see whether all requests have the same unit, if not, we don't set the unit into dataframe (timeseries), only if every unit is the same, we set it in the response. that is the reason why the unit is set here, but not the deeplink

sunker:
Had a quick look at the docs, and it seems like there's more we can do when it comes to unit mappings. For example:

> 10^2.% indicates a metric contains a ratio, typically in the range 0..1, that will be multiplied by 100 and displayed as a percentage (so a metric value 0.03 means "3 percent").

This unit is used for CPU Usage, and it would be nice to map that to percentage in Grafana. But that could go in another PR. I'll create an issue for that.

pkg/tsdb/cloudmonitoring/cloudmonitoring.go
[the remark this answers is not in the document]
papagian: Good points! Decoding can be expensive and it's better to be avoided but checking closer I think it's fine as is: `parseResponse()` calls  `NewDecodedDataFrames()` for creating the Dataframes which actually stores the data decoded:
https://github.com/grafana/grafana/blob/dbfee130b1e1c0e84cac20020eff086ee488ca17/pkg/tsdb/models.go#L218-L222

Then, `Dataframes.Decoded()` since data are decoded just returns them without preforming anything expensive:
https://github.com/grafana/grafana/blob/dbfee130b1e1c0e84cac20020eff086ee488ca17/pkg/tsdb/models.go#L248-L258

WDYT?

sunker:
When I try to create a new AnnotationQuery, the services and metrics dropdowns are not populated. I've tested, and it has to do with [this](https://github.com/grafana/grafana/issues/29667#issuecomment-757947308) bug. I'll fix that in another PR, so we can ignore that now.

papagian:
> When I try to create a new AnnotationQuery, the services and metrics dropdowns are not populated. I've tested, and it has to do with [this](https://github.com/grafana/grafana/issues/29667#issuecomment-757947308) bug. I'll fix that in another PR, so we can ignore that now.

👍  I only tested it with an existing annotation.

sunker (approved):
Tested and it works fine. Great work @ying-jeanne!

hugohaggmark (commented):
I agree, outstanding work @ying-jeanne 🎉

PR #29837, opened 2020-12-15
Select: Changes default menu placement for Select from auto to bottom

### LINKED ISSUE #26709: Select component hidden when the dropdown goes up
While working on a datasource query editor, I got confused by the hidden options when the popup goes up:
image

vs down:
image

When things go up, we can not scroll to the top few items

hugohaggmark:
This PR changes the default menu placement for the `Select` component from `auto` to `bottom`.

Fixes #26709

Details: 

The default menu placement for the `Select` component was previously set to `auto`. This meant that the drop down for the select would open downwards/upwards depending on available space. We've noticed that the `auto` doesn't always work when the drop down for `Select` is opened upwards so therefore we've changed the default to `bottom` instead.


torkelo (approved):
Looks good!

PR #29838, opened 2020-12-15
Docs: Add CloudWatch `resource_arns()` Multiple Tag Example

### LINKED ISSUE #29499: CloudWatch: Providing Multiple Tags to `resource_arns()` Returns Incorrect Response
**What happened**:
I'm trying to create a "Query" dashboard variable using a CloudWatch data source.  

I want to populate the options using the [resource_arns()](https://grafana.com/docs/grafana/latest/datasources/cloudwatch) function.

I want to provide multiple tags to the `resource_arns()` function (for example, a resource tagged with "environment:production" and "service:service-name"):

```
resource_arns(default,elasticloadbalancing:loadbalancer,[{"Component":["$service"]},{"Environment":["$environment"]}])
```

This function takes "tags" as an argument, which through some [digging](https://github.com/grafana/grafana/pull/14803/files#diff-0c6eff99719e840378467688417315b8d9ce22cddf885432bb1885d8aa624f7eR653) suggests that the [ResourceGroupsTaggingApi:GetResources](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html) AWS API is used.

This AWS API should accept a group of tag filters.  When multiple tag filters are provided, the returned resources must satisfy all filters.  From the documentation:

> If you specify more than one filter in a single request, the response returns only those resources that satisfy all specified filters.

However, when I try to provide multiple tag filters, it appears that rather than returning resources which satisfy _both_ filters, I'm getting an excessive list of results.  

I think there is an issue with the way that the tags provided as part of `resource_arns()` are being manipulated to be provided to the AWS API call.  I think that it should be possible to pass a valid --tag-filters object to `resource_arns()` and for it to operate as expected.

When I perform a query like this via the **Variables -> Edit** menu on a CloudWatch data source:

```
resource_arns(default,elasticloadbalancing:loadbalancer,[{"Component":["$service"]},{"Environment":["$environment"]}])
```

I should be returned ARNs for resources which have _both_ a tag of "Environment:foo" and "Component:bar".

Instead, I am returned an excessive list of resources which do not match the constraints I have set in the TagFilter.

The AWS API confirms this is possible and returns the correct response, limited to resources that match _both_ constraints (note the slight difference in `--tag-filters` body required):

```
$ environment=foo
$ service=bar

$ aws resourcegroupstaggingapi get-resources --resource-type-filters 'elasticloadbalancing:loadbalancer' --tag-filters '[ { "Key": "Environment", "Values": ["foo"] }, { "Key":"Component", "Values":["bar"] } ]'
{
    "ResourceTagMappingList": [
        {
            "ResourceARN": "arn:aws:elasticloadbalancing:eu-west-1:<omitted>:loadbalancer/net/foo-bar-1234567890ABC/abcd1234abcd1234",
            "Tags": [
                {
                    "Key": "Component",
                    "Value": "bar"
                },
                {
                    "Key": "Environment",
                    "Value": "foo"
                }
            ]
        }
    ]
}
```

I think there is an issue with the way that the tags provided as part of `resource_arns()` are being manipulated to be provided to the AWS API call.  I think that it should be possible to pass a valid --tag-filters object to `resource_arns()` and for it to operate as expected.

Reproduction steps above.

None

**Environment**:
- Grafana version: Grafana v7.3.3 (2489dc4d3a)
- Data source type & version: CloudWatch (Native, same verson)
- OS Grafana is installed on: CentOS7
- User OS & Browser: MacOSX, Firefox
- Grafana plugins: CloudWatch (Native, same verson)
- Others: None

hybby:
**What this PR does / why we need it**: This PR adds a small example to documentation outlining how multiple AWS tags can be used with the `resource_arns()` function of the AWS CloudWatch integration ([link](https://grafana.com/docs/grafana/latest/datasources/cloudwatch/)).  This example would have helped me when using this function by showing me how to specify multiple tags as part of the query.

Fixes #29499 

**Special notes for your reviewer**: Let me know if there's a different format you'd prefer this to take - I see there is one ELB example already, and I wasn't sure if you'd object to there being two with slightly different tag specification formats.

### DIFF
--- a/docs/sources/datasources/cloudwatch.md
+++ b/docs/sources/datasources/cloudwatch.md
@@ -278,6 +278,7 @@ Example dimension queries which will return list of resources for individual AWS
 | `dimension_values(us-east-1,AWS/S3,BucketSizeBytes,BucketName)`                                                               | S3               |
 | `dimension_values(us-east-1,CWAgent,disk_used_percent,device,{"InstanceId":"$instance_id"})`                                  | CloudWatch Agent |
 | `resource_arns(eu-west-1,elasticloadbalancing:loadbalancer,{"elasticbeanstalk:environment-name":["myApp-dev","myApp-prod"]})` | ELB              |
+| `resource_arns(eu-west-1,elasticloadbalancing:loadbalancer,{"Component":["$service"],"Environment":["$environment"]})`        | ELB              |
 | `resource_arns(eu-west-1,ec2:instance,{"elasticbeanstalk:environment-name":["myApp-dev","myApp-prod"]})`                      | EC2              |
 
 ## ec2_instance_attribute examples


wbrowne (approved):
LGTM 👍  Thanks for the contribution @hybby and glad we got to the bottom of your issues!

achatterjee-grafana (approved):
LGTM. Thanks for contributing!

PR #29770, opened 2020-12-11
Elasticsearch: Use time range from requests or default one

alexanderzobnin:
**What this PR does / why we need it**:
This PR removes using timeSrv from elasticsearch data source. When it's possible, data source uses time range provided in the request. Otherwise, `DefaultTimeRange` from `@grafana/data` is used.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:
We need this as a part of work on extracting Elasticsearch data source into a separate plugin.

PR #29841, opened 2020-12-15
Elasticsearch: remove timeSrv dependency for logs context

Elfo404:
Removes the dependency on timeSrv for log rows context as part of https://github.com/grafana/grafana/pull/29770

instead of limiting the contextual logs to the currently selected timeframe, we use the timestamp of the log we are visualizing the context for as a starting point to request logs.

Since not having a predefined time range doesn't allow us to calculate all the indices we need to query, when generating indices without a `from` or a `to` param, we default to generating 7 indices according to the configured index pattern.


alexanderzobnin (approved):
Looks good!

ivanahuckova (approved):
👍 LGTM! I've mentioned on the call my  1 nitpickiest nitpick about rephrasing the comment to clarify that even though both params are optional, we know that we for sure will receive at least 1 - from the log context. But reading the comment again, it basically says so. So no worries about it. Good job!

PR #29844, opened 2020-12-15
Variables: Fixes so clicking on Selected in drop down will exclude All value from selection

### LINKED ISSUE #28751: click selected with nothing selected selects the all item, as well as all the other items.
**What happened**:
In the template variable selection dropdowns, when selecting `selected` with nothing selected, every value is selected including all. 

the all value should not be selected. 

play.grafana.org, clear the selection by clicking `selected` 
image
then click on `selected` again. 
image

nope. 

**Environment**:
- Grafana version: 7.3.1
- OS Grafana is installed on: n/a
- User OS & Browser: firefox
- Grafana plugins: n/a 
- Others:

hugohaggmark:
This PR changes the behaviour when the user clicks `Selected` in a variable drop down menu. Previously it would include the `All` value when selecting all values. With this PR it will change to exclude the `All` value.

Enjoy the review. 

Fixes #28751

### DIFF
--- a/public/app/features/variables/pickers/OptionsPicker/reducer.test.ts
+++ b/public/app/features/variables/pickers/OptionsPicker/reducer.test.ts
@@ -15,7 +15,7 @@ import {
   updateSearchQuery,
 } from './reducer';
 import { reducerTester } from '../../../../../test/core/redux/reducerTester';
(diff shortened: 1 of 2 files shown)


torkelo (approved):
Looks good

PR #29846, opened 2020-12-15
Variables: Fixes so numerical sort order works for null values

### LINKED ISSUE #8762: [Bug] Numerical sorting of template values breaks on null/empty value
Grafana v4.3.2 (commit: ed4d170) with Prometheus

Sorting template values returned from Prometheus labels which are empty leads to the error:

````
Templating
Template variables could not be initialized: Cannot read property 'match' of undefined
````

Sorting Alphabetically works.

This can break existing dashboards in opaque and hard-to-debug ways.

If this is touched, maybe https://github.com/grafana/grafana/issues/7307 can be revisited as well.

### PR DESCRIPTION
**Pull request by hugohaggmark**
This PR fixes so template options with `null` values are sorted without throwing any errors. 

Fixes #8762

### DIFF
--- a/public/app/features/variables/query/reducer.test.ts
+++ b/public/app/features/variables/query/reducer.test.ts
@@ -1,6 +1,6 @@
 import { reducerTester } from '../../../../test/core/redux/reducerTester';
-import { queryVariableReducer, updateVariableOptions, updateVariableTags } from './reducer';
-import { QueryVariableModel } from '../types';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: public/app/features/variables/query/reducer.test.ts [hunk 1]
```diff
@@ -260,6 +260,28 @@ describe('queryVariableReducer', () => {
   });
 });
 
+describe('sortVariableValues', () => {
```
jackw, line 263: 👏 👏 👏

jackw (approved):
Great work Hugo! 👍

PR #29849, opened 2020-12-15
SharedQuery: Fixes shared query editor now showing queries

### LINKED ISSUE #29836: DashboardDataSource: Shared queries do not show the second time
Dashboard query editor is missing queries after navigating away and then back to query tab.

torkelo:
Fixes #29836

This bug is only in master, not in v7.3.x


hugohaggmark (changes requested):
Hmm changing the which panel you get the data from doesn't refresh the UI unless you change tab. Attaching a small clip
issue-with-shared-queries

hugohaggmark (approved):
Looks good and works as expected now ⭐

PR #29851, opened 2020-12-15
GraphNG: Do not set fillColor from GraphNG only opacity

torkelo:
I am not sure what the logic should be here, it's a bit messy.

The Sparkline needs to send in different line + fill color, but in GraphNG we only have series color, which we passed in via both lineColor and fillColor so
there is no easy way in the builder to figure out when set the fill color and to what.

One option is from GraphNG not to set the fillColor and only set lineColor and opacity, that way the SeriesBuilder will only set the fill when fillOpacity > 0 or when fillColor is set. 

An alternative would be to require a fillOpacity = > 0 for any fillColor to be used. Would require that users of Sparkline set fillColor and fillOpacity > 0 for the fillColor to be used

### DIFF
--- a/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
+++ b/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
@@ -157,12 +157,12 @@ export const GraphNG: React.FC<GraphNGProps> = ({
       builder.addSeries({
         scaleKey,
         drawStyle: customConfig.drawStyle!,
-        lineColor: seriesColor,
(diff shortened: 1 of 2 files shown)


ryantxu:
Looks like something changed in GraphNG -- in previous versions:
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/GraphNG/GraphNG.tsx#L165

was:
```
lineColor: lineColor ?? seriesColor,
```

torkelo:
@ryantxu 

> Looks like something changed in GraphNG -- in previous versions:

you linked to pointColor, not sure what you mean.

PR #29864, opened 2020-12-16
QueryEditors: run queries after changing group options

ryantxu:
This likely should have been part of #29762

This adds an explicit `onRunQueries` after changes in the options group

### DIFF
--- a/public/app/features/query/components/QueryGroupOptions.tsx
+++ b/public/app/features/query/components/QueryGroupOptions.tsx
@@ -144,19 +144,23 @@ export class QueryGroupOptionsEditor extends PureComponent<Props, State> {
       maxDataPoints = null;
     }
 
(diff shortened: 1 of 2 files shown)


public/app/features/query/components/QueryGroupOptions.tsx
```diff
@@ -144,19 +144,23 @@ export class QueryGroupOptionsEditor extends PureComponent<Props, State> {
       maxDataPoints = null;
     }
 
-    onChange({
-      ...options,
-      maxDataPoints: maxDataPoints,
-    });
+    if (maxDataPoints !== options.maxDataPoints) {
```
ryantxu, line 147: avoid the extra `onRunQueries`

torkelo (approved):

PR #29876, opened 2020-12-16
Chore: add as prop to InlineLabel to control rendered element

Elfo404:
Adds a `as` prop to control the rendered element for `InlineLabel`

this way it can be used like: 

```tsx
<InlineLabel  as="span">
// ...
</InlineLabel>
```

that would result in 

```html
<span class="css-1fh8o8p-label">Simple label</span>
```

technically, as can receive any other element including other react components.

the simplest solution here would maybe be to have a boolean prop to switch between `label` and `div`, but I felt this could have been an interesting approach. What do you think @torkelo ?


torkelo (approved):

Clarity-89 (approved):
Looks good altho not 100% sure about this, the name `InlineLabel` kinda implies that it's a `label` element but on the other hand it does use the `label` by default, so I guess it's fine.

PR #29879, opened 2020-12-16
FieldConfig API: Add ability to hide field option or disable it from the overrides.

dprokop:
This PR adds possibility to either:
- hide option from field config UI
- disable possibility to create an option override
- or both of the above

It introduces two new properties to the Field Options API (FieldConfigEditorConfig):
```ts
/** Hides option from the Field config tab */
hideFromDefaults?: boolean;

/** Indicates that option should not be available for the overrides */
hideFromOverrides?: boolean;
```

TODO:
- [ ] Add tests for OverrideEditor.tsx and DefaultFieldConfigEditor.tsx


packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -48,6 +48,12 @@ export interface FieldConfigEditorConfig<TOptions, TSettings = any, TValue = any
    * @param field
    */
   shouldApply?: (field: Field) => boolean;
+
+  /** Indicates that option shoukd not be available in the Field config tab */
+  hideFromFieldOptions?: boolean;
```
dprokop, line 53: Not a fan of this name, happy for any suggestion :)

packages/grafana-data/src/types/fieldOverrides.ts
```diff
@@ -50,10 +50,10 @@ export interface FieldConfigEditorConfig<TOptions, TSettings = any, TValue = any
   shouldApply?: (field: Field) => boolean;
 
   /** Indicates that option shoukd not be available in the Field config tab */
-  hideFromFieldOptions?: boolean;
+  hideFromDefaults?: boolean;
```
jackw, line 53: nit: I'm probably missing some context or a naming convention here but I find the previous name easier to follow.

jackw (approved):
LGTM! 👍

packages/grafana-data/src/types/fieldOverrides.ts
stevenvachon: `customizable?: boolean` ?

torkelo (approved):
Looks good, like the new naming (hideFromDefaults & hideFromOverrides)

packages/grafana-data/src/types/fieldOverrides.ts
dprokop: @jackw to shed some light on this. When we talk about "default field options" we refer to the options that are available from the Fields tab in the panel editor's side pane. This is pretty much a naming convention that we use, hence this suggestion from Torkel.

PR #29882, opened 2020-12-16
Usage Stats: Introduce an interface for usage stats service

vtorosyan:
Adding an interface type for usage stats service allows us to not depend on the implementation outside of the package, for example when testing we can easily mock the service

For more context, please have a look at the conversation in [this PR](https://github.com/grafana/grafana-enterprise/pull/938#discussion_r544296228). 

The usage stats service is not used elsewhere, outside of the package within OSS. It is going to be heavily used in the enterprise, once we start instrumenting enterprise features with more metrics. Thus, the change here is sufficient and we don't need extra refactoring to leverage the interface instead of passing the implementation type.

### DIFF
--- a/pkg/infra/usagestats/service.go
+++ b/pkg/infra/usagestats/service.go
@@ -21,6 +21,12 @@ func init() {
 	registry.RegisterService(&UsageStatsService{})
 }
 
+type UsageStats interface {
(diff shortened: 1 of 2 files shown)


sakjur (approved):

marefr (approved):
LGTM

PR #29884, opened 2020-12-16
API: Add by UID routes for data sources 

### LINKED ISSUE #29852: Backend data source cache: Make UID Version
For provisioning and multi-instance, we prefer UIDs over integer database ids.

However, the current data source cache is based on id: https://github.com/grafana/grafana/blob/master/pkg/services/datasources/cache.go#L34

**Why is this needed**:
For the backend data source code to use UIDs instead of IDs (e.g. data source UIDs in alert definitions), we need a UID based cache I believe. 

**Other:**
 - Will get from that database by UID? If so sqlstore/datasource will need something like `GetDatasourceByUID` (Seems like we have UQE_data_source_org_id_uid index in the table already)
 - Maybe Enterprise's datasourceCacheService would also need updating
 - Feels like we should not have two data source caches ... are we looking to replace get by ID with UID? If so, I think the we would want the frontend to send that instead of datasourceId (or alongside for a period). If we are not looking to replace, maybe we should just have a cache of uid to id, and forward that to the exists ds cache?

kylebrandt:
Adds the ability to get a data source by UID from SQL and via the datasource cache. This is so alerting ng will be able to store data sources by uid instead of id so it works with provisioning.

We may also switch the frontend to operate by UID, so this sets us up for that.

For requests by ID, an additional cache key to the datasource based on the UID is added. For requests by UID, the opposite is also true - an additional id cache key is added.

Also adds HTTP methods for getting + deleting via UID, and adds UID as a property to the response when listing datasources.

Fixes #29852

 - WIP - still need to look at enterprise. 
 - Also, not sure we want/need these API routes. If we do, I may need to also look at what update endpoint does with uids.


torkelo (dismissed):
Overall it looks good but I do feel that the amount of duplication is a bit too much. Would try to share the query operation (see GetDashboardQuery for reference it allows fetching dashboard by id, or uid or slug). And on the API handler it feels like we can also share more

docs/sources/http_api/data_source.md
```diff
@@ -396,6 +445,28 @@ Content-Type: application/json
 {"message":"Data source deleted"}
 ```
 
+## Delete an existing data source by UID
+
+`DELETE /api/datasources/uid/:uid`
+
+**Example Request**:
```
oddlittlebird, line 452:
```suggestion
**Example request:**
```

docs/sources/http_api/data_source.md
```diff
@@ -74,6 +75,53 @@ Content-Type: application/json
...
+  "version": 1,
+  "readOnly": false
+}
+```
+
+## Get a single data source by UID
+
+`GET /api/datasources/uid/:uid`
+
+**Example Request**:
```
oddlittlebird, line 107:
```suggestion
**Example request:**
```

docs/sources/http_api/data_source.md
```diff
@@ -396,6 +445,28 @@ Content-Type: application/json
...
+
+```http
+DELETE /api/datasources/uid/kLtEtcRGk HTTP/1.1
+Accept: application/json
+Content-Type: application/json
+Authorization: Bearer eyJrIjoiT0tTcG1pUlY2RnVKZTFVaDFsNFZXdE9ZWmNrMkZYbk
+```
+
+**Example Response**:
```
oddlittlebird, line 461:
```suggestion
**Example request:**
```

pkg/services/datasources/cache.go
```diff
@@ -48,11 +49,52 @@ func (dc *CacheServiceImpl) GetDatasource(
...
+	dc.CacheService.Set(idKey(ds.Id), ds, time.Second*5)
+	return ds, nil
+}
+
+func idKey(id int64) string {
+	return fmt.Sprintf("ds-%d", id)
+}
+
+func uidKey(uid string) string {
+	return fmt.Sprintf("dsuid-%s", uid)
```
marefr, line 99: This key needs to include the `org id` since uid is unique per org.

pkg/api/datasources.go
```diff
@@ -103,14 +104,61 @@ func DeleteDataSourceById(c *models.ReqContext) Response {
 	return Success("Data source deleted")
 }
 
+// GET /api/datasources/uid/:uid
+func GetDataSourceByUID(c *models.ReqContext) Response {
+	query := models.GetDataSourceQuery{
```
marefr, line 109: Can you use `getRawDataSourceByUID` here instead?

pkg/api/datasources.go
```diff
@@ -103,14 +104,61 @@ func DeleteDataSourceById(c *models.ReqContext) Response {
...
+	ds, err := getRawDataSourceByUID(uid, c.OrgId)
+	if err != nil {
+		return Error(400, "Failed to delete datasource", nil)
+	}
+
+	if ds.ReadOnly {
+		return Error(403, "Cannot delete read-only data source", nil)
+	}
+
+	cmd := &models.DeleteDataSourceByUIDCommand{Uid: uid, OrgId: c.OrgId}
```
marefr, line 144: Nit. Could potentially use one `DeleteDataSourceCommand` similar to `GetDataSourceQuery` for the different kind of deletes to limit some code duplication.

marefr (changes requested):
Great. One important change in cache service related to uid cache key. In addition some nits similar to Torkel's review.

pkg/api/datasources.go
torkelo: yea, I think that makes sense as well. less confusion on what command to use

torkelo (commented):
Think this looks good, marcus found some good things

docs/sources/http_api/data_source.md
kylebrandt, line 461:
```suggestion
**Example response**:
```

pkg/api/datasources.go
```diff
@@ -93,7 +94,49 @@ func DeleteDataSourceById(c *models.ReqContext) Response {
...
+	ds, err := getRawDataSourceByUID(uid, c.OrgId)
+	if err != nil {
+		return Error(400, "Failed to delete datasource", nil)
+	}
+
+	if ds.ReadOnly {
+		return Error(403, "Cannot delete read-only data source", nil)
+	}
+
+	cmd := &models.DeleteDataSourceCommand{UID: uid, OrgID: c.OrgId}
```
marefr, line 139: Nit. This part and below could be reused by DeleteDataSourceByUID and DeleteDataSourceById if extracted to `func deleteDataSource(cmd *models.DeleteDataSourceCommand) Response`. But could be fixed later if you want to get this merged fast.

Nit. If the data source is not found in `getRawDataSourceByUID` that should probably return a 404 not found instead. Same applies to `DeleteDataSourceById`.

marefr (approved):
Great. LGTM. Added a couple nits, but no blocker

pkg/api/datasources.go
kylebrandt: Fixed for 404, will save other for possible future work. Seems we have mix of responses. May want to start returning the UID (which requires the getCmd results), so will save that for possible future work focusing on api consistency.

PR #29886, opened 2020-12-16
Alerting: Copy frame field labels to time series tags

wbrowne:
I noticed we don't always copy frame labels where necessary - I believe this fixes it


kylebrandt (approved):

papagian (approved):
LGTM

PR #29892, opened 2020-12-17
Search: use Card component

### LINKED ISSUE #29870: Search: use Card component for search items
Update search item list to use [new Card component](https://www.figma.com/file/ITCsaXtsANM17oRtU6GQS9/Component-Playground?node-id=712%3A0).

Clarity-89:
Use new Card component for the search list items.

Fixes https://github.com/grafana/grafana/issues/29870

Noticed a few issues/concerns with the new component:
- The search items are way bigger now (almost twice in height) so there's less items displayed on the screen at once. This is even more apparent in the manage dashboards section:
Screenshot 2020-12-17 at 9 18 44

- The virtualisation library used for the search list expects the items to have a fixed size. Setting (an arbitrary) min height on the item seems to work for UI purposes, however the dashboards  without tags and/or folder have a lot of empty space now:

Screenshot 2020-12-17 at 9 21 01
Screenshot 2020-12-17 at 9 27 08
- The Card component doesn't have support for checkbox item on the left. Using `Figure` component works, however it's not vertically centered, so a new component for the left-side centered content will need to be added.


torkelo:
Think we need some help from @jessover9000 here on how to make the dashboards without any tags feels less empty. Think for dashboards in General (root folder) we can still show that (General). Would be nice to get the description in here as well. 

We have some updated designs here as well we add back the dashboard icons, as we will also show other entities in this list (library panels)

Clarity-89:
Maybe for the search items it would be better to stick to the layout where the tags are on the right. It also leaves enough space for the metadata below the dashboard name:
Screenshot 2021-01-05 at 12 08 02

torkelo:
yea, I think we still have to solve the response layout (wide) design for the cards, we hit the same problem for the alert rule list if I remember

Clarity-89:
After some tweaking I think the layout with the tags on the right looks good. The tag list width is limited to 50% of container, otherwise it will overflow on the next row. 
Screenshot 2021-01-07 at 11 14 51

jessover9000:
nice! well done :) some nitpicks:

- I think the tags should be right-aligned, that works but when they overflow to multiple lines they're still left-aligned🤓
- The paddings for the cards are 16px and look too big in my opinion, have you tried 8px? I would rather have smaller paddings and slightly bigger margins between the cards.
- The padding between dashboard name and folder/meta-info is also quite big, I think you could halve that too.

Clarity-89:
@jessover9000 thanks for the review! I have applied your suggestions and now the items look like this: 
Screenshot 2021-01-11 at 17 40 03
Screenshot 2021-01-11 at 18 51 56

> The paddings for the cards are 16px and look too big in my opinion, have you tried 8px? I would rather have smaller paddings and slightly bigger margins between the cards.

8px padding looks too small imo. I see the designs have 11px padding, wonder if we'd try 12 px. But also there was extra vertical padding added due to fixed item's height, I decreased it and it look better now.

torkelo:
The search cards feel a bit too high, but overall looks ok & mergable. 

Really miss a storybook story for the Card component that showcases all the different states & use cases for the card

torkelo (approved):

Clarity-89:
> The search cards feel a bit too high, but overall looks ok & mergable.

The requirement to have all cards fixed height for the list to be virtualised makes it quite tricky to use the card component here. I think we can iterate a bit more on this component during the beta. 
 
> Really miss a storybook story for the Card component that showcases all the different states & use cases for the card

There's an internal story with majority of possible configurations in the docs tab, I'll make it public and change the alpha tag of the component to beta.

PR #29893, opened 2020-12-17
GraphNG: Color series from by value scheme & change to fillGradient to gradientMode 

torkelo:
Implementing the color series by value as outlined in the design doc:
https://docs.google.com/document/d/15G523kih8qj_d-kp-7AetLFRP7Q1D-aHvj4eDAXGVo0/edit# 

* [x] Add seriesBy to FieldColor
* [x] Add bySeriesSupport option to FieldColor settings 
* [x] Updated FieldColor UI to show new option if bySeriesSupport is enabled
* [x] New function getFieldSeriesColor that uses this new option to figure out a series color
* [x] Rename fillGradient to gradientMode, move from AreaConfig to main config type as it will impact line drawing as well
* [x] Introduce new GraphGradientMode value option `Color scheme`  (Color scale feels better but our color option is named color scheme and want the connection to be clear)

I have updated this to be mergable, so the new graph does not take advantage of byValue color schemes (disabled). Want to merge this as it changed fillGradient to gradientMode. 
And updates & improved the code for fill gradients, and improves the hue mode for light theme (making the hue mode theme aware). 

Adds a new test dashboard for hue mode:
Screenshot from 2021-01-17 08-36-12

TODO / Problems (For future PRS) 
* [ ] for series color changes to take effect we need to rebuild the config in GraphNG / re-init uPlot or change quite a bit how the color is passed around. The color is assigned in GraphNG inside the config memoization part. series color is part of legend & uPlot config series builder How to make this part data dynamic will require some tricky changes. Not sure what approach to take yet for this. The easiest thing might be make lineColor & fillColor functions in SeriesProps (for the builder) and for the Legend. Will have to make the legend re-render on every data change. 
* [ ] Introduce support for the Color scheme gradient support

Screenshot from 2020-12-17 08-47-03


dprokop:
> The color is assigned in GraphNG inside the config memoization part. series color is part of legend & uPlot config series builder How to make this part data dynamic will require some tricky changes. Not sure what approach to take yet for this. The easiest thing might be make lineColor & fillColor functions in SeriesProps (for the builder) and for the Legend. Will have to make the legend re-render on every data change.

@torkelo - me and @mckn we have discussed a possible direction to refactor the config memoization a bit. The gist of the solution is to build the config in the frames alignment phase. Not 100% sure yet it will solve the problem we have encountered in the legend stats work and here, but it's workth POCing. I hope to do this after the beta.

packages/grafana-data/src/field/fieldColor.test.ts
```diff
@@ -27,4 +44,47 @@ describe('fieldColorModeRegistry', () => {
     const calcFn = getCalculator({ mode: FieldColorModeId.PaletteClassic, seriesIndex: 1 });
     expect(calcFn(70, 0, undefined)).toEqual('#EAB839');
   });
+
+  it('When color.seriesBy is set to last use that instead of v', () => {
+    const field = getTestField('continuous-GrYlRd');
+
+    field.config.color!.seriesBy = 'last';
+    // min = -10, max -10, last: 5
```
dprokop, line 52:
```suggestion
    // min = -10, max = 10, last = 5
```

torkelo:
> @torkelo - me and @mckn we have discussed a possible direction to refactor the config memoization a bit. The gist of the solution is to build the config in the frames alignment phase

@dprokop would not moving the config to data alignment force a uplot config re-init on ever data change?

dprokop:
> @dprokop would not moving the config to data alignment force a uplot config re-init on ever data change?

Our plan is to memoize, or rather use the previous value of the returned config based on object value equality - config comparison should be fairly cheap to do as these are small simple objects. Hence we would pass the correct UPlotConfig reference to the Plot component if the config itself has not changed.

torkelo:
@dprokop so instead of detecting if data frame config has changed always build the complete uplot config and then do a object diff? The object diff is going to be tricky as some parts of the ConfigBuilder props are functions

packages/grafana-data/src/field/fieldColor.test.ts
```diff
@@ -27,4 +44,47 @@ describe('fieldColorModeRegistry', () => {
...
+    // last percent 75%
+    field.values = new ArrayVector([0, -10, 5, 10, 2, 5]);
+
+    const color = getFieldSeriesColor(field, getTestTheme());
+    const calcFn = getCalculator({ mode: 'continuous-GrYlRd' });
+
+    expect(color.color).toEqual(calcFn(4, 0.75));
```
dprokop, line 59: Took me some time to get this in my head right and I'm not 100 sure I succeeded:) Can you leave a comment in the code where does 4 come from?

packages/grafana-ui/src/components/uPlot/config/UPlotConfigBuilder.test.ts
```diff
@@ -3,7 +3,7 @@
 import { UPlotConfigBuilder } from './UPlotConfigBuilder';
 import { GrafanaTheme } from '@grafana/data';
 import { expect } from '../../../../../../public/test/lib/common';
```
dprokop, line 5:
```suggestion
```

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -90,28 +96,39 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
...
   }
 
-  getFill(): Series.Fill | undefined {
-    const { lineColor, fillColor, fillGradient, fillOpacity } = this.props;
+  private getLineColor(): Series.Stroke {
+    const { lineColor, gradientMode, colorMode, thresholds } = this.props;
 
-    if (fillColor) {
-      return fillColor;
+    if (gradientMode === GraphGradientMode.Scheme) {
```
dprokop, line 102: Can't wait to see this in action

dprokop (approved):
Left one comment about test that I feel like not fully understanding, butother than that this looks great and can't wait to see this in aciton

packages/grafana-data/src/field/fieldColor.test.ts
torkelo: can't really explain this one, the last value is 5, not a 4, confused : )

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
torkelo: Screenshot from 2021-01-18 16-38-15

torkelo: image

packages/grafana-data/src/field/fieldColor.test.ts
torkelo: @dprokop ok now I understand it, the 4 can be anything as this color scale only uses the percent value 0.75 comes from the fact that 5 is 75% in the range -10 to 10

torkelo: added comment to this line

dprokop: heh, is see the confusion hit not only me:) thanks a lot!

PR #29896, opened 2020-12-17
Alerting NG: update API to expect UIDs instead of IDs

### LINKED ISSUE #29339: AlertingNG: Introduce UUID for alert definitions and update API appropriately
**Why is this needed**:
Unique identifiers are preferable over using database IDs.

papagian:
In #29305 alert definitions UIDs are introduced in the database. This modifies the respective API calls to use UIDs instead of IDs.

More specifically:

- Endpoints that used to get `alertDefinitionID` query parameter now get `alertDefinitionUID`
- create and update alert definition responses have changed to return whole alert definition

Fixes #29339


pkg/services/ngalert/api.go
```diff
@@ -54,9 +54,9 @@ func (ng *AlertNG) conditionEvalEndpoint(c *models.ReqContext, dto evalAlertCond
 
 // alertDefinitionEvalEndpoint handles GET /api/alert-definitions/eval/:dashboardId/:panelId/:refId".
```
kylebrandt, line 55: comment out of date

pkg/services/ngalert/api.go
```diff
@@ -54,9 +54,9 @@ func (ng *AlertNG) conditionEvalEndpoint(c *models.ReqContext, dto evalAlertCond
...
 func (ng *AlertNG) alertDefinitionEvalEndpoint(c *models.ReqContext) api.Response {
-	alertDefinitionID := c.ParamsInt64(":alertDefinitionId")
+	alertDefinitionUID := c.ParamsEscape(":alertDefinitionUID")
 
-	condition, err := ng.LoadAlertCondition(alertDefinitionID)
+	condition, err := ng.LoadAlertCondition(alertDefinitionUID, c.SignedInUser.OrgId)
 	if err != nil {
 		return api.Error(400, "Failed to load alert definition conditions", err)
 	}
```
kylebrandt, line 62: on line 70, s/evaludate/evaluate

pkg/services/ngalert/schedule.go
```diff
@@ -221,42 +232,46 @@ func (ng *AlertNG) alertingTicker(grafanaCtx context.Context) error {
 
 type alertDefinitionRegistry struct {
 	mu                  sync.Mutex
-	alertDefinitionInfo map[int64]alertDefinitionInfo
+	alertDefinitionInfo map[string]alertDefinitionInfo
```
kylebrandt, line 235: I think the key can be a struct type with a int64 (orgid) and a string (uid), then the string joining and conversion can go away I perhaps?

kylebrandt: (or could be a tuple array if string [2]string (but I think I like the struct more). Slices can not be map keys but arrays can.

kylebrandt (approved):
LGTM. A couple cleanup comments, and alternatives for the alert def map key for you to consider.

pkg/services/ngalert/schedule.go
```diff
@@ -85,8 +76,8 @@ func (ng *AlertNG) definitionRoutine(grafanaCtx context.Context, definitionUID s
 				}
 			}()
 		case k := <-ng.schedule.stop:
-			if key == k {
-				ng.schedule.log.Debug("stopping alert definition routine", "definitionUID", definitionUID, "orgID", orgID)
+			if k.orgID == key.orgID && k.definitionUID == key.definitionUID {
```
kylebrandt, line 79: can be `if k == key {`

PR #29898, opened 2020-12-17
PanelEdit: Update UI if panel plugin changes field config

torkelo:
Fixes so that the panel edit ui updates when field config is changed from panel plugin

Opted to reuse the event we already have as they both serve the exact same purpose (trigger re-render in PanelEditor)


mckn (approved):
LGTM!

PR #29902, opened 2020-12-17
Folders: Removes the possibility to delete the General folder

### LINKED ISSUE #29880: Deleting "General" folder causing empty grafana state
**What happened**:
Moved all dashboards into custom folder and deleted "General" folder. 
After refresh I see empty list: 
image

In case General is default and mandatory folder:
- I expect to see disabled delete button 
- or notification about my deletion aborted
- or popup to define new default folder.   
In case General is ordinary folder: 
- Correct deletion and remained folders with all created dashboards. 

general_deleted

**Environment**:
- Grafana version: v7.3.5 (11f305f88a) and  v7.0.3 (00ee734baf)
- Data source type & version:  
- OS Grafana is installed on: Mint 20.10 (with docker container) and CentOS (with docker container)
- User OS & Browser: Firefox 83.0 (64-bit)
- Grafana plugins:
- Others:

hugohaggmark:
This PR prevents the Delete button to appear when checking only the `General` folder for deletion, because that leads to a confusing state as described in #29880.

We could also make it so whenever you select folders for deletion, including the `General` folder, the Delete button will not be available. Not sure this is a good user experience, though. Please let me know if I should include this as well.

Enjoy the review 🎉 

Fixes #29880

### DIFF
--- a/public/app/features/search/hooks/useManageDashboards.ts
+++ b/public/app/features/search/hooks/useManageDashboards.ts
@@ -5,8 +5,7 @@ import { DashboardQuery, DashboardSection, OnDeleteItems, OnMoveItems, OnToggleC
 import { DELETE_ITEMS, MOVE_ITEMS, TOGGLE_ALL_CHECKED, TOGGLE_CHECKED } from '../reducers/actionTypes';
 import { manageDashboardsReducer, manageDashboardsState, ManageDashboardsState } from '../reducers/manageDashboards';
 import { useSearch } from './useSearch';
-
-export const GENERAL_FOLDER_ID = 0;
+import { GENERAL_FOLDER_ID } from '../constants';
 
 export const useManageDashboards = (
   query: DashboardQuery,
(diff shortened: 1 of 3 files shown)


public/app/features/search/hooks/useManageDashboards.ts
```diff
@@ -6,6 +6,8 @@ import { DELETE_ITEMS, MOVE_ITEMS, TOGGLE_ALL_CHECKED, TOGGLE_CHECKED } from '..
 import { manageDashboardsReducer, manageDashboardsState, ManageDashboardsState } from '../reducers/manageDashboards';
 import { useSearch } from './useSearch';
 
+export const GENERAL_FOLDER_ID = 0;
```
Clarity-89, line 9: There's also `constants` file in the search folder where this could be stored.

Clarity-89 (approved):
Looks good to me! Also great job with adding tests for the hook 🎉

public/app/features/search/hooks/useManageDashboards.ts
hugohaggmark: Brilliant suggestion, will move it there right away!

PR #29903, opened 2020-12-17
Elasticsearch: Fix broken alerting when using pipeline aggregations

Elfo404:
In the previous data source version, when selecting a field for a pipeline aggregation type, both `field` and `pipelineAgg` were populated in the query model. The react refactoring introduced a bug for which pipeline aggregations don't have `pipelineAgg` set anymore, this is not used in the FE query builder but the BE rely on that field to build the query for alerts.

This PR ensures that when selecting a field for a pipeline metric aggregation, both `field` and `pipelineAgg` in the query model are set.

Also, `onRunQuery` should be called after query change (ie. after each interaction with the query editor), somehow it slipped through and it wasn't being called. https://github.com/grafana/grafana/pull/29903/files#diff-47bf7d4f258d138e6a3eb152d14cfbf5c3d5581b6866a77154e27d627f428354R1 this fixes this behavior.

to test this out I created those 2 dashoards: https://gist.github.com/Elfo404/bd73e0749597e86b8ffb3ed5f4b14208, the interesting bits are in the "targets" array, where the query model is saved. the alert threshold is only based on random data, but when testing the rule you can see in the non-working version `"Condition[0]: Eval: false, Metric: Derivative Average 1, Value: null"` 

```json
// ...
    {
      "field": "1",
      "id": "3",
      "type": "derivative"
    }
```

while the new one:
```json
// ...
    {
      "field": "1",
      "id": "3",
      "pipelineAgg": "1",
      "type": "derivative"
    }
```

### DIFF
--- a/public/app/plugins/datasource/elasticsearch/components/QueryEditor/ElasticsearchQueryContext.test.tsx
+++ b/public/app/plugins/datasource/elasticsearch/components/QueryEditor/ElasticsearchQueryContext.test.tsx
@@ -12,11 +12,19 @@ const query: ElasticsearchQuery = {
 };
 
 describe('ElasticsearchQueryContext', () => {
-  it('Should call onChange with the default query when the query is empty', () => {
+  it('Should call onChange and onRunQuery with the default query when the query is empty', () => {
     const datasource = { timeField: 'TIMEFIELD' } as ElasticDatasource;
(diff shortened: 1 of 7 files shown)


zoltanbedi (approved):
LGTM 👌

PR #29904, opened 2020-12-17
PanelEvents: Isolating angular panel events into it's own event bus + more event refactoring 

### LINKED ISSUE #29815: PanelEdit: After changing from angular to react panel the panel options are no longer updated on change
This is caused by the clean up of the event bus we need to do when removing the angular directive, we need to remove event subscriptions done by the directive. 

https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/state/PanelModel.ts#L485

torkelo:
Fixes #29815 

We have some problems with panel events, for angular we have to destroy and unsub to all the events subscribed by the angular controller and plugin code, but the PanelEditor will then lose its subscriptions. Feels like it's best to separate these into two different emitters. Then it becomes easy to clean up the angular state without any messy logic. 

This also updates the core Refresh & Render events to the new model and removes events that we no longer need.

### DIFF
--- a/public/app/features/dashboard/components/Inspector/QueryInspector.tsx
+++ b/public/app/features/dashboard/components/Inspector/QueryInspector.tsx
@@ -10,8 +10,9 @@ import { getPanelInspectorStyles } from './styles';
 import { supportsDataQuery } from '../PanelEditor/utils';
 import { config } from '@grafana/runtime';
 import { css } from 'emotion';
-import { Unsubscribable } from 'rxjs';
+import { Subscription } from 'rxjs';
 import { backendSrv } from 'app/core/services/backend_srv';
+import { RefreshEvent } from 'app/types/events';
 
 interface DsQuery {
   isLoading: boolean;
@@ -39,9 +40,8 @@ interface State {
 }
 
 export class QueryInspector extends PureComponent<Props, State> {
-  formattedJson: any;
-  clipboard: any;
-  subscription?: Unsubscribable;
+  private formattedJson: any;
+  private subs = new Subscription();
 
   constructor(props: Props) {
     super(props);
@@ -58,11 +58,15 @@ export class QueryInspector extends PureComponent<Props, State> {
   }
 
   componentDidMount() {
-    this.subscription = backendSrv.getInspectorStream().subscribe({
-      next: response => this.onDataSourceResponse(response),
-    });
+    const { panel } = this.props;
 
-    this.props.panel.events.on(PanelEvents.refresh, this.onPanelRefresh);
+    this.subs.add(
+      backendSrv.getInspectorStream().subscribe({
+        next: response => this.onDataSourceResponse(response),
+      })
+    );
+
+    this.subs.add(panel.events.subscribe(RefreshEvent, this.onPanelRefresh));
     this.updateQueryList();
   }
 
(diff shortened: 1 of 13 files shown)


public/app/features/dashboard/components/Inspector/QueryInspector.tsx
```diff
@@ -39,9 +40,8 @@ interface State {
 }
 
 export class QueryInspector extends PureComponent<Props, State> {
-  formattedJson: any;
-  clipboard: any;
-  subscription?: Unsubscribable;
+  private formattedJson: any;
+  private subs = new Subscription();
```
hugohaggmark, line 44: nit: shouldn't this initialization move into the constructor?

public/app/features/dashboard/dashgrid/DashboardGrid.tsx
```diff
@@ -164,12 +161,6 @@ export class DashboardGrid extends PureComponent<Props> {
     this.forceUpdate();
   };
 
-  onWidthChange = () => {
-    for (const panel of this.props.dashboard.panels) {
-      panel.resizeDone();
-    }
-  };
-
```
hugohaggmark, line 172: Out of curiosity, where is this done now?

public/app/features/dashboard/dashgrid/PanelChrome.tsx
```diff
@@ -52,9 +52,8 @@ export interface State {
 }
 
 export class PanelChrome extends Component<Props, State> {
-  readonly timeSrv: TimeSrv = getTimeSrv();
-
-  querySubscription?: Unsubscribable;
+  private readonly timeSrv: TimeSrv = getTimeSrv();
+  private subs = new Subscription();
```
hugohaggmark, line 56: Same nit as above :)

public/app/features/dashboard/state/DashboardModel.ts
```diff
@@ -883,10 +881,12 @@ export class DashboardModel {
     return rowPanels;
   }
 
+  /** @deprecated */
```
hugohaggmark, line 884: Maybe we should also add a deprecation warning to these functions too?

public/app/features/panel/panel_directive.ts
```diff
@@ -22,6 +24,7 @@ module.directive('grafanaPanel', ($rootScope, $document, $timeout) => {
       const panel: PanelModel = scope.ctrl.panel;
 
       let panelScrollbar: any;
+      let subs = new Subscription();
```
hugohaggmark, line 27: Why let?

hugohaggmark (approved):
Looks fantastic, so much cleaner now, great work! I did some testing, and I couldn't find any issues. I left some super-small nits for you to consider.

public/app/features/dashboard/dashgrid/DashboardGrid.tsx
hugohaggmark: Nevermind I saw later in the review where this is done now :)

public/app/features/dashboard/components/Inspector/QueryInspector.tsx
torkelo: yes, conflicted on what is best. This is technically the same as the constructor. The downside of moving it to the constructor is that I need to type it here and then duplicate it in the constructor. So this feels a bit leaner this way, sort of like declaring an instance value variable.

public/app/features/panel/panel_directive.ts
torkelo: mistake, good catch!

public/app/features/dashboard/state/DashboardModel.ts
torkelo: good idea, added!

PR #29906, opened 2020-12-17
Ng Alerting: Remove scroll and fix SplitPane limiters

peterholmberg:
Noticed that the alerting page was longer than the viewport, had to duplicate some css from panel edit. While investigating this I discovered that the `SplitPane` component had built in limiters on size which meant that we could remove our own solution for that.


torkelo (approved):
LGTM

Issue #7690, opened 2017-02-28
[Feature request] Include query results data into alert messages


zax-slaughter:
- Grafana version 4.1.1
- InfluxDB
- Ubuntu 14

- What I did is I tried to set up sensible alerts on many servers. I wanted to have a single query report failures on many servers. Query would return a value based on which notification is sent (say count of errors) + tag information that would allow me to identify the error type and specific server failing.

- What I expected it that it'll be absolutely awesome as all the rest of grafana. However alerts don't work with templates (I know there is another issue on that), so I had to put all errors into a single query. Now the problem is that with many servers there is no way to have meaningful notification sent. Meaningful notification would include information on the errors, such as tags, or error message, ect. That would allow to direct it to a right support group.

- What happens instead, is that I have to set up individual alert panel for each new server, otherwise I can't send meaningful notifications. That of course allows to switch notifications on and off easily with a high fidelity, but requires a lot of panel duplication that is tedious and could have been avoided.

I think this is a relatively small and useful request to allow notification message to include
- alert condition (say we are alerting on count(*) - include threshold end count(*))
- same as alias by allows, such as $tag_system or $tag_error_type, etc

daniellee:
Possible dupe: #5848

zax-slaughter:
I would say that #7121 <https://github.com/grafana/grafana/issues/7121> is
closer, but regardless, I am asking for much smaller feature to make the
feature usable by real world support.

nikskiz:
This feature will be awesome

beingyash:
This is definitely would be very helpful.

Zappes:
This would actually make thed alerts helpful, which they are currently not.

ysoftman:
Isn't this feature completed yet?
(I'm using Grafana version 7.4.5, How can I include query results into the alert notification message?)

rlevytskyi:
Hi All,
Is this feature implemented?
Or maybe there is other way to get some data from query results and pass it to the alert message?

ceeeekay:
Adding to this issue, because I've just spent ages trying to set up a PoC to extract some useful string data for use in alert labels, only to find out the developers don't think doing such a thing is worthwhile.

It's astounding that something so obviously useful has been overlooked (or just straight up denied) by the developers for so long. Alerts _need_ to be enriched with relevant supporting data so that the people receiving them have at least some idea what's going on, instead of dealing with some inscrutable message at 4am.

[...]

Issue #7121, opened 2017-01-03
[Feature request] Email templating in Alert engine


maciejasembler:
**- What Grafana version are you using?**
4.0.1
**- What datasource are you using?**
OpenTSDB
**- What OS are you running grafana on?**
CentOS 7.2

I would like to congratulate for including alerting feature in Grafana 4, this is just what I needed to create alerts for things that our monitoring did not cover. To make email alerting more flexible I would propose to create simple mail templating engine:

**Current Use Case**
We setup alert thresholds and alert recipient as well as static text which informs about the cause, incomplete Graph (no text, no legend) is also added

**What is missing?**
However I think that the email engine lacks templating option (or documentation on this feature if it is possible). What I would like to request is:
- formatting the metric value with units or just manipulate it via variable
- ability to use some HTML tags / BB code / custom grafana rules for text formatting (at least support for generating links)
- use of dasboard-specific variables such as query alias, graph title, dashboard title or template variables defined in dashboard (could be used as a temporary workaround since alerts do not work with template vars in queries), date of alert triggering - this is mainly for customizing links which Grafana could generate when alert occurs
- Create PNG from the graph state when alert triggered - currently no text is drawn onto the graph making it quite unreadable - exact PNG image from configured view would have been sufficient if possible
- Subject templating - simple textbox with variable support

Pros:
- make alerts customizable, lazy users will not be forced to create new mail filters
- make messages clear to understand - at least by formatting it to the unit used in graph - for instance in rates measured in byte difference but displayed in Mbit/s

Cons:
- can't think of any, except somebody has to write some code and docs on it

Keep up the good work !

pdf:
See also potential extended usage: https://github.com/grafana/grafana/issues/6553#issuecomment-272602289

jaimegago:
+1

ddhirajkumar:
+1

rlueckl:
I would like to add:
 * an option to disable the upload of the graph to an external server (make the "fallback" - attaching/embedding the image the default!)
 * don't load resources from external providers (like the logo in the header)

Most of us are using Grafana in an internal network and uploading our metrics to an external site could be fatal. I don't understand how this is the default (and currently the only) option. The only thing saving us currently, is that the servers are not allowed to communicate with the outside world and Grafana is therefore forced to use the fallback method of embedding the picture in the e-mail.

Pros:
 * everybody could decide if his data is sent through the Internet or not

Cons:
 * can't think of any

[...]

PR #29908, opened 2020-12-17
Alerting: Add support for alert notification query label interpolation

wbrowne:
Supports `${}` style templating in alert notification title (alert rule name) and message (alert notification message) fields

Screenshot 2021-01-15 at 16 57 18

Which will get evaluated into the following sample:

```json
{
  "dashboardId": 56,
  "evalMatches": [
    {
      "value": 183.99711804026737,
      "metric": "{handler=\"prometheus\", instance=\"localhost:9090\", job=\"prometheus\", quantile=\"0.5\"}",
      "tags": {
        "handler": "prometheus",
        "instance": "localhost:9090",
        "job": "prometheus",
        "quantile": "0.5"
      }
    },
    {
      "value": 188.1379337949806,
      "metric": "{handler=\"prometheus\", instance=\"localhost:9090\", job=\"prometheus\", quantile=\"0.9\"}",
      "tags": {
        "handler": "prometheus",
        "instance": "localhost:9090",
        "job": "prometheus",
        "quantile": "0.9"
      }
    },
    {
      "value": 150.32021774806873,
      "metric": "{handler=\"query_range\", instance=\"localhost:9090\", job=\"prometheus\", quantile=\"0.5\"}",
      "tags": {
        "handler": "query_range",
        "instance": "localhost:9090",
        "job": "prometheus",
        "quantile": "0.5"
      }
    },
    {
      "value": 148.87523515990875,
      "metric": "{handler=\"query_range\", instance=\"localhost:9090\", job=\"prometheus\", quantile=\"0.9\"}",
      "tags": {
        "handler": "query_range",
        "instance": "localhost:9090",
        "job": "prometheus",
        "quantile": "0.9"
      }
    }
  ],
  "message": "Degraded 0.5, 0.9 latency on localhost:9090",
  "orgId": 1,
  "panelId": 2,
  "ruleId": 6,
  "ruleName": "HTTP latency (0.5, 0.9)",
  "ruleUrl": "http://localhost:3000/d/2Ny2OFJMk/prometheus-play?tab=alert&viewPanel=2&orgId=1",
  "state": "alerting",
  "tags": {},
  "title": "[Alerting] HTTP latency (0.5, 0.9)"
}
```

Note: As the templated message is dependant on the rule evaluation matches, message notifications will look like this for example for state changes


pkg/services/alerting/notifier_test.go
```diff
@@ -18,18 +19,19 @@ import (
...
-		Name:          "Test",
-		Message:       "Something is bad",
-		State:         models.AlertStateAlerting,
-		Notifications: []string{"1"},
-	}
+	testRule := newTestRule("Test", "Something is bad")
 	evalCtx := NewEvalContext(context.Background(), testRule)
 
+	testRuleTemplated := newTestRule("Test {{.quantile}}", "Something is bad on instance {{ .instance }}")
+	evalCtxWithMatch := newEvalContext(context.Background(), testRuleTemplated, &EvalMatch{
```
aknuds1, line 26: Can you just inline newEvalContext? Since it's only used once and super simple, I don't see any benefit to it. It just causes more complexity (especially with the condition) than inlining.

aknuds1 (commented):
Looks good AFAICT, but what is the motivation for the PR (a bug, or a feature request?)? Why is it still in draft mode?

wbrowne:
@aknuds1 need to do some further testing, but wanted to verify the general shape is good 😄

marefr:
Great start. Looks promising, all the functionality seems to be there :+1: Somewhat (partly) related to #7121 and/or #7690

pkg/services/alerting/eval_context.go
```diff
@@ -178,3 +181,51 @@ func getNewStateInternal(c *EvalContext) models.AlertStateType {
...
+	c.Rule.Name = titleBuf.String()
+	return nil
+}
+
+// buildTemplateDataMap builds a map of alert evaluation tag names to a set of associated values (comma separated)
+func buildTemplateDataMap(evalMatches []*EvalMatch) map[string]string {
+	var result = map[string]string{}
+	for _, match := range evalMatches {
+		for tagName, tagValue := range match.Tags {
+			if _, exists := result[tagName]; exists && !strings.Contains(result[tagName], tagValue) {
```
aknuds1, line 223: Isn't there a danger that `result[tagName]` contains a comma separated entry that is a super-string of `tagValue`?That is to say, I think you should instead use a precise regex here to make sure you don't get a partial match.

pkg/services/alerting/notifier_test.go
```diff
@@ -269,6 +299,19 @@ func newTestNotifier(model *models.AlertNotification) (Notifier, error) {
...
 }
 
+func newTestRule(name, description string) *Rule {
+	return &Rule{
+		ID:            1,
+		DashboardID:   1,
+		PanelID:       1,
+		OrgID:         1,
+		Name:          name,
+		Message:       description,
```
aknuds1, line 309: It's less confusing if the parameter is named `message`, same as the field it initializes:
```suggestion
func newTestRule(name, message string) *Rule {
	return &Rule{
		ID:            1,
		DashboardID:   1,
		PanelID:       1,
		OrgID:         1,
		Name:          name,
		Message:       message,
```

pkg/services/alerting/notifier_test.go
```diff
@@ -18,18 +19,21 @@ import (
...
-		OrgID:         1,
-		Name:          "Test",
-		Message:       "Something is bad",
-		State:         models.AlertStateAlerting,
-		Notifications: []string{"1"},
-	}
+	testRule := newTestRule("Test", "Something is bad")
 	evalCtx := NewEvalContext(context.Background(), testRule)
 
+	testRuleTemplated := newTestRule("Test latency {{.quantile}}", "Something is bad on instance {{ .instance }}")
```
aknuds1, line 25: Total nit, but it's less confusing to the reader if the template style is consistent (otherwise it can look as if whitespace is significant):
```suggestion
	testRuleTemplated := newTestRule("Test latency {{.quantile}}", "Something is bad on instance {{.instance}}")
```

aknuds1 (commented):
Generally looks good, but see question and suggestions please.

pkg/services/alerting/eval_context.go
wbrowne: Not 100% sure what you mean - maybe take a look at the tests I just added for this function and if you think there's a missing test case that highlights an issue, let me know!

aknuds1: OK thx, will check the tests.

aknuds1:
I pushed a commit with a failing test case, corresponding to what I had in mind. I also wrapped the test cases in `t.Run`, so you can run each in isolation, makes for a much better debugging experience.

pkg/services/alerting/eval_context.go
```diff
@@ -178,3 +181,55 @@ func getNewStateInternal(c *EvalContext) models.AlertStateType {
...
+	return nil
+}
+
+// buildTemplateDataMap builds a map of alert evaluation tag names to a set of associated values (comma separated)
+func buildTemplateDataMap(evalMatches []*EvalMatch) map[string]string {
+	var result = map[string]string{}
+	for _, match := range evalMatches {
+		for tagName, tagValue := range match.Tags {
+			// skip duplicate values
+			if strings.Contains(result[tagName], tagValue) {
```
aknuds1, line 224: AFAICT, using `strings.Contains` here is buggy, because it will match a substring of an existing value. F.ex., if you have "0.9990" in your string, "0.999" will be determined as a duplicate.

aknuds1 (changes requested):
See my failing test case, I think it makes sense.

pkg/services/alerting/eval_context.go
```diff
@@ -221,7 +221,8 @@ func buildTemplateDataMap(evalMatches []*EvalMatch) map[string]string {
 	for _, match := range evalMatches {
 		for tagName, tagValue := range match.Tags {
 			// skip duplicate values
-			if strings.Contains(result[tagName], tagValue) {
+			match := regexp.MustCompile(fmt.Sprintf("\\b%s\\b", tagValue)).FindString(result[tagName])
```
aknuds1, line 224: You should escape the value string wrt. special characters as shown below. Additionally, `reVal` should be moved outside of the loops.
```suggestion
			reVal := regexp.MustCompile(fmt.Sprintf(`\b%s\b`, regexp.QuoteMeta(tagValue)))
			match := reVal.FindString(result[tagName])
```

aknuds1 (changes requested):

pkg/services/alerting/eval_context.go
wbrowne: Ah yes that's what I was missing - thanks for the reminder. BTW the regex compilation can't be moved outside the loop as it depends on tagValue

aknuds1: @wbrowne Argh yes :man_facepalming:

aknuds1 (approved):
Think this ought to work.

marefr (changes requested):
Looks good, but have some comments in regards to invalid templating syntax, see below. Besides that documentation is needed for this feature, but can be handled in separate PR.

I did test an invalid alert title similar to how variables are referenced in the frontend `{{variable}}`. Named the title `Error on {{instance}} – above threshold` and got the following errors on the server side:
```
INFO[01-12|16:34:24] New state change                         logger=alerting.resultHandler ruleId=1862 newState=alerting prev state=unknown
DBUG[01-12|16:34:24] Rendering alert panel image              logger=alerting.notifier ruleId=1862 urlPath="d-solo/ytbo5mKGk/prometheus-alert?orgId=1&panelId=2"
WARN[01-12|16:34:24] Could not render image, no image renderer found/installed. For image rendering support please install the grafana-image-renderer plugin. Read more at https://grafana.com/
docs/grafana/latest/administration/image_rendering/ logger=rendering                           
DBUG[01-12|16:34:24] Rendered alert panel image               logger=alerting.notifier ruleId=1862 path=/home/marcus/go/src/github.com/grafana/grafana/public/img/rendering_plugin_not_installe
d.png took=33.061µs                            
DBUG[01-12|16:34:24] Uploading alert panel image to external image store logger=alerting.notifier ruleId=1862 path=/home/marcus/go/src/github.com/grafana/grafana/public/img/rendering_plugin_n
ot_installed.png                               
DBUG[01-12|16:34:24] Sending notification                     logger=alerting.notifier type=email uid=000000001 isDefault=true
EROR[01-12|16:34:24] failed trying to evaluate notification template fields logger=alerting.notifier uid=000000001 error="template: :1: function \"instance\" not defined"
EROR[01-12|16:34:24] failed to send notification              logger=alerting.notifier uid=000000001 error="template: :1: function \"instance\" not defined"
DBUG[01-12|16:34:24] Sending notification                     logger=alerting.notifier type=slack uid=000000002 isDefault=true
EROR[01-12|16:34:24] failed trying to evaluate notification template fields logger=alerting.notifier uid=000000002 error="template: :1: function \"instance\" not defined"
EROR[01-12|16:34:24] failed to send notification              logger=alerting.notifier uid=000000002 error="template: :1: function \"instance\" not defined"
DBUG[01-12|16:34:24] Sending notification                     logger=alerting.notifier type=webhook uid=7G5tf01Zz isDefault=true
EROR[01-12|16:34:24] failed trying to evaluate notification template fields logger=alerting.notifier uid=7G5tf01Zz error="template: :1: function \"instance\" not defined"
EROR[01-12|16:34:24] failed to send notification              logger=alerting.notifier uid=7G5tf01Zz error="template: :1: function \"instance\" not defined"
DBUG[01-12|16:34:24] Sending notification                     logger=alerting.notifier type=slack uid=P_vuG8rZk isDefault=true
EROR[01-12|16:34:24] failed trying to evaluate notification template fields logger=alerting.notifier uid=P_vuG8rZk error="template: :1: function \"instance\" not defined"
EROR[01-12|16:34:24] failed to send notification              logger=alerting.notifier uid=P_vuG8rZk error="template: :1: function \"instance\" not defined"
```

In the UI you cannot see that it failed to send the notification, just that something failed:
image

Everything looks okay in the alert rule state changes
image

I'm not sure we can validate this use case properly. That would be done here in that case, https://github.com/grafana/grafana/blob/21e835f3a632c1f1044a7be456ad1caeb8af4f6d/pkg/services/alerting/extractor.go#L125-L127. 

Would suggest that trying to send the notification(s) even if the templating fails, so not returning the error [here](https://github.com/grafana/grafana/pull/29908/files#diff-3539d49b70d590ded053d19235028a016ac0218933e91b3465257b4f4e6afa29R136). WDYT?

public/app/features/alerting/partials/alert_tab.html
```diff
@@ -8,7 +8,10 @@ <h4 class="section-heading">Rule</h4>
     <div class="gf-form-inline">
       <div class="gf-form">
         <span class="gf-form-label width-6">Name</span>
-        <input type="text" class="gf-form-input width-20" ng-model="ctrl.alert.name" />
+        <input type="text" class="gf-form-input width-20 gf-form-input--has-help-icon" ng-model="ctrl.alert.name"" />
```
wbrowne, line 11: @dprokop let me know if you have some extra tricks in mind for helping the user make sure they're formatting a template variable correctly. I thought about adding an `ng-blur` hook to these inputs to check in the case they have tried to use a template variable (search for `"{{"`) and then do a full regex validation to verify it's correctly written - something like `/\{\{\s*\.\w+\s*\}\}/`

marefr (approved):
LGTM

Regarding the help tooltips always hard to add as little information as possible to get the most out of it, e.g. what does `LabelName` mean in the context of an alert rule. But something is better than nothing. The documentation though would allow examples and such which should make all of this more understandable.

Since the backend logs an error if templating fails I don't feel client validation is critical do get this merged.

torkelo:
My main problem with this is that it's not using the Grafana templating syntax we use in the frontend, so users have to learn to interpolation syntaxes. 

Is there anyway we can change the golang template to use `${varname}` syntax? 
https://grafana.com/docs/grafana/latest/variables/syntax/

wbrowne:
@torkelo Yep it's not ideal. Unfortunately, Go doesn't offer that option. Maybe we can support a simple `${}` syntax on the frontend and add some translation logic on the backend to the Go syntax?

torkelo:
> @torkelo Yep it's not ideal. Unfortunately, Go doesn't offer that option. Maybe we can support a simple ${} syntax on the frontend and add some translation logic on the backend to the Go syntax?

That would mean a two way translation is needed. Can we not use a simple regex to support `${variable}` syntax like we do in the frontend instead of using go template?

wbrowne:
Thanks for the poke @torkelo. Should be simpler and more consistent now

pkg/services/alerting/eval_context.go
```diff
@@ -178,3 +179,47 @@ func getNewStateInternal(c *EvalContext) models.AlertStateType {
...
+	return s
+}
+
+// buildTemplateDataMap builds a map of alert evaluation tag names to a set of associated values (comma separated)
+func buildTemplateDataMap(evalMatches []*EvalMatch) map[string]string {
+	var result = map[string]string{}
+	for _, match := range evalMatches {
+		for tagName, tagValue := range match.Tags {
+			// skip duplicate values
+			rVal := regexp.MustCompile(fmt.Sprintf(`\b%s\b`, regexp.QuoteMeta(tagValue)))
```
kylebrandt, line 212: MustCompile panics on failure, so is only really meant when initializing a global var in something like `init()` of a package. Use `.Compile` instead and check the error.

pkg/services/alerting/eval_context.go
```diff
@@ -178,3 +179,47 @@ func getNewStateInternal(c *EvalContext) models.AlertStateType {
...
+	templateDataMap := buildTemplateDataMap(c.EvalMatches)
+	c.Rule.Message = evaluateTemplate(c.Rule.Message, templateDataMap)
+	c.Rule.Name = evaluateTemplate(c.Rule.Name, templateDataMap)
+
+	return nil
+}
+
+func evaluateTemplate(s string, m map[string]string) string {
+	for k, v := range m {
+		re := regexp.MustCompile(fmt.Sprintf(`\${%s}`, regexp.QuoteMeta(k)))
```
kylebrandt, line 199: Same here with MustCompile

pkg/services/alerting/eval_context.go
wbrowne: Good tip - thanks 😄  Just did a quick poke around and noticed the existing uses of this function in the codebase mostly ignores the err 😞  Something the linter should probably pick up..

docs/sources/alerting/notifications.md
```diff
@@ -230,3 +230,14 @@ Notification services which need public image access are marked as 'external onl
 
 All alert notifications contain a link back to the triggered alert in the Grafana instance.
 This URL is based on the [domain]({{< relref "../administration/configuration/#domain" >}}) setting in Grafana.
+
+## Templating
+
+> **Note:** Alert notification templating is only available in Grafana v7.4 and above.
+
+You can apply data templating to an alert notification's title and message by referring to the matching alert condition query's tags/labels. Template variables are formatted as follows: `${Label}`.
```
kylebrandt, line 238: Perhaps references https://github.com/grafana/grafana/blob/master/docs/sources/getting-started/timeseries-dimensions.md#labels with labels.

docs/sources/alerting/notifications.md
```diff
@@ -230,3 +230,14 @@ Notification services which need public image access are marked as 'external onl
 
 All alert notifications contain a link back to the triggered alert in the Grafana instance.
 This URL is based on the [domain]({{< relref "../administration/configuration/#domain" >}}) setting in Grafana.
+
+## Templating
```
oddlittlebird, line 234: This should be a task and explain how to use variables in alert notifications, why you would want to, and some examples. Perhaps the ones you showed in the Thursday meeting?

torkelo (approved):

docs/sources/alerting/notifications.md
```diff
@@ -230,3 +230,14 @@ Notification services which need public image access are marked as 'external onl
...
+
+> **Note:** Alert notification templating is only available in Grafana v7.4 and above.
+
+You can apply data templating to an alert notification's title and message by referring to the matching alert condition query's [labels]({{< relref "../getting-started/timeseries-dimensions.md#labels" >}}). Template variables are formatted as follows: `${Label}`.
```
marefr, line 238: Referring to "Template variables" here can be confusing and mixed up with dashboard template variables.

Nit. A note about how multiple labels are handled would be nice.

public/app/features/alerting/partials/alert_tab.html
```diff
@@ -8,7 +8,10 @@ <h4 class="section-heading">Rule</h4>
...
+        <info-popover mode="right-absolute">
+          If you want to apply templating to the alert rule name, you must use the following syntax - ${LabelName}
```
marefr, line 13: Nit. In docs the syntax is referred to as `${Label}`.

public/app/features/alerting/partials/alert_tab.html
```diff
@@ -186,11 +189,14 @@ <h4 class="section-heading">Notifications</h4>
...
     <span class="gf-form-label width-8">Message</span>
     <textarea
-      class="gf-form-input"
+      class="gf-form-input gf-form-input--has-help-icon"
       rows="10"
       ng-model="ctrl.alert.message"
       placeholder="Notification message details..."
     ></textarea>
+    <info-popover mode="right-absolute">
+      If you want to apply templating to the alert rule name, you must use the following syntax - ${LabelName}
```
marefr, line 198: Nit. In docs the syntax is referred to as `${Label}`.

marefr (approved):
LGTM, but left some comments/nits

docs/sources/alerting/notifications.md
wbrowne: Hopefully, the current state is more in line with what you're thinking @oddlittlebird. Let me know if there's anything I should change 😄

papagian (commented):
I've tested with Cloudwatch, Prometheus, Elasticsearch and works fine.
I have tested also with Cloud monitoring datasource and labels are not interpolated even though there are labels in the response.

Also I think that it would be useful to be able to include [the metric name](https://github.com/grafana/grafana/blob/1c1a800bc0562f475580a4ebba8a09dd4b3ddaaa/pkg/services/alerting/conditions/query.go#L80) in the notification title/text which is returned from all the datasource query (firing) evaluations.

Also in OK and Error notifications the variables remain un-interpolated because this information is not available then. Maybe it would be preferable to replace the variable placeholder in those cases?

For example if I have an alert with `Elastic alert ${@hostname}` as title and text.
The alerting notification looks like:
Screenshot 2021-01-18 at 8 20 59 PM

but the OK notification:
Screenshot 2021-01-18 at 8 19 00 PM

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+title = "Adding notification templating"
+keywords = ["grafana", "documentation", "alerting", "alerts", "notification", "templating"]
+weight = 110
++++
+
+# Alert notification templating
+
+In order to help provide more detailed information to alert notification recipients, you can inject alert query 
+data into an alert notification. This page explains how to use alert query labels in your alert notifications.
+
```
achatterjee-grafana, line 11:
```suggestion
You can provide detailed information to alert notification recipients by injecting alert query data into an alert notification. This topic explains how you can use alert query labels in alert notifications.

```

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+## Adding alert label data into your alert notification
+
+1. Navigate to the panel you want to add or edit an alert rule for, click the title, and then click **Edit**.
```
achatterjee-grafana, line 14:
```suggestion
1. Navigate to the panel you want to add or edit an alert rule for.
1. Click on the panel title, and then click **Edit**.
```

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+1. On the Alert tab, click **Create Alert**. If an alert already exists for this panel, then you can just edit the alert directly on the Alert tab.
```
achatterjee-grafana, line 15:
```suggestion
1. On the Alert tab, click **Create Alert**. If an alert already exists for this panel, then you can edit the alert directly.
```

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+1. Labels that exist from the evaluation of the alert query can be used in the alert rule name or alert notification message fields. Alert data template syntax is formatted as follows: `${Label}`. See the image below for an example.
```
achatterjee-grafana, line 16: This does not look like it should be a numbered step.

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+
+In order to help provide more detailed information to alert notification recipients, you can inject alert query 
+data into an alert notification. This page explains how to use alert query labels in your alert notifications.
+
+## Adding alert label data into your alert notification
+
+1. Navigate to the panel you want to add or edit an alert rule for, click the title, and then click **Edit**.
+1. On the Alert tab, click **Create Alert**. If an alert already exists for this panel, then you can just edit the alert directly on the Alert tab.
+1. Labels that exist from the evaluation of the alert query can be used in the alert rule name or alert notification message fields. Alert data template syntax is formatted as follows: `${Label}`. See the image below for an example.    
+1. When you finish writing your rule, click **Save** in the upper right corner to save the alert rule and the dashboard.
```
achatterjee-grafana, line 17:
```suggestion
1. Click **Save** in the upper right corner to save the alert rule and the dashboard.
```

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,20 @@
...
+1. Finally, when the alert is in the alerting state, the alert label data is injected into the notification fields where appropriate. In the cases where there are multiple unique label values, they will be shown in the notification as comma separated.
```
achatterjee-grafana, line 18: This again is not a numbered step.

achatterjee-grafana, line 18:
```suggestion
When the alert is in the alerting state, the alert label data is injected into the notification fields where appropriate.  Multiple unique label values are comma-separated. 
```

docs/sources/alerting/notifications.md
```diff
@@ -230,3 +230,9 @@ Notification services which need public image access are marked as 'external onl
 
 All alert notifications contain a link back to the triggered alert in the Grafana instance.
 This URL is based on the [domain]({{< relref "../administration/configuration/#domain" >}}) setting in Grafana.
+
+## Notification templating
+
+> **Note:** Alert notification templating is only available in Grafana v7.4 and above.
+
+You can [inject alert query labels into an alert notification]({{< relref "./add-notification-template.md" >}}) by referring to the matching alert condition query's [labels]({{< relref "../getting-started/timeseries-dimensions.md#labels" >}}) using the `${Label}` syntax.
```
achatterjee-grafana, line 238: It is a long and confusing sentence. Can we simplify into 2 shorter and clear sentences.

achatterjee-grafana (changes requested):
Added some copy-edit changes and a request to simplify a line.

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,23 @@
...
+title = "Alert notification templating"
+keywords = ["grafana", "documentation", "alerting", "alerts", "notification", "templating"]
+weight = 110
++++
+
+# Alert notification templating
+
+You can provide detailed information to alert notification recipients by injecting alert query data into an alert notification. 
+
+Labels that exist from the evaluation of the alert query can be used in the alert rule name or alert notification message fields. The alert label data is injected into the notification fields when the alert is in the alerting state. When there are multiple values for the same label, the values are comma-separated.
```
achatterjee-grafana:
```suggestion
Labels that exist from the evaluation of the alert query can be used in an alert rule name or in alert notification message fields. The alert label data is injected into the notification fields when the alert is in the alerting state. When there are multiple values for the same label, the values are comma-separated.
```

docs/sources/alerting/add-notification-template.md
```diff
@@ -0,0 +1,23 @@
...
+
+This topic explains how you can use alert query labels in alert notifications.
```
achatterjee-grafana, line 13: Move this to after the intro sentence "You can provide detailed information to alert notification recipients by injecting alert query data into an alert notification."

achatterjee-grafana (approved):
Looks good! Asked for two minor changes, otherwise good to go.

docs/sources/alerting/add-notification-template.md
wbrowne: I think using `the` might be more correct in this case as the alert rule name and alerting notification message are specific fields on the alert

achatterjee-grafana (approved):
LGTM

yxliaoyx:
> I've tested with Cloudwatch, Prometheus, Elasticsearch and works fine. I have tested also with Cloud monitoring datasource and labels are not interpolated even though there are labels in the response.
> 
> Also I think that it would be useful to be able to include [the metric name](https://github.com/grafana/grafana/blob/1c1a800bc0562f475580a4ebba8a09dd4b3ddaaa/pkg/services/alerting/conditions/query.go#L80) in the notification title/text which is returned from all the datasource query (firing) evaluations.
> 
> Also in OK and Error notifications the variables remain un-interpolated because this information is not available then. Maybe it would be preferable to replace the variable placeholder in those cases?
> 
> For example if I have an alert with `Elastic alert ${@hostname}` as title and text. The alerting notification looks like: Screenshot 2021-01-18 at 8 20 59 PM
> 
> but the OK notification: Screenshot 2021-01-18 at 8 19 00 PM

@papagian 
How to set labels for Elasticsearch?

papagian:
@yxliaoyx this work adds support for using labels in notifications in the legacy alerting. Do you still use this one or the [Grafana Alerting](https://grafana.com/docs/grafana/latest/alerting/)?

yxliaoyx:
@papagian I use grafana:7.5.16. I don't know why you can use `${@hostname}`? Is there any other settings?

papagian:
> @papagian I use grafana:7.5.16. I don't know why you can use `${@hostname}`? Is there any other settings?

can you share the alert query and the evaluation result?

yxliaoyx:
@papagian 

Query
image

Alert
image

My evaluation result. `tags` in `evalMatches` is empty.
```
{
  "Id": 1,
  "Version": 0,
  "OrgId": 1,
  "DashboardId": 5,
  "PanelId": 1,
  "Name": "",
  "Message": "${@hostname} ${_id} ${@timestamp} ${score}",
  "Severity": "",
  "State": "alerting",
  "Handler": 0,
  "Silenced": false,
  "ExecutionError": " ",
  "Frequency": 60,
  "For": 0,
  "EvalData": {
    "evalMatches": [
      {
        "metric": "Max score",
        "tags": {},
        "value": 0.800000011920929
      }
    ]
  },
  "NewStateDate": "2022-09-26T08:47:03Z",
  "StateChanges": 26,
  "Created": "2022-09-22T01:34:19Z",
  "Updated": "2022-09-26T08:45:54Z",
  "Settings": {
    "conditions": [
      {
        "evaluator": {
          "params": [
            0.7
          ],
          "type": "gt"
        },
        "query": {
          "datasourceId": 5,
          "model": {
            "bucketAggs": [
              {
                "field": "@timestamp",
                "id": "2",
                "settings": {
                  "interval": "auto",
                  "min_doc_count": "1"
                },
                "type": "date_histogram"
              }
            ],
            "metrics": [
              {
                "field": "score",
                "id": "1",
                "type": "max"
              }
            ],
            "query": "",
            "refId": "A",
            "timeField": "@timestamp"
          },
          "params": [
            "A",
            "1m",
            "now"
          ]
        },
        "reducer": {
          "type": "max"
        },
        "type": "query"
      }
    ],
    "for": "0",
    "frequency": "1m",
    "message": "${@hostname} ${_id} ${@timestamp} ${score}",
    "noDataState": "ok",
    "notifications": [
      {
        "uid": "notifier1"
      }
    ]
  }
}
```

papagian:
if the `hostname` label in not among [the evalMatch tags](https://github.com/grafana/grafana/blob/dc3eee00a74754dbed05d98ffb46f65ad0addaf7/pkg/services/alerting/eval_context.go#L226) then it won't get interpolated

yxliaoyx:
@papagian So it is impossible in this version? Or I should do something on Grafana or ElasticSearch.

wbrowne:
@yxliaoyx For this to work the eval match would need to look something like:

```
 "EvalData": {
    "evalMatches": [
      {
        "metric": "Max score",
        "tags": {
          "@hostname": "abc",
          "score": 123,
         },
        "value": 0.800000011920929
      }
    ]
  },
```

But there are no tags for your matched query

yxliaoyx:
@wbrowne It only works with `Prometheus` datasource? Or I can use `ElasticSearch` datasource?

wbrowne:
@yxliaoyx  It works with any data source that returns an alert evaluation query response with tag information

rlevytskyi:
Hi All,
I'm using Grafana 10.1 and I was able to define query and extract some results (say, Count=2).
Is there a way I can extract not only the Count but result itself to use some fields from it in my alert messages?

PR #29911, opened 2020-12-17
Docs: Update organization_roles.md

oddlittlebird:
A user on public Slack pointed out that we do not say anywhere in the docs that viewers_can_edit allows Viewers to use Explore. Fixing that now.


matthewhelmke (approved):
LGTM

PR #29916, opened 2020-12-17
Plugins: Convert test data source to use SDK contracts

marefr:
Converts the core testdata data source to use the SDK contracts and by that implementing a backend plugin in core Grafana in similar manner as an external one.

Challenges:
- Testdata data source query model include a _scenarioId_ which is the value for the selected scenario in the query editor. In the SDK there's a [QueryTypeMux](https://github.com/grafana/grafana-plugin-sdk-go/blob/master/backend/datasource/query_type_mux.go) available that you can use for routing certain query types to different query data handlers, however to be able to use this you have to use the _queryType_ property on the query model from the UI. What we ultimately would like to do is to migrate _scenarioId_ to _queryType_, but that could require breaking changes. You can see an example of possible implementation with _queryType_ is use [here](https://github.com/grafana/test-datasource/blob/821d03979cf5944a1a94f25072a38a823c8d5dad/pkg/ds/scenario/scenario.go#L34-L37).

**Temporary testdata:**
Converts testdata plugin using the SDK contracts, this is coming from an old PR #25472 which had a commit that was never merged to master (https://github.com/grafana/grafana/pull/25472/commits/ee30b7ebd2e553411ce4dd2b10f29964c939ecc3). This is mainly for testing/verifying implementing a core backend plugins using the SDK contracts works as expected.

Currently uses a temporary hack to reuse the scenario handlers with tsdb models and converting series to data frames.

There's also some test resources defined that you can call to try it out:
- http://localhost:3000/api/plugins/testdata/resources
- http://localhost:3000/api/plugins/testdata/resources/test
- http://localhost:3000/api/plugins/testdata/resources/test/json
- http://localhost:3000/api/plugins/testdata/resources/boom
- http://localhost:3000/api/plugins/testdata/resources/stream?count=10&sleep=500ms

Fixes #

- Still the query model only uses _scenarioId_ and not queryType, but idea is to migrate it to use queryType but is mostly a frontend refactor. It currently fallbacks to _scenarioId_.
- #30250 changed to use json-iterator library for JSON encoding. That changed map keys in responses coming from /api/ds/query to be non-ordered. Fixed in separate PR by #30732


packages/grafana-runtime/src/utils/queryResponse.ts
```diff
@@ -93,6 +94,79 @@ export function toDataQueryResponse(res: any): DataQueryResponse {
...
 }
 
+/**
+ * Parse the results from /api/ds/query into a DataQueryResponse ordering it by query input RefID's.
+ *
+ * @public
+ */
+export function toDataQueryResponseOrdered(queries: DataQuery[], res: any): DataQueryResponse {
+  const rsp: DataQueryResponse = { data: [], state: LoadingState.Done };
+  if (res.data?.results) {
```
hugohaggmark, line 104: nit: Maybe do a quick exit here
```suggestion
  if (!res.data?.results) {
    return ...
  }
```

packages/grafana-runtime/src/utils/queryResponse.ts
```diff
@@ -93,6 +94,79 @@ export function toDataQueryResponse(res: any): DataQueryResponse {
...
+    const results: KeyValue = res.data.results;
+    for (const q of queries) {
+      const refId = q.refId;
+      const dr = results[refId] as DataResponse;
+      if (dr) {
```
hugohaggmark, line 109: nit: Maybe do a quick exit here if possible too.
```suggestion
      if (!dr) {
        continue;
      }
```

marefr: Copied the code from https://github.com/grafana/grafana/blob/1d399b89eab6ad276e06cc85bc308ecd2aea2776/packages/grafana-runtime/src/utils/queryResponse.ts#L29 so haven't considered any changes to it, but it makes sense.

marefr: The question is more does it make sense to have a `toDataQueryResponseOrdered` function in combination with `toDataQueryResponse` in the grafana/runtime package?

hugohaggmark: Hmm, yes, that is a good question. Something feels a bit odd. Why should I as a consumer have to think about order? 
Shouldn't this be the default behavior handled by the framework in `toDataQueryResponse`, but there are probably valid use cases I'm not understanding, like streaming?

marefr: Discussed with Ryan and he suggested to update `toDataQueryResponse` to take an optional queries argument rather than adding a new method `toDataQueryResponseOrdered`.

marefr: Fixed suggestion in updated `toDataQueryResponse`

packages/grafana-runtime/src/utils/queryResponse.ts
marefr: Not sure what to return here if exit early. The not ok response could still be applicable if data contains results

public/app/plugins/datasource/testdata/datasource.ts
```diff
@@ -80,30 +80,16 @@ export class TestDataDataSource extends DataSourceApi<TestDataQuery> {
...
-        .fetch({
-          method: 'POST',
-          url: '/api/tsdb/query',
-          data: {
-            from: options.range.from.valueOf().toString(),
-            to: options.range.to.valueOf().toString(),
-            queries: queries,
-          },
-        })
-        .pipe(map((res) => this.processQueryResult(queries, res)));
```
ryantxu, line 104: Can we also remove the `processQueryResult` function and the unused `TestData` type?

packages/grafana-runtime/src/utils/queryResponse.ts
```diff
@@ -24,56 +25,63 @@ interface DataResponse {
...
         }
+      }
 
-        if (dr.tables && dr.tables.length) {
-          for (const s of dr.tables) {
-            if (!s.refId) {
-              s.refId = refId;
-            }
-            rsp.data.push(toDataFrame(s));
+      if (dr.tables && dr.tables.length) {
```
ryantxu, line 64:
```suggestion
      if (dr.tables?.length) {
```
maybe 6 vs 1/2 dozen

packages/grafana-runtime/src/utils/queryResponse.ts
```diff
@@ -24,56 +25,63 @@ interface DataResponse {
...
         }
+      }
 
-        if (dr.series && dr.series.length) {
-          for (const s of dr.series) {
-            if (!s.refId) {
-              s.refId = refId;
-            }
-            rsp.data.push(toDataFrame(s));
+      if (dr.series && dr.series.length) {
```
ryantxu, line 55:
```suggestion
      if (dr.series?.length) {
```

packages/grafana-runtime/src/utils/queryResponse.ts
```diff
@@ -24,56 +25,63 @@ interface DataResponse {
...
+ *
  * @public
  */
-export function toDataQueryResponse(res: any): DataQueryResponse {
+export function toDataQueryResponse(res: any, queries?: DataQuery[]): DataQueryResponse {
   const rsp: DataQueryResponse = { data: [], state: LoadingState.Done };
   if (res.data?.results) {
     const results: KeyValue = res.data.results;
-    for (const refId of Object.keys(results)) {
+    const refIDs = queries ? queries.map((q) => q.refId) : Object.keys(results);
```
ryantxu, line 37: does the backend now guarantee that everything has a refId set?  If not, this makes me a bit nervous

ryantxu (approved):
this is a huge step forward -- now we can easily test with the data our SDK can produce!

packages/grafana-runtime/src/utils/queryResponse.ts
hugohaggmark: Hmm, but this is the case when there are no results. So you should just return `rsp`, I guess?

hugohaggmark: I see now this is outdated, never mind.

papagian (approved):
Remarkable work! Modifications look good and I have tested and works great: my existing dashboards/alerts (using the test datasource) work as expected. I've also tested all the test scenarios (expect Grafana Live that it's not setup) in a new panel. Finally, I have tested successfully to use the test datasource in expressions and in an alert definition (alerting NG).

packages/grafana-runtime/src/utils/queryResponse.ts
marefr: It depends on how you see it I guess. The tsdb response model has a results map of refID's with query result:
https://github.com/grafana/grafana/blob/1bdd3eb3dd096862bebf6af59d68ff344cf3693c/pkg/tsdb/models.go#L33-L36

There's nothing in the backend that guarantees that only RefID's part of the queries are part of the response. Very much   up to the developer I would say, but still this should be "best practices" and documented at least. It's not really a new thing that you must return the result with RefID's of the queries to make certain feature in Grafana UI function as expected, for example error handling and so on.

To minimize the risk we could check here if there's any other RefID's in the response and include those in the result of the function. WDYT?

hugohaggmark (approved):
Outstanding work on this 🎉

PR #29935, opened 2020-12-19
Chore: Convert tsdb interval test to standard library

ceh:

Convert tsdb interval test to standard library.

Part of #22844.

I split up the single test function into three separate tests, where each test corresponds to a single method or function.

I also transformed the tests to be table-driven tests, which will simplify adding more test cases. Each test case is executed by a sub-test (`t.Run`), which may be a bit complicated/excessive, but I think it's preferred  in the codebase? (since it allows running the test case individually amongst other things)

### DIFF
--- a/pkg/tsdb/interval_test.go
+++ b/pkg/tsdb/interval_test.go
@@ -4,59 +4,64 @@ import (
 	"testing"
 	"time"
 
-	"github.com/grafana/grafana/pkg/setting"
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
Looks great! I don't know why you think `t.Run` would be excessive on individual test cases, I think it's exactly what it's for?

aknuds1:
Thanks!

PR #29936, opened 2020-12-19
Chore: Convert tsdb request test to standard library

ceh:

Convert tsdb request test to standard library.

Part of #22844.

Renamed the test to `TestHandleRequest` because that's the method under test. Also renamed the file to `request_test.go` because `HandleRequest` is defined in https://github.com/grafana/grafana/blob/c4e141c112f797a9c4724311e9617157964f5c18/pkg/tsdb/request.go#L12, which ought to make it a bit easier to find any related test code.

What do you think?

### DIFF
--- a/pkg/tsdb/request_test.go
+++ b/pkg/tsdb/request_test.go
@@ -5,11 +5,11 @@ import (
 	"testing"
 
 	"github.com/grafana/grafana/pkg/models"
-	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/require"
 )
 
-func TestMetricQuery(t *testing.T) {
-	Convey("When executing request with one query", t, func() {
+func TestHandleRequest(t *testing.T) {
+	t.Run("Should return query result when handling request for query", func(t *testing.T) {
 		req := &TsdbQuery{
 			Queries: []*Query{
 				{RefId: "A", DataSource: &models.DataSource{Id: 1, Type: "test"}},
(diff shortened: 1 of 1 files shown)


pkg/tsdb/request_test.go
```diff
@@ -41,23 +38,22 @@ func TestMetricQuery(t *testing.T) {
...
+	t.Run("Should return error when handling request for query with unknown type", func(t *testing.T) {
 		req := &TsdbQuery{
 			Queries: []*Query{
 				{RefId: "A", DataSource: &models.DataSource{Id: 1, Type: "asdasdas"}},
 			},
 		}
 
 		_, err := HandleRequest(context.TODO(), &models.DataSource{Id: 12, Type: "testjughjgjg"}, req)
-		So(err, ShouldNotBeNil)
+		require.NotNil(t, err)
```
aknuds1, line 56:
```suggestion
		require.Err(t, err)
```

aknuds1 (approved):
Looks great, except for an error assertion, where I think it's more precise to use `require.Error`.

aknuds1 (changes requested):
Looks like I can't apply my suggestion myself before merging, so please just commit it before I re-review :)

pkg/tsdb/request_test.go
```diff
@@ -20,15 +20,12 @@ func TestMetricQuery(t *testing.T) {
 		fakeExecutor.Return("A", TimeSeriesSlice{&TimeSeries{Name: "argh"}})
 
 		res, err := HandleRequest(context.TODO(), &models.DataSource{Id: 1, Type: "test"}, req)
-		So(err, ShouldBeNil)
-
-		Convey("Should return query results", func() {
-			So(res.Results["A"].Series, ShouldNotBeEmpty)
-			So(res.Results["A"].Series[0].Name, ShouldEqual, "argh")
-		})
+		require.Nil(t, err)
```
aknuds1, line 23:
```suggestion
		require.NoError(t, err)
```

pkg/tsdb/request_test.go
```diff
@@ -41,23 +38,22 @@ func TestMetricQuery(t *testing.T) {
 		fakeExecutor.Return("B", TimeSeriesSlice{&TimeSeries{Name: "barg"}})
 
 		res, err := HandleRequest(context.TODO(), &models.DataSource{Id: 1, Type: "test"}, req)
-		So(err, ShouldBeNil)
+		require.Nil(t, err)
```
aknuds1, line 41:
```suggestion
		require.NoError(t, err)
```

aknuds1 (changes requested):

pkg/tsdb/request_test.go
ceh: 👍  Done.

pkg/tsdb/request_test.go
ceh: 👍  Done.

pkg/tsdb/request_test.go
ceh: 👍  Done.

aknuds1 (approved):
LGTM

aknuds1:
Thanks for cleaning things up!

PR #29939, opened 2020-12-20
Drone: Update pipelines for Enterprise

AgnesToulet:
Update Drone pipelines to add steps (test / build / package) for an Enterprise version with build tags. 
Plus, update E2E tests so that PACKAGE_FILE and RUNDIR variables can be set as environment variables.


bergquist (commented):
From my POV this seems like a good way to build them in parallel but I'm definitely not an expert on drone :) @aknuds1 you will be the judge. 

What I would say thou I that I would prefer to merge whatever works first and maybe refactor later.

aknuds1:
@bergquist I've reviewed it, it looks generally pretty good. I want to take the release pipeline for a test spin though, which will require some work.

aknuds1 (commented):
Looks pretty good now, but we should merge the enterprise PR first and for that we have to fix e2e tests. After that I want to do a simulated release to see that everything works. I read through all the release simulation code and fixed the issues I found, so that should be good at this point.

aknuds1 (approved):
LGTM - think we might have to merge this to proceed with the Enterprise PR.

PR #29944, opened 2020-12-21
PanelLibrary: Adds uid and renames title to name

hugohaggmark:
Followup PR to this comment: https://github.com/grafana/grafana/pull/29772#pullrequestreview-552220950. The field Title is renamed Name, and we added a UID field. 
Since this entire feature in behind a feature toggle, we don't currently care about doing proper database migrations (yet).

Relates #29610


torkelo (approved):
Looks good!

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -64,37 +72,43 @@ func TestGetLibraryPanel(t *testing.T) {
 		func(t *testing.T, sc scenarioContext) {
 			response := sc.service.getHandler(sc.reqContext)
 			require.Equal(t, 404, response.Status())
-		}, scenarioConfig{params: map[string]string{":panelId": "74"}})
```
aknuds1, line 67: Why are you removing the panelId parameter instead of renaming it to uid? It seems to me you are introducing undefined behaviour into this scenario, by removing the parameter saying which panel to get.

aknuds1 (commented):
Wondering about the removal of a test parameter, that seems crucial wrt the scenario to me.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Just keeping it simple and that made it so I could remove the scenarioConfig completely. Might reintroduce it when needed. Also the `uid` in this case will be the empty string right?

pkg/services/librarypanels/librarypanels.go
```diff
@@ -52,7 +52,8 @@ func (lps *LibraryPanelService) AddMigration(mg *migrator.Migrator) {
 			{Name: "id", Type: migrator.DB_BigInt, IsPrimaryKey: true, IsAutoIncrement: true},
 			{Name: "org_id", Type: migrator.DB_BigInt, Nullable: false},
 			{Name: "folder_id", Type: migrator.DB_BigInt, Nullable: false},
-			{Name: "title", Type: migrator.DB_NVarchar, Length: 255, Nullable: false},
+			{Name: "uid", Type: migrator.DB_NVarchar, Length: 40, Nullable: false},
```
wbrowne, line 55: Since we're updating the table structure, don't we need to have a new table version (or alternatively just some `ALTER TABLE` commands)? IE
```go
mg.AddMigration("Add uid column", NewAddColumnMigration(libraryPanelV1, &Column{
	Name: "library_panel", Type: DB_NVarchar, Length: 40, Nullable: false,
}))

mg.AddMigration("Rename title column", ...
```

For places where this migration has already successfully run, this would fail AFAICT

hugohaggmark: That is correct, but as this is still behind a feature flag, we're taking this risk.

aknuds1: @wbrowne I think the idea is we can ignore failing migrations with this PR, since this feature is only enabled by developers.

hugohaggmark:
Anymore feedback before I merge this @aknuds1 @wbrowne?

pkg/services/librarypanels/librarypanels_test.go
aknuds1: @hugohaggmark My argument is that it's still needed. Why should this scenario pass an empty string for uid? The scenario is getting a non-existent library panel, not omitting the uid parameter. Omitting it should be a validation error AFAICT.

aknuds1: My opinion here is that the test case should pass a UID for the parameter, it will just not exist as in the code prior to the PR.

aknuds1:
> Anymore feedback before I merge this @aknuds1 @wbrowne?

See my comment, I think the panel UID should be provided in the test case. Removing it looks wrong to me.

pkg/services/librarypanels/librarypanels.go
wbrowne: Always read the label! My bad. Looking good in that case @hugohaggmark!

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Ok, I'll see what I can come up with :)

hugohaggmark: Fixed in the latest commit

aknuds1 (approved):

Issue #29943, opened 2020-12-21
Advanced variable format options issue v.7.3.6


nikola-farshore:
**What happened**:
URL: .....&var-MultiDeviceId=A&var-MultiDeviceId=B
'${MultiDeviceId:raw}' variable format returns:
('A,B')

**What you expected to happen**:
URL: .....&var-MultiDeviceId=A&var-MultiDeviceId=B
'${MultiDeviceId:raw}' variable format should return:
('{A,B}') 

**Environment**:
- Grafana version: Version 7.3.6 (commit: ea06633c34, branch: HEAD)
- Data source type & version: PostgreSQL 11 - Timescale
- OS Grafana is installed on: Centos 7
- User OS & Browser: macOS Big Sur 11.0.1 / Firefox 84.0 (64-bit)

dprokop:
@nikola-farshore thanks for opening the issue. Do you mind sharing the query you are trying to use the `${MultiDeviceId:raw}` in? I think the format is OK and it hasn't been changed for a while. I'm also trying to figure out why the [docs](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/#raw) contains `{}` in the raw syntax description.

nikola-farshore:
`SELECT 
  $__timeGroup("tm", $__interval, 0) as "time", 
  AVG("hist") as "Hourly Avg" 
FROM 
  historicalavgmultiple('${MultiDeviceId:raw}') 
GROUP BY 1 
ORDER BY 1;`

In the previous version/s, list of variables passed to function were wrapped into `{}` so PostgreSQL recognized it as array list.

dprokop:
@nikola-farshore thanks for a quick reply. Trying to uderstad what's going on. When you say "In the previous version/s" what versions do you mean?

nikola-farshore:
Not sure about previous grafana version to be honest, but we were using advanced variable format for a while, so it must be v6 and v7 as well, up to v.7.3.6.

I found old screenshot, so you can check it as a reference:
![Screenshot 2020-12-21 at 13 25 18](https://user-images.githubusercontent.com/53433107/102777198-6c85ed00-4390-11eb-9a8a-d9e3efe3fe80.png)

dprokop:
@nikola-farshore thanks a lot for the screenshot. Seems like it's v7 so i'm gonna dig a little bit deeper. Maybe there has been some changes to the Postgres datasouce that I'm not aware of. I'll let you know when I figure this out.

[...]

PR #29945, opened 2020-12-21
Fix raw format variable docs

dprokop:
Seems like [raw variable formatting docs](https://grafana.com/docs/grafana/latest/variables/advanced-variable-format-options/#raw) are and were incorrect for some time. This probably leads to confusion as the one described in https://github.com/grafana/grafana/issues/29943


hugohaggmark (approved):
LGTM!

oddlittlebird (approved):
Good catch, @dprokop !

PR #29946, opened 2020-12-21
RedirectResponse: Implement all of api.Response

aknuds1:
Fix building of Enterprise, by making RedirectResponse implement all of the api.Response interface.

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -144,17 +144,26 @@ func Respond(status int, body interface{}) *NormalResponse {
 	}
 }
 
+// RedirectResponse represents a redirect response.
 type RedirectResponse struct {
 	location string
 }
 
+// WriteTo writes to a response.
 func (r *RedirectResponse) WriteTo(ctx *models.ReqContext) {
 	ctx.Redirect(r.location)
 }
 
 // Status gets the response's status.
+// Required to implement api.Response.
 func (*RedirectResponse) Status() int {
 	return http.StatusFound
+}
+
+// Body gets the response's body.
+// Required to implement api.Response.
+func (r *RedirectResponse) Body() []byte {
+	return nil
 }
 
 func Redirect(location string) *RedirectResponse {


pkg/api/common.go
```diff
@@ -144,19 +144,28 @@ func Respond(status int, body interface{}) *NormalResponse {
...
 	location string
 }
 
+// WriteTo writes to a response.
 func (r *RedirectResponse) WriteTo(ctx *models.ReqContext) {
 	ctx.Redirect(r.location)
 }
 
 // Status gets the response's status.
+// Required to implement api.Response.
```
sakjur, line 158: Suggestion: Instead of a comment, add `var _ Response = &RedirectResponse{}` to the top of some test file to show the intent of implementing api.Response.

sakjur (approved):

pkg/api/common.go
aknuds1: @sakjur That is a bit ugly. I prefer documenting explicitly each method that corresponds to an interface, because you see why the methods are necessary. If you are linting for unused methods, it really helps to see that methods that look unused implement part of an interface.

If you want to make sure a type implements an interface OTOH, you can also embed the interface in the type. Then you get automatic (unfulfilled) implementations of interface methods you don't implement yourself. That sounds like an alternative to what you propose. I couldn't do that though due to an import cycle.

sakjur: :+1:

That makes sense

sakjur: (I'm _really_ not a fan of embedding  types though 😛)

aknuds1: @sakjur what do you have against it? It's very common in Go, and super useful when you want to make fake implementations of larger interfaces f.ex.

sakjur: I don't think I've ever seen type embedding in a setting that didn't confuse me, and I've yet to see a single case where I thought that using embedding solved a problem that wouldn't be better solved with composition. Having it there allows for Java-esque inheritance-based designs (Macaron, for example, does this a lot), and Go is not a very good language (it loses it's readability, which is really Go's strong side according to me) when you start applying pre-8 Java design principles.

I can definitely see the point for making fake implementations of larger interfaces, but to me that's patching over a design flaw with a hack (which, to be clear, is an ok thing to do, that doesn't make me a fan of the feature though it might help me understand why it can be useful at times).

And providing methods that doesn't actually exist to a type seem like something that might make sense in tests, but in production code it seems like a good way of getting somewhat hard to find segfaults (for example, I would want the following to not compile rather than segfault during runtime https://play.golang.org/p/q89RtjZgujF ).

Should it never be used? Not saying that. I've just yet to find myself wanting it and when I've found it in code it has almost always been in the context of "oh. that'll extend my debugging by five minutes" rather than "that's neat", and I'm not sure I'm convinced that it provides sufficient value to outweigh the costs of having the syntax in Go.

aknuds1: @sakjur Wrt. embedding the interface in this type, is segfaulting so much worse than returning an error for not-implemented methods? IMO in practice a segfault can be better, since "soft" errors can often be swallowed in Go.

sakjur: I agree with what you're saying, but that's not necessarily the alternative, something like

```go
func (t T) Method() {
panic("unexpected call for Method() from Interface, panicking")
}
```

would be an acceptable alternative here. Having the method explicitly panicking means that you can figure out that the type _can_ panic, and more importantly if the interface changes, you'll be forced to update the implementation alongside it for the application to compile again. An empty variable declaration is a way of forcing that compile time check to happen, if you return errors, empties, or panic on unimplemented methods is left to the judgment of the programmer (that's a point in favor of embedding the interface — the behavior of unimplemented methods is defined).

```go
var _ Interface = &Impl{}
```

aknuds1: @sakjur I think there are dynamic cases where it isn't discovered until runtime that a type doesn't implement an interface? I'm worried about the compiler not necessarily being able to catch if a type doesn't implement an interface.

sakjur: :100:

Definitely 🙂

PR #29954, opened 2020-12-22
librarypanels: Fix JSON field casing in tests

aknuds1:
The librarypanels package uses Go case for JSON fields in a test struct, this PR fixes it to use casing idiomatic to JSON.


sunker (approved):
LGTM

PR #29955, opened 2020-12-22
TemplateSrv: Do not throw error for an unknown format but use glob as a fallback

### LINKED ISSUE #29951: Grafana crashes when using a nonexistent variable format in a link
**What happened**: If you use a nonexistent variable format in a panel's link, the whole page crashes and shows this error when you hover the corner:
https://user-images.githubusercontent.com/9257800/102831528-a61b2000-43b1-11eb-93b6-534f7eb28b87.mov

A few possibilities:
- The link on the panel is shown as an error instead
- The link still works but the incorrect variable is not interpolated
- The URL entry field has validation and doesn't let you add a broken variable

Add a panel link using a variable with a format that doesn't exist:
Screen Shot 2020-12-21 at 5 21 14 PM

**Environment**:
- Grafana version: 7.3.6
- OS Grafana is installed on: OSX
- User OS & Browser: OSX & Chrome

dprokop:
Fixes https://github.com/grafana/grafana/issues/29951

Instead of throwing an error, log the problem to the console and use glob as a fallback

### DIFF
--- a/public/app/features/templating/template_srv.test.ts
+++ b/public/app/features/templating/template_srv.test.ts
@@ -295,6 +295,13 @@ describe('templateSrv', () => {
       expect(result).toBe('test');
     });
 
+    it('should use glob format when unknown format provided', () => {
+      let result = _templateSrv.formatValue('test', 'nonexistentformat');
+      expect(result).toBe('test');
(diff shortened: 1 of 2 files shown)


hugohaggmark (approved):
Outstanding work ⭐

PR #29956, opened 2020-12-22
PanelLibrary: Add PATCH to the API

hugohaggmark:
This PR adds `PATCH` to the Panel Library API.

Relates to #29610


pkg/services/librarypanels/database.go
```diff
@@ -63,26 +63,31 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, uid str
 	})
 }
 
+func getLibraryPanel(session *sqlstore.DBSession, uid string, orgID int64) (LibraryPanel, error) {
+	libraryPanels := make([]LibraryPanel, 0)
+	session.Table("library_panel")
+	session.Where("uid=? and org_id=?", uid, orgID)
```
aknuds1, line 69:
```suggestion
	session.Where("uid=? AND org_id=?", uid, orgID)
```

pkg/services/librarypanels/database.go
```diff
@@ -104,3 +108,54 @@ func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]Lib
...
+		}
+
+		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
+			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
+			return err
+		} else if len(res) == 1 {
+			return errLibraryPanelAlreadyExists
+		}
+
+		if rowsAffected, err := session.ID(panelInDb.ID).Update(&libraryPanel); err != nil {
```
aknuds1, line 151:
```suggestion
		panelInDB, err := getLibraryPanel(session, cmd.UID, c.SignedInUser.OrgId)
		if err != nil {
			return err
		}

		libraryPanel = LibraryPanel{
			ID:        panelInDB.ID,
			OrgID:     c.SignedInUser.OrgId,
			FolderID:  cmd.FolderID,
			UID:       cmd.UID,
			Name:      cmd.Name,
			Model:     cmd.Model,
			Created:   panelInDB.Created,
			CreatedBy: panelInDB.CreatedBy,
			Updated:   time.Now(),
			UpdatedBy: c.SignedInUser.UserId,
		}

		if cmd.FolderID == 0 {
			libraryPanel.FolderID = panelInDB.FolderID
		}
		if cmd.Name == "" {
			libraryPanel.Name = panelInDB.Name
		}
		if cmd.Model == nil {
			libraryPanel.Model = panelInDB.Model
		}

		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
			return err
		} else if len(res) == 1 {
			return errLibraryPanelAlreadyExists
		}

		if rowsAffected, err := session.ID(panelInDB.ID).Update(&libraryPanel); err != nil {
```
```suggestion
		panelInDb, err := getLibraryPanel(session, cmd.UID, c.SignedInUser.OrgId)
		if err != nil {
			return err
		}

		libraryPanel = LibraryPanel{
			ID:        panelInDb.ID,
			OrgID:     c.SignedInUser.OrgId,
			FolderID:  cmd.FolderID,
			UID:       cmd.UID,
			Name:      cmd.Name,
			Model:     cmd.Model,
			Created:   panelInDb.Created,
			CreatedBy: panelInDb.CreatedBy,
			Updated:   time.Now(),
			UpdatedBy: c.SignedInUser.UserId,
		}

		if cmd.FolderID == 0 {
			libraryPanel.FolderID = panelInDb.FolderID
		}
		if cmd.Name == "" {
			libraryPanel.Name = panelInDb.Name
		}
		if cmd.Model == nil {
			libraryPanel.Model = panelInDb.Model
		}

		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
			return err
		} else if len(res) == 1 {
			return errLibraryPanelAlreadyExists
		}

		if rowsAffected, err := session.ID(panelInDb.ID).Update(&libraryPanel); err != nil {
```
```suggestion
		panelInDb, err := getLibraryPanel(session, cmd.UID, c.SignedInUser.OrgId)
		if err != nil {
			return err
		}

		libraryPanel = LibraryPanel{
			ID:        panelInDb.ID,
			OrgID:     c.SignedInUser.OrgId,
			FolderID:  cmd.FolderID,
			UID:       cmd.UID,
			Name:      cmd.Name,
			Model:     cmd.Model,
			Created:   panelInDb.Created,
			CreatedBy: panelInDb.CreatedBy,
			Updated:   time.Now(),
			UpdatedBy: c.SignedInUser.UserId,
		}

		if cmd.FolderID == 0 {
			libraryPanel.FolderID = panelInDb.FolderID
		}
		if cmd.Name == "" {
			libraryPanel.Name = panelInDb.Name
		}
		if cmd.Model == nil {
			libraryPanel.Model = panelInDb.Model
		}

		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
			return err
		} else if len(res) == 1 {
			return errLibraryPanelAlreadyExists
		}

		if rowsAffected, err := session.ID(panelInDb.ID).Update(&libraryPanel); err != nil {
```
```suggestion
		panelInDb, err := getLibraryPanel(session, cmd.UID, c.SignedInUser.OrgId)
		if err != nil {
			return err
		}

		libraryPanel = LibraryPanel{
			ID:        panelInDb.ID,
			OrgID:     c.SignedInUser.OrgId,
			FolderID:  cmd.FolderID,
			UID:       cmd.UID,
			Name:      cmd.Name,
			Model:     cmd.Model,
			Created:   panelInDb.Created,
			CreatedBy: panelInDb.CreatedBy,
			Updated:   time.Now(),
			UpdatedBy: c.SignedInUser.UserId,
		}

		if cmd.FolderID == 0 {
			libraryPanel.FolderID = panelInDb.FolderID
		}
		if cmd.Name == "" {
			libraryPanel.Name = panelInDb.Name
		}
		if cmd.Model == nil {
			libraryPanel.Model = panelInDb.Model
		}

		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
			return err
		} else if len(res) == 1 {
			return errLibraryPanelAlreadyExists
		}

		if rowsAffected, err := session.ID(panelInDb.ID).Update(&libraryPanel); err != nil {
```

aknuds1, line 151:
```suggestion
		panelInDB, err := getLibraryPanel(session, cmd.UID, c.SignedInUser.OrgId)
		if err != nil {
			return err
		}

		libraryPanel = LibraryPanel{
			ID:        panelInDB.ID,
			OrgID:     c.SignedInUser.OrgId,
			FolderID:  cmd.FolderID,
			UID:       cmd.UID,
			Name:      cmd.Name,
			Model:     cmd.Model,
			Created:   panelInDB.Created,
			CreatedBy: panelInDB.CreatedBy,
			Updated:   time.Now(),
			UpdatedBy: c.SignedInUser.UserId,
		}

		if cmd.FolderID == 0 {
			libraryPanel.FolderID = panelInDB.FolderID
		}
		if cmd.Name == "" {
			libraryPanel.Name = panelInDB.Name
		}
		if cmd.Model == nil {
			libraryPanel.Model = panelInDB.Model
		}

		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
			return err
		} else if len(res) == 1 {
			return errLibraryPanelAlreadyExists
		}

		if rowsAffected, err := session.ID(panelInDB.ID).Update(&libraryPanel); err != nil {
```

pkg/services/librarypanels/database.go
```diff
@@ -104,3 +108,54 @@ func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]Lib
...
+			UID:       cmd.UID,
+			Name:      cmd.Name,
+			Model:     cmd.Model,
+			Created:   panelInDb.Created,
+			CreatedBy: panelInDb.CreatedBy,
+			Updated:   time.Now(),
+			UpdatedBy: c.SignedInUser.UserId,
+		}
+
+		if cmd.FolderID == 0 {
```
aknuds1, line 134: You shouldn't _conditionally_ update a resource when handling a PUT request, the semantics of PUT in REST is to update the _whole_ resource. If you want partial updates, you should use PATCH.

pkg/services/librarypanels/database.go
```diff
@@ -104,3 +108,54 @@ func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]Lib
...
+		if cmd.Name == "" {
+			libraryPanel.Name = panelInDb.Name
+		}
+		if cmd.Model == nil {
+			libraryPanel.Model = panelInDb.Model
+		}
+
+		if res, err := session.Query("SELECT 1 FROM library_panel WHERE org_id=? AND folder_id=? AND name=? AND uid <>?",
+			libraryPanel.OrgID, libraryPanel.FolderID, libraryPanel.Name, libraryPanel.UID); err != nil {
+			return err
```
aknuds1, line 146: Shouldn't this be handled via  UNIQUE constraint in the database? Checking this in the backend code looks really fragile to me. @sakjur @wbrowne thoughts?

pkg/services/librarypanels/models.go
```diff
@@ -37,3 +37,11 @@ type createLibraryPanelCommand struct {
 	Name     string          `json:"name"`
 	Model    json.RawMessage `json:"model"`
 }
+
+// updateLibraryPanelCommand is the command for updating a LibraryPanel
+type updateLibraryPanelCommand struct {
+	FolderID int64           `json:"folderId"`
+	UID      string          `json:"-"`
```
aknuds1, line 44: This shouldn't be on the command, we should do as for the other *Command types in this package, and pass UID as a separate parameter.

pkg/services/librarypanels/api.go
```diff
@@ -72,3 +73,20 @@ func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response
 
 	return api.JSON(200, util.DynMap{"result": libraryPanels})
 }
+
+// updateHandler handles PUT /api/library-panels/:uid
+func (lps *LibraryPanelService) updateHandler(c *models.ReqContext, cmd updateLibraryPanelCommand) api.Response {
+	cmd.UID = c.Params(":uid")
```
aknuds1, line 79: We should be consistent with the other methods, and pass UID as a separate parameter. It's also less confusing to readers of the code if `cmd` corresponds 1:1 to request parameters (single responsibility).

aknuds1 (changes requested):
I think the PUT verb is being implemented in the wrong way; if you want to support partial updates, it should be the [PATCH](https://restcookbook.com/HTTP%20Methods/patch/) verb as per the standard REST semantics.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
 		})
 }
 
+func TestUpdateLibraryPanel(t *testing.T) {
+	testScenario(t, "When an admin tries to update a library panel that does not exist, it should fail",
+		func(t *testing.T, sc scenarioContext) {
+			cmd := updateLibraryPanelCommand{}
+			response := sc.service.updateHandler(sc.reqContext, cmd)
```
aknuds1, line 185: You should inject a UID here, because the case shouldn't be the UID parameter being omitted, but a non-existent UID being specified.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+							`),
+			}
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": result.Result.UID})
+			response = sc.service.updateHandler(sc.reqContext, cmd)
+			require.Equal(t, 200, response.Status())
+			err = json.Unmarshal(response.Body(), &result)
+			require.NoError(t, err)
+			require.Equal(t, int64(2), result.Result.FolderID)
+			require.Equal(t, "Panel - New name", result.Result.Name)
+			require.Equal(t, "Model - New name", result.Result.Model["name"])
```
aknuds1, line 218: Just a suggestion, but I think it's better to use [go-cmp](https://github.com/google/go-cmp) here, as we do elsewhere in the project, to verify the whole object, not just selected fields.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+		})
+
+	testScenario(t, "When an admin tries to update a library with folder only, it should change folder successfully and return correct result",
```
aknuds1, line 221:
```suggestion
	testScenario(t, "When an admin tries to update a library panel with folder only, it should change folder successfully and return correct result",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, existing.Result.ID, result.Result.ID)
+			require.Equal(t, existing.Result.OrgID, result.Result.OrgID)
+			require.Equal(t, int64(100), result.Result.FolderID)
+			require.Equal(t, existing.Result.Name, result.Result.Name)
+			require.Equal(t, existing.Result.Model["name"], result.Result.Model["name"])
+			require.Equal(t, existing.Result.Created.UTC().Format(time.RFC3339), result.Result.Created.UTC().Format(time.RFC3339))
+			require.Equal(t, existing.Result.CreatedBy, result.Result.CreatedBy)
+			require.True(t, existing.Result.Updated.UTC().Before(result.Result.Updated.UTC()))
+			require.Equal(t, existing.Result.UpdatedBy, result.Result.UpdatedBy)
```
aknuds1, line 248: Safer and more succinct to use go-cmp here.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+		})
+
+	testScenario(t, "When an admin tries to update a library with name only, it should change name successfully and return correct result",
```
aknuds1, line 251:
```suggestion
	testScenario(t, "When an admin tries to update a library panel with name only, it should change name successfully and return correct result",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, existing.Result.ID, result.Result.ID)
+			require.Equal(t, existing.Result.OrgID, result.Result.OrgID)
+			require.Equal(t, int64(1), result.Result.FolderID)
+			require.Equal(t, "New Name", result.Result.Name)
+			require.Equal(t, existing.Result.Model["name"], result.Result.Model["name"])
+			require.Equal(t, existing.Result.Created.UTC().Format(time.RFC3339), result.Result.Created.UTC().Format(time.RFC3339))
+			require.Equal(t, existing.Result.CreatedBy, result.Result.CreatedBy)
+			require.True(t, existing.Result.Updated.UTC().Before(result.Result.Updated.UTC()))
+			require.Equal(t, existing.Result.UpdatedBy, result.Result.UpdatedBy)
```
aknuds1, line 278: Safer and more succinct to use go-cmp here.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+		})
+
+	testScenario(t, "When an admin tries to update a library with model only, it should change model successfully and return correct result",
```
aknuds1, line 281:
```suggestion
	testScenario(t, "When an admin tries to update a library panel with model only, it should change model successfully and return correct result",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, existing.Result.ID, result.Result.ID)
+			require.Equal(t, existing.Result.OrgID, result.Result.OrgID)
+			require.Equal(t, int64(1), result.Result.FolderID)
+			require.Equal(t, existing.Result.Name, result.Result.Name)
+			require.Equal(t, "New Model Name", result.Result.Model["name"])
+			require.Equal(t, existing.Result.Created.UTC().Format(time.RFC3339), result.Result.Created.UTC().Format(time.RFC3339))
+			require.Equal(t, existing.Result.CreatedBy, result.Result.CreatedBy)
+			require.True(t, existing.Result.Updated.UTC().Before(result.Result.Updated.UTC()))
+			require.Equal(t, existing.Result.UpdatedBy, result.Result.UpdatedBy)
```
aknuds1, line 308: Safer and more succinct to use go-cmp here.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+		})
+
+	testScenario(t, "When an other admin tries to update a library, it should change UpdatedBy successfully and return correct result",
```
aknuds1, line 311:
```suggestion
	testScenario(t, "When another admin tries to update a library panel, it should change UpdatedBy successfully and return correct result",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,12 +178,240 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, existing.Result.ID, result.Result.ID)
+			require.Equal(t, existing.Result.OrgID, result.Result.OrgID)
+			require.Equal(t, existing.Result.FolderID, result.Result.FolderID)
+			require.Equal(t, existing.Result.Name, result.Result.Name)
+			require.Equal(t, existing.Result.Model["name"], result.Result.Model["name"])
+			require.Equal(t, existing.Result.Created.UTC().Format(time.RFC3339), result.Result.Created.UTC().Format(time.RFC3339))
+			require.Equal(t, existing.Result.CreatedBy, result.Result.CreatedBy)
+			require.True(t, existing.Result.Updated.UTC().Before(result.Result.Updated.UTC()))
+			require.Equal(t, int64(2), result.Result.UpdatedBy)
```
aknuds1, line 337: Safer and more succinct to use go-cmp here.

aknuds1 (changes requested):

pkg/services/librarypanels/database.go
hugohaggmark: Oh man, sorry I messed up 🙈 . I'll fix it in the next commit.

pkg/services/librarypanels/models.go
hugohaggmark: Ok, that makes sense. I'll look into that.

pkg/services/librarypanels/api.go
hugohaggmark: Yes, I totally agree.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Can you exclude fields to be compared in go-comp?

aknuds1: Yes, you can.

hugohaggmark: That's splendid news! Thank you.

pkg/services/librarypanels/database.go
sakjur: What you're saying sounds correct to me. Do the regular transaction semantics guarantee that you're not inserting "the same" row twice here without anything like a unique constraint?

```
   0
 /  \
A    B
 \  /
  AB
```

aknuds1: I _think_ transaction semantics would also protect you here from inserting a non-unique row, but there might be edge cases/race conditions I'm not thinking of. I feel a bit rusty when it comes to considering transaction edge cases. The UNIQUE constraint should be a much safer bet in any case.

hugohaggmark: This makes sense to me as well. Do we have a way of defining UNIQUE constraints using our current table definitions with xorm?

aknuds1: It has to be added in your migrations I think.

hugohaggmark: @aknuds1 is this what you mean by migrations? Because this doesn't seem to work :( but I'm probably doing something wrong.
```go
		Indices: []*migrator.Index{
			{Cols: []string{"org_id", "folder_id", "name"}, Type: migrator.UniqueIndex},
		}
```

aknuds1: I'm fairly certain it should be done through migrations, but don't know the exact details of how we do it in Grafana since I haven't added unique constraints there. There should be several equivalent ways of defining uniqueness in SQL as I recall. Can you find examples of it in our migrations?

hugohaggmark: Reading up on this here https://github.com/go-xorm/manual-en-US/blob/master/chapter-02/4.columns.md

aknuds1: It's probably wise to look for examples in our migrations also.

sakjur: @hugohaggmark What I've run into before: the index needs it's own IndexMigrator, it won't be picked up by the table migrator.

something like

```go
	mg.AddMigration(indexIdentifier string, migrator.NewAddIndexMigration(migrator.Table, *migrator.Index))
```

hugohaggmark: Yupp, I'm just about to try that. Thank you @sakjur

aknuds1: Thanks @sakjur for chiming in :+1:

hugohaggmark: That worked beautifully 🎉

hugohaggmark: Added `UniqueIndex` and changed both `patchLibraryPanel` and `createLibraryPanel` functions to support this instead.

pkg/services/librarypanels/models.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

pkg/services/librarypanels/database.go
hugohaggmark: Fixed

pkg/services/librarypanels/api.go

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed

torkelo (approved):
Looks good!

aknuds1 (approved):
LGTM

Note: I pushed a commit replacing references to put and update with patch, since the PR is now using PATCH semantics (and updating is associated with the PUT verb).

I also changed some tests to use go-cmp solely instead of combining it with `require`, which is more complex and error prone (the combination that is, not require on its own). A benefit is that the whole edited panel gets checked, not just a selection of fields.

hugohaggmark:
> LGTM
> 
> Note: I pushed a commit replacing references to put and update with patch, since the PR is now using PATCH semantics (and updating is associated with the PUT verb).
> 
> I also changed some tests to use go-cmp solely instead of combining it with `require`, which is more complex and error prone (the combination that is, not require on its own). A benefit is that the whole edited panel gets checked, not just a selection of fields.

Wow, thank you @aknuds1! I'll have a look at your changes and learn.

PR #29964, opened 2020-12-22
LDAP: Update use_ssl documentation

### LINKED ISSUE #22060: Docs: more info about start_tls for LDAP
Documentation clarifying `use_ssl` is required for `start_tls` to work. Or to not require both options set.

**Why is this needed**:
When setting up Grafana to use LDAP login it took me a while to work out I needed both the `use_ssl` option and the `start_tls` option set to true. I only worked it out by looking at the code: https://github.com/grafana/grafana/blob/master/pkg/services/ldap/ldap.go#L114-L123
Almost all applications I have come across separate these options as STARTTLS works differently to LDAPS.

sakjur:

Updates the documentation to make it clearer that LDAP with STARTTLS requires `use_ssl = true`.

Fixes #22060

### DIFF
--- a/conf/ldap.toml
+++ b/conf/ldap.toml
@@ -7,9 +7,9 @@
 host = "127.0.0.1"
 # Default port is 389 or 636 if use_ssl = true
 port = 389
-# Set to true if ldap server supports TLS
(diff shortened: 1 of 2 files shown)


oddlittlebird (approved):
LGTM!

PR #29968, opened 2020-12-23
Influx: Fixes issue with many queries being issued as you type in the variable query field 

### LINKED ISSUE #29965: Template Variable Query sent to Influx upon each key stroke
**What happened**: The Influx Query is parsed with each keystroke (thus leading to lots of error messages while typing in the query for the template variable).

**What you expected to happen**: The influx query will be parsed after the text box loses focus, as happened in previous Grafana versions (and with other data sources).  (It seems odd that this hasn't already been reported.  Maybe I'm overlooking something....)

**How to reproduce it (as minimally and precisely as possible)**: Attempt to add a template variable for the influx data source

query_issue

**Environment**:
- Grafana version: 7.3.1
-  Data source type & version: influxdb 1.8.3
- OS Grafana is installed on: Ubuntu 19.04
- Grafana plugins:
- Others:
- 

There is a component dedicated to editing queries inside variables public/app/plugins/datasource/influxdb/components/VariableQueryEditor.tsx and it’s dedicated to influxDB queries. Although influxdb datasource is a plugin, it’s built into Grafana now (https://grafana.com/grafana/plugins/influxdb).

It’s a little bit different from the default public/app/features/variables/editor/DefaultVariableQueryEditor.tsx query variable editor component, as it doesn’t behave like the default one - the latter does indeed query on blur | when focusing outside the query input box, not on every keystroke.

It’s possible to adjust the influxdb component so it fires the query on blur (when the input loses focus, so as we’d prefer) by changing it to, for example: 
 import React, { PureComponent } from 'react';
import InfluxDatasource from '../datasource';
import { InlineFormLabel, TextArea } from '@grafana/ui';
import { FluxQueryEditor } from './FluxQueryEditor';

interface Props {
 query: string; // before flux, it was always a string
 onChange: (query?: string) => void;
 datasource: InfluxDatasource;
}

interface State {
 value: string;
}

export default class VariableQueryEditor extends PureComponent<Props, State> {
 constructor(props: Props) {
 super(props);
 this.state = { value: props.query };
 }

 onChange = (event: React.FormEvent<HTMLTextAreaElement>) => {
 this.setState({ value: event.currentTarget.value });
 };

 onBlur = (event: React.FormEvent<HTMLTextAreaElement>) => {
 this.props.onChange(event.currentTarget.value);
 };

 onRefresh = () => {
 // noop
 };

 render() {
 let { query, datasource, onChange } = this.props;
 if (datasource.isFlux) {
 return (
 <FluxQueryEditor
 datasource={datasource}
 query= {{ 
 refId: 'A',
 query,
 }} 
 onRunQuery={this.onRefresh}
 onChange={v => onChange(v.query)}
 />
 );
 }

 return (
 <div className="gf-form-inline">
 <InlineFormLabel width={10}>Query</InlineFormLabel>
 <div className="gf-form-inline gf-form--grow kek">
 <TextArea
 value={this.state.value}
 placeholder="metric name or tags query kek"
 rows={1}
 className="gf-form-input"
 onBlur={this.onBlur}
 onChange={this.onChange}
 />
 </div>
 </div>
 );
 }
}

dprokop:
Fixes https://github.com/grafana/grafana/issues/29965

### DIFF
--- a/public/app/plugins/datasource/influxdb/components/VariableQueryEditor.tsx
+++ b/public/app/plugins/datasource/influxdb/components/VariableQueryEditor.tsx
@@ -35,11 +35,11 @@ export default class VariableQueryEditor extends PureComponent<Props> {
         <InlineFormLabel width={10}>Query</InlineFormLabel>
         <div className="gf-form-inline gf-form--grow">
           <TextArea
-            value={query || ''}
+            defaultValue={query || ''}
             placeholder="metric name or tags query"
             rows={1}
             className="gf-form-input"
-            onChange={e => onChange(e.currentTarget.value)}
+            onBlur={e => onChange(e.currentTarget.value)}
           />
         </div>
       </div>


hugohaggmark (approved):
LGTM!

dcooper-cpacket:
Thank you, Dominik.

PR #29985, opened 2020-12-27
Chore: Rewrite login auth test to standard library

ceh:

Rewrite login auth test to standard library. Part of https://github.com/grafana/grafana/issues/22844.

### DIFF
--- a/pkg/login/auth_test.go
+++ b/pkg/login/auth_test.go
@@ -25,7 +25,7 @@ func TestAuthenticateUser(t *testing.T) {
 		require.EqualError(t, err, ErrPasswordEmpty.Error())
 		assert.False(t, sc.grafanaLoginWasCalled)
 		assert.False(t, sc.ldapLoginWasCalled)
-		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
+		assert.Empty(t, sc.loginUserQuery.AuthModule)
 	})
 
 	authScenario(t, "When a user authenticates having too many login attempts", func(sc *authScenarioContext) {
@@ -41,7 +41,7 @@ func TestAuthenticateUser(t *testing.T) {
 		assert.False(t, sc.grafanaLoginWasCalled)
 		assert.False(t, sc.ldapLoginWasCalled)
 		assert.False(t, sc.saveInvalidLoginAttemptWasCalled)
-		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
+		assert.Empty(t, sc.loginUserQuery.AuthModule)
 	})
 
 	authScenario(t, "When grafana user authenticate with valid credentials", func(sc *authScenarioContext) {
@@ -90,7 +90,7 @@ func TestAuthenticateUser(t *testing.T) {
 		assert.True(t, sc.grafanaLoginWasCalled)
 		assert.True(t, sc.ldapLoginWasCalled)
 		assert.False(t, sc.saveInvalidLoginAttemptWasCalled)
-		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
+		assert.Empty(t, sc.loginUserQuery.AuthModule)
 	})
 
 	authScenario(t, "When a non-existing grafana user authenticate and invalid ldap credentials", func(sc *authScenarioContext) {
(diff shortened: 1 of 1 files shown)


pkg/login/auth_test.go
```diff
@@ -4,177 +4,157 @@ import (
...
-				So(sc.grafanaLoginWasCalled, ShouldBeFalse)
-				So(sc.ldapLoginWasCalled, ShouldBeFalse)
-				So(sc.saveInvalidLoginAttemptWasCalled, ShouldBeFalse)
-				So(sc.loginUserQuery.AuthModule, ShouldEqual, "")
-			})
-		})
+		require.EqualError(t, err, ErrPasswordEmpty.Error())
+		assert.False(t, sc.grafanaLoginWasCalled)
+		assert.False(t, sc.ldapLoginWasCalled)
+		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
```
aknuds1, line 28:
```suggestion
		assert.Empty(t, sc.loginUserQuery.AuthModule)
```

pkg/login/auth_test.go
```diff
@@ -4,177 +4,157 @@ import (
...
-				So(sc.saveInvalidLoginAttemptWasCalled, ShouldBeFalse)
-				So(sc.loginUserQuery.AuthModule, ShouldEqual, "")
-			})
-		})
+		require.EqualError(t, err, ErrTooManyLoginAttempts.Error())
+		assert.True(t, sc.loginAttemptValidationWasCalled)
+		assert.False(t, sc.grafanaLoginWasCalled)
+		assert.False(t, sc.ldapLoginWasCalled)
+		assert.False(t, sc.saveInvalidLoginAttemptWasCalled)
+		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
```
aknuds1, line 44:
```suggestion
		assert.Empty(t, sc.loginUserQuery.AuthModule)
```

pkg/login/auth_test.go
```diff
@@ -4,177 +4,157 @@ import (
...
+		mockSaveInvalidLoginAttempt(sc)
+
+		err := authenticateUser(sc.loginUserQuery)
+
+		require.EqualError(t, err, models.ErrUserNotFound.Error())
+		assert.True(t, sc.loginAttemptValidationWasCalled)
+		assert.True(t, sc.grafanaLoginWasCalled)
+		assert.True(t, sc.ldapLoginWasCalled)
+		assert.False(t, sc.saveInvalidLoginAttemptWasCalled)
+		assert.Equal(t, "", sc.loginUserQuery.AuthModule)
```
aknuds1, line 93:
```suggestion
		assert.Equal(t, sc.loginUserQuery.AuthModule)
```

aknuds1 (approved):
LGTM

aknuds1:
Please just commit my suggestions, and we can merge :)

pkg/login/auth_test.go
ceh: Yeah, I see what you mean. I thought about it too, and I actually think it makes sense to explicitly compare the value to a string literal here.

The rationale being if for example the type of `AuthModule` changes from `string` to `*string`, then the test would pass if the value is either `nil` or `""`, while we might want it to fail if it's `nil`.

What do you think?

aknuds1: I guess you're making a valid point, but if we were to follow this logic we would have to generally avoid using `assert.Empty` to test empty strings. I also don't see any reason why `LoginUserQuery` should become a pointer, generally string fields shouldn't be pointers in Go. Their absence should be represented by the empty string rather than `nil`.

aknuds1 (approved):

pkg/login/auth_test.go
ceh: > I also don't see any reason why `LoginUserQuery` should become a pointer, since generally string fields shouldn't be pointers in Go

:+1: I agree. I updated to `assert.Empty` since it's what the project seems to prefer.

(Updated my rewrite rules too for this case too. :smile: )

aknuds1:
Thanks!

PR #29986, opened 2020-12-27
Chore: Rewrite brute force login protection test to standard library

ceh:

Rewrite brute force login protection test to standard library. Part of https://github.com/grafana/grafana/issues/22844.

I chose to split up the test into `TestValidateLoginAttempts` and `TestSaveInvalidLoginAttempt` to make tests somewhat smaller and only test one function.

`TestValidateLoginAttempts` was also reworked into a table driven test because it seemed fitting when the setup and expected assertions was similar. `TestSaveInvalidLoginAttempt` on the other hand is small already and had different assertions depending on whether brute force protection was enabled or not, so a table driven test didn't seem as fitting.

What do you think?


aknuds1 (approved):

aknuds1:
Thanks!

PR #29990, opened 2020-12-28
Graph: Fixes so users can not add annotations in readonly dash

### LINKED ISSUE #28738: annotation creation/edit/deletion on non editable dashboards
As brief summary: I can add annotations but not edit or delete them on a non editable (readonly) dashboard.
I expect adding annotation beeing also disabled when the dashboard is non editable.

**What happened**:
I can create but not edit or delete annotations on a non editable (readonly) dashboard.

I would expect non editable (readonly) means annotations can not get added/inserted.

* create dashboard with graph
* set dashboard setting 'editable' = false
* save & reload
* click on the graph and add annotation
* hover to the bottom of the new annotation and try to edit/delete it

**Environment**:
- Grafana version: 7.2.0 (quite sure this is also true for latest versions)
- Data source type & version: prometheus (but should not matter)
- OS Grafana is installed on: ubuntu 20.04 (but should not matter)
- User OS & Browser: chrome (but should not matter)
- Grafana plugins: piechart, image renderer (but should not matter)

hugohaggmark:
In this PR we add a check if the dashboard is read only before displaying the Add Annotations context menu.

Enjoy the review 🎉 

Fixes #28738

### DIFF
--- a/public/app/plugins/panel/graph/graph.ts
+++ b/public/app/plugins/panel/graph/graph.ts
@@ -24,7 +24,7 @@ import ReactDOM from 'react-dom';
 import { GraphLegendProps, Legend } from './Legend/Legend';
 
 import { GraphCtrl } from './module';
-import { MenuItem, MenuItemsGroup, graphTimeFormat, graphTickFormatter, IconName } from '@grafana/ui';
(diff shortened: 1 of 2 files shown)


mckn (approved):
LGTM!

hugohaggmark:
We will revert the change introduced by this PR when #30419 is merged.

PR #29993, opened 2020-12-28
Alerting: improve alerting default datasource search when extracting alerts

### LINKED ISSUE #29262: SQLStore: customise the limit of retrieved datasources per organisation

ying-jeanne:
**What this PR does / why we need it**: the search default datasource for alerting is not very efficient, we do search query in db and then post-filtering the search result. In the PR, we do directly get from DB.

Following to Fixes #29262 


pkg/services/alerting/extractor_test.go
```diff
@@ -175,6 +175,28 @@ func TestAlertRuleExtraction(t *testing.T) {
...
+			alerts, err := extractor.GetAlerts()
+
+			Convey("Get rules without error", func() {
+				So(err, ShouldBeNil)
+			})
+
+			Convey("Use default datasource", func() {
+				condition := simplejson.NewFromAny(alerts[0].Settings.Get("conditions").MustArray()[0])
+				query := condition.Get("query")
+				So(query.Get("datasourceId").MustInt64(), ShouldEqual, 12)
```
papagian, line 196: This is more readable:
```suggestion
				So(query.Get("datasourceId").MustInt64(), ShouldEqual, defaultDs.Id)
```

papagian (approved):
I have tested and works as expected. I have a small suggestion.

pkg/models/datasource.go
```diff
@@ -207,6 +207,12 @@ type GetDataSourcesQuery struct {
 	Result []*DataSource
 }
 
+type GetDefaultDataSourceQuery struct {
+	OrgId  int64
```
papagian, line 211:
```suggestion
	OrgID  int64
```
You can rename this to conform with the Go's rule for [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

PR #29718, opened 2020-12-09
Chore: add Enterprise dependencies

AgnesToulet:
**What this PR does / why we need it**:
- Add Enterprise dependencies
- Update Auditing doc

**Special notes for your reviewer**:
- I'm not sure about the automatic updates made to the `go.sum` file. 
- Also I can split this into 2 PRs if it's better to separate doc updates from dependencides updates.

PR #29994, opened 2020-12-28
Chore: Add Enterprise dependencies

joanlopez:

- Add Enterprise dependencies
- Update Auditing docs

- Completed based on https://github.com/grafana/grafana/pull/29718
- I created a new PR because I need to merge those changes and @AgnesToulet is away.
- Note minor changes related with the way the Loki transport is defined has been introduced.


docs/sources/enterprise/auditing.md
```diff
@@ -148,11 +149,15 @@ max_file_size_mb = 256
...
+Audit logs are sent to a [Loki](/oss/loki/) service, through HTTP or gRPC.
+
+> The HTTP option for the Loki exporter is only available in Grafana Enterprise v7.4+.
 
 ```ini
 [auditing.logs.loki]
-# Set the URL for writing logs to Loki
+# Set the communication protocol to use with Loki (can be grpc or http)
+type = grpc
+# Set the URL for writing logs to Loki (format must be host:port)
```
sakjur, line 160:
```suggestion
# Set the address for writing logs to Loki (format must be host:port)
```

Minor since people call non-URLs a URL all the time, but if we document this as an address rather than as a URL, someone who knows the formal definition of a URL is potentially slightly less confused.

docs/sources/enterprise/auditing.md
```diff
@@ -14,41 +14,42 @@ Auditing allows you to track important changes to your Grafana instance. By defa
 ## Audit logs
 
 Audit logs are JSON objects representing user actions like:
+
 - Modifications ro resources such as dashboards and data sources.
 - A user failing to log in.
 
 ### Format
 
 Audit logs contain the following fields. The fields followed by **\*** are always available, the others depends on the type of action logged.
```
sakjur, line 23: 🤯 @ `**\***`.

Markdown always reminds me why I kind of like HTML or LaTeX after all 😛

sakjur (approved):

docs/sources/enterprise/auditing.md
joanlopez: Agree!! 💯

pkg/extensions/main.go
```diff
@@ -5,6 +5,7 @@ import (
 	// remove the cron (v1) dependency
 
 	_ "github.com/beevik/etree"
+	_ "github.com/cortexproject/cortex/pkg/util"
```
bboreham, line 8: Could you explain this addition, please?

joanlopez: It's being used [here](https://github.com/grafana/grafana-enterprise/blob/main/src/pkg/extensions/auditing/loggers/loki_http.go#L7-L8). 

So, the underscore import is the way to force that library to be included in Go Modules metadata (`go.mod`/`go.sum`) with no direct usages from Grafana OSS. That's basically the reason of most of the underscore imports present here.

However, going back to this specific import, I see now it's being used at [ngalert](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/api/promql_compat.go#L10). That could allow us to remove this underscore import, but I'm not sure if that's a safe operation because we could stop using that package in the OSS code base at some point, then it won't be easy to figure out back that needed.

Is that explanation / context enough? 🤔

bboreham: Yes, thank you. We are trying to break the dependencies on Cortex to avoid version clashes.

`Backoff` is moving to here: https://github.com/grafana/dskit/blob/main/backoff/backoff.go
Similarly `flagext`.

`TimeFromMillis` is so trivial it should be copied and not reused.

PR #29999, opened 2020-12-28
[dashboard api] manage error when data in dashboard table is not valid json

### LINKED ISSUE #26910: API: Add validation for dashboard JSON
**What happened**:
If the dashboard JSON is for some reason corrupted in the database (migration from other database)
the dashboard is not loaded properly (missing title etc.) but there is no other error message informing the user.

There should be at least an error informing the user about the failure to load the dashboard.

Modify the `data` column in `datasource` table and store an invalid JSON.

**Environment**:
- Grafana version: master

ying-jeanne:
**What this PR does / why we need it**: now when data from the dashboard table is not valid JSON, the response is HTTP 200 and with empty content, which is confusing for the customer.

The fix is done on GetDashboard of API level, it is because the GetDashboardQuery is used not only by retrieve dashboard API, but also by folder service and file reader, etc.

Error code is 500 in this case.
 
Fixes #26910 

### DIFF
--- a/pkg/api/dashboard.go
+++ b/pkg/api/dashboard.go
@@ -50,7 +50,6 @@ func (hs *HTTPServer) GetDashboard(c *models.ReqContext) Response {
 	slug := c.Params(":slug")
 	uid := c.Params(":uid")
 	dash, rsp := getDashboardHelper(c.OrgId, slug, 0, uid)
-
 	if rsp != nil {
 		return rsp
 	}
--- a/pkg/api/dashboard.go
+++ b/pkg/api/dashboard.go
@@ -50,8 +50,23 @@ func (hs *HTTPServer) GetDashboard(c *models.ReqContext) Response {
 	slug := c.Params(":slug")
 	uid := c.Params(":uid")
 	dash, rsp := getDashboardHelper(c.OrgId, slug, 0, uid)
+
 	if rsp != nil {
 		return rsp
+	}
+
+	// When dash contains only keys id, uid that means dashboard data is not valid and json decode failed.
+	if dash.Data != nil {
+		isEmptyData := true
+		for k := range dash.Data.MustMap() {
+			if k != "id" && k != "uid" {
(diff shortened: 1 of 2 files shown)


pkg/api/dashboard_test.go
```diff
@@ -1009,7 +1009,8 @@ func TestDashboardAPIEndpoint(t *testing.T) {
 				return nil
 			})
 			bus.AddHandler("test", func(query *models.GetDashboardQuery) error {
-				query.Result = &models.Dashboard{Id: 1, Data: &simplejson.Json{}}
+				dataValue, _ := simplejson.NewJson([]byte(`{"id": 1, "editable": true, "style": "dark"}`))
```
aknuds1, line 1012: We should make sure the JSON parsing went OK:
```suggestion
				dataValue, err := simplejson.NewJson([]byte(`{"id": 1, "editable": true, "style": "dark"}`))
				require.NoError(t, err)
```

aknuds1 (changes requested):
Generally looks good, but see suggested change please.

pkg/api/dashboard_test.go
aknuds1: Also, can we add a test for the error case?

pkg/api/dashboard.go
```diff
@@ -50,10 +50,25 @@ func (hs *HTTPServer) GetDashboard(c *models.ReqContext) Response {
 	slug := c.Params(":slug")
 	uid := c.Params(":uid")
 	dash, rsp := getDashboardHelper(c.OrgId, slug, 0, uid)
+
```
aknuds1, line 53: Why add this blank line? I think it makes sense as-is:

```suggestion
```

aknuds1 (approved):
LGTM, except I'm wondering if we could add a test case for invalid dashboard data.

aknuds1 (approved):

PR #30000, opened 2020-12-29
Dashboard: Top Share URL icon should share panel URL when on viewPanel page

### LINKED ISSUE #29641: on viewPanel page, the share icon should share the panel URL, not dashboard
**What happened**:
1. user is on a viewPanel page
2. user clicks on top-nav "share" icon, expecting it to share the panel view
3. the generated link is for the dashboard, not the panel

Maybe very experienced Grafana users will know that 1) the top-nav share link is **always** for the dashboard, even if you're on a panel view; and 2) panels have their own share link, inside the panel menu.  However, for the other 90% of people, this is confusing.

When on panelView, the top-nav share link should be for the panel, not the dashboard.

**Environment**:
- Grafana version: 7.3.4

divbhasin:
This PR is to address the issue of clicking the top Share icon when on a viewPanel page and getting the dashboard URL instead of the panel URL.

Fixes #29641 

My understanding on why the following change fixes the issue: When panel is null, we might still be on the viewPanel page, so we do not need to `delete params.viewPanel`. If panel is not null but `params.editPanel == True`, we do not display the Share icon so we do not need to address that case. Thus, I don't think we need the `else`. Please correct me if I'm wrong.


belm0:
Would you take a screenshot of the hovering over the share link on a panel view?  To me that would help confirm that the PR has the expected change.

divbhasin:
@belm0 here are the pictures.

Hovering over Share icon on a viewPanel page:
Screen Shot 2020-12-29 at 9 38 45 AM

Clicking Share icon on a viewPanel page. Observe that a viewPanel parameter is now attached in the URL to share exactly that panel as the issue requested:
Screen Shot 2020-12-29 at 9 39 00 AM

Clicking on Share on the dashboard does not attach a viewPanel and shares the dashboard instead, as expected:
Screen Shot 2020-12-29 at 9 39 24 AM

public/app/features/dashboard/components/DashNav/DashNav.tsx
```diff
@@ -151,7 +151,7 @@ class DashNav extends PureComponent<Props> {
         <ModalsController key="button-share">
           {({ showModal, hideModal }) => (
             <DashNavButton
-              tooltip="Share dashboard"
+              tooltip="Share dashboard/panel"
```
belm0, line 154: perhaps "dashboard or panel" to match wording in the popup

belm0 (approved):
Thank you

This is a sufficient improvement as-is, though it would be nice if the hover and popup were aware of the context and used either "dashboard" or "panel" as appropriate.

hugohaggmark (commented):
Thanks for opening this PR, I'm not sure this is something we're going to add. We'll try to have a definite answer after the holidays.

torkelo:
This this looks good. Just need to update the tooltip wording according to suggestion

torkelo (approved):

divbhasin:
@torkelo I have made the suggested change to the wording. However, I do not have authorization to merge.

PR #30004, opened 2020-12-29
Chore: Changes source map devtool to inline-source-map

hugohaggmark:
This PR helps to debug Grafana in the browser because of better support for breakpoints after `awaits` and the ability to add breakpoints within an expression.

**Before, not able to set breakpoint after `await` see line 29**
image

**After, able to set breakpoint after `await` see line 29 and able to set breakpoint within expressions see line 37**
image

The drawback to this is that it's a bit slower than before as shown below and described at [https://webpack.js.org/configuration/devtool](https://webpack.js.org/configuration/devtool):
|devtool | build | rebuild | production | quality|
|-- | -- | -- | -- | --|
|cheap-module-source-map | slow | slower | yes | original source (lines only)|
|inline-source-map | slowest | slowest | no | original source|

On my dev machine, the slowdown was not noticeable.

### DIFF
--- a/scripts/webpack/webpack.dev.js
+++ b/scripts/webpack/webpack.dev.js
@@ -12,7 +12,7 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
 module.exports = (env = {}) =>
   merge(common, {
-    devtool: 'cheap-module-source-map',
+    devtool: 'inline-source-map',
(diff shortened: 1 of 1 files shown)


mckn (approved):
Works great on my machine, no noticeable difference between this branch and master.

PR #30006, opened 2020-12-29
Usage Stats: Calculate concurrent users as a histogram

vtorosyan:

The metric will help to understand if users are sharing their credentials or using one account for concurrent sessions. This will help to make more informed decisions when enforcing any license limitations on credentials usage. For more information about relevance of the metric, please check out to the [issue](https://github.com/grafana/grafana-enterprise/issues/781).

For the sake of purpose, having histogram like metric is an optimum choice. PR instruments this information by querying `user_auth_token` table directly and adding the result to the existing metrics. Here is how metrics can look like:

```
stats.auth_token_per_user_le_3 -> 1
stats.auth_token_per_user_le_6 -> 2
stats.auth_token_per_user_le_9 -> 3
stats.auth_token_per_user_le_12 -> 7
stats.auth_token_per_user_le_15 -> 14
stats.auth_token_per_user_le_inf -> 30
```
The metric name is composed with the following rule: `stats.auth_token_per_user_le_"<upper inclusive bound>"`, where each bucket is cumulative. 

Fixes https://github.com/grafana/grafana-enterprise/issues/781

**Update**: Once this PR is merged, we need to rollout [this one also](https://github.com/grafana/grafana-enterprise/pull/966)

The job which sends usage reports is running once a day, in addition the usage stats are retrieved when [license token is renewed](https://github.com/grafana/grafana-enterprise/blob/master/src/pkg/extensions/licensing/update.go#L79). That said, in the best case scenario we will query the database once a day. To guard against additional performance degradation, I've added simple in-memory caching, which persists the query result for one hour, before hitting the database again. 

I've had an attempt also to run few tests to understand potential performance, but please note that this is not even a base for a good approximation.

```
Table: user_auth_token
Number of rows: 139818
``` 

This the query used for calculation/retrieval
```
SELECT
    COUNT(CASE WHEN tokens <= 3 THEN 1 END) AS bucket_le_3,
    COUNT(CASE WHEN tokens <= 6 THEN 1 END) AS bucket_le_6,
    COUNT(CASE WHEN tokens <= 9 THEN 1 END) AS bucket_le_9,
    COUNT(CASE WHEN tokens <= 12 THEN 1 END) AS bucket_le_12,
    COUNT(CASE WHEN tokens <= 15 THEN 1 END) AS bucket_le_15,
    COUNT(1) AS bucket_le_inf
FROM (select count(1) as tokens from user_auth_token group by user_id) uat;
```

And the output:

Screenshot 2020-12-30 at 16 01 35

Test results when running the query in MySQL

```
CPUs: 6, Mem: 2gb, SSD
Execution duration: 61ms (first run), 58ms, 56ms, 54ms, 53.4ms, 53.2ms

CPUs: 2, Mem: 2gb, SSD
Execution duration: 55ms (first run), 50.3ms, 47.4ms, 51ms, 52.4ms, 53.2ms

CPUs: 1, Mem: 1gb, SSD
Execution duration: 100ms (first run), 50.3ms, 47.4ms, 51ms, 52.4ms, 53.2ms
```

Test results when running the usage stats job multiple times, the measures are for code execution time ([these lines](https://github.com/grafana/grafana/compare/vt/stats_auth_histogram?expand=1#diff-4dec5e45aed3e553cf36d08338c0bbf9cf691e08e859dac0106b1092442c5172R188)).
- Grafana runs on my local (6 core, 16gb mem, SSD)
- MySQL runs on: CPUs: 1, Mem: 1gb, SSD

```
43.05871ms (first run), 8.581µs, 7.751µs, 7.714µs

38.937798ms (first run), 9.203µs, 6.704µs, 6.356µs

36.834341ms (first run), 6.209µs, 11.69µs, 7.823µs
```


pkg/services/sqlstore/stats.go
```diff
@@ -185,6 +186,58 @@ func GetSystemUserCountStats(ctx context.Context, query *models.GetSystemUserCou
...
+}
+
+func updateConcurrentUsersStats(ctx context.Context) error {
+	/*
+		Retrieves concurrent users stats as a histogram. For buckets, it rounds each token count data point up to the nearest multiple of 5
+		and then groups by that rounded value. If there is no data in a specific bucket, that row will not appear in the results.
+		Note that because ceil() is not a built-in function SQLite, rounding is used instead, which is sufficient approximation for this use case.
+	*/
+	var rawSql = `
+SELECT round(tokens/5.00 + 0.49999)*5	AS bucket_active_tokens,
```
vtorosyan, line 228: I thought distributing data across buckets with increment value of 5 is OK, but we can simply change the formula to the  `round(tokens/1.00 + 0.49999)*1` and have increments of 1.

sakjur:
Would it be possible to move to increments of 3 (I think 3 is the suggested concurrent session limit) and have incremental buckets to make sure we have the same histogram layout as Prometheus? Something like (pseudo SQL, I've omitted the grouping for the buckets as I'd have to play around with a real database to figure that out)

```sql
SELECT
       CASE WHEN tokens <= 3 THEN 1 ELSE 0 AS bucket_le_3
       CASE WHEN tokens <= 6 THEN 1 ELSE 0 AS bucket_le_6
       CASE WHEN tokens <= 9 THEN 1 ELSE 0 AS bucket_le_9
       CASE WHEN tokens <= 12 THEN 1 ELSE 0 AS bucket_le_12
       CASE WHEN tokens <= 12 THEN 1 ELSE 0 AS bucket_le_15
       1 AS bucket_le_inf
       count(*)             AS count
 FROM   (select count(1) as tokens, user_id from` + dialect.Quote("user_auth_token") + ` group by user_id) uat;`
```

Alternatively rename the histogram buckets to have both lower and upper boundaries.

**Expectations**:
- [ ] Use multiples of 3 instead of 5 for buckets.
_and either of_
- [ ] Use Prometheus style incremental buckets.
- [ ] Rename metrics to include the lower boundary.

vtorosyan:
@sakjur I've updated to cumulative histogram, and buckets increment by 3. `auth_token_per_user_le_15` is the upper bound and everything greater goes to `inf.`

Thanks for the recommendation, I think this solution is much better, with no performance difference ✅  The SQL query is more readable and no issues with SQL dialect (when `ceil()` is not available in SQLite) ✅ Predefined buckets and cumulative histogram will ensure that we are consistent and data won't accidentally be misinterpreted. ✅

aknuds1:
Having a look.

aknuds1 (changes requested):
I think the PR should be rewritten to avoid introducing globals. We are generally avoiding the usage of [global state](https://github.com/grafana/grafana/blob/master/contribute/style-guides/backend.md#globals), and we are also reducing the usage of the bus (it tends to require global state f.ex.).

I'm trying to modify the PR locally to avoid global state and bus usage, should be doable.

aknuds1:
I've made a [branch](https://github.com/grafana/grafana/tree/arve/stats_auth_histogram) with my changes, although I haven't tried to fix up the tests yet.

vtorosyan:
We have caught up with @aknuds1 in private, to summarise the conversation:
* The [SQLStore for stats](https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/stats.go) is already getting too big with mixed concerns. We would prefer to start eliminating domain specific implementations from it and keep it for low level SQL needs.
* Related to the above, we can move the changes introduced in this [PR](https://github.com/grafana/grafana/pull/30006/files#diff-484a31f371b93925d8bfa3afe2ce1775d22eb911df01aeb1f327cedbb96b4f58R189) to the [service](https://github.com/grafana/grafana/blob/master/pkg/infra/usagestats/service.go)
* Once the relevant logic is in the service, we can remove the integration through bus and call the service directly from the usage stats

vtorosyan:
@aknuds1 Would appreciate a second look, how does [this](cebefa2d4164a10492f51b2fa307724a863ba5f8) look now? Also sent a Draft PR for now in [Enterprise](https://github.com/grafana/grafana-enterprise/pull/966), as we will need to fix the usage stats usage there once this one is merged.

aknuds1:
Having another look.

pkg/infra/usagestats/service.go
```diff
@@ -70,3 +71,58 @@ func (uss *UsageStatsService) Run(ctx context.Context) error {
...
+
+const concurrentUserStatsCacheLifetime = time.Hour
+
+func (uss *UsageStatsService) GetConcurrentUsersStats(ctx context.Context, mustRefresh bool) (*ConcurrentUsersStats, error) {
+	err := uss.updateConcurrentUsersStatsIfNecessary(ctx, mustRefresh)
+	if err != nil {
+		return nil, err
+	}
+	return uss.concurrentUserStatsCache.stats, nil
+}
```
aknuds1, line 89: The `mustRefresh` parameter isn't used in practice, I think it ought to be removed, since we shouldn't have inactive code.
```suggestion
func (uss *UsageStatsService) GetConcurrentUsersStats(ctx context.Context) (*ConcurrentUsersStats, error) {
	err := uss.updateConcurrentUsersStatsIfNecessary(ctx)
	if err != nil {
		return nil, err
	}
	return uss.concurrentUserStatsCache.stats, nil
}
```

pkg/infra/usagestats/service.go
```diff
@@ -70,3 +71,58 @@ func (uss *UsageStatsService) Run(ctx context.Context) error {
...
+
+func (uss *UsageStatsService) updateConcurrentUsersStatsIfNecessary(ctx context.Context, refresh bool) error {
+	memoizationPeriod := time.Now().Add(-concurrentUserStatsCacheLifetime)
+	if refresh || uss.concurrentUserStatsCache.memoized.Before(memoizationPeriod) {
```
aknuds1, line 93:
```suggestion
func (uss *UsageStatsService) updateConcurrentUsersStatsIfNecessary(ctx context.Context) error {
	memoizationPeriod := time.Now().Add(-concurrentUserStatsCacheLifetime)
	uss.concurrentUserStatsCache.memoized.Before(memoizationPeriod) {
```

pkg/infra/usagestats/types.go
```diff
@@ -0,0 +1,10 @@
+package usagestats
+
+type ConcurrentUsersStats struct {
+	BucketLe3   int32
```
aknuds1, line 4: What does "Le" stand for, little endian? If it's an initialism, it should be written in consistent case, to be comply with standard Go style.

pkg/infra/usagestats/usage_stats_service_test.go
```diff
@@ -0,0 +1,146 @@
...
+				BucketLe9:   3,
+				BucketLe12:  4,
+				BucketLe15:  5,
+				BucketLeInf: 6,
+			}
+
+			assert.Equal(t, expectedResult, actualResult)
+
+			err = createToken(7, sqlStore)
+			assert.NoError(t, err)
```
aknuds1, line 50: `assert` would allow the test to continue after an error.
```suggestion
			require.NoError(t, err)
```

pkg/infra/usagestats/usage_stats_service_test.go
```diff
@@ -0,0 +1,146 @@
...
+				BucketLe6:   3,
+				BucketLe9:   4,
+				BucketLe12:  5,
+				BucketLe15:  6,
+				BucketLeInf: 7,
+			}
+			assert.Equal(t, expectedCachedResult, actualResult)
+
+			err = createToken(8, sqlStore)
+			assert.NoError(t, err)
```
aknuds1, line 82:
```suggestion
			require.NoError(t, err)
```

pkg/infra/usagestats/usage_stats_service_test.go
```diff
@@ -0,0 +1,146 @@
...
+	})
+
+	return nil
+}
+
+func createConcurrentTokens(t *testing.T, sqlStore *sqlstore.SQLStore) {
+	for u := 1; u <= 6; u++ {
+		for tkn := 1; tkn <= u*3; tkn++ {
+			err := createToken(u, sqlStore)
+			assert.NoError(t, err)
```
aknuds1, line 128:
```suggestion
			require.NoError(t, err)
```

pkg/infra/usagestats/usage_stats_service_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+type userAuthToken struct {
+	Id            int64
+	UserId        int64
+	AuthToken     string
+	PrevAuthToken string
+	UserAgent     string
+	ClientIp      string
```
aknuds1, line 139: In Go, [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) should have consistent case:
```suggestion
	ID            int64
	UserID        int64
	AuthToken     string
	PrevAuthToken string
	UserAgent     string
	ClientIP      string
```

pkg/infra/usagestats/usage_stats_service_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	err = sqlStore.WithDbSession(context.Background(), func(dbSession *sqlstore.DBSession) error {
+		_, err = dbSession.Insert(&userAuthToken)
+		return err
+	})
+
+	return nil
+}
+
+func createConcurrentTokens(t *testing.T, sqlStore *sqlstore.SQLStore) {
```
aknuds1, line 124:
```suggestion
func createConcurrentTokens(t *testing.T, sqlStore *sqlstore.SQLStore) {
	t.Helper()
```

pkg/infra/usagestats/usage_stats_test.go
```diff
@@ -419,12 +428,26 @@ func TestMetrics(t *testing.T) {
 			return nil
 		})
 
+		createConcurrentTokens(t, uss.SQLStore)
+
+		t.Run("Should include metrics for concurrent users", func(t *testing.T) {
+			report, err := uss.GetUsageReport(context.Background())
+			assert.NoError(t, err)
```
aknuds1, line 435:
```suggestion
			require.NoError(t, err)
```

pkg/infra/usagestats/types.go
```diff
@@ -0,0 +1,10 @@
+package usagestats
+
+type ConcurrentUsersStats struct {
```
aknuds1, line 3: Looks as if we can avoid exporting this type outside of the package:
```suggestion
type concurrentUsersStats struct {
```

aknuds1 (commented):
Looks much better, but please see comments.

pkg/infra/usagestats/usage_stats_service_test.go
vtorosyan: Yeah I tried that, but does not work unfortunately and I assumed that `sqlstore` cares about case sensitivity when inserting

```
err = sqlStore.WithDbSession(context.Background(), func(dbSession *sqlstore.DBSession) error {
		_, err = dbSession.Insert(&userAuthToken)
		return err
	})
```
The same is in [auth_token](https://github.com/grafana/grafana/blob/10ff4eecefa50c95a7555efb5806df7bfc09182d/pkg/services/auth/auth_token.go#L57) service. I ended up not using the service in the tests because of circular dependency, as far as I can remember.

vtorosyan: Ah never mind, just found out that I can do ``xorm:"user_id"`

aknuds1: Oh yeah, if it's inserted into the database, you need xorm tags on the fields. I'll see if I can find how it's done elsewhere.

vtorosyan:
@aknuds1 here is the commit which [addresses your comments](e1934689c3ab6e231e6ca6934c66cd21269f89b4)

aknuds1:
Thanks, having another look.

aknuds1 (approved):
LGTM, thanks for implementing my suggestions!

PR #30017, opened 2020-12-31
Plugins: prevent app plugin from rendering with wrong location

### LINKED ISSUE #28182: Plugins: App plugin rendered with wrong "path" before unmounting
**What happened**:
Navigating away from an app plugin causes it just before unmounting to be rendered with `path` prop that contains path of the page being navigated to, rather than some path belonging to this plugin. For example, if you navigate from a plugin to explore, it will get a `path` with value `/explore` (gif below).
Bonus bug: in case of AlertingUI, it causes it to call `onNavChange` synchronously, which makes React complain because it happens within `AppRootPage.render` method. I think calling `onNavChange` synchronously from app plugin root component should work.

Demo:
app-plugin-rerender

Logs screenshot:
logs

Should unmount app plugin without rendering it with wrong `path`. 

1. Have an app plugin, make it log it's props
2. Navigate to said app plugin
3. Navigate somewhere else
4. Observe that before unmounting, app plugin was rendered with `path` prop with a path that does not belong to this plugin

These are not big issues and have workarounds. I could try to fix them anyway, but not sure it would be considered worthwhile, due to impending reat-router refac?

**Environment**:
- Grafana version: Grafana v7.3.0-pre (4751923442)
- OS Grafana is installed on: Ubuntu 20.04
- User OS & Browser: Ubuntu 20.04, Chrome 85
- Grafana plugins: reproducible with any app plugin :) tested specifically with Alering UI (master)

domasx2:
Navigating away from an app plugin causes it just before unmounting to be rendered with path prop that contains path of the page being navigated to, rather than some path belonging to this plugin. For example, if you navigate from a plugin to explore, it will render with`path='/explore'`. 
It's unexpected and can cause an error for plugins that trigger backend requests or nav model changes based on injected location data.

It happens because `AppRootPage` gets location data injected from redux, and location in redux store is updated before angular router destroys the `ReactContainer` directive. Angular routing and location state in redux are not perfectly in sync.

This PR fixes it by making `AppRootPage` prevent from rendering  using `shouldComponentUpdate` unless provided path is for the app plugins page, seems the simplest fix. Alternatively could have `ReactContainer` add location data from angular state into props, rather than using location data in redux store.

I know that this is a band-aid and the proper fix is to remove angular routing in favor of  react router :-) will try to work on that eventually.

Fixes #28182

### DIFF
--- a/public/app/features/plugins/AppRootPage.test.tsx
+++ b/public/app/features/plugins/AppRootPage.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import React, { Component } from 'react';
 import { StoreState } from 'app/types';
 import { Provider } from 'react-redux';
@@ -8,6 +8,9 @@ import { getPluginSettings } from './PluginSettingsCache';
 import { importAppPlugin } from './plugin_loader';
 import { getMockPlugin } from './__mocks__/pluginMocks';
 import { AppPlugin, PluginType, AppRootProps, NavModelItem } from '@grafana/data';
+import { updateLocation } from 'app/core/actions';
+import { createRootReducer } from 'app/core/reducers/root';
+import { createStore } from 'redux';
 
 jest.mock('./PluginSettingsCache', () => ({
   getPluginSettings: jest.fn(),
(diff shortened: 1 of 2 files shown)


torkelo (approved):
LGTM

PR #30018, opened 2020-12-31
use sha256 checksum instead of md5

### LINKED ISSUE #29389: Security: Downloaded plugin zip files should be verified with sha256 instead of md5.
`G401 (CWE-326):  Use of weak cryptographic primitive`
 `pkg/cmd/grafana-cli/services/api_client.go:97`

The checksum comes from the grafana.com API so we would have to verify that it has a sha256 checksum of the plugin as well.

ying-jeanne:

Fixes #29389 

### DIFF
--- a/pkg/cmd/grafana-cli/services/api_client.go
+++ b/pkg/cmd/grafana-cli/services/api_client.go
@@ -109,7 +109,7 @@ func (client *GrafanaComClient) DownloadFile(pluginName string, tmpFile *os.File
 		return fmt.Errorf("failed to write to %q: %w", tmpFile.Name(), err)
 	}
 	if len(checksum) > 0 && checksum != fmt.Sprintf("%x", h.Sum(nil)) {
-		return fmt.Errorf("expected sha256 checksum does not match the downloaded archive - please contact security@grafana.com")
+		return fmt.Errorf("expected SHA256 checksum does not match the downloaded archive - please contact security@grafana.com")
 	}
 	return nil
 }
--- a/pkg/cmd/grafana-cli/services/api_client.go
+++ b/pkg/cmd/grafana-cli/services/api_client.go
@@ -2,7 +2,7 @@ package services
 
 import (
 	"bufio"
-	"crypto/md5"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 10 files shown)


ying-jeanne:
I still need to test manually the plugin which is not downloaded from github, the checksum works correctly.

pkg/cmd/grafana-cli/services/api_client.go
```diff
@@ -101,15 +101,15 @@ func (client *GrafanaComClient) DownloadFile(pluginName string, tmpFile *os.File
 	}()
 
 	w := bufio.NewWriter(tmpFile)
-	h := md5.New()
+	h := sha256.New()
 	if _, err = io.Copy(w, io.TeeReader(bodyReader, h)); err != nil {
-		return errutil.Wrap("Failed to compute MD5 checksum", err)
+		return errutil.Wrap("Failed to compute sha256 checksum", err)
```
aknuds1, line 106: Go errors should generally start with a lowercase letter:
```suggestion
		return errutil.Wrap("failed to compute SHA256 checksum", err)
```

pkg/cmd/grafana-cli/services/api_client.go
```diff
@@ -101,15 +101,15 @@ func (client *GrafanaComClient) DownloadFile(pluginName string, tmpFile *os.File
...
 	}
 	if err := w.Flush(); err != nil {
 		return fmt.Errorf("failed to write to %q: %w", tmpFile.Name(), err)
 	}
 	if len(checksum) > 0 && checksum != fmt.Sprintf("%x", h.Sum(nil)) {
-		return fmt.Errorf("expected MD5 checksum does not match the downloaded archive - please contact security@grafana.com")
+		return fmt.Errorf("expected sha256 checksum does not match the downloaded archive - please contact security@grafana.com")
```
aknuds1, line 112:
```suggestion
		return fmt.Errorf("expected SHA256 checksum does not match the downloaded archive - please contact security@grafana.com")
```

aknuds1 (commented):
Generally LGTM, but see suggestions.

aknuds1 (changes requested):
I think we can remove MD5 references in grafana-cli now, want me to push a commit for this? Should also fix the test failure.

aknuds1 (approved):
LGTM

Issue #23991, opened 2020-04-28
Elasticsearch: metric raw document sort


harrybin815:
**What would you like to be added**:

When using the ES data source, the query "metric=raw document" supports the specified field sort

**Why is this needed**:

While timestamp is a common sort, sorting by specified field should be more common

vvvprabhakar:
Fixed this issue for v6.6.x
https://github.com/grafana/grafana/pull/30029

vvvprabhakar:
Fixed this issue for v7.4.x
#32194

matyax:
Closing this one for now. If this is still valid, please follow up. Thanks!

jeremycohensolal-paylead:
Hello,
is it possible to use it with grafana 8.3.6?  or on the last (9.3.6)?  (I use OSS versions)

PR #30029, opened 2021-01-04
metric raw document aggregation with options

vvvprabhakar:
Issue: Not able to add the sort for metric type Raw document 

Added  sort options for Raw documents

What would you like to be added:

When using the ES data source, the query "metric=raw document" supports the specified field sort

Why is this needed:

While timestamp is a common sort, sorting by specified field should be more common


vvvprabhakar:
https://github.com/grafana/grafana/issues/23991

torkelo:
Closing this as we do not accept PRs for new features to old release branches

PR #30035, opened 2021-01-04
fall back to any architecture when getting plugin's checksum #30034

### LINKED ISSUE #30034: grafana-cli
**What happened**:
grafana-cli should fall back to "any" architecture when getting plugin's checksum, if there isn't a more specific architecture.

ying-jeanne:

Fixes #30034

### DIFF
--- a/pkg/cmd/grafana-cli/commands/install_command.go
+++ b/pkg/cmd/grafana-cli/commands/install_command.go
@@ -97,8 +97,11 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
 
 		// Plugins which are downloaded just as sourcecode zipball from github do not have checksum
 		if v.Arch != nil {
-			if configuredChecksum := v.Arch[osAndArchString()].SHA256; len(configuredChecksum) > 0 {
-				checksum = configuredChecksum
+			archMeta, exists := v.Arch[osAndArchString()]
+			if !exists {
+				archMeta = v.Arch["any"]
+			}
+			checksum = archMeta.SHA256
 			}
 		}
 	}
--- a/pkg/cmd/grafana-cli/commands/install_command.go
+++ b/pkg/cmd/grafana-cli/commands/install_command.go
@@ -102,7 +102,6 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
 				archMeta = v.Arch["any"]
 			}
 			checksum = archMeta.SHA256
-			}
 		}
 	}
 
--- a/pkg/cmd/grafana-cli/commands/install_command.go
+++ b/pkg/cmd/grafana-cli/commands/install_command.go
@@ -97,7 +97,9 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
 
 		// Plugins which are downloaded just as sourcecode zipball from github do not have checksum
 		if v.Arch != nil {
-			checksum = v.Arch[osAndArchString()].SHA256
+			if configuredChecksum := v.Arch[osAndArchString()].SHA256; len(configuredChecksum) > 0 {
+				checksum = configuredChecksum
+			}
 		}
 	}
 


pkg/cmd/grafana-cli/commands/install_command.go
```diff
@@ -97,7 +97,9 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
 
 		// Plugins which are downloaded just as sourcecode zipball from github do not have checksum
 		if v.Arch != nil {
-			checksum = v.Arch[osAndArchString()].SHA256
+			if configuredChecksum := v.Arch[osAndArchString()].SHA256; len(configuredChecksum) > 0 {
```
papagian, line 100: I think that you have to check whether `v.Arch` map contains a key for the specific architecture and if not then to fall back to the sha256 checksum of the `any` key.
As an example of a plugin that has architecture specific checksums you can use [Azure Data Explorer Datasource](https://grafana.com/api/plugins/grafana-azure-data-explorer-datasource) and for a plugin that has a generic checksum you can use [Zabbix plugin for Grafana](https://grafana.com/api/plugins/alexanderzobnin-zabbix-app).
So for testing it you can run:
`./bin/grafana-cli plugins install grafana-azure-data-explorer-datasource`
`./bin/grafana-cli plugins install alexanderzobnin-zabbix-app`
In the first case, your architecture specific checksum should be used (if you running the command in Windows should be the `windows-amd64`) and in the latter the `any` checksum should be used.

papagian (dismissed):

pkg/cmd/grafana-cli/commands/install_command.go
```diff
@@ -97,7 +97,9 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
...
+				checksum = configuredChecksum
```
aknuds1, line 101: This is what I had in mind, try specific arch first and eventually fall back to "any":
```suggestion
			archMeta, exists := v.Arch[osAndArchString()]
			if !exists {
				archMeta = v.Arch["any"]
			}
			checksum = archMeta.SHA256
```

aknuds1 (changes requested):
Please see my suggestion for a fallback implementation. We should also have a test.

pkg/cmd/grafana-cli/commands/install_command.go
aknuds1: Agree with @papagian, I made a suggestion in [another](https://github.com/grafana/grafana/pull/30035#discussion_r551438624) comment.

pkg/cmd/grafana-cli/commands/install_command.go
DanCech: this approach make sense, if we get to this point then grafana-com must have returned either the requested arch (if a package for that arch exists) or the "any" fallback

aknuds1: FWIW, I tested this modification before with both "linux-amd64" and "any" arch plugins, and found that it works.

ying-jeanne: What i have as impression is that we took already sha256 from any, if it is the case, the specific architecture when present could just overwrite the configuration, why we need to specifically fallback? maybe there is something i missed.

pkg/cmd/grafana-cli/commands/install_command.go
```diff
@@ -97,7 +97,12 @@ func InstallPlugin(pluginName, version string, c utils.CommandLine, client utils
 
 		// Plugins which are downloaded just as sourcecode zipball from github do not have checksum
 		if v.Arch != nil {
-			checksum = v.Arch[osAndArchString()].SHA256
+			archMeta, exists := v.Arch[osAndArchString()]
+			if !exists {
+				archMeta = v.Arch["any"]
+			}
+			checksum = archMeta.SHA256
+			}
```
aknuds1, line 105: Looks like my previous suggestion included an extra closing curly brace, which fails to compile, sorry about that.
```suggestion
			checksum = archMeta.SHA256
```

aknuds1 (approved):
Looks good to me now, but we should add an automated test to ensure this works (and stays working).

pkg/cmd/grafana-cli/commands/install_command.go
ying-jeanne: checked with both plugin, code works correctly.

aknuds1:
@ying-jeanne Are you able to add automated tests for checksum verification also?

PR #30049, opened 2021-01-04
Elasticsearch: Fix date histogram auto interval handling for alert queries

### LINKED ISSUE #22082: Alert sampling interval does not respect minimum time interval
**What happened**:
An alert isn't working as expected. We want an alert to be triggered when the number of documents matching the search criteria exceeds the specified amount. For instance, more than 3 errors in a minute.
The query used to test the alert rule has an interval of 200 ms. This is far too granular for our sampling interval. The minimum time interval for the data source is 30s. The minimum time interval of the graph is 1m. The alert does not get triggered because of this. The interval used by the graph doesn't match the one used by the alert.

The alert is triggered based on the minimum time interval. Typically this would be the maximum of the minimums. For example: the data source is 30s and the graph is 1m - the alert should use a query with a 1m interval. The interval for the alert should match the graph's interval.

* Create a data source with a minimum time interval (30s).
* Create a graph panel.
  * The metric should be count or unique count.
  * Set the graph's minimum interval to 1m.
* Create an alert for the panel.
  * The rule should be something like `WHEN max() OF query(A,5m,now) IS ABOVE 3`
  * Test the rule. Inspect the request, which shows an interval of 200ms.

This can be worked around by setting the interval of the date histogram to something other than auto. However, this is undesirable, since we can no longer change the dashboard time range and have a graph that isn't packed with sub-pixel bars.

The data source:
image

The metric:
image

The alert:
image

Request query (from Test Rule output):
```
...
"date_histogram":{"field":"@timestamp","interval":"200ms","min_doc_count":0,"extended_bounds":{"min":"1581369398357","max":"1581369698357"},"format":"epoch_millis"}}}
...
```

**Environment**:
- Grafana version: 6.6.0
- Data source type & version: Elasticsearch 7.5.2
- OS Grafana is installed on: Official Docker (Alpine) image
- User OS & Browser: Linux Mint 19 w/ Firefox 72
- Grafana plugins: grafana-worldmap-panel, grafana-clock-panel, grafana-piechart-panel, farski-blendstat-panel, jdbranham-diagram-panel, agenty-flowcharting-panel, goshposh-metaqueries-datasource, grafana-polystat-panel, snuids-radar-panel, grafana-simple-json-datasource, vonage-status-panel, flant-statusmap-panel, snuids-trafficlights-panel, fatcloud-windrose-panel
- Others: -

simianhacker:

The current backend Golang code doesn't honor the minimum interval set in the UI for alerts. This means that the data the alerts are triggering against will never match the data in the visualization. For example, if you're monitoring AWS Cloudwatch EC2 data that only reports every 5 minutes and you set the minimum interval to `5m`, the alerts will not trigger correctly because it will use the `auto` interval instead which for `5m` of data defaults to `200ms`.

This PR fixes that by setting the minimum interval to the same value in the UI.

Fixes #22082

### DIFF
--- a/pkg/tsdb/elasticsearch/time_series_query.go
+++ b/pkg/tsdb/elasticsearch/time_series_query.go
@@ -313,7 +313,7 @@ func (p *timeSeriesQueryParser) parse(tsdbQuery *tsdb.TsdbQuery) ([]*Query, erro
 			return nil, err
 		}
 		alias := model.Get("alias").MustString("")
-		interval := strconv.FormatInt(q.IntervalMs, 10) + "ms"
+		interval := model.Get("interval").MustString("0ms")
 
 		queries = append(queries, &Query{
 			TimeField:  timeField,
--- a/pkg/tsdb/elasticsearch/time_series_query.go
+++ b/pkg/tsdb/elasticsearch/time_series_query.go
@@ -313,7 +313,7 @@ func (p *timeSeriesQueryParser) parse(tsdbQuery *tsdb.TsdbQuery) ([]*Query, erro
 			return nil, err
 		}
 		alias := model.Get("alias").MustString("")
-		interval := model.Get("interval").MustString("0ms")
+		interval := model.Get("interval").MustString("")
 
 		queries = append(queries, &Query{
 			TimeField:  timeField,
--- a/pkg/tsdb/elasticsearch/time_series_query_test.go
+++ b/pkg/tsdb/elasticsearch/time_series_query_test.go
@@ -807,6 +807,7 @@ func TestTimeSeriesQueryParser(t *testing.T) {
 				"timeField": "@timestamp",
 				"query": "@metric:cpu",
 				"alias": "{{@hostname}} {{metric}}",
+        "interval": "10m",
 				"metrics": [
 					{
 						"field": "@value",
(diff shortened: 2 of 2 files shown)


marefr:
Thanks. This change seemed wrong to me at first sight. I was the one implementing the first version of backend support for elasticsearch. I wonder if something did change in Grafana alerting since last 3 years :) Cannot remember whether this has never worked as expected or not.  Could be that Grafana alerting needs an update to solve this problem for "all" data sources I don't know right now. I need to run, but will continue reviewing this tomorrow.

pkg/tsdb/elasticsearch/time_series_query.go
```diff
@@ -313,7 +313,7 @@ func (p *timeSeriesQueryParser) parse(tsdbQuery *tsdb.TsdbQuery) ([]*Query, erro
 			return nil, err
 		}
 		alias := model.Get("alias").MustString("")
-		interval := strconv.FormatInt(q.IntervalMs, 10) + "ms"
+		interval := model.Get("interval").MustString("0ms")
```
marefr, line 316: Should be empty per default for the fallback to configured data source min interval or system default works as expected.

```suggestion
		interval := model.Get("interval").MustString("")
```

marefr (changes requested):
Great catch. Added a suggestion, but other than that this looks good.

marefr (approved):
LGTM

marefr:

PR #30054, opened 2021-01-05
SeriesToRows: Fixes issue in transform so that value field is always named Value

### LINKED ISSUE #29977: Regression in v7.3.5 leads to incorrect Value column name in Table panel
**What happened**:

It appears there has been a regression in the Table panel display between 7.3.4 and 7.3.5. I believe I've tracked this regression down to this commit:
https://github.com/grafana/grafana/commit/70bf01f4f9acdf552bc3e4e2638055f666c3e3f1

In v7.3.4, a `table` panel using a Prometheus datasource and two transforms (_Series to rows_ and _Filter by name_) resulted in the following:

Note the "Value" column in the output.
This "Value" column was also selectable in the _Filter by name_ transform:
image

In other words, regardless of the data this "Value" column could be selected.

Using the [Prometheus - Demo Dashboard](https://play.grafana.org/d/000000029/prometheus-demo-dashboard?tab=transform&orgId=1&refresh=5m&editPanel=1&theme=dark) (play.grafana.net uses v7.3.6 currently):

1. Convert the visualisation to a table
2. Add a _Series to rows_ transform
3. Add a _Filter by name_ transform

Note the "Value" column now takes on the value of the first entry in the given data:
image
image

If we change the query to now change the output (i.e. change from `iowait` to something else (in this case `idle`):
```promql
sum by(mode)(irate(node_cpu_seconds_total{job="node", instance=~"$node:.*", mode!="iowait"}[5m])) > 0
```
We then see the first value showing up at the column name:
image

One can programmatically select the "Metric" and "Value" columns across both v7.3.4 and v7.3.5 using `\b(?!(?:Time)\b).*` as the pattern in the _Filter by name_ transform.

torkelo:
Fixes #29977

### DIFF
--- a/packages/grafana-data/src/transformations/transformers/seriesToRows.test.ts
+++ b/packages/grafana-data/src/transformations/transformers/seriesToRows.test.ts
@@ -134,7 +134,12 @@ describe('Series to rows', () => {
       name: 'A',
       fields: [
         { name: 'Time', type: FieldType.time, values: [100, 150, 200], config: { displayName: 'Random time' } },
-        { name: 'Temp', type: FieldType.number, values: [1, 4, 5], config: { displayName: 'Temp' } },
+        {
+          name: 'Temp',
+          type: FieldType.number,
+          values: [1, 4, 5],
+          config: { displayName: 'Temp', displayNameFromDS: 'dsName' },
+        },
       ],
     });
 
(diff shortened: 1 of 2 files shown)


zoltanbedi (approved):
lgtm

PR #30056, opened 2021-01-05
#30013 - Variables: object queries chained variables

### LINKED ISSUE #30013: Variables: Support chained variables for data sources using object queries
**What happened**:

When variable query is an object, with nested objects, then selecting a value of a variable doesn't refresh chained variables values.

Chained variables are refreshed next to a change.

CF https://github.com/grafadruid/druid-grafana/blob/master/CONTRIBUTE.md
Add a Druid datasource using URL http://broker:8082
Import dashboard.json available in repository root directory.
Comment workarounds line 35 and 42 from `src/VariableQueryEditor.tsx` (or checkout https://github.com/grafadruid/druid-grafana/commit/54c4db89fb132b7d2d27fb974da113c126e4261f )
Go to dashboard, change country, city isn't refreshed.

image

image

See https://github.com/grafana/grafana/blob/master/public/app/features/variables/utils.ts#L52

A quick workaround would be to stringify the object (JSON.stringify), otherwhise, maybe traverse the object recursively to lookup for variables.

NB: Here it's a SQL query, which could drive you wondering why such a nested structure is used; Druid uses JSON for queries (or eventually SQL), which then can produce rather complex JSON like this:

```{
  "queryType": "groupBy",
  "dataSource": "sample_datasource",
  "granularity": "day",
  "dimensions": ["country", "device"],
  "limitSpec": { "type": "default", "limit": 5000, "columns": ["country", "data_transfer"] },
  "filter": {
    "type": "and",
    "fields": [
      { "type": "selector", "dimension": "carrier", "value": "AT&T" },
      { "type": "or",
        "fields": [
          { "type": "selector", "dimension": "make", "value": "Apple" },
          { "type": "selector", "dimension": "make", "value": "Samsung" }
        ]
      }
    ]
  },
  "aggregations": [
    { "type": "longSum", "name": "total_usage", "fieldName": "user_count" },
    { "type": "doubleSum", "name": "data_transfer", "fieldName": "data_transfer" }
  ],
  "postAggregations": [
    { "type": "arithmetic",
      "name": "avg_usage",
      "fn": "/",
      "fields": [
        { "type": "fieldAccess", "fieldName": "data_transfer" },
        { "type": "fieldAccess", "fieldName": "total_usage" }
      ]
    }
  ],
  "intervals": [ "2012-01-01T00:00:00.000/2012-01-03T00:00:00.000" ],
  "having": {
    "type": "greaterThan",
    "aggregation": "total_usage",
    "value": 100
  }
}
```

CF https://druid.apache.org/docs/latest/querying/groupbyquery.html

**Environment**:
- Grafana version: 7.3.6
- Data source type & version: Druid v0.0.0
- User OS & Browser: Linux/Chrome
- Grafana plugins: Druid
- Others: Nothing :) 

I'd be happy to provide a PR in order to deliver a quick fix using `JSON.stringify` and/or maybe discuss a better fix and deliver with brand new Variables API in Grafana 8 ?

jbguerraz:
**What this PR does / why we need it**: Detect chained variables within object queries

**Which issue(s) this PR fixes**: #30013 

Fixes #30013

**Special notes for your reviewer**: Thank you :)


public/app/features/variables/utils.ts
```diff
@@ -51,7 +51,7 @@ export const getSearchFilterScopedVar = (args: {
 
 export function containsVariable(...args: any[]) {
   const variableName = args[args.length - 1];
-  args[0] = isString(args[0]) ? args[0] : Object['values'](args[0]).join(' ');
+  args[0] = isString(args[0]) ? args[0] : JSON.stringify(args[0]);
```
hugohaggmark, line 54: JSON stringify can blow up if the object is circular maybe use something like safeStringifyValye here https://github.com/grafana/grafana/blob/6dfa9b482351edc1da564e4e7917876f9b645d1c/public/app/core/utils/explore.ts#L202

hugohaggmark (commented):
Thank you for creating this PR, I think it would be great if you could add some tests for this as well.

public/app/features/variables/utils.ts
jbguerraz: Thank you @hugohaggmark ! This should be fine now.
nb: I'll add test a bit later and squash the commits then.

jbguerraz (commented):
Still tests to write :)

jbguerraz:
thx @raakasf for writing a test.
Is it fine now @hugohaggmark  ?
Thank you a lot!

hugohaggmark:
@jbguerraz could the test have more than one test case? Thank you?

kaydelaney:
@jbguerraz Hi, is there still interest in having this PR merged?

jbguerraz:
Hello @kaydelaney 
Yes, there is still this issue to get fixed properly (for now we rely on a workaround). The point here is the time we need to find to write some more tests. Can't wait to find such time!

jbguerraz:
now fixed here https://github.com/grafana/grafana/blob/main/public/app/features/variables/utils.ts

PR #30065, opened 2021-01-05
Docs: Update usage-insights.md

oddlittlebird:
Minor edits

### DIFF
--- a/docs/sources/enterprise/usage-insights.md
+++ b/docs/sources/enterprise/usage-insights.md
@@ -1,6 +1,6 @@
 +++
 title = "Usage insights"
-description = "See who is using dashboards and how often"
+description = "See how often dashboards are used"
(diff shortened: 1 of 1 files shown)


docs/sources/enterprise/usage-insights.md
```diff
@@ -1,6 +1,6 @@
 +++
-title = "Usage-insights"
-description = "Usage-insights"
+title = "Usage insights"
+description = "See who is using dashboards and how often"
```
osg-grafana, line 3: This sounds like it is identifying actual people. How about, "See how often dashboards are being used"

osg-grafana: I think the passive is warranted here because the point is to id the dashboards, not the people.

osg-grafana (approved):

PR #30076, opened 2021-01-06
Drone: Fix race conditions between Enterprise and Enterprise2

### LINKED ISSUE #30037: Fix enterprise2 package build
**What happened**:
Enterprise2 package step interferes with enterprise package step (and vice versa), due to the build pipeline tool seemingly trying to sign both enterprise and enterprise2 packages at the same time:

```
Invoking rpm with args: [--addsign dist/grafana-enterprise-7.4.0~2080pre-1.x86_64.rpm dist/grafana-enterprise2-7.4.0~2080pre-1.x86_64.rpm]
```

We need to fix so the packaging steps don't interfere with each other.

aknuds1:
Fix race conditions and file related conflicts between Enterprise and Enterprise2 when performing steps in parallel.

Most of the fixes are done in the [build pipeline tool](https://github.com/grafana/build-pipeline/pull/169).

Fixes #30037.

### DIFF
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -932,10 +932,8 @@ def get_windows_steps(edition, ver_mode, is_downstream=False):
 
     init_cmds = []
     sfx = ''
-    if edition == 'enterprise':
-        sfx = '-enterprise'
-    elif edition == 'enterprise2':
-        sfx = '-enterprise2'
+    if edition in ('enterprise', 'enterprise2'):
+        sfx = '-{}'.format(edition)
     else:
         init_cmds.extend([
             '$$ProgressPreference = "SilentlyContinue"',
--- a/scripts/lib.star
+++ b/scripts/lib.star
@@ -1,4 +1,4 @@
-grabpl_version = '0.5.30'
+grabpl_version = '0.5.31'
 build_image = 'grafana/build-container:1.3.0'
 publish_image = 'grafana/grafana-ci-deploy:1.2.7'
 grafana_docker_image = 'grafana/drone-grafana-docker:0.3.2'
@@ -37,7 +37,7 @@ def pipeline(
         'depends_on': depends_on,
     }
 
-    if edition == 'enterprise':
+    if edition in ('enterprise', 'enterprise2'):
         # We have a custom clone step for enterprise
         pipeline['clone'] = {
             'disable': True,
@@ -118,7 +118,7 @@ def init_steps(edition, platform, ver_mode, is_downstream=False, install_deps=Tr
             'rm dockerize-linux-amd64-v$${DOCKERIZE_VERSION}.tar.gz',
             'yarn install --frozen-lockfile --no-progress',
         ])
-    if edition == 'enterprise':
+    if edition in ('enterprise', 'enterprise2'):
         source_commit = ''
         if ver_mode == 'release':
             committish = '${DRONE_TAG}'
@@ -184,7 +184,7 @@ def init_steps(edition, platform, ver_mode, is_downstream=False, install_deps=Tr
     return steps
 
 def enterprise_downstream_step(edition):
-    if edition == 'enterprise':
+    if edition in ('enterprise', 'enterprise2'):
         return None
 
     return {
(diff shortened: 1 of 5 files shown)


scripts/lib.star
```diff
@@ -596,34 +626,29 @@ def package_step(edition, ver_mode, variants=None, is_downstream=False, build_ta
...
+def e2e_tests_server_step(edition, port=3001):
     sfx = ''
     package_file_pfx = ''
-    if build_tags:
-        sfx = '-' + '-'.join(build_tags)
+    if edition == 'enterprise2':
+        sfx = '-{}'.format(edition)
         package_file_pfx = 'grafana' + sfx
-    elif edition == 'enterprise':
+    elif edition in ('enterprise', 'enterprise2'):
```
AgnesToulet, line 651: You don't need to change this condition as you're already checking for 'enterprise2' above.

scripts/lib.star
```diff
@@ -795,7 +821,7 @@ def release_canary_npm_packages_step(edition):
     }
 
 def deploy_to_kubernetes_step(edition, is_downstream=False):
-    if edition != 'enterprise' or not is_downstream:
+    if edition not in ('enterprise', 'enterprise2') or not is_downstream:
```
AgnesToulet, line 824: I don't think we want this step for 'enterprise2', whatever the value of 'is_downstream' is.

scripts/lib.star
```diff
@@ -906,11 +930,13 @@ def get_windows_steps(edition, ver_mode, is_downstream=False):
...
 
+    init_cmds = []
     sfx = ''
     if edition == 'enterprise':
         sfx = '-enterprise'
-    init_cmds = []
-    if edition != 'enterprise':
+    elif edition == 'enterprise2':
+        sfx = '-enterprise2'
+    else:
```
AgnesToulet, line 939: I don't think we need this step for 'enterprise2'.
But if we do, this can be simplified with something like: 
```
if edition in ('enterprise', 'enterprise2')
    sfx = '-' + edition
```

scripts/lib.star
```diff
@@ -692,6 +717,7 @@ def copy_packages_for_docker_step():
             'package',
         ],
         'commands': [
+            'ls dist/*.tar.gz*',
```
AgnesToulet, line 720: Is this supposed to stay or was it for debugging?

AgnesToulet (commented):
Looks good! Left a few comments and thoughts.

I'm not sure about the updates to the steps we don't use for `enterprise2`. It can be confusing to see ref to `enterprise2` in these steps while we never call them with the `enterprise2`, however it can be helpful to avoid mistakes if we refactor the pipeline code.

scripts/lib.star
aknuds1: @AgnesToulet It's for debugging, but I kept it because I thought it could still be useful? I figure it doesn't hurt when you're checking a build, to see which files get copied.

scripts/lib.star
aknuds1: @AgnesToulet Will check.

scripts/lib.star
aknuds1: @AgnesToulet Thx, will check.

scripts/lib.star
aknuds1: @AgnesToulet Thanks, will check.

aknuds1:
> I'm not sure about the updates to the steps we don't use for `enterprise2`. It can be confusing to see ref to `enterprise2` in these steps while we never call them with the `enterprise2`, however it can be helpful to avoid mistakes if we refactor the pipeline code.

Yeah, I've largely taken 'enterprise2' into account to be robust. It's important to avoid future mistakes.

scripts/lib.star
aknuds1: @AgnesToulet Confirmed and fixed.

scripts/lib.star
aknuds1: @AgnesToulet Confirmed and fixed.

scripts/lib.star
aknuds1: @AgnesToulet Thanks, applied your simplification. I agree the step isn't used for enterprise2, but it''s best to be robust in case the assumption should become invalid.

AgnesToulet (approved):
LGTM!

PR #30087, opened 2021-01-06
Analytics: add data source type in data-request events

AgnesToulet:
Add the data source type information in the meta analytics `data-request` events (as it's expected in the Enterprise backend and I think it's a useful information - although not a key information)

### DIFF
--- a/packages/grafana-runtime/src/types/analytics.ts
+++ b/packages/grafana-runtime/src/types/analytics.ts
@@ -20,7 +20,8 @@ export interface DashboardInfo {
  */
 export interface DataRequestInfo extends Partial<DashboardInfo> {
   datasourceName: string;
-  datasourceId?: number;
+  datasourceId: number;
+  datasourceType: string;
   panelId?: number;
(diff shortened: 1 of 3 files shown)


jongyllen (approved):
LGTM

packages/grafana-runtime/src/types/analytics.ts
```diff
@@ -20,7 +20,8 @@ export interface DashboardInfo {
  */
 export interface DataRequestInfo extends Partial<DashboardInfo> {
   datasourceName: string;
-  datasourceId?: number;
+  datasourceId: number;
```
jongyllen, line 23: Was the change from nullable intentional? :)

AgnesToulet: Yes, when I updated the `DataSourceApi` class, I saw that the `id` field can't be null so it's more logical to have the same type here.

sakjur (approved):
:+1:

PR #30095, opened 2021-01-06
adding curl to plugin ci build container

jhunthrop:
The utility provided for security scanning requires `curl` to be present on the system for it's interactions. This PR is only adding `curl` to the base dependencies for the `grafana-plugin-ci-alpine` docker image.


jhunthrop:
closing this as the dep is already declared, but:

```
docker run -it srclosson/grafana-plugin-ci-alpine /bin/sh
/scripts # curl
/bin/sh: curl: not found
```

PR #30096, opened 2021-01-07
GraphNG: add new alpha XY Chart 

ryantxu:
Replacing #29288

The XY chart will have first class support for things where X != time

image

By seperating `Timeseries` from `XY Chart` we can explore how to best support this while keeping the timeseries chart focused on out 95% use case

Unlike the `timeseries` panel, I think this should only work with a single frame, and suggest flattening it explicitly in a transformation

There is a lot more to do to make this really ready for prime time, but I think it is worth adding soon so that it keeps the GraphNG component honest about supporting `x != time`.  This also points out the need to have hints and suggestions for this and the timeseries panel.  In particular:
* warning when x axis is not sorted
* when time exists, but not two numbers, we should suggest "view as time series" or "view as table"
* The legend does not seem to keep up with changes to the structure


public/app/plugins/panel/xychart/module.tsx
```diff
@@ -0,0 +1,225 @@
...
+      [FieldConfigProperty.Color]: {
+        settings: {
+          byValueSupport: false,
+        },
+        defaultValue: {
+          mode: FieldColorModeId.PaletteClassic,
+        },
+      },
+    },
+    useCustomConfig: builder => {
```
torkelo, line 36: would be nice to extract this and share the common definitions

ryantxu: done ;)

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -70,16 +70,16 @@ export interface LineConfig {
 /**
  * @alpha
  */
-export interface AreaConfig {
+export interface FillConfig {
```
ryantxu, line 73: This change could be its own PR... it touches a bunch of files, but the shared config felt awkward with "Area" rather than "Fill"

packages/grafana-data/src/types/plugin.ts
```diff
@@ -6,7 +6,7 @@ import { LiveChannelSupport } from './live';
 export enum PluginState {
   alpha = 'alpha', // Only included if `enable_alpha` config option is true
   beta = 'beta', // Will show a warning banner
-  stable = 'stable', // Will not show anything 
+  stable = 'stable', // Will not show anything
```
ryantxu, line 9: build fails without this change (lint)

torkelo (approved):
Think this looks mergable, and see that you fixed a master lint error as well

PR #30101, opened 2021-01-07
GraphNG: Overhaul of main test dashboard and update to null & gaps dashboard

torkelo:
* Started overhaul of the main GraphNG test dashboard to test line modes & interpolation modes 

Found a bug in the gaps logic while doing this, if the second series contains nulls they are not shown as gaps. Updated the gaps & connected test dashboard with new test cases. 

Updated main dashboard:
Screenshot from 2021-01-07 10-47-43

Updated gaps dashboard with these panels. They are identical, only difference is the order of the series. But if the series with nulls comes last the gaps are not shown. 

Screenshot from 2021-01-07 10-48-44

### DIFF
--- a/devenv/dev-dashboards/panel-graph/graph-ng-nulls.json
+++ b/devenv/dev-dashboards/panel-graph/graph-ng-nulls.json
@@ -15,7 +15,6 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 345,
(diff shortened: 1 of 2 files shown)


leeoniya:
> Updated gaps dashboard with these panels. They are identical, only difference is the order of the series. But if the series with nulls comes last the gaps are not shown.

let's see what's up here...

PR #30102, opened 2021-01-07
Grafana-UI: Enhances for TimeRangePicker and TimeRangeInput

bfmatei:
**What this PR does / why we need it**:  The PR enhances the `TimeRangePicker` and `TimeRangeInput` in many ways:
- allow hiding relative ranges using a flag called `hideQuickRanges` (applies to both components). This use-case can be found in plugins (for example).
- allow controlling `isReversed` flag in `TimeRangeInput`. This is needed as sometimes opening the picker in the right part of the screen may overflow from the viewport.
- control the height of the picker modals depending on the content that it displays.
- prevent background changing when on a narrow screen.
- improve the stories and enable all the various combinations.

**Special notes for your reviewer**: Sorry for the long PR 😿


packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -146,35 +149,51 @@ interface FormProps extends Omit<Props, 'history'> {
...
+                title="Relative time ranges"
+                options={quickOptions}
+                onSelect={props.onChange}
+                value={props.value}
+                timeZone={props.timeZone}
+              />
+              <div className={styles.spacing} />
+            </>
+          )}
+          {!hideOtherRanges && (
```
Clarity-89, line 188: I think we could use only one flag here, `hideQuickRanges` and hide both relative ranges with it. I don't think there'll be cases where we'll need only either one of them.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -93,7 +99,9 @@ export const TimeRangeInput: FC<Props> = ({
             onChangeTimeZone={onChangeTimeZone || noop}
             className={styles.content}
             hideTimeZone={hideTimeZone}
-            isReversed
+            isReversed={isReversed}
```
Clarity-89, line 102: Just wondering why is there a need to expose this flag?

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.tsx
```diff
@@ -40,6 +43,9 @@ export const TimeRangeInput: FC<Props> = ({
   hideTimeZone = true,
   timeZone = 'browser',
   placeholder = 'Select time range',
+  isReversed = false,
```
Clarity-89, line 46: Just wondering why is there a need to expose this flag?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
bfmatei: Initially I added one single flag but I thought someone might need more granularity on it. I'll switch to one flag only

bfmatei: Btw, I saw your other comment regarding `isReversed`, I responded to it but the comment disappeared completely. Should I do that one?

bfmatei: Oh, nvm. I found a little tiny edge case that I have to fix. Will be back with an update soon

Clarity-89: > Btw, I saw your other comment regarding `isReversed`, I responded to it but the comment disappeared completely. Should I do that one?

That was my bad, I thought the reversed prop is being exposed on the picker but it makes sense to have it being exposed on `TimeRangeInput`.

torkelo (commented):
Looks good 
* The height looks off now that the quick ranges are no longer there
* The background should not change when in narrow (mobile) mode, but do not think that is caused by this PR but would be good to fix at the same time.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
bfmatei: Fixed.

bfmatei:
@Clarity-89 @torkelo  - the PR has been updated with multiple improvements. As there are a lot of edge-cases to be observed (combinations of Wide/narrow screen, with/without ranges, with/without reverse flag, relative/absolute initial values), I suggest looking over storybook when you have the appropriate time.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.story.tsx
```diff
@@ -1,67 +1,122 @@
...
+              timeZone,
+            });
+          }}
+        />
+      );
+    }}
+  </UseState>
+);
+
+export const relative: Story<TimeRangeInputProps> = props => {
```
Clarity-89, line 80:
```suggestion
export const Relative: Story<TimeRangeInputProps> = props => {
```
We're moving to a format where story names are uppercased, altho there are still some stories where we haven't made these changes.

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.story.tsx
```diff
@@ -1,67 +1,122 @@
...
-              action('onChange fired')(timeRange);
-              updateValue(timeRange);
-            }}
-          />
-        );
-      }}
-    </UseState>
+relative.args = args;
+
+export const absolute: Story<TimeRangeInputProps> = props => {
```
Clarity-89, line 102:
```suggestion
export const Absolute: Story<TimeRangeInputProps> = props => {
```

packages/grafana-ui/src/components/TimePicker/TimeRangePicker.story.tsx
```diff
@@ -41,8 +73,64 @@ export const basic = () => {
...
+          >
+            Clear history
+          </Button>
+        </>
+      );
+    }}
+  </UseState>
+);
+
+export const relative: Story<TimeRangePickerProps> = props => {
```
Clarity-89, line 92:
```suggestion
export const Relative: Story<TimeRangePickerProps> = props => {
```

packages/grafana-ui/src/components/TimePicker/TimeRangePicker.story.tsx
```diff
@@ -41,8 +73,64 @@ export const basic = () => {
...
+      timeZone: DefaultTimeZone,
+      history: [],
+    },
+    props
   );
 };
+
+relative.args = args;
+
+export const absolute: Story<TimeRangePickerProps> = props => {
```
Clarity-89, line 115:
```suggestion
export const Absolute: Story<TimeRangePickerProps> = props => {
```

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
```diff
@@ -51,6 +51,40 @@ describe('TimePickerContent', () => {
...
+        onChangeTimeZone={() => {}}
+        onChange={value => {}}
+        timeZone="utc"
+        value={value}
+        isFullscreen={true}
+        hideQuickRanges={true}
+      />
+    );
+
+    expect(wrapper).toMatchSnapshot();
```
Clarity-89, line 69: We're moving away from snapshot tests and Ezyme to using React Testing Library.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -11,58 +11,60 @@ import { TimeRangeForm } from './TimeRangeForm';
 import { TimeRangeList } from './TimeRangeList';
 import { TimePickerFooter } from './TimePickerFooter';
 
-const getStyles = stylesFactory((theme: GrafanaTheme, isReversed) => {
-  const containerBorder = theme.isDark ? theme.palette.dark9 : theme.palette.gray5;
+const getStyles = stylesFactory(
+  (theme: GrafanaTheme, isFullscreen, isReversed, hideQuickRanges, showHistory, isHistoryEmpty) => {
```
Clarity-89, line 15: This seems like too many params for a style function and we're using most of them only once on line 36, so maybe they could be combined into one param outside of this function?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -147,35 +149,50 @@ interface FormProps extends Omit<Props, 'history'> {
...
+    otherOptions = [],
+    isFullscreen,
+    isReversed,
+    hideQuickRanges,
+    showHistory,
+    history,
+  } = props;
   const theme = useTheme();
-  const styles = getStyles(theme, props.isReversed);
+  const isHistoryEmpty = (history?.length ?? 0) === 0;
```
Clarity-89, line 162:
```suggestion
  const isHistoryEmpty = !!history?.length;
```

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -192,43 +209,44 @@ export const TimePickerContent: React.FC<Props> = props => {
...
   if (!props.visible) {
     return null;
   }
 
+  const theme = useTheme();
+  const { value, hideQuickRanges, onChange, timeZone, historyOptions, showHistory } = props;
+  const isAbsolute = isDateTime(value.raw.from) || isDateTime(value.raw.to);
+  const [collapsedFlag, setCollapsedFlag] = useState(!isAbsolute);
+  const collapsed = hideQuickRanges ? false : collapsedFlag;
+  const styles = getNarrowScreenStyles(theme);
```
Clarity-89, line 221: Doesn't this break rules of hooks, since now they will be evaluated conditionally depending on the `visible` prop?

Clarity-89 (commented):
Good job, left a few comments to consider. 
General remarks:
1. Please sing the CLA, otherwise your PR can't be merged. 
2. Not a huge fan of the added complexity but at the same time I see the reasons for it.
3. It works well, except the positioning of the dropdown in the **reversed input picker** is incorrect:
Screenshot 2021-01-08 at 9 42 37
It should start from the beginning of the input, otherwise the absolute range calendar doesn't fit the screen. 
Screenshot 2021-01-08 at 9 50 53
I think this change could've been introduced in another PR but could be fixed here nevertheless.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
bfmatei: Now thinking about it, indeed it is breaking the rules of hooks. I'll fix it ASAP.

I was a little intrigued why the linter did not pop this and I saw that `.eslintrc` in `grafana-ui` is overriding the base eslint config, resulting in errors not being shown. Is this intentional?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
bfmatei: I thought about this and I was wondering if the computations should be performed in the component or the `getStyles` method. My rationale was to avoid increasing complexity in the component itself.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
Clarity-89: > I was a little intrigued why the linter did not pop this and I saw that `.eslintrc` in `grafana-ui` is overriding the base eslint config, resulting in errors not being shown. Is this intentional?

Hmm, not sure about those configs, maybe @jackw knows more.

zoltanbedi: I know. It is not enabled for grafan-ui yet as there are some GraphNG related errors that needs to be fixed first. We can enable it once those are fixed. You can track the progress here https://github.com/grafana/grafana/issues/27391

torkelo:
> allow controlling isReversed flag in TimeRangeInput. This is needed as sometimes opening the picker in the right part of the screen may overflow from the viewport.

This feels like the wrong way to solve this, the component should be able to detect this automatically so that it is not positioned outside the viewport

bfmatei:
> > allow controlling isReversed flag in TimeRangeInput. This is needed as sometimes opening the picker in the right part of the screen may overflow from the viewport.
> 
> This feels like the wrong way to solve this, the component should be able to detect this automatically so that it is not positioned outside the viewport

I agree. I was a little bit surprised it doesn't do it by default and I assumed it was for a particular reason.

bfmatei:
The PR was updated with a mechanism that decides where to open the modal based on the position of the input. Please have a look over and if you're OK with it, I will proceed with writing the necessary tests for it.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -1,181 +1,191 @@
...
+export const TimePickerContent: React.FC<Props> = props => {
+  const theme = useTheme();
+  const isFullscreen = useMedia(`(min-width: ${theme.breakpoints.lg})`);
+  const { width: windowWidth, height: windowHeight } = useWindowSize();
+  const bounding = useMemo(() => props.handleRef.current?.getBoundingClientRect(), [
+    props.handleRef,
+    windowWidth,
+    windowHeight,
+  ]);
+  const fits = checkIfFits(windowWidth, windowHeight, bounding);
```
Clarity-89, line 71: Should this call also be memoized?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.tsx
```diff
@@ -1,181 +1,191 @@
...
+      left: true,
+      right: true,
+      bottom: true,
+      top: true,
+    };
+  }
 
-const getEmptyListStyles = stylesFactory((theme: GrafanaTheme) => {
-  const formBackground = theme.isDark ? theme.palette.gray15 : theme.palette.gray98;
+  const { bodyTallHeight, containerWideWidth, calendarWidth } = stylesVariables;
```
Clarity-89, line 32: This makes this function tightly coupled with `stylesVariables` global variable, should it be passed as an argument instead?

Clarity-89 (commented):
I like the idea of adjustable range picker but at the same time I'm wondering if the users will get confused when seeing the picker's alignment change? Also not a huge fan of the added complexity to this already complex component. 

Noticed a few UI issues: 
TimeRangeInput is only used in Enterprise's reporting feature so far, afaik and there it looks a bit weird when the dropdown is right aligned. Also for some reason it opens to the top while there's still space at the bottom.
Screenshot 2021-01-11 at 12 31 27
Another issue is that the auto positioning doesn't take into account some UI elements like left side navigation bar: 
Screenshot 2021-01-11 at 12 39 33

bfmatei:
I undo'ed the modifications for auto-positioning the dropdown of the picker as there were edge cases where it is more complicated than expected. For the time being it's not feasible to have the complexity as there are very few occurrences of the picker.

The pointed bugs have been fixed, I also tested this in enterprise Grafana just to be sure no other issues appeared.

Please have a look over.

Clarity-89 (commented):
> I undo'ed the modifications for auto-positioning the dropdown of the picker as there were edge cases where it is more complicated than expected. For the time being it's not feasible to have the complexity as there are very few occurrences of the picker.

Yeah, it's a challenge better left for another PR. 

In general I think this looks good, great job! Left a few minor comments, also the tests need to be fixed.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker.story.tsx
```diff
@@ -1,35 +1,67 @@
...
-import { TimeFragment, dateTime } from '@grafana/data';
+import { dateTime, TimeRange, DefaultTimeZone, TimeZone, isDateTime } from '@grafana/data';
+import { TimeRangePickerProps } from './TimeRangePicker';
 
 export default {
   title: 'Pickers and Editors/TimePickers/TimeRangePicker',
   component: TimeRangePicker,
   decorators: [withCenteredStory],
+  argTypes: {
+    hideText: {
```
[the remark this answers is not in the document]
bfmatei: Fixed!

packages/grafana-ui/src/components/TimePicker/TimeRangeInput.story.tsx
```diff
@@ -1,67 +1,122 @@
...
+import { TimeRangeInputProps } from './TimeRangeInput';
 import mdx from './TimeRangeInput.mdx';
 
 export default {
   title: 'Pickers and Editors/TimePickers/TimeRangeInput',
   component: TimeRangeInput,
   decorators: [withCenteredStory],
+  argTypes: {
+    hideTimeZone: {
+      control: 'boolean',
```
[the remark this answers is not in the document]
bfmatei: Fixed!

bfmatei:
The PR has been updated with the changes to StoryBook, fixed tests but also new tests written with RTL.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
```diff
@@ -1,11 +1,12 @@
+import { render, screen } from '@testing-library/react';
 import React from 'react';
 import { shallow } from 'enzyme';
```
Clarity-89, line 3: Could we also refactor other tests to use RTL? Seems weird to use two different libs in the same file.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
```diff
@@ -51,9 +52,121 @@ describe('TimePickerContent', () => {
     );
     expect(wrapper).toMatchSnapshot();
```
Clarity-89, line 53: Also could we get rid of snapshots here?

Clarity-89 (approved):
Looks good! Just think it would be good to get rid of enzyme/snapshot tests completely and replace them with RTL tests.

bfmatei:
Replaced enzyme tests with RTL tests.

@Clarity-89 - one last look over the tests?

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
```diff
@@ -1,59 +1,174 @@
...
+  const absoluteValue = createAbsoluteTimeRange('2019-12-17T07:48:27.433Z', '2019-12-18T07:49:27.433Z');
+  const relativeValue = createRelativeTimeRange();
+  const history = [
+    createAbsoluteTimeRange('2019-12-17T07:48:27.433Z', '2019-12-17T07:49:27.433Z'),
+    createAbsoluteTimeRange('2019-10-18T07:50:27.433Z', '2019-10-18T07:51:27.433Z'),
+  ];
+
+  describe('Wide Screen', () => {
+    it('renders with history', () => {
+      const container = renderComponent({ value: absoluteValue, history });
```
Clarity-89, line 16: Is there a specific reason to use `container` here? The preferred way is to use `screen` for queries instead. 
```
renderComponent({ value: absoluteValue, history });
expect(screen..queryByText(/recently used absolute ranges/i)).toBeInTheDocument();
```

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
```diff
@@ -1,59 +1,174 @@
...
+  hideQuickRanges = false,
+  hideTimeZone = false,
+}: {
+  value: TimeRange;
+  isFullscreen?: boolean;
+  showHistory?: boolean;
+  history?: TimeRange[];
+  hideQuickRanges?: boolean;
+  hideTimeZone?: boolean;
+}): RenderResult {
```
Clarity-89, line 144:
```suggestion
function renderComponent({
  value,
  isFullscreen = true,
  showHistory = true,
  history = [],
  hideQuickRanges = false,
  hideTimeZone = false,
}: PropsWithScreenSize): RenderResult {
```

We can use `PropsWithScreenSize` here (which needs to be exported from `TimePickerContent`) to avoid redeclaring the types.

Clarity-89 (approved):
Great job! Looks good, just two minor comments, when those are addressed I think it's ready to be merged 🎉

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
bfmatei: I used screen initially but I thought container would make more sense for some reason. Fixed.

packages/grafana-ui/src/components/TimePicker/TimeRangePicker/TimePickerContent.test.tsx
bfmatei: Fixed. The workaround with `Pick` and `Partial` is needed as using just `Partial` will throw error regarding `value`

bfmatei: Also, using would require declaring each mandatory prop everytime.

PR #30106, opened 2021-01-07
CloudWatch:  Make sure stats grow horizontally and not vertically in the Query Editor

### LINKED ISSUE #30105: CloudWatch: Stats are being displayed vertically in the query editor
In PR #27648 the CloudWatch Stats components was refactored so that it displays Stats vertically instead of horizontally. That doesn't look so well - all other fields in the editor grow horizontally.

sunker:
fixes #30105


peterholmberg (approved):
LGTM

PR #30003, opened 2020-12-29
Elastic: Fixes so templating queries work

hugohaggmark:
**What this PR does / why we need it**:
As described in #30002 this PR fixes some bugs introduced in the Elastic data source. This PR also removes `DefaultTimeRange` and introduces `getDefaultTimeRange` function instead, this way we always get a valid default TimeRange. 

Enjoy the review 🎉 

**Which issue(s) this PR fixes**:
Fixes #30002

**Special notes for your reviewer**:

PR #30109, opened 2021-01-07
Chore: Fix intermittent time-related test failure in explore datasource instance update

Elfo404:
#30003 Introduced a `getDefaultTimeRange` function that generates a TimeRange object out of the current time.
This lead to an intermittent test failure because we were comparing 2 different TimeRange objects that may be created at 2 different times.

Also made a small change in `getDefaultTimeRange` to ensure that `from` is calculated starting from the `to` property.


torkelo (approved):

hugohaggmark (commented):
Great catch! LGTM!

PR #27449, opened 2020-09-08
Chore: Convert CloudWatch metrics to data frames

aknuds1:
**What this PR does / why we need it**:
Convert CloudWatch metrics queries to return data frames.

Implemented for both back-end and front-end.

PR #30111, opened 2021-01-07
Add query row meta to dataframes in DataQueryResponse

sunker:
When the CloudWatch datasource was refactored to use dataframes in #27449, QueryResult.Meta https://github.com/grafana/grafana/blob/42d569574b5d9f0aea5838592ba53261dc6f102c/pkg/tsdb/models.go#L41 was lost. The meta data is used in the CloudWatch query editor query preview, which is now broken. 

This PR adds meta data to each dataframe within a given query row.

### DIFF
--- a/packages/grafana-runtime/src/utils/queryResponse.ts
+++ b/packages/grafana-runtime/src/utils/queryResponse.ts
@@ -19,6 +19,7 @@ interface DataResponse {
   dataframes?: string[];
   series?: TimeSeries[];
   tables?: TableData[];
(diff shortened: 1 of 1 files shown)


aknuds1 (commented):
I was never involved with the frontend part, I guess @ryantxu is a better reviewer than me at least for this.

sunker:
Hmm I guess a better solution would be to use dataFrame.Meta instead. I'll rewrite the backend.

Issue #16539, opened 2019-04-12
MySQL datasource allows to customize connection string


silencev:
**What would you like to be added**:
In source code [here](https://github.com/grafana/grafana/blob/b9cc476eb7aa57ed6714654e6036c73c9d105919/pkg/services/sqlstore/sqlstore.go#L229) and [here] I found the connection string parameters are hardcoded.

In MySQL Go driver, it supports following parameters
https://github.com/go-sql-driver/mysql#allowcleartextpasswords

**Why is this needed**:
In my use case, I need to add **allowCleartextPasswords=true** in connection string.

It would be convenient for users to has its own parameters.

alexanderilyin:
This is required when using MySQL 8.0 as Data Source.

soufianez75:
Hello I have the same error it's possible to tell me how you have corrected the data source ?

marefr:
@silencev @alexanderilyin are you referring to using MySQL as Grafana config database or querying MySQL using a Grafana data source for plotting data in a dashboard? Or both? Based on referenced code I'm not sure.

alexanderilyin:
In my case I was adding a new MySQL 8 data source for plotting some data on graphs. There was a problem with no default timezone set in MySQL DB and due to some other restrictions, my queries from the panel were failing because of timezone problem. I don't recall how I worked it around but probably found a hacky way. If there was a way to add additional params to the connection string it would be easy to deal with but it's not the option due to how its hardcoded in sources when I checked.

junneyang:
+1 
need connection string parameters:
allowCleartextPasswords=1

PR #30112, opened 2021-01-07
Update mysql.go

junneyang:
when add mysql datasource, it says:
this user requires clear text authentication. If you still want to use it, please add 'allowCleartextPasswords=1' to your DSN
 i suggest add the allowCleartextPasswords=1 as default uri configuration...

### DIFF
--- a/pkg/tsdb/mysql/mysql.go
+++ b/pkg/tsdb/mysql/mysql.go
@@ -36,7 +36,7 @@ func newMysqlQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndpoin
 		protocol = "unix"
 	}
 
-	cnnstr := fmt.Sprintf("%s:%s@%s(%s)/%s?collation=utf8mb4_unicode_ci&parseTime=true&loc=UTC&allowNativePasswords=true",
+	cnnstr := fmt.Sprintf("%s:%s@%s(%s)/%s?collation=utf8mb4_unicode_ci&parseTime=true&loc=UTC&allowNativePasswords=true&allowCleartextPasswords=1",
 		characterEscape(datasource.User, ":"),
 		datasource.DecryptedPassword(),
 		protocol,
--- a/pkg/tsdb/mysql/mysql.go
+++ b/pkg/tsdb/mysql/mysql.go
@@ -36,7 +36,7 @@ func newMysqlQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndpoin
 		protocol = "unix"
 	}
 
-	cnnstr := fmt.Sprintf("%s:%s@%s(%s)/%s?collation=utf8mb4_unicode_ci&parseTime=true&loc=UTC&allowNativePasswords=true&allowCleartextPasswords=1",
+	cnnstr := fmt.Sprintf("%s:%s@%s(%s)/%s?collation=utf8mb4_unicode_ci&parseTime=true&loc=UTC&allowNativePasswords=true",
 		characterEscape(datasource.User, ":"),
 		datasource.DecryptedPassword(),
 		protocol,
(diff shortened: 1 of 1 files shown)


aknuds1 (changes requested):
I'm not sure what causes the need for MySQL clear text passwords, but I don't think it should be hard coded. If we do decide to support this mode, it should most likely be configurable.

aknuds1:
@junneyang Do you know why your MySQL server requires clear text passwords?

junneyang:
> @junneyang Do you know why your MySQL server requires clear text passwords?

because we use a self-signed certificate to connect mysql with SSL:
image

when we use navicate or other mysql client to connect with ssl, it works, but with grafana it fails


@aknuds1
make the option configurable is reasonable

junneyang:
please also ref this issue: #16539
it's the same case for us(allowCleartextPasswords=1)

aknuds1:
> please also ref this issue: #16539
> it's the same case for us(allowCleartextPasswords=1)

Thanks.

aknuds1:
You also need to sign the CLA for us to consider the PR.

junneyang:
> You also need to sign the CLA for us to consider the PR.

i have agreed to the CLA for grafana/grafana just now, thanks

aknuds1:
Do you know why clear text passwords is necessary when you use a self-signed certificate? I didn't get the connection.

junneyang:
> Do you know why clear text passwords is necessary when you use a self-signed certificate? I didn't get the connection between the two.

i do not know the reason a lot, but my operation is:
1、use tsl client auth to connect:
image

2、it says x509: certificate signed by unknown authority
image

3、so i check Skip TLS Verify：
image

4、it says this user requires clear text authentication. If you still want to use it, please add 'allowCleartextPasswords=1' to your DSN

image


so, i do not know how to solve this problem...

aknuds1:
I'm not sure if the need for clear text passwords is necessarily connected to your certificate situation.

sakjur:
Taking a look at this and [client-side cleartext pluggable authentication in the MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/cleartext-pluggable-authentication.html), I think the reasonable thing would be to add this as an option, since if we were to support this by default or as an automatic downgrade, we'd increase the risk of someone getting access to the password by a MITM attack.

There seem to be strong enough reasons (such as the LDAP/PAM cases listed in the docs I linked) to support it, but the combination of increased risks and this not being the typical setup makes a case for not having this on by default.

aknuds1:
@junneyang Can you modify the PR to make this a configurable option instead of hardcoding it?

raymondchen625:
I don't get why there is a connection between clear-text password and TLS either. My guess is the implementation of [MySQL Pluggable Authentication](https://dev.mysql.com/doc/mysql-security-excerpt/5.6/en/pluggable-authentication.html) doesn't support chained plugins. When TLS plugin is used, it assumes this is the only auth plugin and whatever comes out of it must be a clear-text password. That could also explain why it works when TLS is not enabled because the default plugin is mysql_native_password or mysql_old_password. It works fine without allowCleartextPasswords=1.

junneyang:
> I don't get why there is a connection between clear-text password and TLS either. My guess is the implementation of [MySQL Pluggable Authentication](https://dev.mysql.com/doc/mysql-security-excerpt/5.6/en/pluggable-authentication.html) doesn't support chained plugins. When TLS plugin is used, it assumes this is the only auth plugin and whatever comes out of it must be a clear-text password. That could also explain why it works when TLS is not enabled because the default plugin is mysql_native_password or mysql_old_password. It works fine without allowCleartextPasswords=1.

yes, When TLS plugin is used, It must be clear-text password.

aknuds1 (changes requested):
The allowClearTextPasswords setting should be configurable via the data source's UI, not dependent on whether TLS/SSL is used.

ying-jeanne (commented):
hi @junneyang it seems e2e test is failing for the build, here is the guideline https://github.com/grafana/grafana/blob/master/contribute/developer-guide.md, section `Run end-to-end tests` you could find how to run in local e2e testing.

aknuds1:
Closing this PR as stale, since there hasn't been activity since January.

enginecan:
I've created https://github.com/grafana/grafana/pull/63232 in an effort to solve this.

PR #30117, opened 2021-01-07
Alerting NG: Fix alert definition routine stop

papagian:
Fix alert definition routine stop when an alert definition is deleted.
It used to work only if the routine that consumes the message is the routine of the deleted alert definition.
With this fix there is a dedicated stop channel for each alert definition


kylebrandt (approved):
:+1:

PR #30123, opened 2021-01-07
GraphNG: rename "graph3" to "timeseries" panel

ryantxu:
This moves the new uPlot based graph3 into a panel called "timeseries" -- this will become a "beta" plugin that can live next to graph for 7.4

We should also have a "XY Chart" that is not tied to a timeseries format -- it may share many/most graph settings, however it will aim to to focus on X != time

TODO?... but not in this PR
- Sorting? Eventually this should be listed first, like "graph" do we want it 2nd now?  
- Searching? If listed first, maybe it is not an issue... but should searching "graph" show timeseries?
- New Icon?  Perhaps one that does not make the bars option so important -- we will also add a new "Bar char" panel that should have clear bars in the icon  

image
image


leeoniya:
i am 100% behind de-emphasizing bars and removing them from the timeseries icon. we will soon have additional bars-esque icons for new panels like histogram, bars, (already bar guage).

i think a 2-trendline icon (linear + stepped), with visible points, with maybe an area fill for the bottom trend would work well. the stepped variant would signal that there are additional drawStyles and also kind-of looks like bars, but not really.

torkelo (approved):
The only thing that feels a bit bad about the id/panel name is that is's the name of a data form not a visualization, so not very intuitive when looking at the json for example what visualization this is using. 

Alt names
* time-graph
* time-series-graph
* graph-time
* graph-over-time

Thoughts?

leeoniya:
i generally associate the term "graph" with visible `-x` quadrants, and function plotting (do an image search for "graph").

some more:

- time series plot
- trend plot / chart / graph
- trendlines
- time trend
- temporal trend

wikipedia has some other options: https://en.m.wikipedia.org/wiki/Time_series

- run chart
- temporal line chart

i like "trend", as it universally implies a temporal x. on the down side, it's frequently associated with a regression line (like r^2 or moving avg). timeseries is more common and has a really strong association of the chart type viz for this term. try an image search for `time series` and you get prettty much just a single chart type.

ryantxu:
I think timeseries makes sense -- This really is the visualization that is most tied to the concept of time-vs-value.  One could equally argue that 'table' is as much a format as it is a visualization.

In fact before dataframes, graph==timeseries and table==table ;)

PR #30124, opened 2021-01-07
GraphNG: remove graph2 panel (keep the parts needed for explore)

ryantxu:
This removes the react/flot panel -- and moves the options needed for explore into the explore package


torkelo (approved):

zoltanbedi (approved):
We can also remove the remaining getGraphSeriesModel once the Exemplars branch is merged.

PR #30134, opened 2021-01-08
Explore: Make getFieldLinksForExplore more reusable

aocenas:
For service map work and explore it make sense to reuse the same util function to create links to explore but in context of dashboards the signature needs to change so splitOpenFn function is optional. Also allows for passing dataFrame so full interpolation of other column values can take place aligning it bit more with how data links work in dashboards.

### DIFF
--- a/public/app/features/explore/utils/links.ts
+++ b/public/app/features/explore/utils/links.ts
@@ -36,6 +36,7 @@ export const getFieldLinksForExplore = (options: {
     text: 'Raw value',
   };
 
+  // If we have a dataFrame we can allow referencing other columns ant their values in the interpolation.
   if (dataFrame) {
     scopedVars['__data'] = {
       value: {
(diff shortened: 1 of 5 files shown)


public/app/features/explore/utils/links.ts
```diff
@@ -10,20 +19,37 @@ import { splitOpen } from '../state/main';
...
+  const { field, vars, splitOpenFn, range, rowIndex, dataFrame } = options;
+  const scopedVars: any = { ...(vars || {}) };
   scopedVars['__value'] = {
     value: {
       raw: field.values.get(rowIndex),
     },
     text: 'Raw value',
   };
 
+  // If we have a dataFrame we can allow referencing other columns ant their values in the interpolation.
```
zoltanbedi, line 39:
```suggestion
  // If we have a dataFrame we can allow referencing other columns and their values in the interpolation.
```

zoltanbedi (approved):
Nice! 🦖

PR #30141, opened 2021-01-08
Grafana-ui: Allow context menu items to be open in new tab

aocenas:
Sometimes we need both url and onClick work in context menu. This allows to open the link with ctrl/meta + click as would be expected.


aocenas:
This pattern is actually present in multiple places I wonder if it would make sense to have a custom Link component for this behavior so it is harder to forget.

torkelo (approved):
Looks good

dprokop (approved):

PR #30146, opened 2021-01-08
Add option in database config to skip migrations for faster startup.

tomwilkie:
Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>

Adds an option to the DB config to skip running the migrations, to speed up Grafana start up.


torkelo (approved):
Looks good to me!

bergquist (approved):
LGTM

Considering how few will end up using this setting I dont think we have to document it.

PR #30151, opened 2021-01-08
Pagination: Improve pagination for large number of pages

### LINKED ISSUE #29260: Pagination: Improve generic pagination component for grafana/ui
Develop a wrapper around https://github.com/react-component/pagination , similar to how 
https://github.com/ant-design/ant-design/blob/master/components/pagination/Pagination.tsx  has done it but A LOT fewer options, only options that we really will need. 

We have many pages that require pagination, think the only place that use it now is the server admin user list. We do have a Pagination component already in grafana/ui but it's super duper simple .

We need it do be able to handle this 
Screenshot from 2020-11-20 08-54-54

And when too many pages do this:
Screenshot from 2020-11-20 08-56-56

We may also try to just solve those use cases with the component we have instead of using rc-pagination as the base. 

Here is our current Pagination component:
https://developers.grafana.com/ui/latest/index.html?path=/story/buttons-pagination--with-pages 

And in the code: 
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Pagination/Pagination.tsx

This how bad our current pagination component is:
Screenshot from 2020-11-20 09-02-35

nathanrodman:

For large numbers of pages the pagination component would render a button for every single page. This PR adds in helpful next and previous buttons and only renders the relevant page buttons in context to the current page

Screenshot_2021-01-08 Buttons Pagination - With Pages ⋅ Storybook

Fixes https://github.com/grafana/grafana/issues/29260

### DIFF
--- a/packages/grafana-ui/src/components/Pagination/Pagination.tsx
+++ b/packages/grafana-ui/src/components/Pagination/Pagination.tsx
@@ -2,6 +2,9 @@ import React from 'react';
 import { css } from 'emotion';
 import { stylesFactory } from '../../themes';
 import { Button, ButtonVariant } from '../Button';
+import { Icon } from '../Icon/Icon';
+
+const PAGE_LENGTH_TO_CONDENSE = 8;
 
 interface Props {
   /** The current page index being shown.  */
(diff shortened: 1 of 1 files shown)


torkelo:
Nice!

thisisobate (approved):
LGTM! 🎉

Great work @nathanrodman 

@dprokop Please do take a look whenever you're free

packages/grafana-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -16,21 +19,57 @@ export const Pagination: React.FC<Props> = ({ currentPage, numberOfPages, onNavi
...
+          if (condensePages) {
+            if (page === 1 || page === numberOfPages || (page >= currentPage - 2 && page <= currentPage + 2)) {
+              pagesToRender.push(getListItem(page, variant));
+            } else if (page === currentPage - 3 || page === currentPage + 3) {
+              pagesToRender.push(
+                <li key={page} className={styles.item}>
+                  <Icon className={styles.ellipsis} name="ellipsis-v" />
+                </li>
+              );
+            }
```
dprokop, line 57: I think this logic is causing this fairly weird behavior of the navigation arrows changing it's position when navigating, as in the gif below:

packages/grafana-ui/src/components/Pagination/Pagination.tsx
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import { css } from 'emotion';
 import { stylesFactory } from '../../themes';
 import { Button, ButtonVariant } from '../Button';
+import { Icon } from '../Icon/Icon';
+
+const PAGE_LENGTH_TO_CONDENSE = 10;
```
dprokop, line 7: Could this become a property of the Pagination component, so that it could be easily adjusted if necessary? We can still default to 10 :)

dprokop (approved):
@nathanrodman I think this looks great. I have two comments that would be good to address before merging - other than that it's great.

packages/grafana-ui/src/components/Pagination/Pagination.tsx
dprokop: Regarding this comment - maybe @jessover9000 or @hemdrup could chime in and provide some more feedback from the UX POV?

jessover9000: Yes, the jumping buttons are strange. If we want as little as possible jumping when paging through, we need to make sure we always have the same count of elements, so if we have one … separator we need one outer button + 7 buttons on the other side,  or for two … separators two outer buttons and 5 middle buttons. If we think that's a bit too many buttons or we need to make it work for narrow panel widths, we can also just use 3 middle buttons instead of 5, but it should be an odd number so that the active page is centered.
Concretely for this example that would mean:

- Beginning of pagination: Buttons from 1–7 + … + last number
- Somewhere in the middle: 1 + … + 2 previous buttons + 1x current highlighted number + 2 next buttons + … + last number
- End of pagination: 1 + … + 6 buttons prior to last number + last number

Hope it's clear what I mean 🤪

hugohaggmark, line 57: I think this logic is hard to understand and reason about without comments and or tests. Maybe you could add tests or comments so other can understand the magic numbers here without debugging the code?

hugohaggmark (changes requested):

packages/grafana-ui/src/components/Pagination/Pagination.tsx
nathanrodman: This is proving to be a challenging logic problem. I've drawn out a few scenarios here that I'll try to implement

This is using 5 middle buttons, with a total length of elements including navigation buttons at 11

<  1  2  3  ***4***  5  6  7 ...  10  >
<  1 ...  4  5  ***6***  7  8  9  10  >
<  1 ...  8  9  ***10***  11  12 ...  15  >

If it were to be 3 middle buttons it would look something like this

<  1  2  3  ***4***  5  ...  10  >
<  1 ...  5  ***6***  7  ...  10  >
<  1 ...  8  ***9*** 10 >

nathanrodman: it's about to get even more complicated so I will definitely add some helpful comments and see about adding tests :)

hugohaggmark: Thank you 🙏

nathanrodman:
I added some more logic for handling a few important edge cases, namely rendering the same number of elements when scrolling through the pages.

https://user-images.githubusercontent.com/3469390/105530822-36909d00-5c9d-11eb-93f1-4c2a40ae68e9.mov

I created a few named variables to help with some of the complicated logic. Let me know how things look @hugohaggmark @dprokop @thisisobate @jessover9000

hugohaggmark (commented):
Great job on naming and comments and it looks a lot better having the reduce in a separate function, can we move the reduce function outside of the component and add tests for it?

torkelo (approved):
Looks great now, one small improvement could be to update the storybook story with more examples that cover the different states (different nummber of pages, current page at the start, middle or end)

hugohaggmark (approved):
Good to go! Thank you for contributing @nathanrodman!

PR #30154, opened 2021-01-09
AngularQueryEditors: Fixes to Graphite query editor and other who refer to other queries

torkelo:
Fixes to angular query editors that used panel.targets to refer to other queries, caused by recent (in master) changes.

Also fixes so that angular editors that call refresh also triggers a onRunQuery

### DIFF
--- a/public/app/features/query/components/QueryEditorRow.tsx
+++ b/public/app/features/query/components/QueryEditorRow.tsx
@@ -221,7 +221,7 @@ export class QueryEditorRow extends PureComponent<Props, State> {
 
   onDisableQuery = () => {
     this.props.query.hide = !this.props.query.hide;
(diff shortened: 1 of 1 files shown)


jackw (approved):
LGTM! 🚀

public/app/features/query/components/QueryEditorRow.tsx
```diff
@@ -224,7 +221,7 @@ export class QueryEditorRow extends PureComponent<Props, State> {
 
   onDisableQuery = () => {
     this.props.query.hide = !this.props.query.hide;
-    this.onRunQuery();
+    this.props.onRunQuery();
```
dprokop, line 224: oh yes, good catch!

dprokop (approved):

PR #30156, opened 2021-01-09
Gauge: Fixes issue with all null values cause min & max to be null

### LINKED ISSUE #25279: Gauge: Panel throws error when all values are null
**What happened**:
When the query returns no data, the Gauge visualization throws the error "Cannot read Property 'toFixed' of undefined instead of showing 'No Data'. 
image
image

These  2 panels have the same query, and everything works well when there is data
image

'No Data' message displayed, like it is in other panels.

Use Gauge visalization on a query returning no data.

**Environment**:
- Grafana version: 7.0.2
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

torkelo:
Fixes #25279

* [x] Adds e2e test for gauge, checks for rendered gauges & no errors 

The bug was caused by all null series causing min & max to be null. I do not like this fix (in Gauge.tsx). Would prefer to either use:
* [getFieldConfigWithMinMax](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/field/scale.ts) But this is not possible as Gauge does not have access to the full source field only FieldConfig. 
* or some how modify [getFieldDisplayValues](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/field/fieldDisplay.ts#L133) to ensure a default min & max.

### DIFF
--- a/e2e/suite1/specs/gauge.spec.ts
+++ b/e2e/suite1/specs/gauge.spec.ts
@@ -1,0 +1,24 @@
+import { e2e } from '@grafana/e2e';
+
+e2e.scenario({
+  describeName: 'Gauge Panel',
+  itName: 'Gauge rendering e2e tests',
+  addScenarioDataSource: false,
+  addScenarioDashBoard: false,
+  skipScenario: false,
+  scenario: () => {
+    // open Panel Tests - Gauge
+    e2e.flows.openDashboard({ uid: '_5rDmaQiz' });
+
+    cy.wait(1000);
+
+    // check that gauges are rendered
+    e2e()
+      .get('body')
+      .find(`.flot-base`)
+      .should('have.length', 16);
+
+    // check that no panel errors exist
+    e2e.components.Panels.Panel.headerCornerInfo('error').should('not.exist');
+  },
+});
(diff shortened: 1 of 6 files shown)


public/app/features/dashboard/dashgrid/PanelHeader/PanelHeaderCorner.tsx
```diff
@@ -78,9 +79,12 @@ export class PanelHeaderCorner extends Component<Props> {
 
   renderCornerType(infoMode: InfoMode, content: PopoverContent, onClick?: () => void) {
     const theme = infoMode === InfoMode.Error ? 'error' : 'info';
+    const className = `panel-info-corner panel-info-corner--${infoMode.toLowerCase()}`;
+    const ariaLabel = selectors.components.Panels.Panel.headerCornerInfo(infoMode.toLowerCase());
```
hugohaggmark, line 83: Nice!

hugohaggmark (approved):
This looks good 🎉 , not very familiar with Gauge but maybe we should add a test cases for the when min & max are `null`?

e2e/suite1/specs/gauge.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  scenario: () => {
+    // open Panel Tests - Gauge
+    e2e.flows.openDashboard({ uid: '_5rDmaQiz' });
+
+    cy.wait(1000);
+
+    // check that gauges are rendered
+    e2e()
+      .get('body')
+      .find(`.flot-base`)
```
dprokop, line 18: Wonder if we maybe could introduce label selector for Gauge viz rather than relying on this confusing class name? WDYT?

dprokop (approved):
Left one nit to consider :)

e2e/suite1/specs/gauge.spec.ts
torkelo: not sure how without modifying the flot code, want to validate that the flot canvas exists

PR #30162, opened 2021-01-09
InlineSwitch: Added missing InlineSwitch component and fixed two places that used unaligned inline switch

torkelo:
Adds new component for InlineSwitch so we can share this common use case and updates the instances that I found where switch was used inline and was unaligned and without the proper inline input background.

* [x] Adds InlineSwitch component
* [x] Adds storybook story update that show both normal Switch use case & Inline

Before: 
Screenshot from 2021-01-09 14-44-57
Screenshot from 2021-01-09 14-44-44

After:
Screenshot from 2021-01-09 14-44-17
Screenshot from 2021-01-09 14-42-10

Updated story:
Screenshot from 2021-01-09 14-46-58

### DIFF
--- a/e2e/suite1/specs/variables/new-query-variable.ts
+++ b/e2e/suite1/specs/variables/new-query-variable.ts
@@ -66,27 +66,11 @@ describe('Variables - Add variable', () => {
       .within(select => {
         e2e.components.Select.singleValue().should('have.text', 'Disabled');
       });
-    e2e.pages.Dashboard.Settings.Variables.Edit.General.selectionOptionsMultiSwitch()
-      .should('be.visible')
-      .within(select => {
-        e2e()
-          .get('input')
-          .should('not.be.checked');
-      });
-    e2e.pages.Dashboard.Settings.Variables.Edit.General.selectionOptionsIncludeAllSwitch()
-      .should('be.visible')
-      .within(select => {
-        e2e()
-          .get('input')
-          .should('not.be.checked');
-      });
-    e2e.pages.Dashboard.Settings.Variables.Edit.QueryVariable.valueGroupsTagsEnabledSwitch()
-      .should('be.visible')
-      .within(select => {
-        e2e()
-          .get('input')
-          .should('not.be.checked');
-      });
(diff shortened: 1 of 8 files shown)


Clarity-89 (approved):
Looks good! Noticed that clicking on the label doesn't toggle the switch, because the inline label is looking for id of immediate child, which in this case is `div` instead of `input`. This has not been introduced in this PR but wonder if this could be fixed?

torkelo:
> Looks good! Noticed that clicking on the label doesn't toggle the switch, because the inline label is looking for id of immediate child, which in this case is div instead of input. 

@Clarity-89  not sure what you mean? this worked in the old switch that had both label + switch, but this new switch has no label. Cannot find any `for` parameter for the InlineField label that has no effect. Or do you mean something else?

Clarity-89:
> > Looks good! Noticed that clicking on the label doesn't toggle the switch, because the inline label is looking for id of immediate child, which in this case is div instead of input.
> 
> @Clarity-89 not sure what you mean? this worked in the old switch that had both label + switch, but this new switch has no label. Cannot find any `for` parameter for the InlineField label so that is why clicking it has no effect. Or do you mean something else?

Sorry, got confused myself about how this works. But basically, if you add `id` prop to Switch inside `InlineLabel` it will become automatically associated with the label and clicking the label will toggle the Switch.

```jsx
        <InlineFieldRow>
          <InlineField label="My switch">
            <InlineSwitch value={checked} disabled={disabled} onChange={onChange} id={'switch'} />
          </InlineField>
        </InlineFieldRow>
```

torkelo:
> Sorry, got confused myself about how this works. But basically, if you add id prop to Switch inside InlineField it will become automatically associated with the label and clicking the label will toggle the Switch.

Strange, I thought it created a unique id automatically

Clarity-89:
> > Sorry, got confused myself about how this works. But basically, if you add id prop to Switch inside InlineField it will become automatically associated with the label and clicking the label will toggle the Switch.
> 
> Strange, I thought it created a unique id automatically

Indeed, it does:https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Forms/InlineField.tsx#L35-L40, but the InlineField doesn't know anything about it since it's an internal Switch value.

PR #30163, opened 2021-01-09
GraphNG: Minor polish & updates to new time series panel and move it from alpha to beta

torkelo:
* [x] Updates name to `Time series' (white space)
* [x] Updates sort to 2 so it's after the standard graph panel
* [x] Updates default fill to zero
* [x] Fixes default gradient mode 
* [x] Adds beta badge to the panel in viz picker (and changes timeseries panel from alpha to beta)

### DIFF
--- a/packages/grafana-data/src/types/plugin.ts
+++ b/packages/grafana-data/src/types/plugin.ts
@@ -6,6 +6,7 @@ import { LiveChannelSupport } from './live';
 export enum PluginState {
   alpha = 'alpha', // Only included if `enable_alpha` config option is true
   beta = 'beta', // Will show a warning banner
+  stable = 'stable', // Will show a warning banner
   deprecated = 'deprecated', // Will continue to work -- but not show up in the options to add
 }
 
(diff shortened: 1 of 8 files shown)


packages/grafana-data/src/types/plugin.ts
```diff
@@ -6,6 +6,7 @@ import { LiveChannelSupport } from './live';
 export enum PluginState {
   alpha = 'alpha', // Only included if `enable_alpha` config option is true
   beta = 'beta', // Will show a warning banner
+  stable = 'stable', // Will show a warning banner
```
ryantxu, line 9: Will not show?

ryantxu (approved):

packages/grafana-data/src/types/plugin.ts
torkelo: yepp, will not show. Just in case someone put that in plugin.json

leeoniya: @torkelo i think he meant that you missed "not" in the comment :)

ryantxu: yup -- was on my phone so the suggestion button was awkward ;)

torkelo: Oops :) 
https://github.com/grafana/grafana/commit/7ca66b204004d4d0089de652cd47c51d4195cf36

PR #30165, opened 2021-01-10
Legends: Refactoring and rewrites of legend components to simplify components & reuse

torkelo:
This tries to untangle the GraphLegend and GraphLegendListItem & GraphLegendTableRow from the old Graph component and simplify the other legend components and storybook stories. I found the current components very hard to follow and reuse in other instances as the generic components were basically useless without using GraphLegend. 

Much more left to do to have more style resue and refactor some of the components and simplify them further. 

* [x] Move GraphLegend and other GraphLegend types from Graph/ to Legend/  , renamed to Legend (but later found this to have a conflict with Forms Legend) 
* [x] Make the base Legend component take in display mode and decide what legend type to render (LegendList or LegendTable)
* [x] Make the LegendList & LegendTable include the default item renderers to simplify usage
* [x] Rename Legend to VizLegend as Forms already export a Legend component name
* [ ] Is VizLegend and ok name? rename all other types and components with this prefix? 
* [x] Clean up styles and remove more of the inline css styles 
* [x] Update table styles
* [x] Update the color icon styles  (from line to something bigger like a square or circle) 
* [x] Storybook story works with color change and click highlighting 
* [x] Updated table row style & added row hover style 
* [x] Remove onSeriesAxisToggle
* [ ] What other interface changes do we need? Remove axis selection? Axis selection won't be possible with the new Graph is requires setting a unit as well. Technically possible if a right y-axis already exists. 
* [ ] Use of event bus?  (for hover events, and other things) 

 Two main stories, one with legend values 
Screenshot from 2021-01-10 21-18-01

And one without:
Screenshot from 2021-01-10 21-19-36

Color picker & clicking on items to highlight works in the storybook


packages/grafana-ui/src/components/Legend/LegendListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+
+      {item.displayValues && <LegendStatsList stats={item.displayValues} />}
+    </div>
+  );
+};
+
+LegendListItem.displayName = 'LegendTableItem';
+
+const getStyles = stylesFactory((theme: GrafanaTheme) => {
```
dprokop, line 52: FYI `stylesFactory` is not necessary until you need more args than theme. For theme only case yo can use [useStyles hook](https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md#usestyles-hook)

dprokop (commented):
This is looking good @torkelo!

dprokop:
> What other interface changes do we need? Remove axis selection? Axis selection won't be possible with the new Graph is requires setting a unit as well. Technically possible if a right y-axis already exists.

Yeah, this needs to be dropped as now will be configured via field config. I also always found it confusing to have the axis side config via the color picker, so I guess this will be a UX win at the end of the day.

dprokop:
@torkelo I think the circle works nicely as a legend indicator. The rounded square looks odd to me, does not feel like a legend item indicator:

image

What I have noticed though is that sometimes the legend icon is represented differently depending on the chart type. For instance it would be:
- a circle in Pie Chart. 
- a line with a dot or a shape representing series (-⚬- , -□-, -▵-) in Line Chart.

Longer-term I think we could maybe introduce `VizLegendItemStyle` interface to describe the behaviour above if we agree this is a good idea.

packages/grafana-ui/src/components/Legend/LegendListItem.tsx
torkelo: not sure I understand the advantage to useStyles? seems identical to stylesFactory but less. flexible? Why do we have two methods for the same thing?

torkelo:
Going with slightly beefier lines for now 
Screenshot from 2021-01-11 14-58-55

packages/grafana-ui/src/components/VizLegend/VizLegendList.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const VizLegendList: React.FunctionComponent<Props> = ({
+  items,
+  itemRenderer,
+  onSeriesColorChange,
+  onLabelClick,
+  placement,
+  className,
+}) => {
+  const theme = useContext(ThemeContext);
+  const styles = getStyles(theme);
```
dprokop, line 22: useStyles maybe?

dprokop, line 22: useStyles maybe?

packages/grafana-ui/src/components/VizLegend/VizLegendListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface Props {
+  item: VizLegendItem;
+  className?: string;
+  onLabelClick?: (item: VizLegendItem, event: React.MouseEvent<HTMLDivElement>) => void;
+  onSeriesColorChange?: SeriesColorChangeHandler;
+}
+
+export const VizLegendListItem: React.FunctionComponent<Props> = ({ item, onSeriesColorChange, onLabelClick }) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
```
dprokop, line 20: useStyles maybe?

packages/grafana-ui/src/components/VizLegend/VizLegendTableItem.tsx
```diff
@@ -1,113 +1,41 @@
...
-
-export const GraphLegendTableRow: React.FunctionComponent<GraphLegendItemProps> = ({
-  item,
-  onSeriesColorChange,
-  onToggleAxis,
   onLabelClick,
   className,
 }) => {
   const theme = useContext(ThemeContext);
   const styles = getStyles(theme);
```
dprokop, line 25: useStyles maybe?

packages/grafana-ui/src/components/Legend/LegendListItem.tsx
dprokop: Less code is I guess the advantage.

dprokop (approved):
No more comments that the useTheme vs useStyles. Great work!

dprokop:
Ah, one nit, or rather a question. Is the thead/tr/th rendered on purpose when there is no table legend header? The line at the top of the table looks off to me, think it looks better without it:

image

image


Here's the code that causes it:
image

torkelo:
> Ah, one nit, or rather a question. Is the thead/tr/th rendered on purpose when there is no table legend header? The line at the top of the table looks off to me, think it looks better without it:

@dprokop  have not changed that part, might be an existing bug in the table mode component when there are no stats

torkelo:
updated to use useStyles no :)

dprokop:
> @dprokop have not changed that part, might be an existing bug in the table mode component when there are no stats

probably, will fix it with stats work

simPod:
Dropping class on series icon somehow does this now on dac7add457e65b7e976af49fb8b2fd3483e2f409

image

public/app/plugins/panel/graph/Legend/LegendSeriesItem.tsx
```diff
@@ -197,7 +197,7 @@ class LegendSeriesIcon extends PureComponent<LegendSeriesIconProps, LegendSeries
         enableNamedColors
       >
         {({ ref, showColorPicker, hideColorPicker }) => (
-          <span ref={ref} onClick={showColorPicker} onMouseLeave={hideColorPicker} className="graph-legend-icon">
+          <span ref={ref} onClick={showColorPicker} onMouseLeave={hideColorPicker}>
```
simPod, line 200: reverting this change fixes the issue 🤔

PR #30166, opened 2021-01-10
Links: Add underline on hover for links in NewsPanel

torkelo:
We had no hover state for title link in NewsPanel.

This changes external links to follow similar design to GitHub. That is
blue color and show underline on hover.

Think this needs a more long term solution. Thinking of a component that
can capture the more common link usages. Like a StyledLink
component with an option for color = default | text | text-weak
(default would be blue). text would use text color and text-weak the weak (muted) text color
The hover state would always use blue + underline.

Looking at GitHub this would cover most of the link usage.


dprokop (approved):
LGTM

PR #30176, opened 2021-01-11
Settings: add config options to disable config variable expansion

tsharju:

Provides options for the server config to disable the filename variable expansion as well as the environment variable expansion in the server config file. This could be used to prevent possible security issues, if for instance config file is generated based on user input. An example of a possible exploit could be to set `google_analytics_ua_id` to something like `$__file{/proc/cpuinfo}` to obtain information or perhaps even security credentials stored to files.

The PR adds config variables `disable_env_variable_expansion` and `disable_file_variable_expansion` with boolean values that allow disabling each setting expander individually.

n/a

n/a

### DIFF
--- a/conf/defaults.ini
+++ b/conf/defaults.ini
@@ -227,6 +227,11 @@ x_content_type_options = true
 # when they detect reflected cross-site scripting (XSS) attacks.
 x_xss_protection = true
 
+# Set to true to disable config variable expansion from environment variables
+disable_env_variable_expansion = false
+
+# Set to true to disable config variable expansion from a file path
+disable_file_variable_expansion = false
 
 #################################### Snapshots ###########################
 [snapshots]
--- a/pkg/setting/expanders.go
+++ b/pkg/setting/expanders.go
@@ -43,6 +43,17 @@ func AddExpander(name string, priority int64, e Expander) {
 	})
 }
 
+func removeExpander(name string) {
+	n := 0
+	for _, x := range expanders {
+		if x.name != name {
+			expanders[n] = x
+			n++
+		}
+	}
+	expanders = expanders[:n]
+}
+
 var regex = regexp.MustCompile(`\$(|__\w+){([^}]+)}`)
 
 func expandConfig(file *ini.File) error {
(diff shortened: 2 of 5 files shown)


aknuds1:
@torkelo @bergquist @marefr @daniellee What do you think of this feature?

bergquist:
Thank you for contributing security features :)

But if the user has access to the grafana.ini file to set those file expansions they could also disable that setting. right? 
I wonder if this actually increases security or just introduces another setting.

tsharju:
Very good point. Our use case has been more that there are some configuration parameters that we allow users to specify e.g. the google analytics id and having the variable expansion available feels not too safe. Alternative option would be to have a server startup option to disable this functionality that cannot be changed during runtime.

marefr:
Isn't "the proper thing" in this situation to limit the permission of the account running Grafana? 

From my perspective this seems like an edge case since we've had no other users requesting this and there doesn't seem to be any existing feature request issue. I understand your problem though. We usually suggest opening a feature request issue to track the activity of certain features to allow us to understand whether this is a problem shared by many users and in that case is an indication that we need to prioritize it/accept a pull request.

tsharju:
Limiting permissions of the account running Grafana is naturally one option. As the environment variable expansion is something that we wanted to disable as well, I thought it would handle the files the same. If needed I can open an issue to discuss the topic and see whether there is need for such feature.

aknuds1:
You're saying that a case you have in mind is generating the configuration file based on user input, couldn't you instead sanitize the user input to make sure it doesn't contain variables? I might be overlooking something, but it sounds to me as if this issue should be caught on a level above Grafana.

marefr:
Thank you for your contribution, but I'm closing this based on earlier discussions.

paulwouters:
Obviously writing a parser to deal with complicated parsing to disable things and trying work around various ways of quoted, backslashing, and  other tweaks to avoid detection is more complicated and error prone than just allowing a "service mode" option that restricts grafana when offered "as a service" without the enduser having host access to grafana. The patch is more trivial than doing this elsewhere.

PR #30192, opened 2021-01-11
GraphNG: Custom point symbols

dprokop:
WIP

Adds possitbility to customize the shape of a point on the graph. Available shapes are dot, square, triangle, marble, star and cross:

image
image

TODO:

-  [ ] Tinker more with shape sizings

@leeoniya is ther any way to customize the hover highlight of the point? Not sure why it renders a circle with a transparent border, probably you have encountered some alignment issues? At least that's what I've seen when I've played with modifying uPlot code. The highlight with custom shapes work quite weird with the shapes going outside of the inner circle and that looks not so nice.

### DIFF
--- a/packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
+++ b/packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
@@ -5,9 +5,10 @@ import {
   DrawStyle,
   LineConfig,
   FillConfig,
-  PointsConfig,
-  PointVisibility,
   LineInterpolation,
+  PointsConfig,
+  PointSymbol,
+  PointVisibility,
   FillGradientMode,
 } from '../config';
 import { PlotConfigBuilder } from '../types';
@@ -29,6 +30,7 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
       showPoints,
       pointColor,
       pointSize,
+      pointSymbol,
       scaleKey,
       spanNulls,
       show = true,
@@ -70,6 +72,13 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
       pointsConfig.points!.show = false;
     } else if (showPoints === PointVisibility.Always) {
       pointsConfig.points!.show = true;
+    }
+
+    if (pointsConfig.points!.show) {
+      if (pointSymbol && pointSymbol !== PointSymbol.Dot) {
+        console.log(pointSize);
+        pointsConfig.points!.show = getCustomPointRenderer(pointSymbol, pointColor!, pointSize!);
+      }
     }
 
     return {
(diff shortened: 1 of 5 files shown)


packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -204,3 +213,104 @@ function getCanvasGradient(opts: AreaGradientOptions): (self: uPlot, seriesIdx:
...
+      drawFn = drawTriangle;
+      break;
+    case PointSymbol.Marble:
+      shapeSize = size * 3;
+      drawFn = drawMarble;
+      break;
+    case PointSymbol.Cross:
+      shapeSize = size * 2;
+      drawFn = drawCross;
+      break;
```
dprokop, line 241: sizing is quick hacking here to make the shapes look sort of ok. Needs some love

public/app/plugins/panel/timeseries/TimeSeriesPanel.tsx
```diff
@@ -41,6 +41,7 @@ export const TimeSeriesPanel: React.FC<TimeSeriesPanelProps> = ({
       <TooltipPlugin mode={options.tooltipOptions.mode as any} timeZone={timeZone} />
       <ZoomPlugin onZoom={onChangeTimeRange} />
       <ContextMenuPlugin timeZone={timeZone} replaceVariables={replaceVariables} />
+      {/*<CustomPointsPlugin />*/}
```
dprokop, line 44:
```suggestion
```

leeoniya:
> is there any way to customize the hover highlight of the point? Not sure why it renders a circle with a transparent border, probably you have encountered some alignment issues?

so, right now the hover points are html elements initialized and returned from `cursor.points.show` [1]. i *think* you can return an custom shape by crafting and returning an inline `<svg>` element.

longer-term (after upcoming release), it makes sense for uPlot to standardize around svg rather than html, even the canvas drawn points can be initialized via a Path2D object constucted from and svg pathstring. this would allow using the same svg shapes for both in-dom and on-canvas points.

[1] https://github.com/leeoniya/uPlot/blob/e870fc16cef352fb283a050957ec08374e21a47f/dist/uPlot.d.ts#L377

ryantxu:
For reference, in plotly studio, this is the UI:  
image

These are the options:
https://plotly.com/javascript/reference/scatter/#scatter-marker-symbol

https://github.com/plotly/plotly.js/blob/c72264cfa66dde899a0dfcd4da08c79dedb75ff7/src/components/drawing/symbol_defs.js

ryantxu:
Rather than using custom canvas path rendering... what about an SVG/font?  this would have the added benefit that we can use it in the dom OR canvas?

Is this something we should punt till after 7.4?  Its cool, but seems less important than some some of the other issues

dprokop:
> Rather than using custom canvas path rendering... what about an SVG/font? this would have the added benefit that we can use it in the dom OR canvas?
> Is this something we should punt till after 7.4? Its cool, but seems less important than some some of the other issues

AFAIK we can render svg on canvas. No rush with this for 7.4

Issue #9305, opened 2017-09-19
[Feature request] Allow hiding grid lines


tuukkamustonen:
To make dashboards a little cleaner, it would be nice to be able to remove the horizontal grid lines in Graph panel, to achieve something like this:

![image](https://user-images.githubusercontent.com/94327/30594753-d49189a0-9d57-11e7-9f05-209959f390f0.png)

I only found https://community.grafana.com/t/hiding-gridlines-in-grafana/1684 that states that this is not possible at the moment.

Diegonarducci88:
+1

dastrix80:
Was this completed? id love to have the lines removed...

jperkins909:
+1

tyvich:
Some research work on the benefits on grid lines in graphs:
https://www.perceptualedge.com/articles/dmreview/grid_lines.pdf 

Gridlines can add a lot of visual clutter to dashboards. A toggle to remove would be great, especially for visually dense dashboards

fasavard:
+1

zamashal:
+1
Any updates on this feature?

bpthur:
+1

finnhuman:
+1

cpmoore:
Is implementation planned for this?

ojek:
+1

benickj:
+1

this would be a very helpful feature! 
We have several timeseries data in a graph with single vertical line for each value. Sometimes it is very hard to say if the line belongs to a real value or is just a grid line.
So would be great if we have the possibility to toggle the horizontal and vertical grid lines seperate.

felipegutierrez:
+1

guokailun:
+1

flipp3003:
+1

I've been waiting for this for a long time

gurpartb:
+1

windworship:
+1

mordowiciel:
+1!

Varrah:
+1

josephctyler:
+1

eulers-identity:
+1

GregoryKowalik:
+1

henfiber:
+1

marunu:
+1

frogale:
+1

sceppi:
+1

nicolaihenriksen:
+1

durgaprasad1709:
+1

Wailshire:
+1

[...]

PR #30195, opened 2021-01-11
GraphNG: Hide grid for right-y axis if left x-axis exists

torkelo:
Fixes so we do not render double grid lines for both left and right axis

Todo
* Fix so it works if you only have a right axis


torkelo:
working on an alternative solution now

leeoniya:
any chance of squashing https://github.com/grafana/grafana/issues/9305 as part of this work?

torkelo:
not sure I want new options for that yet, maybe once we have a way to hide advanced options

torkelo:
Switched to a better soution

ryantxu (approved):

dprokop (approved):

Issue #29895, opened 2020-12-17
PanelEdit: Clean field config properties when switching panel plugin


torkelo:
We currently clean all top-level properties and the panel options when switching visualization.  We also remember these options so when switching back the are restored. 

We hover preserve all field config options. The main reason is that we want to preserve the standard field options. 

TODO
* [ ] Clean/remove any custom field config options from defaults and overrides. And remember them so we can restore them if we switch back to the same visualisation
* [ ] Clean/remove any standard field options that are not used by the new panel, remember them in case switching to a new panel (or back) that does. 

Relevant area where this logic is: 
https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/state/PanelModel.ts#L373

PR #30197, opened 2021-01-11
Field overrides: skipping overrides for properties no longer existing in plugin

mckn:
Prior to this PR the field overrides editor would explode if trying to render an override containing properties that does not longer exists in the panel.

This PR prevents that from happening by rendering the properties as "unkown property" and adds a possibility to remove it from the configuration.

Skärmavbild 2021-01-12 kl  07 29 33


mckn:
I'm just going to add some tests before moving this from draft.

public/app/features/dashboard/components/PanelEditor/OverrideEditor.tsx
```diff
@@ -114,8 +103,13 @@ export const OverrideEditor: React.FC<OverrideEditorProps> = ({
     });
 
   const renderOverrideTitle = (isExpanded: boolean) => {
-    const overriddenProperites = override.properties.map(p => registry.get(p.id).name).join(', ');
+    const overriddenProperites = override.properties
+      .map(p => registry.getIfExists(p.id)?.name)
+      .filter(name => !!name)
+      .join(', ');
+
```
torkelo, line 110: why do we do this (map to just name)? and then map over the properties (the names) and call registry.getIfExists again? why not keep this be the registry items and map over them later. 

So would be jus to filter out the undefined / null ones.

dprokop:
Pushed minor UI change @mckn :

image
image

public/app/features/dashboard/components/PanelEditor/OverrideEditor.tsx
dprokop: same question :)

dprokop (approved):

public/app/features/dashboard/components/PanelEditor/OverrideEditor.tsx
mckn: Not really sure I am following. Can you explain a bit more how you think I could change it?

torkelo: Sorry I misread the diff wrong. My comment does not make sense. 

But after reading the code more, It feels like we can get properties and filter out null items once, now it's in two places.  (in renderOverrideTitle and in the body), and two calls to registry.get, think we can simplify the code a bit that way.  

```ts
const configProps =  override.properties
      .map(p => registry.getIfExists(p.id))
      .filter(p => p !== null)
```

torkelo:
Oh shit, maybe this won't need to show the properties, if we did this: https://github.com/grafana/grafana/issues/29895 this should technically not happen

public/app/features/dashboard/components/PanelEditor/OverrideEditor.tsx
mckn: Super, I will fix that tomorrow.

PR #30198, opened 2021-01-11
Plugins: Prevent users with no enough permission to navigate to applications

### LINKED ISSUE #30196: Plugins: Prevent users with no enough permission to navigate to applications
**What happened**:

I copy-pasted the URL of an application plugin that someone shared with me. 
That URL corresponds to an app "includes" only visible for admins, so I got weird errors.

To be redirected to the Grafana main page, as I'm redirected when I try to navigate to any `/admin` URL logged in with a non-admin user.

1. Install any data app plugin with admins-only includes.
2. Log in with a non-admin user.
3. Try to navigate (manually introducing the URL) to the include path.

- Environment: N/A
- Grafana version: latest 
- User OS & Browser: N/A
- Grafana plugins: any app plugin with admin-only includes.

joanlopez:

Adds a new middleware middleware to check for permission on loading app plugin paths.

Fixes #30196 

- Expected behaviour:
  - If the application has no includes, do nothing (since there's no role setting at application/plugin level).
  - Otherwise, it looks for the (URL) matching include and checks if the logged in user has enough permission.
  - If it's not, it's redirected to the home page. 
- Few additional comments left inline.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -73,7 +73,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/plugins/:id/", reqSignedIn, hs.Index)
 	r.Get("/plugins/:id/edit", reqSignedIn, hs.Index) // deprecated
 	r.Get("/plugins/:id/page/:page", reqSignedIn, hs.Index)
-	r.Get("/a/:id/*", reqSignedIn, hs.Index) // App Root Page
+	r.Get("/a/:id/", reqSignedIn, middleware.ReqAppRoles, hs.Index)  // App root page
+	r.Get("/a/:id/*", reqSignedIn, middleware.ReqAppRoles, hs.Index) // App sub pages
 
 	r.Get("/d/:uid/:slug", reqSignedIn, redirectFromLegacyPanelEditURL, hs.Index)
 	r.Get("/d/:uid", reqSignedIn, redirectFromLegacyPanelEditURL, hs.Index)
--- a/pkg/middleware/plugins.go
+++ b/pkg/middleware/plugins.go
@@ -22,11 +22,11 @@ func ReqAppRoles(c *models.ReqContext) {
 
 	ok := false
 	for _, include := range plugin.Includes {
-		if !c.OrgRole.Includes(include.Role) {
+		if !strings.Contains(c.Req.RequestURI, include.Path) {
 			continue
 		}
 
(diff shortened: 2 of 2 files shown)


pkg/api/api.go
```diff
@@ -73,7 +73,8 @@ func (hs *HTTPServer) registerRoutes() {
 	r.Get("/plugins/:id/", reqSignedIn, hs.Index)
 	r.Get("/plugins/:id/edit", reqSignedIn, hs.Index) // deprecated
 	r.Get("/plugins/:id/page/:page", reqSignedIn, hs.Index)
-	r.Get("/a/:id/*", reqSignedIn, hs.Index) // App Root Page
+	r.Get("/a/:id/", reqSignedIn, middleware.ReqAppRoles, hs.Index)  // App root page
+	r.Get("/a/:id/*", reqSignedIn, middleware.ReqAppRoles, hs.Index) // App sub pages
```
joanlopez, line 77: I had to duplicate the route definition for `/` and `/*` because the latter is mandatory (not empty `*` is allowed), so the `/` is needed to support those applications that only uses the root page (i.e. tabs-based app).

Note that this pattern (defining both `/` and `/*`) is not something new, but something already present multiple times in the same file.

ryantxu:
does this need to be done on the backend?   Would it be better to have a good fronted pattern so we can have more fine grained control?  In grafana 8, this will be react router based

joanlopez:
> does this need to be done on the backend? Would it be better to have a good fronted pattern so we can have more fine grained control? In grafana 8, this will be react router based

I feel like I have not enough context to say where's the best place to make that happen (either backend / frontend). However, what seems to be clear is that we need to rethink the way we handle routes in both (frontend & backend), since in this case is currently working because of the "allow by default" policy which is extremely error-prone IMHO.

Do you think someone with a larger experience in both (backend & frontend) can bring some for feedback? 🤔

sakjur:
I hadn't thought of it before.
I'd agree with the notion that the best place to place this is in the frontend. All validation of frontend logic is there for the user's sake rather than securing the Grafana instance, and I'm not sure why we'd want to enforce that from the backend.

It'll help us reduce the connection between the frontend and backend, which I think is good for separation of concerns.

That being said, I'm not sure how urgent this is and if it's sensible to wait until Grafana 8 to get this feature in? (or maybe it can be done today already?)

PR #30203, opened 2021-01-11
Auth: add expired token error and update CreateToken function

AgnesToulet:
- Add expired token error and save it into context data (with the token ID and the user ID)
- Update CreateToken function signature to have more information about user and request

These updates allow us to have more information in the Auditing feature


pkg/services/auth/auth_token.go
```diff
@@ -133,6 +141,10 @@ func (s *UserAuthTokenService) LookupToken(ctx context.Context, unhashedToken st
 		return nil, models.ErrUserTokenNotFound
 	}
 
+	if model.CreatedAt <= s.createdAfterParam() || model.RotatedAt <= s.rotatedAfterParam() {
+		return &models.UserToken{Id: model.Id, UserId: model.UserId}, models.ErrUserTokenExpired
```
marefr, line 145: Feels weird to return both a token and an error. Have you considered handling this scenario in a background job instead? I'm not sure that makes things easier though.

pkg/services/auth/auth_token.go
```diff
@@ -60,7 +61,7 @@ func (s *UserAuthTokenService) ActiveTokenCount(ctx context.Context) (int64, err
 	return count, err
 }
 
-func (s *UserAuthTokenService) CreateToken(ctx context.Context, userId int64, clientIP net.IP, userAgent string) (*models.UserToken, error) {
+func (s *UserAuthTokenService) CreateToken(c *models.ReqContext, user *models.User) (*models.UserToken, error) {
```
marefr, line 64: I understand you want more context available for auditing purpose. Is the idea to extract additional request headers and so on in enterprise from models.ReqContext or what's the purpose of this?

pkg/models/user_token.go
```diff
@@ -33,7 +34,7 @@ type RevokeAuthTokenCmd struct {
 
 // UserTokenService are used for generating and validating user tokens
 type UserTokenService interface {
-	CreateToken(ctx context.Context, userId int64, clientIP net.IP, userAgent string) (*UserToken, error)
```
xlson, line 36: What prompted this change? It doesn't make sense to me as we're now mixing db calls and request logic in the same method, making the scope of this method really large.

pkg/services/auth/auth_token.go
AgnesToulet: I created a new struct `TokenExpiredError` to hold more information in the error itself and not return the token anymore, WDYT?

pkg/services/auth/auth_token.go
```diff
@@ -133,6 +141,13 @@ func (s *UserAuthTokenService) LookupToken(ctx context.Context, unhashedToken st
 		return nil, models.ErrUserTokenNotFound
 	}
 
+	if model.CreatedAt <= s.createdAfterParam() || model.RotatedAt <= s.rotatedAfterParam() {
```
marefr, line 144: This changes the logic compared to earlier which uses greater than compared to greater than or equal and in addition AND instead of OR. Any reason for this change?

AgnesToulet: Yes, we want the opposite of what were done before. Before, we wanted to only retrieve valid tokens so the condition was equal to "token is not expired". Now, we want to retrieve valid and expired tokens but if the token is expired we return an error, so the condition here is "token is expired".

marefr: Yeah of course. Sorry for the confusion

marefr:
Before merge you should run the auth_token load tests to verify before and after changes behaves without increased latency or errors: https://github.com/grafana/grafana/tree/master/devenv/docker/loadtest

pkg/api/login.go
```diff
@@ -258,10 +259,12 @@ func (hs *HTTPServer) loginUserWithUser(user *models.User, c *models.ReqContext)
 	}
 
 	hs.log.Debug("Got IP address from client address", "addr", addr, "ip", ip)
-	userToken, err := hs.AuthTokenService.CreateToken(c.Req.Context(), user.Id, ip, c.Req.UserAgent())
+	ctx := context.WithValue(c.Req.Context(), models.RequestURIKey{}, c.Req.RequestURI)
```
aknuds1, line 262: Where is the `RequestURIKey` key/value pair on the context used?

AgnesToulet: It's used in the associated Enterprise PR, maybe I should add it in the `RequestURIKey` comment?

aknuds1: Aha, yeah it might be a good idea to add a comment so it doesn't look unused.

AgnesToulet: I added the comment!

aknuds1: Yeah I saw, thanks :)

pkg/services/auth/auth_token.go
AgnesToulet: Besides the client IP address and the user agent, I needed the request URI information. Leo helped me find a solution where I only add the information I need and not the whole `ReqContext` object.

aknuds1 (approved):
LGTM

PR #30209, opened 2021-01-11
Use Node.js 14.x in plugin CI

stevenvachon:
Trying to avoid https://github.com/facebook/jest/issues/8769


aknuds1 (approved):
Looks good AFAICT.

Issue #3269, opened 2015-11-17
the `org_user` table schema should utilize foreign keys for `user_id` and `org_id`


drawks:
Lack of foreign keys here makes the data unconstrained by the reality of the contents of `users` and `orgs` which compounds the complexity of troubleshooting various issues with auth backend integrations.

utkarshcmu:
:+1:

torkelo:
Grafana supports a lot of different SQL data sources, and foreign keys can have some major differences that will be problematic. Another reason we skip foreign keys is that we do a log of schema migrations and changes and not having to deal with foreign keys when doing those can be a huge time saver. 

But I agree we have had some bugs due to this, but overall they have been very few an non critical

PR #30212, opened 2021-01-12
LibraryPanels: adds connections

hugohaggmark:
This PR adds the ability to connect a dashboard with library panel by adding a new table and new endpoints. I'm not sure this should be part of the LibraryPanel service, or if this should be a separate service, please let me know. Thank you.

Enjoy the review 🎉 

Relates #29610 

### DIFF
--- a/pkg/services/librarypanels/api.go
+++ b/pkg/services/librarypanels/api.go
@@ -107,7 +107,7 @@ func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response
 
 // getConnectedDashboardsHandler handles GET /api/library-panels/:uid/dashboard/.
 func (lps *LibraryPanelService) getConnectedDashboardsHandler(c *models.ReqContext) api.Response {
-	dashboardIds, err := lps.getConnectedDashboards(c, c.Params(":uid"))
+	dashboardIDs, err := lps.getConnectedDashboards(c, c.Params(":uid"))
 	if err != nil {
 		if errors.Is(err, errLibraryPanelNotFound) {
 			return api.Error(404, errLibraryPanelNotFound.Error(), err)
@@ -115,7 +115,7 @@ func (lps *LibraryPanelService) getConnectedDashboardsHandler(c *models.ReqConte
 		return api.Error(500, "Failed to get connected dashboards", err)
 	}
 
-	return api.JSON(200, util.DynMap{"result": dashboardIds})
+	return api.JSON(200, util.DynMap{"result": dashboardIDs})
 }
 
 // patchHandler handles PATCH /api/library-panels/:uid
--- a/pkg/services/librarypanels/api.go
+++ b/pkg/services/librarypanels/api.go
@@ -18,9 +18,12 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
 
 	lps.RouteRegister.Group("/api/library-panels", func(libraryPanels routing.RouteRegister) {
 		libraryPanels.Post("/", middleware.ReqSignedIn, binding.Bind(createLibraryPanelCommand{}), api.Wrap(lps.createHandler))
+		libraryPanels.Post("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.connectHandler))
 		libraryPanels.Delete("/:uid", middleware.ReqSignedIn, api.Wrap(lps.deleteHandler))
+		libraryPanels.Delete("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.disconnectHandler))
 		libraryPanels.Get("/", middleware.ReqSignedIn, api.Wrap(lps.getAllHandler))
 		libraryPanels.Get("/:uid", middleware.ReqSignedIn, api.Wrap(lps.getHandler))
+		libraryPanels.Get("/:uid/dashboard/", middleware.ReqSignedIn, api.Wrap(lps.getConnectedDashboardsHandler))
 		libraryPanels.Patch("/:uid", middleware.ReqSignedIn, binding.Bind(patchLibraryPanelCommand{}), api.Wrap(lps.patchHandler))
 	})
 }
@@ -38,7 +41,19 @@ func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd createLi
 	return api.JSON(200, util.DynMap{"result": panel})
 }
 
-// deleteHandler handles DELETE /api/library-panels/:uid
+// connectHandler handles POST /api/library-panels/:uid/dashboard/:dashboardId.
+func (lps *LibraryPanelService) connectHandler(c *models.ReqContext) api.Response {
+	if err := lps.connectDashboard(c, c.Params(":uid"), c.ParamsInt64(":dashboardId")); err != nil {
+		if errors.Is(err, errLibraryPanelNotFound) {
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to connect library panel", err)
+	}
+
+	return api.Success("Library panel connected")
+}
+
+// deleteHandler handles DELETE /api/library-panels/:uid.
 func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response {
 	err := lps.deleteLibraryPanel(c, c.Params(":uid"))
 	if err != nil {
@@ -51,7 +66,23 @@ func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response
 	return api.Success("Library panel deleted")
 }
 
-// getHandler handles GET /api/library-panels/:uid
+// disconnectHandler handles DELETE /api/library-panels/:uid/dashboard/:dashboardId.
+func (lps *LibraryPanelService) disconnectHandler(c *models.ReqContext) api.Response {
+	err := lps.disconnectDashboard(c, c.Params(":uid"), c.ParamsInt64(":dashboardId"))
+	if err != nil {
+		if errors.Is(err, errLibraryPanelNotFound) {
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		if errors.Is(err, errLibraryPanelDashboardNotFound) {
+			return api.Error(404, errLibraryPanelDashboardNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to disconnect library panel", err)
+	}
+
+	return api.Success("Library panel disconnected")
+}
+
+// getHandler handles GET /api/library-panels/:uid.
 func (lps *LibraryPanelService) getHandler(c *models.ReqContext) api.Response {
 	libraryPanel, err := lps.getLibraryPanel(c, c.Params(":uid"))
 	if err != nil {
@@ -64,7 +95,7 @@ func (lps *LibraryPanelService) getHandler(c *models.ReqContext) api.Response {
 	return api.JSON(200, util.DynMap{"result": libraryPanel})
 }
 
-// getAllHandler handles GET /api/library-panels/
+// getAllHandler handles GET /api/library-panels/.
 func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response {
 	libraryPanels, err := lps.getAllLibraryPanels(c)
 	if err != nil {
@@ -72,6 +103,19 @@ func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response
 	}
 
 	return api.JSON(200, util.DynMap{"result": libraryPanels})
+}
+
+// getConnectedDashboardsHandler handles GET /api/library-panels/:uid/dashboard/.
+func (lps *LibraryPanelService) getConnectedDashboardsHandler(c *models.ReqContext) api.Response {
+	dashboardIds, err := lps.getConnectedDashboards(c, c.Params(":uid"))
+	if err != nil {
+		if errors.Is(err, errLibraryPanelNotFound) {
+			return api.Error(404, errLibraryPanelNotFound.Error(), err)
+		}
+		return api.Error(500, "Failed to get connected dashboards", err)
+	}
+
+	return api.JSON(200, util.DynMap{"result": dashboardIds})
 }
 
 // patchHandler handles PATCH /api/library-panels/:uid
--- a/pkg/services/librarypanels/api.go
+++ b/pkg/services/librarypanels/api.go
@@ -18,12 +18,12 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
 
 	lps.RouteRegister.Group("/api/library-panels", func(libraryPanels routing.RouteRegister) {
 		libraryPanels.Post("/", middleware.ReqSignedIn, binding.Bind(createLibraryPanelCommand{}), api.Wrap(lps.createHandler))
-		libraryPanels.Post("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.connectHandler))
+		libraryPanels.Post("/:uid/dashboards/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.connectHandler))
 		libraryPanels.Delete("/:uid", middleware.ReqSignedIn, api.Wrap(lps.deleteHandler))
-		libraryPanels.Delete("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.disconnectHandler))
+		libraryPanels.Delete("/:uid/dashboards/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.disconnectHandler))
 		libraryPanels.Get("/", middleware.ReqSignedIn, api.Wrap(lps.getAllHandler))
 		libraryPanels.Get("/:uid", middleware.ReqSignedIn, api.Wrap(lps.getHandler))
-		libraryPanels.Get("/:uid/dashboard/", middleware.ReqSignedIn, api.Wrap(lps.getConnectedDashboardsHandler))
+		libraryPanels.Get("/:uid/dashboards/", middleware.ReqSignedIn, api.Wrap(lps.getConnectedDashboardsHandler))
 		libraryPanels.Patch("/:uid", middleware.ReqSignedIn, binding.Bind(patchLibraryPanelCommand{}), api.Wrap(lps.patchHandler))
 	})
 }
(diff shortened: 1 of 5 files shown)


pkg/services/librarypanels/api.go
```diff
@@ -74,6 +105,19 @@ func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response
 	return api.JSON(200, util.DynMap{"result": libraryPanels})
 }
 
+// getConnectedDashboardsHandler handles GET /api/library-panels/:uid/dashboard/.
+func (lps *LibraryPanelService) getConnectedDashboardsHandler(c *models.ReqContext) api.Response {
+	dashboardIds, err := lps.getConnectedDashboards(c, c.Params(":uid"))
```
aknuds1, line 110:
```suggestion
	dashboardIDs, err := lps.getConnectedDashboards(c, c.Params(":uid"))
```

pkg/services/librarypanels/database.go
```diff
@@ -40,6 +40,32 @@ func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd cre
 	return libraryPanel, err
 }
 
+// connectDashboard adds a connection between a Library Panel and a Dashboard.
+func (lps *LibraryPanelService) connectDashboard(c *models.ReqContext, uid string, dashboardId int64) error {
```
aknuds1, line 44:
```suggestion
func (lps *LibraryPanelService) connectDashboard(c *models.ReqContext, uid string, dashboardID int64) error {
```

pkg/services/librarypanels/database.go
```diff
@@ -59,6 +85,29 @@ func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, uid str
 	})
 }
 
+// disconnectDashboard deletes a connection between a Library Panel and a Dashboard.
+func (lps *LibraryPanelService) disconnectDashboard(c *models.ReqContext, uid string, dashboardId int64) error {
```
aknuds1, line 91:
```suggestion
func (lps *LibraryPanelService) disconnectDashboard(c *models.ReqContext, uid string, dashboardID int64) error {
```

pkg/services/librarypanels/database.go
```diff
@@ -105,6 +154,33 @@ func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]Lib
 	return libraryPanels, err
 }
 
+// getConnectedDashboards gets all dashboards connected to a Library Panel.
+func (lps *LibraryPanelService) getConnectedDashboards(c *models.ReqContext, uid string) ([]int64, error) {
+	connectedDashboardIds := make([]int64, 0)
```
aknuds1, line 161:
```suggestion
	connectedDashboardIDs := make([]int64, 0)
```

aknuds1, line 161: Nit, but I don't think there's any need to create the array, only to declare it.
```suggestion
	var connectedDashboardIds []int64
```

pkg/services/librarypanels/database.go
```diff
@@ -105,6 +154,33 @@ func (lps *LibraryPanelService) getAllLibraryPanels(c *models.ReqContext) ([]Lib
...
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		panel, err := getLibraryPanel(session, uid, c.SignedInUser.OrgId)
+		if err != nil {
+			return err
+		}
+
+		libraryPanelDashboards := make([]LibraryPanelDashboard, 0)
```
aknuds1, line 168: Nit, but I think there's no need to create the array, only to declare it.
```suggestion
		var libraryPanelDashboards []LibraryPanelDashboard
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -27,6 +27,33 @@ func TestCreateLibraryPanel(t *testing.T) {
 		})
 }
 
+func TestConnectLibraryPanel(t *testing.T) {
+	testScenario(t, "When an admin tries to create a connection for a library panel that does not exists, it should fail",
```
aknuds1, line 31:
```suggestion
	testScenario(t, "When an admin tries to create a connection for a library panel that does not exist, it should fail",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -67,6 +94,47 @@ func TestDeleteLibraryPanel(t *testing.T) {
 		})
 }
 
+func TestDisconnectLibraryPanel(t *testing.T) {
+	testScenario(t, "When an admin tries to disconnect a connection with a library panel that does not exist, it should fail",
```
aknuds1, line 98:
```suggestion
	testScenario(t, "When an admin tries to remove a connection with a library panel that does not exist, it should fail",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -67,6 +94,47 @@ func TestDeleteLibraryPanel(t *testing.T) {
...
+		func(t *testing.T, sc scenarioContext) {
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": "unknown", "dashboardId": "1"})
+			response := sc.service.disconnectHandler(sc.reqContext)
+			require.Equal(t, 404, response.Status())
+		})
+
+	testScenario(t, "When an admin tries to disconnect a connection that does not exist, it should fail",
```
aknuds1, line 105:
```suggestion
	testScenario(t, "When an admin tries to remove a connection that does not exist, it should fail",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -67,6 +94,47 @@ func TestDeleteLibraryPanel(t *testing.T) {
...
+			var result libraryPanelResult
+			err := json.Unmarshal(response.Body(), &result)
+			require.NoError(t, err)
+
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": result.Result.UID, "dashboardId": "1"})
+			response = sc.service.disconnectHandler(sc.reqContext)
+			require.Equal(t, 404, response.Status())
+		})
+
+	testScenario(t, "When an admin tries to disconnect a connection that does exists, it should succeed",
```
aknuds1, line 120:
```suggestion
	testScenario(t, "When an admin tries to remove a connection that does exist, it should succeed",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,6 +246,64 @@ func TestGetAllLibraryPanels(t *testing.T) {
 		})
 }
 
+func TestGetConnectedDashboards(t *testing.T) {
+	testScenario(t, "When an admin tries to get connected dashboards for a library panel that does not exists, it should fail",
```
aknuds1, line 250:
```suggestion
	testScenario(t, "When an admin tries to get connected dashboards for a library panel that does not exist, it should fail",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,6 +246,64 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+		func(t *testing.T, sc scenarioContext) {
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": "unknown"})
+			response := sc.service.getConnectedDashboardsHandler(sc.reqContext)
+			require.Equal(t, 404, response.Status())
+		})
+
+	testScenario(t, "When an admin tries to get connected dashboards for a library panel that does exists but has no connections, it should return none",
```
aknuds1, line 257:
```suggestion
	testScenario(t, "When an admin tries to get connected dashboards for a library panel that exists, but has no connections, it should return none",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -178,6 +246,64 @@ func TestGetAllLibraryPanels(t *testing.T) {
...
+			response = sc.service.getConnectedDashboardsHandler(sc.reqContext)
+			require.Equal(t, 200, response.Status())
+
+			var dashResult libraryPanelDashboardsResult
+			err = json.Unmarshal(response.Body(), &dashResult)
+			require.NoError(t, err)
+			require.Equal(t, 0, len(dashResult.Result))
+		})
+
+	testScenario(t, "When an admin tries to get connected dashboards for a library panel that does exists and has connections, it should return connected dashboard ids",
```
aknuds1, line 277:
```suggestion
	testScenario(t, "When an admin tries to get connected dashboards for a library panel that exists and has connections, it should return connected dashboard IDs",
```

aknuds1 (changes requested):
Found typos.

pkg/services/librarypanels/database.go
hugohaggmark: Here I want the api to return an empty array instead of "null", so using `make` is correct in this case.

pkg/services/librarypanels/api.go

pkg/services/librarypanels/database.go

pkg/services/librarypanels/database.go

pkg/services/librarypanels/database.go

pkg/services/librarypanels/database.go
hugohaggmark: Great catch, thank you, and I've fixed it now!

pkg/services/librarypanels/database.go
aknuds1: A nil array should be treated the same as an empty one in Go. Can you see any practical difference?

hugohaggmark: Yes, json.Marshal will treat a nil array as "null" value and I don't want that behavior here.

aknuds1: Aha, I was wondering if there might be a difference there.

pkg/services/librarypanels/librarypanels.go
```diff
@@ -67,4 +67,21 @@ func (lps *LibraryPanelService) AddMigration(mg *migrator.Migrator) {
...
 	mg.AddMigration("add index library_panel org_id & folder_id & name", migrator.NewAddIndexMigration(libraryPanelV1, libraryPanelV1.Indices[0]))
+
+	libraryPanelDashboardV1 := migrator.Table{
+		Name: "library_panel_dashboard",
+		Columns: []*migrator.Column{
+			{Name: "id", Type: migrator.DB_BigInt, IsPrimaryKey: true, IsAutoIncrement: true},
+			{Name: "librarypanel_id", Type: migrator.DB_BigInt, Nullable: false},
+			{Name: "dashboard_id", Type: migrator.DB_BigInt, Nullable: false},
+			{Name: "created", Type: migrator.DB_DateTime, Nullable: false},
+			{Name: "created_by", Type: migrator.DB_BigInt, Nullable: false},
```
aknuds1, line 78: Shouldn't created_by be marked as a foreign key?

aknuds1 (commented):
I'm wondering if created_by shouldn't be marked as a foreign key in the database, to ensure integrity.

pkg/services/librarypanels/librarypanels.go
hugohaggmark: Why, I don't think I understand can you elaborate?

hugohaggmark: Do you mean a foreign key to the `user` table?

aknuds1: created_by refers to a user ID, so shouldn't it be a foreign key relation (FK pointing from library_panel_dashboard to the user table)?

aknuds1: Come to think of it, shouldn't librarypanel_id and dashboard_id also be FKs?

hugohaggmark: I agree that librarypanel_id and dashboard_id should be foreign keys. Do you know if we have an example of foreign keys usage in Grafana?

hugohaggmark: But I think `created_by` doesn't make sense as we're using the signed in users id. If that is wrong, then something else is seriously wrong, don't you think?

aknuds1: I don't know directly how to make foreign key migrations in Grafana since I haven't had the need yet, but I would search the code to look for examples. There ought to be something.

About `created_by`, what's the concrete argument for not making it a foreign key referring to the user table (I'm not sure off the top of my head what is the exact table's name)? I don't follow you here.

hugohaggmark: My argument was strictly from an api usage scenario, but if someone adds rows in the database manually then of course it makes sense, I'll add it as well. And I'll also add foreign keys to `folder_id`, `org_id`, `updated_by` and `created_by` in the `library_panel` table as well. The challenge I've is that I can't seem to find any foreign key usages in Grafana.

aknuds1: @marefr @papagian we should use foreign key constraints for the aforementioned fields here, right?

aknuds1: @hugohaggmark I don't really see what the FK has to do with it being used from the API or not. When for example a user is removed from the database, it should be detected that any corresponding `created_by` references are stale.

hugohaggmark: I agree with you @aknuds1, but I can't see it happening using xorm and I can't see any examples throughout our codebase. Do you think we can add this when we know how to in a follow up PR instead?

marefr: We don't use foreign keys for historical and technical reasons. Not something we should add in this PR ass Hugo suggest, but potentially something to investigate on its own

aknuds1: @marefr Thanks for clarifying that, I've been able to dig up a [comment](https://github.com/grafana/grafana/issues/3269#issuecomment-383328548) by @torkelo that sheds light on why foreign key constraints aren't used in Grafana. I agree that there's no point in asking for FKs in this PR considering they're not used in Grafana.

pkg/services/librarypanels/api.go
```diff
@@ -18,9 +18,12 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
 
 	lps.RouteRegister.Group("/api/library-panels", func(libraryPanels routing.RouteRegister) {
 		libraryPanels.Post("/", middleware.ReqSignedIn, binding.Bind(createLibraryPanelCommand{}), api.Wrap(lps.createHandler))
+		libraryPanels.Post("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.connectHandler))
```
aknuds1, line 21: I think the dashboards part of the path should be called "dashboards", since its a collection of resources, for example the endpoint /api/library-panels/dashboard/ will returns a collection of dashboards. It's also consistent with /api/library-panels.

```suggestion
		libraryPanels.Post("/:uid/dashboards/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.connectHandler))
```

pkg/services/librarypanels/api.go
```diff
@@ -18,9 +18,12 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
...
 		libraryPanels.Delete("/:uid", middleware.ReqSignedIn, api.Wrap(lps.deleteHandler))
+		libraryPanels.Delete("/:uid/dashboard/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.disconnectHandler))
```
aknuds1, line 23:
```suggestion
		libraryPanels.Delete("/:uid/dashboards/:dashboardId", middleware.ReqSignedIn, api.Wrap(lps.disconnectHandler))
```

pkg/services/librarypanels/api.go
```diff
@@ -18,9 +18,12 @@ func (lps *LibraryPanelService) registerAPIEndpoints() {
...
 		libraryPanels.Get("/", middleware.ReqSignedIn, api.Wrap(lps.getAllHandler))
 		libraryPanels.Get("/:uid", middleware.ReqSignedIn, api.Wrap(lps.getHandler))
+		libraryPanels.Get("/:uid/dashboard/", middleware.ReqSignedIn, api.Wrap(lps.getConnectedDashboardsHandler))
```
aknuds1, line 26:
```suggestion
		libraryPanels.Get("/:uid/dashboards/", middleware.ReqSignedIn, api.Wrap(lps.getConnectedDashboardsHandler))
```

pkg/services/librarypanels/api.go
```diff
@@ -38,7 +41,19 @@ func (lps *LibraryPanelService) createHandler(c *models.ReqContext, cmd createLi
 	return api.JSON(200, util.DynMap{"result": panel})
 }
 
-// deleteHandler handles DELETE /api/library-panels/:uid
+// connectHandler handles POST /api/library-panels/:uid/dashboard/:dashboardId.
```
aknuds1, line 44:
```suggestion
// connectHandler handles POST /api/library-panels/:uid/dashboards/:dashboardId.
```

pkg/services/librarypanels/api.go
```diff
@@ -51,7 +66,23 @@ func (lps *LibraryPanelService) deleteHandler(c *models.ReqContext) api.Response
 	return api.Success("Library panel deleted")
 }
 
-// getHandler handles GET /api/library-panels/:uid
+// disconnectHandler handles DELETE /api/library-panels/:uid/dashboard/:dashboardId.
```
aknuds1, line 69:
```suggestion
// disconnectHandler handles DELETE /api/library-panels/:uid/dashboards/:dashboardId.
```

pkg/services/librarypanels/api.go
```diff
@@ -74,6 +105,19 @@ func (lps *LibraryPanelService) getAllHandler(c *models.ReqContext) api.Response
 	return api.JSON(200, util.DynMap{"result": libraryPanels})
 }
 
+// getConnectedDashboardsHandler handles GET /api/library-panels/:uid/dashboard/.
```
aknuds1, line 108:
```suggestion
// getConnectedDashboardsHandler handles GET /api/library-panels/:uid/dashboards/.
```

pkg/services/librarypanels/models.go
```diff
@@ -22,11 +22,24 @@ type LibraryPanel struct {
 	UpdatedBy int64
 }
 
+// LibraryPanelDashboard is the model for library panel connections.
+type LibraryPanelDashboard struct {
```
aknuds1, line 26: I think this type can be un-exported.
```suggestion
// libraryPanelDashboard is the model for library panel connections.
type libraryPanelDashboard struct {
```

aknuds1 (approved):
Generally looks good now, but I think the path component should be "dashboards" instead of "dashboard", to be consistent and because it represents a collection of dashboards, not a single one.

hugohaggmark:
> Generally looks good now, but I think the path component should be "dashboards" instead of "dashboard", to be consistent and because it represents a collection of dashboards, not a single one.

Thanks for all the great feedback, and I agree that the path component is better this way. Thank you!

pkg/services/librarypanels/api.go

pkg/services/librarypanels/api.go

pkg/services/librarypanels/api.go

pkg/services/librarypanels/api.go

pkg/services/librarypanels/api.go

pkg/services/librarypanels/api.go

pkg/services/librarypanels/models.go
hugohaggmark: Fixed

PR #30214, opened 2021-01-12
Field overrides: extracting the field config factory into its own reusable module.

mckn:
When using the GraphNG in explore we need to recreate the field config registry in Explore to be able to use the `applyFieldOverrides` helper functionality to apply the proper state onto the `DataFrame[]` data set.

I'm not sure we want to expose this from the `@grafana/data` but this PR is used as a reference of discussion.

The idea would then be to do something like the following when applying the field config overrides when processing the data set in Explore:

```ts
// imagine that the data is the incoming data set.
const registry = createFieldConfigRegistry(getGraphFieldConfig(defaultGraphConfig), 'Explore');
const applied = applyFieldOverrides({
        fieldConfig: configWithOverrides,
        data,
        timeZone,
        replaceVariables: value => value, // TODO: replace with real replace
        theme,
        fieldConfigRegistry: registry,
});

```

### DIFF
--- a/packages/grafana-data/src/panel/registryFactories.ts
+++ b/packages/grafana-data/src/panel/registryFactories.ts
@@ -1,0 +1,58 @@
+import { FieldConfigOptionsRegistry } from '../field/FieldConfigOptionsRegistry';
+import { standardFieldConfigEditorRegistry } from '../field/standardFieldConfigEditorRegistry';
+import { FieldConfigProperty } from '../types/fieldOverrides';
+import { FieldConfigEditorBuilder } from '../utils/OptionsUIBuilders';
+import { SetFieldConfigOptionsArgs } from './PanelPlugin';
+
+export function createFieldConfigRegistry<TFieldConfigOptions>(
+  config: SetFieldConfigOptionsArgs<TFieldConfigOptions> = {},
+  pluginName: string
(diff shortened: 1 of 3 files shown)


packages/grafana-data/src/panel/registryFactories.ts
```diff
@@ -0,0 +1,58 @@
+import { FieldConfigOptionsRegistry } from '../field/FieldConfigOptionsRegistry';
+import { standardFieldConfigEditorRegistry } from '../field/standardFieldConfigEditorRegistry';
+import { FieldConfigProperty } from '../types/fieldOverrides';
+import { FieldConfigEditorBuilder } from '../utils/OptionsUIBuilders';
+import { SetFieldConfigOptionsArgs } from './PanelPlugin';
+
+export function createFieldConfigRegistry<TFieldConfigOptions>(
```
torkelo, line 7: Shold this have jsdoc state tag? something like 

/** @beta */

torkelo (approved):
Looks good

packages/grafana-data/src/panel/registryFactories.ts
mckn: Yupp, just wanted to get your input first.

PR #30215, opened 2021-01-12
PanelOptionsGroups: Only restore styles from PanelOptionsGroup

torkelo:
I removed the PanelOptionsGroup component but the styles where still in use by angular panels so adding the styles only back

#30157


hugohaggmark (approved):
LGTM!

PR #30223, opened 2021-01-12
AlertingNG: Save alert instances

papagian:
Co-authored-by: Kyle Brandt <kyle@grafana.com>

Follow up of work in #29276

Fixes #

Requires postgres >= 9.5 (ON CONFLICT Clause)


pkg/services/ngalert/instance.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// AlertInstance represents a single alert instance.
+type AlertInstance struct {
+	AlertDefinitionOrgID int64  `xorm:"org_id"`
+	AlertDefinitionUID   string `xorm:"alert_definition_uid"`
```
kylebrandt, line 11: Maybe Since these are a property of the go `AlertInstance` and the `alert_instance` table. We can Drop the `Alert` prefix from these properties in both Go and the db. So just `DefinitionOrgID` and `DefinitionUID`

pkg/services/ngalert/schedule.go
```diff
@@ -50,11 +50,17 @@ func (ng *AlertNG) definitionRoutine(grafanaCtx context.Context, key alertDefini
 				results, err := eval.ConditionEval(&condition, ctx.now)
 				end = timeNow()
 				if err != nil {
+					// consider saving alert instance on error
 					ng.schedule.log.Error("failed to evaluate alert definition", "key", key, "attempt", attempt, "now", ctx.now, "duration", end.Sub(start), "error", err)
 					return err
 				}
 				for _, r := range results {
 					ng.schedule.log.Info("alert definition result", "key", key, "attempt", attempt, "now", ctx.now, "duration", end.Sub(start), "instance", r.Instance, "state", r.State.String())
```
kylebrandt, line 58: Now that we save these, I think this can change to Debug so as not to add to many line to the log.

Could collect summary for all results within a run, and log a single info line summary (X instances, N alerting, O Normal).

pkg/services/sqlstore/migrator/dialect.go
```diff
@@ -34,6 +34,8 @@ type Dialect interface {
 
 	IndexCheckSQL(tableName, indexName string) (string, []interface{})
 	ColumnCheckSQL(tableName, columnName string) (string, []interface{})
+	//UpsertSQL returns the upsert sql statement for a dialect
```
kylebrandt, line 37:
```suggestion
	// UpsertSQL returns the upsert sql statement for a dialect.
```

pkg/services/sqlstore/migrator/mysql_dialect.go
```diff
@@ -199,3 +199,28 @@ func (db *MySQLDialect) ErrorMessage(err error) string {
 func (db *MySQLDialect) IsDeadlock(err error) bool {
 	return db.isThisError(err, mysqlerr.ER_LOCK_DEADLOCK)
 }
+
+//UpsertSQL returns the upsert sql statement for PostgreSQL dialect
```
kylebrandt, line 203:
```suggestion
// UpsertSQL returns the upsert sql statement for PostgreSQL dialect.
```

pkg/services/sqlstore/migrator/postgres_dialect.go
```diff
@@ -215,3 +215,40 @@ func (db *PostgresDialect) PostInsertId(table string, sess *xorm.Session) error
 	}
 	return nil
 }
+
+//UpsertSQL returns the upsert sql statement for PostgreSQL dialect
```
kylebrandt, line 219:
```suggestion
// UpsertSQL returns the upsert sql statement for PostgreSQL dialect.
```

pkg/services/sqlstore/migrator/sqlite_dialect.go
```diff
@@ -147,3 +148,40 @@ func (db *SQLite3) IsUniqueConstraintViolation(err error) bool {
 func (db *SQLite3) IsDeadlock(err error) bool {
 	return false // No deadlock
 }
+
+//UpsertSQL returns the upsert sql statement for SQLite dialect
```
kylebrandt, line 152:
```suggestion
// UpsertSQL returns the upsert sql statement for SQLite dialect.
```

kylebrandt (approved):
Works! :tada: 

Couple naming, logging changes suggested.

We will want the ability to filter and having different listing scenarios, but can be in future PR.

pkg/services/ngalert/schedule.go
kylebrandt: Also might be nice to have definition name in log as well (as UIDs are hard to read/keep track of)

pkg/services/ngalert/database_mig.go
```diff
@@ -68,3 +68,28 @@ func addAlertDefinitionVersionMigrations(mg *migrator.Migrator) {
...
+			{Name: "labels", Type: migrator.DB_Text, Nullable: false},
+			{Name: "labels_hash", Type: migrator.DB_NVarchar, Length: 190, Nullable: false},
+			{Name: "current_state", Type: migrator.DB_NVarchar, Length: 190, Nullable: false},
+			{Name: "current_state_since", Type: migrator.DB_BigInt, Nullable: false},
+			{Name: "last_eval_time", Type: migrator.DB_BigInt, Nullable: false},
+		},
+		PrimaryKeys: []string{"definition_org_id", "definition_uid", "labels_hash"},
+		Indices: []*migrator.Index{
+			{Cols: []string{"definition_org_id", "definition_uid", "current_state"}, Type: migrator.IndexType},
+			{Cols: []string{"definition_org_id", "current_state"}, Type: migrator.IndexType},
```
papagian, line 87: @kylebrandt I've added this index since I think it's required by [listAlertInstances](https://github.com/grafana/grafana/blob/6f371b644b92ec6f00f12f78cf09c14545ab2cd6/pkg/services/ngalert/instance_database.go#L48) if the definition UID is missing by the command.

pkg/services/ngalert/database_mig.go
```diff
@@ -68,3 +68,28 @@ func addAlertDefinitionVersionMigrations(mg *migrator.Migrator) {
...
+		},
+	}
+
+	// create table
+	mg.AddMigration("create alert_instance table", migrator.NewAddTableMigration(alertInstance))
+	mg.AddMigration("add index in alert_instance table on definition_org_id, definition_uid and current_state columns", migrator.NewAddIndexMigration(alertInstance, alertInstance.Indices[0]))
```
papagian, line 93: This causes MySQL to fail with:
`common_test.go:26: Failed to init test database: migration failed: Error 1059: Identifier name 'IDX_alert_instance_definition_org_id_definition_uid_current_state' is too long`
the index name [is derived ](https://github.com/grafana/grafana/blob/4dd7b7a82d476a7a0c29b30a8fce5b45a4c29cff/pkg/services/sqlstore/migrator/types.go#L55) from the columns of the index therefore I have to rename the columns to something shorter.

papagian:
I've committed two additional modifications:
- delete alert instances when deleting an alert definition
- include alert definition title in the alert instance listing api response

@kylebrandt would you like to check once again, thanks!

kylebrandt (approved):
Looks good to me, instances seem to delete as expected with the defintion.

Outside of this PR, but I noticed we allow different alert defs to have the same name, so maybe title + orgId should be a unique index. Or at least maybe create / update def doesn't allow:
 - Duplicate title to be created
 - empty title to be created

PR #30227, opened 2021-01-12
runRequest: Fixes issue with request time range and time range returned to panels are off causing data points to be cut off (outside)

torkelo:
We have a problem with runRequest in that in always interpolates the time range after a response has been received.

This causes the time range that is returned in the request and later past to panels to be off from what the actual request from & to where (by the time the query took)

So if query takes 50ms the from time is 50ms later than the first data point, and this causes it to not be renderd as it outside the time range bounds.

This is a very simple initial fix, might need a more complex solution.

### DIFF
--- a/public/app/features/query/state/runRequest.ts
+++ b/public/app/features/query/state/runRequest.ts
@@ -17,6 +17,7 @@ import {
   guessFieldTypes,
   LoadingState,
   PanelData,
+  TimeRange,
   toDataFrame,
 } from '@grafana/data';
 import { toDataQueryError } from '@grafana/runtime';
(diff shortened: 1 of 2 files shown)


public/app/features/query/state/runRequest.ts
```diff
@@ -90,6 +83,20 @@ export function processResponsePacket(packet: DataQueryResponse, state: RunningQ
...
+  if (!isString(range.raw.from) || loadingState !== LoadingState.Streaming) {
+    return range;
+  }
+
+  return {
+    ...range,
+    from: dateMath.parse(range.raw.from, false)!,
+    to: dateMath.parse(range.raw.to, true)!,
+  };
+}
```
torkelo, line 98: So this changes logic to always use the request range as originally passed in, and leaves the time range update & re-evaluation to only on streaming

PR #30231, opened 2021-01-12
Chore: Better logging of plugin scanning errors

aknuds1:
Improve logging of plugin scanning errors. Found this necessary when plugin scanning failed due to a stale symlink.


papagian (approved):
LGTM!

PR #30238, opened 2021-01-12
WIP/GraphNG: support soft min max axis

### LINKED ISSUE #979: Graph: Limits on Y auto scale (min max)
I've got a graph displaying error rates, and I've set the axis max to `auto`.  The problem is that on quiet days, with a low error rate, the axis shrinks down to display tiny blips as large mountains.  I could set the max to a particular vaule. But, I don't want to aim to low and experience a graph that goes off the chart. And, I don't want to set the max so high as to not get readable "medium" graphs.  (I don't want to use thresholds, as I've got graphs on the other axis, and I don't want to cause confusion with those ranges.)

An ideal solution would to have a "min-max" setting.  I could set the min-max to 100, and the axis would never get smaller than that.  But, it would grow if need be.

ryantxu:
Fixes #979

Status:  this is currently just the settings scaffolding and does not implement the actual scaling

This adds an axis mode option:
image

TODO:
- [ ] words!  what language makes sense?
- [ ] implement the auto scalling

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,7 +17,7 @@ steps:
   image: grafana/build-container:1.3.1
   commands:
   - mkdir -p bin
-  - curl -fL -o bin/grabpl https://grafana-downloads.storage.googleapis.com/grafana-build-pipeline/v0.5.32/grabpl
+  - curl -fL -o bin/grabpl https://grafana-downloads.storage.googleapis.com/grafana-build-pipeline/v0.5.33/grabpl
   - chmod +x bin/grabpl
   - ./bin/grabpl verify-drone
   - curl -fLO https://github.com/jwilder/dockerize/releases/download/v$${DOCKERIZE_VERSION}/dockerize-linux-amd64-v$${DOCKERIZE_VERSION}.tar.gz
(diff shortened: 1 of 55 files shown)


leeoniya:
i can land the remainder by the end of this week.

leeoniya:
another option supports both hard limits (clamping) and soft limits by using the Min/Max from "Standard options" for hard, and adds `Axis > SoftMin` and `Axis > SoftMax` numeric fields to the "Time series options" for the soft limits.

The soft limits will be inside the hard limits. The scale/axis range will start at the soft limits and expand to the hard limits when the data exceeds the softs, and will clip afterwards.

image

torkelo:
> should the JSON store the full config (rather than a unary value), so we have more flexibility to expose more subtle behavior in the future?

I am not sure that is a good idea, then we have to support that more complex model forever. The simpler the persisted model, the easier for us to change this & improve defaults.

torkelo:
I like this `Axis min & max mode` approach, only one new setting needed.

leeoniya:
@torkelo please read the recent comments onwards from https://github.com/grafana/grafana/issues/979#issuecomment-759187310

i think they're valid concerns, and the there's quite a lot of added flexibility gained from having 2 numeric fields instead of a toggle that neuters/repurposes clamping ability. i'm not sure of another way to accomodate spike-squashing in a softMin/max scenario is you cannot set hard limits as well.

torkelo:
@leeoniya yea, maybe your right that separate soft min & max options would allow hard limits. Wish we could hide some of these advanced options so they are not always expanded/visible (to not overwhelm new users)

dprokop:
> @leeoniya yea, maybe your right that separate soft min & max options would allow hard limits. Wish we could hide some of these advanced options so they are not always expanded/visible (to not overwhelm new users)

@torkelo I was thinking along the same lines about this. I assume when we talk about the advanced options we consider the soft min/max numeric fields. I find the pattern we use for the axis scales something that could fit this use case as well when the numeric inputs are only shown when the soft mode is selected:

image

torkelo:
@dprokop but that would still require an always present soft mode option, and what would soft mode mean if you have not specified any values in the inputs? feels better to have soft min & soft max inputs but hide them & axis width, (and other future advanced options) in an advanced section or something :)

leeoniya:
superseded by #30326.

PR #30239, opened 2021-01-12
feat(Loki):  loki label_values() function with left hand subquery component

### LINKED ISSUE #25205: Loki: label_values() - Use series API instead of labels API

Ability to filter variables in loki using other variables. 

Ie in prometheus i can use 
```
label_values(disk_used_percent{host=~"$server"},path)
```
Where $server is
```
label_values(cpu_usage_idle, host)
```
To get list of mountpoints on this specific server. 

Will be nice to have same feature for loki, for example i have promtail collecting systemd journal on all machines, and needless to say they have different services running. Using label_values(unit) lists tons of services and it's not easy to understand how they correlate to specific servers.

Looks like loki already have matching functionality built-in, providing only units running on that node in case i already specified host:

image

**Why is this needed**:
Much cleaner selectors in dashboards and overall usability improvement

secustor:
At the moment it is not possible to run nested queries in Grafana variables for Loki     

e.g. `label_values({compose_service=~$service, compose_project=~$project}, container_name)`

Fixes #25205

I need some direction regarding topics like:
-  how to get  specific data like the other defined variables for later interpolation   
- do I need a start time if the range has been provided as both are required
- the same goes for `interval` and `intervalMs`

### DIFF
--- a/public/app/plugins/datasource/loki/datasource.test.ts
+++ b/public/app/plugins/datasource/loki/datasource.test.ts
@@ -542,6 +542,40 @@ describe('LokiDatasource', () => {
         expect(res).toEqual([{ text: 'label1' }]);
       });
     });
+
+    mocks.forEach((mock, index) => {
+      it(`should return label values for subquery for Loki v${index}`, async () => {
+        const { ds } = getTestContext(mock);
+
+        const res = await ds.metricFindQuery('label_values({app="foo", namespace="prod"},instance)');
+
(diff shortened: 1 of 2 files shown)


sherifkayad:
@secustor do you plan to work on this further? It's definitely a great thing that we want badly

secustor:
I would look into it again, if I get the directions mentioned in the PR

PR #30242, opened 2021-01-13
DashboardSchemas: OpenAPI Schema Generation

trotttrotttrott:

Adds a Go program for creating OpenAPI schema JSON. Prior to this, we anticipated using `cue export`. The problem is you cannot configure the openapi encoder to expand references from the command line. This results in references to unexported schemas.

<details>
<summary>Details and examples...</summary>

```
▸ cue export --out openapi -o - ./panels/ | grep ref
                                        "$ref": "#/components/schemas/_thresholds"
                                            "$ref": "#/components/schemas/_mapping"
                                            "$ref": "#/components/schemas/_dataLink"
                                    "$ref": "#/components/schemas/_override"
                        "$ref": "#/components/schemas/_panel"
                                    "$ref": "#/components/schemas/_override"
                                    "$ref": "#/components/schemas/_dataLink"
                        "$ref": "#/components/schemas/_thresholds"
                        "$ref": "#/components/schemas/_panel"
                        "$ref": "#/components/schemas/_gridPos"
```

There's no way to tell Cue to expand the above references from the command line. This makes the document incomplete and would likely make anything consuming it fail.

```
▸ go run . ./panels
{"Gauge":{"description":"Gauge is a...
```

The Go program expands references by configuring the encoder to do so. Also, it only outputs the schemas since they're the only part of an OpenAPI document that's relevant here.

</details>

In addition, this fixes number syntax in many Cue files. This resolves error, `unsupported op for number &`, when loading from Go.


aknuds1 (approved):

aknuds1:
Is this practically interesting to Grafana users?

trotttrotttrott:
> Is this practically interesting to the Grafana community?

I see this as a step toward practical interests...

One being dashboards as code library generation with [grafana/dashboard-spec](https://github.com/grafana/dashboard-spec). This project maintains a separate collection of schemas that should to be retired. Getting the Cue -> OpenAPI piece worked out is a prerequisite. That is unless we rewrite the generator to work with Cue directly, but it's uncertain if that's even worthwhile.

Another interest is to facilitate exploration of generating TypeScript from Cue. We're not certain of how this should be done yet. There's already tools that can convert OpenAPI to Typescript, but no tools that I'm aware of that generate TypeScript directly from Cue. We'd like to at least evaluate our OpenAPI options. I think there's also a chance we end up developing our own Typescript generator in which case it may be what this Go program becomes. Anything extraneous would be removed.

aknuds1:
Sounds awesome :)

PR #30243, opened 2021-01-13
Docs: cloudwatch filter should be JSON format

tomowang:

Document example was not working

Fixes #

Example `ec2_instance_attribute(default, InstanceId, { 'tag:Environment': ['production'] });` will cause `SyntaxError: Unexpected token ' in JSON at position 2`


oddlittlebird (approved):
Thanks for helping improve Grafana documentation!

PR #30245, opened 2021-01-13
Metrics: Stream response

aknuds1:
This PR modifies the /api/ds/query endpoint to stream the JSON response, instead of encoding it all into memory before writing it back to the client. This _should_ improve memory consumption, although I wasn't able to measure a big improvement.

I discovered this potential for optimization while investigating Azure Data Explorer performance.

As part of the PR I also move the different response types into pkg/api/response.go, for easier navigation since I add another type.

### DIFF
--- a/pkg/api/common.go
+++ b/pkg/api/common.go
@@ -16,6 +16,7 @@ var (
 )
 
 type Response interface {
+	// WriteTo writes to the provided context.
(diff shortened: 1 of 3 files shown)


aknuds1:
Closing for now, since from additional research, it looks like there's no performance improvement. Will look into other libraries instead.

PR #30250, opened 2021-01-13
Metrics API: Use jsoniter for JSON encoding

aknuds1:
For the /api/ds/query endpoint, use [jsoniter](https://github.com/json-iterator/go) library for JSON encoding, for performance gains. I haven't seen huge gains in the actual request processing time, but profiling indicates a 3x-ish speedup. Not sure why that's not visible in terms of wall time, we should investigate.

I've been told that Prometheus and Cortex use jsoniter already.

## CPU profile jsoniter
pprof-jsoniter

## CPU profile stdlib
pprof-stdlib


papagian (approved):
I haven't seen a big improvement in request processing time either but I think that this depends on the amount of the response data so maybe the query I have used was not the best. However I have tested and it works.

PR #30251, opened 2021-01-13
GraphNG: stats in legend

dprokop:
Draft for early feedback and ideas.

The problem is trickier that I innitially assumed. In current implementation, the legend items are built together with the UPlotConfigBuilder. This means they are only built, when the config is invalidated. This creates an issue, when there is data-only update. Basically the calc values are not being updated.

To mitigate this I have moved the legend items creation outside of the memo function that memoizes the UPlotConfigBuilder. I have also extended the builders addSeries interface to transpoirt some of the information that's required for the legend to be built :

```ts
// The following properties are not used in the uPlot config, but are utilized as transport for legend config
dataFrameFieldIndex,
fieldName: getFieldDisplayName(field, alignedFrame),
hideInLegend: customConfig.hideFrom?.legend,
```

These are not used to construct uPlot config, but to build the legend only.

I have also changed the interface of VizLegendItem from 

```
export interface VizLegendItem {
  label: string;
  color: string;
  yAxis: number;
  disabled?: boolean;
  displayValues?: DisplayValue[];
  fieldIndex?: DataFrameFieldIndex;
}
```
to 

```
export interface VizLegendItem {
  label: string;
  color: string;
  yAxis: number;
  disabled?: boolean;
  getDisplayValues?: () => DisplayValue[];
  fieldIndex?: DataFrameFieldIndex;
}
```

to enable reducing the field only when the legend is shown.

Unfortunately, this implementation suffers from a hack thas shown here https://github.com/grafana/grafana/compare/graph-ng-stats-legend?expand=1#diff-65862696b913d4fc89d7ba62cdf603a251ce3e5764c7c9cec5fa3b64e9e6b609R219 as the config vs data might not be in sync during the render when the legend is created.

Not sure how to solve it yet so any ideas are more than welcome <3


torkelo:
Great job! I see you hit similar problems I did in this PR https://github.com/grafana/grafana/pull/29893  (where color depends on data, and so also ran into an issue with legend items being only built on config change). So the way you fixing it here will unblock that PR :smile: 

I found a small issue with the table calcs are lowercase in legend table but upper case in the calc selection editor:
Screenshot from 2021-01-13 15-16-19

dprokop:
> So the way you fixing it here will unblock that PR 

Yeah, but I'm still on a fence whether or not the "fix" is a good way to go. I have concerns and don't fully understand the actual cause for the hack i need to use here https://github.com/grafana/grafana/pull/30251/files#diff-65862696b913d4fc89d7ba62cdf603a251ce3e5764c7c9cec5fa3b64e9e6b609R219 so I'm gonna give it a night to ferment.

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -212,16 +203,58 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
+
+      if (seriesConfig.hideInLegend || !fieldIndex) {
+        return undefined;
+      }
+
+      const field = data[fieldIndex.frameIndex]?.fields[fieldIndex.fieldIndex];
+
+      // I don't know what's going on... ;)
+      if (!field) {
+        return undefined;
```
torkelo, line 221: @dprokop I think I know why this happens, when the series count change you get one render here with new data, and old configBuilder.getSeries

dprokop: That's exactly what happens, I just can't think of any nicer way of handling this now. And this feels like something that might lead to confusing issues I believe. I mean, this pretty much stops anything wrong to happen as we are making sure the field exists, so undefined will not be referenced anywhere down the road.

dprokop:
>I found a small issue with the table calcs are lowercase in legend table but upper case in the calc selection editor:

@torkelo this is fixed now

torkelo (approved):

PR #30252, opened 2021-01-13
Grafana-UI: Add story/docs for FilterPill

Clarity-89:
 Add story and docs for FilterPill component.

### DIFF
--- a/packages/grafana-ui/src/components/FilterPill/FilterPill.story.tsx
+++ b/packages/grafana-ui/src/components/FilterPill/FilterPill.story.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { Story } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { FilterPill, FilterPillProps } from './FilterPill';
+import { withCenteredStory } from '@grafana/ui/src/utils/storybook/withCenteredStory';
+import mdx from './FilterPill.mdx';
+import { getAvailableIcons } from '../../types';
+
+export default {
+  title: 'General/FilterPill',
+  component: FilterPill,
+  decorators: [withCenteredStory],
+  argTypes: {
+    icon: { control: { type: 'select', options: getAvailableIcons() } },
+  },
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const Basic: Story<FilterPillProps> = args => {
+  return <FilterPill {...args} onClick={action('Pill clicked')} />;
+};
+
+Basic.args = {
+  selected: false,
+  label: 'Test',
+  icon: undefined,
+};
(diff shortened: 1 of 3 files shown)


packages/grafana-ui/src/components/FilterPill/FilterPill.story.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { withCenteredStory } from '@grafana/ui/src/utils/storybook/withCenteredStory';
+import mdx from './FilterPill.mdx';
+import { getAvailableIcons } from '../../types';
+
+export default {
+  title: 'General/FilterPill',
+  component: FilterPill,
+  decorators: [withCenteredStory],
+  argTypes: {
+    icon: { control: { type: 'select', options: getAvailableIcons() } },
```
jackw, line 14: Nitpick: do we have any preference over usage of `getAvailableIcons` and `iconOptions` found [here](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/utils/storybook/knobs.ts#L6-L14)? 

Button and Badge stories currently use `iconOptions` maybe whichever we settle on we put it somewhere for sharing across all stories?

packages/grafana-ui/src/components/FilterPill/FilterPill.story.tsx
```diff
@@ -0,0 +1,31 @@
...
+  },
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const Basic: Story<FilterPillProps> = args => {
+  return <FilterPill {...args} onClick={action('Pill clicked')} />;
```
jackw, line 24: Storybook has an alternative syntax for actions without importing:

```
argTypes: {
    onClick: { action: 'Pill Clicked' },
}
```

packages/grafana-ui/src/components/FilterPill/FilterPill.tsx
```diff
@@ -39,7 +39,6 @@ const getFilterPillStyles = stylesFactory((theme: GrafanaTheme, isSelected: bool
       padding: ${theme.spacing.xxs} ${theme.spacing.sm};
       background: ${theme.colors.bg2};
       border-radius: ${theme.border.radius.sm};
-      display: inline-block;
```
jackw, line 42: Could this cause some use cases to render each `FilterPill` on a new line or are they always wrapped in a `HorizontalGroup` or similar?

jackw (approved):
LGTM! 🚀

packages/grafana-ui/src/components/FilterPill/FilterPill.story.tsx
Clarity-89: Didn't know we had `iconOptions`. I don't have any preferences but in this case the `None` option doesn't work because the component always has an icon by default, even tho `None` is selected in the dropdown.

packages/grafana-ui/src/components/FilterPill/FilterPill.tsx
Clarity-89: This was a duplicate declaration, there's `display: flex` rule below, so I removed this one as it is overwritten anyway.

PR #30256, opened 2021-01-13
GraphNG: added support to change series color from legend.

mckn:
This adds support to change series color by pressing the icon in the legend. It will alter the field overrides and use that mechanism to change the series color. The benefit of doing this is that the configuration will be persisted (if saved) and will survive page reloads.

This is a first implementation of this functionality but I think it could be improved in a couple of ways.

Change the `onSeriesColorChange` callback signature to have dataFrameFieldIndex instead of label. That would be more stable in scenarios where series have the same name.


torkelo (commented):
Nice!!

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -33,6 +33,7 @@ export interface GraphNGProps extends Omit<PlotProps, 'data' | 'config'> {
   legend?: LegendOptions;
   fields?: XYFieldMatchers; // default will assume timeseries data
   onLegendClick?: (event: GraphNGLegendEvent) => void;
+  onSeriesColorChange?: (label: string, color: string) => void;
```
ryantxu, line 36: is `label: string` our best identifier?  I get a bit nervous about checking equality based on the display names that get changed through the override process :grimacing:

ryantxu (commented):
@mckn -- this looks great!  only concern is how we identify the field -- perhaps an event with both the text and raw field?

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
torkelo: Since this is get's saved as an override, I think this is our only identifier that will work across query refreshes.

torkelo (approved):
Works great! Code looks good!

dprokop (approved):
Looks great @mckn!

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
mckn: @ryantxu yeah, for now I think this is what we have to go with. But as I stated in the special notes above it would be better to do some refactoring and use the `dataFrameFieldIndex` as the identifier.

But as @torkelo highlighted, we probably still need to use the label in the fieldOverride to is will work across query refreshes.

PR #30262, opened 2021-01-13
Decimals: Improving auto decimals logic for high numbers and scaled units 

### LINKED ISSUE #23561: Stat/Gauge: Displaying wrong number of decimals for even thousands, millions, billions
**What happened**:

Currently `ValueFormatter` is not respecting `scaledDecimals`, producing variable-decimal results:

```js
getValueFormat('short')(1, 0, 1)           // => 1
getValueFormat('short')(1000, 0, 1)        // => 1.0000 K
getValueFormat('short')(1000000, 0, 1)     // => 1.0000000 Mil
getValueFormat('short')(1000000000, 0, 1)  // => 1.0000000000 Bil
```

Consider some integer metrics like "Execution times", displaying with a decimal is not reasonable for such integer metrics when there is no scales: "Executed 10.0 times". However it is fine to keep a resonable decimal when the number is scaled, like "Executed 10.0 K times".

```js
getValueFormat('short')(1, 0, 1)           // => 1
getValueFormat('short')(1000, 0, 1)        // => 1.0 K
getValueFormat('short')(1000000, 0, 1)     // => 1.0 Mil
getValueFormat('short')(1000000000, 0, 1)  // => 1.0 Bil
```

The above can be reproduced in [valueFormats.test.ts](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/valueFormats/valueFormats.test.ts)

https://github.com/grafana/grafana/blob/61460ea3a211ef4655a1ef600da324fa84979102/packages/grafana-data/src/valueFormats/valueFormats.ts#L130

Maybe changing this line to the following can fix this bug:

```js
decimals = scaledDecimals;
```

**Environment**:
- Grafana version: master
- Data source type & version:
- OS Grafana is installed on:
- User OS & Browser:
- Grafana plugins:
- Others:

torkelo:
Fixes #23561

This improves all panels with auto decimals that use scaled units that go above 1K that before had to many decimals. 

Also fixes decimals in GraphNG y-axis ticks and the above issues did not impact the old graph panel as that used special decimal logic. 

Before
Screenshot from 2021-01-13 22-12-20

After
Screenshot from 2021-01-13 22-01-47

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.ts
+++ b/packages/grafana-data/src/field/displayProcessor.ts
@@ -22,6 +22,10 @@ interface DisplayProcessorOptions {
    * Will pick 'dark' if not defined
    */
   theme?: GrafanaTheme;
+  /**
+   * Used by auto decimals logic
+   **/
+  tickSize?: number;
 }
 
 // Reasonable units for time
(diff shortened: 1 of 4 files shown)


packages/grafana-data/src/field/displayProcessor.ts
```diff
@@ -22,6 +22,10 @@ interface DisplayProcessorOptions {
    * Will pick 'dark' if not defined
    */
   theme?: GrafanaTheme;
+  /**
+   * Used by auto decimals logic
+   **/
+  tickSize?: number;
```
ryantxu, line 28: is this used?

torkelo: Ah no, i did go down a much more complex route of modifying GrapNG and AxisBuiler to support the same logic flot has where it has the tick size (delta) to aid the decimals logic. But found a solution that did not need that so reverted all of those changes. Must have missed reverting this one

mckn (approved):
Awesome improvement! Works as expected when I am testing it.

rovo89:
I'm not 100% sure about the implications of this change. Does decimals = auto mean **all** decimals now, is that the intention?

Here's a test dashboard on Grafana 7.3.7 with different static values and unit *bytes (SI)*:
grafana_737

Same on 7.4.0-beta1:
grafana_740

In short, the Stats panels used to show 3-4 digits in total, which I think was nice. I don't really care about the kilobytes if I have 100+ MB, but for a one/two-digit number, a few decimals help. Now all decimals are shown, which is never what I want. Alternatively, I can configure a fixed amount of decimals, but the former implementation was quite clever to reduce precision where it doesn't matter.

Graph panels didn't change at all - the tooltip continues to show one decimal more than configured. The y-axis respects the setting, but maybe it shouldn't - cutting off the zeros would work fine, maybe the decimals setting could be understood as *maximum* precision for the y-axis.

Finally, the new Time series panel looks like this (with a single point):
grafana_740_timeseries

The tooltips show the same precision as the Stats panel, so same comments about this and also about the y-axis.

PR #30266, opened 2021-01-14
CODEOWNERS: Make backend squad owners of backend style guidelines

aknuds1:
In .github/CODEOWNERS, make backend squad owners of backend code style guidelines. This change makes sense since we are responsible for the Grafana backend code, and reviewers from said squad should be picked when PRs are made to this document.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -23,7 +23,7 @@ go.mod @grafana/backend-platform
 go.sum @grafana/backend-platform
 
 # Backend code docs
-/contribute/style-guides/backend.md
+/contribute/style-guides/backend.md @grafana/backend-platform
 
 /e2e @grafana/grafana-frontend-platform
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -22,6 +22,9 @@
 go.mod @grafana/backend-platform
 go.sum @grafana/backend-platform
 
+# Backend code docs
+/contribute/style-guides/backend.md
```
sakjur, line 26: This might just be my limited knowledge of the CODEOWNERS syntax, but shouldn't this be

```suggestion
/contribute/style-guides/backend.md @grafana/backend-platform
```

aknuds1: Thanks @sakjur, I probably was too sleepy when adding this. It looks correct to me.

sakjur (approved):

papagian (approved):

PR #30268, opened 2021-01-14
GraphNG: support fill below to (bands)

ryantxu:
This migrates the flot behavior to GraphNG:
image

The process for building charts with bands is pretty complex and requires some mental gymnastics.  I propose that we:
1. include this PR that will allow full control of all options
2. before 8, add a new panel "Bands"?  that makes it very easy to setup banded chart between min/max/mean/samples and has simple style and color pickers.


packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -169,6 +171,21 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
+        if (!indexByName) {
+          indexByName = getNamesToFieldIndex(alignedFrame);
+        }
+        const t = indexByName.get(getFieldDisplayName(field, alignedFrame));
+        const b = indexByName.get(customConfig.fillBelowTo);
+        if (isNumber(b) && isNumber(t)) {
+          console.log('todo bands', t, b);
+          builder.addBand({
+            series: [t, b],
+            fill: 'rgba(255,0,0,0.1)',
```
leeoniya, line 184: if you leave this null, it will use the fill of the high series.

ryantxu:
@torkelo -- can you take a look?  I am not wild about the code, but the behavior is worth evaluating.  In particular picking an opacity when not configured.

Also -- in angular, adding 'fillBelowTo' will also remove lines for both series:
image

torkelo:
Yes will do!

torkelo (approved):
It works well but is a bit of a pain to setup as you have to override both min & max and set, set fill below to & line width on max, and line width on min. So at minimum 2 overrides with 3 properties total. 

Would be nice if we could default to no lines unless specifically set in an override but there is no way for GraphNG to know if a line width was derived from a default or set via override 

Screenshot from 2021-01-15 09-59-18

PR #30272, opened 2021-01-14
TimeSeriesPanel: The new graph panel now supports y-axis value mapping 

### LINKED ISSUE #9292: [Feature Request] . Value-to-text for Y Axis in Graphs.
- What Grafana version are you using?  4.3.2
- What datasource are you using?  MySQL
- What OS are you running grafana on? Centos 7

I tried doing a search first and didn't come up with anything relevant for just graphs, so apologies if this is a dupe.   I know there are value-to-text mappings for single stats, but I'd like to advocate for a Y-axis value-to text mapping (or custom range...if you will).    Take the below image:
image

This particular graph will only have a value range of 1-7, never higher or lower.  As is, I've had to add a single row text field with a legend showing the Number-to-text mapping.   Given that this graph will never change besides 1-7, I'd like to be able to label the Y-axis something like "Success, FAIL, error, etc... instead of just the numbers.   I realize this is a somewhat limited scope, but I'm sure there are others out there in similar situations that might find this useful.

torkelo:
Closes #9292 

Screenshot from 2021-01-14 09-52-45

### DIFF
--- a/devenv/dev-dashboards/panel-graph/graph-ng.json
+++ b/devenv/dev-dashboards/panel-graph/graph-ng.json
@@ -1328,30 +1328,30 @@
           "stringInput": "6,6"
         },
         {
-          "points": [],
-          "stream": {
-            "type": "signal",
-            "speed": 250,
-            "spread": 3.5,
-            "noise": 2.2,
-            "bands": 1
-          },
-          "pulseWave": {
-            "timeStep": 60,
-            "onCount": 3,
-            "onValue": 2,
-            "offCount": 3,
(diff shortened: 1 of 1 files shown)


dprokop (approved):

lealceldeiro:
How do I use this feature. It's not documented yet on https://grafana.com/docs/grafana/latest/panels/visualizations/graph-panel/ ?

lealceldeiro:
OK, found: https://grafana.com/docs/grafana/latest/panels/visualizations/time-series/

PR #30277, opened 2021-01-14
Icons: Adds custom icon support ands new panel and interpolation icons

torkelo:
* [x] Adds the possibility to refer to custom svg with a `gf-` name prefix. 
* [x] Adds new interpolation icons
* [x] Adds new panel icons
* [x] Makes it possible to have icon only & description for tooltips in RadioButtons / Group
* [x] Updates RadioButtonGroup story 

Screenshot from 2021-01-14 11-51-34
Screenshot from 2021-01-14 11-51-44
Screenshot from 2021-01-14 11-51-53
Screenshot from 2021-01-14 11-52-00


packages/grafana-ui/src/components/Icon/custom/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const IconNotFound: FC<SvgProps> = ({ ...rest }) => {
+  return <svg {...rest} />;
+};
+
+export const customIcons: Record<string, ComponentType<SvgProps>> = {
+  'gf-interpolation-linear': InterpolationLinear,
+  'gf-interpolation-smooth': InterpolationSmooth,
+  'gf-interpolation-step-before': InterpolationStepBefore,
+  'gf-interpolation-step-after': InterpolationStepAfter,
```
dprokop, line 76: @jessover9000 have we requested those icons to be included in the Unicons package?

dprokop (approved):
Nice!

packages/grafana-data/src/types/select.ts
```diff
@@ -5,6 +5,7 @@ export interface SelectableValue<T = any> {
   label?: string;
   value?: T;
   imgUrl?: string;
+  icon?: string;
```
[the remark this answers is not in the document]
torkelo: yea, was thinking the same

packages/grafana-ui/src/components/Icon/custom/index.tsx
torkelo: @dprokop I do not think we need to, these are custom to Grafana

jessover9000: We haven't, but I also haven't consciously seen them / don't know what they even look like. If Torkel says they're custom and they work with unicons ok, we can leave them? If we want them tweaked to match the style better, just tell me :)

dprokop: No, just asking out of curiosity. I thing the custom ones are just fine

torkelo: Yea these Graph line interpolation modes feels very custom to Grafana

PR #30278, opened 2021-01-14
PanelLibrary: adding library panels to Dashboard Api

hugohaggmark:
This PR is a first exploratory POC of how loading of dashboards could work with Panel Library. I've done some workarounds I'm not particularly happy with, but that's because I don't know better. I'll explain them in comments below.

This draft only works for loading dashboards at the moment, saving is not implemented yet.

My thinking is that a library panel model json will have an extra property if it's connected to a dashboard like so:
```json
...
"libraryPanel": {
  "name": "<some unique name>",
  "uid": "<someuid>"
},
...
```

When loading a dashboard we will loop through all the panels in the dashboard json and if we find the `libraryPanel` property we try to match it with the connected library panels from the database. If the match is ok, then we replace the panel in the dashboard json with the library panel from the database and add back any dashboard specific properties (id, gridPos at the moment).

When storing a dashboard ~~(not implemented yet)~~ we ~~would need to~~ do the reverse and loop through all panels and find library panels. For each library panel we would just keep the `libraryPanel` property and any dashboard specific properties  (id, gridPos at the moment).

When deleting a dashboard ~~(not implemented yet)~~ we ~~would need to~~ remove any existing connections.

**To test this PR**
1. Add feature toggle
```
   [feature_toggles]
    enable = panelLibrary
```
2. Run SQL script
```sql
DELETE FROM library_panel WHERE ID > 0;
DELETE FROM library_panel_dashboard WHERE ID > 0;

INSERT INTO library_panel (id, org_id, folder_id, uid, name, model, created, created_by, updated, updated_by) VALUES (1,1,2, 'g1sNpCaMz','React Gauge','{"cacheTimeout":null,"datasource":null,"fieldConfig":{"defaults":{"color":{"mode":"thresholds"},"custom":{},"links":[{"targetBlank":true,"title":"Go to drilldown","url":"/d/O6GmNPvWk/dashboard-tests-nested-template-variables-drilldown?orgId=1&${__all_variables}&${__url_time_range}"}],"mappings":[],"max":100,"min":0,"nullValueMode":"connected","thresholds":{"mode":"absolute","steps":[{"color":"green","value":null},{"color":"red","value":80}]},"unit":"none"},"overrides":[]},"gridPos":{"h":9,"w":4,"x":16,"y":0},"id":8,"links":[],"options":{"orientation":"horizontal","reduceOptions":{"calcs":["mean"],"fields":"","values":false},"showThresholdLabels":false,"showThresholdMarkers":true,"text":{}},"pluginVersion":"7.4.0-pre","targets":[{"refId":"A","scenarioId":"random_walk"}],"timeFrom":null,"timeShift":null,"title":"React gauge datalink","type":"gauge"}','2021-01-14 10:12:22',1,'2021-01-14 10:12:22',1);
INSERT INTO library_panel (id, org_id, folder_id, uid, name, model, created, created_by, updated, updated_by) VALUES (2,1,2, 'MAnX2ifMk','React Table','{"datasource":"gdev-testdata","fieldConfig":{"defaults":{"color":{"mode":"thresholds"},"custom":{"align":"center","displayMode":"color-background","filterable":false},"mappings":[],"thresholds":{"mode":"percentage","steps":[{"color":"green","value":null},{"color":"blue","value":20},{"color":"orange","value":60},{"color":"red","value":70}]},"unit":"degree"},"overrides":[{"matcher":{"id":"byName","options":"Max"},"properties":[{"id":"custom.width","value":84}]},{"matcher":{"id":"byName","options":"Last"},"properties":[{"id":"custom.width","value":78}]},{"matcher":{"id":"byName","options":"Mean"},"properties":[{"id":"custom.width","value":74}]},{"matcher":{"id":"byName","options":"Field"},"properties":[{"id":"custom.align","value":"left"}]}]},"gridPos":{"h":16,"w":7,"x":0,"y":1},"id":4,"options":{"showHeader":true,"sortBy":[{"desc":true,"displayName":"Last"}]},"pluginVersion":"7.4.0-pre","targets":[{"refId":"A","scenarioId":"random_walk","seriesCount":15,"stringInput":""}],"timeFrom":null,"timeShift":null,"title":"Colored background","transformations":[{"id":"reduce","options":{"reducers":["max","mean","last"]}}],"type":"table"}','2021-01-14 10:12:22',1,'2021-01-14 10:12:22',1);

INSERT INTO library_panel_dashboard (id, librarypanel_id, dashboard_id, created, created_by)
SELECT 1,1,id, '2021-01-14 05:40:52',1 FROM dashboard WHERE title = 'Panel - Panel Library';
INSERT INTO library_panel_dashboard (id, librarypanel_id, dashboard_id, created, created_by)
SELECT 2,2,id, '2021-01-14 05:41:15',1 FROM dashboard WHERE title = 'Panel - Panel Library';
```

3. Navigate to [http://localhost:3000/d/imQX6j-Gz/panel-panel-library?orgId=1](http://localhost:3000/d/imQX6j-Gz/panel-panel-library?orgId=1)
4. You should see the following:
image
5. Use the API to update the library panels and they should be updated 🤞 

Relates #29610


pkg/models/dashboards.go
```diff
@@ -461,3 +462,24 @@ type GetDashboardRefByIdQuery struct {
...
+	UID      string `xorm:"uid"`
+	Name     string
+	Model    json.RawMessage
+
+	Created time.Time
+	Updated time.Time
+
+	CreatedBy int64
+	UpdatedBy int64
+}
```
hugohaggmark, line 485: This struct is already defined in the librarypanels package but when I tried to use that I got circular dependencies. Any ideas how to solve this?

pkg/services/sqlstore/dashboard_librarypanels.go
```diff
@@ -0,0 +1,35 @@
...
+
+	libraryPanelMap := make(map[string]models.LibraryPanel)
+	for _, panel := range libraryPanels {
+		libraryPanelMap[panel.UID] = panel
+	}
+
+	query.Result = libraryPanelMap
+
+	return nil
+}
```
hugohaggmark, line 35: Maybe this should be in the librarypanel service instead? I don't know what would be the preferred approach here because the library panel service doesn't use the bus at all. Should librarypanel service be a dependency to dashboard? I feel like my backend skills are a bit weak to reason about this and could need help here.

pkg/models/dashboards.go
torkelo: Can the dashboard api layer use the PanelLibraryService? you can add a new dependency here: https://github.com/grafana/grafana/blob/master/pkg/api/http_server.go#L81  

Then the api handler can just call the libraryService .GetLibararyPanelsForDashboard (or something like that)

pkg/services/sqlstore/dashboard_librarypanels.go
torkelo: think you can move this to a PanelLibraryService  function

torkelo (commented):
I do like some aspect of doing json model mergin in the backend. 

The downside I can see is for provisioning you do not want this. So say in the frontend you load a provisioned dashboard, you do modifications and want to save (see json) to copy it to git then you want the dashboard json without any library panel properies (just uid & name), so for this case we sort of need at least the "remove library panel properies" code in the frontend as well. 

Doing that cleanup should be straight forward and not a lof code to duplicate in backend & frontend though.

hugohaggmark:
Gotcha, so the workflow to save a provisioned dashboard today is Share =>  Export for sharing externally => Save to File?

pkg/models/dashboards.go
hugohaggmark: Yes, I saw there were already plenty of dependencies there, so I wasn't sure that was the correct way. I'll add the dependency to http server as suggested. Thank you!

pkg/services/sqlstore/dashboard_librarypanels.go
hugohaggmark: Yepp will do that. Thank you for clarifying!

pkg/models/dashboards.go
hugohaggmark: Doesn't seem to work though
```
Version: 7.4.0-pre, Linux Version: 7.4.0, Package Iteration: 1610632398pre
go build -ldflags -w -X main.version=7.4.0-pre -X main.commit=9fdb934535 -X main.buildstamp=1610621026 -X main.buildBranch=hugoh/dashboard-loadlibpanels -o ./bin/grafana-server ./pkg/cmd/grafana-server
package github.com/grafana/grafana/pkg/cmd/grafana-server
	imports github.com/grafana/grafana/pkg/server
	imports github.com/grafana/grafana/pkg/api
	imports github.com/grafana/grafana/pkg/services/librarypanels
	imports github.com/grafana/grafana/pkg/api: import cycle not allowed
```

Maybe the simpler micro service pattern doesn't work here, do you think I need to refactor the LibraryPanelService to the "old" style? Need to dig deeper and see if there is some examples already.

torkelo:
> Gotcha, so the workflow to save a provisioned dashboard today is Share => Export for sharing externally => Save to File?

Sort of, you can just hit save and you get a special save provisioned dashboard modal (try saving any of the devenv dashboards)

pkg/models/dashboards.go
torkelo: interesting, hm.. might need to break out either http server from the root api package, or the api.Wrap & util functions into a separate api sub package. 

If dashboard has it's own service (that handled http api & db) then this would not be a problem, this looks like it's only a problem due to api package having dashboards api handlers & the base api interfaces & util functions & types that all "micro" services use.  

thoughts @aknuds1  @marefr on how to solve this? (handler in api package being able to reference a service that defines api handlers itself)

hugohaggmark: I moved `common.go` into a `utils` folder. Naming here is tricky I tried `common`, `helpers`, `apiHelpers` but stuck with `utils`. After that I moved structs and functions into the LibraryPanelService instead. I highly appreciate any feedback about the naming and the move, thank you 🙏

hugohaggmark: Hmm on second thought I think I'll create a separate PR with that move.

torkelo: > Hmm on second thought I think I'll create a separate PR with that move.
:+1:

hugohaggmark: Opened https://github.com/grafana/grafana/pull/30298 instead

pkg/services/librarypanels/database.go
```diff
@@ -229,3 +229,30 @@ func (lps *LibraryPanelService) patchLibraryPanel(c *models.ReqContext, cmd patc
...
+	libraryPanelMap := make(map[string]LibraryPanel)
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		sql := `SELECT
+				lp.id, lp.org_id, lp.folder_id, lp.uid, lp.name, lp.model, lp.created, lp.created_by, lp.updated, updated_by
+			FROM
+				library_panel_dashboard lpd
+			INNER JOIN
+				library_panel lp ON lpd.librarypanel_id = lp.id AND lpd.dashboard_id=?`
+
+		libraryPanels := make([]LibraryPanel, 0)
```
aknuds1, line 243:
```suggestion
		var libraryPanels []LibraryPanel
```

aknuds1 (changes requested):

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"errors"
+
+	"github.com/grafana/grafana/pkg/components/simplejson"
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/grafana/grafana/pkg/services/librarypanels"
+)
+
+func (hs *HTTPServer) loadLibraryPanels(dash *models.Dashboard) error {
```
aknuds1, line 11: How about moving the `loadLibraryPanels` method to `LibraryPanelService` instead (and call it LoadDashLibraryPanels for clarity)? It would make HTTPServer less fat, and I don't see any reason not to do it.

hugohaggmark: Good suggestion, I'll change that.

pkg/services/librarypanels/database.go
hugohaggmark: Damn, I never learn when to use which. Thank you for suggesting the change.

pkg/services/librarypanels/database.go
```diff
@@ -229,3 +229,30 @@ func (lps *LibraryPanelService) patchLibraryPanel(c *models.ReqContext, cmd patc
...
+
+func (lps *LibraryPanelService) GetLibraryPanelsForDashboard(query GetLibraryPanelsQuery) (map[string]LibraryPanel, error) {
+	libraryPanelMap := make(map[string]LibraryPanel)
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		sql := `SELECT
+				lp.id, lp.org_id, lp.folder_id, lp.uid, lp.name, lp.model, lp.created, lp.created_by, lp.updated, updated_by
+			FROM
+				library_panel_dashboard lpd
+			INNER JOIN
+				library_panel lp ON lpd.librarypanel_id = lp.id AND lpd.dashboard_id=?`
```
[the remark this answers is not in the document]
hugohaggmark: Will fix

pkg/services/librarypanels/database.go
```diff
@@ -229,3 +229,30 @@ func (lps *LibraryPanelService) patchLibraryPanel(c *models.ReqContext, cmd patc
...
 	return libraryPanel, err
 }
+
+func (lps *LibraryPanelService) GetLibraryPanelsForDashboard(query GetLibraryPanelsQuery) (map[string]LibraryPanel, error) {
+	libraryPanelMap := make(map[string]LibraryPanel)
+	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		sql := `SELECT
+				lp.id, lp.org_id, lp.folder_id, lp.uid, lp.name, lp.model, lp.created, lp.created_by, lp.updated, updated_by
+			FROM
+				library_panel_dashboard lpd
```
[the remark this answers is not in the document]
hugohaggmark: Will fix

pkg/api/utils/common.go
```diff
@@ -55,6 +63,16 @@ func (r *NormalResponse) Body() []byte {
 	return r.body
 }
 
+// Err gets the response's body.
+func (r *NormalResponse) Err() error {
+	return r.err
+}
+
+// ErrMessage gets the response's body.
```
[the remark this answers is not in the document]
hugohaggmark: This will be merged with the other PR so these changes will be corrected.

pkg/api/utils/common.go
```diff
@@ -55,6 +63,16 @@ func (r *NormalResponse) Body() []byte {
 	return r.body
 }
 
+// Err gets the response's body.
```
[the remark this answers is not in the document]
hugohaggmark: This will be merged with the other PR so these changes will be corrected.

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+
+		// set the library panel json as new panel json in dashboard json
+		dash.Data.Get("panels").SetIndex(i, libraryPanelModelAsJson.Interface())
+
+		// set dashboard specific props
+		dash.Data.Get("panels").GetIndex(i).Set("gridPos", panelAsJson.Get("gridPos").MustMap())
+		dash.Data.Get("panels").GetIndex(i).Set("id", panelAsJson.Get("id").MustInt64())
+		dash.Data.Get("panels").GetIndex(i).Set("libraryPanel", make(map[string]interface{}))
+		dash.Data.Get("panels").GetIndex(i).Get("libraryPanel").Set("uid", libraryPanelInDb.UID)
+		dash.Data.Get("panels").GetIndex(i).Get("libraryPanel").Set("name", libraryPanelInDb.Name)
```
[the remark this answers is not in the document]
hugohaggmark: Yes, this looks more elegant. Thank you for suggestion.

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+		// we have a match between what is stored in db and in dashboard json
+		libraryPanelModel, err := libraryPanelInDb.Model.MarshalJSON()
+		if err != nil {
+			return errors.New("could not marshal library panel json")
+		}
+
+		libraryPanelModelAsJson, err := simplejson.NewJson(libraryPanelModel)
+		if err != nil {
+			return errors.New("could not convert library panel to json")
+		}
```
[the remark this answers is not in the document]
hugohaggmark: I'll keep simpleJson for now but add the original error to error message.

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+
+		libraryPanelInDb, ok := libraryPanels[uid]
+		if !ok {
+			return errors.New("found a library panel that does not exists as a connection")
+		}
+
+		// we have a match between what is stored in db and in dashboard json
+		libraryPanelModel, err := libraryPanelInDb.Model.MarshalJSON()
+		if err != nil {
+			return errors.New("could not marshal library panel json")
```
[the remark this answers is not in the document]
hugohaggmark: Good find!

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+			continue
+		}
+
+		// we have a library panel
+		uid := libraryPanel.Get("uid").MustString()
+		if len(uid) == 0 {
+			return errors.New("found a library panel without uid")
+		}
+
+		libraryPanelInDb, ok := libraryPanels[uid]
```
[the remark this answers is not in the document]
hugohaggmark: My frontend naming still there, I'll fix.

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+	query := librarypanels.GetLibraryPanelsQuery{DashboardId: dash.Id}
+	libraryPanels, err := hs.LibraryPanelService.GetLibraryPanelsForDashboard(query)
+	if err != nil {
+		return err
+	}
+
+	panels := dash.Data.Get("panels").MustArray()
+	for i, panel := range panels {
+		panelAsJson := simplejson.NewFromAny(panel)
+		libraryPanel := panelAsJson.Get("libraryPanel")
```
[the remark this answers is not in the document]
hugohaggmark: My frontend naming still there, I'll fix.

pkg/api/dashboard_librarypanels.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+
+	"github.com/grafana/grafana/pkg/components/simplejson"
+	"github.com/grafana/grafana/pkg/models"
+	"github.com/grafana/grafana/pkg/services/librarypanels"
+)
+
+func (hs *HTTPServer) loadLibraryPanels(dash *models.Dashboard) error {
+	query := librarypanels.GetLibraryPanelsQuery{DashboardId: dash.Id}
+	libraryPanels, err := hs.LibraryPanelService.GetLibraryPanelsForDashboard(query)
```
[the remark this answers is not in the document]
hugohaggmark: I used the bus in a previous commit and forgot to change, great catch I'll change it.

pkg/services/librarypanels/database.go
aknuds1: @hugohaggmark It's just a nuance, but I figure there's no point in allocating an array if it's just going to be replaced.

hugohaggmark:
Sorry everyone, but I needed to force push to reset and merge the changes from master.

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed the error formatting!

pkg/api/dashboard_librarypanels.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/services/librarypanels/database.go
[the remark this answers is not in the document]
hugohaggmark: Fixed!

pkg/services/librarypanels/database.go
hugohaggmark: Fixed!

pkg/api/dashboard_librarypanels.go
hugohaggmark: Fixed!

pkg/services/librarypanels/librarypanels.go
```diff
@@ -39,6 +44,63 @@ func (lps *LibraryPanelService) IsEnabled() bool {
...
+
+		// we have a library panel
+		uid := libraryPanel.Get("uid").MustString()
+		if len(uid) == 0 {
+			return errors.New("found a library panel without uid")
+		}
+
+		libraryPanelInDB, ok := libraryPanels[uid]
+		if !ok {
+			return errors.New("found a library panel that does not exists as a connection")
```
aknuds1, line 74: There's a typo in the error message, and I think the following is more explanatory for someone debugging:
```suggestion
			return fmt.Errorf("found connection to library panel %q that isn't in database", uid)
```

pkg/services/librarypanels/librarypanels.go
```diff
@@ -39,6 +44,63 @@ func (lps *LibraryPanelService) IsEnabled() bool {
...
+
+		// we have a match between what is stored in db and in dashboard json
+		libraryPanelModel, err := libraryPanelInDB.Model.MarshalJSON()
+		if err != nil {
+			return fmt.Errorf("could not marshal library panel JSON: %w", err)
+		}
+
+		libraryPanelModelAsJSON, err := simplejson.NewJson(libraryPanelModel)
+		if err != nil {
+			return fmt.Errorf("could not convert library panel to JSON: %w", err)
```
aknuds1, line 85: I realize that this is converting what is already JSON to simplejson, right?
```suggestion
			return fmt.Errorf("could not convert library panel to simplejson model: %w", err)
```

aknuds1 (approved):
Generally LGTM, but see suggestions for error string improvements.

pkg/services/librarypanels/librarypanels.go
hugohaggmark: You're correct

pkg/api/dashboard.go
```diff
@@ -230,6 +230,13 @@ func deleteDashboard(c *models.ReqContext) response.Response {
 		return response.Error(500, "Failed to delete dashboard", err)
 	}
 
+	if hs.Cfg.IsPanelLibraryEnabled() {
```
aknuds1, line 233: Doesn't it make more sense to disconnect library panels from the dashboard _before_ deleting the dashboard? If the database had foreign key constraints, deleting the dashboard first could fail due to the references. Also, references could potentially be left dangling after deleting the dashboard, in case the following deletion doesn't work.

hugohaggmark: Ok, that makes sense, just remember there are no foreign keys in Grafana :)

aknuds1: @hugohaggmark Yeah I am aware, but it's good to think as if there could be. The assumption could break at some point.

hugohaggmark: Fixed!

torkelo (approved):
Code looks solid to me! :tada:

hugohaggmark:
@aknuds1 any more feedback before I merge this? No rush, if you have more pressing beta1 PRs to review.

aknuds1:
@hugohaggmark Checking.

pkg/services/librarypanels/librarypanels.go
```diff
@@ -39,6 +43,158 @@ func (lps *LibraryPanelService) IsEnabled() bool {
...
+			"uid":  libraryPanelInDB.UID,
+			"name": libraryPanelInDB.Name,
+		})
+	}
+
+	return nil
+}
+
+// CleanLibraryPanelsForDashboard cleans library panels JSON before storing a dashboard to the database.
+func (lps *LibraryPanelService) CleanLibraryPanelsForDashboard(dash *models.Dashboard) error {
```
aknuds1, line 104: Looking at the usage of this method and the tests, can't it just be called from `ConnectLibraryPanelsForDashboard` instead? Is there any reason to burden the client with calling both of these methods?

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -516,6 +519,578 @@ func TestPatchLibraryPanel(t *testing.T) {
...
+				Data: simplejson.NewFromAny(dashJSON),
+			}
+
+			err = sc.service.ConnectLibraryPanelsForDashboard(sc.reqContext, &dash)
+			require.EqualError(t, err, errLibraryPanelHeaderUIDMissing.Error())
+		})
+}
+
+func TestDisconnectLibraryPanelsForDashboard(t *testing.T) {
+	testScenario(t, "When an admin tries to delete a dashboard with a library panel, it should disconnect the dashboard",
```
aknuds1, line 981:
```suggestion
	testScenario(t, "When an admin tries to delete a dashboard with a library panel, it should disconnect the two",
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -516,6 +519,578 @@ func TestPatchLibraryPanel(t *testing.T) {
...
+				Data: simplejson.NewFromAny(dashJSON),
+			}
+
+			err = sc.service.CleanLibraryPanelsForDashboard(&dash)
+			require.EqualError(t, err, errLibraryPanelHeaderNameMissing.Error())
+		})
+}
+
+func TestConnectLibraryPanelsForDashboard(t *testing.T) {
+	testScenario(t, "When an admin tries to store a dashboard with a library panel, it should connect the dashboard",
```
aknuds1, line 874:
```suggestion
	testScenario(t, "When an admin tries to store a dashboard with a library panel, it should connect the two",
```

aknuds1:
Basically, I don't understand why `CleanLibraryPanelsForDashboard` is a separate public method.

pkg/services/librarypanels/librarypanels.go
hugohaggmark: When the first function is called, the dashboard doesn't have an Id/uid and we want to clean up the JSON that is stored before it's stored. That's why it's a 2 shot call. Does that make sense?

hugohaggmark:
> Basically, I don't understand why `CleanLibraryPanelsForDashboard` is a separate public method.

I tried to explain in my comment, I hope that makes sense?

pkg/services/librarypanels/librarypanels.go
aknuds1: @hugohaggmark OK, makes sense.

hugohaggmark: To clarify, we can't connect the dashboard and library panels when there is no dashboardId and we don't want to store the whole library panel JSON in the db because that is already stored in the library_panel table.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -516,6 +519,578 @@ func TestPatchLibraryPanel(t *testing.T) {
...
+			require.NoError(t, err)
+
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": existing.Result.UID})
+			response = sc.service.getConnectedDashboardsHandler(sc.reqContext)
+			require.Equal(t, 200, response.Status())
+
+			var dashResult libraryPanelDashboardsResult
+			err = json.Unmarshal(response.Body(), &dashResult)
+			require.NoError(t, err)
+			require.Equal(t, 1, len(dashResult.Result))
```
aknuds1, line 928:
```suggestion
			require.Len(t, dashResult.Result, 1)
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -516,6 +519,578 @@ func TestPatchLibraryPanel(t *testing.T) {
...
+			require.NoError(t, err)
+
+			sc.reqContext.ReplaceAllParams(map[string]string{":uid": existing.Result.UID})
+			response = sc.service.getConnectedDashboardsHandler(sc.reqContext)
+			require.Equal(t, 200, response.Status())
+
+			var dashResult libraryPanelDashboardsResult
+			err = json.Unmarshal(response.Body(), &dashResult)
+			require.NoError(t, err)
+			require.Equal(t, 0, len(dashResult.Result))
```
aknuds1, line 1039:
```suggestion
			require.Empty(t, dashResult.Result)
```

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -516,6 +519,578 @@ func TestPatchLibraryPanel(t *testing.T) {
...
+		})
+
+	testScenario(t, "When an admin tries to delete a dashboard with a library panel without uid, it should fail",
```
aknuds1, line 1042: Actually, thinking about this scenario, I wonder if this type of discrepancy should be ignored (and just logged)? It's not really a problem is it that a connection isn't found in the database when you try to delete it?

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Thank you for suggesting this!

aknuds1:
I read the code some more and it dawned on me that you should probably be removing all dashboard/panel connections in one SQL transaction, not one per panel. You should also avoid returning an error, I think, if fewer connections are found than expected. Maybe you should log (or return?) an error if connections are left behind.

aknuds1:
Same goes for creating the connections, it's probably more performant/safe to create them all in one transaction?

hugohaggmark:
> I read the code some more and it dawned on me that you should probably be removing all dashboard/panel connections in one SQL transaction, not one per panel. You should also avoid returning an error, I think, if fewer connections are found than expected. Maybe you should log (or return?) an error if connections are left behind.

This is a brilliant suggestion, I'll look into this ASAP. Thank you!

aknuds1:
Thinking on it some more, probably best to not return an error if connections are left behind after deleting, but maybe log a warning so the admin can clean up?

hugohaggmark:
> Same goes for creating the connections, it's probably more performant/safe to create them all in one transaction?

@aknuds1 I'm struggling to solve this one as the SQL becomes something like the following:
`INSERT INTO library_panel_dashboard (librarypanel_id, dashboard_id, created, created_by) SELECT id,?,?,? FROM library_panel WHERE id IN (?` + strings.Repeat(",?", len(uids)-1) + `) `

I don't understand how  I resolve the array of uids in the (IN clause)?

aknuds1:
If you can't insert all connections in a single query, an improvement should at least to use only one transaction.

torkelo:
Moved this to v8 milestone

hugohaggmark:
My last commit adds code to delete all connections for a dashboard in one call and connects all library panels in the same transaction.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: I added a log message now, please let me know if that looks as it should.

hugohaggmark:
> Moved this to v8 milestone

We could also merge this during beta period @torkelo as it's behind a feature toggle. Or do you think we should wait?

torkelo:
No that is fine as well

torkelo:
Merge any time

hugohaggmark:
Cool, just waiting for some more feedback before merge.

pkg/services/librarypanels/database.go
```diff
@@ -110,6 +131,26 @@ func (lps *LibraryPanelService) disconnectDashboard(c *models.ReqContext, uid st
...
+	}
+	return lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		result, err := session.Exec("DELETE FROM library_panel_dashboard WHERE dashboard_id=?", dashboardID)
+		if err != nil {
+			return err
+		}
+		if rowsAffected, err := result.RowsAffected(); err != nil {
+			return err
+		} else if rowsAffected != panelCount {
+			return fmt.Errorf("disconnected %d panels, while expecting %d", rowsAffected, panelCount)
```
aknuds1, line 147: As discussed previously, I think it's a bad idea to return an error if a different number of connections were removed than expected. The reason is that a parallel process may have added or removed connections in the meantime. I think the only concern we should have here is whether connections were left behind in the database, and if they were it should only be logged as a warning, not returned as an error.

aknuds1 (commented):
Looks better, but I don't think we should treat it as an error if too many or too few connections were deleted.

pkg/services/librarypanels/database.go
hugohaggmark: Aah, ok now I understand what you mean. That makes sense.

pkg/services/librarypanels/database.go
```diff
@@ -110,6 +131,26 @@ func (lps *LibraryPanelService) disconnectDashboard(c *models.ReqContext, uid st
 	})
 }
 
+// disconnectLibraryPanelsForDashboard deletes connections for all Library Panels in a Dashboard.
+func (lps *LibraryPanelService) disconnectLibraryPanelsForDashboard(dashboardID int64, panelCount int64) error {
+	if panelCount == 0 {
```
aknuds1, line 136: This should be removed, since it's subject to race conditions, i.e. connections may have been added by a parallel process.

aknuds1 (changes requested):

pkg/services/librarypanels/database.go
hugohaggmark: Hmm, it's hard to tell what you want me to remove. Do you mean the if statement?

hugohaggmark: @aknuds1 I think you need to clarify this because I don't really understand.

aknuds1: I mean the `if` statement. It makes an assumption that may be wrong, i.e. we don't know the number of panels at this point in time.

aknuds1: It's connected to my [other](https://github.com/grafana/grafana/pull/30278#discussion_r560729613) comment, i.e. both are subject to the same race condition (another process may have added or removed connections).

hugohaggmark: Ok, I need to switch to multithreaded thinking here. Good catch!

pkg/services/librarypanels/database.go
```diff
@@ -144,7 +141,7 @@ func (lps *LibraryPanelService) disconnectLibraryPanelsForDashboard(dashboardID
 		if rowsAffected, err := result.RowsAffected(); err != nil {
 			return err
 		} else if rowsAffected != panelCount {
-			return fmt.Errorf("disconnected %d panels, while expecting %d", rowsAffected, panelCount)
+			lps.log.Warn("Number of disconnects does not match number of panels", "dashboard", dashboardID, "rowsAffected", rowsAffected, "panelCount", panelCount)
```
aknuds1, line 144: I think the only condition that should be logged is _if there are connections remaining in the DB_, we shouldn't worry the Grafana admin if we removed more connections than expected. It's not a blocker though, so I'll approve.

aknuds1 (approved):

pkg/services/librarypanels/database.go
hugohaggmark: @aknuds1 the check for the remaining connections would have to be done in a separate transaction, correct?

aknuds1: I don't think testing in another transaction would be worth it, not catch all cases.

hugohaggmark: But wouldn't the count always be `0` within the transaction?

aknuds1: True, within the transaction there shouldn't be any more connections I think.

hugohaggmark: I think I'll leave at as is for now.

aknuds1: It's still subject to a race condition though, since the panel count may have become stale before the transaction starts.

PR #30280, opened 2021-01-14
ReleaseNotes: Updated changelog and release notes for 7.3.7

grafanabot:
This exciting! So much has changed!


alexanderzobnin (approved):

CHANGELOG.md
```diff
@@ -50,6 +50,19 @@ In order not to break all angular panel plugins and data sources we have some cu
 
 <!-- 7.4.0 END -->
 
+<!-- 7.3.7 START -->
+
+# 7.3.7 (2021-01-14)
+
+### Bug fixes
```
osg-grafana, line 57: Please fix the # to ### heading level jump:

### --> ##

I suspect this would be handled upstream.

AgnesToulet: This is everywhere in the changelog file. If we want to change that for the whole changelog, I think it should be done in a separate PR.

PR #30281, opened 2021-01-14
Grafana UI: EmptySearchResult docs

Clarity-89:

Add story/docs for `EmptySearchResult` components and remove sass styles.


torkelo (approved):

ricardozv:
I need to create an icon on the left bar using react js, can I use your package for that? That's right?

PR #30283, opened 2021-01-14
AngularGraph: Fixes issues with legend wrapping after legend refactoring

torkelo:
The legends rewrite made the angular graph legend in table mode wrap.

* [x] Update angular graph legend symbol to be the new fatter one
* [x] Update angular graph legend table so that it does not wrap


dprokop (approved):
LGTM

PR #30284, opened 2021-01-14
Transformations: Format date

joanlopez:

This PR introduces a new transformation that allows the user to add new field from data format calculation which would allow performing aggregations over time fields. For example, given a `created_at` timestamp, the user can create a new field containing only the exact hour, day, week of year, month, etc.

It uses [Moment.js format](https://momentjs.com/docs/#/displaying/format/) to perform transformations.

**Very useful** for data sources like [GitHub](https://github.com/grafana/github-datasource) to perform aggregations based on datetimes. See the attached video for more details.

https://user-images.githubusercontent.com/5459617/104608877-b6f24500-5682-11eb-83e6-1ee6f96501a3.mp4

- It'd also make sense to (either complementary or alternatively) add support for grouping by date (same day, week, month), but the code behind the [GroupByTransformerEditor.tsx](https://github.com/grafana/grafana/blob/5f4b52812277cd919bec01e731cc988d605a62d3/public/app/core/components/TransformersUI/GroupByTransformerEditor.tsx) looked a bit more scary. Also, it follows performance optimization strategies (like calculating all the groups for all the fields to do a single traverse) that might be a bit harder to implement there or at least would make the code more complex.

- Also, note that the date formatting might be also used for table panels, in which case, the output shouldn't be the datetime Unix milliseconds but the raw string. I can add that functionality with checkbox on the transformation UI (inline) if you think that'd make sense.

### DIFF
--- a/public/app/core/components/TransformersUI/CalculateFieldTransformerEditor.tsx
+++ b/public/app/core/components/TransformersUI/CalculateFieldTransformerEditor.tsx
@@ -15,11 +15,12 @@ import {
   TransformerRegistyItem,
   TransformerUIProps,
 } from '@grafana/data';
-import { FilterPill, HorizontalGroup, Input, LegacyForms, Select, StatsPicker } from '@grafana/ui';
+import { FilterPill, HorizontalGroup, InlineLabel, Input, LegacyForms, Select, StatsPicker } from '@grafana/ui';
 import {
   BinaryOptions,
   CalculateFieldMode,
   CalculateFieldTransformerOptions,
+  DateOptions,
   getNameFromOptions,
   ReduceOptions,
 } from '@grafana/data/src/transformations/transformers/calculateField';
@@ -32,11 +33,13 @@ interface CalculateFieldTransformerEditorState {
   include: string[];
   names: string[];
   selected: string[];
+  timeFieldNames: Set<string>;
 }
 
 const calculationModes = [
   { value: CalculateFieldMode.BinaryOperation, label: 'Binary operation' },
   { value: CalculateFieldMode.ReduceRow, label: 'Reduce row' },
+  { value: CalculateFieldMode.FormatDate, label: 'Format date' },
 ];
 
 export class CalculateFieldTransformerEditor extends React.PureComponent<
@@ -50,16 +53,19 @@ export class CalculateFieldTransformerEditor extends React.PureComponent<
       include: props.options?.reduce?.include || [],
       names: [],
       selected: [],
+      timeFieldNames: new Set(),
     };
   }
 
   componentDidMount() {
     this.initOptions();
+    this.extractTimeFieldNames();
   }
 
   componentDidUpdate(oldProps: CalculateFieldTransformerEditorProps) {
     if (this.props.input !== oldProps.input) {
       this.initOptions();
+      this.extractTimeFieldNames();
     }
   }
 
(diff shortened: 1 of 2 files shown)


ryantxu:
Excellent!  I think this makes more sense as its own transformation rather than options on "Calculate field"  --- the concept is pretty unique and odd to think about as a calculation ;)

Maybe "Add field from date"?

We should also only require entering the explicit date pattern as a last result.  For grouping, the key ones we should promote would be:
* month  (jan-dec)
* day-of-week (mon-sunday)
* year (YYYY)
* hour-in-day (0-24)

For graphing reasons... we may want the month/day-of-week to be numbers, but then attach [ValueMapping](https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/types/dataFrame.ts#L68) so they render with words

dprokop:
> I think this makes more sense as its own transformation rather than options on "Calculate field" --- the concept is pretty unique and odd to think about as a calculation ;)
> Maybe "Add field from date"?

I agree with @ryantxu about this one. My first expectation when I tried to test this was to have a separate transformer rather than this being hidden behind the calculation. I also think about this more as a field creation rather than "calculation"

> We should also only require entering the explicit date pattern as a last result

@ryantxu do you suggest a list of predefined patterns?

public/app/core/components/TransformersUI/CalculateFieldTransformerEditor.tsx
```diff
@@ -324,6 +354,66 @@ export class CalculateFieldTransformerEditor extends React.PureComponent<
...
+          <Select
+            isClearable={false}
+            allowCustomValue={false}
+            className="width-18"
+            options={timeFieldOptions}
+            placeholder={'Select field'}
+            value={options?.fieldName ?? ''}
+            onChange={this.onFieldToFormatChanged}
+          />
+          <InlineLabel width="auto" tooltip={'Look at the Moment.js docs for allowed formats'}>
```
dprokop, line 403: I suggest, rather than referring to moment.js docs, to have the allowed tokens listed or linked to in this particular transformation's docs. Or ideally if the allowed tokens were listed in this inline help. Problem though is it's gonna be fairly long list.

joanlopez:
> > I think this makes more sense as its own transformation rather than options on "Calculate field" --- the concept is pretty unique and odd to think about as a calculation ;)
> > Maybe "Add field from date"?
> 
> I agree with @ryantxu about this one. My first expectation when I tried to test this was to have a separate transformer rather than this being hidden behind the calculation. I also think about this more as a field creation rather than "calculation"
> 
> > We should also only require entering the explicit date pattern as a last result
> 
> @ryantxu do you suggest a list of predefined patterns?

Yup, I discussed it in depth offline with @ryantxu, so I think I have a clear idea about how to re-do it in a better / more UX friendly way considering both ideas (new separated transformation + providing a list of choices). 

I didn't push any change yet because I've been focused on some issues planned for the `v7.4`, but I'll come back with a new implementation soon.

Thanks anyway for your feedback @dprokop, I'm glad to see you spent time trying it out and bringing some feedback 😍

mckn (commented):
Really nice addition! Looking forward for the new implementation 🙏

joanlopez:
> Really nice addition! Looking forward for the new implementation 🙏

Yeah, I've been focused on many different and somehow urgent things recently, but I hope to be able to submit the PR with the completely new transformation suggested by @ryantxu, sooner than later 😇 

Thanks for your feedback!! 😊

PR #30289, opened 2021-01-14
Graph: Consider reverse sorted data points on isOutsideRange check

joanlopez:

The current implementation of the `isOutsideRange` check within the `Graph` panel assumes the data points are chronologically sorted, so this PR introduces the changes needed to add support also for reverse sorted data points.

During some tests related with the transformation introduced [here](https://github.com/grafana/grafana/pull/30284), I realized that due to the date adjustments applied by the transformation (all the data points that happened on `YYYY-MM-DD HH:MM:SS` are grouped at `YYYY-MM-DD 00:00:00`), sometimes you end up with a data point before the `timeRange.from` (the left-most one, because`YYYY-MM-DD 00:00:00` < `YYYY-MM-DD HH:MM:SS`).

So, after some debugging with the super valuable help of @ryantxu, we realized the `isOutsideRange` check is only calculating the difference between the `last` data point and the `timeRange.from`, assuming the "last" (right-most) data point is the last in the array, so assuming the data points are chronologically sorted, which makes the warning to appear for reverse sorted data points in situations it's not expected. See the screenshots attached below.

image
image (1)

### DIFF
--- a/public/app/plugins/panel/graph/data_processor.ts
+++ b/public/app/plugins/panel/graph/data_processor.ts
@@ -94,7 +94,11 @@ export class DataProcessor {
       const from = range.from;
 
       if (last - from.valueOf() < -10000) {
-        series.isOutsideRange = true;
+        // If the data is in reverse order
+        const first = datapoints[0][1];
+        if (first - from.valueOf() < -10000) {
+          series.isOutsideRange = true;
+        }
       }
     }
     return series;


ryantxu (approved):
Looks good -- we should also make sure reverse sorted stuff behaves OK in the new timeseries panel!

PR #30294, opened 2021-01-15
Email: change the year in templates

adeptg:
Happy New Year! Let it be better than the previous one :)


dprokop (approved):
Happy New Year to you too @adeptg! Thank you for the contribution!

PR #30297, opened 2021-01-15
AlertingNG: Modify queries and transform endpoint to get datasource UIDs

papagian:
It modifies query model in alert definitions to expect `datasourceUid` (string) property instead of `datasourceId` (int).
However, for keep supporting queries with expressions (which have `datasourceId`) the transform endpoint should work with both (either data source UID or ID).

Fixes #

Requires grafana/grafana-plugin-sdk-go/pull/249


go.mod
```diff
@@ -43,7 +43,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/gosimple/slug v1.9.0
 	github.com/grafana/grafana-plugin-model v0.0.0-20190930120109-1fc953a61fb4
-	github.com/grafana/grafana-plugin-sdk-go v0.81.0
+	github.com/grafana/grafana-plugin-sdk-go v0.82.0
```
kylebrandt, line 46: Seem comments in https://github.com/grafana/grafana-plugin-sdk-go/pull/249 (In particular the convert to proto, maybe do that and cut .83 with this?

pkg/expr/transform.go
```diff
@@ -131,14 +131,17 @@ func QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.Que
...
+		datasourceUID = req.PluginContext.DataSourceInstanceSettings.UID
 	}
 
 	getDsInfo := &models.GetDataSourceQuery{
 		OrgId: req.PluginContext.OrgID,
 		Id:    datasourceID,
+		Uid:   datasourceUID,
 	}
 
 	if err := bus.Dispatch(getDsInfo); err != nil {
```
kylebrandt, line 147: I think this should perhaps be `ng.DatasourceCache`, looks like this is tech debt for an that results in extra db hits.

I think: if UID != "" ng.DatasourceCache.GetDatasourceByUID else  ng.DatasourceCache.GetDatasourceByID (So uid is preferred, but ID will still work).

papagian: I'm afraid that requires more refactoring: this function does not know about the cache service.

kylebrandt: Can consider for future pr

kylebrandt (approved):

PR #30298, opened 2021-01-15
Chore: Moves common and response into separate packages

hugohaggmark:
This PR is needed to avoid circular refrences when adding micro services [as described here](https://github.com/grafana/grafana/pull/30278#discussion_r557412182).

Relates #29610


aknuds1 (commented):
I'm reading your originating PR first, to understand the need.

pkg/api/utils/common.go
```diff
@@ -30,30 +31,26 @@ func Wrap(action interface{}) macaron.Handler {
...
 // JSON creates a JSON response.
-func JSON(status int, body interface{}) *NormalResponse {
+func JSON(status int, body interface{}) *response.NormalResponse {
 	return Respond(status, body).Header("Content-Type", "application/json")
 }
 
-// jsonStreaming creates a streaming JSON response.
-func jsonStreaming(status int, body interface{}) streamingResponse {
+// JsonStreaming creates a streaming JSON response.
+func JsonStreaming(status int, body interface{}) response.StreamingResponse {
```
aknuds1, line 39:
```suggestion
// JSONStreaming creates a streaming JSON response.
func JSONStreaming(status int, body interface{}) response.StreamingResponse {
```

aknuds1 (changes requested):
Why not move the functions creating responses to the api/response package instead? Would be more logically coherent/self-documenting to me, and less ambiguous ("utils" is an ambiguous package name).

hugohaggmark:
> Why not move the functions creating responses to the api/response package instead? Would be more logically coherent/self-documenting to me, and less ambiguous ("utils" is an ambiguous package name).

Thanks for the feedback, exactly which functions are you referring to?

hugohaggmark:
So I moved Wrap into the routing package and the rest of the function from common into the response package. I hope that makes it clearer.

aknuds1:
> > Why not move the functions creating responses to the api/response package instead? Would be more logically coherent/self-documenting to me, and less ambiguous ("utils" is an ambiguous package name).
> 
> Thanks for the feedback, exactly which functions are you referring to?
> I mean should everything in common.go be move back to response.go? We split common.go into 2 files recently and I was just respecting that split up.

I mean the functions creating responses, i.e. not Wrap.

I'll have a look at your latest revision.

marefr (approved):
Great :star2: LGTM

aknuds1 (approved):
LGTM

PR #30301, opened 2021-01-15
PanelEditor: Render panel field config categories as separate option group sections 

dprokop:
This PR introduces **a breaking**  change to the way how panel/field options categories configuration is interpreted in the UI.

Try Time series Beta panel to see this in action.

### Field options categories (collapsible by default after this PR)

API:

```
category: ['Axis']
```
First category in the list serves as options group indicator.

| BEFORE   | AFTER   |
|---|---|
|image|  image|
---
### Field options sub-categories

API:
`category: ['Axis', 'Subcategory']`
First category in the list serves as options group indicator.

| BEFORE   | AFTER   |
|---|---|
| image| image|
---
### Field options sub-categories in the main field options pane

API:
| BEFORE   | AFTER   |
|---|---|
|`category: ['Main options subcategory']`|`category: [undefined, 'Main options subcategory'],`

First category set to `undefined` marks the option to be categorized under the panel's main options group

| BEFORE   | AFTER   |
|---|---|
|image| image|

### DIFF
--- a/packages/grafana-data/src/types/options.ts
+++ b/packages/grafana-data/src/types/options.ts
@@ -39,7 +39,7 @@ export interface OptionEditorConfig<TOptions, TSettings = any, TValue = any> {
   /**
    * Array of strings representing category of the option. First element in the array will make option render as collapsible section.
    */
-  category?: string[];
+  category?: Array<string | undefined>;
 
   /**
    * Set this value if undefined
--- a/public/app/features/dashboard/components/PanelEditor/DynamicConfigValueEditor.tsx
+++ b/public/app/features/dashboard/components/PanelEditor/DynamicConfigValueEditor.tsx
@@ -35,7 +35,10 @@ export const DynamicConfigValueEditor: React.FC<DynamicConfigValueEditorProps> =
   // eslint-disable-next-line react/display-name
   const renderLabel = (includeDescription = true, includeCounter = false) => (isExpanded = false) => (
     <HorizontalGroup justify="space-between">
-      <Label category={item.category?.splice(1)} description={includeDescription ? item.description : undefined}>
+      <Label
+        category={item.category?.filter(c => c !== undefined) as string[]}
+        description={includeDescription ? item.description : undefined}
+      >
         {item.name}
         {!isExpanded && includeCounter && item.getItemsCount && <Counter value={item.getItemsCount(property.value)} />}
       </Label>
--- a/packages/grafana-data/src/panel/registryFactories.ts
+++ b/packages/grafana-data/src/panel/registryFactories.ts
@@ -24,7 +24,6 @@ export function createFieldConfigRegistry<TFieldConfigOptions>(
 
     for (const customProp of builder.getRegistry().list()) {
       customProp.isCustom = true;
-      customProp.category = [`${pluginName} options`].concat(customProp.category || []);
       // need to do something to make the custom items not conflict with standard ones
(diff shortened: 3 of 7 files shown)


torkelo:
Nice! 

It would be cool if the first section that has the panel name could be customized so instead of "Time series options" it could be:
* Series rendering
* Display
* Lines, Bars, Points, Areas modes 
* Rendering 
* Graphic options

packages/grafana-data/src/types/options.ts
```diff
@@ -39,7 +39,7 @@ export interface OptionEditorConfig<TOptions, TSettings = any, TValue = any> {
   /**
    * Array of strings representing category of the option. First element in the array will make option render as collapsible section.
    */
-  category?: string[];
+  category?: Array<string | undefined>;
```
marcusolsson, line 42: Out of curiosity, when would you have an undefined (sub)category?

dprokop: In a case when you want to have subcategories under the main category of the field options, see **Field options sub-categories in the main field options pane** in the PR description.

marcusolsson:
Just to make sure, because it doesn't seem to have a breaking API change, this merely changes how categories are displayed, i.e. plugins will still build and run?

dprokop:
> Just to make sure, because it doesn't seem to have a breaking API change, this merely changes how categories are displayed, i.e. plugins will still build and run?

Yes, it's breaking in terms of the UI interpretation, not the API per se.

ryantxu:
Nice work.   The "[Breaking]" notice seems a bit bold to me -- as it does not actually break anything ;)  If it gets listed in the breaking change section of the changelog, make sure it is clear that the only change is categories in FieldConfig are now shown in collapsable sections like they are for panel options.

ryantxu (approved):

public/app/features/dashboard/components/PanelEditor/DynamicConfigValueEditor.tsx
```diff
@@ -35,7 +35,10 @@ export const DynamicConfigValueEditor: React.FC<DynamicConfigValueEditorProps> =
   // eslint-disable-next-line react/display-name
   const renderLabel = (includeDescription = true, includeCounter = false) => (isExpanded = false) => (
     <HorizontalGroup justify="space-between">
-      <Label category={item.category?.splice(1)} description={includeDescription ? item.description : undefined}>
+      <Label
+        category={item.category?.filter(c => c !== undefined) as string[]}
```
torkelo, line 39: Wonder if we should remove this category on label and instead just support nested OptionGroup sections

torkelo:
I do not see this as a breaking change, Updated PR title. Think of this just as a UX change

public/app/features/dashboard/components/PanelEditor/DynamicConfigValueEditor.tsx
dprokop: @torkelo I think we should have a fallback for such situation, when someone adds more than two cattegories. Otherwise it may be confusing some of the strings in the list are missing. We could pottentially use a tuple to define the categories array `[string | undefined, string]`

But TBH - not sure if it's worth it.

torkelo: Yea not sure either, let's keep it and revisit as we do https://github.com/grafana/grafana/pull/30329

PR #30304, opened 2021-01-15
Grafana-UI: Add story/docs for ErrorBoundary

Clarity-89:

Add ErrorBoundary to Storybook. 

It seems like Storybook has it's own mechanism for catching errors, which comes on top of our alerts and needs to be closed first:
Screenshot 2021-01-15 at 14 54 32

I couldn't find a way to disable it for specific stories, maybe @jackw or @peterholmberg have any ideas here?


peterholmberg (approved):
I have completely missed this, sorry. It looks good!

PR #30305, opened 2021-01-15
Prettier: Update to version 2

torkelo:
Updates prettier to 2.2.1 from 1.19.

Prettier 2 has a lot better formatting for chained function calls. This accounts for most changes in this PR.

The other change is always paranthesis in arrow functions.

Not sure when is the best time to do this given the merge conflicts in ongoing PRs this will cause. Maybe just before beta branch of next week? That
should be a time where we have the fewest big PRs in progress.


torkelo:
Think we have to change from eslint-plugin-prettier to eslint-plugin-prettier to support 2.0

PR #30316, opened 2021-01-15
Expressions: Remove feature toggle

kylebrandt:
Makes expressions on my default.

 - Will need a UI indicator about breaking changes before merging this
 - Maybe we this should just be a different API endpoint -- seems cleaner. Can we change the front end to send to that when there is an expression if I do that @peterholmberg ?
 - Updates the endpoint so the DS is checked for every query in the case of Transforms. Need to make sure data source permissions can not be bypassed.
 - Docs will be in another PR, can start next week. Will be based on draft in https://github.com/kylebrandt/tempExprDoc/blob/main/expr.md


papagian:
I think that it should be also removed from the frontend:
https://github.com/grafana/grafana/blob/f9b3f4d47464cd731f9237638a138de9c146a3cb/packages/grafana-data/src/types/config.ts#L36
and then the [docs](https://grafana.com/docs/grafana/latest/packages_api/data/featuretoggles/#featuretoggles-interface) will automatically get synched.

torkelo:
We need to only show this for backend data sources

marefr (commented):
I suggested to break out transform to a new HTTP API endpoint to separate the concepts/logic.  Doing this you wouldn't have to "fake" a data source with a certain expression name. However, guess it simplifies things having this since alerting can use same query models for ordinary queries and expression queries. You said that you didn't want to change this in first version at least. 

I still think it's a bit hard to understand the query logic if there's expression or not so at least maybe divide the code paths if it's an expression or not in separate functions as a first step? 

In general a few tests of this would be good, but something to consider in a separate PR/later.

kylebrandt:
for reference beta badge looks like:

image

kylebrandt:
@torkelo Is showing backend only a requirement for beta1? Not sure if that is tricky or not. Also a bit complicated by multiple queries. (e.g. backend query + expression, backend query is removed (or frontend query is added)). Maybe better to show some sort of error/warning if an expression and frontend datasource exist together?

kylebrandt:
@marefr up against beta1 deadline, so do not think we will get to different endpoint for 7.4. Bit of crunch time, but will look into 

> code paths if it's an expression or not in separate functions as a first step

tomorrow.

torkelo:
Made the expression button only show for backend data sources (if alerting: true is set in plugin.json) 
And changed the beta badge that looked disconnected from the button to just plain text inside the button 
Screenshot from 2021-01-18 18-29-49

kylebrandt:
@torkelo Did you mean to push something with that last comment? (Screenshot looks good I think)

torkelo:
@kylebrandt yes, pushed now :)

kylebrandt:
@torkelo Can we show it with "Mixed"? Needed to be able to demonstrate the `$A + $B` sort of operations from different datasources.

image

Would also allow for standalone expression case if we show it in mixed, not that useful very often, but I use it sometimes for testing or demonstrating something.

kylebrandt:
We also should probably always show it in the `/ngalerting` view I think.

torkelo:
Yes, but there we should also filter the datasource picker to only show backend datasources. Problem is mixed datasource is not a backend datasource, might need some special logic to handle this case

torkelo:
Added logic to show it for mixed as well

kylebrandt:
Mixed is working for me now :+1:  . Is it possible to have a tool tip on the button? I want make sure that users understand that expressions may break with version upgrades - and I am not sure "beta" means that to them.

torkelo:
Added tooltip, not sure I am 100% comfortable having this not behind a feature toggle , feels strange to have an experimental feature that will likely break dashboards in a primary location next the add query button

torkelo:
I think we will need a way to disable it, I am sure a lot of Grafana admins will not want this experimental button there that will just confuse users and create dashboards that will break

kylebrandt:
> I think we will need a way to disable it, I am sure a lot of Grafana admins will not want this experimental button there that will just confuse users and create dashboards that will break

Would it be default "expressions enabled" or "expressions disabled". If the default behavior is disabled (e.g. `[Expressions] disabled = true`) the effectively it isn't really any different than the feature flag.

I think for me it is about visibility and a bit of a chicken and egg (if off by default, will not be seen, will not get feed back, etc). @daniellee was wanting this for 7.4 - so he may have thoughts.

kylebrandt:
image

Looks good to me.

torkelo:
> I think for me it is about visibility and a bit of a chicken and egg (if off by default, will not be seen, will not get feed back, etc). @daniellee was wanting this for 7.4 - so he may have thoughts.

yes, very good point. But think a way to disable it would be good still. Not sure the current feature toggle system can handle that

torkelo:
But maybe we can maybe do that during beta

kylebrandt:
> Not sure the current feature toggle system can handle that

I do not see a lot of risk in creating an `[Expressions]` section in the conf. And have then have a disabled property under that. Okay with that?, if so okay to get that disable option in after beta1 (but before stable) but still go forward with this?

torkelo:
> I do not see a lot of risk in creating an [Expressions] section in the conf. And have then have a disabled property under that. Okay with that?, if so okay to get that disable option in after beta1 (but before stable) but still go forward with this?

:+1:

kylebrandt:
@marefr pushed change to move the expression path to different function.

papagian (approved):
LGTM

PR #30320, opened 2021-01-15
grafana/ui: export TLS auth component

ryantxu:
Some plugins need to create TLSConfig but can not use the full datasource settings.  Can we expose this component directly?  This is needed for:
* mongodb
* sitewise iot


scottlepp (changes requested):
Thanks Ryan.  Though I can't really use this component (I tried but I ended up changing it a bit), I would like to use the CertificationKey component.  Can you export that one too?

scottlepp (approved):

PR #30323, opened 2021-01-15
BarChart: add alpha bar chart panel

ryantxu:
test with this data: https://docs.google.com/spreadsheets/d/1XP7Ze8kw6X3iaAnqi_UZAnwqfZR8yKmbKfdIITHs36g/view#gid=0

image

image

Test with multiple series and a `reduce` transformer:
image
image


torkelo:
Pushed the new icon 
Screenshot from 2021-01-16 07-35-39

public/app/plugins/panel/timeseries/config.ts
```diff
@@ -257,8 +257,10 @@ export function addLegendOptions(builder: PanelOptionsEditorBuilder<OptionsWithL
         ],
       },
       showIf: (c) => c.legend.displayMode !== LegendDisplayMode.Hidden,
-    })
-    .addCustomEditor<StatsPickerConfigSettings, string[]>({
+    });
+
+  if (!!!excludeCalcs) {
```
leeoniya, line 262: lol wtf :D

torkelo:
Great progress! looks good!

Looks like negative values are not working but otherwise a good base line :)

packages/grafana-ui/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,447 @@
...
+  width: number;
+  data: DataFrame;
+  onLegendClick?: (event: GraphNGLegendEvent) => void;
+  onSeriesColorChange?: (label: string, color: string) => void;
+}
+
+/**
+ * @alpha
+ */
+export const BarChart: React.FunctionComponent<Props> = ({
```
torkelo, line 41: a big complex component / function, is there a way to break this up a bit, move some of the nested functions out (drawBars, drawPoints), pointWithin, walkTo etc. 

A unit test or two and 1-2 storybook stories would also help

packages/grafana-ui/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,257 @@
...
+    // bar orientation -> x scale orientation & direction
+    const xOri = orientation === VizOrientation.Horizontal ? 1 : 0;
+    const xDir = orientation === VizOrientation.Horizontal ? -1 : 1;
+
+    const formatValue =
+      showValue !== BarValueVisibility.Never
+        ? (seriesIdx: number, value: any) => formattedValueToString(data.fields[seriesIdx].display!(value))
+        : undefined;
+
+    const config = getConfig(xOri, xDir, groupWidth, barWidth, formatValue);
```
leeoniya, line 85: probably should accept an options object instead of continuing to grow the args list.

packages/grafana-ui/src/components/BarChart/BarChart.tsx
leeoniya: pulled all the core stuff out of the react component.

fixed negative value handling.

added hover integration points for adding tooltip.

torkelo:
Pushed a minor style update that adds a line width and opacity. Might be panel-level options we need, but can be added later. 

Screen Shot 2021-01-28 at 20 02 04

torkelo:
@leeoniya try to avoid force pushing on PRs unless needed, makes it hard for others to test & update, but sometimes needed :)

leeoniya:
> @leeoniya try to avoid force pushing on PRs unless needed, makes it hard for others to test & update, but sometimes needed :)

sure, i did clear this one with ryan first :)

ryantxu:
I added some annotations to the VisLegend https://github.com/grafana/grafana/pull/30323/files#diff-0d79e34b6f7a84bb1429c1ccfff7feb32aa51cde7d698f47316ba667342f27e2R7 so it now passes the docs-build :tada:

packages/grafana-ui/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,310 @@
...
+  legend,
+  onLegendClick,
+  onSeriesColorChange,
+  ...plotProps
+}) => {
+  if (!data || data.fields.length < 2) {
+    return <div>Missing data</div>;
+  }
+
+  // dominik? TODO? can this all be moved into `useRevision`
```
dprokop, line 63: @leeoniya useRevision is a more generic hook that can be uses in different scenarios as well.

Alternative, to make this simpler and reusable in BarChart and TimeSeries panel etc would be to have `useDataFrameStructureCompare` hook, that would return the comparison function for useRevision to use. It would be sth like this:

```ts
// useDataFrameStructureCompare.ts

export function useDataFrameStructureCompare() {
  const compareFrames = useCallback((a?: DataFrame | null, b?: DataFrame | null) => {
    if (a && b) {
      return compareDataFrameStructures(a, b);
    }
    return false;
  }, []);

  return compareFrames;
}
```

then this code in BarChart/TimeSeries panel would change to:

```ts
const compareFrames = useDataFrameStructureCompare();
const configRev = useRevision(data, compareFrames);
```

But me and @mckn think we do not need the revision anymore if we slightly change the way we create the config. Basically move it to joinDataFrames logic when both aligned data and the config would be created removing the need for the config to be memoized. Then we would compare the not by reference but by value and pass either previous(unchanged) or new config to the UPlotChart component. But since the config can contain functions it's gonna require those functions memoization for the compare to give correct results. I hope to start POCing this approach later today/early next week.

packages/grafana-ui/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,310 @@
...
+      if (!field) {
+        return undefined;
+      }
+
+      return {
+        disabled: !seriesConfig.show ?? false,
+        fieldIndex,
+        color: seriesConfig.lineColor!,
+        label: seriesConfig.fieldName,
+        yAxis: 1,
```
dprokop, line 258: Think we need to account for the axis being either left or right. See TimeSeries panel legend creation.

Then this function will become identical and I think we could turn it into some utility function `getGraphLegendItems(builder: UPlotConfigBuilder): VizLegendItem[]`

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -170,18 +170,18 @@ export const GraphNG: React.FC<GraphNGProps> = ({
...
-        // The builder will manage unique scaleKeys and combine where appropriate
-        builder.addScale({
-          scaleKey,
-          distribution: customConfig.scaleDistribution?.type,
-          log: customConfig.scaleDistribution?.log,
-          min: field.config.min,
-          max: field.config.max,
-          softMin: customConfig.axisSoftMin,
-          softMax: customConfig.axisSoftMax,
-        });
```
dprokop, line 183: @leeoniya are you sure about this? I assume we are delegating combining the scales to uPlot?

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -328,7 +328,7 @@ export const GraphNG: React.FC<GraphNGProps> = ({
   );
 };
 
-const mapMouseEventToMode = (event: React.MouseEvent): GraphNGLegendEventMode => {
+export const mapMouseEventToMode = (event: React.MouseEvent): GraphNGLegendEventMode => {
```
dprokop, line 331: Think we can move this to new file utils.ts in the GraphNG dir

public/app/plugins/panel/timeseries/config.ts
```diff
@@ -151,85 +153,97 @@ export function getGraphFieldConfig(cfg: GraphFieldConfig): SetFieldConfigOption
...
-          process: (value) => value,
         });
+
+      addAxisConfig(builder);
+      addHideFrom(builder);
     },
   };
 }
 
+export function addHideFrom(builder: FieldConfigEditorBuilder<AxisConfig>) {
```
torkelo, line 164: nice to share these! :heart:

public/app/plugins/panel/timeseries/config.ts
```diff
@@ -151,85 +153,97 @@ export function getGraphFieldConfig(cfg: GraphFieldConfig): SetFieldConfigOption
...
+    override: SeriesConfigEditor,
+    shouldApply: () => true,
+    hideFromDefaults: true,
+    hideFromOverrides: true,
+    process: (value) => value,
+  });
+}
+
+export function addAxisConfig(builder: FieldConfigEditorBuilder<AxisConfig>, hideScale?: boolean) {
+  builder
```
torkelo, line 185: nice to share these! :heart:

public/app/plugins/panel/barchart/module.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export const plugin = new PanelPlugin<BarChartOptions, BarChartFieldConfig>(BarChartPanel)
+  .useFieldConfig({
+    standardOptions: {
+      [FieldConfigProperty.Color]: {
+        settings: {
+          byValueSupport: false,
+        },
+        defaultValue: {
+          mode: FieldColorModeId.PaletteClassic,
```
torkelo, line 21: hopefully, by v8 we will have a new better default palette (don't like that the first two colors are green and yellow)

torkelo (approved):
Looks great! 

Of course, more work is needed to reduce duplication, etc, and a unit tests, gdev test dashboard etc. 

Created an epic to keep track of more work related to this panel
https://github.com/grafana/grafana/issues/30741

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
ryantxu: I removed this export for now since we are not using it.  I think the click behavior we want here is opposite to the graph, but we can sort that out in 7.5+

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
ryantxu: discussed, and yes

packages/grafana-ui/src/components/BarChart/BarChart.tsx
ryantxu: I added this as a TODO on https://github.com/grafana/grafana/issues/30741

packages/grafana-ui/src/components/BarChart/BarChart.tsx
ryantxu: I added this as a TODO on  https://github.com/grafana/grafana/issues/30741

I don't see any small changes that make what we have any better/different

maxdd:
Is it already possible to use it with flux query?

PR #30324, opened 2021-01-15
Panels: remove beta flag from stat and bargauge panels

ryantxu:
These have not changed in a while, and feel pretty stable:
image

for sure a different category from Timeseries panel!


torkelo:
Think these beta badges have never been visible before , must have come after I added support for beta badges and these had beta states in plugin json :)

torkelo (approved):

Issue #5541, opened 2016-07-08
[Feature request] Add "include zero" to graph Y-axis


kipe:
* **I'm submitting a ...**
* [ ] Bug report
* [x] Feature request
* [ ] Question / Support request: **Please do not** open a github issue. [Support Options](http://grafana.org/support/)

Please include this information:

* What Grafana version are you using? 3.0.4
* What datasource are you using? Influx DB 0.13
* What OS are you running grafana on? Linux

When drawing graphs currently the only possible way to control the Y-axis is to set value limits or use automatic limits. While this is enough for most use cases, it would be nice to have a feature to always include zero to the Y-axis.

For example, when viewing a graph with positive values it's quite normal to set Y-min to 0. By setting this, the graph can be "normalized" to always include 0 value. However, when there are some points in which the value is below zero, the values are lost from the graph with this solution.

Couple screencaps to demonstrate the issue:
[without Y-min, only positive values](https://kipe.kapsi.fi/rnd/grafana_issue/include_zero1.png)
[without Y-min, a few negative values](https://kipe.kapsi.fi/rnd/grafana_issue/include_zero2.png)
[Y-min = 0, only positive values](https://kipe.kapsi.fi/rnd/grafana_issue/include_zero3.png)
[Y-min = 0, a few negative values](https://kipe.kapsi.fi/rnd/grafana_issue/include_zero4.png)

The logic of "include zero" as pseudo-code:

```js
if (include_zero) {
if (min(y) > 0) {
y_axis_min = 0;
}
if (max(y) < 0) {
y_axis_max = 0;
}
}
```

mgaugusch:
I'd also really like to see this implemented, especially for temperature graphs (which may include values less than zero in my area)

Issue #4156, opened 2016-02-26
Feature Request:  Y axis auto min/max that aligns min/max across many graphs


sslupsky:
Hi,  I am wondering if others would find it helpful to be able to configure the "Y-Span" in the Axes & Grid settings of the Graph panel?  That is, in addition to specifying the Y-Max or Y-Min, there is a third box for selecting the "Y-Span".  By specifying a span here, the panel is rendered using the span specified.

This would help alleviate the situation where two graphs are rendered with vastly different spans because one graph has a large gradual change in magnitude (graph A) and the other graph has a small change on a basically flat line (graph B).   Generally, the eye would be drawn to the small change in graph B because the auto scale would make this look like a large change but when you look at the Y axis it becomes apparent the span is much smaller than graph A.  If the "average" value of these two graphs is the same, then setting the Y-Max and the Y-Min would suffice.  But if one graph is offset compared to the other, and that offset changes over time, then using Y-Max and Y-Min is not practical because you have to set the span so large as to make it difficult to see any changes.  So, it would be preferable to view the two graphs with the same span.

torkelo:
can you describe more how the span would work ? How would it be implemented?

CamJN:
There are a few options that might work, one limit could be left as auto (max or min) and the other could be a fixed offset from it (say y-max:auto y-min: max - 10) another option would be setting the max and min as relative to the average value (say y-max:avg + 10 y-min: avg - 10).

thoj:
I understand that this function is probably only useful for very few people. But I've been hacking on this a bit. And I've come up with something that works for me. The commit in my fork adds a Y-Span box to the Axes tab.

Where X is a integer or float:
~X Span X around average
=X Span X around current value
>X Span is atleast X
<X Span is clamped to X

https://github.com/thoj/grafana/commit/7dcdccbd42e9f63b7388e439f7194fe7ead8039a

Example of why i need it: 
![y-span](https://cloud.githubusercontent.com/assets/153440/16985695/7ce60908-4e82-11e6-9b8d-60334877008b.PNG)

Any interest in a PR?

tomnewport:
:+1: This would make me happy

hahnjo:
As my issue #11783 has been closed as a duplicate of this I wanted to see if somebody is working on this? The currently referenced PR's (#5720 and the (merged) refactor in #6051) only seem to deal with extending functionality of `Y-Min` and `Y-Max`, but not with aligning the axes across different graphs.

[...]

PR #30326, opened 2021-01-16
GraphNG: Adds support for soft min and soft max for better auto-scaling axis limits

### LINKED ISSUE #979: Graph: Limits on Y auto scale (min max)

### LINKED ISSUE #5984: Always keep thresholds visible in graphs ('minimum' maximum)
- **I'm submitting a ...**
- [ ] Bug report
- [x] Feature request
- [ ] Question / Support request: **Please do not** open a github issue. [Support Options](http://grafana.org/support/)

- What Grafana version are you using?
  3.1.1
- What datasource are you using?
  Elasticsearch 2.x
- What OS are you running grafana on?
  RHEL 6.x
- What did you do?
  Added thresholds to a graph which could go far above the upper threshold
- What was the expected result?
  Threshold lines always visible
- What happened instead?
  Threshold lines not visible until datapoints exceed them.

I would like to have the option to keep thresholds visible on the graph at all times _without_ needing to set a fixed maximum in the panel settings.

The graph plots the age of replicated data on the y-axis, and our thresholds are 20 minutes and 30 minutes. We would like the upper threshold to be visible all the time to highlight how close (or ideally how far) the age of data is from the threshold, but also to allow the graph to scale beyond the threshold when required.

Effectively this would mean setting a 'minimum' on the maximum, or padding the upper value in the graph to ensure that the threshold is always visible.

Currently I am doing this by using graphite to plot a single value, but this makes the tooltip show the date on the far left of the x axis regardless of where you hover on the graph.

leeoniya:
softMin & softMax can prevent blips from turning into mountains when the data is mostly flat, and hardMin/hardMax can prevent intermittent spikes from flattening useful detail by clipping the spikes past a defined point. this PR includes the work started by @ryantxu in https://github.com/grafana/grafana/pull/30238 and should close at least #979 and #5984. (maybe more).

also, soft limits can be used to "always show 0" or "only autoscale positive values". (in most cases)

Closes #979 
Closes #5984.

this includes a bump for uPlot to 1.6.1 that fixes a bug in the default y auto-ranging behavior. i checked all the test panels that have stable y ranges (the y axis ticks panel) and there is only a small difference that improves last panel by bringing the 0 into range.

it should be noted that there is no support for log scales here since the ranging functions are completely different. i'm inclined to wait until someone complains before putting any work into it because very often enabling log scales serves the same purpose as setting soft limits on linear scales (to squash spikes).

below...

top panels are the default ranging behavior. the wide version is the full dataset, and each smaller panel is different interesting subset of the same data. the bottom copies of the same panels all have the following limits: `Hard: 1.5 - 6.0, Soft: 2.5 - 5.0`:

image

### DIFF
--- a/devenv/dev-dashboards/panel-graph/graph-ng-soft-limits.json
+++ b/devenv/dev-dashboards/panel-graph/graph-ng-soft-limits.json
@@ -1,0 +1,2238 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "links": [
+    {
+      "icon": "external link",
+      "tags": ["gdev", "graph-ng"],
+      "type": "dashboards"
+    }
+  ],
+  "panels": [
+    {
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axis": {
+              "grid": true,
+              "label": "",
+              "side": 3,
+              "width": 60
+            },
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "bars": {
+              "show": false
+            },
+            "drawStyle": "line",
+            "fill": {
+              "alpha": 0
+            },
+            "fillGradient": "none",
+            "fillOpacity": 15,
+            "hideFrom": {
+              "graph": false,
+              "legend": false,
+              "tooltip": false
+            },
+            "line": {
+              "show": true,
+              "width": 1
+            },
+            "lineInterpolation": "linear",
+            "lineWidth": 1,
+            "nullValues": "null",
+            "pointSize": 5,
+            "points": {
+              "radius": 4,
+              "show": false
+            },
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "showPoints": "auto",
+            "spanNulls": false
+          },
+          "links": [],
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          },
+          "unit": "none"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 11,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 7,
+      "options": {
+        "legend": {
+          "asTable": false,
+          "calcs": [],
+          "displayMode": "list",
+          "isVisible": true,
+          "placement": "bottom"
+        },
+        "tooltipOptions": {
+          "mode": "single"
+        }
+      },
+      "pluginVersion": "7.4.0-pre",
+      "targets": [
+        {
+          "alias": "",
+          "csvWave": {
+            "timeStep": 60,
+            "valuesCSV": "0,0,2,2,1,1"
+          },
+          "lines": 10,
+          "points": [],
+          "pulseWave": {
+            "offCount": 3,
+            "offValue": 1,
+            "onCount": 3,
+            "onValue": 2,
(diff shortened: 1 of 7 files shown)


torkelo (approved):
Awesome! Lots will be happy with this new capability!

leeoniya:
@torkelo 

i think this also closes https://github.com/grafana/grafana/issues/4156

and maybe closes https://github.com/grafana/grafana/issues/5541

torkelo:
Yea, both looks like we can close with this change😃😃

adamkdean:
@leeoniya if this closes #4156, does that mean that we can now have say four graphs side by side which all have their min-max synchronised?

For example, imagine four temperature sensor graphs with an auto y-axis scale of 30 ºC to 40º C and values around 35 ºC. If one of the sensor values were to spike to 50 ºC and the y-axis range of it's respective graph to increase from 30 ºC to 50 ºC to cover the spike, would the other three graphs also be able to have their y-axes synchronised? If not, then I don't believe this closes #4156.

leeoniya:
yeah, it's not a 100% fix but significantly improves the situation. i'm still pretty new to Grafana, so not sure how realistic a cross-panel y axis sync is.

coincidentally, there's a recent similar request in uPlot for this that i'm gonna work through: https://github.com/leeoniya/uPlot/issues/433. disabling uPlot's built-in "zoom on selection end" looks like prerequisite, which Grafana already does. but this is more complicated because it needs to cross-sync specific units across specific panels, not just on zoom but also on data updates. and there needs to be some UI for defining _which_ panels & units to sync.

@torkelo can re-open if he thinks this is a realistic thing for the future.

adamkdean:
Thanks both. The synchronisation of graphs would be super useful, especially when you're monitoring grouped yet separate graphs and want to be able keep things visually consistent, so I think #4156 still remains an "in demand" request. That being said, this ticket is a great feature, just not the same one 😄 .

simPod:
According to the diff, this is not available to the graph panel, right?

leeoniya:
the graph panel is on its way out and won't be getting much love going forward.

its replacement is called "Time series", which is where this will be available.

mmatrosov:
Where can I find the exact formula on how these values are used? It is not [the docs](https://grafana.com/docs/grafana/latest/visualizations/time-series/change-axis-display/#soft-min-and-soft-max), they are ambiguous. 

My understanding is the following:

```
y_max = clamp(max_over_data, soft_max, hard_max),
y_min = clamp(min_over_data, hard_min, soft_min),
where clamp(x, a, b) = max(a, min(x, b))
```

But I'm not sure what happens if `soft_max` is greater than `hard_max` or `soft_min` is less than `hard_min`.

bwims:
Hi,

I'm running 8.4.4  Is that a high enough version?  I cannot see anything related to hard and soft in the axes definitions.

My interest is in always displaying the threshold line.  (FYI I am displaying barometric pressure and want the "normal" threshold to be visible.

I got sent over here because the original question asking for threshold visibility was closed pointing here.

Any help / advice gratefully received!

leeoniya:
> I cannot see anything related to hard and soft in the axes definitions.

the "hard" limits are just `min` and `max`. and the soft limits are `softMin` and `softMax`.

image

bwims:
Hi, thanks for replying.  I had my pressure graph on "Graph", not "Time series" which looks like it got upgraded  to remain the default.   So that solves one problem.  The other is that I don't quite understand what "soft" means.   Do I put the min and max at the extremes and let the system work it out?  Does it automatically take note of the threshold? 

Thanks!

leeoniya:
> Edit: as far as I can see, it behaves just like the previous min, max.

if you _only_ set min and max, then yes. they are treated as hard limits. we did not want to break that.

for your case, you can omit `min` and `max` and only set `softMin` and/or `softMax` to values that encompass minimum range you want to see. if your data exceeds those, then it will autoscale up to the hard limits (or to infinity if hard limits are not set).

bwims:
I see!

Many many thanks!  That's great!

B.

PR #30328, opened 2021-01-16
Badge: Design improvement & reduce contrast

torkelo:
I felt the badges where to bright distracting from the main icon/text.

Before:  
Screenshot from 2021-01-16 14-02-58

After:
Screenshot from 2021-01-16 14-01-40

 
Before Light
Screenshot from 2021-01-16 14-04-37

After Light:
Screenshot from 2021-01-16 14-06-26

### DIFF
--- a/packages/grafana-ui/src/components/Badge/Badge.tsx
+++ b/packages/grafana-ui/src/components/Badge/Badge.tsx
@@ -62,7 +62,7 @@ const getStyles = stylesFactory((theme: GrafanaTheme, color: BadgeColor) => {
       .setAlpha(0.1)
       .toString();
     borderColor = tinycolor(sourceColor)
-      .lighten(15)
+      .lighten(25)
       .toString();
     textColor = tinycolor(sourceColor)
       .darken(15)
--- a/packages/grafana-ui/src/components/Badge/Badge.tsx
+++ b/packages/grafana-ui/src/components/Badge/Badge.tsx
@@ -49,23 +49,23 @@ const getStyles = stylesFactory((theme: GrafanaTheme, color: BadgeColor) => {
 
   if (theme.isDark) {
     bgColor = tinycolor(sourceColor)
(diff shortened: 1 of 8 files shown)


packages/grafana-ui/src/components/Badge/Badge.tsx
```diff
@@ -49,23 +49,23 @@ const getStyles = stylesFactory((theme: GrafanaTheme, color: BadgeColor) => {
 
   if (theme.isDark) {
     bgColor = tinycolor(sourceColor)
-      .darken(38)
+      .setAlpha(0.1)
```
Clarity-89, line 52: I feel like these colors should come from the design system, instead of being arbitrarily set. Wonder if they could be aligned with the [new tag colors](https://www.figma.com/file/I34UgAga9JvqjlxfKYZnve/Design-System?node-id=453%3A971)?

hugohaggmark (approved):
LGTM!

packages/grafana-ui/src/components/Badge/Badge.tsx
torkelo: Maybe, these are from the design system, but programmatic alpha. This is what Luke advocates as well (to have fundamental colors and then programmatic changes for different use cases like hover). But maybe this should change should be in a mixin. If we have this pattern in more places (IE programmatically turn a palette color into subtle background then I think this should be mixin), and if it's for a common background we use for many components then it should be in the theme, but here it's just an automatic background based on a named color 

This badge should work for many colors (see GitHub labels) and not sure how to deal with that in terms of the theme, currently the named colors palette is outside the theme (but changes slighly depending on theme)

Clarity-89: Yeah, I think as long as these color adjustment functions are predictable and have a common pattern, it should work well.

PR #30333, opened 2021-01-16
ToolbarButton: New emotion based component to replace all navbar, DashNavButton and scss styles

torkelo:
The navbar styles are among the oldest and messiest and reused in and overriden in different parts here is a first stab at migrating them to a more robust react & emotion component.

* [x] Main button style
* [x] Only icon
* [x] Icon + text
* [x] Svg image + text
* [x] Right side angle up/down for open/closed state (Used by SelectButtons and new [VisualisationButton](https://github.com/grafana/grafana/pull/30329) I am working on)
* [x] Tight & attached look for refresh button
* [x] Adds ButtonGroup component that works with both normal Button and ToolbarButton to group buttons close together (without spacing merging them) 
* [x] I tried to make the toolbar a variant of the main button and ended up rewrtiting the main button code quite a bit, but then gave up. They are separate things. But keeping some of the refactorings of the main button component
* [x] Big update to main Button storybook

For follow-up PR: 
* [ ] ResponsiveButton
* [ ] Replace dashboard buttons
* [ ] Handle all responsive hiding rules in dashboard
* [ ] Replace explore buttons
* [ ] Handle explore responsive rules
* [ ] Add a active (orange outline) state & style (used in explore live mode)  

Storybook story:
Screenshot from 2021-01-18 07-58-53

Dam, thought I had solved every use case but looks like I also need to support:
Screenshot from 2021-01-17 18-46-16

The above is the trickiest, combines the old sass button classes & the navbar-button classes. Makes me think that this toolbar button should be a variant of normal Button, a bit tricky as this button variant has different paddings & color 

Screenshot from 2021-01-17 18-46-32

Updated main button storybook: 
Screenshot from 2021-01-17 22-42-00


packages/grafana-ui/src/components/Button/Button.story.tsx
```diff
@@ -26,11 +30,53 @@ export default {
...
 };
 
-export const Simple: Story<ButtonProps> = ({ children, ...args }) => <Button {...args}>{children}</Button>;
-Simple.args = {
-  variant: 'primary',
-  size: 'md',
-  disabled: false,
-  children: 'Button',
-  icon: undefined,
+export const Variants: Story<ButtonProps> = ({ children, ...args }) => {
```
Clarity-89, line 33: I think the point of the story is to provide controls for the button and then allow users to modify them to be able to see different variants. Removing controls and showing only some variants for the buttons instead is quite limited and doesn't showcase all the possible options. I think a better place for this Variants would be in the docs, i.e. the mdx file where we can show different variants and add a bit of description of where and how they'd be used. The canvas on the other hand should have mostly one component, which can then be modified via controls.

packages/grafana-ui/src/components/Button/Button.tsx
```diff
@@ -186,11 +75,155 @@ export const LinkButton = React.forwardRef<HTMLAnchorElement, ButtonLinkProps>(
...
+    img: css`
+      width: 16px;
+      height: 16px;
+      margin-right: ${theme.spacing.sm};
+      margin-left: -${theme.spacing.xs};
+    `,
+    icon: css`
+      margin-left: -${padding / 2}px;
+      ${!iconOnly && `margin-right: ${padding / 2}px;`}
+      ${iconOnly && `margin-right: -${padding / 2}px;`}
```
Clarity-89, line 146:
```suggestion
      margin-right: ${(iconOnly ? -padding : padding) / 2}px;
```

packages/grafana-ui/src/components/Button/Button.tsx
```diff
@@ -186,11 +75,155 @@ export const LinkButton = React.forwardRef<HTMLAnchorElement, ButtonLinkProps>(
...
+    `,
+    content: css`
+      display: flex;
+      align-items: center;
+      flex-direction: row;
+      align-items: center;
```
Clarity-89, line 152: Duplicate `align-items` declaration.

packages/grafana-ui/src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,55 @@
...
+export interface Props extends HTMLAttributes<HTMLDivElement> {
+  noSpacing?: boolean;
+}
+
+export const ButtonGroup = forwardRef<HTMLDivElement, Props>(({ noSpacing, children, ...rest }, ref) => {
+  const styles = useStyles(getStyles);
+  const className = cx({
+    [styles.wrapper]: !noSpacing,
+    [styles.wrapperNoSpacing]: noSpacing,
+  });
```
Clarity-89, line 15:
```suggestion
  const className = noSpacing ? styles.wrapperNoSpacing : styles.wrapper;
```
Not sure we need `cx` here, the styles seem to be mutually exclusive so we can just pick one based on condition.

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    const buttonStyles = cx(
+      {
+        [styles.button]: true,
+        [styles.buttonFullWidth]: fullWidth,
+        [styles.narrow]: narrow,
+      },
+      className
+    );
```
Clarity-89, line 41:
```suggestion
    const contentStyles = cx(styles.content, {
      [styles.contentWithIcon]: !!icon,
      [styles.contentWithRightIcon]: isOpen !== undefined,
    });

    const buttonStyles = cx(
      styles.button,
      {
        [styles.buttonFullWidth]: fullWidth,
        [styles.narrow]: narrow,
      },
      className
    );
```
I think `cx` works the same way as `classNames`, so if the condition is always the true the styles can just be added outside of the object.

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const body = (
+      <button ref={ref} className={buttonStyles} {...rest}>
+        {icon && <Icon name={icon} size={'lg'} />}
+        {imgSrc && <img className={styles.img} src={imgSrc} />}
+        {children && <span className={contentStyles}>{children}</span>}
+        {isOpen === false && <Icon name="angle-down" />}
+        {isOpen === true && <Icon name="angle-up" />}
```
Clarity-89, line 49:
```suggestion
      {isOpen ? <Icon name="angle-up" /> : <Icon name="angle-down" />}
```
I think using a ternary here makes the code cleaner.

packages/grafana-ui/src/components/Forms/commonStyles.ts
```diff
@@ -86,30 +87,29 @@ export const inputSizesPixels = (size: string) => {
...
-  const { hasText, hasIcon, size } = props;
-  const { spacing, typography, height } = props.theme;
+export function getPropertiesForButtonSize(size: ComponentSize, theme: GrafanaTheme) {
+  const { typography, height } = theme;
 
   switch (size) {
     case 'sm':
       return {
-        padding: `0 ${spacing.sm}`,
+        padding: 8,
```
Clarity-89, line 95: I think we can use `SPACING_BASE` from `defaults.ts` as the base here, so the padding stay consistent with the ones from the design system .

Clarity-89 (commented):
I think the changes look good! My main point is about how we use canvas in storybook, this is smth we haven't done in all the cases but I think with the introduction of controls the canvas should be just one component that can be modified via controls and if we want to show some possible configurations for that component, they are better placed in the mdx files to be displayed in the Docs tab.

packages/grafana-ui/src/components/Button/Button.tsx
```diff
@@ -131,26 +16,28 @@ type CommonProps = {
...
+      children,
     });
 
     return (
       <button className={cx(styles.button, className)} {...otherProps} ref={ref}>
-        <ButtonContent icon={icon} size={otherProps.size}>
-          {children}
-        </ButtonContent>
+        {icon && <Icon name={icon} size={size} className={styles.icon} />}
+        {children && <span className={styles.content}>{children}</span>}
```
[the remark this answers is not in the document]
torkelo: I found it just complicating things, and the reuse was minimal,  it split the style generation into two files & getStyles and made things harder to follow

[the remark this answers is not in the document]
torkelo: Could consider adding it back and passing the styles from the main button styles maybe if there is more need to share the component, but to only share 2 lines it felt not worth it

packages/grafana-ui/src/components/Button/Button.story.tsx
torkelo: I find this type of story really bad for development & for developers to see examples. I find showing all varaiants at the same time is so much better for development and to showcase examples.

packages/grafana-ui/src/components/Button/Button.tsx
torkelo: ah, thanks

packages/grafana-ui/src/components/Button/ButtonGroup.tsx
torkelo: ah, great spotted!

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
torkelo: I know, just find it a bit more readable.

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
torkelo: just find ? less readable in this case

packages/grafana-ui/src/components/Button/Button.story.tsx
Clarity-89: But we can achieve exactly the same showcase by moving the components to the docs tab while keeping canvas as a sort of playground to try out various options.

packages/grafana-ui/src/components/Forms/commonStyles.ts
torkelo: added spacing.base to the theme type so I could use it here

packages/grafana-ui/src/components/Button/Button.story.tsx
torkelo: the playground was just changing icon, not that valuable a playground. But yea a separate playground could be added if that is useful. But I prefer the canvas showing the important state for easy development & verification

packages/grafana-ui/src/themes/default.ts
```diff
@@ -75,6 +75,7 @@ const theme: GrafanaThemeCommons = {
     xxl: '1440px',
   },
   spacing: {
+    base: SPACING_BASE,
```
torkelo, line 78: @Clarity-89 added base to the theme type to make it accessible code that uses the theme (really with all the spacings where numbers and not strings, this mixing of strings with 'px' and numbers is so messy)

packages/grafana-ui/src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+      border-radius: 0;
+      border-right: 0;
+
+      &:last-child {
+        border-radius: 0 3px 3px 0;
+        border-right: 1px solid ${theme.colors.border2};
+      }
+
+      &:first-child {
+        border-radius: 3px 0 0 3px;
```
Clarity-89, line 50:
```suggestion
        border-radius: ${theme.border.radius.md} 0 0 ${theme.border.radius.md};
```

Clarity-89 (approved):
Looks good! Just a minor comment.

packages/grafana-ui/src/themes/default.ts
Clarity-89: Yeah, I think this is a good workaround 👍

packages/grafana-ui/src/components/Button/ButtonGroup.tsx
torkelo: ah, nice. This should actually be sm as it looks like buttons use sm radius

torkelo:
thanks for the reviews @Clarity-89 !

PR #30334, opened 2021-01-17
Grafana-UI: Fix HTML character entity error

baijunyao:
Fix HTML character entity error

image


Clarity-89 (approved):
Great catch! Thank you for the contribution! 🎉

PR #30346, opened 2021-01-18
Docs: Add jwtTokenAuth to plugin metadata schema

marcusolsson:

Adds missing `jwtTokenAuth` section to the plugin metadata schema. 

From what I can see, `jwtTokenAuth` has the same schema as `tokenAuth`. Let me know if there are any semantic differences in how we use the properties.


ryantxu (approved):
i had no idea :)

PR #30350, opened 2021-01-18
Grafana-ui: Fix context menu item always using onClick instead of href

aocenas:
fix after https://github.com/grafana/grafana/pull/30141
There is always onClick handler from the parent component. Moved the code checking special keys there.


hugohaggmark (approved):
LGTM!

dprokop (approved):

dprokop (approved):

PR #30351, opened 2021-01-18
Loki: Label browser

davkal:
- replaces stream cascader widget which made it hard to find relevant streams
- multi-step selection allows for selecting a couple of labels first, then find the relevant values
- supports facetting which makes impossible label combinations hard to choose
- removes useLokiLabels which was used to manage the cascader
- shows errors when loading more than 100 labels, or 10000 label values.
- button to validate selector before using it
- button to run selector as metrics query to show logs volume
- taking time range for metadata requests directly from timeSrv (not sync'd yet with timepicker changes, will do in separate PR)

https://user-images.githubusercontent.com/859729/104908112-4602a400-5986-11eb-9a26-82d5b714198f.mov


public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,526 @@
...
+import LokiLanguageProvider from '../language_provider';
+import { css } from 'emotion';
+import store from 'app/core/store';
+import { FixedSizeList } from 'react-window';
+
+import { GrafanaTheme } from '@grafana/data';
+import { LokiLabel } from './LokiLabel';
+
+// Hard limit on labels to render
+const MAX_LABEL_COUNT = 100;
```
ivanahuckova, line 22: Is there a reason for hard limit? Also, isn't the limit too low (I checked and our loki-ops has around ~70-80 labels, but maybe some of our users have even more)? 

**Blocker:**
If we are slicing, I think that we should let our users know that we have reached the limit and not all labels are visible. 

**Maybe next iteration/during the beta:**
We could maybe add a search to specify label names - so we wouldn't slice the labels, but just show first ~100 and then users could specify more which labels are they interested.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,526 @@
...
+    linkColor: theme.colors.text,
+    linkColorHover: white,
+    wrapperBg: dark2,
+    wrapperShadow: black,
+    itemColor: white,
+    groupLabelColor: theme.colors.textWeak,
+    itemBgHover: dark7,
+    headerBg: dark1,
+    headerSeparator: dark7,
+  };
```
ivanahuckova, line 131: **Maybe next iteration/during the beta**:
There are styles that are not used (linkColor, linkColorHover,...). This should be cleaned up and we should have here only styles that are used.

public/app/plugins/datasource/loki/components/LokiLabel.tsx
```diff
@@ -0,0 +1,122 @@
...
+        {...rest}
+      >
+        <Highlighter textToHighlight={text} searchWords={[searchTerm]} highlightClassName={styles.matchHighLight} />
+      </span>
+    );
+  }
+);
+
+const getLabelStyles = (theme: GrafanaTheme, name: string) => {
+  const colors = ['#FF7368', '#459EE7'];
```
ivanahuckova, line 64: I see that these colors are used as a border colors for the tags. I am wondering if we can't use a red & blue color from our palette/colors because this doesn't seem consistent. Or we can use `getTagStyles` and style this as tags.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+  });
+  it('returns an empty selector for selected labels with no values', () => {
+    const labels: SelectableLabel[] = [{ name: 'foo', selected: true }];
+    expect(buildSelector(labels)).toEqual('{}');
+  });
+  it('returns an empty selector for one selected label with no selected values', () => {
+    const labels: SelectableLabel[] = [{ name: 'foo', selected: true, values: [{ name: 'bar' }] }];
+    expect(buildSelector(labels)).toEqual('{}');
+  });
+  it('returns a simple selector from a selected label with a selected value', () => {
```
ivanahuckova, line 21: I would also add a scenario with multiple labels and values `=~`

ivanahuckova (commented):
As mentioned, I really like this approach. See below some comments and feedback: 

1. Clicking outside of labels popover should close the labels popover. Otherwise it is in front of everything.
image
2. I don't think we have a pattern for loading where colors change red -> blue. It seems not consistent, but not sure what would be the better option. I would check with UX team if we don't have some loading pattern that could be applicable.
3. I think that max width of browser should be the width of screen and when selecting label values, it should wrap. 
4. When I select label value, I think it would make sense to add spinner/visually show that the labels are being processed and that we will show only the relevant ones.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,526 @@
...
+    const { languageProvider } = this.props;
+    if (languageProvider) {
+      const selectedLabels: string[] = store.getObject(LAST_USED_LABELS_KEY, []);
+      languageProvider.start().then(() => {
+        const labels: SelectableLabel[] = languageProvider
+          .getLabelKeys()
+          .slice(0, MAX_LABEL_COUNT)
+          .map(label => ({ name: label, selected: selectedLabels.includes(label), loading: false }));
+        this.setState({ labels }, () => {
+          this.state.labels.forEach(label => {
```
aocenas, line 297: I see this on multiple places, couldn't we just fire the `this.fetchValues` on `componentDidUpdate`? Comparing the previous `labels` with current ones and fire the load depending on that? I assume we want to do that every time labels or their selected state changes.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,526 @@
...
+  async fetchValues(name: string) {
+    const { languageProvider } = this.props;
+    this.updateLabelState(name, { loading: true });
+    try {
+      const values: FacettableValue[] = (await languageProvider.getLabelValues(name))
+        .slice(0, MAX_VALUE_COUNT)
+        .map(value => ({ name: value }));
+      this.updateLabelState(name, { values, loading: false });
+    } catch (error) {
+      console.error(error);
```
aocenas, line 333: Is error here likely?  maybe we could use the app events to show an error alert to the user.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,526 @@
...
+    let selectedLabels = labels.filter(label => label.selected && label.values);
+    if (searchTerm) {
+      // TODO extract from render() and debounce
+      try {
+        matcher = new RegExp(searchTerm.split('').join('.*'), 'i');
+        selectedLabels = selectedLabels.map(label => ({
+          ...label,
+          values: label.values?.filter(value => value.selected || matcher.test(value.name)),
+        }));
+      } catch (error) {}
```
aocenas, line 372: What error are we catching here? it seems like only error here could be programatic one.

aocenas:
Nice, I think @ivanahuckova did a more thorough review than I did though :D

davkal:
I've addressed the blockers and other feedback, PTAL

ivanahuckova (commented):
Hello, hello. | love the work that you have done around time ranges. The code in language_provider is much cleaner. Here is (mostly) UI/UX feedback, plus I've left a feedback in the code as well. 
1. I think that those borders between label values column made browser look cleaner. I think it would be nice to have them back. 🙂  
2. When I use Validate selector, I got jumpy visualisation (see below)
Kapture 2021-01-20 at 14 18 01
3. Clicking outside doesn't close the Label Browser and label browser is often in the way.
image
4. When I scroll, label browser "dances around":
Kapture 2021-01-20 at 14 26 42
5. Error `Too many values for...` doesn't clean up when new label is selected/that label with too many values is deselected. 
image

davkal:
@ivanahuckova I think this is ready for another review. PTAL
The only thing not addressed is the timepicker sync, will do this in a separate PR.

ivanahuckova (commented):
I have just tested this out (didn't review code very closely), but I am starting to have a feeling that we should replace popover/popup browser with something else/other UX, as it popup is quite often behaving unexpectedly and it looks buggy.

When I was testing it now, it was still jumpy, didn't correctly resize on smaller screens, wasn't positioned correctly on split screen. I feel that using pattern that we have currently with the **Help** dropdown, could save us a lot of smaller UI/UX fixes in the future as it is part of the query row - would have correct size and no jumpiness.

image

What are your thoughts @davkal? 

See the video of the jumpy/unexpected behaviour:

https://user-images.githubusercontent.com/30407135/108072385-1f0fbe80-7067-11eb-9ebf-1955a1b758f4.mov


Btw I love the rate function addition! In the future, we could even add more predefined functions! :)

ivanahuckova:
Also, when tested in dashboard panels the browser was bigger than available space and I wasn't able to access all label names:
image

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+const MAX_VALUE_COUNT = 10000;
+const MAX_AUTO_SELECT = 4;
+const EMPTY_SELECTOR = '{}';
+export const LAST_USED_LABELS_KEY = 'grafana.datasources.loki.browser.labels';
+
+type onChange = (selector: string) => void;
+
+export interface BrowserProps {
+  languageProvider: LokiLanguageProvider;
+  onChange: onChange;
```
ivanahuckova, line 22:
```suggestion
export interface BrowserProps {
  languageProvider: LokiLanguageProvider;
  onChange: (selector: string) => void;
```

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+  labels: SelectableLabel[];
+  searchTerm: string;
+  status: string;
+  error: string;
+  validationStatus: string;
+}
+
+interface FacettableValue {
+  name: string;
+  // facetted?: boolean; // True if value is possible given the selector
```
ivanahuckova, line 38:
```suggestion
```

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+        facets: undefined,
+      }));
+      return { labels, searchTerm: '', status: '', error: '', validationStatus: '' };
+    });
+    store.delete(LAST_USED_LABELS_KEY);
+  };
+
+  onClickLabel = (name: string, value: string | undefined, event: React.MouseEvent<HTMLElement>) => {
+    // Stop early before it reaches the ClickOutsideWrapper
+    event.stopPropagation();
```
ivanahuckova, line 211:
```suggestion
```

This can be probably removed as we are not using ClickOutsideWrapper anymore.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+    }
+    // Toggle selected state
+    const selected = !label.selected;
+    let nextValue: Partial<SelectableLabel> = { selected };
+    if (label.values && !selected) {
+      // Deselect all values if label was deselected
+      const values = label.values.map((value) => ({ ...value, selected: false }));
+      nextValue = { ...nextValue, facets: 0, values };
+    }
+    this.updateLabelState(name, nextValue, '', () => {
```
ivanahuckova, line 224: I think that readability would be improved if this was split  to something similar as you did in `    this.updateLabelState(name, { values }, '', () => this.doFacetting(name));`

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+      return;
+    }
+    // Toggle selected state
+    const selected = !label.selected;
+    let nextValue: Partial<SelectableLabel> = { selected };
+    if (label.values && !selected) {
+      // Deselect all values if label was deselected
+      const values = label.values.map((value) => ({ ...value, selected: false }));
+      nextValue = { ...nextValue, facets: 0, values };
+    }
```
ivanahuckova, line 223: I think if we created a helper function in `utils` to hide all of this logic, the readability would be improved.

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+      } else {
+        // Only need to facet when deselecting labels
+        this.doFacetting();
+      }
+    });
+  };
+
+  onClickValue = (name: string, value: string | undefined, event: React.MouseEvent<HTMLElement>) => {
+    // Stop early before it reaches the ClickOutsideWrapper
+    event.stopPropagation();
```
ivanahuckova, line 241:
```suggestion
```

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
ivanahuckova: +1

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
ivanahuckova: +1

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+    // Label is facetted out, hide all values
+    return { ...label, loading: false, hidden: !possibleValues, values: undefined, facets: 0 };
+  });
+}
+
+const getStyles = stylesFactory((theme: GrafanaTheme) => ({
+  wrapper: css`
+    background-color: ${theme.colors.bg2};
+    padding: ${theme.spacing.md};
+    max-width: 100rem;
```
ivanahuckova, line 98: Is there a reason why we don't do `width: 100%` as we are in query editor - we can use the full provided witdth. Here (if not that much labels) the width is too small and "ready" text is in the way of the button. 
image

ivanahuckova: Also I am not sure if `max-width: 100rem;` does anything because when I am adding labels to search in, the label browser window is growing and growing and pushing query editor outside the view, so I have to scroll.
image

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
```diff
@@ -0,0 +1,493 @@
...
+  selector: css`
+    font-family: ${theme.typography.fontFamily.monospace};
+    margin-bottom: ${theme.spacing.sm};
+  `,
+  status: css`
+    padding: ${theme.spacing.xs};
+    color: ${theme.colors.textSemiWeak};
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
ivanahuckova, line 122: I am not sure why, but these ellipsis are causing weird border. Not sure if we can do anything about it, but noticed it.

ivanahuckova (commented):
I really like this new view! I think it is much better than the popup. Really, really like this! 

I have left quite a lot of comments, but they are mostly very small/easy to fix/do it later/questions/ui feedback. This is really, really close to be merged. 🎉

public/app/plugins/datasource/loki/components/LokiLabelBrowser.tsx
ivanahuckova: image

public/app/plugins/datasource/loki/language_provider.ts
```diff
@@ -431,10 +425,9 @@ export default class LokiLanguageProvider extends LanguageProvider {
    * they can change over requested time.
    * @param name
    */
-  fetchSeriesLabels = async (match: string, absoluteRange: AbsoluteTimeRange): Promise<Record<string, string[]>> => {
```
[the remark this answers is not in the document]
davkal: This will change soon when Loki's label API takes a selector. Keeping this for now.

ivanahuckova (approved):
Let's get this to 7.5. 🚀 This is going to improve the experience of our Loki users a lot!  
I still have some feedback and I think someone from UX (@jessover9000 or @hemdrup) could review it as well so we could fully polish it before 7.5- stable. Yaayy! LGTM.

PR #30353, opened 2021-01-18
PostgreSQL: Allow providing TLS/SSL certificates as text in addition to file paths

### LINKED ISSUE #29958: Paste in an SSL cert for Postgres, like for MySQL
Allow me to paste an SSL cert in the Postgres data source config, instead of entering a file path.

**Why is this needed**:
The MySQL data source allows me to paste an SSL certificate into the UI, and it stores the cert in the Grafana DB. This is especially useful in hosted Grafana, where I don't have access to the filesystem. Postgres requires that I enter a file path for the SSL cert, which doesn't work in hosted Grafana. Would love to see the "paste" functionality for Postgres as an additional option.

ying-jeanne:

Fixes #29958 


pkg/tsdb/postgres/postgres.go
```diff
@@ -18,15 +21,19 @@ import (
...
 	tsdb.RegisterTsdbQueryEndpoint("postgres", newPostgresQueryEndpoint)
+
 }
 
-func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
-	logger := log.New("tsdb.postgres")
-	logger.Debug("Creating Postgres query endpoint")
+func cleanUpFiles(config sslAuthenticationConfig) {
+	if config.tmpFilesPath != "" {
+		defer os.RemoveAll(config.tmpFilesPath)
```
papagian, line 29: Lint complains because you should check the error returned value from `os.RemoveAll`.

Also, I think that you don't need `defer` here. Defer is used when we want to execute something after the surrounding function returns. So probably it's required [when cleanUpFiles is called](https://github.com/grafana/grafana/blob/58ade53a2b62245f41243b356e38275dc4cffcca/pkg/tsdb/postgres/postgres.go#L70).

aknuds1:
Should the PR still be in draft mode?

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,8 +76,61 @@ func escape(input string) string {
...
+			return config, err
+		}
+		currentPath, err = ioutil.TempDir(currentPath, "tmpSSLCerts")
+		if err != nil {
+			return config, err
+		}
+		config.tmpFilesPath = currentPath
+
+		if len(decrypted["tlsClientCert"]) > 0 {
+			clientCertification := path.Join(currentPath, "client.crt")
```
aknuds1, line 104: You should use `filepath.Join`, not `path.Join` since it's a filesystem path.

aknuds1 (changes requested):
You need to use filepath.Join for filesystem paths, and you should also ensure that the temporary directory consistently gets deleted as @papagian pointed out.

pkg/tsdb/postgres/postgres.go
aknuds1: Agree with @papagian that it needs to be ensured that the temporary directory gets removed. Regarding error check, you probably just want to log a warning if the temporary directory fails to be removed.

ying-jeanne: actually, I wanted to ask you guys if I put the directory name which is dynamic, there would be risks if the program crashed in the middle, there would be "trash files" forever. how about make it permanent? would it be a concurrent kind of problem?

aknuds1: It would require some more thought, but I don't think it should be a problem to create a temporary directory and ensure it gets cleaned up.

aknuds1:
Reading through the PR now.

public/app/plugins/datasource/postgres/config_ctrl.ts
```diff
@@ -62,6 +62,18 @@ export class PostgresConfigCtrl {
     this.showTimescaleDBHelp = !this.showTimescaleDBHelp;
   }
 
+  lssModeMapping() {
```
aknuds1, line 65:
```suggestion
  sslModeMapping() {
```

public/app/plugins/datasource/postgres/partials/config.html
```diff
@@ -30,36 +30,17 @@ <h3 class="page-heading">PostgreSQL Connection</h3>
   <div class="gf-form">
     <label class="gf-form-label width-10">SSL Mode</label>
     <div class="gf-form-select-wrapper max-width-15 gf-form-select-wrapper--has-help-icon">
-      <select class="gf-form-input" ng-model="ctrl.current.jsonData.sslmode" ng-options="mode for mode in ['disable', 'require', 'verify-ca', 'verify-full']" ng-init="ctrl.current.jsonData.sslmode"></select>
+	  <select class="gf-form-input" ng-model="ctrl.current.jsonData.sslmode" 
+	  ng-options="mode for mode in ['disable', 'require', 'verify-ca', 'verify-full']" 
+	  ng-init="ctrl.current.jsonData.sslmode" ng-change="ctrl.lssModeMapping()"></select>
```
aknuds1, line 35:
```suggestion
	  ng-init="ctrl.current.jsonData.sslmode" ng-change="ctrl.sslModeMapping()"></select>
```

aknuds1 (changes requested):
The PR removes the option to specify paths to certificate files. It should not do this, but provide pasting certificates as an _additional_ option. That's stated in the referenced issue, and I don't think we should remove the old functionality either.

marefr (commented):
In general looks good. There's some tricky things that needs to addressed related to multiple data sources instances, caching and race-conditions, see comments.

public/app/plugins/datasource/postgres/config_ctrl.ts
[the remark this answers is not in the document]
marefr: Noticed Arve suggested on this one so resolving this.

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,8 +77,61 @@ func escape(input string) string {
...
+	sslMode := strings.TrimSpace(strings.ToLower(ds.JsonData.Get("sslmode").MustString("verify-full")))
+	config := sslAuthenticationConfig{}
+	config.sslMode = sslMode
+	if sslMode != "disable" {
+		decrypted := ds.SecureJsonData.Decrypt()
+		currentPath, err := os.Getwd()
+		if err != nil {
+			return config, err
+		}
+		currentPath, err = ioutil.TempDir(currentPath, "tmpSSLCerts")
```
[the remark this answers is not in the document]
aknuds1: @marefr Do we need to keep the certificate files after establishing the connection? If we don't, I think we can just delete the temporary directory containing the files after the connection is made.

aknuds1:
@marefr Don't know if you noticed, but a fundamental issue with the PR is that support for specifying certificate files is removed. That isn't part of the original request, and I don't think we should do so.

aknuds1:
Since, to my knowledge, we should preserve the old functionality of specifying certificate files, there's the question of how to switch between certificate file paths and text (for pasting) in the UI. I think this UX question should be resolved before we program on the backend.

marefr:
@aknuds1 I suggested that to minimize the complexity/maintenance burden moving forward with the addition that existing file paths will continue to work until the data source is updated. Can be discussed, but this feature adds quite a lot of complexity I would say.

aknuds1:
@marefr I don't think it would be all that complex, especially considering we need the backwards compatibility anyway. Shouldn't such a big UX change be discussed more in depth before a decision is made?

aknuds1:
@torkelo has clarified that we need to keep supporting specifying certificate filesystem paths. So there has to be a switch in the UI to choose between filesystem paths and text inputs..

ying-jeanne:
discussed with mitch, agreed at best 7.4.0 and at worst 7.5.0 is still ok.

aknuds1:
> discussed with mitch, agreed at best 7.4.0 and at worst 7.5.0 is still ok.

Sounds good; let's avoid rushing it and make sure we get it right.

ying-jeanne:
> @torkelo has clarified that we need to keep supporting specifying certificate filesystem paths. So there has to be a switch in the UI to choose between filesystem paths and text inputs. I suggest we also add a field to the data source model that specifies whether to use filesystem paths or raw text, so it's completely deterministic from where to pick the certificates.

frontend done, now we support both authentication methods.

pkg/tsdb/postgres/postgres.go
```diff
@@ -49,10 +57,18 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
 
 	logger.Debug("Successfully connected to Postgres")
+	return endpoint, sslCertificationCfg, err
+}
+
+func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
+	logger := log.New("tsdb.postgres")
+	logger.Debug("Creating Postgres query endpoint")
+	endpoint, sslCertificationCfg, err := databaseConnection(datasource, logger)
+	defer cleanUpFiles(sslCertificationCfg)
```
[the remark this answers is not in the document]
ying-jeanne: put a permanent folder now, and the folder is kept until all the certification input text is removed. the only thing that is not done yet is to put the files into the data folder.

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,8 +77,61 @@ func escape(input string) string {
...
+		decrypted := ds.SecureJsonData.Decrypt()
+		currentPath, err := os.Getwd()
+		if err != nil {
+			return config, err
+		}
+		currentPath, err = ioutil.TempDir(currentPath, "tmpSSLCerts")
+		if err != nil {
+			return config, err
+		}
+		config.tmpFilesPath = currentPath
```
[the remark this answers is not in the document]
ying-jeanne: now uid of datasource is part of the folder name, each datasource would have their own temporary folder.

pkg/tsdb/postgres/postgres.go
```diff
@@ -56,15 +57,117 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+}
+
+func writeConnectionFiles(ds *models.DataSource, logger log.Logger) error {
+	sslMode := strings.TrimSpace(strings.ToLower(ds.JsonData.Get("sslmode").MustString("verify-full")))
+	decrypted := ds.SecureJsonData.Decrypt()
+	tlsCACert := decrypted["tlsCACert"]
+	tlsClientCert := decrypted["tlsClientCert"]
+	tlsClientKey := decrypted["tlsClientKey"]
+
+	if sslMode != "disable" {
```
papagian: Better checking if the `sslMode` is disabled and then return directly in order to reduce the nested block.

pkg/tsdb/postgres/postgres.go
```diff
@@ -56,15 +57,115 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+			log.Warnf("failed to delete temporary folder generated %v : %v", currentPath, err)
+		}
+	}
+	mutex.Unlock()
+
+	return nil
+}
+
 func generateConnectionString(datasource *models.DataSource, logger log.Logger) (string, error) {
+	sslConfigureMethod := strings.TrimSpace(strings.ToLower(datasource.JsonData.Get("sslconfiguremethod").MustString("file-content")))
```
papagian, line 159:
```suggestion
	sslConfigureMethod := strings.TrimSpace(strings.ToLower(datasource.JsonData.Get("sslconfiguremethod").MustString("file-path")))
```

I think for legacy reasons the default should be `file-path` instead, so that existing datasources, which don't have this field set (since it's new), will end up using the file paths.
In fact, I have tested and if the default is `file-content`, the generated string does not contains the certifications:
`DBUG[01-20|15:33:14] getEngine                                logger=tsdb.postgres connection="user='grafana' password='password' host='localhost' dbname='grafana' sslmode='verify-ca' port=5432"`

If I switch it to default to `file-path`, it works:
`DBUG[01-20|15:45:07] getEngine                                logger=tsdb.postgres connection="user='grafana' password='password' host='localhost' dbname='grafana' sslmode='verify-ca' port=5432 sslrootcert='devenv/docker/blocks/postgres_ssl/root.crt' sslcert='devenv/docker/blocks/postgres_ssl/client.crt' sslkey='devenv/docker/blocks/postgres_ssl/client.key'"`

public/app/plugins/datasource/postgres/partials/config.html
```diff
@@ -27,39 +27,63 @@ <h3 class="page-heading">PostgreSQL Connection</h3>
...
-    <info-popover mode="right-absolute">
-      To authenticate with a client SSL/TLS certificate, provide the path to the corresponding key file here.
-      Be sure that the file is <i>only</i> readable by the user executing the grafana process.
-    </info-popover>
+
+
+  <div class="gf-form-group" ng-if="ctrl.current.jsonData.sslmode != 'disable' && ctrl.current.jsonData.sslconfiguremethod === 'file-path'">
+	<div class="gf-form">
+		<h6>TLS Auth Details</h6>
+		<info-popover mode="header">TLS Certs file path are stored in the Grafana database.</info-popover>
```
marefr, line 57: Don't think this this tooltip is needed - it doesn't add any useful information and I already suggested to mention it in regards to SSL configuration method.

```suggestion
```

public/app/plugins/datasource/postgres/partials/config.html
```diff
@@ -27,39 +27,63 @@ <h3 class="page-heading">PostgreSQL Connection</h3>
...
+	</div>
+	<div class="gf-form max-width-30">
+		<span class="gf-form-label width-10">SSL Root Certificate</span>
+		<input type="text" class="gf-form-input gf-form-input--has-help-icon" ng-model='ctrl.current.jsonData.sslRootCertFile' placeholder="SSL/TLS root cert file"></input>
+		<info-popover mode="right-absolute">
+		If the selected SSL mode requires a server root certificate, provide the path to the file here.
+		Be sure that the file is readable by the user executing the grafana process.
```
marefr, line 64: Not sure the "Be sure that the file is readable by the user executing the grafana process." information is needed on each file. I tried to explain that in SSL configuration mode tooltip. Same applies for the other file path input fields below. WDYT?

```suggestion
```

public/app/plugins/datasource/postgres/partials/config.html
```diff
@@ -27,39 +27,63 @@ <h3 class="page-heading">PostgreSQL Connection</h3>
       />
     </div>
   </div>
+  <div class="gf-form">
+    <label class="gf-form-label width-10">SSL Configure Method</label>
+    <div class="gf-form-select-wrapper max-width-15 gf-form-select-wrapper--has-help-icon">
+	  <select class="gf-form-input" ng-model="ctrl.current.jsonData.sslconfiguremethod"
```
marefr, line 33: Nit. Would consider renaming `sslconfiguremethod` to `sslConfigurationMethod`.

marefr (commented):
Left some comments/suggestions on the frontend changes.

pkg/tsdb/postgres/xxxxxxgeneratedSSLCerts/ca.crt
```diff
@@ -0,0 +1 @@
+i am CA certification
```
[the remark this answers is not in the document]
ying-jeanne: it is generated by the unit test, just removed it and put the pattern into .gitignore.

pkg/tsdb/postgres/postgres.go
```diff
@@ -56,15 +57,124 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+	if err != nil {
+		return err
+	}
+
+	var mutex = &sync.Mutex{}
+
+	currentPath = filepath.Join(currentPath, ds.Uid+"generatedSSLCerts")
+	if _, err := os.Stat(currentPath); os.IsNotExist(err) {
+		mutex.Lock()
+		err = os.Mkdir(currentPath, 0600)
```
papagian, line 122:
```suggestion
		err = os.Mkdir(currentPath, 0700)
```
the directory should have execute permission otherwise it fails to access the directory contents.

papagian (changes requested):
I have a frontend issue when I create a datasource and select `Certificate Content` as `SSL Configuration Method` the file path fields are hiding but the text area field do not unhide:
Screenshot 2021-01-21 at 12 28 08 PM
I have to change the `SSL Mode` for them to appear.
I'm get the same behaviour both in Chrome and Firefox.

papagian:
It would be nice also to provide some examples of provisioning a datasource using certificate file paths and content.

aknuds1:
Having a look.

ying-jeanne:
> I have a frontend issue when I create a datasource and select `Certificate Content` as `SSL Configuration Method` the file path fields are hiding but the text area field do not unhide:
> Screenshot 2021-01-21 at 12 28 08 PM
> I have to change the `SSL Mode` for them to appear.
> I'm get the same behaviour both in Chrome and Firefox.

bug fixed.

pkg/tsdb/postgres/postgres.go
```diff
@@ -56,15 +57,124 @@ func newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndp
...
+	var generatedFilePath string
+	if tlsjs, ok := ds.JsonData.CheckGet(jsonFieldName); ok {
+		generatedFilePath = tlsjs.MustString("")
+	}
+	if fileContent != "" {
+		if generatedFilePath == "" {
+			generatedFilePath = filepath.Join(currentPath, certFileName)
+		}
+
+		if err := ioutil.WriteFile(generatedFilePath, []byte(fileContent), 0700); err != nil {
```
aknuds1, line 83: This file should definitely not be executable.
```suggestion
		if err := ioutil.WriteFile(generatedFilePath, []byte(fileContent), 0600); err != nil {
```

aknuds1 (changes requested):

aknuds1:
I will review more next week, too tired now.

aknuds1 (changes requested):
I've tested the PR thoroughly now, and made a [branch](https://github.com/grafana/grafana/tree/arve/29958) with improvements that I think ought to be merged into the PR. Let's chat if you're wondering about my changes in the branch. The branch has for example the improvement that certificate file system paths are validated, and an error is returned if they don't exist (and are non-empty). It also has more extensive debug logging to help diagnosing the two different TLS modes of the data source.

I think the approach of writing certificate files in a sub-directory of the current directory is problematic though. I would prefer that we use a temporary directory, if possible, otherwise we should use Grafana's data directory. Dumping files beneath the current directory isn't so good. I am also running into this problem that the certificate files get their permission changed to 0644 and I can't understand why (they get created with 0600). When this happens, the Postgres driver emits an error, because the permissions should be 0600 or stricter.

aknuds1:
@ying-jeanne should we really backport this to 7.4.x?? Only bug fixes should be backported in general, and last I heard this was headed for 7.5.

ying-jeanne:
from pq community discussion https://github.com/lib/pq/issues/489, and how windows permission is changed https://medium.com/@MichalPristas/go-and-file-perms-on-windows-3c944d55dd44. Seems pq package is not really checking the client key file permission. it should not be an issue even the windows file permission could only be 0666 or 0444.

aknuds1:
> from pq community discussion [lib/pq#489](https://github.com/lib/pq/issues/489), and how windows permission is behaving in os.Stat(file) https://medium.com/@MichalPristas/go-and-file-perms-on-windows-3c944d55dd44. Seems pq package is not really checking the client key file permission for windows. it should not be an issue even on windows os.Stat(file) could only return 0666 or 0444. I need to continually test on windows locally.

That's how I remember it too, that on Windows POSIX file permissions don't really work. I guess it's only read-only or read-write.

aknuds1 (approved):
Looks pretty good to me now, but I think the concurrency patterns need thinking some more about to ensure that we don't have race conditions. I haven't had the time to think properly about it.

I also wish I knew why the cert files have the wrong permissions after creation, but calling `os.Chmod` seems to fix the issue. It might also be a good idea to chmod them in any case, in case someone changed the permissions outside of our control.

NB: The tests need fixing up, I didn't find the time to rectify everything after my latest changes.

marefr (changes requested):
Some suggestions for removing import what seems like an error converting from global state

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,15 +86,161 @@ func escape(input string) string {
...
+		s.logger.Debug("File path not stored in data source")
+	}
+	if fileContent != "" {
+		if generatedFilePath == "" {
+			// Use standard filename
+			generatedFilePath = filepath.Join(dataDir, filename)
+		}
+
+		s.logger.Debug("Writing cert file", "path", generatedFilePath)
+		if err := ioutil.WriteFile(generatedFilePath, []byte(fileContent), 0600); err != nil {
```
papagian, line 145: I think that it would be good to remove any leading spaces (since the certificate will fail otherwise) and add a test for this case similar to this [one](https://github.com/grafana/grafana/blob/966078eb60979c87d18473f84a4a9f8257c41997/pkg/tsdb/postgres/postgres_test.go#L109).

papagian (approved):
I've tested and it works fine! I have a minor suggestion to trim the leading spaces from the content.
Also the [documentation](https://grafana.com/docs/grafana/latest/datasources/postgres/) should be updated appropriately and it would be handful to update the [provisioning](https://grafana.com/docs/grafana/latest/datasources/postgres/#configure-the-data-source-with-provisioning) section with some examples for configuring TLS/SSL certificates (by providing file paths or content)

pkg/tsdb/postgres/postgres.go
```diff
@@ -17,20 +23,38 @@ import (
...
+
+func (s *postgresService) newPostgresQueryEndpoint(datasource *models.DataSource) (tsdb.TsdbQueryEndpoint, error) {
+	s.logger.Debug("Creating Postgres query endpoint")
+
+	cnnstr, err := s.generateConnectionString(datasource)
 	if err != nil {
 		return nil, err
 	}
 
 	if setting.Env == setting.Dev {
```
marefr, line 56: Migrating away from using global settings

```suggestion
	if s.Cfg.Env == setting.Dev {
```

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,15 +86,161 @@ func escape(input string) string {
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	workDir := filepath.Join(s.Cfg.DataPath, "tls", ds.Uid+"generatedTLSCerts")
+	exists, err := fs.Exists(workDir)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		if err := os.MkdirAll(workDir, 0700); err != nil {
```
marefr, line 192: Do user really needs execute permission in the directory?

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,15 +86,161 @@ func escape(input string) string {
...
+}
+
+func (s *postgresService) generateConnectionString(datasource *models.DataSource) (string, error) {
+	tlsConfigurationMethod := strings.TrimSpace(
+		strings.ToLower(datasource.JsonData.Get("tlsConfigurationMethod").MustString("file-path")))
 	tlsMode := strings.TrimSpace(strings.ToLower(datasource.JsonData.Get("sslmode").MustString("verify-full")))
 	isTLSDisabled := tlsMode == "disable"
 
+	if !isTLSDisabled && tlsConfigurationMethod == "file-content" {
+		if err := s.writeCertFiles(datasource); err != nil {
```
marefr, line 234: Is there any reason writing the cert files and populating `datasource.JsonData` with generated cert files rather  than doing it at line 273 below?

pkg/tsdb/postgres/postgres.go
```diff
@@ -61,15 +86,161 @@ func escape(input string) string {
...
+	decrypted := ds.SecureJsonData.Decrypt()
+	tlsRootCert := decrypted["tlsCACert"]
+	tlsClientCert := decrypted["tlsClientCert"]
+	tlsClientKey := decrypted["tlsClientKey"]
+
+	if tlsRootCert == "" && tlsClientCert == "" && tlsClientKey == "" {
+		s.logger.Debug("No TLS/SSL certificates provided")
+	}
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
```
marefr, line 184: I think this might be problematic, basically holding a lock and making every postgres data source instance that uses `file-content` method wait here. Something we need to think a bit more about before being able to merge this.

marefr (changes requested):
Looks great in general. Added some comments/suggestions. Still needs some more thinking in regards to the mutex.

pkg/tsdb/postgres/postgres.go
ying-jeanne: I tested with leading spaces, it seems pq package is removing spaces and it worked directly.

papagian: If I understand correctly `pq` trims the spaces from the connection URL; what I meant was to trim the spaces from the input (certificate or key) before saving it in the file that will be provided to the connection string.
However I have tested again and my problem weer not the leading spaces but leading spaces/tabs on each line of the input which is an invalid input actually. So not bother with this, my bad!

pkg/tsdb/postgres/postgres.go
aknuds1: @marefr the directory won't be possible to cd into without execute bit, right?

pkg/tsdb/postgres/postgres.go
```diff
@@ -80,41 +99,41 @@ func generateConnectionString(datasource *models.DataSource, logger log.Logger)
...
-		if tlsCert != "" && tlsKey != "" {
-			logger.Debug("Setting TLS/SSL client auth", "tlsCert", tlsCert, "tlsKey", tlsKey)
-			connStr += fmt.Sprintf(" sslcert='%s' sslkey='%s'", tlsCert, tlsKey)
-		} else if tlsCert != "" || tlsKey != "" {
-			return "", fmt.Errorf("TLS/SSL client certificate and key must both be specified")
-		}
+	connStr += fmt.Sprintf(" sslmode='%s'", escape(tlsSettings.Mode))
+
+	// Attach root certificate if provided
+	// Attach root certificate if provided
```
marefr, line 122:
```suggestion
```

pkg/tsdb/postgres/tlsmanager.go
```diff
@@ -0,0 +1,228 @@
...
+	return generatedFilePath
+}
+
+// writeCertFile writes a certificate file.
+func writeCertFile(
+	ds *models.DataSource, logger log.Logger, fileContent string, generatedFilePath string) error {
+	fileContent = strings.TrimSpace(fileContent)
+	if fileContent != "" {
+		logger.Debug("Writing cert file", "path", generatedFilePath)
+		if err := ioutil.WriteFile(generatedFilePath, []byte(fileContent), 0600); err != nil {
```
marefr, line 126: Think that we need to handle existing file exists. If files are written, cache set. Then at a later point someone updates the data source and potentially the certificate content. Then the files need to be overwritten.

pkg/tsdb/postgres/tlsmanager_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+	t.Run("Check datasource cache creation", func(t *testing.T) {
+		var id, index int64
+		var wg sync.WaitGroup
+		wg.Add(10)
+		mutex := new(sync.Mutex)
+		index = 1
+		for id = 1; id <= 10; id++ {
+			go func() {
+				mutex.Lock()
```
marefr, line 59: Why is a mutex used here? Doesn't look like it's needed.

pkg/tsdb/postgres/tlsmanager_test.go
```diff
@@ -0,0 +1,291 @@
...
+				go func() {
+					ds := &models.DataSource{
+						Id:             1,
+						Version:        2,
+						Database:       "database",
+						JsonData:       jsonDataValue,
+						SecureJsonData: securityjsonValue,
+						Uid:            "testData",
+					}
+					mutex.Lock()
```
marefr, line 103: Don't think mutex is needed here either. But I may be wrong/overlook something.

pkg/tsdb/postgres/tlsmanager.go
```diff
@@ -0,0 +1,228 @@
...
+	tlsRootCert := decrypted["tlsCACert"]
+	tlsClientCert := decrypted["tlsClientCert"]
+	tlsClientKey := decrypted["tlsClientKey"]
+
+	if tlsRootCert == "" && tlsClientCert == "" && tlsClientKey == "" {
+		m.logger.Debug("No TLS/SSL certificates provided")
+	}
+
+	// Calculate all files path
+	workDir := filepath.Join(m.dataPath, "tls", ds.Uid+"generatedTLSCerts")
```
marefr, line 161: Better to use id here instead of uid since uid can be changed.

```suggestion
	workDir := filepath.Join(m.dataPath, "tls", fmt.Sprintf("%dgeneratedTLSCerts", ds.Id))
```

marefr (dismissed):
Awesome work with this :+1: Looks really close to get this merged :tada: 

I've added some comments/suggestions. Some minor things. Not sure if doing suggested changes will create some issues that would need to be addressed.

pkg/tsdb/postgres/tlsmanager_test.go
ying-jeanne: for the index we talked about, I would remove the index as you suggested, and use back id as datasource id, then i would not need the mutex anymore

marefr: Yes, in my head that should work :+1:

PR #30361, opened 2021-01-18
ColorPicker: Removed unused enableNamedColors property

### LINKED ISSUE #30356: ColorPicker: Remove enableNamedColors property it does no seem to be used
Think this was mainly for legacy uses and does not look like it's wired up anymore so let's remove it .

Remove this prop:
https://github.com/grafana/grafana/blob/bedd662c0727fbfd02578f84852f5afdfeb4ba19/packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx#L19

thisisobate:
This PR removes enableNamedColors property in the ColorPicker component

Fixes #30356

### DIFF
--- a/packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
+++ b/packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
@@ -16,7 +16,6 @@ export interface ColorPickerProps extends Themeable {
    * @deprecated Use onChange instead
    */
   onColorChange?: ColorPickerChangeHandler;
-  enableNamedColors?: boolean;
 }
 
 export interface Props<T> extends ColorPickerProps, PopoverContentProps {
@@ -51,12 +50,9 @@ export class ColorPickerPopover<T extends CustomPickersDescriptor> extends React
   };
 
   handleChange = (color: any) => {
-    const { onColorChange, onChange, enableNamedColors, theme } = this.props;
+    const { onColorChange, onChange, theme } = this.props;
     const changeHandler = onColorChange || onChange;
 
-    if (enableNamedColors) {
-      return changeHandler(color);
-    }
     changeHandler(getColorForTheme(color, theme));
   };
 
--- a/packages/grafana-ui/src/components/ColorPicker/ColorPicker.story.tsx
+++ b/packages/grafana-ui/src/components/ColorPicker/ColorPicker.story.tsx
@@ -1,17 +1,10 @@
 import React from 'react';
-import { boolean } from '@storybook/addon-knobs';
 import { SeriesColorPicker, ColorPicker } from '@grafana/ui';
 import { action } from '@storybook/addon-actions';
 import { withCenteredStory } from '../../utils/storybook/withCenteredStory';
 import { UseState } from '../../utils/storybook/UseState';
 import { renderComponentWithTheme } from '../../utils/storybook/withTheme';
 import mdx from './ColorPicker.mdx';
-
-const getColorPickerKnobs = () => {
-  return {
-    enableNamedColors: boolean('Enable named colors', false),
-  };
-};
 
 export default {
   title: 'Pickers and Editors/ColorPicker',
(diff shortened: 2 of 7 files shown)


thisisobate:
@torkelo

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
```diff
@@ -51,12 +50,9 @@ export class ColorPickerPopover<T extends CustomPickersDescriptor> extends React
   };
 
   handleChange = (color: any) => {
-    const { onColorChange, onChange, enableNamedColors, theme } = this.props;
+    const { onColorChange, onChange, theme } = this.props;
     const changeHandler = onColorChange || onChange;
 
-    if (enableNamedColors) {
-      return changeHandler(color);
```
torkelo, line 58: this should be the default

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
```diff
@@ -51,12 +50,9 @@ export class ColorPickerPopover<T extends CustomPickersDescriptor> extends React
...
-    }
     changeHandler(getColorForTheme(color, theme));
```
torkelo, line 56: this is the path that can go away

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
thisisobate: okay

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
thisisobate: done! 🥳

torkelo (approved):

torkelo:
looks good!

thisisobate:
@dprokop

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
dprokop: @torkelo I can't recall when this has changed but it seems like the color picker popover always returns a hex color now, not the named color when the color from the palette is chosen. Is that something you have changed?

dprokop (approved):
Looks good @thisisobate !

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
torkelo: @dprokop that can't be true, how did you test that? in master?

torkelo: testing in master I always get named colors when I click on the circle colors

torkelo: in the panel json

dprokop: My bad - I have not noticed this hasn't been updated in this PR yet. You are right - line 58 from the above code should be the default rather than `changeHandler(getColorForTheme(color, theme));`

packages/grafana-ui/src/components/ColorPicker/ColorPickerPopover.tsx
dprokop: @thisisobate seems like you have not updated this yet - let us know when it's updated. It should be `changeHandler(color)`

dprokop (changes requested):

torkelo:
@thisisobate I am so sorry, the support escalation request that reported that the enableNamedColors property did not work, that you always got named colors back (even when this is false) appears to not be true. This does work, and tested some old angular plugins that use the color picker they rely on non named colors so we cannot remove this as I thought. So we have to close this.

PR #30369, opened 2021-01-18
MS SQL: Integrated security

### LINKED ISSUE #11876: [Feat] MSSQL Integrated Security
v5.1.1
MSSQL
Windows 2012R2
### Feature request
Being able to setup a data source using Integrated Security including Encrytion.
For example an jdbc url: jdbc:sqlserver://myhost.acme.com:1433;databaseName=myDB;integratedSecurity=true;encrypt=true
Connectivity to the MSSQL database server using the credentials of the Grafana service.

daniellee:
Fixes #11876

## The Xorm workaround

This feature works if the [login/user id is empty](https://github.com/denisenkom/go-mssqldb#common-parameters). It will then fallback to Windows authentication (providing the user is on Windows and Sql Server has Windows Auth enabled and the user has been given access). 

~~I couldn't figure out how to set the user id to empty (same problem has always existed with the password field). Xorm will ignore empty fields when [updating the database table](https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/datasource.go#L257).~~

~~My workaround/hack is to set the user id to `' '` (one space). Is there a better way to do this? Should I instead send a reset flag in the update command or explicitly overwrite the User field with the empty value in datasource.go?~~

## How to test

This feature adds support for Windows authentication and therefore can only really be tested on Windows. 

On Linux, the user will see the error message:

`Login error: mssql: Login failed for user ''.`

Steps for testing:
1. Install Sql Server (Developer Edition and Managment Studio). There are a [couple of settings that will need to set](https://stackoverflow.com/questions/12774827/cant-connect-to-localhost-on-sql-server-express-2012-2016). 
2. Create a database for unit tests ([create a database and user](https://github.com/grafana/grafana/blob/master/pkg/tsdb/mssql/mssql_test.go#L22))
3. Set the mixed mode authentication for your Sql Server instance to be able to run the unit tests and remove the [Skip from this line](https://github.com/grafana/grafana/blob/master/pkg/tsdb/mssql/mssql_test.go#L32) 
4. Run [devenv setup](https://github.com/grafana/grafana/blob/master/devenv/setup.sh) to get the provisioned data source and dashboards for MSSQL. You will need to restart Grafana on Windows afterwards for Grafana to detect them. 
5. To run the unit tests with Windows authentication then change the [connection string](https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/sqlutil/sqlutil.go#L65) to: `server=%s;port=%s;database=grafanatest`
6. `go test ./pkg/tsdb/mssql/...`

### DIFF
--- a/docs/sources/datasources/mssql.md
+++ b/docs/sources/datasources/mssql.md
@@ -12,18 +12,19 @@ Grafana ships with a built-in Microsoft SQL Server (MS SQL) data source plugin t
 
 ### Data source options
 
-Name           | Description
-------------   | -------------
-`Name`         | The data source name. This is how you refer to the data source in panels and queries.
-`Default`      | Default data source means that it will be pre-selected for new panels.
-`Host`         | The IP address/hostname and optional port of your MS SQL instance. If port is omitted, default 1433 will be used.
-`Database`     | Name of your MS SQL database.
-`User`         | Database user's login/username
-`Password`     | Database user's password
-`Encrypt`      | This option determines whether or to which extent a secure SSL TCP/IP connection will be negotiated with the server, default `false`.
-`Max open`     | The maximum number of open connections to the database, default `unlimited`.
-`Max idle`     | The maximum number of connections in the idle connection pool, default `2`.
-`Max lifetime` | The maximum amount of time in seconds a connection may be reused, default `14400`/4 hours.
+| Name             | Description                                                                                                                           |
+| ---------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `Name`           | The data source name. This is how you refer to the data source in panels and queries.                                                 |
+| `Default`        | Default data source means that it will be pre-selected for new panels.                                                                |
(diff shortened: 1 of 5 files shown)


aknuds1:
Having a look.

aknuds1 (approved):
Backend changes look fine to me, though AFAICT only tests changed between revisions b91950f658ec and 1e08a3fab204 in go-mssqldb? Am I missing something?

daniellee:
@aknuds1 you seem to be right. Maybe only the frontend changes are needed. The SSPI support was added years ago and then [NTLMv2 support](https://github.com/denisenkom/go-mssqldb/pull/551) was added just before b91950f658ec.

aknuds1 (approved):

daniellee:
Used the [MustCols](https://godoc.org/github.com/go-xorm/xorm#Engine.MustCols) to always update the user field and that sets the field to empty when Windows auth is chosen. Think that should be fine for other data sources too - I tested Prometheus and a few others.

PR #30370, opened 2021-01-19
Transforms: Add sort by transformer

ryantxu:
This transform will sort data before sending it to the visualizations.
image

TODO:
- [x] style for the reverse toggle


public/app/core/components/TransformersUI/SortByTransformerEditor.tsx
```diff
@@ -0,0 +1,73 @@
...
+                }}
+              />
+            </InlineField>
+            <InlineField
+              label="Reverse"
+              onClick={() => {
+                onSortChange(index, { ...s, desc: !!!s.desc });
+              }}
+            >
+              <Switch
```
ryantxu, line 53: @dprokop is there a way to get this centered reasonably without custom emotion css?

torkelo: There is an InlineSwitch

torkelo (approved):
Looks good!

osg-grafana:
@oddlittlebird Do we need to add a label for `add to What’s New`?

oddlittlebird:
Yes, we do. Good catch @osg-grafana.

PR #30371, opened 2021-01-19
Chore: adds library panel header to JSON

hugohaggmark:
This PR makes sure we include a `libraryPanel` property in the json model for library panels to distinguish them from "normal" panels. This could be easily be done in the Frontend instead, but I figure red that it would make more sense to have it in the backend.

First, I thought that we shouldn't store this in the db as the information is already in the table row as `uid` and `name` but that meant that we needed to add this header to all user facing requests like `post`, `patch`, `get`, `getAll`. 

Instead of doing the processing on all user facing requests I stored the header instead for `post` and `patch`. Let me know if what you think about this.

Enjoy the review 🎉 

Relates #29610

### DIFF
--- a/pkg/services/librarypanels/database.go
+++ b/pkg/services/librarypanels/database.go
@@ -47,11 +47,12 @@ func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd cre
 		CreatedBy: c.SignedInUser.UserId,
 		UpdatedBy: c.SignedInUser.UserId,
 	}
-	err := lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
-		err := setLibraryPanelHeader(&libraryPanel)
-		if err != nil {
-			return err
-		}
+	err := setLibraryPanelHeader(&libraryPanel)
+	if err != nil {
+		return libraryPanel, err
+	}
+
+	err = lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
 		if _, err := session.Insert(&libraryPanel); err != nil {
 			if lps.SQLStore.Dialect.IsUniqueConstraintViolation(err) {
 				return errLibraryPanelAlreadyExists
--- a/pkg/services/librarypanels/database.go
+++ b/pkg/services/librarypanels/database.go
@@ -2,6 +2,7 @@ package librarypanels
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/services/librarypanels/database.go
```diff
@@ -28,12 +48,17 @@ func (lps *LibraryPanelService) createLibraryPanel(c *models.ReqContext, cmd cre
 		UpdatedBy: c.SignedInUser.UserId,
 	}
 	err := lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
+		err := setLibraryPanelHeader(&libraryPanel)
```
aknuds1, line 51: Can you move the `setLibraryPanel` call out of this database callback? I don't see any reason for having it there. This change  will make the code much easier to read and verify.

pkg/services/librarypanels/librarypanels_test.go
```diff
@@ -25,6 +25,19 @@ func TestCreateLibraryPanel(t *testing.T) {
...
+		func(t *testing.T, sc scenarioContext) {
+			command := getCreateCommand(1, "Text - Library Panel")
+			response := sc.service.createHandler(sc.reqContext, command)
+			require.Equal(t, 200, response.Status())
+
+			var result libraryPanelResult
+			err := json.Unmarshal(response.Body(), &result)
+			l := result.Result.Model["libraryPanel"].(map[string]interface{})
+			require.NoError(t, err)
+			require.Equal(t, l["name"], "Text - Library Panel")
```
aknuds1, line 39: Can you instead assert the equivalence of `l` itself, e.g. with `cmp.Diff`? Otherwise, there can be unaccounted for fields in it, it also makes the code easier to verify for a reviewer.

pkg/services/librarypanels/database.go
hugohaggmark: Yeah, sure I think I can do that.

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Yes, I'll fix that!

pkg/services/librarypanels/database.go
hugohaggmark: Fixed!

pkg/services/librarypanels/librarypanels_test.go
hugohaggmark: Fixed!

aknuds1 (approved):
Looks good AFAICT

hugohaggmark:
@torkelo any feedback before I merge this?

torkelo (commented):
Not sure why we need this in the library panel json model, this is part of the library panel entity already. But I am sure there is a good reason. Is not name & uid part of the library panel api (get / list)

torkelo:
>  json model for library panels to distinguish them from "normal" panels 

I agree this is needed when library panels are embedded inside dashboards, but this looks its part of the library panel json itself. Not sure when we need it there

hugohaggmark:
> Not sure why we need this in the library panel json model, this is part of the library panel entity already. But I am sure there is a good reason. Is not name & uid part of the library panel api (get / list)

My thinking is that it simplifies [loading/deleting/saving dashboards](https://github.com/grafana/grafana/pull/30278) as we can look for this prop in the json. Maybe my thinking is wrong and this could all be constructed in the Frontend.

hugohaggmark:
Closing this after discussion with Torkel, we need to handle this in the Fronend instead. This way we don't get data duplication and syncing.

PR #30374, opened 2021-01-19
POC/Events: show eventBus activity in a panel

ryantxu:
This adds a debug panel to help see events as they happen. 
events


packages/grafana-data/src/events/common.ts
```diff
@@ -0,0 +1,45 @@
+import { DataFrame } from '../types';
+import { eventFactory } from './eventFactory';
+import { BusEventWithPayload } from './types';
+
+/** @alpha */
+export interface DataHoverPayload {
```
ryantxu, line 6: @torkelo I think this is close to a reasonable payload for the hover event.  It would allow multi-axis hover and a consistent way to to hover a series row/column.

packages/grafana-data/src/events/types.ts
```diff
@@ -16,6 +16,7 @@ export interface BusEvent {
 export abstract class BusEventBase implements BusEvent {
   readonly type: string;
   readonly payload?: any;
+  readonly source?: any; // panel
```
ryantxu, line 19: can we automatically inject this based on which panel it is broadcast from?

packages/grafana-data/src/events/common.ts
```diff
@@ -0,0 +1,45 @@
...
+  raw: any; // Original mouse event (includes pageX etc)
+
+  x: Record<string, number>; // { time: 5678 },
+  y: Record<string, number>; // { __fixed: 123, lengthft: 456 }  // each axis|scale gets a value
```
torkelo, line 10: always forcing numbers on these, wont that be a problem for BarChart / categorical x

packages/grafana-data/src/events/common.ts
```diff
@@ -0,0 +1,45 @@
...
+export const DataEvents = {
+  // mouse move & drag (down will be set)
+  dataHover: eventFactory<DataHoverEvent>(DataHoverEvent.type),
+
+  // mouse out
+  dataHoverClear: eventFactory<DataHoverEvent>(DataHoverClearEvent.type),
+
+  // mouse up and click
+  dataSelect: eventFactory<DataHoverEvent>(DataSelectEvent.type),
+};
```
torkelo, line 45: Why do we need these old style events?

packages/grafana-data/src/events/types.ts
torkelo: We could create a wrapper/adapter style event bus where all it did was to add this every payload. 

So let's create a EventBusWithSource that takes in an eventbus and a & source (and all it does is to add the source to the payload, otherwise just a wrapper)

torkelo (commented):
Created https://github.com/grafana/grafana/pull/30379 to further isolate our usage of PanelModel.events to only internal usage. So we are free to change this if we want to. 

A bit torn on how to proceed. 

* Get rid of PanelModel.events and only use a dashboard/app level emitter with different sources & payloads
* Make PanelModel.events a wrapper around dashboard emitter but adds source (so same events), then all consumers need to check source. 
* Keep PanelModel.events as is (isolated) and create a specific wrapper in PanelChrome that  wraps dashboard emitter and adds source. 

But the even with source, how can say inside UPlotChart know what events comes from itself and others? source is fine, but it will need a reference to itself as well so it can now if an event comes from it's own legend & or hover.

packages/grafana-data/src/events/common.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  x: Record<string, number>; // { time: 5678 },
+  y: Record<string, number>; // { __fixed: 123, lengthft: 456 }  // each axis|scale gets a value
+
+  data?: DataFrame; // source data
+  rowIndex?: number; // the hover row
+  columnIndex?: number; // the hover column
+
+  // Drag events
+  down?: Omit<DataHoverPayload, 'down'>;
```
torkelo, line 17: not sure what your thinking here, a nested event for where we started?

ryantxu: yes -- this would be the "drag" event.  where the mouse was when the event started

packages/grafana-data/src/events/common.ts
ryantxu: switched to any

oscarkilhed:
Some concepts from this used here #32941

ryantxu:
Closing in favor of #32941

PR #30380, opened 2021-01-19
AlertingNG: Enforce unique alert definition title (non empty)/UID per organisation

papagian:
Following [this](https://github.com/grafana/grafana/pull/30223#pullrequestreview-570541856) comment, it drops and creates alert definition indices as unique ones.
It modifies alert definition creation to fail if the title is missing.
Also it changes the JSON properties to camelCase.

Fixes #

If there are conflicting alert definitions, they should be deleted (using the respective API endpoint) before applying the database migration (otherwise it will fail).


kylebrandt (approved):

kylebrandt:
- The name property is actually "title", we changed it at some point. So should update UI field name and the property name that is sent.
- The save call is not setting expressions in the `queriesAndExpressions`, it is creating the queries, but seems to be setting the data source model from the first query for each (or something like that):

image

PR #30388, opened 2021-01-19
Grafana-UI: Make Card story public

Clarity-89:
- Make Card story public
- Make metadata separator customisable with default to vertical line `|`.


torkelo (approved):
Nice!

PR #30389, opened 2021-01-19
PanelOptions: Refactoring applying panel and field options out of PanelModel and add property clean up for properties not in field config registry 

### LINKED ISSUE #29895: PanelEdit: Clean field config properties when switching panel plugin

torkelo:
PanelModel is getting too big, this is moving the options & field config defaults & option merging (and validation) out of PanelModel to simplify the code & tests and reduce the size of PanelModel. 

* [x] Move apply field and options out 
* [x] Move tests 
* [x] Write some new tests
* [x] Handle cleaning custom properties that are not part of the registry 
* [x] Handle cleaning out overrides and overrides with no properties
* [x] How to restore options when switching back and forth between visualizations 
* [x] Clear all custom options & custom overrides before switch
* [x] Remember and restore all custom defaults and overrides 

Closes #29895


packages/grafana-data/src/panel/PanelPlugin.ts
```diff
@@ -137,6 +139,10 @@ export class PanelPlugin<TOptions = any, TFieldConfigOptions extends object = an
     configDefaults.custom = {} as TFieldConfigOptions;
 
     for (const option of this.fieldConfigRegistry.list()) {
+      if (option.defaultValue === null || option.defaultValue === undefined) {
+        continue;
+      }
+
       set(configDefaults, option.id, option.defaultValue);
     }
```
torkelo, line 147: We where setting props with a default value of undefined, maybe we should set props with a default value of null as there is a distinction there

torkelo: changed this to only ignore props with undefined as default

packages/grafana-data/src/types/panel.ts
```diff
@@ -131,7 +131,8 @@ export type PanelMigrationHandler<TOptions = any> = (panel: PanelModel<TOptions>
 export type PanelTypeChangedHandler<TOptions = any> = (
   panel: PanelModel<TOptions>,
   prevPluginId: string,
-  prevOptions: any
+  prevOptions: Record<string, any>,
+  prevFieldConfig: FieldConfigSource
```
torkelo, line 135: adding the previous panel's field config here to the PanelTypeChangedHandler @ryantxu

public/app/features/dashboard/state/getPanelOptionsWithDefaults.test.ts
```diff
@@ -0,0 +1,390 @@
...
+                ],
+              },
+              "unit": "bytes",
+            },
+            "overrides": Array [],
+          },
+          "options": Object {
+            "name": "hello",
+            "number": 20,
+            "showThresholds": false,
```
dprokop, line 162: Ok, so the options that are currently set are preserved and the rest is set to defaults?

public/app/features/dashboard/state/getPanelOptionsWithDefaults.test.ts
```diff
@@ -0,0 +1,390 @@
...
+              preferThresholdsMode: true,
+            },
+          },
+        },
+      });
+      expect(result.fieldConfig.defaults.color!.mode).toBe(FieldColorModeId.Thresholds);
+    });
+  });
+
+  describe('when applying defaults clean properties that no longer part of the registry', () => {
```
dprokop, line 231:
```suggestion
  describe('when applying defaults clean properties that are no longer part of the registry', () => {
```

dprokop (approved):
This works as expected and just wow, I don't know how you can shoot with such PRs with this speed :D The color mode settings changes were slightly tricky to understand, but other than that beautiful PR.

public/app/features/dashboard/state/getPanelOptionsWithDefaults.test.ts
torkelo: yes

PR #30395, opened 2021-01-19
Dashboard: Remove template variables option from ShareModal

### LINKED ISSUE #29191: ShareModal: Form updates to share link
Todo
* [ ] Remove the template variable option and close #21131
* [ ] Change the time range option to instead be "Lock time range"  with a description "Will transform a relative to now time range to a fixed absolute time range". Only show this if the current time range is relative to now. 
* [ ] Update form styles to new form components. 

Code https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/components/ShareModal/ShareLink.tsx 
 
Current modal: (found in dashboard share & panel share) 
Screenshot from 2020-11-18 10-35-32

oscarkilhed:
Also updates the styling of the ShareModal

The option to not use the template variables does not make sense if the dashboard/panel depends on template variables. The option is not respected by the link generator.

The styling is out of date and should be updated.

Fixes #29191

### DIFF
--- a/public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
+++ b/public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
@@ -6,9 +6,9 @@ import { appEvents } from 'app/core/core';
 import { buildIframeHtml } from './utils';
 
 const themeOptions: Array<SelectableValue<string>> = [
-  { label: 'current', value: 'current' },
-  { label: 'dark', value: 'dark' },
-  { label: 'light', value: 'light' },
+  { label: 'Current', value: 'current' },
+  { label: 'Dark', value: 'dark' },
+  { label: 'Light', value: 'light' },
 ];
 
 interface Props {
--- a/public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
+++ b/public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
@@ -1,5 +1,5 @@
 import React, { PureComponent } from 'react';
-import { Select, Switch, Field, FieldSet } from '@grafana/ui';
+import { RadioButtonGroup, Switch, Field, FieldSet, TextArea } from '@grafana/ui';
 import { SelectableValue } from '@grafana/data';
 import { DashboardModel, PanelModel } from 'app/features/dashboard/state';
 import { buildIframeHtml } from './utils';
@@ -17,7 +17,7 @@ interface Props {
 
 interface State {
   useCurrentTimeRange: boolean;
-  selectedTheme: SelectableValue<string>;
+  selectedTheme: string;
   iframeHtml: string;
 }
 
@@ -26,7 +26,7 @@ export class ShareEmbed extends PureComponent<Props, State> {
     super(props);
     this.state = {
       useCurrentTimeRange: true,
-      selectedTheme: themeOptions[0],
+      selectedTheme: 'current',
       iframeHtml: '',
     };
   }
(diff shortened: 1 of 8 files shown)


torkelo (commented):
Looks great! I did some design tweaks that simplified the design (removed the big icon that made the design look off center). And improved the spacing a bit. 

Think the Embedd tab can be further improved. 

Try using a Field for the text area (and use our TextArea component), and put the paragraph above the text area in a label description (see how i did that in the Link tab)

public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
```diff
@@ -73,32 +62,30 @@ export class ShareEmbed extends PureComponent<Props, State> {
...
-              <InlineField labelWidth={24} label="Template variables">
-                <Switch
-                  id="share-template-variables"
-                  value={includeTemplateVars}
-                  onChange={this.onIncludeTemplateVarsChange}
-                />
-              </InlineField>
-              <InlineField labelWidth={24} label="Theme">
+              </Field>
+              <Field label="Theme">
```
dprokop, line 85: Since the select has only 3 options available I suggest using [RadioButtonGroup](https://developers.grafana.com/ui/latest/index.html?path=/story/forms-radiobuttongroup--simple) component instead

public/app/features/dashboard/components/ShareModal/ShareLink.tsx
```diff
@@ -98,60 +85,49 @@ export class ShareLink extends PureComponent<Props, State> {
...
-              <InlineField labelWidth={24} label="Template variables">
-                <Switch
-                  id="share-template-vars"
-                  value={includeTemplateVars}
-                  onChange={this.onIncludeTemplateVarsChange}
-                />
-              </InlineField>
-              <InlineField labelWidth={24} label="Theme">
+              </Field>
+              <Field label="Theme">
```
dprokop, line 112: Same with RadioButtonGroup :)

dprokop (approved):
Good stuff @oscarkilhed, left one nit to update, other than that it seems great.

public/app/features/dashboard/components/ShareModal/ShareEmbed.tsx
```diff
@@ -73,32 +62,30 @@ export class ShareEmbed extends PureComponent<Props, State> {
...
                 <Select width={20} options={themeOptions} value={selectedTheme} onChange={this.onThemeChange} />
-              </InlineField>
-            </div>
+              </Field>
+            </FieldSet>
             <p className="share-modal-info-text">
```
dprokop, line 89: Refering to @torkelo comment: you can change this to be `<Field ... description="The html code below...."><TextArea .../></Field>`, then we will have a consistent form for this particular view

dprokop:
When switched the theme selection to radio button I've noticed a minor inconsistency with the labels. This is how the theme switcher looks like on the profile page (http://localhost:3000/profile), mind the capitalization of the options names:

image


If we could align this in the share modal with this PR it would be awesome.

oscarkilhed:
> When switched the theme selection to radio button I've noticed a minor inconsistency with the labels. This is how the theme switcher looks like on the profile page (http://localhost:3000/profile), mind the capitalization of the options names:
> 
> image
> 
> If we could align this in the share modal with this PR it would be awesome.

Yeah I'll fix that!

When playing around with those radio buttons a bit I realized that "Current" doesn't make sense. The link doesn't reflect my current selected theme, it just omits the theme from the link. So I guess the theme is up to the user that clicks the shared link.

torkelo (approved):

torkelo:
Looks great!

dprokop:
> The link doesn't reflect my current selected theme, it just omits the theme from the link. So I guess the theme is up to the user that clicks the shared link.

Exactly - it's not about the theme of the user that is creating the share link, but the one that is opening it.

dprokop (approved):
Let's merge! Looks good

PR #30407, opened 2021-01-20
Transform: improve the "outer join" performance/behavior

ryantxu:
This replaces the "outer join" transformer with an implementation from uPlot.

With this version:
* by default, it will join by fist time or number that all frames share (time vs Time vs _time)
* fills values with "undefined" rather than "null" -- uplot will always treat that as connected series (added just for align)

TODO:
- [x] GraphNG should use the same function
- [x] Sort behavior on single field?  Currently this short circuits when there is only one frame -- this means things are not necessarily sorted when only a single field, but are always sorted when a real join happens.

LATER:
* support duplicate timestamps -- currently a Set is used to find the distinct values.  We could check if a value is already in the set, add something small to it, then remove that offset to it at the end (or not) if we want to keep duplicates


leeoniya:
@ryantxu let's wait till uPlot 1.6.2?

the join() signature will change and will return `data` directly, rather than `{data, isGap}`.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -31,26 +31,34 @@ export function pickBestJoinField(data: DataFrame[]): FieldMatcher {
   return fieldMatchers.get(FieldMatcherID.byName).get(common[0]);
 }
 
+export interface JoinOptions {
+  data: DataFrame[];
```
leeoniya, line 35: @ryantxu data -> frames

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export interface JoinOptions {
+  data: DataFrame[];
+  joinBy?: FieldMatcher;
+
+  /**
+   * @internal -- used when we need to keep a reference to the original frame/field index
+   */
+  keepOriginIndexes?: boolean;
```
leeoniya, line 40: keepOriginIndices

packages/grafana-data/src/transformations/transformers/seriesToColumns.test.ts
```diff
@@ -45,58 +44,102 @@ describe('SeriesToColumns Transformer', () => {
...
-            state: {
-              displayName: 'temperature even',
+            Object {
+              "config": Object {},
+              "labels": Object {
+                "name": "even",
+              },
+              "name": "temperature",
+              "state": Object {
+                "displayName": "even temperature",
```
ryantxu, line 72: `temperature even` vs `even temperature` is the real change here  --- BUT that does no actually make its way to the end user :shrug:

ryantxu:
@mckn @torkelo -- the names for tests changed slightly, but the real behavior looks identical.  Can you guys see if things behave as expected?   I still want to make sure that we use this same function direclty within GraphNG

torkelo (approved):
Looks this for a manual test and could not find anything that looked wrong :)

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,230 @@
...
+// https://github.com/leeoniya/uPlot/blob/master/src/utils.js#L325
+// This avoids needing to import uplot into the data package
+//--------------------------------------------------------------------------------
+
+// Copied from uplot
+type AlignedData = [number[], ...Array<Array<number | null>>];
+
+// nullModes
+const NULL_IGNORE = 0; // all nulls are ignored, converted to undefined (e.g. spanGaps: true)
+const NULL_GAP = 1; // nulls are retained, alignment artifacts = undefined values (default)
```
leeoniya, line 143: maybe rename this to `NULL_RETAIN`. "gap" is pretty obscure outside a timeseries / uPlot context. i'll likely rename it on the uPlot side, too.

leeoniya:
yeah the behavior for a single _frame_ (not field) is a tough call - a trade-off between perf and function behavior uniformity.

it's not just sorting that won't happen but also null modes will be ignored (will behave as NULL_RETAIN).

we should just drop the early exit. any other behvior would be a footgun for sure. generally, we should expect that people dont try to join single frames, and if they do, then they can expect to eat the cost.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,230 @@
...
+const NULL_EXPAND = 2; // nulls are expanded to include adjacent alignment artifacts (undefined values)
+
+// mark all filler nulls as explicit when adjacent to existing explicit nulls (minesweeper)
+function nullExpand(yVals: Array<number | null>, nullIdxs: number[], alignedLen: number) {
+  for (let i = 0, xi, lastNullIdx = Number.NEGATIVE_INFINITY; i < nullIdxs.length; i++) {
```
leeoniya, line 148: NEGATIVE_INFINITY was actually a left-over from the previous strategy when i was testing against values. i've changed this to -1 for index testing: https://github.com/leeoniya/uPlot/commit/9324c4263a548a70a1a273e43c07717736e9fd63

ryantxu:
@leeoniya can you take a look again?

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      data.push(yVals);
+    }
+  }
+
+  return data;
+}
+
+// Quick test if the first and last points look to be ascending
+export function isLikelyAscendingVector(data: Vector): boolean {
```
leeoniya, line 265: this util function is not join-specific, probably pull out of this file?

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,293 @@
...
+      while (xi < alignedLen && yVals[xi] == null) {
+        yVals[(lastNullIdx = xi++)] = null;
+      }
+    }
+  }
+}
+
+// nullModes is a tables-matched array indicating how to treat nulls in each series
+function join(tables: AlignedData[], nullModes: number[][]) {
+  if (tables.length === 1) {
```
leeoniya, line 199: let's kill this early exit block to get uniform nullModes behavior for all inputs. the single frame bailout should happen outside of this fn (one level up), i think.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,293 @@
...
+ * This will return a single frame joined by the first matching field.  When a join field is not specified,
+ * the default will use the first time field
+ */
+export function outerJoinDataFrames(options: JoinOptions): DataFrame | undefined {
+  if (!options?.frames?.length) {
+    return undefined;
+  }
+  const joinFieldMatcher = options.joinBy ?? pickBestJoinField(options.frames);
+
+  if (options.frames.length < 2) {
```
leeoniya, line 73: === 1 ?

leeoniya:
so far just did review half asleep, on tiny phone screen. will follow up tmrw.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,293 @@
...
+   */
+  keepOriginIndices?: boolean;
+}
+
+/**
+ * This will return a single frame joined by the first matching field.  When a join field is not specified,
+ * the default will use the first time field
+ */
+export function outerJoinDataFrames(options: JoinOptions): DataFrame | undefined {
+  if (!options?.frames?.length) {
```
leeoniya, line 68: neither options nor frames is optional here, so drop the `?` ?

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
ryantxu: it is still a bit too specific to be generally exported -- if we have a use elsewhere we can move it

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
ryantxu: I droped the first, but will keep the second -- although the types require it to be set, the runtime environment does not.  Since this function gets called often on raw data it seems like a cheap way to avoid exploding

leeoniya: so then it's missing on `frames` here?: https://github.com/grafana/grafana/pull/30407#pullrequestreview-578023461

ryantxu: no -- since we already checked it is valid, it does not need to be checked again.

Interestingly, if `frames` was defined as optional, typescript would know it is  not optional after this point

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * This will return a single frame joined by the first matching field.  When a join field is not specified,
+ * the default will use the first time field
+ */
+export function outerJoinDataFrames(options: JoinOptions): DataFrame | undefined {
+  if (!options.frames?.length) {
+    return undefined;
+  }
+  const joinFieldMatcher = options.joinBy ?? pickBestJoinField(options.frames);
```
leeoniya, line 73: can we avoid making this here and just declare it with `let`. if there is only 1 frame and `enforceSort` is absent (the typical case), then this doesnt get used at all.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,294 @@
...
+          fieldIndex,
+        };
+      }
+    }
+
+    if (!join) {
+      continue; // skip the frame
+    }
+
+    if (originalFields.length < 1) {
```
leeoniya, line 138: === 0?

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,294 @@
...
+
+      if (options.keepOriginIndices) {
+        field.state!.origin = {
+          frameIndex,
+          fieldIndex,
+        };
+      }
+    }
+
+    if (!join) {
```
leeoniya, line 134: not a huge fan of negating stuff that may or may not be objects, always preferred `join != null`, but can leave as-is.

packages/grafana-data/src/transformations/transformers/joinDataFrames.ts
```diff
@@ -0,0 +1,294 @@
...
+    const a: AlignedData = [join.values.toArray()]; //
+    for (const field of fields) {
+      a.push(field.values.toArray());
+      originalFields.push(field);
+    }
+    allData.push(a);
+  }
+
+  const joined = join(allData, nullModes);
+  if (!joined) {
```
leeoniya, line 152: i don't think this guard is necessary. `join()` will always return something.

leeoniya (approved):

PR #30416, opened 2021-01-20
GraphNG: assume uPlot's series stroke is always a function

dprokop:
Fixes toolitps's series icon not being visible.

But... I have questions and this PR will require followup. Since in the tooltip plugin we are using uPlot's series config to retrieve the series color, we used assume stroke is not a function (according to https://github.com/leeoniya/uPlot/blob/71994d3a6f79cbf478ecf3880205880cc4a17037/dist/uPlot.d.ts#L645 and https://github.com/leeoniya/uPlot/blob/71994d3a6f79cbf478ecf3880205880cc4a17037/dist/uPlot.d.ts#L544 typings). Investigating the code though I found this line https://github.com/leeoniya/uPlot/blob/master/src/uPlot.js#L759 that converts no matter what stroke value to a function. Which leads me to a question what is the missmatch here? Is the typing wrong or is it the stroke type returned incorrect @leeoniya?

### DIFF
--- a/packages/grafana-ui/src/components/uPlot/plugins/TooltipPlugin.tsx
+++ b/packages/grafana-ui/src/components/uPlot/plugins/TooltipPlugin.tsx
@@ -48,8 +48,8 @@ export const TooltipPlugin: React.FC<TooltipPluginProps> = ({ mode = 'single', t
             <SeriesTable
               series={[
                 {
-                  // stroke is typed as CanvasRenderingContext2D['strokeStyle'] - we are using strings only for now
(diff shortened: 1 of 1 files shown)


torkelo (approved):
Looks good for now :)

leeoniya:
@dprokop anything that _can_ be a function will be converted to a function at init-time. this adds a level of uniformity for internal ops. i'd have to create a whole new fn-only variant for readback typings vs opts typings for them to be 100% accurate, which would lead to typings bloat :(

PR #30419, opened 2021-01-20
Graph: Fixes so only users with correct permissions can add annotations

### LINKED ISSUE #29972: Annotations: Viewer can open "Add Annotation" modal (but they cannot create annotation, which is correct)
**What happened**:
Users with the role viewer are allowed and able to create annotations but when they click the save-button they receive the message "Access denied to this dashboard".

annotations
 
The expected behaviour is that once the user with viewer role clicks the save-button that the annotation is saved. The alert pop-up should then say: "Annotation added" and have a green background.

* Log in as a user with the role viewer
* open a dashboard with a graph panel
* click the graph panel and select 'add annotation'
* add a description
* click save-button

I found this issue: https://github.com/grafana/grafana/issues/9991
This issue is labeled as a a feature request, however, it does not mention that a viewer can already add an annotation, but cannot save it.

**Environment**:
- Grafana version: 7.3.6
- Data source type & version: data source used -- grafana -- with query randomwalk to create the graph panel
- OS Grafana is installed on: ms windows os version 10
- User OS & Browser: ms windows os version 10 - firefox
- Grafana plugins: default plugins
- Others: none

hugohaggmark:
When fixing https://github.com/grafana/grafana/pull/29990 I forgot to add cases for when users don't have permissions to add annotations. This PR extracts the logic for when a user can add an annotation into the `canAddAnnotations` function and uses it accordingly where needed.

This is probably something that needs to migrate to GraphNG moving forward.

Enjoy the review 🎉 

Fixes #29972

- test adding annotation by clicking in Graph (viewers shouldn't see Add annotations menu item)
- test adding annotation by selecting an area in the graph while holding down `cmd|ctrl` key
- test adding annotation by clicking in Graph while while holding down `cmd|ctrl` key
- test it with a user that has edit permissions
- test it with a user that has edit permissions and make the dashboard not editable => Settings => General => Editable toggle
image
- test it with a user that has viewer permissions

### DIFF
--- a/public/app/plugins/panel/graph/graph.ts
+++ b/public/app/plugins/panel/graph/graph.ts
@@ -180,7 +180,7 @@ class GraphElement {
       return;
     }
 
-    if ((ranges.ctrlKey || ranges.metaKey) && (this.dashboard.meta.canEdit || this.dashboard.meta.canMakeEditable)) {
+    if ((ranges.ctrlKey || ranges.metaKey) && this.canAddAnnotations()) {
       // Add annotation
       setTimeout(() => {
         this.eventManager.updateTime(ranges.xaxis);
@@ -201,7 +201,7 @@ class GraphElement {
   ): (() => MenuItemsGroup[]) => {
     return () => {
       // Fixed context menu items
-      const items: MenuItemsGroup[] = this.dashboard?.editable
+      const items: MenuItemsGroup[] = this.canAddAnnotations()
         ? [
             {
               items: [
@@ -253,7 +253,7 @@ class GraphElement {
       }
 
       // skip if dashboard is not saved yet (exists in db) or user cannot edit
-      if (!this.dashboard.id || (!this.dashboard.meta.canEdit && !this.dashboard.meta.canMakeEditable)) {
+      if (!this.dashboard.id || !this.canAddAnnotations()) {
         return;
       }
 
@@ -924,6 +924,10 @@ class GraphElement {
       }
       return formattedValueToString(formatter(val, axis.tickDecimals, axis.scaledDecimals));
     };
+  }
+
+  canAddAnnotations() {
+    return (this.dashboard.meta.canEdit || this.dashboard.meta.canMakeEditable) && this.dashboard.editable;
   }
 }
 
--- a/public/app/plugins/panel/graph/graph.ts
+++ b/public/app/plugins/panel/graph/graph.ts
@@ -180,7 +180,7 @@ class GraphElement {
       return;
     }
 
-    if ((ranges.ctrlKey || ranges.metaKey) && this.canAddAnnotations()) {
+    if ((ranges.ctrlKey || ranges.metaKey) && this.dashboard.canAddAnnotations()) {
       // Add annotation
       setTimeout(() => {
         this.eventManager.updateTime(ranges.xaxis);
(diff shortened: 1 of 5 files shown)


public/app/plugins/panel/graph/graph.ts
```diff
@@ -925,6 +925,10 @@ class GraphElement {
       return formattedValueToString(formatter(val, axis.tickDecimals, axis.scaledDecimals));
     };
   }
+
+  canAddAnnotations() {
+    return (this.dashboard.meta.canEdit || this.dashboard.meta.canMakeEditable) && this.dashboard.editable;
```
torkelo, line 930: this.dashboard.editable should be here, then you cannot add annotations to dashboards that are not editable (which you should be able to do)

hugohaggmark: Sorry I don't understand, but `this.dashboard.editable` is part of the logic?

torkelo: yes, think (this.dashboard.meta.canEdit || this.dashboard.meta.canMakeEditable)  should cover it

hugohaggmark: It doesn't, unfortunately. It's very confusing with all these props. But when `editable = false` then an editor could still add annotations. See #29990.

hugohaggmark: I'll do another test run to confirm

hugohaggmark: I did another test run without `this.dashboard.editable` in the statement and it fails for non editable dashboards. For non editable dashboards someone with edit permissions would still be able to add annotations so `this.dashboard.editable` still needs to be in the statement.

torkelo: But that seems to be the current and correct behavior (that you can still add annotations to noneditable dashboards):
https://play.grafana.org/d/Z9NCtxBMz/test-editable?orgId=1

hugohaggmark: Probably because #29990 hasn't been released yet.

hugohaggmark: Hmm, now I'm a bit confused. Then this issue is not a correct then https://github.com/grafana/grafana/issues/28738

torkelo: yea, think the bug in https://github.com/grafana/grafana/issues/28738 was not that you cannot add an annotation but that you cannot then delete or edit it

hugohaggmark: Ah I see, I'll try to fix this in this PR as well. Thank you for clarifing this.

hugohaggmark: I've moved the `canAddAnnotations` function to the `DashboardModel` instead and added a check for this in annotation_tooltip as well.

hugohaggmark: This means that 7.4.0-beta1 will have the intermittent behavior from #29990 until this PR gets merged. 
How do we do with back port labels and betas? Should I also add a back port label to this PR @torkelo?

torkelo: yes, that is needed

torkelo (approved):
look good

PR #30420, opened 2021-01-20
Docs: update license expiration behavior for reporting

AgnesToulet:
Update documentation about license expiration and reporting.

For information, we display this in the reporting UI when the license is expired:
> Creating new reports is not available with an expired license. Existing reports continue to be processed but you need to update your license to create a new one.


osg-grafana (approved):
Verbiage itself looks good, and given logic is valid, approving.

PR #30421, opened 2021-01-20
Docs: Add config settings for fonts in reporting

undef1nd:

### DIFF
--- a/docs/sources/enterprise/reporting.md
+++ b/docs/sources/enterprise/reporting.md
@@ -135,12 +135,12 @@ concurrent_render_limit = 4
 image_scale_factor = 2
 # Path to the directory containing font files
 fonts_path =
-# Name of the TrueType font file with Regular style, eg. DejaVuSansCondensed.ttf
-font_regular =
-# Name of the TrueType font file with Bold style, eg. DejaVuSansCondensed-Bold.ttf
-font_bold =
-# Name of the TrueType font file with Italic style, eg. DejaVuSansCondensed-Oblique.ttf
-font_italic =
+# Name of the TrueType font file with Regular style
+font_regular = DejaVuSansCondensed.ttf
+# Name of the TrueType font file with Bold style
+font_bold = DejaVuSansCondensed-Bold.ttf
+# Name of the TrueType font file with Italic style
+font_italic = DejaVuSansCondensed-Oblique.ttf
 ```
 
 ## Report time range
--- a/docs/sources/enterprise/reporting.md
+++ b/docs/sources/enterprise/reporting.md
@@ -120,7 +120,7 @@ When generating reports, each panel renders separately before being collected in
 
 To modify the panels' clarity you can set a scale factor for the rendered images. A higher scale factor is more legible but will increase the file size of the generated PDF.
 
-It's also possible to specify custom fonts supporting different Unicode scripts. DejaVu that is the default font used for PFD rendering.
+You can also specify custom fonts that support different Unicode scripts. The DejaVu font is the default used for PDF rendering.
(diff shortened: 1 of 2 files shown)


docs/sources/enterprise/reporting.md
```diff
@@ -120,6 +120,8 @@ When generating reports, each panel renders separately before being collected in
 
 To modify the panels' clarity you can set a scale factor for the rendered images. A higher scale factor is more legible but will increase the file size of the generated PDF.
 
+It's also possible to specify custom fonts supporting different Unicode scripts. DejaVu that is the default font used for PFD rendering.
```
osg-grafana, line 123:
```suggestion
You can also specify a custom font that support different Unicode scripts. The DejaVu font is the default used for PDF rendering.
```

osg-grafana: Can a user specify "a" font or "any" font that fulfills that condition? I left it as "a" here because I assume that would otherwise require testing all all possible fonts.

docs/sources/enterprise/reporting.md
```diff
@@ -120,6 +120,8 @@ When generating reports, each panel renders separately before being collected in
 
 To modify the panels' clarity you can set a scale factor for the rendered images. A higher scale factor is more legible but will increase the file size of the generated PDF.
```
osg-grafana, line 121:
```suggestion
To make a panel more legible, you can set a scale factor for the rendered images. However, a higher scale factor increases the file size of the generated PDF.
```

osg-grafana (changes requested):

docs/sources/enterprise/reporting.md
undef1nd: I wanted to stress that there's an option for specifying font different from the default one. The default font does not support, for example, Chinese glyphs. So having this option users are able to bypass the limitation of the default font by using some other one.

osg-grafana (approved):

AgnesToulet:
This [documentation file](https://github.com/grafana/grafana/blob/master/docs/sources/enterprise/enterprise-configuration.md) should also be updated with the new configuration options available.

Also, you don't need the `backport v7.4.x` label on documentation PR as documentation doesn't follow the same release process as the Grafana binaries.

osg-grafana (approved):

osg-grafana (approved):

undef1nd:
Thanks @AgnesToulet for pointing it out! Updated the doc.

sakjur (approved):

docs/sources/enterprise/reporting.md
```diff
@@ -131,6 +133,14 @@ concurrent_render_limit = 4
 # Set the scale factor for rendering images. 2 is enough for monitor resolutions
 # 4 would be better for printed material. Setting a higher value affects performance and memory
 image_scale_factor = 2
+# Path to the directory containing font files
+fonts_path =
+# Name of the TrueType font file with Regular style
```
osg-grafana, line 138:
```suggestion
# Name of the TrueType font file with regular style
```

What is the difference in meaning between Regular and regular? Similar comment for content in lines 140 and 142.

osg-grafana (changes requested):

docs/sources/enterprise/reporting.md
undef1nd: There's no difference in meaning. It's just a font's style. You're right, will change to lowercase.

osg-grafana (approved):

osg-grafana (approved):

osg-grafana (approved):

PR #30431, opened 2021-01-20
Alerting: Hides threshold handle for percentual thresholds

### LINKED ISSUE #20092: Percent Diff alert methods seem to be broken.
**What happened**:
I’ve tried the percent_diff() formula, but it treats the value as absolute, not as percentage.

trigger alert by percent_diff, not the value as absolute.

image

**Environment**:
- Grafana version:V6.4.3
- User OS & Browser: Chrome

hugohaggmark:
This PR hides the threshold handle when the user changes to a percentual threshold. Probably not the best solution but we're currently working on the next generation Alerting so I didn't want to spend too much effort fixing this.

alerts-hide-threshold-handler

Enjoy the review 🎉 

Fixes #20092


torkelo (approved):
Looks good!

PR #30440, opened 2021-01-20
Cloud Monitoring: Fix legend naming with display name override

wbrowne:
Since setting the frame field labels, the legend will end up being more extensive. So this change is to override it with the field config, so that we just use the dataframe name (which is the default when there are no frame field labels)

Before:
before

After:
after


sunker (changes requested):
looks good! 

can we add this also for MQL queries? MQL queries are built and parsed into dataframes in  time_series_filter.go, so the same logic for handling DisplayNameFromDS and Labels should go somewhere  [here](https://github.com/grafana/grafana/blob/cloudwatch/30434-field-config-fix/pkg/tsdb/cloudmonitoring/time_series_filter.go#L255)

sunker (approved):
Tested and it works as expected!

papagian (approved):
Good catch! It works as expected.

PR #30443, opened 2021-01-20
Footer: Fixes layout issue in footer 

torkelo:
A recent change (I made) to Icon component made icons that are not found render svg without a height property causing footer layout issues


ivanahuckova (approved):
👍

PR #30444, opened 2021-01-20
Expressions: Fix button icon

torkelo:
Minor fix to icon placement in Expressions button


kylebrandt (approved):
Looks better to me.

Is now:

image

Was:

image

PR #30445, opened 2021-01-20
Security: Request validator

joanlopez:

Introduces the concept of `PluginRequestValidator` as an abstraction level for the component responsible of filtering plugin requests based on some attributes (data source url, request headers, etc).

- We should discuss if would make sense to add the current proxy ~whitelisting~ allowlist functionality into the OSS implementation of the `PluginRequestValidator`.
- Note I'm returning "Access Denied" as the error message, but we need to discuss what's the preferred message on each case.
- I may have forgotten any "entrypoint" where this validation should be added.  
- I'm concerned about how this could affect the user experience.
- I left few comments inline.


pkg/models/validations.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"net/http"
+)
+
+type PluginRequestValidator interface {
+	// Validate performs a request validation based
+	// on the data source URL and some of the request
+	// attributes (headers, cookies, etc).
+	Validate(dsURL string, req *http.Request) error
```
joanlopez, line 11: There are many alternatives suggested:

- `Validate(pCtx backend.PluginContext, req *http.Request) error` - I decided to only use the data source URL for this PoC because we don't have the plugin context everywhere. However, if I'm not wrong, it may be retrieved.
- Since `dsURL` is sometimes empty (`""`), and `req` is sometimes `nil`, we may split it up into two different validation methods (`ValidateURL`, `ValidateRequest` or something), but that would imply calling both when required (which isn't something bad per se).
- @xlson suggested using a custom data structure with some data attributes (e.g. headers, cookies, etc) instead of using the raw `http.Request`.
- Other alternatives are more than welcome 😛

pkg/plugins/backendplugin/manager.go
```diff
@@ -195,6 +196,19 @@ func (m *manager) CollectMetrics(ctx context.Context, pluginID string) (*backend
...
 // CheckHealth checks the health of a registered backend plugin.
 func (m *manager) CheckHealth(ctx context.Context, pluginContext backend.PluginContext) (*backend.CheckHealthResult, error) {
+	var dsURL string
+	if pluginContext.DataSourceInstanceSettings != nil {
+		dsURL = pluginContext.DataSourceInstanceSettings.URL
+	}
+
+	err := m.PluginRequestValidator.Validate(dsURL, nil)
+	if err != nil {
+		return &backend.CheckHealthResult{
```
joanlopez, line 206: Alternatively, we may return a typed (sentinel) error instead and add the corresponding error handling on the `CheckHealth` consumer's side. Additionally, we may move the validation also to the `CheckHealth` consumer's side (the same applies for the `CallResource` method), but that would imply adding the validation twice (once per consumer: `plugins.go` & `datasources.go`).

ryantxu:
I don't understand what the goal is with this -- how are plugins expected to implement this?

joanlopez:
> I don't understand what the goal is with this -- how are plugins expected to implement this?

It's not plugins expected to implement this but to introduce an abstract filtering layer that can have multiple implementations  (like licensing; i.e. oss vs enterprise). So, the idea is that the OSS implementation of this filtering layer can encapsulate the current ~whitelist~ allowlist functionality (or just be a dummy implementation that does nothing) while the enterprise one can implement a more complex filtering (based on data source URL, requests headers, etc).

xlson:
We were originally intending to make it possible to block requests based on host:port combinations as well as all ports on a host. But that is not possible currently as we don't always know the port that is being used, for example, mysql datasources does not need a port set to know to make requests to 3306. We'd need a way to figure out the port to be able to implement a reliable block-logic based on host:port.

xlson:
@joanlopez Will the current interceptions cover alerting?

joanlopez:
> @joanlopez Will the current interceptions cover alerting?

After doing some research about how the `alerting` service works, I realized in fact the current proposal does not control the alerting requests. Here's some information I gathered during the research:

- We only [register](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/conditions/query.go#L23) one condition type.
- That condition type (`QueryCondition`) [uses the `alerting.EvalContext` to perform the requests.](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/conditions/query.go#L112)
- So, we should be able to inject the `PluginRequestValidator` into the [`AlertEngine`](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/engine.go#L25) then [inject it into the `alerting.EvalContext`](https://github.com/grafana/grafana/blob/master/pkg/services/alerting/engine.go#L167) to let the `QueryCondition` use it. Does that make sense for you @xlson?

That being said, I want to raise my concern about the maintainability of this proposal in terms of how can we ensure we won't introduce new code pieces that do not use the `PluginRequestValidator`. One example might be if at some point we introduce a new `alerting.Condition`.

xlson:
That sounds reasonable.

I agree, this feature will be hard to maintain and that is something we will have to take into consideration going forward.

joanlopez:
> That sounds reasonable.
> 
> I agree, this feature will be hard to maintain and that is something we will have to take into consideration going forward.

PR updated, see I added validation on the alerting service at https://github.com/grafana/grafana/pull/30445/commits/0551555ef771851d0686043246eeb59eb9df3d18

marefr (approved):
LGTM since to be released in 7.4 tomorrow. If I had more time to review this before the release would probably have some opinions/feedback. 

In general, in the future would like to see some potential larger refactorings to allow things like this to be added thru middleware, hooks or decorators (since it's a somewhat cross-cutting concern). Right now this needs to be "sprinkled" all over the place and is a bit unfortunate, like deep inside Grafana alerting business logic. 

One specific thing would be data source http client outgoing request middleware ([code](https://github.com/grafana/grafana/blob/1c4bf2fa4839c7157f6bcf757215484165e1d39f/pkg/models/datasource_cache.go#L171-L176)), but it's currently hard because that is implemented in the models package and cannot use any service - we're planning to refactor this though in regards to plugins.

Nit. Naming using `plugin`, like `PluginRequestValidator`, is perhaps a bit unfortunate since it risk making people confused about it (I was and also others reviewing this). Maybe something to discuss later?

Tested this quickly locally and seems fine. Something to watch out for would be if the service dependency graph has changed and by that causes any problems for Enterprise services for example. Earlier had issues with reporting and image renderer services. But would expect you to have tested this already.

Something missing here might be a validation when saving a data source. Something for the future to discuss maybe?

joanlopez:
> LGTM since to be released in 7.4 tomorrow. If I had more time to review this before the release would probably have some opinions/feedback.
> 
> In general, in the future would like to see some potential larger refactorings to allow things like this to be added thru middleware, hooks or decorators (since it's a somewhat cross-cutting concern). Right now this needs to be "sprinkled" all over the place and is a bit unfortunate, like deep inside Grafana alerting business logic.

I cannot agree more here! 💯 
That's precisely what I tried to mean at the end of [this comment](https://github.com/grafana/grafana/pull/30445#issuecomment-767511821). However, I guess that'd imply a huge refactor which was incompatible with delivering this feature on the v7.4 release (what was required externally).

> One specific thing would be data source http client outgoing request middleware ([code](https://github.com/grafana/grafana/blob/1c4bf2fa4839c7157f6bcf757215484165e1d39f/pkg/models/datasource_cache.go#L171-L176)), but it's currently hard because that is implemented in the models package and cannot use any service - we're planning to refactor this though in regards to plugins.

I'm curious about what's the proposal here. If I'm not wrong, it's a bit tricky because it also implies outgoing non-HTTP traffic, so a single HTTP client with a RoundTripper (or similar) for all the outgoing HTTP traffic wouldn't be enough.  

> Nit. Naming using `plugin`, like `PluginRequestValidator`, is perhaps a bit unfortunate since it risk making people confused about it (I was and also others reviewing this). Maybe something to discuss later?

I'm completely open for suggestions, in fact I opened this PR with the aim of receiving feedback and discussing pros & cons together, but it hasn't had activity till now that the deadline is burning. 

Anyway, starting by a simple rename should be easy, right?

> Tested this quickly locally and seems fine. Something to watch out for would be if the service dependency graph has changed and by that causes any problems for Enterprise services for example. Earlier had issues with reporting and image renderer services. But would expect you to have tested this already.

Didn't test it with the actual Enterprise implementation of the validator, but with a dummy one which denies all and it worked as expected. Also, I suppose @xlson did some tests.

> Something missing here might be a validation when saving a data source. Something for the future to discuss maybe?

Absolutely. We discussed it, but we didn't take a decision. Another missing thing here is the validation over the `routes` plugin settings.

Issue #21708, opened 2020-01-24
Lint: Blacklist path imports from @grafana/* packages


dprokop:
We need to blacklist path imports in @grafana/* as paths are different in bundle vs source. In bundled packages there is no `src` directory, which can lead to weird issues, i.e.

![image](https://user-images.githubusercontent.com/2376619/73051633-64bd8000-3e83-11ea-84be-8f276707b772.png)

https://github.com/grafana/grafana/pull/21707

dprokop:
Fixed with #30453

PR #30453, opened 2021-01-20
Chore: Restrict internal imports from other packages

aocenas:
Imports directly from `package/src/...` won't work when package is published and use in a plugin.


dprokop (approved):
Should work :)

torkelo:
Nice!

dprokop:
Closes https://github.com/grafana/grafana/issues/21708

PR #30461, opened 2021-01-20
Timeseries: only migrage point size when configured

### LINKED ISSUE #30454: Beta Time series panel: point size hidden when set to "never"
**What happened**:

I converted my panel to the new time series panel, and got massive points on hover:

In the settings, there is no config for point size:

If I change the points to always appear, now the point size config shows up (and works):

Also, I am not sure if this is a bug or some issue on my end, but I converted my panel, which had "reasonable" point sizes:

<details>

```json
{
  "aliasColors": {},
  "bars": false,
  "dashLength": 10,
  "dashes": false,
  "datasource": "Prometheus",
  "fill": 1,
  "gridPos": {
    "h": 7,
    "w": 6,
    "x": 18,
    "y": 7
  },
  "id": 4,
  "legend": {
    "avg": false,
    "current": false,
    "max": false,
    "min": false,
    "show": false,
    "total": false,
    "values": false
  },
  "lines": true,
  "linewidth": 1,
  "links": [],
  "nullPointMode": "null",
  "percentage": false,
  "pointradius": 5,
  "points": false,
  "renderer": "flot",
  "seriesOverrides": [],
  "spaceLength": 10,
  "stack": false,
  "steppedLine": false,
  "targets": [
    {
      "expr": "go_goroutines{app=\"istiod\"}",
      "format": "time_series",
      "intervalFactor": 2,
      "legendFormat": "Number of Goroutines",
      "refId": "A",
      "step": 2
    }
  ],
  "thresholds": [],
  "timeFrom": null,
  "timeRegions": [],
  "timeShift": null,
  "title": "Goroutines",
  "tooltip": {
    "shared": true,
    "sort": 0,
    "value_type": "individual"
  },
  "type": "graph",
  "xaxis": {
    "buckets": null,
    "mode": "time",
    "name": null,
    "show": true,
    "values": []
  },
  "yaxes": [
    {
      "format": "short",
      "label": "",
      "logBase": 1,
      "max": null,
      "min": null,
      "show": true
    },
    {
      "format": "short",
      "label": null,
      "logBase": 1,
      "max": null,
      "min": null,
      "show": true
    }
  ],
  "yaxis": {
    "align": false,
    "alignLevel": null
  },
  "options": {
    "alertThreshold": true
  },
  "fieldConfig": {
    "defaults": {
      "custom": {}
    },
    "overrides": []
  },
  "pluginVersion": "7.4.0-beta1",
  "fillGradient": 0,
  "hiddenSeries": false
}
```
</details>

I then switch this to the new panel type, and the point sizes were automatically set to a large size.

The size of points on hover is easily configurable without having to go through (small) hoops like enabling point size temporarily.

Just make a time series panel, configure point size, and set to "never"

**Environment**:

ryantxu:
Fixes: #30454

The flot panel ignores point size configuration for hover, I think we should still use it, but this will not migrate point size unless it is actually used


torkelo (approved):

torkelo:
when points are never we should maybe have better default

ryantxu:
> when points are never we should maybe have better default 

I think our behavior is OK -- it shows the configured point size for the hover.  In the example here, the configured point size is BIG

PR #19416, opened 2019-09-25
Dashboard: Adds support for a global minimum dashboard refresh interval

lfroment0:
**What this PR does / why we need it**:
Allowing users to setup dashboards with low refresh rates can apply too much pressure on datasources.
Instead of having to run scripts to scrap all dashboards and make updates manually, this feature would provide a way for administrators to limit this behavior and avoid investigation on why data sources fail.
 
**What it does:**
- filters out the refresh rates available in the time picker that are lower than the set minimum refresh rate in the configuration .ini file
- adds the minimum refresh rate as available in the timepicker
- If the user tries to enter a refresh rate that is lower than the minimum in the url, defaults to the minimum rate
- when trying to update the json via the API, rejects the update if the dashboard's refesh rate is lower than the minimum
 
**Which issue(s) this PR fixes**:

Fixes #3356

**Documentation** 

See Documentation [here](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#min_refresh_interval)

PR #30474, opened 2021-01-21
RefreshPicker: Make valid intervals in url visible in RefreshPicker

### LINKED ISSUE #30468: v7.4-beta1: Auto refresh interval is not visible in refresh picker after re-load
To replicate: 

* Open dashboard
* add `&refresh=13s` **important that the refresh interval is not a existing refresh interval**
* Do a full page reload

The panels still auto refresh every 5s but the refresh selector makes you think there is no refresh active

hugohaggmark:
This PR is one way to solve #30468 by making so that any valid refresh interval in url is shown in RefreshPicker. The other alternative would be to open a PR that ignores any refresh interval that is not valid. Let me know what you think.

This PR also refactors out `intervalsToOptions` to a separate function.

Enjoy the review 🎉 

Fixes #30468


torkelo:
Not sure the problem is in the picker, is the problem not that an invalid refresh interval is used & trigger refreshes. They should either be ignored (in both refresh picker & timeSrv) or we should pick the closest valid option and use that

hugohaggmark:
> Not sure the problem is in the picker, is the problem not that an invalid refresh interval is used & trigger refreshes. They should either be ignored (in both refresh picker & timeSrv) or we should pick the closest valid option and use that

Sure, this behavior was introduced in https://github.com/grafana/grafana/pull/19416. I'm going to go with picking the closest one in the Picker then.

torkelo:
> Sure, this behavior was introduced in #19416. I'm going to go with picking the closest one in the Picker then.

I think the bug is here: https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/services/TimeSrv.ts#L155 

The check there only checks the server min refresh setting, not the options in the dashboard. The refresh picker checks the dashboard options & the server min refresh interval setting, and so shows emtpy (off), but TimeSrv thinks it's valid and uses it

hugohaggmark:
> > Sure, this behavior was introduced in #19416. I'm going to go with picking the closest one in the Picker then.
> 
> I think the bug is here: https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/services/TimeSrv.ts#L155
> 
> The check there only checks the server min refresh setting, not the options in the dashboard. The refresh picker checks the dashboard options & the server min refresh interval setting, and so shows emtpy (off), but TimeSrv thinks it's valid and uses it

Gotcha, I'm looking at `this.dashboard.timepicker.refresh_intervals` but this is always undefined. Maybe the dashboard isn't loaded at this stage?

hugohaggmark:
> > > Sure, this behavior was introduced in #19416. I'm going to go with picking the closest one in the Picker then.
> > 
> > 
> > I think the bug is here: https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/services/TimeSrv.ts#L155
> > The check there only checks the server min refresh setting, not the options in the dashboard. The refresh picker checks the dashboard options & the server min refresh interval setting, and so shows emtpy (off), but TimeSrv thinks it's valid and uses it
> 
> Gotcha, I'm looking at `this.dashboard.timepicker.refresh_intervals` but this is always undefined within this function at load time. Maybe the dashboard isn't loaded at this stage?

hugohaggmark:
> > > > Sure, this behavior was introduced in #19416. I'm going to go with picking the closest one in the Picker then.
> > > 
> > > 
> > > I think the bug is here: https://github.com/grafana/grafana/blob/master/public/app/features/dashboard/services/TimeSrv.ts#L155
> > > The check there only checks the server min refresh setting, not the options in the dashboard. The refresh picker checks the dashboard options & the server min refresh interval setting, and so shows emtpy (off), but TimeSrv thinks it's valid and uses it
> > 
> > 
> > Gotcha, I'm looking at `this.dashboard.timepicker.refresh_intervals` but this is always undefined within this function at load time. Maybe the dashboard isn't loaded at this stage?

Aah I think I understand now, if there is no intervals then defaults apply 🙈

hugohaggmark:
So I pushed my latest changes, this got a lot more complicated than I thought when first looked into this because there are so many variants.

jackw (approved):
LGTM! 🚀 

Lovely work with the tests Hugo! 👏👏👏

hugohaggmark:
@torkelo any final feedback before I merge this. Have I forgotten something?

torkelo (approved):
Looks great!

PR #30480, opened 2021-01-21
Dashboard: Remove Icon and change copy -> Copy to clipboard in the share embedded panel modal

oscarkilhed:

Small update to the copy and styling of the copy button in Share embedded panel, after some UX consultation.


hugohaggmark (approved):
LGTM! 🎉

Issue #24607, opened 2020-05-13
Updated look and feel of Modals


dprokop:
Figma sketch: https://www.figma.com/file/eFgDg4iNTY8bbjPitUyiCI/Modals?node-id=98%3A58

- [x] Agree on the design and UI patterns for the NG modals
- [ ] Dev modals as part of grafana/ui
  - [ ] Add corresponding documentation page
  - [ ] Add UX gudelines
- [ ] Update documentation pages which contain screenshots with Modals

Refs:
- https://github.com/grafana/grafana/pull/30481

dprokop:
Think this was already implemented with theme changes in v8

PR #30481, opened 2021-01-21
Admin: Fixes so whole org drop down is visible when adding users to org

### LINKED ISSUE #22749: Grafana add to org dropdown is not showing all elements in json
**What happened**:
In grafana admin page while adding a user to an organization the dropdown it's not showing all list returned in the JSON.
It has to show all elements returned in the JSON.
grafana admin -> users -> select user -> add organization, scroll bar actually going out of div
Screenshot 2020-03-12 at 4 33 23 PM

**Environment**:
- Grafana version: 6.6.2
- Data source type & version:
- User OS & Browser: Chrome
- Grafana plugins: 
- Others:

hugohaggmark:
This PR refactors `Modal` into a Functional Component and adds a render property so consumers can customize modal content part of `Modal`. I made `render` optional so we don't break existing usage of `Modal`. 

Would love some feedback on this change, maybe we should refactor all existing Modals to use render pattern instead? Or should we introduce a CustomModal or a third option?

We're probably doing a lot of work for this in v8 #24607, so I didn't want to do too much in this PR.

Enjoy the review 🎉 

Fixes #22749

### DIFF
--- a/packages/grafana-ui/src/components/Modal/Modal.test.tsx
+++ b/packages/grafana-ui/src/components/Modal/Modal.test.tsx
@@ -9,10 +9,10 @@ describe('Modal', () => {
 
   it('renders nothing by default or when isOpen is false', () => {
     const wrapper = mount(<Modal title={'Some Title'} />);
-    expect(wrapper.html()).toBe('');
+    expect(wrapper.html()).toBe(null);
 
     wrapper.setProps({ ...wrapper.props(), isOpen: false });
-    expect(wrapper.html()).toBe('');
+    expect(wrapper.html()).toBe(null);
   });
 
   it('renders correct contents', () => {
--- a/packages/grafana-ui/src/components/Modal/Modal.tsx
+++ b/packages/grafana-ui/src/components/Modal/Modal.tsx
@@ -1,13 +1,13 @@
-import React from 'react';
+import React, { PropsWithChildren, useCallback } from 'react';
 import { Portal } from '../Portal/Portal';
 import { cx } from 'emotion';
-import { withTheme } from '../../themes';
-import { IconName, Themeable } from '../../types';
+import { useTheme } from '../../themes';
+import { IconName } from '../../types';
 import { getModalStyles } from './getModalStyles';
 import { ModalHeader } from './ModalHeader';
 import { IconButton } from '../IconButton/IconButton';
 
-export interface Props extends Themeable {
+export interface Props {
   icon?: IconName;
   iconTooltip?: string;
   /** Title for the modal or custom header element */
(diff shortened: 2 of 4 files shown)


packages/grafana-ui/src/components/Modal/Modal.tsx
```diff
@@ -19,48 +19,51 @@ export interface Props extends Themeable {
...
+          <div className={styles.modalHeaderClose}>
+            <IconButton surface="header" name="times" size="lg" onClick={onDismiss} />
           </div>
-          <div className={styles.modalContent}>{this.props.children}</div>
         </div>
-        <div className={styles.modalBackdrop} onClick={this.props.onClickBackdrop || this.onClickBackdrop} />
-      </Portal>
-    );
-  }
+        {render && render({ styles })}
```
Clarity-89, line 53: Is there a specific reason we'd need `render` in addition to `children` prop? Seems like they both are used to render content. If we need to customize content container styling maybe introducing `contentClassName` prop would work instead?

packages/grafana-ui/src/components/Modal/Modal.tsx
```diff
@@ -19,48 +19,51 @@ export interface Props extends Themeable {
...
+        {!render && <div className={styles.modalContent}>{children}</div>}
```
Clarity-89, line 54: I think this might get a bit confusing when users provide both render and children prop. Maybe adding a warning, if both props are present, that says that the render prop is used, would be a good idea here?

Clarity-89 (commented):
Great job! Left a few comments.

Correct me if I'm wrong, but it seems that render prop is introduced here to be able to override the content styling. Maybe it would be simpler to introduce a prop to customise container styles instead, same as `className`, but for content?

packages/grafana-ui/src/components/Modal/Modal.test.tsx
```diff
@@ -9,10 +9,10 @@ describe('Modal', () => {
 
   it('renders nothing by default or when isOpen is false', () => {
     const wrapper = mount(<Modal title={'Some Title'} />);
-    expect(wrapper.html()).toBe('');
+    expect(wrapper.html()).toBe(null);
```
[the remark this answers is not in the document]
hugohaggmark: Good idea, but something for #24607 I think.

packages/grafana-ui/src/components/Modal/Modal.tsx
```diff
@@ -19,48 +19,51 @@ export interface Props extends Themeable {
...
-  onClickBackdrop = () => {
-    this.onDismiss();
-  };
+interface RenderApi {
+  styles: ReturnType<typeof getModalStyles>;
+}
 
-  renderDefaultHeader(title: string) {
-    const { icon, iconTooltip } = this.props;
+export function Modal(props: PropsWithChildren<Props>): JSX.Element | null {
```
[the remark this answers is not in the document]
hugohaggmark: I'm using `function` not `const` here so I can't use FC? Any suggestions?

packages/grafana-ui/src/components/Modal/Modal.tsx
hugohaggmark: The reason for this is that I didn't want to break any existing Modals.

packages/grafana-ui/src/components/Modal/Modal.tsx
hugohaggmark: Great suggestion, but this depends on if I'll go ahead and refactor all existing Modals or not. But I guess that based on your feedback we should either refactor all Modal or introduce yet another className prop. The latter one is something that we should avoid IMHO.

hugohaggmark: @dprokop would love your feedback on this before I take my next steps

packages/grafana-ui/src/components/Modal/Modal.tsx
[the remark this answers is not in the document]
hugohaggmark: I updated the return type to `ReturnType<FC<Props>>` instead.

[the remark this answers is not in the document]
Clarity-89: Any particular reason we'd want to use `function` here? I think in this case const + FC make code a bit cleaner. Not sure there's a way to use FC with `function` keyword.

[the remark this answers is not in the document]
hugohaggmark: We should have a Frontend Weekly discussion about this. Seems like we don't really have a clear guideline regarding this (yet). Until we have I'll keep using `function` 😉

packages/grafana-ui/src/components/Modal/Modal.tsx
torkelo: Agree with Alex that this solution adds some unnecessary complexity to the Moda. the fundamental problem is really that the Select does not support portal for the dropdown.

I would add an optional property on the Modal component to disable the overflow: auto property in the modal content, or add a contentClassName, not sure which is best out of those two :)

hugohaggmark:
I reverted to using a simple prop called `contentClassName`.

public/app/features/admin/UserOrgs.tsx
```diff
@@ -224,9 +225,21 @@ export class AddToOrgModal extends PureComponent<AddToOrgModalProps, AddToOrgMod
     const { isOpen } = this.props;
     const { role } = this.state;
     const styles = getAddToOrgModalStyles();
-
+    const modalStyles = getModalStyles(this.props.theme);
+    const contentClassName = cx(
+      modalStyles.modalContent,
```
torkelo, line 230: This this is a left over from the previous solution, this is already applied in the Modal component

torkelo (approved):
Look good :)  found one thing you left from the first solution

public/app/features/admin/UserOrgs.tsx
```diff
@@ -224,9 +225,21 @@ export class AddToOrgModal extends PureComponent<AddToOrgModalProps, AddToOrgMod
...
+      css`
+        overflow: visible;
+      `
+    );
```
Clarity-89, line 234:
```suggestion
     const contentClassName = css`
      overflow: visible;
    `;
```

The modal content already has content styles applied, so this shouldn't be necessary here.

Clarity-89 (approved):
Looks good! 🎉  Just one minor comment.

public/app/features/admin/UserOrgs.tsx
hugohaggmark: Yes, I realized that when I saw Torkel's comment. Will fix ASAP.

public/app/features/admin/UserOrgs.tsx
hugohaggmark: Awesome catch, thank you!

packages/grafana-ui/src/components/Modal/Modal.tsx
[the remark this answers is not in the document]
dprokop: My two cents about this. For a matter of readability, I would opt for:

- `function` - when a function actually is a utility
- `const` - in case of _function component_. It's easier to type with `:FC<..>` than `ReturnType<FC<...>>` and I guess I've  never seen typing function components this way. Guess event React guide uses const all over the place rather than functions.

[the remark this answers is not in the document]
hugohaggmark: Hahaha, I love being the first one using that instead of `ReactElement | null`. We should really have a Frontend discussion about the `function` vs `const` usage. I know that @torkelo has some opinions about this too.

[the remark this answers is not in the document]
torkelo: yes, I prefer function over a const variable, except for react components, where the typing is easier / more readable with a const variable.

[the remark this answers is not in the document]
hugohaggmark: Ok, I didn't realize you preferred const for ReactComponents. Good to know.

PR #30482, opened 2021-01-21
Logging: add frontend logging helpers to @grafana/runtime package

domasx2:
Adds public `logInfo`, `logWarning`, `logDebug` and `logError` helpers to `@grafana/runtime`. Intended to be used by plugins for any custom reporting, so plugin doesn't have to install own Sentry client. Also exposes stricter interface than Sentry client.

Related to #27993

### DIFF
--- a/packages/grafana-runtime/src/index.ts
+++ b/packages/grafana-runtime/src/index.ts
@@ -9,5 +9,6 @@ export * from './types';
 export * from './measurement';
 export { loadPluginCss, SystemJS, PluginCssOptions } from './utils/plugin';
 export { reportMetaAnalytics } from './utils/analytics';
+export { LogLevel, logMessage, logError } from './utils/logging';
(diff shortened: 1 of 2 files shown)


mckn (commented):
This looks great! What do you think about exposing different log functions instead of an optional level parameter?

That would fit more into the way you have structured the `logError` function.

So I am basically thinking of exposing:

`logInfo`
`logError`
`logWarning`

Etc.

domasx2:
> This looks great! What do you think about exposing different log functions instead of an optional level parameter?
> 
> That would fit more into the way you have structured the `logError` function.
> 
> So I am basically thinking of exposing:
> 
> `logInfo`
> `logError`
> `logWarning`
> 
> Etc.

Makes sense, done! I only exposed info, error, warning and debug as these map directly to backend logging levels

mckn (approved):
Looks really good! I think you need to merge master into this branch then is should be ready to go.

PR #30496, opened 2021-01-21
AlertingNG: modify backend to conform with the query model send by the frontend

papagian:
The backend should not expect a `datasourceUid` model property but should pass the `datasource` value as the `datasoureUid` in the transformation.

```
"model": {
            "alias": "just-testing",
            "datasource": "gdev-testdata",
            "datasourceUid": "000000004",
            "intervalMs": 1000,
            "maxDataPoints": 100,
            "orgId": 0,
            "refId": "A",
            "scenarioId": "csv_metric_values",
            "stringInput": "1,20,90,30,5,0"
        }
```
becomes:
```
"model": {
            "alias": "just-testing",
            "datasource": "000000004",
            "intervalMs": 1000,
            "maxDataPoints": 100,
            "orgId": 0,
            "refId": "A",
            "scenarioId": "csv_metric_values",
            "stringInput": "1,20,90,30,5,0"
        }
```

### DIFF
--- a/pkg/services/ngalert/eval/alert_query.go
+++ b/pkg/services/ngalert/eval/alert_query.go
@@ -19,10 +19,12 @@ func (d Duration) String() string {
 	return time.Duration(d).String()
 }
 
+// MarshalJSON returns JSON encoding of the Duration
 func (d Duration) MarshalJSON() ([]byte, error) {
 	return json.Marshal(time.Duration(d).Seconds())
 }
 
+// UnmarshalJSON parses the JSON-encoded data and stores nn Duration
 func (d *Duration) UnmarshalJSON(b []byte) error {
 	var v interface{}
 	if err := json.Unmarshal(b, &v); err != nil {
@@ -68,7 +70,7 @@ type AlertQuery struct {
 	// RelativeTimeRange is the relative Start and End of the query as sent by the frontend.
 	RelativeTimeRange RelativeTimeRange `json:"relativeTimeRange"`
 
-	DatasourceUID string `json:"-"`
+	Datasource string `json:"-"`
 
 	// JSON is the raw JSON query and includes the above properties as well as custom properties.
 	Model json.RawMessage `json:"model"`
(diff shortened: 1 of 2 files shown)


kylebrandt:
From the panel query route the frontend sends `datasource` property filled with the name, and `datasourceId` with the id. So seems using `datasource` for different things would be confusing and inconsistent, whereas `datasourceUid` says what it is, and is not inconsistent. `datasourceUid` is also the property currently when saving an alert definition. 

So clearer to change frontend property instead I think? @peterholmberg / @torkelo

torkelo:
@kylebrandt not sure what is best here, to change this in the frontend would complicate things,  we would have to change the query model to something like 
```
refId: string;
datasource: {
  name?:string;
  uid?: string;
}
```

So we have a single data source property that is an object that can have both name & uid,  but that would break a lot of existing stuff, so instead the frontend can do lookups based on name & uid transparently. 

I am hesitant to have both datasource & datasourceUid as the code would always have to always check both, unlike now where we just have a single property, and dataSourceSrv.get  can look up data source by name or uid (first try to lookup by uid then name). But agree it's a bit messy. 

datasourceId has never been part of the frontend query model, only added to queries as they are fired through the tsdb query api.

pkg/services/ngalert/eval/alert_query.go
```diff
@@ -68,7 +70,7 @@ type AlertQuery struct {
 	// RelativeTimeRange is the relative Start and End of the query as sent by the frontend.
 	RelativeTimeRange RelativeTimeRange `json:"relativeTimeRange"`
 
-	DatasourceUID string `json:"-"`
+	Datasource string `json:"-"`
```
kylebrandt, line 73: Now I am confused :-/ Can we add a comment saying what this will be?

torkelo:
Ok had a think I think this makes sense, having a single property named `datasource` that can be either name or uid (we will move to this always being uid) but for backward compatibility, we have to support name for a while

PR #30499, opened 2021-01-21
GraphNG: add bar alignment option

### LINKED ISSUE #18220: Graph: Bar placement option to have it placed before data point
**What happened**: Bars are displayed on the wrong side of the data point on a prometheus query with the rate() or increase() functions.
barsOnWrongSide
In the Image above both panels have the same query: 'increase(myPrometheusCounter[$__interval])’. The top panel has no ‘min step’ so it shows with each spike a tick-up increase in the counter.
The bottom has a ‘15m’ min step and is set to ‘bar’ and not ‘line’ graph. Notice how the queried data points from Prometheus are at 10:00, 10:15, 10:30 etc. Counter increases that happened between 11:15-11:30 are at the 11:30 datapoint. And that bar sits from 11:30-11:40.5

**What you expected to happen**: Bar to sit within the time interval that the data came from

**How to reproduce it (as minimally and precisely as possible)**: Use increase() or rate() on a prometheusCounter[$__interval] and use a large 'min step' to make large intervals to make it easy to see.

[From Docs](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors)
> Range vector literals … select a range of samples back from the current instant. Syntactically, a range duration is appended in square brackets ([]) at the end of a vector selector to specify how far back in time values should be fetched for each resulting range vector element.

So my ‘increase(myPrometheusCounter[15m])’ will return the counter delta from each datapoint back 15 minutes.

**Environment**:
- Grafana version: v6.2.4, v5.4.3

- OS Grafana is installed on: Ubuntu and Docker on Ubuntu
- User OS & Browser: Windows and Ubuntu Firefox
- Grafana plugins:
- Others:

dprokop:
Adds new option, `barAlignment`, that allows bars rendering alignment relative to data point.

UI:
image

Closes https://github.com/grafana/grafana/issues/18220

### DIFF
--- a/packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
+++ b/packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
@@ -9,12 +9,14 @@ import {
   PointVisibility,
   LineInterpolation,
   GraphGradientMode,
+  BarConfig,
+  BarAlignment,
 } from '../config';
 import { PlotConfigBuilder } from '../types';
 import { DataFrameFieldIndex } from '@grafana/data';
 import { getScaleGradientFn, getOpacityGradientFn, getHueGradientFn } from './gradientFills';
 
-export interface SeriesProps extends LineConfig, FillConfig, PointsConfig {
+export interface SeriesProps extends LineConfig, BarConfig, FillConfig, PointsConfig {
   scaleKey: string;
   gradientMode?: GraphGradientMode;
   /** Used when gradientMode is set to Scheme */
@@ -40,6 +42,7 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
       lineInterpolation,
       lineWidth,
       lineStyle,
+      barAlignment,
       showPoints,
       pointColor,
       pointSize,
@@ -66,7 +69,7 @@ export class UPlotSeriesBuilder extends PlotConfigBuilder<SeriesProps, Series> {
         lineConfig.dash = lineStyle.dash ?? [10, 10];
       }
       lineConfig.paths = (self: uPlot, seriesIdx: number, idx0: number, idx1: number) => {
-        let pathsBuilder = mapDrawStyleToPathBuilder(drawStyle, lineInterpolation);
+        let pathsBuilder = mapDrawStyleToPathBuilder(drawStyle, lineInterpolation, barAlignment);
         return pathsBuilder(self, seriesIdx, idx0, idx1);
       };
     }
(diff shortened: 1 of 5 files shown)


leeoniya:
i'll review this in a bit, currently on phone.

i don't think this fully addresses more than the title of that issue. if you read through the initial report details, the bar _width_ is also relevant as it represents the bucket time range. that's what make that issue a histogram request rather than a styling tweak.

a true histogram panel (that we'll be working on soon) will have additional controls such as placing ticks only at the bucket boundaries, bar width based on bucket size, left alignment.

i was gonna suggest to add bar width control also, but now i wonder if we should leave further tweaks for the new histogram panel (though we weren't planning on a temporal x in that panel).

@ryantxu @torkelo what do you gentlemen think?

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
```diff
@@ -146,16 +148,15 @@ let builders: PathBuilders | undefined = undefined;
...
+  const barWidthFactor = 0.6;
+  const barMaxWidth = Infinity;
   if (!builders) {
+    // This should be global static, but Jest initalization was failing so we lazy load to avoid the issue
     const pathBuilders = uPlot.paths;
-    const barWidthFactor = 0.6;
-    const barMaxWidth = Infinity;
 
     builders = {
-      bars: pathBuilders.bars!({ size: [barWidthFactor, barMaxWidth] }),
```
leeoniya, line 158: @dprokop the point of having them cached here is so they don't need to be re-initialized on ever redraw() call (like during resize). please do the same thing here as the stepBefore & stepAfter. make something like `barCenter`, `barBefore`, `barAfter`

leeoniya (changes requested):

packages/grafana-ui/src/components/uPlot/config/UPlotSeriesBuilder.ts
dprokop: This one cannot be cached, at least not now, as it depends on the position config which is dynamic as compared to width factor and max width.

leeoniya: i'm confused. how is this any different than what we do with stepBefore/stepAfter and lineInterpolation?

leeoniya: image

leeoniya: i also tried pulling that whole block up and out in https://github.com/grafana/grafana/pull/30518, but that's still failing Jest/CI for mysterious reasons :(

dprokop: @leeoniya oh shit, u totally right with the separate builder for bar types, rather than single one that I have proposed. It makes sense- thans!

leeoniya (approved):
the icons feel too narrow, and it's hard to see where the datapoints are. looks like they're based on portions of the step versions, but should probably be grown horizontally to the same width as the linear/smooth ones.

PR #30502, opened 2021-01-21
Docs: Update 7.4 What's New to use more correct description of alerting notification template feature

wbrowne:
There may be some confusion over this feature and what it supports - namely full templating support for alert notifications. This updated wording is more specific and therefore hopefully clearer to users


docs/sources/whatsnew/whats-new-in-v7-4.md
```diff
@@ -79,7 +79,7 @@ The main use case is for [multi-dimensional](https://grafana.com/docs/grafana/la
 
 > **Note:** Queries built with this feature might break with minor version upgrades until Grafana 8 is released. This feature does not work with the current Grafana alerting.
 
-### Variable support in alert notifications
+### Alert notification query label interpolation
```
oddlittlebird, line 82: The problem with this change is that it removes all mention of templates, variables, or support of either of those. In fact, reading through the description, I would have no idea that this gives me support for adding variables to my alert notifications.

Can you add that information in here somehow? Keep in mind that we are writing for beginners with Grafana, not just engineer admins.

oddlittlebird (approved):
Much better, thank you!

Issue #30354, opened 2021-01-18
Migrate grafana/ui sass to emotion


dprokop:
SASS styles to migrate to Emotion, all in grafana/ui package:

- [x] [FormField](https://github.com/grafana/grafana/blob/1c58202b26cc738e1b51e1add6606b7e602cfda8/packages/grafana-ui/src/components/FormField/FormField.tsx)
- [x] [ColorPicker](https://github.com/grafana/grafana/blob/1c58202b26cc738e1b51e1add6606b7e602cfda8/packages/grafana-ui/src/components/ColorPicker/ColorPicker.tsx)
- [x] [RefreshPicker](https://github.com/grafana/grafana/blob/1c58202b26cc738e1b51e1add6606b7e602cfda8/packages/grafana-ui/src/components/RefreshPicker/RefreshPicker.tsx) 
- [x] [CustomScrollbar](https://github.com/grafana/grafana/blob/1c58202b26cc738e1b51e1add6606b7e602cfda8/packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx)
- [x] [Tooltip](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Chart/Tooltip.tsx) 
- [x] [TableInputCSV](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx)

Migrated scss files should be removed both from the codebase and from the grafana/ui scss index file (https://github.com/grafana/grafana/blob/61b9d811cb7878660acdef82d3154462d5e3e287/packages/grafana-ui/src/components/index.scss)

Some getting sarted guide for the styling with Emotion at Grafana: https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md

PR #30506, opened 2021-01-21
CustomScrollbar: migrated styles from sass to emotion

thisisobate:
This PR migrates styles from sass to emotion

Relates #30354 


packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,8 +1,60 @@
...
+    `,
+    view: css`
+      display: flex;
+      flex-grow: 1;
+      flex-direction: column;
+    `,
+  };
+};
+
+const styles = getStyles();
```
torkelo, line 56: this looks like it's called out side the component, and the page argument is never passed from the component props to this

thisisobate: oh...my bad.
would fix this right away

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -75,11 +128,15 @@ export class CustomScrollbar extends Component<Props> {
       passedProps.style.display = 'none';
     }
 
-    return <div {...passedProps} className={track} />;
+    return (
+      <div {...passedProps} className={track === 'track-vertical' ? styles.trackVertical : styles.trackHorizontal} />
```
hugohaggmark, line 132: Instead of passing the string argument `track` ('track-vertical' | 'track-horizontal') shouldn't we pass the correct className instead? This would eliminate the ternary in className:
`className={track === 'track-vertical' ? styles.trackVertical : styles.trackHorizontal}`

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -75,11 +128,15 @@ export class CustomScrollbar extends Component<Props> {
...
+    );
   };
 
-  renderThumb = (thumb: 'thumb-horizontal' | 'thumb-vertical', passedProps: any) => {
-    return <div {...passedProps} className={thumb} />;
+  renderThumb = (thumb: 'thumb-vertical' | 'thumb-horizontal', passedProps: any) => {
+    return (
+      <div {...passedProps} className={thumb === 'thumb-vertical' ? styles.thumbVertical : styles.thumbHorizontal} />
+    );
```
hugohaggmark, line 139: Instead of passing the string argument `thumb` ('thumb-vertical' | 'thumb-horizontal') shouldn't we pass the correct className instead? This would eliminate the ternary in className:
`className={thumb === 'thumb-vertical' ? styles.thumbVertical : styles.thumbHorizontal}`

hugohaggmark (commented):
This is looking good, I left some comments for you to consider.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
hugohaggmark: Thinking more about this, I don't see why `renderThumb` exists at all. Maybe just move this code to `renderThumbHorizontal` and `renderThumbVertical` instead?

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,8 +1,58 @@
 import React, { Component } from 'react';
 import isNil from 'lodash/isNil';
 import classNames from 'classnames';
+import { css } from 'emotion';
 import Scrollbars from 'react-custom-scrollbars';
 
+const getStyles = (page?: boolean) => {
```
torkelo, line 7: Can you move this to the bottom of the file, we usually move the getStyles to the bottom

thisisobate: Alright 👍

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -70,36 +72,36 @@ export class CustomScrollbar extends Component<Props> {
...
-    return <div {...passedProps} className={track} />;
+    return <div {...passedProps} className={styles.trackVertical | styles.trackHorizontal} />;
   };
 
-  renderThumb = (thumb: 'thumb-horizontal' | 'thumb-vertical', passedProps: any) => {
-    return <div {...passedProps} className={thumb} />;
-  };
+  // renderThumb = (thumb: styles.thumbVertical | styles.thumbHorizontal, passedProps: any) => {
+  //   return <div {...passedProps} className={thumb} />;
+  // };
```
hugohaggmark, line 85: Please remove commented code.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -70,36 +72,36 @@ export class CustomScrollbar extends Component<Props> {
...
   }
 
-  renderTrack = (track: 'track-vertical' | 'track-horizontal', hideTrack: boolean | undefined, passedProps: any) => {
+  renderTrack = (styles: any, hideTrack: boolean | undefined, passedProps: any) => {
     if (passedProps.style && hideTrack) {
       passedProps.style.display = 'none';
     }
 
-    return <div {...passedProps} className={track} />;
+    return <div {...passedProps} className={styles.trackVertical | styles.trackHorizontal} />;
```
hugohaggmark, line 80: Are you sure about this ternary? Aren't you're already passing the correct styles to this function? I think if you add a proper type to styles you'd get a transpile error here.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -140,3 +143,52 @@ export class CustomScrollbar extends Component<Props> {
...
+        transition: opacity 0.3s ease-in-out;
+      }
+    `,
+    view: css`
+      display: flex;
+      flex-grow: 1;
+      flex-direction: column;
+    `,
+  };
+};
```
hugohaggmark, line 194: nit: I think you should add a return type to this function. Then you should use that return type in the functions above instead of using `styles: any`. Types help you avoid unforced errors at transpile time instead of runtime. They're your best TypeScript friend :)
```suggestion
interface XYZ {
 customScrollbar: xyz,
 ...
 ...
}
const getStyles = (page?: boolean): XYZ => {
  return {
    customScrollbar: css`
      display: flex;
      flex-grow: 1;
      &:hover {
        opacity: 0.8;
        transition: opacity 0.3s ease-in-out;
      }
    `,
    trackVertical: css`
      border-radius: 3px;
      width: 8px !important;
      right: ${page ? 0 : 2}px;
      bottom: 2px;
    `,
    trackHorizontal: css`
      border-radius: 3px;
      height: 8px !important;
      right: 2px;
      bottom: 2px;
      left: 2px;
    `,
    thumbVertical: css`
      @include gradient-vertical($scrollbarBackground, $scrollbarBackground2);
      border-radius: 6px;
      opacity: 0;
      &:hover {
        opacity: 0.8;
        transition: opacity 0.3s ease-in-out;
      }
    `,
    thumbHorizontal: css`
      @include gradient-horizontal($scrollbarBackground, $scrollbarBackground2);
      border-radius: 6px;
      opacity: 0;
      &:hover {
        opacity: 0.8;
        transition: opacity 0.3s ease-in-out;
      }
    `,
    view: css`
      display: flex;
      flex-grow: 1;
      flex-direction: column;
    `,
  };
};
```

public/app/core/components/Page/Page.tsx
```diff
@@ -47,7 +47,7 @@ class Page extends Component<Props> {
     const { navModel, children, ...otherProps } = this.props;
     return (
       <div {...otherProps} className="page-scrollbar-wrapper">
-        <CustomScrollbar autoHeightMin={'100%'} className="custom-scrollbar--page">
+        <CustomScrollbar autoHeightMin={'100%'} page={true}>
```
hugohaggmark, line 50: Good catch, but do you think page a suitable name for this property? I think it seems a bit too generic?

public/app/features/dashboard/containers/DashboardPage.tsx
```diff
@@ -316,7 +316,7 @@ export class DashboardPage extends PureComponent<Props, State> {
             scrollTop={updateScrollTop}
             hideHorizontalTrack={true}
             updateAfterMountMs={500}
-            className="custom-scrollbar--page"
+            page={true}
```
hugohaggmark, line 319: Outstanding catch, but do you think page a suitable name for this property? I think it seems a bit too generic?

public/app/features/explore/Wrapper.tsx
```diff
@@ -31,7 +31,7 @@ export class Wrapper extends Component<WrapperProps> {
 
     return (
       <div className="page-scrollbar-wrapper">
-        <CustomScrollbar autoHeightMin={'100%'} autoHeightMax={''} className="custom-scrollbar--page">
+        <CustomScrollbar autoHeightMin={'100%'} autoHeightMax={''} page={true}>
```
hugohaggmark, line 34: Nice catch, but do you think page a suitable name for this property? I think it seems a bit too generic?

hugohaggmark (commented):
It's looking better, but I left some more comments for you to consider.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
thisisobate: wow...would fix it right away.
Thanks for bringing this up!

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -140,3 +147,52 @@ export class CustomScrollbar extends Component<Props> {
...
+    `,
+    trackHorizontal: css`
+      border-radius: 3px;
+      height: 8px !important;
+      right: 2px;
+      bottom: 2px;
+      left: 2px;
+    `,
+    thumbVertical: css`
+      @include gradient-vertical($scrollbarBackground, $scrollbarBackground2);
```
thisisobate, line 175: Do I need to convert this to CSS equivalent? @torkelo @hugohaggmark

torkelo: yes

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -16,6 +17,16 @@ interface Props {
   setScrollTop: (event: any) => void;
   autoHeightMin?: number | string;
   updateAfterMountMs?: number;
+  scrollbarStickLeft?: boolean;
```
torkelo, line 20: What does this property do? and why was it added?

torkelo: ah, I see now.

thisisobate: It replaces the former `page` I used earlier.

thisisobate: Basically, It makes the page scrollbar stick to the left side to aid in hitting it.

thisisobate: https://github.com/grafana/grafana/blob/10aabe8ce34656e1e3d669afeae08dd41f70c048/packages/grafana-ui/src/components/CustomScrollbar/_CustomScrollbar.scss#L51

torkelo: left? you mean right?

"isPageScrollbar"  property name would make more sense

thisisobate: I think so too. 
Would update right away

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -2,9 +2,14 @@ import React, { Component } from 'react';
...
 import { css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+import { Themeable } from '../../types/theme';
+import { stylesFactory } from '../../themes';
+import { withTheme } from '../../themes/index';
 import Scrollbars from 'react-custom-scrollbars';
 
-interface Props {
+interface Props extends Themeable {
+  theme: GrafanaTheme;
```
hugohaggmark, line 12: No need to add this as the `Props` interface extends Themeable. Mabye you should take the time to dive into what extending an Interface means before using it? https://www.typescriptlang.org/docs/handbook/interfaces.html#extending-interfaces

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -14,7 +19,7 @@ interface Props {
   hideHorizontalTrack?: boolean;
   hideVerticalTrack?: boolean;
   scrollTop?: number;
-  setScrollTop: (event: any) => void;
+  setScrollTop?: (event: any) => void;
   autoHeightMin?: number | string;
   updateAfterMountMs?: number;
   isPageScrollbar?: boolean;
```
hugohaggmark, line 25: Much better naming, great work!

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -146,53 +153,56 @@ export class CustomScrollbar extends Component<Props> {
...
+        width: 8px !important;
+        right: ${isPageScrollbar ? 0 : 2}px;
+        bottom: 2px;
+      `,
+      trackHorizontal: css`
+        border-radius: 3px;
+        height: 8px !important;
+        right: 2px;
+        bottom: 2px;
+        left: 2px;
```
hugohaggmark, line 181: nit: Maybe see if you can find any spacing in the theme that matches these `px` values?

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
thisisobate: Alright. would fix right away

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -14,7 +19,7 @@ interface Props {
   hideHorizontalTrack?: boolean;
   hideVerticalTrack?: boolean;
   scrollTop?: number;
-  setScrollTop: (event: any) => void;
+  setScrollTop?: (event: any) => void;
```
hugohaggmark, line 22: What made you change this prop from mandatory to optional?

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
hugohaggmark: Great, but more important is to understand why `theme` isn't needed in this context. This is an opportunity to learn something very important.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
thisisobate: I noticed some pre-existing HOC did not provide the `setScrollTop` props to this component thus leading to some errors during runtime. I figured it's best to make this optional.

torkelo: think this is correct

hugohaggmark: Great, I can see now that it's in the defaults.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -127,16 +140,68 @@ export class CustomScrollbar extends Component<Props> {
...
+const getStyles = stylesFactory(
+  (theme: GrafanaTheme, isPageScrollbar?: boolean): StylesInterface => {
+    return {
+      customScrollbar: css`
+        display: flex;
+        flex-grow: 1;
+        &:hover {
+          opacity: 0.8;
+          transition: opacity 0.3s ease-in-out;
+        }
```
torkelo, line 167: this makes the whole view get opacity & get darker

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
thisisobate: Nice!

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -13,15 +18,25 @@ interface Props {
...
+  isPageScrollbar?: boolean;
+}
+
+interface StylesInterface {
+  customScrollbar: string;
+  trackVertical: string;
+  trackHorizontal: string;
+  thumbVertical: string;
+  thumbHorizontal: string;
+  view: string;
```
hugohaggmark, line 33: nit: Maybe move this closer to the `getStyles` function so you can easily see the component props.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
thisisobate: okay. I guess we should remove this entirely then

hugohaggmark (commented):
I think this looks very good now. Skilled work 🎉   I left one final nit for better readability. 
But when I test this, I still can't see any scrollbar. Have you tested the PR?

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -127,16 +140,64 @@ export class CustomScrollbar extends Component<Props> {
...
+        linear-gradient(#404357, ${theme.palette.dark10});
+        border-radius: ${theme.border.radius.md};
+        opacity: 0;
+        &:hover {
+          opacity: 0.8;
+          transition: opacity 0.3s ease-in-out;
+        }
+      `,
+      thumbHorizontal: css`
+        linear-gradient(#404357, ${theme.palette.dark10});
```
torkelo, line 188: not valid css, have to set a background property

thisisobate: oh...didn't see that coming 😅
Thanks alot Torkel...would fix it right away

thisisobate:
After a brief conversation with @torkelo, we figured it's best to make this a functional component. We also resulted in wrapping the original CSS class names into a wrapper as illustrated in this PR

hugohaggmark:
> After a brief conversation with @torkelo, we figured it's best to make this a functional component. We also resulted in wrapping the original CSS class names into a wrapper as illustrated in this PR

Ok, so now we have two challenges migrating from sass and class component. I don't agree with this added complexity, but I'm sure there are valid reasons to do both in this PR. Looking at changes now.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
-  updateAfterMount() {
-    if (this.ref && this.ref.current) {
-      const scrollbar = this.ref.current as any;
+  const updateAfterMount = () => {
+    if (ref && ref.current) {
+      const scrollbar = ref.current as any;
       if (scrollbar.update) {
         scrollbar.update();
       }
     }
```
hugohaggmark, line 56: nit: Maybe using a quick exit here would make the code cleaner.
```suggestion
    if (!ref || !ref.current) {
      return;
    }

    const scrollbar = ref.current as any;
    if (scrollbar.update) {
      scrollbar.update();
    }
```

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
+  hideVerticalTrack,
+  scrollTop,
+  updateAfterMountMs,
+  children,
+}) => {
+  const ref = useRef<Scrollbars>(null);
+  const styles = useStyles(getStyles);
+
+  const updateScroll = () => {
+    console.log('updateScroll');
```
hugohaggmark, line 44: Please remove log message.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
-    }
-
-    return <div {...passedProps} className={track} />;
   };
 
-  renderThumb = (thumb: 'thumb-horizontal' | 'thumb-vertical', passedProps: any) => {
-    return <div {...passedProps} className={thumb} />;
-  };
+  /** onMount */
+  useEffect(() => {
```
hugohaggmark, line 60: nit: Try to have all hooks follow each other at the top of the component, makes it easier to reason about.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
-        renderThumbHorizontal={this.renderThumbHorizontal}
-        renderThumbVertical={this.renderThumbVertical}
-        renderView={this.renderView}
-      >
-        {children}
-      </Scrollbars>
-    );
+    // this logic is to make scrollbar visible when content is added body after mount
+    if (ref.current === null && updateAfterMountMs) {
+      console.log('scrollbar mount updateAfterMountMs');
```
hugohaggmark, line 65: Please remove console message.

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
+      setTimeout(updateAfterMount, updateAfterMountMs);
+    }
+  });
+
+  const renderTrackHorizontal = useCallback((passedProps: any) => {
+    return renderTrack('track-horizontal', hideHorizontalTrack, passedProps);
+  }, []);
```
hugohaggmark, line 72: This callback is dependent on `hideHorizontalTrack` so it should be part of dependencies.
```suggestion
  }, [hideHorizontalTrack]);
```

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -1,142 +1,170 @@
...
+
+  const renderTrackVertical = useCallback((passedProps: any) => {
+    return renderTrack('track-vertical', hideVerticalTrack, passedProps);
+  }, []);
```
hugohaggmark, line 76: This callback is dependent on `hideVerticalTrack ` so it should be part of dependencies.
```suggestion
  }, [hideVerticalTrack]);
```

hugohaggmark (approved):
Substantial work on this 🎉   I found a couple of smaller things and some nits but I think it works as expected now when I try it out. The color seems slightly greener now on my screen though:
**Before**:
image

**After**:
image

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
torkelo: ah, good catch!

packages/grafana-ui/src/components/CustomScrollbar/CustomScrollbar.tsx
```diff
@@ -56,15 +46,24 @@ export const CustomScrollbar: FC<Props> = ({
...
+  if (updateAfterMountMs) {
+    useEffect(() => {
+      setTimeout(() => {
+        const scrollbar = ref.current as any;
+        if (scrollbar?.update) {
+          scrollbar.update();
+        }
+      }, updateAfterMountMs);
+    }, []);
+  }
```
hugohaggmark: Doesn't this break the rules of hooks? [https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

torkelo: How do I do an onMount (only on mount ) one time effect otherwise?

torkelo: Which rule does it break?

hugohaggmark: `Don’t call Hooks inside loops, conditions, or nested functions.`

hugohaggmark: This looks like it's calling a hook inside a condition.

torkelo: ah, true. Wonder if it still is ok. Just did not want this onMount effect if was not needed

torkelo: I think this should be fine, as updateAfterMountMs should never change during a scrollbars lifetime

PR #30510, opened 2021-01-21
ButtonSelect & RefreshPicker: Rewrite of components to use new emotion based ToolbarButton & Menu

torkelo:
Continuing work on getting rid of navbar-button styles, the trickiest is the old SelectButton

Merges logic spread out and overridden over many components now much simplified in RefreshPicker. 

* [x] Use ToolbarButton for the button
* [x] Use Menu for the menu
* [x] Fix the hiding of the Off value (RefreshPicker Off value is emtpy)
* [x] Fix making this work as as button addon to the blue Run Query button in Explore
* [x] Handle all the explore use cases inside RefreshPicker 
* [x] Update RefreshPicker story to show case all states. 
* [x] Make ToolbarButton support primary and destructive states
* [x] Make Icon component handle spinner state to simplify the logic 
* [x] Remove a newer version of ButtonSelect that uses our SelectBase but was not used anywhere (and could not find any plugin that used it) 

This also fixes some bugged styles in explore with the blue run query button and hover state.

Changes
* Changes so when live streaming in explore the Run query button just shows a spinner instead of "Run query" like it does today where this button does nothing when live streaming. 

Updated refresh picker story 
Screenshot from 2021-01-22 15-37-35


packages/grafana-ui/src/components/Button/ToolbarButton.tsx
```diff
@@ -19,27 +20,31 @@ export interface Props extends HTMLAttributes<HTMLButtonElement> {
...
     const buttonStyles = cx(
       {
         [styles.button]: true,
         [styles.buttonFullWidth]: fullWidth,
         [styles.narrow]: narrow,
+        [styles.primaryVariant]: variant === 'primary',
+        [styles.destructiveVariant]: variant === 'destructive',
       },
       className
     );
```
Clarity-89, line 40: If the styles are renamed to 'primary' and 'destructive', we'd then simplified this code a bit, which is also easier to maintain in case more variants are added. 
```suggestion
    const buttonStyles = cx(
      {
        [styles.button]: true,
        [styles.buttonFullWidth]: fullWidth,
        [styles.narrow]: narrow,
      },
      styles[variant]
      className
    );
```

packages/grafana-ui/src/components/Icon/Icon.tsx
```diff
@@ -84,6 +84,16 @@ export const Icon = React.forwardRef<HTMLDivElement, IconProps>(
 
 Icon.displayName = 'Icon';
 
+function getFontAwesomeIconStyles(iconName: string, className?: string): string {
+  return cx(
+    iconName,
+    {
+      ['fa-spin']: iconName === 'fa fa-spinner',
```
Clarity-89, line 91:
```suggestion
      'fa-spin': iconName === 'fa fa-spinner',
```

packages/grafana-ui/src/components/Menu/Menu.tsx
```diff
@@ -36,7 +39,7 @@ export interface MenuProps extends React.HTMLAttributes<HTMLDivElement> {
   onClose?: () => void;
 }
 
-/** @public */
+/** @alpha */
```
Clarity-89, line 42: Moving from public to alpha seems a bit strange as this component might be already used publicly?

packages/grafana-ui/src/components/Menu/Menu.tsx
```diff
@@ -116,86 +120,82 @@ interface MenuItemProps {
   target?: string;
```
Clarity-89, line 120:
```suggestion
  target?: LinkTarget;
```

packages/grafana-ui/src/components/Menu/Menu.tsx
```diff
@@ -116,86 +120,82 @@ interface MenuItemProps {
...
+      {
+        [styles.item]: true,
+        [styles.activeItem]: active,
+      },
+      className
+    );
+
+    return (
+      <div className={itemStyle}>
+        <a href={url ? url : undefined} target={target} className={styles.link} onClick={onClick}>
```
Clarity-89, line 139:
```suggestion
        <a href={url ? url : undefined} target={target} className={styles.link} onClick={onClick} rel={target === '_blank' ? "noopener noreferrer" : undefined}>
```

packages/grafana-ui/src/components/Forms/Legacy/Select/ButtonSelect.tsx
```diff
@@ -1,101 +1,81 @@
...
+  return {
+    wrapper: css`
+      position: relative;
+      display: inline-flex;
+    `,
+    menuWrapper: css`
+      position: absolute;
+      z-index: ${theme.zIndex.dropdown};
+      top: ${theme.spacing.formButtonHeight + 1}px;
+      right: 0;
```
Clarity-89, line 78: Should this be `left` instead? Otherwise the dropdown is not aligned (at leat in storybook).

packages/grafana-ui/src/components/RefreshPicker/RefreshPicker.tsx
```diff
@@ -1,42 +1,28 @@
...
-  hasLiveOption?: boolean;
-  // You can supply your own refresh button element. In that case onRefresh and tooltip are ignored.
-  refreshButton?: React.ReactNode;
-  buttonSelectClassName?: string;
-  theme: GrafanaTheme;
+  isLoading?: boolean;
+  isLive?: boolean;
+  text?: string;
+  noIntervalPicker?: boolean;
+  'aria-label'?: string;
```
Clarity-89, line 20: We probably should standardise this as lots of the components define the prop as `ariaLabel`.

public/app/features/explore/RunButton.tsx
```diff
@@ -1,27 +1,10 @@
...
-      .select-button-value {
-        color: white !important;
-      }
-    `,
-  };
-});
-
 export type Props = {
-  splitted: boolean;
+  small?: boolean;
```
Clarity-89, line 7:
```suggestion
  isSmall?: boolean;
```
To be consistent with `isLive` prop below.

packages/grafana-ui/src/components/Forms/Legacy/Select/ButtonSelect.story.internal.tsx
```diff
@@ -1,18 +1,19 @@
-import React from 'react';
+import React, { FC } from 'react';
 import { action } from '@storybook/addon-actions';
-import { withKnobs, object, text } from '@storybook/addon-knobs';
+import { withKnobs, object } from '@storybook/addon-knobs';
```
Clarity-89, line 3: Would be nice to switch this to use controls and add controls for button variant.

Clarity-89 (commented):
Looks good! Left a few comments to consider. 

Also it seems to me like the option have less padding on the right than on the left:
Screenshot 2021-01-22 at 17 54 02

ivanahuckova (commented):
Nice work! 💪  
When I tested this with Live tailing, "Execute query" tooltip doesn't really worked with it. Also we are not able to do anything with that button when live tailing, so maybe we could disable it or change the hover mode?

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
torkelo: ah good idea!

packages/grafana-ui/src/components/Menu/Menu.tsx
torkelo: this was recently added in 7.4 and only used by ContextMenu, I do not like it's API. WIll try to get this change to alpha  into 7.4

packages/grafana-ui/src/components/Forms/Legacy/Select/ButtonSelect.tsx
torkelo: no, the only uses we have right now should be right aligned, have updated the story

packages/grafana-ui/src/components/RefreshPicker/RefreshPicker.tsx
torkelo: hm... never seend that, we should use the html attribute name for it other wise it's going to be messy to refer to this property using two names

packages/grafana-ui/src/components/Menu/Menu.tsx
```diff
@@ -116,86 +120,82 @@ interface MenuItemProps {
...
-    groupLabelColor: theme.colors.textWeak,
-    itemBgHover: dark7,
-    headerBg: dark1,
-    headerSeparator: dark7,
-  };
-
-  const styles = theme.isDark ? darkThemeStyles : lightThemeStyles;
+  const linkColor = theme.colors.text;
+  const linkColorHover = theme.colors.linkHover;
+  const wrapperBg = theme.colors.formInputBg;
```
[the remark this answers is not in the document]
torkelo: yea this was a refactoring of the the old menu component and they had these variables (duplicated twice for both theme), and I agree they are unnessary

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
torkelo: except when variant is undefined

torkelo: could not make this suggestion work, without introducing messy logic around undefined

packages/grafana-ui/src/components/Menu/Menu.tsx
torkelo: fixed!

torkelo:
Thanks for the greet feedback, have pushed updates.

packages/grafana-ui/src/components/RefreshPicker/RefreshPicker.tsx
torkelo: Removed the need for this aria-label, moded this to a component aria-label that is defined in RefreshPicker

hugohaggmark:
I just merged this change into my PR and when the RefreshPicker is set to a value it's not that obvious anymore (missing the orange color) is this correct @torkelo?
image

torkelo:
Yea decided to skip the orange color for now

hugohaggmark:
Ok thank you, I thought I'd messed up my merge.

vickyyyyyyy:
When this ready to be released, this PR also needs to be released at the same time: https://github.com/grafana/grafana/pull/31093

PR #30511, opened 2021-01-21
Prometheus: Fix show query instead of Value if no __name__ and metric

### LINKED ISSUE #29466: grafana-7.1.3 graph field values are not updating
all the fields are shown as "values" "values" "values" instead of original fields in the graph.

Legend is shown as value. The following query is applied:

[sum(node_memory_MemTotal)](url) - sum(node_memory_MemFree) - sum(node_memory_Buffers) - sum(node_memory_Cached)

This was working in the previous releases of grafana. But it not working in Grafana - 7.1.3 release.

**This image is from newer grafana release**

[
grafana
](url)

**This image is from older grafana release:**

grafana-old

zoltanbedi:
Fixes #29466

### DIFF
--- a/public/app/plugins/datasource/prometheus/result_transformer.test.ts
+++ b/public/app/plugins/datasource/prometheus/result_transformer.test.ts
@@ -353,6 +353,32 @@ describe('Prometheus Result Transformer', () => {
       expect(result[0].name).toEqual('test{job="testjob"}');
     });
 
+    it('should use query as series name when __name__ is not available and metric is empty', () => {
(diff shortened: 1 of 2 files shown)


ivanahuckova (approved):
LGTM! 🐝

Not sure if you talked to @torkelo, but maybe he could have a look as well. I remember that there was some issue about why we use Value (if I recall correctly - to make it work with some Transformations), but maybe I am remembering this incorrectly.

torkelo (approved):

PR #30514, opened 2021-01-21
Expressions: Measure total transformation requests and elapsed time

### LINKED ISSUE #30402: Expressions: Add a metric(s) for Expression hits.
At least a metric that counts server side expression hits.

**Why is this needed**:
Need to make sure there is at least an operational metric for expression requests so this can be monitored in cloud etc for 7.4.

papagian:

Fixes #30402

Example metrics:
```
# HELP expressions_queries_duration_milliseconds Expressions query summary
# TYPE expressions_queries_duration_milliseconds summary
expressions_queries_duration_milliseconds{respStatus="failure",quantile="0.5"} 0.03297
expressions_queries_duration_milliseconds{respStatus="failure",quantile="0.9"} 0.052602
expressions_queries_duration_milliseconds{respStatus="failure",quantile="0.99"} 0.052602
expressions_queries_duration_milliseconds_sum{respStatus="failure"} 0.105187
expressions_queries_duration_milliseconds_count{respStatus="failure"} 3
expressions_queries_duration_milliseconds{respStatus="success",quantile="0.5"} 0.133245
expressions_queries_duration_milliseconds{respStatus="success",quantile="0.9"} 1.567416
expressions_queries_duration_milliseconds{respStatus="success",quantile="0.99"} 1695.7309
expressions_queries_duration_milliseconds_sum{respStatus="success"} 3258.419529
expressions_queries_duration_milliseconds_count{respStatus="success"} 23
```

Note: failing alert definition evaluations are attempted 3 times.


pkg/expr/transform.go
```diff
@@ -11,11 +11,35 @@ import (
...
+		Name:      "transform_queries_total",
+		Help:      "The total number of tranform queries",
+	}, []string{"status"})
+	prometheus.MustRegister(transformQueryCounter)
+
+	transformQueryHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "grafana",
+		Name:      "transform_queries_duration_seconds",
+		Help:      "Tranform query histogram",
+		Buckets:   prometheus.ExponentialBuckets(0.0001, 2, 24),
```
papagian, line 37: I'm not completely sure about this bucket configuration; another one could better serve our purposes.

kylebrandt:
s/transform/expression? Else it is confusing with the front end feature.

pkg/expr/transform.go
bergquist: Thats alot of buckets :) I think we should reduce the amount. Might make more sense to define these manually and update them if we need more granularity.

pkg/expr/transform.go
```diff
@@ -11,11 +11,35 @@ import (
...
 	"google.golang.org/grpc/status"
 )
 
+var (
+	transformQueryCounter   *prometheus.CounterVec
+	transformQueryHistogram *prometheus.HistogramVec
+)
+
+func init() {
+	transformQueryCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
```
bergquist, line 26: The counter is not really useful here since the histograms exposes a counter as well. 

`grafana_transform_queries_duration_seconds_sum{status="failure"} 0.0026933020000000003
grafana_transform_queries_duration_seconds_count`

The reason to use histograms and counters would be if we want to have more labels for the counter.

pkg/expr/transform.go
papagian: thanks, I can make configurable the `count`.

pkg/expr/transform.go
```diff
@@ -69,7 +94,20 @@ func WrapTransformData(ctx context.Context, query *tsdb.TsdbQuery) (*tsdb.Respon
...
+	defer func() {
+		var respStatus string
+		switch {
+		case err == nil:
+			respStatus = "success"
+		default:
+			respStatus = "failure"
+		}
+		transformQueryCounter.WithLabelValues(respStatus).Inc()
+		transformQueryHistogram.WithLabelValues(respStatus).Observe(time.Since(start).Seconds())
```
papagian, line 108: @bergquist do you think it would be better to store milliseconds instead of seconds?

kylebrandt: I would think ms

pkg/expr/transform.go
papagian: @bergquist 
I have changed it to:
`Buckets:   prometheus.ExponentialBuckets(0.001, 4, 9),`
do you think that is fine or should we better let the users override it?

marefr (changes requested):
Looks good in general. Added some comments/suggestions.

Any description of why you've chosen to use [histogram](https://www.robustperception.io/how-does-a-prometheus-histogram-work) over [summary](https://www.robustperception.io/how-does-a-prometheus-summary-work) (which we use in all other places of the code base I think)?

On another note, given that latency of transforms increasing in a production system how would you track down for which certain dashboard/query/alert rule problem originate? I don't think metrics is the answer to all of this, but having more granular metrics of each transform method moving forward I think would be logical. Nothing to handle in this PR though. In addition, distributed tracing could probably be very useful.

pkg/expr/transform.go
marefr: I'm no export at this. But your suggestion would results in the following buckets ([Go playground](https://play.golang.org/p/dqD5LkgBoXm)): [0.001 0.004 0.016 0.064 0.256 1.024 4.096 16.384 65.536]

This would mean the lowest bucket (0.001) means every transform taking from zero to 0.001ms (1 Microsecond) which seems very low and the highest bucket from 65.536ms and infinity which doesn't seem very useful.

Using for example `ExponentialBuckets(200, 1.5, 9)` would produce the following buckets `[200 300 450 675 1012.5 1518.75 2278.125 3417.1875 5125.78125]` with lowest going from zero to 200ms and highest from ~5s and infinity. 

But at the end it depends on what your normal expected durations are, whether you're planning to use percentiles and/or how you're planning to use the metrics. More information at https://prometheus.io/docs/practices/histograms/#errors-of-quantile-estimation

pkg/expr/transform.go
```diff
@@ -11,11 +11,30 @@ import (
...
+)
+
+func init() {
+	expressionsQuerySummary = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name:       "expressions_queries_duration_milliseconds",
+			Help:       "Expressions query summary",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"respStatus"},
```
marefr, line 31: Nit. Think this can be just status

```suggestion
		[]string{"status"},
```

pkg/expr/transform.go
```diff
@@ -69,7 +88,24 @@ func (s *Service) WrapTransformData(ctx context.Context, query *tsdb.TsdbQuery)
 
 // TransformData takes Queries which are either expressions nodes
 // or are datasource requests.
-func (s *Service) TransformData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
+func (s *Service) TransformData(ctx context.Context, req *backend.QueryDataRequest) (r *backend.QueryDataResponse, err error) {
+	start := time.Now()
+	defer func() {
+		if s.isDisabled() {
```
marefr, line 94: Nit. Now we're checking `isDisabled` in two places.

Doing something like this instead would only require one check:

```go
func (s *Service) TransformData(ctx context.Context, req *backend.QueryDataRequest) (r *backend.QueryDataResponse, err error) {
	if s.isDisabled() {
		return nil, status.Error(codes.PermissionDenied, "Expressions are disabled")
	}

	start := time.Now()
	defer func() { ... }
```

marefr (approved):
LGTM but added a couple nits/suggestions

PR #30519, opened 2021-01-21
Decimals: Big Improvements to auto decimals and fixes to auto decimals bug found in 7.4-beta1 

### LINKED ISSUE #30479: 7.4.0-beta1 Decimals displayed incorrectly
7.2.x, 7.3.x:

2021-01-21 10_15_03-PostgreSQL Database - Grafana

7.4.0-beta1:

2021-01-21 10_12_46-PostgreSQL Database - Grafana

2021-01-21 07_59_51-Clipboard

**Environment**:
- Data source type & version: Prometheus 2.23.0

torkelo:
Fixes #30479 
Cures #Cancer

This is the perfect RP, people will talk about this PR for ages. Everyone in fact already talks about it and what they all say is that it is perfect. I am sure that is what they say. And that it is very smart. The person who invented scaled decimals was not very smart. But thanks to this glorious PR you can now live in a world where decimals are calculated NOT BEFORE a scaled unit is divided by a factor but AFTER, image that!  

I tried this about a year ago but for some reason gave up. This changes decimals logic so instead of trying to calculate the auto decimals before we scale a unit (ie 1250 000 => 1.25 K) we do that at the last moment (ie based on the value 1.25 and not 1250000). Turns out it is much easier to calculate the decimals for `1.25` than it is to calculate the possible decimal needed for `1250000` in case it was going to be divided multiple times later. And this also changes the old graph panel to not use its internal decimals logic but instead use the code platform decimal logic. 

The core platform auto decimal logic is derived from the flot code base. I can't get rid of scaledDecimals in the ValueFormatter function type without breaking things, but maybe not so many plugins that rely on sending timezone in the 4th parameter? 

Looks to be working. But could need some help testing this. 

context: Me going a bit crazy over how many hours I have spent trying to make the system of scaled decimals (that I invented) work over the years, to finally figure out the solution was so simple and did not involve scaled decimals at all.


ryantxu:
down with scaledDecimals!

hugohaggmark (approved):
This PR is perfect, hats off!

packages/grafana-data/src/valueFormats/valueFormats.ts
```diff
@@ -45,10 +45,15 @@ export function toFixed(value: number, decimals?: DecimalCount): string {
...
     return '';
   }
+
   if (value === Number.NEGATIVE_INFINITY || value === Number.POSITIVE_INFINITY) {
     return value.toLocaleString();
   }
 
+  if (decimals === null || decimals === undefined) {
+    decimals = getDecimalsForValue(value);
+  }
```
torkelo, line 55: This is the important part, after changing so auto decimals are not assigned in the displayProcessor before the format function is called and not in toFixedScaled it comes in as null or undefined here, but the value is scaled and it is now easier to figure out the auto decimals!

PR #30521, opened 2021-01-21
GraphNG: uPlot 1.6.2

leeoniya:
this simplifies `uPlot.join()`, dropping `isGap()` in favor of using `null` _and_ `undefined` values in `alignedData`. join() performance is increased by 2x-4x. it is a prerequisite for #30407 plus future improvements to `DataFrame[]` alignment.


ryantxu (approved):
Looks good -- works for me

PR #30529, opened 2021-01-22
DataFrame: cache frame/field index in field state

ryantxu:
In the GraphNG, we need to keep track of the original frame/field index -- we currently have an awkward AlignedFrameWithGapTest class that wraps the frame and holds the index.


packages/grafana-ui/src/components/GraphNG/utils.test.ts
```diff
@@ -22,35 +21,62 @@ describe('alignDataFrames', () => {
...
-          name: 'time',
-          type: FieldType.time,
-          values: new ArrayVector([1000, 2000, 3000, 4000]),
-        },
-        {
-          config: {},
-          state: {
-            displayName: 'temperature A',
-            seriesIndex: 0,
+      expect(aligned?.fields).toMatchInlineSnapshot(`
```
ryantxu, line 24: I changed the style to inline snapshot -- is there any advantage to the non snapshot style?

leeoniya:
`index: { frameIndex, fieldIndex }` feels rather redundant and obscure

better:
`origin: { frameIndex, fieldIndex }`

less better:
`source: { frameIndex, fieldIndex }`

leeoniya:
also wouldn't mind standardizing on join/joined instead of align/aligned.

"joined" conveys a sense of being composed from multiple parts, vs "aligned" which to me just means somehow pre-processed.

ryantxu:
Lets look at aligned vs joined in another issue

packages/grafana-data/src/types/dataFrame.ts
```diff
@@ -162,6 +162,11 @@ export interface FieldState {
    * Useful for assigning color to series by looking up a color in a palette using this index
    */
   seriesIndex?: number;
+
+  /**
+   * Location of this field within the context frames results
+   */
+  origin?: DataFrameFieldIndex;
```
torkelo, line 169: Name feels ambiguous, maybe sourceFieldIndex ? And what is this index type? (On mobile now)

torkelo (approved):
Looks good, but only did a quick review on mobile :)

packages/grafana-data/src/types/dataFrame.ts
ryantxu: DataFrameFieldIndex was added by @mckn for the legend support -- I am not 100% convinced it is necessary, but not sure how to fix it yet

ryantxu: I added `@internal` flags to it -- lets see if we can make it go away

leeoniya: since uPlot operates in seriesIdx, i'm not too sure how we can make this go away without joining the frames much higher and having everything downstream work on the joined frame. probably not realistic or necessary since the joining will only be needed in limited cases.

leeoniya (approved):
LGTM :rocket:

PR #30533, opened 2021-01-22
Docs: Define TLS/SSL terminology

aknuds1:
Add a Markdown file to define and standardize terminology used in the Grafana codebase. Starting out, I've added a definition for the TLS/SSL term that we've decided to use previously. This is quite useful since we refer to TLS/SSL a lot, and it's completely inconsistent at the moment.


contribute/style-guides/terminology.md
```diff
@@ -0,0 +1,15 @@
+# Terminology used throughout Grafana code base
```
osg-grafana, line 1: Can we call this "Grafana technical glossary" and move it up a directory?

contribute/style-guides/terminology.md
```diff
@@ -0,0 +1,15 @@
...
+
+This document aims to define and standardize terminology used throughout Grafana source code.
```
osg-grafana, line 3:
```suggestion
This document defines technical terms used in Grafana.
```

contribute/style-guides/terminology.md
```diff
@@ -0,0 +1,15 @@
...
+
+## TLS/SSL
+
+The acronyms [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) (Transport Layer Security and
```
osg-grafana, line 7: I recommend that we use a definition list of some sort rather than paragraph style. Can you test this format out:

```
kramdown
: A Markdown-superset converter
```

contribute/style-guides/terminology.md
```diff
@@ -0,0 +1,15 @@
...
+
+The acronyms [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) (Transport Layer Security and 
+[SSL](https://en.wikipedia.org/wiki/SSL) (Secure Socket Layer) are both used to describe the HTTPS security layer,
+and are in practice synonymous. However, TLS is considered the current name for the technology, and SSL is considered
+[deprecated](https://tools.ietf.org/html/rfc7568). 
+
+As such, while both terms are in use (also in our codebase) and are indeed interchangeable, TLS is the preferred term. 
+That said however, we have at Grafana Labs decided to use both acronyms in combination when referring to this type of 
+technology, i.e. _TLS/SSL_. This is in order to not confuse those who may not be aware of them being synonymous,
+and SSL still being so prevalent in common discourse.
```
osg-grafana, line 15:
```suggestion
[TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security)

: The preferred term to use when describing an HTTPS security layer.

[SSL](https://en.wikipedia.org/wiki/SSL) (Secure Socket Layer)

 : This term is [deprecated](https://tools.ietf.org/html/rfc7568). Use TLS.
```

osg-grafana (changes requested):

contribute/style-guides/terminology.md
aknuds1: @osg-grafana OK, if that makes sense. I'm still unsure personally.

osg-grafana:
While there are many more terms that need to be added and approved, let’s keep the scope of this PR to just these two / four terms because we are laying the groundwork for how we want the technical glossary to be.

Very happy to be seeing this PR. Sorely needed.

contribute/style-guides/terminology.md
osg-grafana: Ideally there are links from "Use TLS." to the TLS entry, and after we get the definition list working, I would like to figure out how to do that.

osg-grafana: We might also choose to write this entire glossary source in HTML, which might make things much easier.

contribute/style-guides/terminology.md
osg-grafana: It is not a style guide.

aknuds1: Yeah, but I'm unsure where it fits in still.

osg-grafana: I just tried to look for it at https://grafana.com/docs/grafana/latest/style-guide/.

contribute/style-guides/terminology.md
aknuds1: @osg-grafana I'm not familiar with using kramdown, where do you want this to be inserted?

contribute/style-guides/terminology.md
aknuds1: @osg-grafana I don't think these changes are good, because they drop the actually vital information, which is to use the term TLS/SSL throughout Grafana.

contribute/style-guides/terminology.md
osg-grafana: From a content-type perspective, I would expect it at https://github.com/grafana/grafana/tree/master/contribute/technical-glossary.md and in "production" at https://grafana.com/docs/grafana/latest/technical-glossary/ and aliased to https://grafana.com/docs/grafana/latest/glossary/.

contribute/style-guides/terminology.md
osg-grafana: And here in lies the power of a glossary. We get to decide once and then it is settled for everyone. Thank you for the pushback. I would like to hear from additional members of the @grafana/docs-squad to weigh in here as to whether or not we back in a deprecated term into usage.

Two decisions (at least are happening here):
1. How do we want to format the glossary. (I vote for a definition list at minimum.)
2. Do we document status quo (at least as a start), and in this case bake in the use of the deprecated term (SSL)? If so, do we treat it as an exception? How do we move away from a deprecated term if we bake it in? What tripwire can we set such that a deprecated term is discouraged and remove from technical content at a future date?

contribute/style-guides/terminology.md
osg-grafana: This is going to need more input from @grafana/docs-squad. As simple as this PR sounds, we are laying some groundwork here in terms of a model that others can contribute to, so please bear with us as we come to consensus. We have a meeting today where we can discuss this.

osg-grafana: I included this particular markdown style in the code changes in the other parts of the doc. Based on what we do here, we could stand to update the Markdown guide (if we go with Markdown for this doc) to include definition list style at https://github.com/grafana/grafana/blob/master/contribute/style-guides/documentation-markdown-guide.md#lists.

contribute/style-guides/terminology.md
aknuds1: @osg-grafana It was already decided months ago, in conjunction with an external PR, that we would use the term TLS/SSL. @sarlinska made the decision as I recall, as the UX designer.

osg-grafana (changes requested):

contribute/style-guides/terminology.md
aknuds1: We also need to think in terms of documentation and Grafana code, those are two different things. TLS/SSL was chosen as the term to use throughout the Grafana application. Whether the same decision will apply for documentation might be the case, but it's an independent decision.

osg-grafana: Great, then we can reference that PR from here. Where is that link to the PR?

aknuds1:
@osg-grafana To make sure we're on the same page, which term to use for TLS/SSL in the Grafana application isn't purely a docs squad decision. The decision was already made by a UX designer (@sarlinska). If the decision should be re-tried, I don't know who would have to be involved, but we should at the very least check with @sarlinska first..

osg-grafana:
> @osg-grafana To make sure we're on the same page, which term to use for TLS/SSL in the Grafana application isn't purely a docs squad decision. The decision was already made by a UX designer (@sarlinska). If the decision should be re-tried, I don't know who would have to be involved, but we should at the very least check with @sarlinska first..

100% agreement from my side.

contribute/style-guides/terminology.md
osg-grafana: > We also need to think in terms of documentation and Grafana code, those are two different things. TLS/SSL was chosen as the term to use throughout the Grafana application. Whether the same decision will apply for documentation might be the case, but it's an independent decision.

The documentation is the tool to convey that information and aims to align, so I am not sure why there is an independent decision. I believe we are moving in pretty much the same direction, and all the added technical and decision-making context helps. Thank you.

aknuds1: > Great, then we can reference that PR from here. Where is that link to the PR?
> 
> Also, because several months have passed, would it make sense to revisit this decision, given we have a new element to the argument (time)? (I am asking neutrally, and because I do not know the answer to it.)

@osg-grafana I don't remember any longer. It was too many PRs ago.

aknuds1: > > We also need to think in terms of documentation and Grafana code, those are two different things. TLS/SSL was chosen as the term to use throughout the Grafana application. Whether the same decision will apply for documentation might be the case, but it's an independent decision.
> 
> The documentation is the tool to convey that information and aims to align, so I am not sure why there is an independent decision. I believe we are moving in pretty much the same direction, and all the added technical and decision-making context helps. Thank you.

@osg-grafana I think Grafana UI and documentation needs to be thought of as independent concerns. They may align, but they're not the same thing.

aknuds1: In any case the application itself carries more weight, the documentation exists to support the former.

contribute/style-guides/terminology.md
oddlittlebird: Is kramdown supported by the Blackfriday engine? That's what we use according to [Contributing to docs](https://github.com/grafana/grafana/blob/master/contribute/documentation.md#style-and-formatting).

contribute/style-guides/terminology.md
oddlittlebird: Is there a reason not to add this to the Word usage section in the style guide? There's nothing that says we cannot add definitions as well as usage information.

As for format, I think GitHub handles Markdown better than HTML (I could be wrong, feel free to correct me).

aknuds1: @oddlittlebird Markdown is a superset of HTML, so any HTML should be valid.

aknuds1: @oddlittlebird @osg-grafana maybe there's a misunderstanding? I'm addressing terminology in the code/the Grafana application, not in docs.

contribute/style-guides/terminology.md
aknuds1: I'm still not sure exactly what kind of changes you have mind. I don't think it's worth blocking this PR until a style is decided on though. It's probably best to agree on the basics, and experiment with style independently.

oddlittlebird: Agreed.

contribute/style-guides/terminology.md
oddlittlebird: The UI text follows the Grafana style guide in almost all cases. I cannot think of an exception except for using ampersands when they need to save space.

oddlittlebird (approved):

contribute/style-guides/terminology.md
aknuds1: This glossary has to be specific to the codebase in any case, because it may need to distinguish between the term used in the UI and that used in the code. In this case f.ex. the UI term is TLS/SSL, whereas in code it's TLS.

osg-grafana:
I get the feeling that I am blocking development based on all the exchanges. Will build a Grafana technical glossary starting with a more straightforward term. Unblocking this PR so we can move forward.

osg-grafana (approved):

osg-grafana (approved):

aknuds1:
Moving the file to contribute/engineering/, since we agree that contribute/style-guides is the wrong directory. I chose "engineering" as a new directory since the document is aimed at the engineering side of the org.

PR #30544, opened 2021-01-22
Admin: Fixes so form values are filled in from backend

### LINKED ISSUE #27121: Signup form for newly invited user does not have default fields populated
**What happened**:
When an invited user clicks on a Grafana invite link, they are presented with a blank signup form. E.g.
Screen Shot 2020-08-21 at 10 24 36 AM

The signup form should be pre-populated with the user's email/username.

This appears to work fine for older Grafana versions that use AngularJS to render this form. Judging by the code, it's also meant to be pre-populated in the React version, but something's going wrong somewhere.

As an admin, invite a new user to an org, and load the form on the invite link that's sent to them.

The code that generates the form (https://github.com/grafana/grafana/blob/master/public/app/features/users/SignupInvited.tsx) appears to successfully load the `initFormModel` from the back-end - which contains the new user's email. This model should be applied as the form's `defaultValues`. At runtime, in the React inspector in Google Dev Tools, those `defaultValues` do in fact show up as being set correctly for the form. But due to glitch somewhere, the values don't actually show up in the UI.

My strong suspicion is that the underlying issue is this: https://github.com/react-hook-form/react-hook-form/issues/1150. This recommends that for default values pulled from an API to be shown, `reset({ defaultValues })` should be called at `useEffect` in order to populate the form. Unfortunately my React-fu is insufficient to properly attempt a fix along these lines.

**Environment**:
- Grafana version: 7.1.4 (also reproduced on other 7.x; bug does not exist in 6.4.1)
- OS Grafana is installed on: Linux or macOS
- User OS & Browser: macOS + Chrome or Safari
- Grafana plugins: N/A
- Others: N/A

hugohaggmark:
The issue #27121 describes beautifully what the issue is. This PR solves the default values issue using a small workaround. We wait for the async form data to arrive before rendering the form. 
This could probably also be solved using `useForm` hook directly in the component and the `reset` function.

I also took the time and added some tests and refactored away redux connect in favor of useSelector.

Enjoy the review: 🎉 

Fixes #27121

[You must setup stmp settings in your custom.ini file to send an invite.](https://community.grafana.com/t/gmail-for-notifications/1895)

### DIFF
--- a/public/app/features/users/SignupInvited.test.tsx
+++ b/public/app/features/users/SignupInvited.test.tsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import * as redux from 'react-redux';
+import { render, screen, waitFor, within } from '@testing-library/react';
+import SignupInvitedPage from './SignupInvited';
+import { backendSrv } from '../../core/services/backend_srv';
+import userEvent from '@testing-library/user-event';
+
+jest.mock('app/core/core', () => ({
+  contextSrv: {
+    user: { orgName: 'Invited to Org Name' },
+  },
+}));
+
+jest.mock('@grafana/runtime', () => ({
+  ...((jest.requireActual('@grafana/runtime') as unknown) as object),
+  getBackendSrv: () => backendSrv,
+}));
+
+const defaultGet = {
+  email: 'some.user@localhost',
+  name: 'Some User',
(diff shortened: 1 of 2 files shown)


Clarity-89 (approved):
Great job 🎉

PR #30554, opened 2021-01-22
TableInputCSV: migrated styles from sass to emotion

thisisobate:
This PR migrates styles from sass to emotion

Relates to #30354


packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -16,6 +17,27 @@ interface State {
...
+        bottom: 15px;
+        right: 15px;
+        border: 1px solid #222;
+        background: $online;
+        padding: 1px 4px;
+        font-size: 80%;
+      }
+    `,
+  };
+};
```
dprokop, line 39: Please place the getStyles function at the end of the file.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -16,6 +17,27 @@ interface State {
...
+      textarea {
+        height: 100%;
+        width: 100%;
+      }
+      footer {
+        position: absolute;
+        bottom: 15px;
+        right: 15px;
+        border: 1px solid #222;
+        background: $online;
```
dprokop, line 33: This line won't work, as there is not $online variable. Please use theme colors.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -16,6 +17,27 @@ interface State {
...
+        padding: 1px 4px;
+        font-size: 80%;
```
dprokop, line 35: In general we should not use any statically defined sizing/colors. Insead, please use theme variables :)

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -61,8 +83,9 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
   render() {
     const { width, height } = this.props;
     const { data } = this.state;
+    const styles = getStyles();
```
dprokop, line 86: As a matter of consistency please use `useStyles` hook for retrieving the styles. You can find some more info about this in the [styling documentation](https://github.com/grafana/grafana/blob/master/contribute/style-guides/styling.md#styling-with-theme)

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -61,8 +83,9 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
...
     return (
-      <div className="gf-table-input-csv">
+      <div className={styles.gfTableInputCsv}>
```
dprokop, line 88: For the naming convention you don't need to be 1-1 with the original class names. In this case I suggest changing the class name to `tableInputCsv`
```suggestion
      <div className={styles.tableInputCsv}>
```

dprokop (changes requested):
Left few comments to change:)

hugohaggmark (commented):
You're getting there but I agree with Dominiks comments on this PR.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
thisisobate: In a situation where the sizing is not defined in the theme variables file, is it advisable to use the statically defined sizing regardless?

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
thisisobate: Gotcha 😀🙌

thisisobate: It seems like I can't use the `useStyles` hook inside a class component.
Is there a workaround to this? I'm considering converting the component from a class to a functional component. WDYT? @dprokop

hugohaggmark: You're correct about hooks, Uche. But instead of converting this into a Functional component, (which is a splendid exercise for another time) you should use the `stylesFactory` function. There are plenty of examples of stylesFactory and Class components like [https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Logs/LogLabelStats.tsx](https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/Logs/LogLabelStats.tsx) for instance. Study how that Components interface `extends Themable` and uses `withTheme`.

I've also noticed that you have a lot of PRs opened at the same time now. I think you're wasting time context switching between these.
Try to focus on the task at hand i.e. migrate the styles and closing this PR. Then move on to the next PR, read all the comments, fix them, test the PR and try to get that PR closed and repeat.

thisisobate: Alright

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.story.internal.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 
-import TableInputCSV from './TableInputCSV';
+import { TableInputCSV } from './TableInputCSV';
+import { Meta } from '@storybook/react/types-6-0';
```
jackw, line 4: Minor - this is likely an auto import but there shouldn't be a need to reference `/types-6-0` here. `import { Meta } from '@storybook/react` should be enough.

hugohaggmark (approved):
This is looking great and works as expected, nice work 🎉  I left some final nits for your consideration.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -87,4 +97,28 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
...
+          width: 100%;
+        }
+        footer {
+          position: absolute;
+          bottom: 15px;
+          right: 15px;
+          border: 1px solid #222;
+          background: ${theme.palette.online};
+          padding: 1px 4px;
+          font-size: 80%;
```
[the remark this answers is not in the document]
thisisobate: I could only find `4px` as a sizing value in the theme.
Although I could find values such as `1px` in the theme, it references a different type of property- in this case, borders.

I think it makes sense for us to always use values that relate to the given CSS property at hand.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -87,4 +97,28 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
...
-export default TableInputCSV;
+export const TableInputCSV = withTheme(unThemedTableInputCSV);
+TableInputCSV.displayName = 'TableInputCSV';
+
+const getStyles = stylesFactory(
+  (theme: GrafanaTheme): TableInputCSVStyle => {
+    return {
+      TableInputCsv: css`
+        position: relative;
+        textarea {
```
jackw, line 108: Nit: It could be a nice addition to swap these element selectors out for class names if possible. My reasoning is that element selectors tie semantics to styling which can lead to increased maintenance as a future developer who edits the HTML tags will also need to change the CSS selectors. Class names allows us to separate these concerns.

jackw (commented):
Looking really good @thisisobate. 👏 

I gave it quick scan and left a couple of minor comments.

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.story.internal.tsx
hugohaggmark: Good catch Jack!

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
thisisobate: Alright 👍
Thanks @jackw

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -78,10 +81,10 @@ export class UnThemedTableInputCSV extends React.PureComponent<Props, State> {
           placeholder="Enter CSV here..."
           value={this.state.text}
           onChange={this.onTextChange}
-          className="gf-form-input"
+          className={classNames(styles.textarea, 'gf-form-input')}
```
hugohaggmark, line 84: @thisisobate shouldn't you use `cx` from emoition to combine class names instead? Maybe I'm missing something, WDYT @jackw

thisisobate: You're right Hugo.
Would update the changes right away

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -59,19 +70,20 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
...
     return (
-      <div className="gf-table-input-csv">
+      <div className={styles.tableInputCsv}>
         <textarea
           style={{ width, height }}
           placeholder="Enter CSV here..."
           value={this.state.text}
           onChange={this.onTextChange}
-          className="gf-form-input"
+          className={cx(styles.textarea, 'gf-form-input')}
```
hugohaggmark, line 83: Aha now I see what is going on, I think maybe you should use the `TextArea` component from grafana-ui and then you don't need to combine the `gf-form-input` style?

thisisobate: Yeah...I think that would work perfectly fine

hugohaggmark:
Still looks and works as expected!

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -1,16 +1,28 @@
...
   height: string | number;
   onSeriesParsed: (data: DataFrame[], text: string) => void;
 }
 
+interface TableInputCSVStyle {
+  tableInputCsv: string;
+  textarea: string;
+  footer: string;
+}
+
```
hugohaggmark, line 25: nit: move this closer to the getStyles function. This will make it easier to read the props for the Component.

hugohaggmark (approved):
Still looks and works as expected!

hugohaggmark:
@dprokop friendly reminder that you're blocking this PR, than you

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -87,4 +94,34 @@ export class TableInputCSV extends React.PureComponent<Props, State> {
...
+TableInputCSV.displayName = 'TableInputCSV';
+
+interface TableInputCSVStyle {
+  tableInputCsv: string;
+  textarea: string;
+  footer: string;
+}
+
+const getStyles = stylesFactory(
+  (theme: GrafanaTheme): TableInputCSVStyle => {
```
dprokop, line 107: FYI- TS did a good job so far inferring the return type of the style getters, so I think you should not need return type annotation:)

dprokop (approved):
Works good!

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
thisisobate: Alright. would update right away

packages/grafana-ui/src/components/TableInputCSV/TableInputCSV.tsx
```diff
@@ -1,9 +1,15 @@
 import React from 'react';
 import debounce from 'lodash/debounce';
+import { css } from 'emotion';
 import { DataFrame, CSVConfig, readCSV } from '@grafana/data';
 import { Icon } from '../Icon/Icon';
+import { GrafanaTheme } from '@grafana/data';
+import { Themeable } from '../../types/theme';
+import { TextArea } from '../TextArea/TextArea';
+import { stylesFactory } from '../../themes';
+import { withTheme } from '../../themes/index';
```
dprokop, line 10:
```suggestion
import { stylesFactory, withTheme } from '../../themes';
```

dprokop (approved):

dprokop (approved):

PR #30565, opened 2021-01-22
GraphOG: Fix height issue in Firefox

zoltanbedi:

@slim-bean found an issue after #29776 that the logs panel in Explore looks like this in Firefox.

Screenshot 2021-01-22 at 21 05 36

After the changes it looks like this.
Screenshot 2021-01-22 at 21 00 41


leeoniya:
we've been calling it GraphOG :D

ryantxu (approved):

PR #30571, opened 2021-01-23
Explore: Replaces navbar-button and overriden explore button css classes with ToolbarButton and cleans up scss & markup, removes ResponsiveButton

torkelo:
* [x] Replace all the toolbar buttons in explore to use Toolbar
* [x] Move some of the responsive logic of ResponsiveButton into ToolbarButton
* [x] Remove now unused scss
* [x] Remove ResponsiveButton
* [x] Fixes styling issues with the live buttons (the hover state was bad, and the paused state)
* [x] Fixes so that in paused live state the button says paused
* [x] Needed to support a transparent toolbar button for the share & star button in dashboard, so refactored toolbar button to use a new variant type (default, active, transparent, etc) 

Tested a bunch and think all the existing behavior is there (responsive etc)


packages/grafana-ui/src/components/Button/ToolbarButton.tsx
```diff
@@ -21,24 +22,28 @@ export interface Props extends HTMLAttributes<HTMLButtonElement> {
   /** reduces padding to xs */
   narrow?: boolean;
   /** variant */
-  variant?: ButtonVariant;
-  /** Changes border color to orange */
-  active?: boolean;
+  variant?: ToolbarButtonVariant;
+  /** Hide any children and only show icon */
+  iconOnly?: boolean;
```
torkelo, line 27: this is a utility prop that makes the ResponsiveButton obsolete, a nice way to toggle rendering the children (text) via bool property

hugohaggmark (commented):
Code looks great, but I'll leave it to @grafana/observability-squad to review and test this too.

aocenas:
Seems like the Live functionality in Explore is a bit broken. Checking what is wrong now, because it seems like some of it is broken even from previous releases.

public/app/features/explore/LiveTailButton.tsx
```diff
@@ -107,25 +96,23 @@ export function LiveTailButton(props: LiveTailButtonProps) {
...
-          splitted={splitted}
-          buttonClassName={classNames('btn navbar-button', styles.liveButton, {
-            [`btn--radius-right-0 explore-active-button ${styles.noRightBorderStyle}`]: isLive,
-            [styles.isLive]: isLive && !isPaused,
-            [styles.isPaused]: isLive && isPaused,
-          })}
+        <ToolbarButton
+          iconOnly={splitted}
+          variant={buttonVariant}
           icon={!isLive ? 'play' : 'pause'}
```
aocenas, line 109: This should be using `isPaused`. The confusion is that `isLive` means you are in live tailing mode, which then can be running or paused. This bug is here since 7.0 though.

public/app/features/explore/LiveTailButton.tsx
```diff
@@ -107,25 +96,23 @@ export function LiveTailButton(props: LiveTailButtonProps) {
...
-          iconClassName={isLive ? 'icon-brand-gradient' : undefined}
           onClick={onClickMain}
-          title={'\xa0Live'}
-        />
+        >
+          {isPaused ? 'Live' : 'Paused'}
```
aocenas, line 112: This should be also other way around.

aocenas (changes requested):

public/app/features/explore/LiveTailButton.tsx
torkelo: ah, of course. :see_no_evil:

public/app/features/explore/LiveTailButton.tsx
torkelo: fixed

aocenas (approved):
Looks good now

PR #30578, opened 2021-01-24
Provisioning: Add organizations provisioning

### LINKED ISSUE #12119: Organisations provisioning
Currently configuring organisations and LDAP group mapping to organisation roles in an automated way is quite difficult. If you start with no organisations and want to configure them in an automatic way, you need to:

1. Start grafana with no group mapping in ldap.toml
2. Call the grafana API multiple times to create the organisations and get their IDs
3. Generate a new ldap.toml with the IDs of the organisations
4. Restart grafana

Especially in a Kubernetes environment this becomes very difficult to engineer right. We want to be able to recreate the environment with all organisations in an automated way. It would be a lot easier if provisioning at startup-time of organisations would be possible.

Maybe grafana-cli might help there, so that you could create the orgs before starting up grafana. Also, it should be possible to specify fixed IDs, so that you can also provide a static ldap.toml file (that you could generate with an init container, for example).

philschleier:

This PR adds a provisioner that allows the creation, modification and deletion of orgs from a config file as follows:
```
orgs:
- id: 1
  name: First Org
- id: 2
  name: Second Org
  preferences:
    # Needs to be known up front
    homeDashboardId: 1
    timezone: browser
    theme: light
deleteOrgs:
- id: 3 
```

Closes #12119

From my point of view, this already implements the core requirement, which is org provisioning. I'm still unclear on a couple things, hence the Draft PR:

* I've only added a v1 config - is a v0 config needed?
* There's no auto assigned admin for these orgs. In case of initial_admin_creation, should I do a get for the admin_user and add that userId as admin to these orgs?
* As it is now, this will simply run after the initial creation of admin and org, hence it can rename/overwrite the default org. That appears to be fine to me, but it might lead to some confusion.
* Selecting any of theses orgs as auto_assign_org_id should work fine.
* Org preferences: Are optionally updated, if present in the config (pointer), which seems to be uncommon for the codebase (Comments?). Needing to know the dashboard ID in order to set it as home dashboard defeats the purpose I think. Would it make sense to add a IsHomeDashboard option to the dashboard provisioning, similar to "default datasource"?

I have not yet updated any docs until I have more info on the above.

### DIFF
--- a/pkg/services/provisioning/orgs/config_reader_test.go
+++ b/pkg/services/provisioning/orgs/config_reader_test.go
@@ -1,6 +1,7 @@
 package orgs
 
 import (
+	"context"
 	"testing"
 
 	"github.com/grafana/grafana/pkg/bus"
@@ -33,7 +34,7 @@ func TestOrgAsConfig(t *testing.T) {
 		Convey("Two configured orgs", func() {
 			Convey("no org in database", func() {
 				dc := newOrgProvisioner(logger)
-				err := dc.applyChanges(twoOrgsConfig)
+				err := dc.applyChanges(context.Background(), twoOrgsConfig)
 				if err != nil {
 					t.Fatalf("applyChanges return an error %v", err)
 				}
@@ -51,7 +52,7 @@ func TestOrgAsConfig(t *testing.T) {
 
 				Convey("should update one org", func() {
 					dc := newOrgProvisioner(logger)
-					err := dc.applyChanges(twoOrgsConfig)
+					err := dc.applyChanges(context.Background(), twoOrgsConfig)
 					if err != nil {
 						t.Fatalf("applyChanges return an error %v", err)
 					}
@@ -73,7 +74,7 @@ func TestOrgAsConfig(t *testing.T) {
 
 				Convey("should have two new orgs", func() {
 					dc := newOrgProvisioner(logger)
-					err := dc.applyChanges(twoOrgsConfigPurgeOthers)
+					err := dc.applyChanges(context.Background(), twoOrgsConfigPurgeOthers)
 					if err != nil {
 						t.Fatalf("applyChanges return an error %v", err)
 					}
@@ -88,13 +89,13 @@ func TestOrgAsConfig(t *testing.T) {
 
 		Convey("broken yaml should return error", func() {
 			reader := &configReader{}
-			_, err := reader.readConfig(brokenYaml)
+			_, err := reader.readConfig(context.Background(), brokenYaml)
 			So(err, ShouldNotBeNil)
 		})
 
 		Convey("skip invalid directory", func() {
 			cfgProvider := &configReader{log: log.New("test logger")}
-			cfg, err := cfgProvider.readConfig("./invalid-directory")
+			cfg, err := cfgProvider.readConfig(context.Background(), "./invalid-directory")
 			if err != nil {
 				t.Fatalf("readConfig return an error %v", err)
 			}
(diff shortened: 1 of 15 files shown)


zifter:
Any updates for this pr?

bhiravabhatla:
Team,

Any movement on this.

haarchri:
LGTM - can we merge this feature after rebase ?

joschi36:
@philschleier Thanks for your contribution.
Maybe @wbrowne @marefr can have a look on this?

darkleaf:
ping

tanandy:
up

jessabe:
Will be discussing this with @pkolyvas when they are back from holidays. :)

pkg/services/provisioning/orgs/orgs.go
```diff
@@ -0,0 +1,96 @@
...
+		cmd := &models.GetOrgByIdQuery{Id: org.Id}
+		err := bus.Dispatch(ctx, cmd)
+		if err != nil && !errors.Is(err, models.ErrOrgNotFound) {
+			return err
+		}
+
+		if errors.Is(err, models.ErrOrgNotFound) {
+			provisioner.log.Info("inserting org from configuration ", "id", org.Id, "name", org.Name)
+			insertCmd := createInsertCommand(org)
+			if err := bus.Dispatch(ctx, insertCmd); err != nil {
```
ying-jeanne, line 49: where the bus removal, we would need to inject org/preferences service here and remove all the buses.

pkg/services/sqlstore/org.go
```diff
@@ -151,11 +164,11 @@ func createOrg(name string, userID int64, engine *xorm.Engine) (models.Org, erro
 
 // CreateOrgWithMember creates an organization with a certain name and a certain user as member.
 func (ss *SQLStore) CreateOrgWithMember(name string, userID int64) (models.Org, error) {
-	return createOrg(name, userID, ss.engine)
+	return createOrg(0, name, userID, ss.engine)
 }
 
 func CreateOrg(ctx context.Context, cmd *models.CreateOrgCommand) error {
-	org, err := createOrg(cmd.Name, cmd.UserId, x)
+	org, err := createOrg(cmd.Id, cmd.Name, cmd.UserId, x)
```
ying-jeanne, line 171: here we need to make sure we don't use UserUID instead

pkg/services/provisioning/orgs/config_reader_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	twoOrgsConfig            = "testdata/two-orgs"
+	twoOrgsConfigPurgeOthers = "testdata/insert-two-delete-two"
+	brokenYaml               = "testdata/broken-yaml"
+
+	fakeRepo *fakeRepository
+)
+
+func TestOrgAsConfig(t *testing.T) {
+	Convey("Testing org as configuration", t, func() {
```
ying-jeanne, line 25: all test should be convey free now.

ying-jeanne:
Hi @philschleier, 

Thanks for your contribution, since our backend structure changed quite a bit recently, this PR would have quite some changes needed before merging. I saw you are creating the branch from your own repository, could you pls whether give me write permission or create from Grafana directly. So we can corp easily. 

Best Regards,

MrGadroo:
This seems like a basic feature for provisioning other things. Can anyone merge this or provide any support on this?

jicki:
Has this feature been updated?

MrGadroo:
Initially i wrote a utility to create org/dashbords/datasource using api to achieve this, but that had some issues so I finally found a better workaround for this. If anyone is stuck please follow below steps. Might not be the best approach but works like a charm. 
1. Linked grafana with postgres so i could do manipulation in db.
2. Created a blank DB added two relations for org and org_user(extracted from manually created environment) during initial setup. It can be handled with kubernetes jobs as well. 
3. Created an image for grafana dashboards (separated via folder for different orgs), used it for provisioning.
4. Added above image in initcontainer with emptyDir volume sharing both init and main container.
5. Used a cm for datasource provisioning.
PS: With this I am finally able to achieve dashboard provisioning in different orgs and also able to maintain versioning.

paulwer:
any updates?

ahamboeck:
any updates?

PR #30581, opened 2021-01-25
Fix CircleCI config syntax

simPod:

The syntax was invalid

### DIFF
--- a/packages/grafana-toolkit/config/circleci/config.yml
+++ b/packages/grafana-toolkit/config/circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
     - run:
         name: Continue if this plugin has a provisioning path
         command: |
-          [ -z "$PROVISIONING_REPO_PATH" ] && circleci-agent step halt
+          [ -z "${PROVISIONING_REPO_PATH}" ] && circleci-agent step halt
     - checkout
     - add_ssh_keys:
         fingerprints:
@@ -206,7 +206,7 @@ jobs:
     - run:
         name: Continue if this plugin has a provisioning path
         command: |
-          [ -z "$PROVISIONING_REPO_PATH" ] && circleci-agent step halt
+          [ -z "${PROVISIONING_REPO_PATH}" ] && circleci-agent step halt
     - checkout
     - attach_workspace:
         at: .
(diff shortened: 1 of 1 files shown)


aknuds1 (commented):
How is the syntax wrong? To me it doesn't look as if you're fixing anything. I can only see whitespace changes and removing curly braces from shell variable expansions (which shouldn't be a good change).

simPod:
Whitespace changes are significant in YAML. The whitespace indent is invalid and won't run in CircleCI (syntax error).

The same applies for `"${` combination.

After this change CI runs. Before it does not because of invalid syntax. CircleCI cannot parse it.

simPod:
https://app.circleci.com/pipelines/github/simPod/GrafanaJsonDatasource/171/workflows/2c908635-0af6-4def-8963-9f03ff883a57/jobs/358

aknuds1:
> Whitespace changes are significant in YAML. The whitespace indent is invalid and won't run in CircleCI (syntax error).

Yeah I know whitespace is significant in YAML, but bad indentation can be difficult to spot. I see the problem you're fixing now, thanks.

> The same applies for `"${` combination.

Are you sure `${}` is a syntax problem in this YAML? I see you removed those changes in the meantime?

aknuds1:
@srclosson Can you review this? I don't work on this CircleCI logic.

simPod:
@aknuds1 

> Are you sure ${} is a syntax problem in this YAML?

no, seems insignificant. I must have misinterpreted something CircleCI tried to say to me. Nvm. Only indentation change matters here.

dprokop (approved):
This looks good to me, but let's see what @srclosson says :)

srclosson (approved):
Yes, looks fine to me too. We don't really use this circleci config anymore. It's a long story, and frankly I suspect a lot of this can be cleaned out since we are moving to drone in enterprise plugins eventually anyway. It does make me wonder though, if this needed to get fixed, who's using it?

simPod:
I'm using it for building my plug-in. Is there better way now?

srclosson:
Not right now. Go ahead and keep using it so long as it's working for you. There is no intent to update it anymore though. Heads up.

simPod:
Just FYI I used the grafana toolkit command to update the CircleCI definition.

PR #30588, opened 2021-01-25
PageToolbar: Extract navbar styles and layout into a modern emotion based component

torkelo:
This component aims to replace all the top toolbar / navbars we have (Dashboard, Explore, Panel Edit & Alerting). It's a bit tricky to support all use cases, for example, the dashboard has a clickable page title that can be composed of parent (folder) and dashboard title, not something Explore has. Panel Edit has a back button etc. But opted to handle all uses cases in the component to more easily share the styles and alignment but can be refactored out later when the full picture emerges. 

This only updates the dashboard toolbar, leaving the other views for follow-up PRs.  The aim here is to improve the responsive behavior for the dashboard toolbar and have buttons wrap instead of just hiding them. 

TODO:
* [x] Basic PageToolbar, support pageIcon and title, and right side action buttons via children
* [x] Add support for clickable title & parent
* [x] Add support for left side actions (share & favorite etc) 
* [x] Add support for responsive layouts where buttons wrap
* [x] Folder name hides on smaller screens
* [x] Fix text ellipsis on long titles & folder names
* [x] Hide some actions in playlist mode
* [x] Support view panel mode (back button) 
* [x] Restore support fo tv & kiosk modes
* [x] Fix mobile view when the menu is absoluty position in top left corner 
* [ ] Update Enterprise and test their buttons and precense avatars

Story so far:
Screenshot from 2021-01-25 09-16-46

### DIFF
--- a/packages/grafana-ui/src/components/Button/ToolbarButton.tsx
+++ b/packages/grafana-ui/src/components/Button/ToolbarButton.tsx
@@ -94,24 +94,19 @@ const getStyles = (theme: GrafanaTheme) => {
       label: toolbar-button;
       display: flex;
       align-items: center;
-      label: toolbar-button;
-      background: ${theme.colors.bg1};
-      border: 1px solid ${theme.colors.border2};
       height: ${theme.height.md}px;
       padding: 0 ${theme.spacing.sm};
       border-radius: ${theme.border.radius.sm};
       line-height: ${theme.height.md - 2}px;
       font-weight: ${theme.typography.weight.semibold};
-
+      border: 1px solid ${theme.colors.border2};
       &:focus {
         outline: none;
       }
-
       &[disabled],
       &:disabled {
         cursor: not-allowed;
         opacity: 0.5;
-
         &:hover {
           color: ${theme.colors.textWeak};
           background: ${theme.colors.bg1};
@@ -121,28 +116,15 @@ const getStyles = (theme: GrafanaTheme) => {
     default: css`
       color: ${theme.colors.textWeak};
       background-color: ${theme.colors.bg1};
-
       &:hover {
         color: ${theme.colors.text};
         background: ${styleMixins.hoverColor(theme.colors.bg1, theme)};
-      }
-
-      &[disabled],
-      &:disabled {
-        cursor: not-allowed;
-        opacity: 0.5;
-
-        &:hover {
-          color: ${theme.colors.textWeak};
-          background: ${theme.colors.bg1};
-        }
       }
     `,
     active: css`
       color: ${theme.palette.orangeDark};
       border-color: ${theme.palette.orangeDark};
       background-color: transparent;
-
       &:hover {
         color: ${theme.colors.text};
         background: ${styleMixins.hoverColor(theme.colors.bg1, theme)};
@@ -169,11 +151,6 @@ const getStyles = (theme: GrafanaTheme) => {
     `,
     content: css`
       flex-grow: 1;
-      display: none;
-
-      @media only screen and (min-width: ${theme.breakpoints.md}) {
-        display: block;
-      }
     `,
     contentWithIcon: css`
       padding-left: ${theme.spacing.sm};
--- a/packages/grafana-ui/src/components/Button/ToolbarButton.tsx
+++ b/packages/grafana-ui/src/components/Button/ToolbarButton.tsx
@@ -7,6 +7,7 @@ import { Tooltip } from '../Tooltip/Tooltip';
 import { Icon } from '../Icon/Icon';
 import { getPropertiesForVariant } from './Button';
 import { isString } from 'lodash';
+import { selectors } from '@grafana/e2e-selectors';
 
 export interface Props extends ButtonHTMLAttributes<HTMLButtonElement> {
   /** Icon name */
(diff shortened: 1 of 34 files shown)


packages/grafana-ui/src/components/PageLayout/PageToolbar.tsx
```diff
@@ -0,0 +1,199 @@
...
+                size="xxl"
+                surface="dashboard"
+                aria-label={selectors.components.BackButton.backArrow}
+                onClick={onGoBack}
+              />
+            </div>
+          )}
+          <div className={styles.titleWrapper}>
+            {parent && onClickParent && (
+              <a onClick={onClickParent} className={cx(styles.titleLink, styles.parentLink)}>
```
Clarity-89, line 78: This should be either a link with ref or a button with onClick.

packages/grafana-ui/src/components/PageLayout/PageToolbar.tsx
```diff
@@ -0,0 +1,199 @@
...
+                {parent} <span className={styles.parentIcon}>/</span>
+              </a>
+            )}
+            {onClickTitle && (
+              <a onClick={onClickTitle} className={styles.titleLink}>
```
Clarity-89, line 83: Same as above.

packages/grafana-ui/src/components/PageLayout/PageToolbar.tsx
```diff
@@ -0,0 +1,199 @@
...
+                {title}
+              </a>
+            )}
+            {!onClickTitle && <div className={styles.titleText}>{title}</div>}
+          </div>
+          {leftItems &&
+            leftItems.map((child, index) => (
```
Clarity-89, line 90:
```suggestion
            {leftItems?.map((child, index) => (
```

public/app/features/dashboard/components/DashNav/DashNav.tsx
```diff
@@ -172,74 +182,42 @@ class DashNav extends PureComponent<Props> {
...
-          classSuffix="save"
-          tooltip="Add panel"
-          icon="panel-add"
-          onClick={onAddPanel}
-          iconType="mono"
-          iconSize="xl"
-          key="button-panel-add"
-        />
-      );
+    const tvButton = <ToolbarButton tooltip="Cycle view mode" icon="monitor" onClick={this.onToggleTVMode} />;
```
Clarity-89, line 201: I think this also needs `key`.

public/app/features/dashboard/components/DashNav/DashNav.tsx
```diff
@@ -172,74 +182,42 @@ class DashNav extends PureComponent<Props> {
...
+    const timeControls = (
+      <DashNavTimeControls dashboard={dashboard} location={location} onChangeTimeZone={updateTimeZoneForSession} />
```
Clarity-89, line 203: This too is missing `key`.

Clarity-89 (approved):
Great job! A few minor comments. 

Also not sure if relevant for this PR but noticed two issues:
1. `ButtonSelect` seems to have an unused prop `maxMenuHeight`, which ends up being passed to `button` causing a warning, as that's not a valid HTML attribute. 
2. The selected value in the time picker looks weird on slightly narrower screens:
Screenshot 2021-01-26 at 13 45 09

public/app/features/dashboard/components/DashNav/DashNav.tsx
torkelo: ah, good catch!

packages/grafana-ui/src/components/PageLayout/PageToolbar.tsx
torkelo: yes, was copied from DashNav this way but your right this should be a button

packages/grafana-ui/src/components/Button/ToolbarButton.tsx
```diff
@@ -31,7 +32,20 @@ export type ToolbarButtonVariant = 'default' | 'primary' | 'destructive' | 'acti
...
+      icon,
+      className,
+      children,
+      imgSrc,
+      fullWidth,
+      isOpen,
+      narrow,
+      variant = 'default',
+      iconOnly,
+      'aria-label': ariaLabel,
```
hugohaggmark, line 46: Nice trick, I'll try to remember this one.

public/app/features/explore/LiveTailButton.tsx
```diff
@@ -97,21 +97,19 @@ export function LiveTailButton(props: LiveTailButtonProps) {
...
+      <Tooltip content={isLive ? <>Pause the live stream</> : <>Live stream your logs</>} placement="bottom">
         <ToolbarButton
           iconOnly={splitted}
           variant={buttonVariant}
-          icon={!isLive || isPaused ? 'play' : 'pause'}
+          icon={!isLive ? 'play' : 'pause'}
           onClick={onClickMain}
         >
-          {isLive && isPaused ? 'Paused' : 'Live'}
+          {isPaused ? 'Live' : 'Paused'}
```
hugohaggmark, line 112: Should this be the other way around or maybe this is the call to action?
```suggestion
          {isPaused ? 'Paused' : 'Live'}
```

hugohaggmark (approved):
Super monumental work on this 🎉   I've tested some different dashboards in several responsive sizes. The only thing I found was the stop button for playlists that seems a bit different from before:

image

public/app/features/explore/LiveTailButton.tsx
torkelo: ah, good catch. Wrong merge when rebasing this, fixed this in another PR.

oddlittlebird:
@grafana/docs-squad, this might affect some of our dashboard documentation and docs for dashboard features such as variables.

PR #30590, opened 2021-01-25
Logging: sourcemap support for frontend stacktraces

domasx2:
Adds source map support for logging frontend stacktraces. Will attempt on best effort basis to find sourcemaps for compiled core & plugin js bundles and indicate original source location instead of compiled file location in the stacktrace. It's much more useful to see the original source code location, as compiled files are minimized which effectively obfuscates the code. 
Using https://github.com/go-sourcemap/sourcemap. Note that it seems to fail to determine function name for react components. But not a big deal as long as file/line/column are correct. 

Before (no sourcemap usage):
error_no_sourcemap

After (using sourcemap):
error_with_sourcemap

Todos:
- [x] refactor for better testability
- [x] add unit tests
- [x] sanitize paths before attempting to read from fs?

Related to #27993

Would be awesome to get opinions on a few things:

Currently all the logic is smushed together in `frontend_logging.go`. I'd like to decouple file loading, source map resolution and the actual logging and test these separately. What is the best way to go about it? Should I create an injectable `SourceMapService` and add it to `HTTPServer`, which is then passed to the frontend logging handler?

Also, current impl. keeps a cache of loaded sourcemaps, which can increase memory footprint somewhat. Is that ok? Maybe should use some cache impl that limits lifetime of the cached items, or scope the cache only to the processing of a single stacktrace?

Thanks!


pkg/api/frontend_logging.go
```diff
@@ -2,17 +2,55 @@ package api
...
 )
 
 var frontendLogger = log.New("frontend")
 
+type sourceLocation struct {
+	file     string
+	function string
+	line     int
+	col      int
+	pluginId string
```
aknuds1, line 31:
```suggestion
	pluginID string
```

pkg/api/frontend_logging.go
```diff
@@ -2,17 +2,55 @@ package api
...
+}
+
+type sourceMapLocation struct {
+	dir      string
+	path     string
+	pluginId string
```
aknuds1, line 37:
```suggestion
	pluginID string
```

pkg/api/frontend_logging.go
```diff
@@ -2,17 +2,55 @@ package api
...
+}
+
+type sourceMap struct {
+	consumer *sourcemap.Consumer
+	pluginId string
```
aknuds1, line 42:
```suggestion
	pluginID string
```

aknuds1 (commented):
Just some style nits, since Go initialisms should be in consistent case.

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+
 func (value *frontendSentryExceptionValue) FmtMessage() string {
 	return fmt.Sprintf("%s: %s", value.Type, value.Value)
 }
 
 func (value *frontendSentryExceptionValue) FmtStacktrace() string {
 	var stacktrace = value.FmtMessage()
 	for _, frame := range value.Stacktrace.Frames {
-		stacktrace += fmt.Sprintf("\n  at %s (%s:%v:%v)", frame.Function, frame.Filename, frame.Lineno, frame.Colno)
+		mappedLocation, err := resolveSourceLocation(frame.Filename, frame.Lineno, frame.Colno)
```
aknuds1, line 171: Just a nit, but could you just pass on the frame? It would make for more ergonomic code, especially when the function will eventually need more or less arguments. There's no benefit to passing individual fields of `frame`.
```suggestion
		mappedLocation, err := resolveSourceLocation(frame)
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+		if err != nil {
+			frontendLogger.Error("Error resolving stack trace frame source location.", "err", err)
```
aknuds1, line 173: You should return here, since you shouldn't use the other return value(s) when an error is returned.
```suggestion
			frontendLogger.Error("Error resolving stack trace frame source location.", "err", err)
			return stacktrace
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
 }
 
+func guessSourceMapLocation(sourceUrl string) (*sourceMapLocation, error) {
+	u, err := url.Parse(sourceUrl)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasPrefix(u.Path, "/public/build/") {
+		return &sourceMapLocation{
+			dir:      setting.StaticRootPath,
```
aknuds1, line 76: Can you pass in a `setting.Cfg` parameter, and avoid using this global variable? We are doing our very best to get rid of global variables in the backend.

Not sure how to get `setting.Cfg` passed into your code, but that's how it should be done.

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
 
 func (value *frontendSentryExceptionValue) FmtStacktrace() string {
 	var stacktrace = value.FmtMessage()
 	for _, frame := range value.Stacktrace.Frames {
-		stacktrace += fmt.Sprintf("\n  at %s (%s:%v:%v)", frame.Function, frame.Filename, frame.Lineno, frame.Colno)
+		mappedLocation, err := resolveSourceLocation(frame.Filename, frame.Lineno, frame.Colno)
+		if err != nil {
+			frontendLogger.Error("Error resolving stack trace frame source location.", "err", err)
+		}
+		if mappedLocation != nil {
```
aknuds1, line 175: Nit, but it's better to return early here, for easier reading.
```suggestion
		if mappedLocation == nil {
			return stacktrace
		}
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+	}
+	return nil, nil
+}
+
+func resolveSourceLocation(url string, line int, column int) (*sourceLocation, error) {
+	smap, err := getSourceMap(url)
+	if err != nil {
+		return nil, err
+	}
+	if smap != nil {
```
aknuds1, line 146: Nit, but it's better to return early to reduce nesting.
```suggestion
	if smap == nil {
		return nil, nil
	}
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+		file, function, line, col, ok := smap.consumer.Source(line, column)
+		if ok {
```
aknuds1, line 148: It's more idiomatic to return early.
```suggestion
		if !ok {
			return nil, nil
		}
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
 
+func guessSourceMapLocation(sourceUrl string) (*sourceMapLocation, error) {
+	u, err := url.Parse(sourceUrl)
+	if err != nil {
+		return nil, err
+	}
+	if strings.HasPrefix(u.Path, "/public/build/") {
+		return &sourceMapLocation{
+			dir:      setting.StaticRootPath,
+			path:     path.Join("build", u.Path[len("/public/build/"):]) + ".map",
```
aknuds1, line 77: Should this be a filesystem path or a network path? If it's a filesystem path, you should use `filepath.Join`.

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+	defer sourceMapCache.Unlock()
+
+	if smap, ok := sourceMapCache.cache[sourceUrl]; ok {
+		return smap, nil
+	}
+	sourceMapLocation, err := guessSourceMapLocation(sourceUrl)
+	if err != nil {
+		return nil, err
+	}
+	if sourceMapLocation != nil {
```
aknuds1, line 106: It's more idiomatic to return early here, and avoid nesting. Also wondering if you actually want to delete the cache entry?
```suggestion
	if sourceMapLocation == nil {
		delete(sourceMapCache.cache, sourceURL)
		return nil, nil
	}
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+					path:     u.Path[len(pluginPrefix):] + ".map",
+					pluginId: route.PluginId,
+				}, nil
+			}
+		}
+	}
+	return nil, nil
+}
+
+func getSourceMap(sourceUrl string) (*sourceMap, error) {
```
aknuds1, line 95: In Go, initialisms should be written in consistent case.
```suggestion
func getSourceMap(sourceURL string) (*sourceMap, error) {
```

pkg/api/frontend_logging.go
```diff
@@ -28,14 +66,121 @@ type frontendSentryEvent struct {
...
+		if err != nil {
+			if os.IsNotExist(err) {
+				sourceMapCache.cache[sourceUrl] = nil
+				return nil, nil
+			}
+			return nil, err
+		}
+		defer func() {
+			if err := f.Close(); err != nil {
+				frontendLogger.Error("Failed to close source map file.", "err", err)
```
aknuds1, line 118: We generally don't punctuate log messages:
```suggestion
				frontendLogger.Error("Failed to close source map file", "err", err)
```

aknuds1 (commented):
Made some suggestions for more readable code mostly.

aknuds1:
> Currently all the logic is smushed together in frontend_logging.go. I'd like to decouple file loading, source map resolution and the actual logging and test these separately. What is the best way to go about it? Should I create an injectable SourceMapService and add it to HTTPServer, which is then passed to the frontend logging handler?

I'm not sure. Why would you need an injectable service? Adding those is best avoided unless actually necessary, since it's a heavy abstraction and our DI framework is also fragile. I think there's only one way to figure out how to test these things in separation (if that's necessary), and it would be to try to do it. In general, try to not abstract more than necessary, since it makes the code harder to maintain.

> Also, current impl. keeps a cache of loaded sourcemaps, which can increase memory footprint somewhat. Is that ok? Maybe should use some cache impl that limits lifetime of the cached items, or scope the cache only to the processing of a single stacktrace?

I'm not sure. @marefr do you have an opinion?

pkg/api/frontend_logging.go
domasx2: If there's no cache entry, and sourcemap is not found on the fs, I want to add cache entry with `nil` value so we don't try to find it in fs again later

pkg/api/frontend_logging.go
domasx2: right! :)  
- [ ] pass the frame

pkg/api/frontend_logging.go
aknuds1: OK thanks, updated my suggestion

pkg/api/frontend_logging.go
domasx2: That would return a partial stacktrace. My intention here was to treat unexpected sorucemap error as non critical and just log the untransformed line, still better than nothing. Makes sense?

pkg/api/frontend_logging.go
domasx2: oops
- [ ] use `filepath.Join`

pkg/api/frontend_logging.go
domasx2: - [ ] get `setting.Cfg` in here

domasx2: I was thinking that instead of this static location resolution here it would be better to somehow co-opt the existing static file resolution that is used for the `public/*` endpoitns. But could not figure out a way that does not require some extensive refac...

pkg/api/frontend_logging.go
aknuds1: In that case, you should put an `else` before the next `if`. Let me see if I can make a suggestion.

pkg/api/frontend_logging.go
aknuds1, line 175: I think this shouldn't be checked if `err` is non-nil, since you shouldn't trust other return values when an error is returned IMO:
```suggestion
		else if mappedLocation != nil {
```

aknuds1 (commented):
Updated a couple of suggestions.

pkg/api/frontend_logging.go
domasx2: I'll start remembering this, someday... :) 
- [ ] initialisims in all caps

marefr:
Sorry, I don't understand the full picture. Is this a problem only with plugins or in general all frontend code? Can you include some more details of log results before this PR, like errors in Grafana core itself and plugin errors.

pkg/api/frontend_logging.go
aknuds1: Dunno.

domasx2:
> Sorry, I don't understand the full picture. Is this a problem only with plugins or in general all frontend code? Can you include some more details of log results before this PR, like errors in Grafana core itself and plugin errors.

Updated description, added before/after examples of a logged stack trace, for the same error. It affects all frontend code in general, the example stack trace has code locations both from grafana core and from integrations plugin.

domasx2:
> I'm not sure. Why would you need an injectable service? Adding those is best avoided unless actually necessary, since it's a heavy abstraction and our DI framework is also fragile. I think there's only one way to figure out how to test these things in separation (if that's necessary), and that's to try to do it. In general, try to not abstract more than necessary, since it makes the code harder to maintain.

My concern was that tight integration makes tests harder to write & maintain, and wasn't sure what's the accepted practice in this codebase. Testing it as a whole works for me as well, will do that then. I think I only need to abstract away the file system interaction, but hopefully will be easy to do ad-hoc

> Actually one thing that I think would make sense would be to move frontend_logging.go to a sub-package api/frontendlogging, since the api package namespace is already too populated and difficult to reason about as a result.

Makes sense!
- [ ] move to separate subpackage

aknuds1:
> My concern was that tight integration makes tests harder to write & maintain, and wasn't sure what's the accepted practice in this codebase. Testing it as a whole works for me as well, will do that then. I think I only need to abstract away the file system interaction, but hopefully will be easy to do ad-hoc

Depends what you're after, but if you want to have loose coupling, you don't need DI.

On the subject of integration though, we do write integration tests and I find these easier to maintain in practice than tests that use a lot of stubbing/faking/mocking since the latter breaks down easily, and cause a lot of cognitive overhead.

pkg/api/frontend_logging.go
domasx2: done!

pkg/api/frontend_logging.go
domasx2: went over the code and rfaced to early returns where applicable

pkg/api/frontend_logging.go
domasx2: updated!

domasx2:
> On the subject of integration though, we do write integration tests and I find these easier to maintain in practice than tests that use a lot of stubbing/faking/mocking since the latter breaks down easily, and cause a lot of cognitive overhead to figure out how (and whether) things actually fit together.

Makes sense!
I refactored everything but the http handler into `api/frontendlogging` package and made a `SourceMapStore` struct to hold config & abstraction over file access,  wrote test that tests the whole thing in one piece

aknuds1:
Will have a look soon, but busy with a bug at the moment.

pkg/api/frontendlogging/source_maps.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+type sourceMap struct {
+	consumer *sourcemap.Consumer
+	pluginID string
+}
+
+type ReadSourceMapFn func(dir string, path string) ([]byte, error)
+
+func ReadSourceMapFromFs(dir string, path string) ([]byte, error) {
```
aknuds1, line 32: Nit, but initialisms should be in consistent case in Go.
```suggestion
func ReadSourceMapFromFS(dir string, path string) ([]byte, error) {
```

pkg/api/frontend_logging.go
```diff
@@ -1,104 +1,40 @@
...
-type frontendSentryEvent struct {
-	*sentry.Event
-	Exception *frontendSentryException `json:"exception,omitempty"`
-}
+type frontendLogMessageHandler func(c *models.ReqContext, event frontendlogging.FrontendSentryEvent) response.Response
 
-func (value *frontendSentryExceptionValue) FmtMessage() string {
-	return fmt.Sprintf("%s: %s", value.Type, value.Value)
-}
+func NewFrontendLogMessageHandler(store *frontendlogging.SourceMapStore) frontendLogMessageHandler {
```
aknuds1, line 15: Is this part kept in the `api` package to avoid cyclic imports? Otherwise it would be neat to move this as well to `api/frontendlogging` so `api` can be slimmed down :)

aknuds1 (approved):
Looks good to me now AFAICT! Thanks for moving stuff into api/frontendlogging :)

pkg/api/frontend_logging.go
domasx2: yup, exactly. Test utils for http handlers are in `api`. In such case they should be extracted as well, but felt like too big a scope for this PR :)

aknuds1: OK, agree then :)

pkg/api/frontendlogging/source_maps.go
domasx2: goddamit! one day...
- [ ] fix

PR #30597, opened 2021-01-25
Data proxy: Fix encoded characters in URL path should be proxied as encoded

### LINKED ISSUE #26870: Jaeger operations list does not work if service name contains forward slash
**What happened**: Attempting to list operations available for a service in the Jaeger plugin fails with a JS error if the service name contains a forward slash.

jaeger-services
console-error
api-request

URL-encoding of the service name was done in #25674 but the overall functionality still does not work.

**What you expected to happen**: Operations for the service are listed

**How to reproduce it (as minimally and precisely as possible)**: Have a service name that contains a forward slash. Try to use the Jaeger plugin to find and select a trace for this service.

**Anything else we need to know?**: Making the same API call to Jaeger directly does return a response. Possibly the URL-encoding may be getting un-done by the proxy. There's no way to use direct mode due to #25699.

**Environment**:
- Grafana version: 7.1.3
- User OS & Browser: Safari 13.1.2
- Grafana plugins:
- Others:

### LINKED ISSUE #31438: Datasource proxy is incorrectly handling escaped slashes in URL path
**What happened**:
A client who is using Cloud Alerting UI complained that trying to delete alert rule group whose name is prefixed with a slash "/" , eg `/my_group`, fails.
I investigated and found that:
1. Cloud Alerting UI plugin sends the delete rule group request via grafana's datasource proxy
2. Rule group name is encoded as a path part, eg: `DELETE https://domaslapinskas.grafana.net/api/datasources/proxy/5/rules/default/%2Ftest`
3. Grafana datasource proxy then calls `DELETE https://prometheus-us-central1.grafana.net/api/prom/rules/default//test` with the leading slash for the group name unescaped. This causes cortex to return 301 to `https://prometheus-us-central1.grafana.net/api/prom/rules/default/test`, removing the slash

Escaped slashes should be preserved when proxying datasource requests.

1. Spin up a grafana instance on grafana.com
2. Go to Cloud Alerting UI and create a rule group with leading slash, eg `/test`
3. Attempt to delete the single rule in this group, which should cause the rule to be deleted. It fails

Browser request from debug tools:
```
Request URL: https://domaslapinskas.grafana.net/api/datasources/proxy/5/rules/default/%2Ftest
Request Method: DELETE
Status Code: 301 
```

Logs from grafana backend:
```
2021-02-24 08:22:47 | t=2021-02-24T08:22:47+0000 lvl=info msg="Request Completed" logger=context userId=2 orgId=1 uname=domaslapinskas method=DELETE path=/api/datasources/proxy/5/rules/default//test status=301 remote_addr=88.119.159.231 time_ms=31 size=0 referer="https://domaslapinskas.grafana.net/a/grafana-alerting-ui-app/?tab=rules&rulessource=grafanacloud-domaslapinskas-prom" traceID=3678479b3ed0557e
-- | --
  |   | 2021-02-24 08:22:47 | t=2021-02-24T08:22:47+0000 lvl=info msg=Requesting logger=data-proxy-log url=https://prometheus-us-central1.grafana.net/api/prom/rules/default//test



```

**Environment**:
- Grafana version: 7.4.2

marefr:
Fix encoded characters in URL path should be proxied encoded in data proxy.

Fixes #26870 
Fixes #31438

Is it viable to replace `c.Params("*")` with a replace of the raw URL path?


pkg/api/dataproxy.go
```diff
@@ -59,3 +62,9 @@ func ensureProxyPathTrailingSlash(originalPath, proxyPath string) string {
 
 	return proxyPath
 }
+
+var proxyPathRegexp = regexp.MustCompile("^\\/api\\/datasources\\/proxy\\/[\\d]+\\/?")
+
+func getProxyPath(originalRawPath string) string {
+	return proxyPathRegexp.ReplaceAllString(originalRawPath, "")
```
aknuds1, line 69: Can you just move the call to `ensureProxyPathTrailingSlash` here, is there any need to require the caller to take care of this?

pkg/api/dataproxy_test.go
```diff
@@ -7,28 +7,55 @@ import (
...
+				"/api/datasources/proxy/1/some/thing",
+				"some/thing",
+			},
+			{
+				"/api/datasources/proxy/54/api/services/afsd%2Fafsd/operations",
+				"api/services/afsd%2Fafsd/operations",
+			},
+		}
+		for _, tc := range testCases {
+			t.Run("Given raw path should equal expected proxy path", func(t *testing.T) {
```
aknuds1, line 56: I struggle with making sense of the original description? Does the following make sense?
```suggestion
			t.Run("Given raw path, should produce expected proxy path", func(t *testing.T) {
```

zoltanbedi:
It is still not working but thank you very much for picking this up.
In here https://github.com/grafana/grafana/blob/master/pkg/api/pluginproxy/ds_proxy.go#L193
proxyPath is still good with url encoded

In here https://github.com/grafana/grafana/blob/master/pkg/api/pluginproxy/ds_proxy.go#L101
proxy.ctx.Req.URL.Path is decoded

pkg/api/dataproxy.go
```diff
@@ -33,8 +34,10 @@ func (hs *HTTPServer) ProxyDataSourceRequest(c *models.ReqContext) {
 		return
 	}
 
+	proxyPath := getProxyPath(c.Req.URL.EscapedPath())
+
 	// macaron does not include trailing slashes when resolving a wildcard path
-	proxyPath := ensureProxyPathTrailingSlash(c.Req.URL.Path, c.Params("*"))
+	proxyPath = ensureProxyPathTrailingSlash(c.Req.URL.Path, proxyPath)
```
domasx2, line 40: I think `ensureProxyPathTrailingSlash` is no longer needed? Previously `c.Params()` would remove the trailing slash, but afaik `c.Req.URL.EscapedPath()` does not

domasx2:
I tested this a bit, and seems that now the slash gets escape twice :disappointed: 

Eg if you set `req.URL.Path = "/api/prom/rules/default/%2Ftest"` you get `req.URL.String() == "https://prometheus-us-central1.grafana.net/api/prom/rules/default/%252Ftest"`

marefr:
I'll have a look and see if I can resolve these problems again

marefr:
@zoltanbedi @domasx2 can you please help out testing this out and/or provide some way to easily verify my changes. I tried to use the docker/blocks/slow_proxy, but with my changes and trying to include escaped slash doesn't arrive at the slow proxy :confused:

marefr:
I think I might be able to verify using the jaeger docker block...

marefr:
@zoltanbedi @domasx2 Using jaeger with a service named `grafana/service` I can list the operations so it seems to work as expected now:
image
image

zoltanbedi (approved):
Thank you @marefr lgtm.

domasx2:
Works for me as well, thanks @marefr  :+1:

pkg/api/dataproxy.go
marefr: Good catch :+1:

pkg/api/dataproxy.go
marefr: @domasx2 pointed out that ensureProxyPathTrailingSlash is no longer needed. Please have a look again

domasx2 (approved):

PR #30605, opened 2021-01-25
Chore: bump redux toolkit to 1.5.0

jackw:
A prototype pollution vulnerability exists in [immer < 8.0.1](https://github.com/advisories/GHSA-9qmh-276g-x5pj) which is bundled in Grafana through reduxjs/toolkit. Bumping to the [latest version](https://github.com/reduxjs/redux-toolkit/releases/tag/v1.5.0) of reduxjs/toolkit resolves this. 

Due to my limited knowledge of the codebase please consider the following quote from the release notes of RTK when reviewing:

>Immer recently released v8.0.0, which changes its default behavior around auto-freezing state values. Previously, Immer only auto-froze data in development mode, partly under the assumption that freezing would be slower overall. Due to internal changes in Immer 7, Immer now checks if data is frozen and can bail out of some of its processing early. As a result, Immer 8 switches to always freezing return values, even in production, Per discussion in the Immer issues linked from the v8 release announcement, this apparently is actually faster on average.
>
>This is a noticeable change in behavior, but we consider it not breaking for RTK on the grounds that you shouldn't be mutating state outside of a reducer anyway, so there shouldn't be any visible effect on correctly-written Redux logic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -200,7 +200,7 @@
   "dependencies": {
     "@grafana/slate-react": "0.22.9-grafana",
     "@popperjs/core": "2.5.4",
-    "@reduxjs/toolkit": "1.3.4",
+    "@reduxjs/toolkit": "1.5.0",
     "@sentry/browser": "5.25.0",
     "@sentry/types": "5.24.2",
     "@sentry/utils": "5.24.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4982,12 +4982,12 @@
     prop-types "^15.6.1"
     react-lifecycles-compat "^3.0.4"
 
-"@reduxjs/toolkit@1.3.4":
-  version "1.3.4"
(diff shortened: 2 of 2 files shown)


torkelo (approved):
LGTM

hugohaggmark (approved):
Great work Jack, any broken Redux would show up in e2e tests, especially for template variables. 

But maybe @grafana/observability-squad can have a look too as Explore uses Redux a lot too?

dprokop (approved):

zoltanbedi (approved):
Explore looks good.

aocenas:
@jackw just fyi we opt out of the freezing in explore, because it created issues with the Graph component which modified the data in some way. @zoltanbedi at some point we may actually remove this exception if we use the new Graph component and it won't need data to be modifiable.

PR #30609, opened 2021-01-25
Search: Add sort information in dashboard results

AgnesToulet:
This PR allows the sort filters to have a select clause to retrieve sort information and add it to the search results in the `SortMeta` field (along with the `MetaName` that the sort option can specify). This will be used for Enterprise sort filters.
It also adds the possibility to add an index to a sort option to specify its order in the list.

This PR also includes a bit of refactoring of the field names.

**Examples**:
Search response example:
```json
[{
    "id": 1,
    "uid": "XXXX",
    "title": "Grafana Demo",
    "uri": "db/grafana-demo",
    "url": "/d/XXXX/grafana-demo",
    "slug": "",
    "type": "dash-db",
    "tags": [
	    "prometheus"
    ],
    "isStarred": false,
    "sortMeta": "69 errors"
}, ...]
```

Sorting options:
```json
{"sortOptions":
    [{
        "description":"Sort results based on recent errors in a descending order",
        "displayName":"Errors 30 days (most)",
        "meta":"errors",
        "name":"errors-recently-desc"
    },
    ...]
}
```


pkg/services/sqlstore/dashboard.go
```diff
@@ -307,27 +311,31 @@ func makeQueryResult(query *search.FindPersistedDashboardsQuery, res []Dashboard
...
 				Tags:        []string{},
 			}
 
-			if item.FolderId > 0 {
-				hit.FolderUrl = models.GetFolderUrl(item.FolderUid, item.FolderSlug)
+			if item.FolderID > 0 {
+				hit.FolderURL = models.GetFolderUrl(item.FolderUID, item.FolderSlug)
+			}
+
+			if len(query.Sort.MetaName) > 0 {
```
sakjur, line 333: This is the nitpickiest of nitpicks, but for strings specifically I like comparing to the empty string as it tells the reader what type to expect there without having to use a lookup for it.

```suggestion
			if query.Sort.MetaName != "" {
```

pkg/services/search/sorting.go
```diff
@@ -29,6 +31,8 @@ type SortOption struct {
 	Name        string
 	DisplayName string
 	Description string
+	Index       int
+	MetaName    string
```
sakjur, line 35: One problem here, and I mean this in the "we'll probably see this as a request in a few months" not as "we should fix this before merging", is that we don't support plural and singular differentiation here.

The way I'd prefer to solve this is (_not in this PR_) to have a type specifically for this kind of value+text combinations which would make future internationalization and localization efforts easier 🙂

Naïvely I'm thinking something that could start out as

```go
type IntUnit struct {
    Unit string
    // String function uses Unit if SingularUnit is empty.
    SingularUnit string
    Value int64
}

type (u IntUnit) String() string {
    if u.Value == 1 && u.SingularUnit != "" {
        return fmt.Sprintf("%d %s", u.Value, u.SingularUnit)
    }
    return fmt.Sprintf("%d %s", u.Value, u.Unit)
}
``` 

This is implemented in a sufficient number of places that I guess it should take ten minutes or less to find a blog post outlining a much better solution. (even before accounting for internationalization)

sakjur (approved):
:+1: Neat solution!

Ping @lukasztyrala who have been working on exploration for [2021-01 New Search for 8.0 (internal)](https://docs.google.com/document/d/1UnGCYieH61GuwTxCYJXX_AyHk8YAb2qBofDN9afjtGc/view) to make sure he's aware of this being added and can object if there's a reason to do so.

PR #30628, opened 2021-01-26
Graph: Fixes auto decimals issue in legend and tooltip

### LINKED ISSUE #30619: Decimals: Graph tooltips are less precise with decimals = auto
**What happened**:
I tried out how decimals are displayed in 7.4.0-beta1 and had rather surprising results: https://github.com/grafana/grafana/pull/30262#issuecomment-764597588
With https://github.com/grafana/grafana/pull/30519, the situation has improved a lot and it's pretty good now (almost perfect!), but the Graph panel now shows numbers in the tooltip without any decimals when decimals are set to *auto*.

With nightly build Grafana v7.5.0-10813pre (80294b2dbf):
grafana_30519

With Grafana 7.3.7:
grafana_737

The tooltip formats the number exactly like the Stats panel, i.e. 4.57 GB in the first panel, 457 GB in the second and 45.7 GB in the third. The axis uses "appropriate" decimals, in this case no decimals are totally sensible.

Apart from that, I would generally expect that the fixed number of decimals also applies to the tooltips - but Graph always displays one decimal more than configured. It's not very important, I just think it's strange.

Note that the new Time Series panel is working fine in all of these things.

Create a Graph panel. Use Static datasource to have a value like 4,567,890,123. Set unit to *bytes(SI)* and decimals to *auto*.

**Environment**:
- Grafana version: nightly build v7.5.0-10813pre (80294b2dbf)
- Data source type & version: Static
- User OS & Browser: Windows / Chrome
- Grafana plugins: nothing relevant
- Others:

torkelo:
Fixes #30619

When fixing the decimals issue (and removing scaled decimals) I forgot this part in TimeSeries2 that is used by legend & tooltip

### DIFF
--- a/public/app/core/specs/time_series.test.ts
+++ b/public/app/core/specs/time_series.test.ts
@@ -430,25 +430,11 @@ describe('TimeSeries', () => {
       };
     });
 
-    it('should set decimals based on Y axis (expect calculated decimals = 1)', () => {
+    it('should set decimals to null if no decimals set', () => {
       const data = [series];
       // Expect ticks with this data will have decimals = 1
       updateLegendValues(data, panel, height);
-      expect(data[0].decimals).toBe(2);
-    });
-
-    it('should set decimals based on Y axis to 0 if calculated decimals = 0)', () => {
-      testData.datapoints = [
-        [10, 2],
-        [0, 3],
-        [100, 4],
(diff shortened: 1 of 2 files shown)


hugohaggmark (approved):
LGTM!

PR #30638, opened 2021-01-26
Grafana/UI: Unit picker should not set a category as unit

### LINKED ISSUE #30599: Unit picker: clicking on the units category sets the category as a unit
Please see attached video:

https://user-images.githubusercontent.com/2376619/105712629-5bde0f00-5f1a-11eb-9cb1-438986932567.mp4

**How to replicate:**

1. Go to i.e. graph panel Edit mode
2. Select Field options
3. Open standard options 
4. Click on the unit picker
5. Click on any of the unit categories (i.e. Acceleration)
6. Bum: after step 5 the unit is set to Acceleration.

oscarkilhed:
When clicking categories for units, the category is set as the unit, until a leaf node is selected.
This change exposes the changeOnSelect property from rc-cascader, and sets it to false for the unit picker.

Fixes #30599

### DIFF
--- a/packages/grafana-ui/src/components/Cascader/Cascader.tsx
+++ b/packages/grafana-ui/src/components/Cascader/Cascader.tsx
@@ -13,6 +13,8 @@ interface CascaderProps {
   separator?: string;
   placeholder?: string;
   options: CascaderOption[];
+  /** Changes the value for every selection, including branch nodes */
+  changeOnSelect?: boolean;
   onSelect(val: string): void;
   /** Sets the width to a multiple of 8px. Should only be used with inline forms. Setting width of the container is preferred in other cases.*/
   width?: number;
@@ -68,6 +70,8 @@ export class Cascader extends React.PureComponent<CascaderProps, CascaderState> 
       activeLabel,
     };
   }
+
+  static defaultProps = { changeOnSelect: true };
 
   flattenOptions = (options: CascaderOption[], optionPath: CascaderOption[] = []) => {
     let selectOptions: Array<SelectableValue<string[]>> = [];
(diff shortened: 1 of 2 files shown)


packages/grafana-ui/src/components/Cascader/Cascader.tsx
```diff
@@ -13,6 +13,8 @@ interface CascaderProps {
   separator?: string;
   placeholder?: string;
   options: CascaderOption[];
+  /** Changes the value for every selection, including branch nodes */
```
hugohaggmark, line 16: Super small nit, maybe worth mentioning default value?
```suggestion
  /** Changes the value for every selection, including branch nodes. Defaults to true. */
```

packages/grafana-ui/src/components/Cascader/Cascader.tsx
```diff
@@ -195,7 +199,7 @@ export class Cascader extends React.PureComponent<CascaderProps, CascaderState>
           <RCCascader
             onChange={onChangeCascader(this.onChange)}
             options={this.props.options}
-            changeOnSelect
+            changeOnSelect={changeOnSelect}
```
oscarkilhed, line 202: This property name in the rc-cascade component is a bit weird. The documentation doesn't do a good job of describing what it does either.

hugohaggmark (approved):
Looks good 🎉  , I just left a super small nit for consideration.

packages/grafana-ui/src/components/Cascader/Cascader.tsx
oscarkilhed: Absolutely!

PR #30644, opened 2021-01-26
bump cypress to 6.3.0

vickyyyyyyy:

@grafana/e2e is currently using a much older version of cypress - we want to bump this up to their latest version.

In particular, `<5.2.0` versions of cypress seem to bring up a templating error when adding a template variable in our variable editor. Detailed investigation notes [here](https://github.com/grafana/plugins-private/issues/1582). We want this fixed so we can enable our e2e tests for our enterprise plugins which currently try to run tests in our variable editor.

There seems to be a few breaking changes in [v6](https://docs.cypress.io/guides/references/changelog.html#6-0-0) which affects the current implementation of our e2e tests:

- We cannot use `not.be.visible` because they will now fail on elements which do not exist in the DOM. For our elements where they are asserting on elements that do not exist in the DOM, we need to use `not.exist` instead.

- We can no longer use `be.visible` on some of our elements because they will now fail on elements with `opacity:0`. From the changelog, it states although these assertions can no longer be used, they can still be actionable (e.g. we can still use `.click()` etc) so I have just removed the `be.visible` assertions from the chain and have applied the suggestions of replacing with `exist` instead.


vickyyyyyyy:
The E2E tests passed locally for me so I will have a look at why they are failing here 😭

hugohaggmark:
> The E2E tests passed locally for me so I will have a look at why they are failing here 😭

I restarted the build and the same e2e tests failed so I think this might be real failing tests and not flakiness. Let me know if you need help with this @vickyyyyyyy

e2e/suite1/specs/select-focus.spec.ts
```diff
@@ -17,15 +17,15 @@ e2e.scenario({
 
         e2e.components.Select.option().should('be.visible').first().click();
 
-        e2e.components.Select.input().should('be.visible').should('have.focus');
+        e2e.components.Select.input().should('have.focus');
```
hugohaggmark, line 20: Maybe we should add back the exist?
```suggestion
        e2e.components.Select.input().should('exist').should('have.focus');
```

e2e/suite1/specs/select-focus.spec.ts
```diff
@@ -17,15 +17,15 @@ e2e.scenario({
...
       });
 
     e2e.pages.Dashboard.Settings.General.title().click();
 
     e2e.components.FolderPicker.container()
       .should('be.visible')
       .within(() => {
-        e2e.components.Select.input().should('be.visible').should('not.have.focus');
+        e2e.components.Select.input().should('not.have.focus');
```
hugohaggmark, line 28: Maybe we should add back the exist?
```suggestion
        e2e.components.Select.input().should('exist').should('not.have.focus');
```

e2e/suite1/specs/inspect-drawer.spec.ts
```diff
@@ -75,36 +75,36 @@ const expectDrawerTabsAndContent = () => {
         expect(li.text()).equals('Data');
       });
       e2e.components.PanelInspector.Data.content().should('be.visible');
-      e2e.components.PanelInspector.Stats.content().should('not.be.visible');
-      e2e.components.PanelInspector.Json.content().should('not.be.visible');
-      e2e.components.PanelInspector.Query.content().should('not.be.visible');
+      e2e.components.PanelInspector.Stats.content().should('not.exist');
```
hugohaggmark, line 78: Is there any documentation from Cypress that mentions this change?

e2e/suite1/specs/select-focus.spec.ts
vickyyyyyyy: Updated the PR description for why this needs to be removed for v6 of cypress 😄

e2e/suite1/specs/inspect-drawer.spec.ts
vickyyyyyyy: I was in the middle of updating the PR after pushing my changes, just finished updating it 👍

e2e/suite1/specs/variables/new-query-variable.ts
```diff
@@ -105,7 +105,11 @@ describe('Variables - Add variable', () => {
 
     e2e.pages.Dashboard.SubMenu.submenuItemLabels('a label').should('be.visible');
 
-    e2e.pages.Dashboard.SubMenu.submenuItemValueDropDownValueLinkTexts('A').eq(1).should('be.visible').click();
+    e2e.pages.Dashboard.SubMenu.submenuItemValueDropDownValueLinkTexts('A')
+      .should('be.visible')
```
vickyyyyyyy, line 109: @hugohaggmark this test seemed to be a bit flaky - it passed before I fixed the other tests with the v6 breaking changes but then failed after, even though I didn't change anything (with this). When running locally (both headless and not), I could never get this to fail like it did in the build but thought I'd add in this additional assertion, just in case. Do you reckon it's fine to do so?

hugohaggmark: Yes, this one is unfortunately a bit flaky. Can you try and add 500ms to 
https://github.com/grafana/grafana/blob/027944dfd6db6142e9415e4e5c7b135a7a368be2/e2e/suite1/specs/variables/new-query-variable.ts#L102
and
https://github.com/grafana/grafana/blob/027944dfd6db6142e9415e4e5c7b135a7a368be2/e2e/suite1/specs/variables/new-query-variable.ts#L170

e2e/suite1/specs/inspect-drawer.spec.ts
hugohaggmark: Awesome, thank you 🎉

e2e/suite1/specs/variables/new-query-variable.ts
vickyyyyyyy: If we use specific `wait`s before `should` assertions, are we pretty much prolonging the `should` assertion timeout? Are the `should`s not enough to avoid flakiness?

hugohaggmark: I wasn't aware you could change the individual timeouts, if we can change the timeout on those that would be a lot better. Thank you.

vickyyyyyyy: I'm reading up on Cypress (I've not used it before, so bear with me 🙏 ) to have a better understanding of our timeouts and correct me if I'm wrong, but it looks like we have the following options:

- Increase the timeouts for single assertions. In this case, increase the timeout for the `should` assertions after the `wait`s suggested: https://docs.cypress.io/api/commands/should.html#Timeouts

- Increase the timeouts for all assertions for all tests by changing the `defaultCommandTimeout` in the config: https://docs.cypress.io/guides/references/configuration.html#Timeouts

- Increase the timeouts for all assertions for just one test: https://stackoverflow.com/questions/55019061/increase-timeout-for-specific-its-method-of-cypress-io

Which do you think we should implement? Is there a reason why we wouldn't want to increase the default timeout to prevent flakiness across all tests?

hugohaggmark: Yes, we need to keep everything as fast as possible, just adding those waits kind of made me very sad. We should make exceptions at micro level if you ask me, so I'd increase timeout for the individual assertions that fail here and remove the waits.

vickyyyyyyy: Awesome, thanks! I will increase the timeout in the individual assertions to 4500 (for the default 4000 + the 500 suggested)

hugohaggmark: With the previous waits it should be 5000, correct? Or are you keeping the waits. Sorry for the confusion.

vickyyyyyyy: I was thinking instead of the `wait`s we can pass in the individual timeouts and if the default command timeout for cypress is 4000 and we wanted to have a wait of 500, we would be able to add in a timeout of 4500 for both `should` assertions that come after?

```
// instead of this e2e().wait(500);
// the should assertion has default of 4000 but we want to add 500 to make it 4500 below
e2e.components.BackButton.backArrow().should('be.visible').click({ force: true });
```

Let me know if I've misunderstood?

I also had a look at passing in the individual timeouts and it looks like we would need to update the wrappers here: https://github.com/grafana/grafana/blob/master/packages/grafana-e2e/src/support/types.ts#L60 with

```
return e2e().get(Selector.fromAriaLabel(value), { timeout });
return e2e().get(selector, { timeout })
return e2e().get(Selector.fromAriaLabel(selector, { timeout }));
```

Just to double check, is my understanding correct? If that's the case then I can add the `wait`s as suggested here first to help with the flakiness and then I can create a separate PR for allowing individual timeouts to be passed into our wrappers? Just to keep the PR smaller and I don't want to block other work that's waiting on this PR 😄

vickyyyyyyy: Oh, I misread your previous comment and thought you meant to add the `wait`s but didn't realise they were already there 🤦‍♀️  Yup, you are right, it would be 5000 😄

vickyyyyyyy: Increased the wait and it still seems flaky 😔

hugohaggmark: Regarding the order of PRs, I totally agree with you, that sounds like a great approach. Regarding the flakiness with the increase, try another 500ms if that doesn't work then there is probably something else wrong and I can help look into it too.

vickyyyyyyy: Awesome, thanks!

I've increased it and it looks like it still fails at the exact same place:

```
1143 | 1) Variables - Add variable
1144 | adding a multi value query variable:
1145 | AssertionError: Timed out retrying after 4000ms: Expected to find element: `1`, but never found it. Queried from element: <a.css-10l6kcd>
1146 | at Context.eval (http://end-to-end-tests-server:3001/__cypress/tests?p=../../e2e/suite1/specs/variables/new-query-variable.ts:1855:87)
```

I can't reproduce it locally but I haven't merged latest master in yet, so I'll do that and see if I can get it failing

vickyyyyyyy: Merged with latest master locally to test and I still can't reproduce locally:

Screenshot 2021-01-29 at 09 55 14

The error in the build never shows up and it's fine with the `eq(1)` selector:

Screenshot 2021-01-29 at 09 57 24

Any suggestions on how I can debug this further? In the e2e tests I've been debugging for our enterprise plugins, I was always able to see the same errors locally as the build. Is there anything the build does differently here, in particular? 🤔

hugohaggmark: Sorry for the late response, I've been in meetings. No, this flakiness is hard to repro locally. I'm taking it for a spin now.

vickyyyyyyy: Thank you! I tried removing the waits locally to see if I can make it fail but they still passed 😓 

Is this test failing more for this PR in particular or does it tend to fail a lot in other PRs too? I opened another PR this morning and the build for that was fine - I was wondering if maybe v6 of cypress makes it more flaky for some reason?

hugohaggmark:
Pushed some changes now, let's see if those help :)

hugohaggmark:
If the changes don't work, feel free to add `skip` to the flaky tests and I'll have another look on Monday. This way you can close this PR. Does that make sense?

hugohaggmark (approved):
Great work on bumping this, it was way overdue. Thank you!!

vickyyyyyyy:
> If the changes don't work, feel free to add `skip` to the flaky tests and I'll have another look on Monday. This way you can close this PR. Does that make sense?

Makes sense, thanks so much! The test is still failing so I've skipped them for now 👍

hugohaggmark:
I pushed a small refactor that might help out.

vickyyyyyyy:
> I pushed a small refactor that might help out.

Aha, I also pushed the `skip` addition, let me revert that and see 👍

vickyyyyyyy:
> I pushed a small refactor that might help out.

It looks like that worked! Thanks so much for your help!! 🔥

PR #30653, opened 2021-01-26
Frontend changes for library panels feature

kaydelaney:
Just a draft PR to show some of the work I've been doing for the frontend side of the library panels feature.
There are quite a few changes here so I may very well break out some of the changes into independent PRs.
In its current state this PR enables users to add, delete, update and unlink library panels, though I can almost guarantee that there are some bugs which I still need to iron out. There are also still a lot of things I need to refactor and consolidate, but it's a starting point.


packages/grafana-ui/src/components/NewTable/NewTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+  tableClass?: string;
+}
+
+enum SortDirection {
+  Ascending = 'ascending',
+  Descending = 'descending',
+  None = 'none',
+}
+
+export const NewTable: FC<Props> = (props: Props) => {
```
kaydelaney, line 26: this is still heavily a WIP and will be split into a separate PR

hugohaggmark (commented):
Massive work 🎉  going for a test drive now. 

The only thing that struck me while browsing the code is the absence of the featureToggle `panelLibrary` :)

Will do a proper review after test drive. Again, monumental work on this so far!

hugohaggmark:
Took this for a testdrive and I can add existing lib panels, splendid work on that!  I'm struggling to save the dashboard after adding library panels, haven't figured out what is wrong yet but I'll continue to look into it.

And I've looked into the PR some more and I've identified the following fixes to the api. Let me know if you agree I should fix these @kaydelaney:

1. Introduce a proper frontend DTO with correct casing so we don't need to translate for instance, `libraryPanel.ID => librayPanel.id` in the frontend. 
2. Return user names instead of ids in the api response, this way we don't need the extra call to `/api/org/users`
3. `getLibraryPanels` should only return library panels that the user has permissions for

We should also discuss if saving and updating the panels should be done in the frontend or backend. Not sure what the best approach is right now.

hugohaggmark:
I think the next prio steps for this PR is:

- [ ] get everything behind a feature toggle
- [ ] _maybe split into several PRs (the NewTable component is something I think should be in a separate PR)_
- [ ] fix build errors
- [ ] merge to master

hugohaggmark:
I pushed some changes that fixes some of issue when saving a dashboard with library panels

dprokop:
Wow, loads to take a look at :) Will do that during the day today

packages/grafana-data/src/types/theme.ts
```diff
@@ -243,6 +243,9 @@ export interface GrafanaTheme extends GrafanaThemeCommons {
     formCheckboxBgChecked: string;
     formCheckboxBgCheckedHover: string;
     formCheckboxCheckmark: string;
+
+    cardBackground: string;
+    cardBackgroundHover: string;
```
torkelo, line 248: use bg2  (it is the card barckground), and then styleMixins.hoverColor(bg2, theme)

torkelo: Or if you can use the Card component

torkelo (commented):
Great job! Good feedback from Hugo on structural changes. 

Not sure about the NewTable component, would be good to list out the requirements and how we can solve this with the main table component and the pros and cons of a new separate component, or if we can extend/modify the core Table component to cover this use case as well. (Good topic for a separate PR) 

In the future try to open draft PRs with lists of Todos and things missing / left to figure out, draft PRs can be ugly & incomplete but gives the squad and tech leads a better sense of where things stand and an opportunity to gauge progress & jump in with early feedback if needed

packages/grafana-ui/src/components/Icon/Icon.tsx
```diff
@@ -77,6 +77,7 @@ export const Icon = React.forwardRef<HTMLDivElement, IconProps>(
             style={style}
           />
         )}
+        {divElementProps.children}
```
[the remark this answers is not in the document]
kaydelaney: Ah, makes sense

kaydelaney:
Re: changes to the api @hugohaggmark 
>     1. Introduce a proper frontend DTO with correct casing so we don't need to translate for instance, `libraryPanel.ID => libraryPanel.id` in the frontend.

👍 

>     2. Return user names instead of ids in the api response, this way we don't need the extra call to `/api/org/users`
Would also be nice to return the url of the user's avatar if we want to avoid additional api calls.

>     3. `getLibraryPanels` should only return library panels that the user has permissions for

👍 

> I pushed some changes that fixes some of issue when saving a dashboard with library panels

Thanks!

Working on getting everything behind a feature toggle.

public/app/features/dashboard/components/AddPanelWidget/AddPanelWidget.tsx
```diff
@@ -197,12 +235,51 @@ const getStyles = stylesFactory((theme: GrafanaTheme) => {
...
+      margin-top: 39px;
+      margin-bottom: 42px;
+    `,
+    buttonMargin: css`
+      margin-right: 10px;
+    `,
+    libraryPanelsWrapper: css`
+      padding-left: 32px;
+      padding-right: 25px;
+      padding-top: 40px;
```
dprokop, line 282: Can we use theme spacings for these fixed values?

public/app/features/dashboard/components/AddPanelWidget/AddPanelWidget.tsx
```diff
@@ -228,20 +305,20 @@ const getAddPanelWigetHandleStyles = stylesFactory((theme: GrafanaTheme) => {
...
+//       margin-bottom: ${theme.spacing.lg};
+//       &:hover {
+//         background: ${theme.colors.bg2};
+//       }
+//     `,
+//     icon: css`
+//       color: ${theme.colors.textWeak};
+//     `,
+//   };
+// });
```
dprokop, line 324: Don't forget to remove:)

public/app/features/dashboard/components/LibraryPanelCard/LibraryPanelCard.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [showDeletionModal, setShowDeletionModal] = useState(false);
+
+  const onDeletePanel = () => {
+    onDelete?.();
+    setShowDeletionModal(false);
+  };
+
+  return (
+    <>
+      <div className={cx(styles.panelCard)}>
```
dprokop, line 41: Please use cx only when composing class names:)

public/app/features/dashboard/components/LibraryPanelCard/LibraryPanelCard.tsx
```diff
@@ -0,0 +1,162 @@
...
+              <span>{title}</span>
+              <div className={cx(styles.panelFigures)} onClick={onClick}>
+                <span>Reusable panel</span>
+                <Icon name="apps" className={cx(styles.detailIcon)} />
+                {connectedDashboards.length}
+                <Icon name="x" className={cx(styles.detailIcon)} />
+                {varCount}
+                <span>
+                  Last edited {lastEdited} by {lastAuthor}
+                </span>
```
dprokop, line 55: image

From a user perspective would be nice to see what each icon/counter represents. Maybe in a tooltip?

public/app/features/dashboard/components/LibraryPanelCard/LibraryPanelCard.tsx
```diff
@@ -0,0 +1,162 @@
...
+      display: flex;
+      padding: 10px;
+    `,
+    cardControls: css`
+      margin-top: 11px;
+      margin-bottom: 11px;
+      margin-left: 16px;
+      margin-right: 16px;
+      padding-top: 11px;
+      padding-left: 22px;
```
dprokop, line 118: Please use theme spacing variables:)

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -26,6 +50,9 @@ export const PanelOptionsTab: FC<Props> = ({
   onPanelConfigChange,
   onPanelOptionsChanged,
 }) => {
+  const theme = useTheme();
+  const styles = getStyles(theme);
```
dprokop, line 54: You can use `useStyles` instead when styles getter depends on  theme only (https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md#usestyles-hook)

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -26,6 +50,9 @@ export const PanelOptionsTab: FC<Props> = ({
...
+  const [showingAddPanelModal, setShowingAddPanelModal] = useState<boolean>(false);
```
dprokop, line 55: Think stype annotation is not neccesary when initial value is set

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -46,7 +73,50 @@ export const PanelOptionsTab: FC<Props> = ({
...
+  };
+
+  const onAddToPanelLibrary = (e: React.MouseEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+
+    setShowingAddPanelModal(true);
+  };
+
+  if (panel.libraryPanel) {
```
dprokop, line 101: and feature toggle?:)

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -152,5 +222,56 @@ export const PanelOptionsTab: FC<Props> = ({
     </OptionsGroup>
   );
 
+  if (config.featureToggles.panelLibrary) {
+    elements.push(
```
dprokop, line 226: image

I think going forward we could optimize the card's display depending on the available space in the options pane. Maybe show less information when pane is not wide enough.

I'm also not sure about the information being shown on the card (at least when adding a new panel to a dashboard). I think the information bout how many dashboards the panel is used is probably relevant when the user wants to remove the panel. In other cases, it feels like it may be adding unnecessary noise. WDYT @jessover9000?

dprokop (commented):
Plenty of effort here @kaydelaney, awesome work!

public/app/core/services/backend_srv.ts
```diff
@@ -397,6 +410,40 @@ export class BackendSrv implements BackendService {
...
+    return this.patch(`/api/library-panels/${panelSaveModel.libraryPanel.uid}`, {
+      folderId,
+      name: panelSaveModel.title,
+      model: panelSaveModel,
+    }).then(({ result }: { result: LibraryPanel }) => result);
+  }
+
+  async deleteLibraryPanel(uid: string) {
+    return this.delete(`/api/library-panels/${uid}`);
+  }
```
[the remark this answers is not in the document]
hugohaggmark: I agree @dprokop I already asked @kaydelaney to move all feature specific things from backend_srv into a separate class

public/app/features/dashboard/components/LibraryPanelCard/LibraryPanelCard.tsx
torkelo: We should use the Card component here (and extend it where needed)

public/app/features/dashboard/components/SaveDashboard/useDashboardSave.tsx
```diff
@@ -8,12 +8,63 @@ import appEvents from 'app/core/app_events';
...
 const saveDashboard = async (saveModel: any, options: SaveDashboardOptions, dashboard: DashboardModel) => {
   let folderId = options.folderId;
   if (folderId === undefined) {
-    folderId = dashboard.meta.folderId || saveModel.folderId;
+    folderId = dashboard.meta.folderId ?? saveModel.folderId;
   }
+
+  if (config.featureToggles.panelLibrary) {
+    await saveLibraryPanels(saveModel, folderId!);
+  }
```
torkelo, line 66: this is the wrong place, saving a library panel should be a specific action done from the panel edit mode. As is only for one panel at a time

public/app/features/library-panels/components/LibraryPanelCard/LibraryPanelCard.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [showDeletionModal, setShowDeletionModal] = useState(false);
+
+  const onDeletePanel = () => {
+    onDelete?.();
+    setShowDeletionModal(false);
+  };
+
+  return (
+    <>
+      <div className={cx(styles.panelCard)}>
```
torkelo, line 41: I do no think we should duplicate and create yet another card component, but try to use the core Card component and extend it to handle this scenario.

torkelo (commented):
A nice job! it's a good start :) 

I could not add a library panel, the button in the modal does not work for me. Not sure if it's hooked up yet. 

Found some issues (Card styles, we want to try to reuse and extend the core Card component here), and updating a library panel should be an explicit action from the panel edit mode, not linked to saving a dashboard.

packages/grafana-ui/src/components/NewTable/NewTable.tsx
torkelo: Think finding a way to re-use the core Table component would be preferable, what was the main reason for not using it?

public/app/features/dashboard/components/SaveDashboard/useDashboardSave.tsx
kaydelaney: Agree. This was always just a temporary measure

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -46,7 +73,50 @@ export const PanelOptionsTab: FC<Props> = ({
       visTabInputRef.current.focus();
     }
   };
-  // Fist common panel settings Title, description
+
+  const useLibraryPanel = ({ model, uid, title, lastEdited, lastAuthor }: LibraryPanelCardProps) => {
```
torkelo, line 77: Would be nice to move the library panel stuff out to it's own component thís file & component is already quite big

public/app/features/library-panels/components/LibraryPanelCard/LibraryPanelCard.tsx
kaydelaney: Makes sense. Will work on consolidating the components

packages/grafana-ui/src/components/NewTable/NewTable.tsx
kaydelaney: Main reason was that the existing Table component uses a `DataFrame` as the interface for displaying data, which makes sense for datasource responses but seemed like kind of a strange fit for simpler data.

public/app/features/dashboard/components/SaveDashboard/useDashboardSave.tsx
kaydelaney: Just to clarify something: So the only way to save a library panel should be from the panel edit view?
If that's the case, if a user creates a new library panel - which automatically gets added to the current dashboard - and the user saves the dashboard before that panel is explicitly saved, should a modal be presented warning the user that they have unsaved library panels or something to that effect?

packages/grafana-ui/src/components/NewTable/NewTable.tsx
torkelo: But creating a DataFrame with say toDataFrame is pretty simple. This feels pretty complex as well. Maybe a wrapper around the current Table to make it easier to use for non DataFrame could be an option?

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
torkelo: yea, agree, would try to keep this card as simple as possible to start with (star, delete, copy). 

What does the `(x) 3` mean

kaydelaney: Per jess's design, it means the number of template variables used by the panel. Currently just a placeholder though

torkelo: ah, ok. Let's skip that for now

hugohaggmark:
Took this for a test spin again and most things are working nicely together. One thing I found was that the Add panel to library button doesn't seem to be hooked up to something 
image

public/app/core/services/backend_srv.ts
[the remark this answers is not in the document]
kaydelaney: https://github.com/grafana/grafana/pull/30993

jessover9000:
Hello! Managed to look at the feature now :)

**Panel library dropdown in panel settings sidepane**
- Search input should be big enough to show the entire placeholder text (could use entire width up to the filter dropdown), if it's too small the placeholder text should get shortened to just "Search library" instead of being cut off
- Spacing between search input and "Popular panels from the library" could be bigger. Maybe give the search input a margin-bottom:8px.
- "No library panels found" message has the same text style as the headline saying Popular panels. Instead the "No panels found should look different – maybe make it grey or add a headline style to the "Popular panels" line? This also applies to the Add panel > Panel library view.
image

**Panel library Card list**
- Icon div is 32px wide, but icon is only 16px. If we want to use a small 16x16 icon for the library panels, it's fine to move the title, button etc further to the left. Right now the space between icon and rest of the card content is a bit large.
- The position of the tag is a bit high, therefore although the all-around padding is the same it looks as if the top padding was bigger because the panel title is a bit lower than the tag.
image

**Add this panel to library modal**
Generally this looks good. My local Grafana isn't set up with lots of folders, so I'm wondering how that dropdown will work if you have more nested folders? Will it become a cascader? Also, when I click the dropdown, I get a blinking cursor but can't seem to edit the text in the dropdown / doesn't seem searchable in this case, but I think a typing/search option might be helpful for people with excessive folder structures.
image

**Add new panel > Add reusable panel**
If we don't have any reusable panels added yet, does it make sense to show the panel library at all? Or should we move directly to the reusable panel editor with a default new panel?
image

**Things that I noticed that maybe just aren't done yet**
- Panel edit for reusable panels doesn't show that you're creating a reusable panel in the title as of now. I think it should have the subline that says "Reusable panel / Library panel" or something like in the designs, otherwise it might be easy to miss that you're working on a "special" panel and not just a random test one.
- Save/Update buttons for reusable panel edit still need to be changed.
- If I see it correctly, the reusable panels currently don't appear in the dashboard list as of now? Because the modals tell me they are saved in a folder but I cannot see the actual reusable panels if I go to dashboards>manage.

hugohaggmark:
@kaydelaney I took this for another spin this morning and I think we're really close a MVP, fantastic work on this!

kaydelaney:
Just a note on that last commit: I said as much in a comment, but I suspect that handling of library panel disconnections would be better suited to the backend. Though if someone disagrees I'd love to hear it.

torkelo:
Good progress! 

Some points (that can be done in follow-up PRs)
* No Apply button for a library panel (should only be an option to Save or Discard, Esc should open modal so user can choose to discard or save). So no way to leave library panel edit without saving or discarding changes

Add panel view:
* The padding is not aligned (same left and right)
Screenshot from 2021-02-12 15-28-50

The library panel view:

* The heading "Popular panels from the panel library"  should be above the filter & search inputs 
* The `Create new reusable panel`  is something we can wait with I feel to follow-up PR as that requires special logic for discard (needs to delete panel) 

Screenshot from 2021-02-12 15-23-40

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -149,6 +149,16 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
   onFieldConfigChange = (config: FieldConfigSource) => {
+    if (this.props.panel.libraryPanel?.uid !== undefined) {
+      this.props.dashboard.panels.forEach((panel) => {
+        if (panel.libraryPanel?.uid === this.props.panel.libraryPanel!.uid) {
+          panel.updateFieldConfig({
+            ...config,
+          });
+        }
+      });
+    }
```
torkelo, line 160: not sure I understand this condition and what is going on here

torkelo: the panel edit cannot update other panels in the dashboard (would break discard mechanic)

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -157,12 +167,30 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
   };
 
   onPanelOptionsChanged = (options: any) => {
+    if (this.props.panel.libraryPanel?.uid !== undefined) {
+      this.props.dashboard.panels.forEach((panel) => {
+        if (panel.libraryPanel?.uid === this.props.panel.libraryPanel!.uid) {
+          panel.updateOptions(options);
+        }
+      });
+    }
```
torkelo, line 176: same as above, need to remove this and handle this some other way

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -157,12 +167,30 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
   onPanelConfigChanged = (configKey: keyof PanelModel, value: any) => {
+    if (this.props.panel.libraryPanel?.uid !== undefined) {
+      this.props.dashboard.panels.forEach((panel) => {
+        if (panel.libraryPanel?.uid === this.props.panel.libraryPanel!.uid) {
+          // @ts-ignore
+          panel[configKey] = value;
+          panel.render();
+        }
+      });
+    }
```
torkelo, line 192: Same as above

public/app/features/dashboard/components/SaveDashboard/useDashboardSave.tsx
```diff
@@ -9,50 +9,19 @@ import { updateLocation } from 'app/core/reducers/location';
...
+  const libraryPanels = dashboard.panels
+    .filter((panel) => panel.libraryPanel?.uid !== undefined)
+    .reduce<Record<string, boolean>>((libPanels, curPanel) => {
+      libPanels[curPanel.libraryPanel!.uid!] = true;
+      return libPanels;
+    }, {});
 
-  return Promise.all(panelPromises).then(() => saveModel);
+  const panelsToUnlink = dashboard.originalLibraryPanels.filter((libPanelId) => !libraryPanels[libPanelId]);
+  return panelsToUnlink.map((panelId) => getLibrarySrv().disconnectLibraryPanel(panelId, dashboard.id));
```
torkelo: not sure I understand why this is, is this not an action in panel edit?

torkelo: I guess this is for deleted panels? then yes that should be done in backend not here

public/app/features/dashboard/state/DashboardModel.ts
```diff
@@ -125,6 +127,9 @@ export class DashboardModel {
     this.links = data.links || [];
     this.gnetId = data.gnetId || null;
     this.panels = _.map(data.panels || [], (panelData: any) => new PanelModel(panelData));
+    this.originalLibraryPanels = this.panels
+      .filter((panel) => panel.libraryPanel !== undefined)
+      .map((panel) => panel.libraryPanel!.uid!);
```
torkelo, line 132: think originalLibraryPanels is only used for unlink on save, so can remove this, is handled in backend (or should). so not sure we need this?

torkelo (commented):
Sorry for short and many comments, great progress and looks close to mergable. 

But found some big problems in panel edit there and the actions that modify other panel models, which will break the discard mechanic. Panel edit is working on a totally isolated panel model clone in order to easily discard all changes. So making changes to other panel models is not allowed. 

Not sure why this code exists there, is it if you have multiple panels linked to the same library panel? Think we easily handle that logic onSave/Apply when you leave panel edit

hugohaggmark:
Let me know if we need to move stuff to the backend, I'm glad to help out.

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -46,7 +67,57 @@ export const PanelOptionsTab: FC<Props> = ({
...
+              <img
+                width="22"
+                height="22"
+                className={styles.userAvatar}
+                src={panel.libraryPanel.meta.updatedBy.avatarUrl}
+                alt={`Avatar for ${panel.libraryPanel.meta.updatedBy.name}`}
+              />
+            )}
+            {panel.libraryPanel.meta.updatedBy.name}
+          </p>
```
torkelo, line 109: Would like to see this in its own component/file

public/app/features/dashboard/components/PanelEditor/PanelOptionsTab.tsx
```diff
@@ -152,5 +223,67 @@ export const PanelOptionsTab: FC<Props> = ({
...
+          <AddLibraryPanelModal
+            panel={panel}
+            onDismiss={() => setShowingAddPanelModal(false)}
+            initialFolderId={dashboard.meta.folderId}
+            isOpen={showingAddPanelModal}
+          />
+        )}
+      </OptionsGroup>
+    );
+  }
```
torkelo, line 265: Can this also be moved to its own component/file?

hugohaggmark:
I think this works really good @kaydelaney, I've added all bugs/questions from my latest run through to the continuation issue (#31305) so we can merge this PR to master.

hugohaggmark (approved):
Looks outstanding, let's get this merged! 🎉

kaydelaney:
Some notes on ensuring library panels are always either saved or discarded when exiting the editor:
Currently how this works is when the user exits the panel editor (either by pressing the (non-browser) back arrow, or pressing esc), an action `exitPanelEditor` is dispatched which compares the current panel model with the "source" panel model which is initialized along with the panel editor. If the panel models differ, the user is prompted with a modal to save the panel, or cancel. If they save, they are brought back to the dashboard, and if they cancel are brought back to the panel editor.
One case isn't handled here, which is when the user presses the browser's back button. If we want it to behave the same way as pressing the grafana back button, we would need to implement a panel change tracker similar to the dashboard change tracker, which would monitor the current path and display the save modal if the user leaves the edit panel view.
Unfortunately with our current infrastructure which makes use of angular's router, and our solution for keeping angular location state and redux location state in sync, bridgeSrv, it seems like it would be very difficult to implement this in a clean manner, and it might be a better idea to wait until we switch to react router.
An alternative would be to use the `panelEditorCleanUp` action which is dispatched when the panel editor is unmounted as an opportunity to show the modal. This would take the user away from the panel editor, but it's better than nothing.

public/app/core/services/backend_srv.ts
```diff
@@ -398,6 +398,10 @@ export class BackendSrv implements BackendService {
   getFolderByUid(uid: string) {
     return this.get<FolderDTO>(`/api/folders/${uid}`);
   }
+
+  async getOrgUsers(): Promise<OrgUser[]> {
+    return this.get('/api/org/users');
+  }
```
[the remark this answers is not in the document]
kaydelaney: nope, nice catch. will remove.

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -156,6 +181,19 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
+    appEvents.emit(CoreEvents.showModalReact, {
+      component: UnlinkModal,
+      props: {
+        onConfirm: () => {
+          delete this.props.panel.libraryPanel;
+          this.props.panel.render();
+          this.forceUpdate();
+        },
+      },
+    });
```
torkelo, line 194: would be better to use react or redux & ModalsController  to show modals, using the event is mostly for legacy angular use cases

public/app/features/dashboard/components/PanelEditor/PanelEditor.tsx
```diff
@@ -255,18 +293,48 @@ export class PanelEditorUnconnected extends PureComponent<Props> {
...
-        Save
-      </ToolbarButton>,
+      this.props.panel.libraryPanel ? (
+        <ToolbarButton
+          onClick={this.onSavePanel}
+          variant="primary"
+          title="Apply changes and save library panel"
+          key="save-panel"
+        >
+          Save Panel
```
torkelo, line 303:
```suggestion
   Save library panel
```

torkelo:
Creating a new library panel, after saving it says 0 dashboards, but the list contains a lot

Screenshot from 2021-02-22 20-03-39

hugohaggmark:
> Creating a new library panel, after saving it says 0 dashboards, but the list contains a lot
> 
> Screenshot from 2021-02-22 20-03-39

@torkelo thanks for noticing this, but I think we've already added this to the continuation issue 31305

torkelo:
Ah good 👍

PR #30654, opened 2021-01-26
Logs: Do not filter default message field

FliegenKLATSCH:
- Elasticsearch data source
- (Explore-Mode)
- Logs view

The field, which is configured as default message field in the data source settings, is filtered out.
It's not possible to _add_ another field in addition to the default message field (toggling the visibility of the available fields in a log message).

To be checked: Maybe this makes sense for other data sources, to filter this out?

### DIFF
--- a/packages/grafana-ui/src/components/Logs/logParser.ts
+++ b/packages/grafana-ui/src/components/Logs/logParser.ts
@@ -64,10 +64,8 @@ const getDerivedFields = memoizeOne(
     return (
       row.dataFrame.fields
         .map((field, index) => ({ ...field, index }))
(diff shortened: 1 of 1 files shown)


Elfo404:
Hi @FliegenKLATSCH , thanks for the PR!

I think I understood why you would want to do that, but I suspect in other datasources (eg. loki) this doesn't make much sense.

Would you please open an issue describing your problem and your use-case so that we can better discuss about it?

Thanks!

Elfo404:
I'm going to close this issue as it would cause troubles with other logging datasosurces, but it raised a good point about the overall logging experience in grafana. Let's keep the discussion alive in https://github.com/grafana/grafana/issues/30817

Thanks @FliegenKLATSCH for bringing this up!

PR #30663, opened 2021-01-27
Add downsampling info to prometheus datasource and make $__rate_interval downsampling-aware (#30662)

### LINKED ISSUE #30662: Downsampling- and retention-aware $__rate_interval for Thanos-backed prometheus data sources.
$__rate_interval should take downsampling policies into account for Thanos-backed prometheus datasources. "Zooming into the past" should not fail to produce meaningful results due to `$__rate_interval` being set, for example, to `150s` when retention policy only keeps one datapoint per hour.

**Why is this needed**:
Currently `$__rate_interval` is defined as `max( $__interval + Scrape interval, 4 * Scrape interval)`, which is fine for pure prometheus data sources. However, if datasource performs downsampling and separate retention policies for each downsampling level, that could be insufficient for a query to "just work" for all the use cases.

Say, we have a Thanos installation set as a Prometheus datasource. Thanos retention policies are set as follows:
RAW: 30 days
5 Minute: 90 days
1 Hour: 3 years
One has a panel with a query `rate(metric_name{}[$__rate_interval])`. It would work fine for the "Last N days" queries, but it would fail if one wishes to see a graph for a single day set 5 months into the past, the actual query would be something like:
`rate(metric_name{}[150s])`
But this specific datasource only has datapoints every 3600 seconds, which leads to empty result.
It would be marginally better to send a query like 
`rate(metric_name{}[7200s])`
and actually have sensible data displayed. After all, downsampled data is there for people to use.

krya-kryak:
This PR adds downsampling retention info to prometheus datasource and tweaks $__rate_interval to make use of it, as described in #30662.

The core purpose is to allow user to see all the data present in the datasource, regardless of the time window selected.
Here's an example. screenshots are made against the same thanos-query-frontend for the same query and the same time window set several months in the past. Admittedly, `up` is a lousy choice, but it still proves the point.

Retention policies unset:
retention-unset-small

Retention policies set:
retention-set-small

At this point only 1h downsampled datapoints are present in Thanos, and retention policies in grafana are set accordingly. Note the difference in computed `$__rate_interval`, allowing the rate query to produce results.

Fixes #30662

I would be happy to update tests and docs, just please tell if Grafana team is OK with such an approach.

Relevant Thanos PR: https://github.com/thanos-io/thanos/pull/3723


davkal:
Hi! First of all thanks for your contribution. I think this would make a real difference for Thanos users who use the dynamic interval in long term queries. The issue I still see is that at this point we don't want to make the datasource config page more complex than it needs to be by introducing fields that are only valid for one prometheus flavor. We got around this in the past by adding the "neutral" query param field. Perhaps there is a way to do this transparently here too? Eg., if those downsample periods are just default, then perhaps we can detect the presence of the custom query parameter and then still make the interval adjustment under the hood, based on the default downsample periods. Or put another way, how common is it that someone changes the periods?

krya-kryak:
@davkal David, thank you for an answer.

> if those downsample periods are just default, then perhaps we can detect the presence of the custom query parameter and then still make the interval adjustment under the hood, based on the default downsample periods. Or put another way, how common is it that someone changes the periods?

Tha defaults for Thanos are simple - no retention, data never gets deleted. Bless those lucky souls who can afford that :)
They wouldn't have the problem addressed here in the first place.

In my experience setting some sort of retention policy is extemely common for production usage. Personally, I've never seen any large scale installation without it - not just Thanos, but Graphite installations and two different in-house monitoring solutions. Every retention policy was different, balancing the requirements and resources available.

So, I doubt hardcoding some sort of defaults is a viable option.

> The issue I still see is that at this point we don't want to make the datasource config page more complex than it needs to be by introducing fields that are only valid for one prometheus flavor

I can think of two ways to reduce cluttering:
1. Hide some implementation-specific options (like this retention policies) either under the collapsable settings options (like HTTP auth credentials are normally hidden unless "basic auth" checkbox is active)
2. Add freeform text field that will just contain JSON with settings considered implementation-specific and/or experimental. That's where cluttering will be encapsulated.

davkal:
Sorry for the delayed response, we're in the middle of a release. I would rule out a free text form, that's simply confusing. I'd rather suggest to add a dropdown in the Misc section to select a "flavor", which could have "Prometheus" and "Thanos" as values. And if "Thanos" is selected, a new Thanos section appears. Wdyt?

davkal:
One more question: doesn't thanos have the knowledge on how the requested range matches the downsampled ranges, and could adjust the interval accordingly?

krya-kryak:
> One more question: doesn't thanos have the knowledge on how the requested range matches the downsampled ranges, and could adjust the interval accordingly?

While Thanos does have the knowledge of its downsampling policies, it does not modify the query in such a way. One would think it's kinda weird if you get results for `rate(metric[10m])` when request specifically names `rate(metric[1m])`. I believe it would take extending the PromQL language itself: https://github.com/prometheus/prometheus/issues/3806

krya-kryak:
Whoops, sorry for requesting review from everyone. Accidentally rebased my branch on v7.4.x instead of master.

Changes over original commit:
* UI changed the way @davkal suggested
* Instant query for downsampled-only timestamp now provides max_source_resolution, allowing for correct lookback interval for instant queries. Thanks you, @ahurtaud, for raising the issue in Slack.
* Tests added

public/app/plugins/datasource/prometheus/configuration/PromSettings.tsx
```diff
@@ -105,6 +121,17 @@ export const PromSettings = (props: Props) => {
             />
           </div>
         </div>
+        <div className="gf-form">
+          <InlineFormLabel width={14} tooltip="Datasource backend.">
```
osg-grafana, line 125:
```suggestion
          <InlineFormLabel width={14} tooltip="Data source backend.">
```

public/app/plugins/datasource/prometheus/configuration/PromSettings.tsx
```diff
@@ -116,6 +143,85 @@ export const PromSettings = (props: Props) => {
...
+                      placeholder="0"
+                    />
+                  }
+                />
+              </div>
+            </div>
+            <div className="gf-form-inline">
+              <div className="gf-form max-width-30">
+                <FormField
+                  label="Retention for 5m resolution"
```
osg-grafana, line 177: What does _m_ stand for here?

public/app/plugins/datasource/prometheus/configuration/PromSettings.tsx
```diff
@@ -116,6 +143,85 @@ export const PromSettings = (props: Props) => {
...
+                  labelWidth={14}
+                  tooltip="Thanos 5m downsampled data retention. 0 for unlimited."
```
osg-grafana, line 179: See comment on line 177.

public/app/plugins/datasource/prometheus/configuration/PromSettings.tsx
```diff
@@ -116,6 +143,85 @@ export const PromSettings = (props: Props) => {
...
+                      placeholder="0"
+                    />
+                  }
+                />
+              </div>
+            </div>
+            <div className="gf-form-inline">
+              <div className="gf-form max-width-30">
+                <FormField
+                  label="Retention for 1h resolution"
```
osg-grafana, line 201: I assume _h_ means _hour_ or _hours_ here. See also line 203. Probably makes sense in context...

torkelo (commented):
If you decide to accept this , Would try I move the Thanos specific stuff out of the main data source class (it's getting big) to another file (would of course need to be called from the main data source class) but sort similar to the result processor etc , maybe an interval processor? That can be different depending on flavor ?

public/app/plugins/datasource/prometheus/configuration/PromSettings.tsx
krya-kryak: Thank you. It did stand for minutes, as in PromQL duration. You're right though, it is unnecessary here. I've replaced promql abbreviations with English.

krya-kryak:
> If you decide to accept this , Would try I move the Thanos specific stuff out of the main data source class (it's getting big) to another file (would of course need to be called from the main data source class) but sort similar to the result processor etc , maybe an interval processor? That can be different depending on flavor ?

Thank you, Torkel. Fair point. Moved virtually everything from `datasource.ts` to `language_utils.ts` and `flavour_provider.ts`. The latter contains a class providing helper functions. Changes to `datasource.ts` are now reduced to bare minimum.

krya-kryak:
Merged master to fix a newly-emerged conflict.

krya-kryak:
Merged master to fix yet another merge conflict.

Folks, any sort of feedback would be much appreciated.

krya-kryak:
@davkal David, do you mind having a look please? Kudos.

krya-kryak:
Merged master to fix fresh merge conflict. Folks, I would really appreciate some feedback. Thanks!

davkal:
I'm still discussing this with @gouthamve and also reached out to bartek to get his thoughts. I realize it's taking quite some time, but i want this to be based on a consensus decision.

krya-kryak:
Hey folks, have you come to some sort of decision?

krya-kryak:
folks?

davkal:
This has been open way too long. The indecision also stems from the fact that it's a sizeable contribution. After re-reading the original issue #30662 it's clear that we don't want to fix it. I'm really sorry this took so long, and I realize that we've done you a disservice. But as it stands we won't add these Thanos-specific code paths to the prometheus datasource for the reasons stated in #30662 . Apologies again

PR #30664, opened 2021-01-27
TSDB: Use non-global config

aknuds1:
Refactor the TSDB backend code to use non-global configuration (no functionality changes!). This is necessary for #30353, unless we want to keep the approach of injecting the Postgres data source as a service (which I don't think we should) since it requires access to config in the Postgres data source.

Sorry for the huge PR, but it was a domino effect.o

Also performing some minor cleanup, such as renaming symbols to follow standard Go style.

### DIFF
--- a/pkg/api/alerting.go
+++ b/pkg/api/alerting.go
@@ -128,7 +128,7 @@ func GetAlerts(c *models.ReqContext) response.Response {
 }
 
 // POST /api/alerts/test
-func AlertTest(c *models.ReqContext, dto dtos.AlertTestCommand) response.Response {
+func (hs *HTTPServer) AlertTest(c *models.ReqContext, dto dtos.AlertTestCommand) response.Response {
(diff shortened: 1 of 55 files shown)


aknuds1:
Regarding CI failing, I have to fix up integration tests.

torkelo (commented):
The goal of removing global state is great, but I am afraid this breaks the existing patterns in a way that feels inconsistent and complex, and is working around the problem of the DI framework by not using DI at all.

I view system/service Cfg as a service level dependency, having it be passed as part of a command or query does not feel right to me. 

Passing it the dashboards.NewService function look good and however (but could also see us making this a proper service like the rest and injecting Cfg like the other services do)

aknuds1:
Closing this after discussing with @torkelo, since he doesn't think the PR follows the right pattern. We're also both unsure which other pattern to follow concretely.

PR #30668, opened 2021-01-27
PanelLibrary: changes casing of responses and adds meta property

hugohaggmark:
This PR changes the responses of the library panel api:
- adds a meta property with user names and canEdit flag (always true for now)
- changes casing so the api is more easily consumed by the frontend

Relates #29610

### DIFF
--- a/pkg/services/librarypanels/database.go
+++ b/pkg/services/librarypanels/database.go
@@ -5,13 +5,11 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
-
-	"github.com/grafana/grafana/pkg/util"
-
+	"github.com/grafana/grafana/pkg/api/dtos"
 	"github.com/grafana/grafana/pkg/models"
-
 	"github.com/grafana/grafana/pkg/services/sqlstore"
+	"github.com/grafana/grafana/pkg/services/sqlstore/migrator"
+	"github.com/grafana/grafana/pkg/util"
 )
 
 // createLibraryPanel adds a Library Panel.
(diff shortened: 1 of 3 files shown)


pkg/services/librarypanels/database.go
```diff
@@ -148,50 +166,109 @@ func (lps *LibraryPanelService) disconnectLibraryPanelsForDashboard(dashboardID
...
-	session.Table("library_panel")
-	session.Where("uid=? AND org_id=?", uid, orgID)
-	err := session.Find(&libraryPanels)
+func getLibraryPanel(session *sqlstore.DBSession, uid string, orgID int64) (LibraryPanelWithMeta, error) {
+	libraryPanels := make([]LibraryPanelWithMeta, 0)
+	sql := `SELECT
+				lp.id, lp.org_id, lp.folder_id, lp.uid, lp.name, lp.model, lp.created, lp.created_by, lp.updated, lp.updated_by
+				, 0 as can_edit
+				, (SELECT u.login from user u where u.id = lp.created_by and u.org_id = lp.org_id) as created_by_name
+				, (SELECT u.login from user u where u.id = lp.updated_by and u.org_id = lp.org_id) as updated_by_name
```
hugohaggmark, line 175: Is there a more better way to do this SELECT without sub selects? I'm a bit rusty here and appreciate any suggestions.

pkg/services/librarypanels/database.go
```diff
@@ -148,50 +166,109 @@ func (lps *LibraryPanelService) disconnectLibraryPanelsForDashboard(dashboardID
...
 	orgID := c.SignedInUser.OrgId
-	libraryPanels := make([]LibraryPanel, 0)
+	libraryPanels := make([]LibraryPanelWithMeta, 0)
 	err := lps.SQLStore.WithDbSession(context.Background(), func(session *sqlstore.DBSession) error {
-		err := session.SQL("SELECT * FROM library_panel WHERE org_id=?", orgID).Find(&libraryPanels)
+		sql := `SELECT
+				lp.id, lp.org_id, lp.folder_id, lp.uid, lp.name, lp.model, lp.created, lp.created_by, lp.updated, lp.updated_by
+				, 0 as can_edit
+				, (SELECT u.login from user u where u.id = lp.created_by and u.org_id = lp.org_id) as created_by_name
+				, (SELECT u.login from user u where u.id = lp.updated_by and u.org_id = lp.org_id) as updated_by_name
```
hugohaggmark, line 235: Same comment as above, could this SELECT be done in a better way?

pkg/services/librarypanels/database.go
torkelo: should be able to just do a join, important to get the user email so you can return the avatar image url as well, see https://github.com/grafana/grafana/blob/master/pkg/services/sqlstore/team.go#L409 for reference

hugohaggmark: Thank you, I'll look into this.

hugohaggmark: I couldn't figure out how to use `xorm` with two joins using the same table name so I wrote it like an SQL statement instead.

pkg/services/librarypanels/models.go
```diff
@@ -22,6 +22,56 @@ type LibraryPanel struct {
...
+	CreatedByName  string
+	CreatedByEmail string
+	UpdatedByName  string
+	UpdatedByEmail string
+}
+
+// LibraryPanelDTO is the frontend DTO for library panels.
+type LibraryPanelDTO struct {
+	ID       int64               `json:"id"`
+	OrgID    int64               `json:"orgID"`
```
torkelo, line 49: should this be orgId (in the json)?

torkelo (approved):
Think this looks good, just one minor json prop name issue

pkg/services/librarypanels/models.go
hugohaggmark: Yes, of course. Good catch!

hugohaggmark:
Any last feedback from @kaydelaney, @aknuds1 or @kylebrandt before I merge this?

hugohaggmark:
I'm merging this PR now. Please leave any additional comments and I'll fix them in a later PR @aknuds1 @kylebrandt. Thank you.

PR #30671, opened 2021-01-27
Grafana-UI: Fix setting default value for MultiSelect

### LINKED ISSUE #30649: MultiSelect: Setting defaultValue doesn't work
**What happened**:
Setting `defaultValue` for `MultiSelect` or `<Select isMulti />` doesn't do anything. Underlying react-select library disregards `defaultValue` if `value` is set, which in case of `MultiSelect` defaults to empty array. Instead the value for MultiSelect should be `undefined` if no value is provided.

Providing `defaultValue` prop would correctly set default value and display it as currently selected.

Clarity-89:
Default to `undefined` instead of an empty array for Select's `value`, to enable `defaultValue` to be set. 

Fixes https://github.com/grafana/grafana/issues/30649


torkelo (approved):
Looks good

PR #30676, opened 2021-01-27
AlertingNG: Edit Alert Definition

peterholmberg:
Enables editing an alert. This also removes the sidemenu button and moves "Add NG Alert" to the alert rules list page.

Fixes #

@kylebrandt @torkelo I think we need to discuss how we want the datasource picker to work. If we're forcing it to mixed we need some way to disable or hide the datasource pickers so that users can't use anything but mixed.


kylebrandt:
Somehow, when loading some of my existing ones, the refIds are blank:

image

In my console when this happens I get:

image

(Could be some wrong with my model based on changes we made and not this PR - need to look closer)

kylebrandt:
> I think we need to discuss how we want the datasource picker to work...

Forcing mixed I think may just be a temp fix.

In the case of the data source picker, at least currently, I don't believe it has anything to do with the backend behavior around this. However, we do need a place to persist the information (although you could perhaps scan all the queries for their datasource value, and pick the value of the datasource picker based on that).

At least with data sources, I have found storing the front end model information and the query information together to be confusing when working with backend data sources. So I wonder if it is better to have a separate property on the alert definition that is a json blob for properties that are solely front-end model settings.

Although if it turns out a property needs to be backend processing as well, then we would need the discipline to move the property (because if in the backend we start fetching properties from the frontend model section, things will end up even more confusing than mixing them).

public/app/features/alerting/EditNGAlert.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const mapDispatchToProps: MapDispatchToProps<DispatchProps, OwnProps> = {
+  getAlertDefinition,
+  updateAlertDefinition,
+};
+
+export default hot(module)(
+  connectWithCleanUp(mapStateToProps, mapDispatchToProps, (state) => state.alertDefinition)(EditNGAlert)
+);
```
hugohaggmark, line 56: What is the reason why we use PureComponent instead of Functional Component?

public/app/features/alerting/NewNGAlert.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+};
+
+const mapDispatchToProps: MapDispatchToProps<DispatchProps, OwnProps> = {
+  createAlertDefinition,
+};
+
+export default hot(module)(
+  connectWithCleanUp(mapStateToProps, mapDispatchToProps, (state) => state.alertDefinition)(NewNGAlert)
+);
```
hugohaggmark, line 41: Why not a Functional Component?

hugohaggmark (commented):
I did a quick read thru and now I'm going to test it. My key question is why you're not using Functional Components?

hugohaggmark (changes requested):
1. I think the default data source should be selected for new queries
2. It seems like the state is mutated when I try to edit a Prometheus Query.
image

public/app/features/alerting/EditNGAlert.tsx
peterholmberg: I want these to connect to the store and while I think that's kind of possible in functional components too I think  it looks cleaner in a PureComponent, is there a huge upside to using functional components here?

hugohaggmark: I was just curious really, seems like such a small Component so it felt like it should be a Functional Component and all the ceremony of connecting to Redux is much less in a Functional Component.

peterholmberg: Yeah, maybe it's a bit overkill.

public/app/features/alerting/EditNGAlert.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const { pageId, getAlertDefinition } = this.props;
+    getAlertDefinition(pageId);
+  }
+
+  render() {
+    const { alertDefinition, updateAlertDefinition } = this.props;
+
+    if (alertDefinition.id < 1) {
+      return <Spinner />;
+    }
```
dprokop, line 36: @peterholmberg why not using an additional loading state property rather than relying on the id? Feels like a pottential source of problems/confussion when the contract changes.

public/app/features/alerting/__snapshots__/AlertRuleList.test.tsx.snap
```diff
@@ -73,6 +73,12 @@ exports[`Render should render alert rules 1`] = `
       <div
         className="page-action-bar__spacer"
       />
+      <LinkButton
```
dprokop, line 76: Maybe this PR is a good opportunity to get rid of this snapshot?

public/app/features/alerting/components/AlertDefinitionItem.tsx
```diff
@@ -21,6 +21,12 @@ export const AlertDefinitionItem: FC<Props> = ({ alertDefinition, search }) => {
           <span key="text">{alertDefinition.description}</span>
         </span>
       </Card.Meta>
+      <Card.Actions>
+        <div />
```
dprokop, line 25: Is this empty div necessary?

public/app/features/alerting/EditNGAlert.tsx
peterholmberg: Problem solved :slightly_smiling_face:

public/app/features/alerting/components/AlertDefinitionItem.tsx
peterholmberg: Well, the `<Card.Actions>` components expects an array of children. So it's either changing the Actions or having the dummy div or wrapping the LinkButton in an array.

public/app/features/alerting/__snapshots__/AlertRuleList.test.tsx.snap
peterholmberg: Sure!

kylebrandt:
Working for me <3

hugohaggmark (commented):
I'm going to remove my request for changes for this PR so it can make it into 7.4.0.

hugohaggmark (approved):
I'm going to remove my request for changes for this PR so it can make it into 7.4.0.

PR #30680, opened 2021-01-27
AlertingNG: Update UX to use new PageToolbar component

torkelo:
Updates alerting UX to use new PageToolbar & Toolbar buttons. 

It does change the UI a bit, but we can tweak that later. Need some UX feedback on that, easy to change now that all toolbars use the same component 

Screenshot from 2021-01-27 16-47-21


peterholmberg:
Should it not have a different background? 

Screenshot from 2021-01-29 10-17-57


Screenshot from 2021-01-29 10-14-49

Issue #30607, opened 2021-01-25
Influx: Incomplete graphs on wider screens when using Flux


davkal:
**What happened**:

When using a Flux query with `aggregateWindow` with `v.windowPeriod` on multi-day data on a wide screen, the resulting graph is incomplete (works fine on small displays)

**What you expected to happen**:

All requested data should be drawn.

**How to reproduce it (as minimally and precisely as possible)**:

0. Have an Influx datasource
1. Add a dashboard and graph panel with a Flux query where you know you have multiple days of data, and make sure to use `aggregateWindow(every: v.windowPeriod)`
2. View panel on wide screen (e.g, 2560 px wide)
3. Adjust the timepicker to an odd `From`, say `now-5d`
Result: Only part of the series is drawn (and if the query returns multiple series, only the first series is shown)

**Anything else we need to know?**:

On my 2560 px screen, the max datapoint calculated is 2110, while for the range of `now-5d` the auto window period is `2m`. This is clearly off since 5 days of data on a 2m window are: 5 * 24 * 60 / 2 = 3600 datapoints. The query stats panel shows `Total number rows | 3166`, which happens to be 3600 / 1.5. [And 1.5x happens to be the max data cutoff point for the influx/flux backend engine](https://github.com/grafana/grafana/blob/a2a6b9448ff0516a50738ac53fea28a3365d016f/pkg/tsdb/influxdb/flux/executor.go#L30).

davkal:
> On my 2560 px screen, the max datapoint calculated is 2110, while for the range of now-5d the auto window period is 2m. This is clearly off since 5 days of data on a 2m window are: 5 * 24 * 60 / 2 = 3600 datapoints. The query stats panel shows Total number rows | 3166, which happens to be 3600 / 1.5. And 1.5x happens to be the max data cutoff point for the influx/flux backend engine.

@torkelo  something must be odd in the maxDatapoints and Interval interplay. It seems to be calculated somewhere else before being handed to the influx/flux go backend, so I suspect other datasources that have a maxDatapoints cut off might suffer too?

FYI My hotfix here just raises the maxDatapoints factor from 1.5 to 2.

R-Studio:
It looks like the same issue as: https://github.com/grafana/grafana/issues/26484 
_The issue occurs also on the newest version of Grafana v7.4.0._

davkal:
[...]

PR #30688, opened 2021-01-27
InfluxDB: Show all datapoints for dynamically windowed flux query

davkal:
Hotfix for #30607 by raising the max point limit during dataframe construction.

### DIFF
--- a/pkg/tsdb/influxdb/flux/executor.go
+++ b/pkg/tsdb/influxdb/flux/executor.go
@@ -27,7 +27,7 @@ func executeQuery(ctx context.Context, query queryModel, runner queryRunner, max
 		glog.Warn("Flux query failed", "err", err, "query", flux)
 		dr.Error = err
 	} else {
-		dr = readDataFrames(tables, int(float64(query.MaxDataPoints)*1.5), maxSeries)
+		dr = readDataFrames(tables, int(float64(query.MaxDataPoints)*2), maxSeries)
 	}
 
 	// Make sure there is at least one frame


pkg/tsdb/influxdb/flux/executor.go
```diff
@@ -27,7 +27,7 @@ func executeQuery(ctx context.Context, query queryModel, runner queryRunner, max
 		glog.Warn("Flux query failed", "err", err, "query", flux)
 		dr.Error = err
 	} else {
-		dr = readDataFrames(tables, int(float64(query.MaxDataPoints)*1.5), maxSeries)
```
ryantxu, line 30: I wonder if we should go even bigger?  the real goal is avoiding selecting way too many things maybe 2.5x or even 5x?

ryantxu (approved):

pkg/tsdb/influxdb/flux/executor.go
davkal: It's a bit tough, because 10k datapoints sounds like a lot, but that's the max default for flux query models if it's not set via the frontend, so I think we're safe here. And if you get here via the frontend, then you don't really want more because it can't be drawn properly?

ryantxu: the real solution is to add a warning/notice to the result when the data is clipped using this maxDataPoints -- but I think this is a good quick fix

davkal: Ha, there is an error generated when it's clipped, but the error is overwritten AFAI can tell: https://github.com/grafana/grafana/blob/5f1f820bb99eeca7a6dfe5f428b2b593d38585d4/pkg/tsdb/influxdb/flux/executor.go#L86

ryantxu: rather than an error I mean something like:
```
frame.AppendNotices(data.Notice{
		Severity: data.NoticeSeverityWarning,
		Text: "Results are truncated based on maxDataPoints",
	})
```

PR #30691, opened 2021-01-27
CDN: Adds support for serving assets over a CDN

### LINKED ISSUE #6844: [Feature Request]: Support serving static assets of CDN

torkelo:
Closes #6844

Adds option to load js & css from cdn path. Tested by starting a local nginx with a custom hostname and copying everything under public to it.

This does not currently impact loading plugins, they are still loaded from instance.

It does impact the content_security_policy, if enabled it needs to be modified allow the cdn. Not sure how exactly.

TODO
* [ ] How to make it apply to all images 

Screenshot from 2021-01-27 22-39-34

### DIFF
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -263,7 +263,7 @@ Path where the socket should be created when `protocol=socket`. Make sure that G
 
 > **Note**: Only available in v7.4+
 
-Specify a full http url address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
+Specify a full HTTP URL address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
 
 For example, given a cdn url like `https://cdn.myserver.com` grafana will try to load a javascript file from
 `http://cdn.myserver.com/oss/v7.4.0/public/build/app.<hash>.js`.
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -261,7 +261,7 @@ Path where the socket should be created when `protocol=socket`. Make sure that G
 
 ### cdn_url
 
-> **Note**: Only available in v7.4+
+> **Note**: Available in Grafana v7.4 and later versions.
 
 Specify a full HTTP URL address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
 
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -266,7 +266,7 @@ Path where the socket should be created when `protocol=socket`. Make sure that G
 Specify a full HTTP URL address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
 
 For example, given a cdn url like `https://cdn.myserver.com` grafana will try to load a javascript file from
-`http://cdn.myserver.com/oss/v7.4.0/public/build/app.<hash>.js`.
+`http://cdn.myserver.com/grafana-oss/v7.4.0/public/build/app.<hash>.js`.
 
 <hr />
 
--- a/docs/sources/administration/configuration.md
+++ b/docs/sources/administration/configuration.md
@@ -259,6 +259,15 @@ Path to the certificate key file (if `protocol` is set to `https` or `h2`).
 
 Path where the socket should be created when `protocol=socket`. Make sure that Grafana has appropriate permissions before you change this setting.
 
+### cdn_url
+
+> **Note**: Only available in v7.4+
+
+Specify a full http url address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
+
+For example, given a cdn url like `https://cdn.myserver.com` grafana will try to load a javascript file from
+`http://cdn.myserver.com/oss/v7.4.0/public/build/app.<hash>.js`.
+
 <hr />
 
 ## [database]
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -1309,8 +1309,8 @@ func (cfg *Cfg) readServerSettings(iniFile *ini.File) error {
 	return nil
 }
 
-// Returns full CDN Path with BuildVersion and edition added
-func (cfg *Cfg) GetFullCDNURL(edition string) string {
+// GetContentDeliveryURL returns full content delivery URL with /<edition>/<version> added to URL
+func (cfg *Cfg) GetContentDeliveryURL(edition string) string {
 	if cfg.CDNRootURL != nil {
 		url := *cfg.CDNRootURL
 		preReleaseFolder := ""
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -197,7 +197,7 @@ type Cfg struct {
 	SocketPath       string
 	RouterLogging    bool
 	Domain           string
-	CDNPath          string
+	CDNRootURL       *url.URL
 
 	// build
 	BuildVersion string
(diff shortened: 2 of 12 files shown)


public/app/index.ts
```diff
@@ -1,4 +1,6 @@
-import app from './app';
+declare let __webpack_public_path__: string;
+__webpack_public_path__ = window.public_cdn_path;
```
ryantxu, line 2: is this used?

torkelo: Yes this is the key part of telling webpack where to load files

torkelo: Without this all the dynamic webpack chunks are loaded from localhost

hugohaggmark (approved):
I think the code looks ok, but I'm not very familiar with all the details.

marefr (approved):
Awesome :+1: 

Other things to consider referencing CDN might be email templates for example.

Another thing to take into consideration would be to not having to update the CDN path config each time you upgrade Grafana - I guess this partly will be solved by the build hashes, but thinking about 10 Grafana releases would currently require 10x all the files with hashes in the same CDN folder, unless CDN path config is not changed on upgrade. Maybe would make sense to always enforce/include version number/commit in the CDN path?

pkg/setting/setting.go
```diff
@@ -1297,6 +1298,7 @@ func readServerSettings(iniFile *ini.File, cfg *Cfg) error {
 		return err
 	}
 
+	cfg.CDNPath = valueAsString(server, "cdn_path", "")
```
torkelo, line 1301: @marefr think here we should probably add Grafana version string to the path (so v7.4.0-beta1) for example. (or in index.go) so that on the cdn each published release is under different paths , the config path you set in the ini file is just the root cdn path

bergquist: using a hash would work with master builds and pre-releases as well. That would be nice.

marefr: So maybe both version and hash then? Or only using hash for pre-releases and version for stable ones, not sure how hard that would be.

torkelo: The asset files are already hashed, this should just add versio folder. Master builds already build a number of some sort I think

marefr: True, makes sense

torkelo:
@bergquist @marefr added edition & and BuildVersion to the url as well so paths becomes 

https://cdn.grafana.com/oss/v7.5.0-11124pre 
https://cdn.grafana.com/enterprise/v7.5.0-11124pre

if the cdn_path is set to https://cdn.grafana.com   depending on what verison & edition grafana is

torkelo:
Added one final thing, also segrate master builds into a separate path for easier clean-up. So it will be 
```
/oss
  /v7.4.0-beta1
  /v7.4.0
  /master
     /v7.5.0-121pre
/enterprise
 .... 
```

pkg/setting/setting.go
```diff
@@ -1297,9 +1298,27 @@ func readServerSettings(iniFile *ini.File, cfg *Cfg) error {
...
 	return nil
 }
 
+// Returns full CDN Path with BuildVersion and edition added
+func (cfg *Cfg) GetFullCDNPath(edition string) string {
+	if cfg.CDNPath != "" {
+		cdnUrl, err := url.Parse(cfg.CDNPath)
+		if err == nil {
```
marefr, line 1309: I think this should be treated as an error and either logged or failing to start since the intent of user has been to configure this and if that fails the user should be notified.

pkg/setting/setting_test.go
```diff
@@ -389,3 +389,28 @@ func TestAuthDurationSettings(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, maxLifetimeDurationTest, cfg.LoginMaxLifetime)
 }
+
+func TestGetCDNPath(t *testing.T) {
```
marefr, line 393: Would be nice with one test where cdn path includes a path as well, e.g. `http://cdn.grafana.com/something/`

marefr (changes requested):
Awesome. This is great :+1: Spotted some minor problems, see comments. .ini files and documentation needs update as well

pkg/setting/setting.go
torkelo: good point

pkg/setting/setting_test.go
torkelo: :+1:

torkelo:
@marefr pushed changes, added error handling in the read config stage (of the url parsing), changed the property name from cdn_path to cd_url as it's a url not just path. Added docs

public/views/index-template.html
```diff
@@ -199,14 +197,18 @@
         </p>
         <p>
           1. This could be caused by your reverse proxy settings.<br /><br />
-          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath. If not using a reverse proxy make sure to set serve_from_sub_path to true.<br />
+          2. If you host grafana under subpath make sure your grafana.ini root_url setting includes subpath. If not
+          using a reverse proxy make sure to set serve_from_sub_path to true.<br />
```
achatterjee-grafana, line 201:
```suggestion
          use a reverse proxy to set serve_from_sub_path to true.<br />
```

public/views/index-template.html
```diff
@@ -199,14 +197,18 @@
...
           <br />
           3. If you have a local dev build make sure you build frontend using: yarn start, yarn start:hot, or yarn
           build<br />
           <br />
           4. Sometimes restarting grafana-server can help<br />
           <br />
-          5. You might be using a non-supported browser. To make sure read the documentation on <a href="https://grafana.com/docs/grafana/latest/installation/requirements/#supported-web-browsers">supported browsers</a>.
+          5. You might be using a non-supported browser. To make sure read the documentation on
```
achatterjee-grafana, line 208:
```suggestion
          5. Check if you are using a non-supported browser. For more information, refer to the list of 
```

docs/sources/administration/configuration.md
```diff
@@ -259,6 +259,15 @@ Path to the certificate key file (if `protocol` is set to `https` or `h2`).
 
 Path where the socket should be created when `protocol=socket`. Make sure that Grafana has appropriate permissions before you change this setting.
 
+### cdn_url
+
+> **Note**: Only available in v7.4+
```
achatterjee-grafana, line 264:
```suggestion
> **Note**: Available in Grafana v7.4 and later versions.
```

docs/sources/administration/configuration.md
```diff
@@ -259,6 +259,15 @@ Path to the certificate key file (if `protocol` is set to `https` or `h2`).
...
+
+Specify a full http url address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
```
achatterjee-grafana, line 266:
```suggestion
Specify a full HTTP URL address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
```

achatterjee-grafana (commented):
Added some copy-edit suggestions for the doc content.

public/views/index-template.html
torkelo: the suggestion has a different meaning which is incorrect.

public/views/index-template.html
torkelo: updated!

docs/sources/administration/configuration.md
```diff
@@ -259,6 +259,15 @@ Path to the certificate key file (if `protocol` is set to `https` or `h2`).
 
 Path where the socket should be created when `protocol=socket`. Make sure that Grafana has appropriate permissions before you change this setting.
 
+### cdn_url
+
+> **Note**: Available in Grafana v7.4 and later versions.
+
+Specify a full HTTP URL address to the root of your Grafana CDN assets. Grafana will add edition and version paths.
+
+For example, given a cdn url like `https://cdn.myserver.com` grafana will try to load a javascript file from
```
achatterjee-grafana, line 268:
```suggestion
For example, given a CDN URL like `https://cdn.myserver.com` grafana will try to load a javascript file from
```

achatterjee-grafana (approved):
Added the suggestion to fix a typo that got missed earlier.

marefr (approved):
LGTM

PR #30700, opened 2021-01-28
TraceViewer: Fix lazy loading by changing the scroll element

zoltanbedi:

Lazy loading broke again for trace view, this pr fixes it by changing the scroll element.

## TODO
- [x] Add e2e test after https://github.com/grafana/grafana/pull/30644


e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
+import { e2e } from '@grafana/e2e';
+
+e2e.scenario({
+  describeName: 'Trace view',
+  itName: 'Can lazy load big traces',
+  addScenarioDataSource: false,
+  addScenarioDashBoard: false,
+  skipScenario: false,
```
hugohaggmark, line 8: Just a small nit, you don't need to use e2e.scenario in this case as you're not adding a dash or data source. Could be easier to just do:
```ts
describe('Trace view', () => {
  it('Can lazy load big traces', () => {
    e2e.flows.login('admin', 'admin');
```

e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
...
+  scenario: () => {
+    e2e().intercept('GET', '/api/datasources/proxy/29/api/traces/long-trace', {
+      fixture: 'long-trace-response.json',
+    });
```
hugohaggmark, line 12: Nice, this is Cypress new api right?

e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
...
+      .should('be.visible')
+      .within(() => {
+        e2e.components.Select.input().should('be.visible').click();
+
+        e2e().contains('gdev-jaeger').scrollIntoView().should('be.visible').click();
+      });
+
+    e2e.components.QueryField.container().should('be.visible').type('long-trace');
+
+    e2e.components.RefreshPicker.runButton().click();
```
hugohaggmark, line 26: To avoid flakiness:
```suggestion
    e2e.components.RefreshPicker.runButton().should('be.visible').click();
```

e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    e2e().get('[data-test-id="SpanBar--wrapper"]').should('have.length', 100);
```
hugohaggmark, line 28: This should be a proper e2e selector, you can still use `[data-test-id="SpanBar--wrapper"]` in the e2e selector.

e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
...
+    e2e().get('.scrollbar-view').scrollTo('center');
```
hugohaggmark, line 29: This should also be a proper e2e selector.

e2e/suite1/specs/trace-view-scrolling.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // After scrolling we should have 140 spans instead of the first 100
+    e2e().get('[data-test-id="SpanBar--wrapper"]').should('have.length', 140);
```
hugohaggmark, line 32: Same comment as above.

hugohaggmark (commented):
Looking good but would be better to use proper e2e selectors I think.

e2e/suite1/specs/trace-view-scrolling.ts
zoltanbedi: It is yes. Works like a charm 😄

zoltanbedi:
Thank you for the review @hugohaggmark, I knew I needed your eyes on this 😄

hugohaggmark (approved):
Looks great now 🎉 !

PR #30701, opened 2021-01-28
Chore: grafana-toolkit uses grafana-ui and grafana-data workspaces

### LINKED ISSUE #31185: Should grafana-toolkit consume the local monorepo versions of grafana-ui and grafana-data?
Right now, it [depends](https://github.com/grafana/grafana/blob/9629dded42e5457266a2448ce5e9851b2bcb6749/packages/grafana-toolkit/package.json#L31) on version `next` of these packages, causing them to be pulled from npm, rather than linked from the local monorepo.  Their dependencies are also pulled down, possibly creating version conflicts with the hoisted dependencies in the repo root.  One consequence is that `@types/slate` is pulled in as both 0.47.1 and 0.47.2, requiring the TypeScript compiler to waste a bunch of time establishing that the types are equivalent across versions.

jackw:
Currently `@grafana/toolkit` package.json references `@grafana/ui: "next"` and `@grafana/data: "next"`. Because of the `yarn.lock` file this results in stale packages (`@grafana/ui@7.0.0` and `@grafana/data@7.0.0`) being installed. This adds outdated packages to the dependency tree which brings both a lot of noise and potentially causes users to believe we use outdated / vulnerable packages or worse developers importing these outdated / obsolete dependencies.

Fixes #31185 

Note that popperjs v1 types were still being referenced in the grafana-ui codebase. This PR updates those too.

### DIFF
--- a/packages/grafana-toolkit/package.json
+++ b/packages/grafana-toolkit/package.json
@@ -28,10 +28,10 @@
   "dependencies": {
     "@babel/core": "7.9.0",
     "@babel/preset-env": "7.9.0",
-    "@grafana/data": "next",
+    "@grafana/data": "7.5.0-pre.0",
     "@grafana/eslint-config": "2.2.0",
     "@grafana/tsconfig": "^1.0.0-rc1",
-    "@grafana/ui": "next",
+    "@grafana/ui": "7.5.0-pre.0",
     "@types/command-exists": "^1.2.0",
     "@types/execa": "^0.9.0",
     "@types/expect-puppeteer": "3.3.1",
--- a/packages/grafana-ui/src/components/Forms/Legacy/Switch/Switch.tsx
+++ b/packages/grafana-ui/src/components/Forms/Legacy/Switch/Switch.tsx
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react';
 import uniqueId from 'lodash/uniqueId';
+import { Placement } from '@popperjs/core';
 import { Tooltip } from '../../../Tooltip/Tooltip';
-import * as PopperJS from 'popper.js';
 import { Icon } from '../../..';
 
 export interface Props {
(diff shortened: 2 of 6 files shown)


packages/grafana-toolkit/package.json
```diff
@@ -28,10 +28,10 @@
   "dependencies": {
     "@babel/core": "7.9.0",
     "@babel/preset-env": "7.9.0",
-    "@grafana/data": "next",
+    "@grafana/data": "7.5.0-pre.0",
```
dprokop, line 31: I wonder if we had specified this as ^7.0.0, wouldn't it by default resolve to workspace package?

dprokop: Would be good to see why https://github.com/grafana/grafana/pull/30428 did not update toolkit's dependencies

jackw: > I wonder if we had specified this as ^7.0.0, wouldn't it by default resolve to workspace package? Asking because otherwise, we will have to update the version string with each release.

I believe Lerna (which uses yarn workspaces under the hood) will only bump versions if the dependency matches the workspace version. Otherwise it will treat it as a package to pull from the registry. As noted [here](https://classic.yarnpkg.com/en/docs/workspaces/#toc-limitations-caveats) in the yarn docs...

>In the example above, if workspace-b depends on a different version than the one referenced in workspace-a’s package.json, the dependency will be installed from npm rather than linked from your local filesystem. This is because some packages actually need to use the previous versions in order to build the new ones (Babel is one of them).

> Would be good to see why #30428 did not update toolkit's dependencies

I kind of confirmed this by running the `yarn packages:prepare` script in master and compared it to running the same script in this branch. This branch will bump the versions used by grafana-toolkit whilst in master it leaves them untouched.

amcasey: I also understood that they had to match exactly.

amcasey (approved):

dprokop (approved):

Issue #30686, opened 2021-01-27
X-axis time scrolling is broken due to commit b52f9fed68


faxm0dem:
This commit https://github.com/grafana/grafana/commit/b52f9fed68 breaks the time scrolling, when new data arrives e.g. through streaming datasources.

torkelo:
no it still works as it did, https://play.grafana.org/d/nP8rcffGk/1-new-features-in-v7-4?viewPanel=29&orgId=1

the time range needs to be relative

faxm0dem:
Well, my time range is relative, so there must be something else wrong here !
I see on your example there is a dot toggling "streaming start/stop", I guess I'm missing that in my case. How do I enable that ?
@torkelo

faxm0dem:
For the record, @torkelo answered on another channel that the trick was to set the **state** when calling the subscriber:

```typescript
subscriber.next({
  [...]
  state: LoadingState.Streaming,
});
```

PR #30704, opened 2021-01-28
Fix documentation for streaming data sources

faxm0dem:
The added code is now mandatory since 7.4.x
See https://github.com/grafana/grafana/issues/30686


marcusolsson (approved):
LGTM. Thanks for helping keeping the plugin docs up-to-date!

PR #30706, opened 2021-01-28
DashboardPicker: switch to promise-based debounce, return dashboard UID

Clarity-89:
- Use [debounce-promise](https://github.com/bjoerge/debounce-promise) instead of lodash's debounce for `loadOptions` function. The promise-based approach is necessary because otherwise the results fo the first call will overwrite the results for the debounced call when filtering options. The issue is described in great detail [here](https://github.com/JedWatson/react-select/issues/3075#issuecomment-450194917). 
- Return dashboard UID from the search, so it can be used for retrieving dashboards via API.


package.json
```diff
@@ -89,6 +89,7 @@
     "@types/d3": "5.7.2",
     "@types/d3-force": "^2.1.0",
     "@types/d3-scale-chromatic": "1.3.1",
+    "@types/debounce-promise": "^3.1.3",
```
hugohaggmark, line 92: nit: always use exact versions

package.json
```diff
@@ -234,6 +235,7 @@
     "d3-force": "^2.1.1",
     "d3-scale-chromatic": "1.5.0",
     "dangerously-set-html-content": "1.0.6",
+    "debounce-promise": "^3.1.2",
```
hugohaggmark, line 238: nit: always use exact versions

package.json
Clarity-89: True, great catch!

hugohaggmark (approved):
Looks great and works as expected 🎉

PR #30709, opened 2021-01-28
PanelLibrary: better handling of deleted panels

hugohaggmark:
This PR does two things:

1. Whenever a library panel is deleted we also clean up any dashboard connections
2. When loading a dashboard that includes a library panel that isn't connected to the dash would previously fail the load. This PR changes this behavior by setting a minimal JSON that ends up showing something like this 
image
This should probably be even more refined but this will do for now.
3. Adds name to panels in the test dashboard

Relates #29610

### DIFF
--- a/pkg/services/librarypanels/database.go
+++ b/pkg/services/librarypanels/database.go
@@ -113,13 +113,20 @@ func (lps *LibraryPanelService) connectLibraryPanelsForDashboard(c *models.ReqCo
 
 // deleteLibraryPanel deletes a Library Panel.
 func (lps *LibraryPanelService) deleteLibraryPanel(c *models.ReqContext, uid string) error {
-	orgID := c.SignedInUser.OrgId
 	return lps.SQLStore.WithTransactionalDbSession(context.Background(), func(session *sqlstore.DBSession) error {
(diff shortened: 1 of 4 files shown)


pkg/services/librarypanels/database.go
```diff
@@ -113,13 +113,20 @@ func (lps *LibraryPanelService) connectLibraryPanelsForDashboard(c *models.ReqCo
...
 		}
 
+		if _, err := session.Exec("DELETE FROM library_panel_dashboard WHERE librarypanel_id=?", panel.ID); err != nil {
+			return err
+		}
+
+		result, err := session.Exec("DELETE FROM library_panel WHERE id=?", panel.ID)
+		if err != nil {
+			return err
+		}
```
hugohaggmark, line 129: I went back and forth on the order for this. Should we delete the connections before the panel or the other way around?

aknuds1: Connections first, but it's all in a transaction anyway isn't it, so shouldn't matter to the outside world.

aknuds1 (approved):

pkg/services/librarypanels/database.go
hugohaggmark: Yeah, I finally got it right 😉

PR #30713, opened 2021-01-28
Auth: Use SigV4 lib from grafana-aws-sdk 

wbrowne:


pkg/models/datasource_cache.go
```diff
@@ -319,3 +321,14 @@ func ClearDSDecryptionCache() {
...
 }
+
+func awsServiceNamespace(dsType string) string {
+	switch dsType {
+	case DS_ES, DS_ES_OPEN_DISTRO:
+		return "es"
+	case DS_PROMETHEUS:
+		return "aps"
+	default:
+		panic(fmt.Sprintf("Unsupported datasource %s", dsType))
```
aknuds1, line 332: I would quote `dsType`, in case it contains whitespace:

```suggestion
		panic(fmt.Sprintf("Unsupported datasource %q", dsType))
```

aknuds1 (approved):

go.mod
```diff
@@ -42,6 +42,7 @@ require (
 	github.com/google/go-cmp v0.5.4
 	github.com/google/uuid v1.2.0
 	github.com/gosimple/slug v1.9.0
+	github.com/grafana/grafana-aws-sdk v0.0.0-20210112151715-f082803d987a
```
marefr, line 45: Would be nice to tag the github.com/grafana/grafana-aws-sdk, e.g. v0.1.0, to make it easier to understand what version/changes of the lib grafana is using. Probably a good idea to upgrade github.com/grafana/grafana-plugin-sdk-go to the latest version in github.com/grafana/grafana-aws-sdk as well.

marefr (approved):
LGTM, but added a comment that would be nice to address now or later

go.mod
wbrowne: Good idea - will follow up with that!

PR #30716, opened 2021-01-28
GrafanaUI: Add a way to persistently close InfoBox

### LINKED ISSUE #24664: FeatureInfoBox: add possibility to close/collapse the box
This `Transformations info box` is massive – it should be possible to close it without clicking on transformation type. Especially that it keeps coming back if you delete all transformations.
image

_Originally posted by @sarlinska in https://github.com/grafana/grafana/issues/23551#issuecomment-628082464_

oscarkilhed:
InfoBox and FeatureInfoBox can take up a lot of screen real estate. This makes it easy to let the user close the boxes and not have them come back. It also enables this feature on the TransformationsEditor.

Fixes #24664

### DIFF
--- a/packages/grafana-ui/src/components/InfoBox/InfoBox.story.tsx
+++ b/packages/grafana-ui/src/components/InfoBox/InfoBox.story.tsx
@@ -1,11 +1,17 @@
 import React from 'react';
-import { number, select, text } from '@storybook/addon-knobs';
 import { FeatureState } from '@grafana/data';
 import { InfoBox, FeatureInfoBox } from '@grafana/ui';
 import mdx from './InfoBox.mdx';
-import { DismissableFeatureInfoBox, FEATUREINFOBOX_PERSISTENCE_ID_PREFIX } from './DismissableFeatureInfoBox';
+import {
+  DismissableFeatureInfoBox,
+  DismissableFeatureInfoBoxProps,
+  FEATUREINFOBOX_PERSISTENCE_ID_PREFIX,
+} from './DismissableFeatureInfoBox';
 import { Button } from '../Button';
 import { css } from 'emotion';
+import { Story } from '@storybook/react';
+import { FeatureInfoBoxProps } from './FeatureInfoBox';
+import { InfoBoxProps } from './InfoBox';
 
 export default {
   title: 'Layout/InfoBox',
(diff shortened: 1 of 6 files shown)


packages/grafana-ui/src/components/InfoBox/InfoBox.tsx
```diff
@@ -24,24 +27,61 @@ export interface InfoBoxProps extends Omit<React.HTMLAttributes<HTMLDivElement>,
...
+      ref
+    ) => {
       const theme = useTheme();
       const styles = getInfoBoxStyles(theme, severity);
       const wrapperClassName = branded ? cx(styles.wrapperBranded, className) : cx(styles.wrapper, className);
 
+      let dismiss;
+
+      if (dismissPersistenceId) {
+        const persistanceId = FEATUREINFOBOX_PERSISTENCE_ID_PREFIX.concat(dismissPersistenceId);
```
Clarity-89, line 61:
```suggestion
        const persistenceId = FEATUREINFOBOX_PERSISTENCE_ID_PREFIX.concat(dismissPersistenceId);
```

packages/grafana-ui/src/components/InfoBox/InfoBox.tsx
```diff
@@ -24,24 +27,61 @@ export interface InfoBoxProps extends Omit<React.HTMLAttributes<HTMLDivElement>,
...
+
+        const [dismissed, setDismissed] = useLocalStorage(persistanceId, { isDismissed: false });
```
Clarity-89, line 63: I think this will violate the rules of hooks as `useLocalStorage` is called conditionally here. We have a lint rule for this in the main codebase but it's currently disabled in grafana-ui.

public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -30,8 +32,9 @@ import { TransformationOperationRows } from './TransformationOperationRows';
 import { TransformationsEditorTransformation } from './types';
 import { PanelNotSupported } from '../PanelEditor/PanelNotSupported';
 import { AppNotificationSeverity } from '../../../../types';
+import theme from '@grafana/ui/src/themes/default';
```
Clarity-89, line 35: Unnecessary import.

public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -208,9 +211,16 @@ export class TransformationsEditor extends React.PureComponent<TransformationsEd
...
   renderNoAddedTransformsState() {
     return (
-      <VerticalGroup spacing={'lg'}>
+      <>
         <Container grow={1}>
-          <FeatureInfoBox title="Transformations" url={getDocsLink(DocsId.Transformations)}>
+          <FeatureInfoBox
+            title="Transformations"
+            className={css`
+              margin-bottom: ${theme.spacing.lg};
```
Clarity-89, line 219: When `withTheme` is used, `theme` is available on the props.

packages/grafana-ui/src/components/InfoBox/InfoBox.tsx
```diff
@@ -24,24 +27,61 @@ export interface InfoBoxProps extends Omit<React.HTMLAttributes<HTMLDivElement>,
...
+        onDismiss,
+        dismissPersistenceId,
+        severity = 'info',
+        ...otherProps
+      },
+      ref
+    ) => {
       const theme = useTheme();
       const styles = getInfoBoxStyles(theme, severity);
       const wrapperClassName = branded ? cx(styles.wrapperBranded, className) : cx(styles.wrapper, className);
```
Clarity-89, line 56:
```suggestion
      const wrapperClassName = cx(branded ? styles.wrapperBranded : styles.wrapper, className);
```

Clarity-89 (commented):
Good job! Left a few comments.

packages/grafana-ui/src/components/InfoBox/InfoBox.tsx
oscarkilhed: Ah, yeah I was having some issues with this approach and useLocalStorage. If dismissPersistenceId isn't set it would create an entry in localstorage with an empty string as the key. I had this suspicion that moving the persistence stuff into a separate component that wraps the info boxes might be a better idea. I think I'll try that out instead.

packages/grafana-ui/src/components/InfoBox/InfoBox.story.tsx
```diff
@@ -77,3 +80,44 @@ export const featureInfoBox = () => {
     </div>
   );
 };
+
+export const dismissableFeatureInfoBox = () => {
+  const { containerWidth, severity, title, url } = getKnobs();
```
Clarity-89, line 85: Would it be a good idea to move this story to use controls at the same time?

Clarity-89 (approved):
Looks good! 🎉

torkelo:
@dprokop sorry missed reviewing this, not sure I agree that a grafana-ui component should keep this state & store this in local storage, that should be up to the app to control where this state is stored. 

For example, this should mostly use the user help flags feature and that is stored in the database (and requires an API call to update), that way we store dismissing info boxes on the user, not in local storage. 

Having a separate component DismissableFeatureInfoBox also feels not optimal to me there is no new component here just storage logic, to me this should just be an onDismiss on the InfoBox component (which it already has). Then the storage could be done as a hook or something in the app that uses this.

dprokop:
> @dprokop sorry missed reviewing this, not sure I agree that a grafana-ui component should keep this state & store this in local storage, that should be up to the app to control where this state is stored.
> For example, this should mostly use the user help flags feature and that is stored in the database (and requires an API call to update), that way we store dismissing info boxes on the user, not in local storage.
> Having a separate component DismissableFeatureInfoBox also feels not optimal to me there is no new component here just storage logic, to me this should just be an onDismiss on the InfoBox component (which it already has). Then the storage could be done as a hook or something in the app that uses this.

@torkelo I missed this one too, no worries. I agree with you that it's app responsibiility to define the persisted storage, rather than then library.

Me and @oscarkilhed gonna talk this through.

PR #30730, opened 2021-01-29
Variables: Fixes so text format will show All instead of custom all value

### LINKED ISSUE #30685: Variables: Value with custom all value does not show the text representation `All` when used in panel title
**What happened**:
when rendering variable on panel title it renders variable value

we were expecting it to render variable text. This was the case prior version 7.x

create a variable with `all` option
add custom value for all with `_some_text_`
set title as `abc - ${app_type}`

we expect the title to show up as `abc - all`
but we see `abc - _some_text_`

we have tried using `abc - ${app_type:raw}` & `abc - ${app_type:text}`

**Environment**:
- Grafana version: 7.3.7
- Data source type & version: NA
- OS Grafana is installed on:
- User OS & Browser: Mac OSX, Brave, chrome
- Grafana plugins:
- Others:

hugohaggmark:
This PR changes so whenever you have a custom `all value` in a variable and use the `:text` formatting for a variable it will return `All` as `text` and custom `all value` as `value`.

Fixes #30685

### DIFF
--- a/public/app/features/templating/formatRegistry.ts
+++ b/public/app/features/templating/formatRegistry.ts
@@ -1,7 +1,8 @@
 import kbn from 'app/core/utils/kbn';
-import { Registry, RegistryItem, VariableModel, textUtil, dateTime } from '@grafana/data';
-import { map, isArray, replace } from 'lodash';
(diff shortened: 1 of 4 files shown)


torkelo (approved):
Looks good!

PR #30732, opened 2021-01-29
Datasource: Use json-iterator configuration compatible with standard library

marefr:
This will make sure that any map keys in JSON is ordered.

Fixes #

Seems like #30250 removed ordering of map keys in /api/ds/query responses. Using the standard lib JSON encoder per defaults order the map keys for you. With ordered map keys in responses we would have a stable order of the results.

First fixed in #29916, but opened this to include in 7.4 stable.


wbrowne (approved):
Wonder if it makes sense to extract `jsoniter.ConfigCompatibleWithStandardLibrary` in a var block with a comment, just so it's clear why this is the way it is.

PR #30742, opened 2021-01-29
TraceViewer: Fix show log marker in spanbar

### LINKED ISSUE #27885: Mark log/event annotations on the Span bar in Trace viewer
**What would you like to be added** / **Why is this needed**:

In the Jaeger UI, each event / log of a span is denoted by a vertical line on the span bar - helping visualise when the event occurred in the timeline. This is missing in the Grafana Trace Viewer.
Screenshot 2020-09-29 at 4 17 03 PM

zoltanbedi:
Fixes #27885


packages/jaeger-ui-components/src/common/CopyIcon.test.js
```diff
@@ -52,19 +53,19 @@ describe('<CopyIcon />', () => {
     expect(copySpy).toHaveBeenCalledWith(props.copyText);
   });
 
-  it('updates state when tooltip hides and state.hasCopied is true', () => {
+  it.skip('updates state when tooltip hides and state.hasCopied is true', () => {
```
Elfo404, line 56: is this skip wanted?

packages/jaeger-ui-components/src/common/CopyIcon.test.js
```diff
@@ -52,19 +53,19 @@ describe('<CopyIcon />', () => {
...
     expect(wrapper.state().hasCopied).toBe(false);
 
     const state = wrapper.state();
-    wrapper.find(UITooltip).prop('onVisibleChange')(false);
+    wrapper.find(Tooltip).prop('onVisibleChange')(false);
     expect(wrapper.state()).toBe(state);
   });
 
-  it('persists state when tooltip opens', () => {
+  it.skip('persists state when tooltip opens', () => {
```
Elfo404, line 66: is this skip wanted?

packages/jaeger-ui-components/src/common/CopyIcon.tsx
```diff
@@ -78,20 +80,18 @@ export default class CopyIcon extends React.PureComponent<PropsType, StateType>
   render() {
     const styles = getStyles();
     return (
-      <UITooltip
-        arrowPointAtCenter
-        mouseLeaveDelay={0.5}
-        onVisibleChange={this.handleTooltipVisibilityChange}
+      <Tooltip
+        // onVisibleChange={this.handleTooltipVisibilityChange}
```
Elfo404, line 84: I know there's a big TODO on top of this component but Is this a leftover?

zoltanbedi: Yes. This component was not showing before and the deeplinking feature it has is not working right now. This will be fixed in a separate pr.

packages/jaeger-ui-components/src/TraceTimelineViewer/ReferencesButton.tsx
```diff
@@ -16,10 +16,11 @@ import React from 'react';
 import { css } from 'emotion';
 import NewWindowIcon from '../common/NewWindowIcon';
 import { TraceSpanReference } from '@grafana/data';
-import { UITooltip, UIDropdown, UIMenuItem, UIMenu, TooltipPlacement } from '../uiElementsContext';
+import { UIDropdown, UIMenuItem, UIMenu } from '../uiElementsContext';
 
 import ReferenceLink from '../url/ReferenceLink';
 import { createStyle } from '../Theme';
+import { Tooltip } from '@grafana/ui';
```
aocenas, line 23: Why did we change from `UITooltip` to `Tooltip`? The point of using `UITooltip` was to not tie this to grafan/ui in any way, in case we still wanted to make this into a separate project (which may not happen but still). So I wonder if there were some technical issue with the approach of using injected components?

zoltanbedi: Yes I changed it because the tooltip was not showing for me for some reason when used with the wrapper.

packages/jaeger-ui-components/src/common/CopyIcon.test.js
zoltanbedi: yes

packages/jaeger-ui-components/src/common/CopyIcon.test.js
zoltanbedi: yes

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -161,7 +161,7 @@ function SpanBar(props: TInnerProps) {
     <div
       className={cx(styles.wrapper, className)}
       onClick={onClick}
-      onMouseOut={setShortLabel}
+      onMouseLeave={setShortLabel}
```
aocenas, line 164: Out of curiosity why this had to be changed?

aocenas (commented):
Code looks good but there are some visual issues
In dark mode the log marker should probably be some light color to be more visible:
Screenshot from 2021-02-19 15-03-27

The tooltip contents have very little visibility in light theme:
Screenshot from 2021-02-19 15-01-05

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
zoltanbedi: I had a problem with the tooltip not showing. And after changing that to onMouseLeave it showed up.

From [w3schools](https://www.w3schools.com/jsref/event_onmouseleave.asp)
> Tip: The onmouseleave event is similar to the onmouseout event. The only difference is that the onmouseleave event does not bubble (does not propagate up the document hierarchy).

zoltanbedi:
Fixed the popover/log marker color problem.

Dark theme:
Screenshot 2021-02-22 at 17 39 40

Light theme:
Screenshot 2021-02-22 at 17 40 44

aocenas (approved):
Nice work, looks great now.

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -65,14 +65,14 @@ const getStyles = createStyle(() => {
...
-      background-color: rgba(0, 0, 0, 0.5);
+      background-color: ${theme.colors.text};
       cursor: pointer;
       height: 60%;
       min-width: 1px;
       position: absolute;
       top: 20%;
       &:hover {
-        background-color: #000;
+        background-color: ${theme.colors.textSemiWeak};
```
torkelo, line 75: Why using text color for a background?

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -65,14 +65,14 @@ const getStyles = createStyle(() => {
     `,
     logMarker: css`
       label: logMarker;
-      background-color: rgba(0, 0, 0, 0.5);
+      background-color: ${theme.colors.text};
```
torkelo, line 68: This looks wrong , use bg1 or bg2 or bg3 for backgrounds

torkelo (changes requested):
The theme variable choice look strange

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
torkelo: or are these text-like elements?

zoltanbedi: not really. I just needed a color that is the reverse of bgcolor. so dark in light theme and light in dark theme.

torkelo: can you show what elements in the UI is the logMarker? maybe border2 can work as an alternative

torkelo (approved):

packages/jaeger-ui-components/src/TraceTimelineViewer/SpanBar.tsx
zoltanbedi: These are the log markers
Screenshot 2021-02-23 at 16 21 59

border2 would be light for light which wouldn't work.

PR #30745, opened 2021-01-29
Modals: Fix closing of modals with escape key

### LINKED ISSUE #30592: Modal: Esc key no longer closes modals
I noticed I can no longer close modals with the Esc key.

jackw:
This PR adds a listener to the grafana ui modal to handle closing by pressing the Esc key. 

Fixes #30592

Initial testing suggests this works but I'm not sure if this is the right direction or if we should be making use of the `KeybindingSrv` class to handle this within the Grafana app.

### DIFF
--- a/packages/grafana-ui/src/components/Modal/Modal.tsx
+++ b/packages/grafana-ui/src/components/Modal/Modal.tsx
@@ -40,7 +40,7 @@ export function Modal(props: PropsWithChildren<Props>): ReturnType<FC<Props>> {
     }
   }, [propsOnDismiss]);
 
-  const onEscKey = (ev: KeyboardEvent): void => {
+  const onEscKey = (ev: KeyboardEvent) => {
     if (ev.key === 'Esc' || ev.key === 'Escape') {
       onDismiss();
     }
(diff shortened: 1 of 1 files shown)


packages/grafana-ui/src/components/Modal/Modal.tsx
```diff
@@ -40,6 +40,23 @@ export function Modal(props: PropsWithChildren<Props>): ReturnType<FC<Props>> {
     }
   }, [propsOnDismiss]);
 
+  const onEscKey = (ev: KeyboardEvent): void => {
```
Clarity-89, line 43:
```suggestion
  const onEscKey = (ev: KeyboardEvent) => {
```
I think it's better to infer as much as possible, so it's less type maintenance.

Clarity-89 (approved):
Looks good! I think `KeybindingSrv` is used for global events, whereas here `Esc` key should work per specific Modal component.

dprokop (approved):
Looks good. One thing to consider is if this behaviour should not be controlled via property, i.e. `closeOnEscape` or sth similar.

Issue #30365, opened 2021-01-18
TimeSeriesPanel: Add support for transform negative


simPod:
**What would you like to be added**:

In current Graph panel, I'm able to take a series, put it in a stack and flip it

![image](https://user-images.githubusercontent.com/327717/104948377-7b76b400-59bd-11eb-9e22-743dca11c82a.png)

It is not possible in new Timeseries panel now

**Why is this needed**:

Let's say we have two timeseries representing `in` and `out` directions and want to show them in one graph. Therefore, it makes sense to use both sides of axis Y, thus flipping one direction to the other side.

torkelo:
@leeoniya think you added support for this in uPlot?

leeoniya:
not yet, we've only finished `fillBelowTo` so far: https://github.com/grafana/grafana/pull/30268

stacking support will be built on top of this. we're still gonna need to figure out how the two will interact, i'm thinking they'll have to be mutually exclusive.

leeoniya:
@simPod can you attach the data you're working with?

if you have an `in` series, and an `out` series, then you just need to flip the out series, but this isn't stacking related, right?

simPod:
@leeoniya I tried to simplified the use case, I might have `in a` `in b` `out a` `out b` :)

This is what I'm working with through https://github.com/simPod/GrafanaJsonDatasource. I can generate second series if it helps

<details>
<summary>Dataset</summary>
```json
[...]

Issue #7949, opened 2017-03-27
BarChart: Support rendering some series as lines


ChamMach:
Please include this information:
- What Grafana version are you using ? 4.x
- What datasource are you using ? ElasticSearch.
- What OS are you running grafana on ? Docker (grafana/grafana)
- What did you do ? Trying to stack result of request with X-Axis mode Series.
- What was the expected result? Something like that : 

![5](https://cloud.githubusercontent.com/assets/23166349/24347256/f9c15f58-12d7-11e7-9f6f-f45b48e03a47.png)

- What happened instead ? Something like that : 

![6](https://cloud.githubusercontent.com/assets/23166349/24347289/1777a638-12d8-11e7-9166-a75b713b44d6.png)

As you can see, points are not "stacking" on bars. For each new query (like query with points), Grafana create graph at the right (even if I use Group by : Date Histogram : "@timestamp" or something like that). 
So, I want the possibility to "stacking" result of query for graphs with x-axis mode series like x-axis mode date : 

![7](https://cloud.githubusercontent.com/assets/23166349/24347556/fd8bfb4c-12d8-11e7-94d2-c40f2ca8c0a8.png)

OR the possibility to change the "dimension" of bars.

Thank you and have a good day.

luiscruz:
Any update on this?

rentes:
Hi everyone, I'm not sure what to do to speed up this feature release, except saying that it would be of tremendous help to have it. I too want to stack X-axis by series, not by time. What can I do to have this on my grafana charts? Thank you.

jeliasson:
One would expect that stacking of metrics would be a native feature by now. When thinking of it, it does not seem like a big challenge. However, would really love to see this feature in the future. Thanks and keep it up!

Deltablox:
@jeliasson i agree... 2.5 years after opening that feature request and nothing happened until now.
Feelsbadman.
I would also really love to see this feature soon. It would be a nice improvement

ChamMach:
Hi everyone ! 

Glad to see that I'm not the only interested for this feature. Even 2 years later.

@torkelo the team have any plan to add this feature in the future ?

Thank's.

MattGill98:
+1

trierra:
+1

kvannorel:
+1

mohdhallal:
+1

Psyf:
+1

ThomasMadland:
+1

jceddy:
+1

[...]

PR #30749, opened 2021-01-29
GraphNG: stacking

dprokop:
See Panel Tests - Graph NG - stacking for tests (data set taken from @leeoniya demo)

Important change apart from stacking is the UPlotChart API change- data is now AlignedData. dataFrame property left for TooltipPlugin compatibility, but to be removed going forward to keep UPlotChart as Grafana-agnostic as possible 

TODO:
- [x] make bands creation consider field config
- [x] update components that use GraphNG (i.e. Sparkline)
- [x] migration
- [x] get rid of DataFrame dependency from UPlotChart (followup)
- [ ] Documentation (followup PR)
- [ ] percent stacking (TODO in a separate PR)


packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -77,6 +82,13 @@ export const GraphNG: React.FC<GraphNGProps> = ({
     return outerJoinDataFrames({ frames: data, joinBy: fields.x, keep: fields.y, keepOriginIndices: true });
   }, [data, fields]);
 
+  const plotData = useMemo(() => {
+    if (!frame) {
+      return null;
+    }
+    return preparePlotData(frame, stacking) as AlignedData;
```
dprokop, line 89: moved out from UPlotChart

packages/grafana-ui/src/components/GraphNG/GraphNG.tsx
```diff
@@ -110,7 +122,7 @@ export const GraphNG: React.FC<GraphNGProps> = ({
   const configRev = useRevision(frame, compareFrames);
 
   const configBuilder = useMemo(() => {
-    const builder = new UPlotConfigBuilder();
+    const builder = new UPlotConfigBuilder(stacking.enable);
```
dprokop, line 125: Not sure the best approach, maybe builder.enableStacking() would align better with the builder's api?

packages/grafana-ui/src/components/GraphNG/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+    return frame.fields.map((f) => f.values.toArray());
+  }
+
+  // TODO: percent stacking
+  const acc = Array(frame.fields[0].values.length).fill(0);
+  const result = [];
+  for (let i = 1; i < frame.fields.length; i++) {
+    for (let j = 0; j < frame.fields[i].values.length; j++) {
+      const v = frame.fields[i].values.get(j);
+      acc[j] += v === null || v === undefined ? 0 : +v;
```
leeoniya, line 31: btw, `== null` covers both `null` and `undefined`, and only those two. (TS is happy with this as well).

ryantxu: I had no idea about this either -- eslint trains us to use `===` everywhere.

This revelation lead to us supporting `gaps` and `connected` with `null` and `undefined` -- greatly simplifying/removing the whole `isGap()` business.

packages/grafana-ui/src/components/GraphNG/types.ts
```diff
@@ -18,3 +18,11 @@ export interface GraphNGLegendEvent {
   fieldIndex: DataFrameFieldIndex;
   mode: GraphNGLegendEventMode;
 }
+
+/**
+ * @alpha
+ */
+export interface StackingOptions {
```
ryantxu, line 25: Lets reuse the config from:
https://github.com/grafana/grafana/blob/master/packages/grafana-ui/src/components/BarChart/types.ts#L8

dprokop: done!

leeoniya:
works pretty well from an initial run-through :+1: 

looks like legend toggle is busted?

additional observations: leaving grafana's defaults and simply enabling stacking can result in some pretty misleading graphs. i've generally also had to set `min` or `softMin` to 0, and also enable some level of fill opacity. without any fill it's impossible to know that there is stacking going on and without 0, the stack is from some arbitrary baseline. we've had the same oddness happen in the bar chart panel, so we set a default softMin of 0 there, to avoid bars that start at 50 and end at 60. would be great if it's possible to recommend or auto-set some defaults when toggling stacking.

leeoniya:
@dprokop also do you think this PR will address splitting positive and negative stacks? see https://github.com/grafana/grafana/issues/30365.

this involves splitting series with all-negatives and stacking them independently and inverted from all-positives series.

torkelo:
I think the stacking option needs a field-level option to support the stacking groups (that can be defined via overrides)

leeoniya:
> I think the stacking option needs a field-level option to support stacking groups (that can be defined via overrides)

i'm not sure i've ever seen a timeseries chart with more than 1 stacking group for positive and 1 for negative values, or a single group that logically combined both. do you have any pictures that illustrate how you envision anything other than just splitting positive and negative series?

leeoniya:
maybe you're talking about something like https://github.com/grafana/grafana/issues/7949,  where you want some as unstacked line series, but some others as stacked bar?

torkelo:
I was mainly thinking on negative and positive stacking but might be some other scenarios

leeoniya:
so, from what i understand, the plan is to split first by `drawStyle` && (`lineInterpolation` or `barAlignment`), and then by all-positive and all-negatives. this would allow stacking e.g. left-aligned-bars without stacking line/stepAfter. if we do this automatically, there's no need to manually configure stacking groups, as there's no reasonable way to represent multiple stacking groups with identical drawstyles and alignments or lineInterpolations. (probably over several PRs?).

dprokop:
> @dprokop also do you think this PR will address splitting positive and negative stacks? see #30365. 

Yes, I'll resume the work soon :)

packages/grafana-ui/src/components/uPlot/config/UPlotConfigBuilder.ts
```diff
@@ -118,16 +122,22 @@ export class UPlotConfigBuilder {
...
+        config.bands = this.bands;
+        const keepFill = new Set<number>();
+        for (const b of config.bands) {
+          keepFill.add(b.series[0]);
+        }
+
+        for (let i = 1; i < config.series.length; i++) {
+          if (!keepFill.has(i)) {
+            config.series[i].fill = undefined;
+          }
```
dprokop, line 140: This makes the stacking and fillBelowTo options mutually exclusive. We talked this through with @leeoniya and this makes most sense as it's most probably super uncommon to have these two combined.

packages/grafana-ui/src/components/uPlot/plugins/TooltipPlugin.tsx
```diff
@@ -58,10 +58,11 @@ export const TooltipPlugin: React.FC<TooltipPluginProps> = ({ mode = 'single', t
 
         // when interacting with a point in single mode
         if (mode === 'single' && originFieldIndex !== null) {
-          const field = otherProps.data[originFieldIndex.frameIndex].fields[originFieldIndex.fieldIndex];
+          const field = graphContext.alignedData.fields[focusedSeriesIdx!];
```
dprokop, line 61: @leeoniya having the aligned data coming into the panel would actually help reducin this context consumption. We could rely on the single data prop rather than distinguishing between aligned and not aligned data.

packages/grafana-ui/src/components/uPlot/utils.ts
```diff
@@ -51,12 +54,43 @@ export function preparePlotData(frame: DataFrame, ignoreFieldTypes?: FieldType[]
...
       continue;
     }
+
+    if (f.config.custom?.stackingMode !== StackingMode.None && f.config.custom?.stackingGroup) {
+      if (!stackingGroups[f.config.custom.stackingGroup]) {
+        stackingGroups[f.config.custom.stackingGroup] = [result.length];
+      } else {
+        stackingGroups[f.config.custom.stackingGroup].push(result.length);
+      }
+    }
```
dprokop, line 67: this map is vreated in two places currently - one for results stacking(here) and one for bands creation.

hugohaggmark (commented):
I did some manual testing and to be honest I don't know much about how this should work so I concentrated my efforts on diffs. These are 2 diffs between old/new panel that I found that might be issues:

Filtering on series in old panel removes the stacking:
bug1

Filtering on series with bars in new panel shows hollow bars:
bug2

Will also have a run through the code now.

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -231,4 +249,10 @@ export const graphFieldOptions = {
     { label: 'Hue', value: GraphGradientMode.Hue },
     //  { label: 'Color scheme', value: GraphGradientMode.Scheme },
   ] as Array<SelectableValue<GraphGradientMode>>,
+
+  stacking: [
+    { label: 'None', value: StackingMode.None },
+    { label: 'Standard', value: StackingMode.Standard },
+    { label: 'Percent(TODO)', value: StackingMode.Percent },
```
hugohaggmark, line 256: Maybe remove this until we have a working implementation?

packages/grafana-ui/src/components/GraphNG/utils.ts
```diff
@@ -219,8 +223,28 @@ export function preparePlotConfigBuilder(
...
+    const groups = Object.keys(stackingGroups);
+    for (let i = 0; i < groups.length; i++) {
+      const group = stackingGroups[groups[i]];
+      for (let j = group.length - 1; j > 0; j--) {
+        builder.addBand({
+          series: [group[j], group[j - 1]],
+        });
+      }
+    }
+  }
```
hugohaggmark, line 247: Maybe we should extend or add tests for this? If I was ever to change things in this code I would feel a lot safer if there were tests.

packages/grafana-ui/src/components/uPlot/config/UPlotConfigBuilder.ts
hugohaggmark, line 140: Is it possible to add tests for this component?

packages/grafana-ui/src/components/uPlot/utils.ts
```diff
@@ -33,8 +34,10 @@ export function buildPlotConfig(props: PlotProps, plugins: Record<string, PlotPl
 }
 
 /** @internal */
+
 export function preparePlotData(frame: DataFrame, ignoreFieldTypes?: FieldType[]): AlignedData {
   const result: any[] = [];
+  const stackingGroups: Record<string, number[]> = {};
```
hugohaggmark, line 40: Nit: maybe Map would be a better choice here?

hugohaggmark (commented):
Skilled work on this PR and sorry for taking time to review it. I found some tiny things that you might want to consider.

packages/grafana-ui/src/components/uPlot/config.ts
```diff
@@ -175,6 +175,23 @@ export interface HideableFieldConfig {
...
+  Percent = 'percent',
+}
+
+/**
+ * @alpha
+ */
+export interface StackingConfig {
+  stackingMode?: StackingMode;
+  stackingGroup?: string;
+}
```
torkelo, line 193: I would try to not have have two stacking options, instead combine these into a single stacking (string) prop: 

stacking?: string;

valid values would be undefined (disabled), 'normal'  = normal stacking, we can add 'percent' as a special value later.  Or if any other value it's a custom stacking group.  

A bit similar to https://api.highcharts.com/highcharts/plotOptions.series.stacking

torkelo: Or a separate option is not so bad, hm.. tricky

torkelo (approved):
Looks good! Awesome job!

leeoniya:
:heart:

MichaelTendoSsemwanga:
This was an amazing successful feat. Apologies if this is not the correct way to report a bug, but I thought it convenient to note the issue here since you may almost instantly recognise the problem. As I have noted in this issue: https://github.com/grafana/grafana/issues/36151, Stacking negative values does not paint the fill.

PR #30752, opened 2021-01-29
Chore: Remove warning when calling resource

marefr:
Fixes so that the warning _WARN Failed to close stream logger=plugins.backend err="cannot close a closed stream"_ is not logged every time calling a resource.

Fixes #


wbrowne (approved):
LGTM

PR #30784, opened 2021-02-01
Panels: Fixes so panels are refreshed when scrolling past them fast

### LINKED ISSUE #30300: Panels don't load if scrolling down too fast
How to reproduce:
1. Open e.g. https://play.grafana.org/d/avzwehmz/demo-wind-farm?orgId=1
2. Click browser reload button
3. Scroll down right away while the page is still loading
4. Scroll back up
5. Some panels have not been loaded and they remain unloaded

Expected:
Panels are loaded when I scroll back up

Tested on:
MacOS Mojave 10.14.6
Chrome 87.0.4280.141
Grafana 7.3.7

recorded

hugohaggmark:
It seems like scrolling by a panel real fast can in some very rare edge cases cause the loading of data to be triggered but at the same time the `isInView` is false because of the scrolling. This PR attempts to solve this by adding `refreshWhenInView` to state so the next time the panel is in view the loading will reoccur.

I also added a test that proves this behavior.

Fixes #30300

### DIFF
--- a/public/app/features/dashboard/dashgrid/PanelChrome.test.tsx
+++ b/public/app/features/dashboard/dashgrid/PanelChrome.test.tsx
@@ -1,0 +1,80 @@
+import React, { FC } from 'react';
+import { ReplaySubject } from 'rxjs';
+import { act, render, screen } from '@testing-library/react';
+import { getDefaultTimeRange, LoadingState, PanelData, PanelPlugin, PanelProps } from '@grafana/data';
+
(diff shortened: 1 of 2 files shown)


dprokop (approved):
Seems to work fine, and the test is 👌

hugohaggmark:
> Seems to work fine, and the test is 👌

Thank you, a lot of mocks to prove my scenario, but I got it to work :)

PR #30796, opened 2021-02-01
GraphNG: improve behavior when switching between solid/dash/dots

ryantxu:
When switching between dash/dots, we also need to update the spacing -- this PR updates the segment spacing when the mode changes


leeoniya:
LGTM :rocket:

leeoniya (approved):

PR #30797, opened 2021-02-01
Variable editor: Add missing callback dependency

Clarity-89:

`SelectionOptionsEditor` was missing variable dependency in `onMultiChanged` function. As a result, after changing variable name, toggling multi value didn't work, since `onMultiChanged` was receiving the old vale of `variable`. 


hugohaggmark (approved):
Great catch! 🎉

PR #30802, opened 2021-02-01
Transforms: allow boolean in field calculations

ryantxu:
This will will explicitly support boolean values in the calculation transform.

This has always kinda worked (true * 6 = 6, false * 6 = 0), but this makes it easier to see and visualize in the UI.


ryantxu:
cc @bymattoa

nenadilic84:
I've tested this with the Grafana v7.4.3 (010f20c1c8) on play.grafana.org and it didn't work as I would expect.
1. There was no autocompletion of the series with the boolean type
2. Typing in the series manually didn't show the calculated values and didn't create the new series

Screenshot 2021-02-25 at 21 05 02

ryantxu:
@nenadilic84 What is your original query?  are you sure the column type is actually boolean?  the behavior I see above would happen with string column with values "true" and "false"

PR #30806, opened 2021-02-02
BarChart: inside-align strokes, upgrade uPlot to 1.6.4.

leeoniya:
prior to this PR, the BarChart stroke was center-aligned to the bar edge, so if it was wide, it would run into neighboring bars and the value text. this aligns the stroke inside (see below). also, the bars would become blurry at odd stroke widths due to the way uPlot tries to align line paths, this adds a new opt-out called `pxAlign`, so we can maintain bar crispness at all stroke widths.

uPlot gets bumped to 1.6.4. nothing too risky in this changelog, but please test all the panels just in case!

https://github.com/leeoniya/uPlot/releases/tag/1.6.4

the bump also fixes a bug that affects fillBelowTo for bars and dominik's stacking PR, where the bar strokes were not clipped to the proper level (the fill was clipped correctly).

image

### DIFF
--- a/packages/grafana-ui/package.json
+++ b/packages/grafana-ui/package.json
@@ -72,7 +72,7 @@
     "react-transition-group": "4.4.1",
     "slate": "0.47.8",
     "tinycolor2": "1.4.1",
-    "uplot": "1.6.3"
+    "uplot": "1.6.4"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "16.0.0",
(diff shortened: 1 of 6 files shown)


torkelo:
Nice, this might fix the bars at the edge of canvas was had a border that looked thinner than the others

torkelo (approved):

leeoniya:
thanks!

PR #30808, opened 2021-02-02
Chore: pad unknown values with undefined

ryantxu:
MutableVectors currently pad unknown values with null.  When this was developed, there was no semantic difference between null and undefined.  With the new graph panel, `undefined` values will appear connected while `null` values are treated as gaps.


hugohaggmark (approved):
LGTM!

PR #30810, opened 2021-02-02
Variables: Clears drop down state when leaving dashboard

### LINKED ISSUE #30786: Variables drop-down not clearing state when navigating away from the dashboard
**What happened**:
Dashboard variable drop-down doesn't clear state when navigating away from the dashboard.

The dropdown should be collapsed next time the dashboard is opened.

1. Open dashboards with multi selection variables.
2. Click on the variable drop-down to open the list and select one of the options.
3. Navigate out to a different view (data source for example) without confirming the selection.
4. Navigate back to the previous dashboard.
5. 
variables

**Environment**:
- Grafana version: N/A
- User OS & Browser: N/A
- Grafana plugins: N/A
- Others: N/A

hugohaggmark:
When leaving dashboards, we forgot to clear the Redux state for Picker and Editor. This PR attempts to remedy this. 

Fixes #30786

### DIFF
--- a/public/app/features/variables/editor/reducer.test.ts
+++ b/public/app/features/variables/editor/reducer.test.ts
@@ -4,6 +4,7 @@ import {
   changeVariableEditorExtended,
   changeVariableNameFailed,
   changeVariableNameSucceeded,
+  cleanEditorState,
   clearIdInEditor,
   initialVariableEditorState,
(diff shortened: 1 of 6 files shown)


Clarity-89 (approved):
Works as expected, great job!

PR #30821, opened 2021-02-02
Grafana-ui: fixes no data message in Table component

### LINKED ISSUE #30545: New-table panel does not show "no data to show" when there is no data.
**What happened**:
The old table panel used to show the string "no data to show" when there was no data. 
The new table panel instead shows the column headings for the data, but does not show "no data to show" leaving some users confused if there is in fact any data or not. 
See screen shots. 
image (3)

The new-table panel to have a clear visual clue such as saying "no data to show" when there is no data.

**Environment**:
- Grafana version: 7.3.7 
- Data source type & version: Azure logs analytics. 
- User OS & Browser: n/a
- Grafana plugins: n/a
- Others:

jackw:
The Table component in grafana-ui is lacking a  "no data" message if the data passed in has no rows. This PR fixes this.

| Before | After |
| --- | --- |
| Screenshot 2021-02-02 at 11 19 01 | Screenshot 2021-02-02 at 11 19 43 |

Fixes #30545

### DIFF
--- a/packages/grafana-ui/src/components/Table/Table.tsx
+++ b/packages/grafana-ui/src/components/Table/Table.tsx
@@ -206,15 +206,19 @@ export const Table: FC<Props> = memo((props: Props) => {
               })}
             </div>
           )}
-          <FixedSizeList
-            height={height - headerHeight}
-            itemCount={rows.length}
-            itemSize={tableStyles.rowHeight}
-            width={'100%'}
-            style={{ overflow: 'hidden auto' }}
-          >
-            {RenderRow}
-          </FixedSizeList>
+          {rows.length > 0 ? (
+            <FixedSizeList
+              height={height - headerHeight}
+              itemCount={rows.length}
+              itemSize={tableStyles.rowHeight}
+              width={'100%'}
+              style={{ overflow: 'hidden auto' }}
+            >
+              {RenderRow}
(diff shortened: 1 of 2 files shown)


dprokop (approved):
This looks good but I have a question- @torkelo just curious why you want this to be part of the table component? Maybe instead we should handle this on the PanelChrome level as a matter of consistency?

torkelo:
@dprokop 

To me showing an info message in the table that there are no rows feels like a pretty clear Table component feature, not sure why this would be anywhere else. We show Tables in multiple places (drawer for example), and want this info message everywhere.

torkelo:
think handling this in PanelChrome is tricky as different visualizations might want to show fields and the no rows or no fields states differently

torkelo:
for example in a singlestat / gauge, the panel will render a gauge for each field and no data in each

dprokop:
So I consider two alternative situations actually:

1. There is no data in the response (PanelChrome no data info) - basically empty/no data frame - then this look/feel is OK 
2. There is a data frame returned- but with no rows(Table panel no info). In such scenario I would expect a header row being rendered an "no data" text shown, pretty much sth like this:
  image

torkelo:
@dprokop yes exactly, and i was under the impression that this was fixing scenario 2

PR #30822, opened 2021-02-02
CDN: Fixes cdn path when Grafana is under sub path

torkelo:
Fixes issues with the CDN change that added `/` prefix to assets paths even when no CDN was configured which made assets not load when Grafana is under sub path.


marefr (approved):
LGTM

PR #30827, opened 2021-02-02
InfluxDB: Add http configuration when selecting InfluxDB v2 flavor

### LINKED ISSUE #29303: Influx: There is not auth configuration when Flux language is selected
**What happened**:
Can't connect to influxdb 2.0 when using flux datasource

Somewhere to set https settings or certs

Add Flux datasource when influx is setup using TLS

**Environment**:
- Grafana version: 7.3.2
- Data source type & version: Influx Version 2.0.0 (37cc047)
- OS Grafana is installed on: ubuntu
- User OS & Browser: chrome

aocenas:
Fixes: https://github.com/grafana/grafana/issues/29303

Adds the HTTP default config section. Also adds basic InfluxDB2 dev env block for testing.

### DIFF
--- a/devenv/docker/blocks/influxdb2/docker-compose.yaml
+++ b/devenv/docker/blocks/influxdb2/docker-compose.yaml
@@ -14,7 +14,7 @@
       - influxdb
     image: quay.io/influxdb/influxdb:v2.0.3
     # Use these same configurations parameters in your telegraf configuration, mytelegraf.conf.
-    entrypoint: influx setup --bucket mybucket -t mytoken -o myorg --username=grafana --password=grafana --host=http://influxdb:8086 -f
+    entrypoint: influx setup --bucket mybucket -t mytoken -o myorg --username=grafana --password=grafana12345 --host=http://influxdb:8086 -f
       # Wait for the influxd service in the influxdb container has fully bootstrapped before trying to setup an influxdb instance with the influxdb_cli service.
     restart: on-failure:10
     depends_on:
@@ -38,4 +38,3 @@
       - ./docker/blocks/influxdb2/telegraf.conf:/etc/telegraf/telegraf.conf:ro
       - /var/log:/var/log
       - ../data/log:/var/log/grafana
-
--- a/devenv/docker/blocks/influxdb2/docker-compose.yaml
+++ b/devenv/docker/blocks/influxdb2/docker-compose.yaml
@@ -1,0 +1,41 @@
+  influxdb:
+    image: quay.io/influxdb/influxdb:v2.0.3
+    container_name: influxdb2
+    ports:
+      - '8086:8086'
+    environment:
+      INFLUXDB_REPORTING_DISABLED: 'true'
+    volumes:
+      - ./docker/blocks/influxdb2/influxdb.conf:/etc/influxdb/influxdb.conf
+
+  # Use the influx cli to set up an influxdb instance.
+  influxdb_cli:
+    links:
+      - influxdb
+    image: quay.io/influxdb/influxdb:v2.0.3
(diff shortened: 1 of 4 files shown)


ivanahuckova (approved):
Question - not necessarily to this PR, but when I want to add Flux data source, the basic auth toggle is already switched on, which I think is not necessary, as Flux data source works even if I turn it off. 

Awesome that you have added flux docker block. I wasn't able to test it, but that is probably just my problem with generally running docker blocks (I have to figure that out). 

I would still wait for @ryantxu to do a review, but from me - LGTM.

aocenas:
> the basic auth toggle is already switched on

Honestly not sure about that one it was already there in the code so did not change that. @ryantxu do you know if there was a purpose for it?

devenv/docker/blocks/influxdb2/docker-compose.yaml
```diff
@@ -0,0 +1,41 @@
...
+    volumes:
+      - ./docker/blocks/influxdb2/influxdb.conf:/etc/influxdb/influxdb.conf
+
+  # Use the influx cli to set up an influxdb instance.
+  influxdb_cli:
+    links:
+      - influxdb
+    image: quay.io/influxdb/influxdb:v2.0.3
+    # Use these same configurations parameters in your telegraf configuration, mytelegraf.conf.
+    entrypoint: influx setup --bucket mybucket -t mytoken -o myorg --username=grafana --password=grafana --host=http://influxdb:8086 -f
```
papagian, line 17:
```suggestion
    entrypoint: influx setup --bucket mybucket -t mytoken -o myorg --username=grafana --password=grafana12345 --host=http://influxdb:8086 -f
```
otherwise I get an error like:
`Error: failed to setup instance: password is too short`

aocenas: Thanks, seems like I forgot to push the update

public/app/plugins/datasource/influxdb/components/ConfigEditor.tsx
```diff
@@ -95,25 +86,14 @@ export class ConfigEditor extends PureComponent<Props> {
...
-              <Input
-                className="width-20"
-                value={options.url || ''}
-                placeholder="http://localhost:9999"
-                onChange={this.onUpdateInflux2xURL}
-              />
-            </div>
-          </div>
-        </div>
+        <DataSourceHttpSettings
```
ryantxu, line 89: How is this actually used?  I expect it would need to be passed to https://github.com/influxdata/influxdb-client-go

papagian (approved):
I've tested and AFAICT works great! I have tested without TLS and by selecting TLS Client Auth without filling the necessary fields and it complains as expected (`tls: failed to find any PEM data in certificate input`).
 It would also be nice to include a flux data source together with the other [gdev provisioned datasources](https://github.com/grafana/grafana/blob/master/devenv/datasources.yaml) by adding:
```
  - name: gdev-flux 
    type: influxdb
    access: proxy
    url: http://localhost:8086
    secureJsonData:
      token: mytoken
    jsonData:
      version: Flux
      organization: myorg
      defaultBucket: mybucket
```

ryantxu:
I don't get how this does anything?  are we using the frontend proxy for anything with flux?

public/app/plugins/datasource/influxdb/components/ConfigEditor.tsx
aocenas: We create the http client the same way for flux and then pass it to the client lib here https://github.com/grafana/grafana/blob/a2a6b9448ff0516a50738ac53fea28a3365d016f/pkg/tsdb/influxdb/flux/flux.go#L84. So I assume all the standard HTTP settings for the transport are respected.

PR #30829, opened 2021-02-02
Zipkin: Show success on test data source

### LINKED ISSUE #28838: Zipkin: test data source is not returning success message
**What happened**:

Adding Zipkin data source the success message is not shown.

Show success message when the data source is reached.

zoltanbedi:

Add zipkin example app to devenv
Migrate to fetch api in data source. Part of #27222

Fixes #28838


public/app/plugins/datasource/zipkin/datasource.ts
```diff
@@ -24,7 +24,7 @@ export class ZipkinDatasource extends DataSourceApi<ZipkinQuery> {
     super(instanceSettings);
   }
 
-  query(options: DataQueryRequest<ZipkinQuery>): Observable<DataQueryResponse> {
```
hugohaggmark, line 27: Why remove all the return types? I think we agreed that although TS can infer this, that it adds clarity?

hugohaggmark (approved):
Skilled work on this 🎉 , I just have one comment.

public/app/plugins/datasource/zipkin/datasource.ts
zoltanbedi: ohh I was not aware of that. I agree it adds clarity but I would enforce that "rule" (maybe we should add a rule) to public facing api. I like that ts can infer type for me so any time I change something I can see that it is breaking something or not without updating types.

public/app/plugins/datasource/zipkin/datasource.ts
```diff
@@ -35,31 +35,29 @@ export class ZipkinDatasource extends DataSourceApi<ZipkinQuery> {
...
     return query.query;
   }
 
-  private request<T = any>(apiUrl: string, data?: any, options?: Partial<BackendSrvRequest>): Observable<{ data: T }> {
-    // Hack for proxying metadata requests
-    const baseUrl = `/api/datasources/proxy/${this.instanceSettings.id}`;
+  private request<T = any>(apiUrl: string, data?: any, options?: Partial<BackendSrvRequest>) {
     const params = data ? serializeParams(data) : '';
-    const url = `${baseUrl}${apiUrl}${params.length ? `?${params}` : ''}`;
+    const url = `${this.instanceSettings.url}${apiUrl}${params.length ? `?${params}` : ''}`;
```
aocenas, line 54: Isn't this basically changing the access mode from proxy to direct? Before we were using the data source proxy with the `/api/datasources/proxy` but `this.instanceSettings.url` should contain direct url for the datasource.

aocenas (changes requested):

public/app/plugins/datasource/zipkin/datasource.ts
aocenas: Enforcing to public API could make sense. Personally I prefer the return types explicitly because otherwise it can be a harder to get to the type in IDE and also if you have multiple return types it is easy to get to state where you have `: TypeA | TypeB` return type which is valid and the resulting compilation errors can be misleading of what is wrong.

public/app/plugins/datasource/zipkin/datasource.ts
zoltanbedi: Nah. I was trying to figure out how but the url is the datasource proxy see 
Screenshot 2021-02-04 at 13 19 31

aocenas: Ah interesting didn't know we swap that for the proxy one

aocenas (approved):

PR #30854, opened 2021-02-03
Transformers: Add search to transform selection

ryantxu:
The list of transformations is getting pretty long and hard to pick.  This adds search on the initial page:

image


public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -209,6 +217,31 @@ class UnThemedTransformationsEditor extends React.PureComponent<TransformationsE
   };
 
   renderNoAddedTransformsState() {
+    const { search } = this.state;
+    let suffix: React.ReactNode | undefined = undefined;
```
jackw, line 221: Nit: Could use `null` here rather than the union?

public/app/features/dashboard/components/TransformationsEditor/TransformationsEditor.tsx
```diff
@@ -40,6 +42,7 @@ interface TransformationsEditorProps extends Themeable {
 interface State {
   data: DataFrame[];
   transformations: TransformationsEditorTransformation[];
+  search?: string;
```
jackw, line 45: nit: defaulting this to an empty string rather than `undefined` might make the code easier to reason with.

jackw (approved):
Works great! 🥇 

Left a couple of nits. I think this PR also needs milestone, project and labels adding.

torkelo:
Hmm feels some how duplicating effort in having search in initial view and in the picker select, and for them to be different , maybe it's time to skip the initial list and make that picker better?

ryantxu:
Yes -- I think we can drop the picker and have it show the same initial list+search

oddlittlebird:
What release will this be in? I don't see a milestone.

oddlittlebird:
Nice feature, but I do not see docs. Was Docs Squad just added for visibility? 

We will definitely want to mention this feature in the What's New and update the docs when it gets close to release.

ryantxu:
this would be in 7.5 --- docs was only added because of a merge :shrug:

dprokop (approved):
> Yes -- I think we can drop the picker and have it show the same initial list+search

Agree with that! Good improvement in general:)

oddlittlebird:
> this would be in 7.5 --- docs was only added because of a merge 🤷

Gah, hate it when that happens. Super confusing, lol. Carry on. :)

PR #30858, opened 2021-02-03
Variables: Adds queryparam formatting option

### LINKED ISSUE #21459: Templating: Links and multi valued variables that set variables via URL
Hello, 
I'm currently trying to embed defined variables, which can take several or all values, into custom html links of a panel. But the problem is that the available options:

- $var_name
- [[var_name]]
- ${var_name:glob}
- ${var_name:regex}
- ${var_name:pipe}
- ${var_name:csv}
- ${var_name:json}
- ${var_name:distributed}
- ${var_name:lucence}
- ${var_name:percentencode}
- ${var_name:glob}

do not allow correct use. 

The own links of grafana create for example the following link if selected more than once: 
.../?var_name=xy&var_name=yz
However, I cannot illustrate this with the above possibilities. 

Therefore I would like to ask if there is a possibility for it or if something like this can be created. 

The reason not to create links with the integrated possibility is the missing overview of many links. 
Therefore the whole thing should be sorted and displayed clearly using HTML.

With best regards

hugohaggmark:
This PR adds the `queryparam` formatting options for variables as `${foo:queryparam}`. 
This PR also changes the default suggested variable inserted text in Data Links from `var-${name}=${value}` to `${name:queryparam}` 

**Example**
| Variable Name | Current Variable Value | Result
| --- | --- | --- |
| foo | All | var-foo=All
| foo | A | var-foo=A
| foo | B | var-foo=B
| foo | A + B | var-foo=A&var-foo=B

Fixes #21459

### DIFF
--- a/docs/sources/variables/advanced-variable-format-options.md
+++ b/docs/sources/variables/advanced-variable-format-options.md
@@ -150,9 +150,9 @@ String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
 ```
 
-## Query Parameters
+## Query parameters
 
-Formats single- and multi-valued variables into their query parameter representation like `var-foo=value1&var-foo=value2`.
+Formats single- and multi-valued variables into their query parameter representation. Example: `var-foo=value1&var-foo=value2`
 
 ```bash
 servers = ["test1", "test2"]
--- a/docs/sources/variables/advanced-variable-format-options.md
+++ b/docs/sources/variables/advanced-variable-format-options.md
@@ -149,3 +149,13 @@ servers = ["test1", "test2"]
 String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
 ```
+
+## Query Parameters
+
+Formats single- and multi-valued variables into their query parameter representation like `var-foo=value1&var-foo=value2`.
+
+```bash
+servers = ["test1", "test2"]
+String to interpolate: '${servers:queryparam}'
+Interpolation result: "var-servers=test1&var-servers=test2"
+```
--- a/devenv/dev-dashboards/feature-templating/global-variables-and-interpolation.json
+++ b/devenv/dev-dashboards/feature-templating/global-variables-and-interpolation.json
@@ -27,14 +27,14 @@
         "overrides": []
       },
       "gridPos": {
-        "h": 16,
+        "h": 18,
         "w": 24,
         "x": 0,
         "y": 0
       },
       "id": 11,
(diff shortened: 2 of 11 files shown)


hugohaggmark:
@oddlittlebird @achatterjee-grafana this feature is new in Grafana 7.4.1 do we need to add a note about this in the docs?

torkelo (approved):
Looks really good!

oddlittlebird:
Will it be in the release tomorrow? If so, then we should mention it in the What's New now. 

If later, will it be included in the auto-generated release notes?

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -149,3 +149,13 @@ servers = ["test1", "test2"]
 String to interpolate: '${servers:text}'
 Interpolation result: "test1 + test2"
 ```
+
+## Query Parameters
```
oddlittlebird, line 153:
```suggestion
## Query parameters
```

docs/sources/variables/advanced-variable-format-options.md
```diff
@@ -149,3 +149,13 @@ servers = ["test1", "test2"]
...
+
+Formats single- and multi-valued variables into their query parameter representation like `var-foo=value1&var-foo=value2`.
```
oddlittlebird, line 155:
```suggestion
Formats single- and multi-valued variables into their query parameter representation. Example: `var-foo=value1&var-foo=value2`
```

osg-grafana:
> Will it be in the release tomorrow? If so, then we should mention it in the What's New now.
> 
> If later, will it be included in the auto-generated release notes?

7.4.1 is the target release version.

oddlittlebird (changes requested):

oddlittlebird:
> 7.4.1 is the target release version.

Yes, I saw that. Version != release date.  The questions stand.

torkelo:
> If later, will it be included in the auto-generated release notes?

yes will be part of the auto-generated release notes

docs/sources/variables/advanced-variable-format-options.md
hugohaggmark: Fixed!

docs/sources/variables/advanced-variable-format-options.md
hugohaggmark: Fixed!

oddlittlebird (approved):
Looks good

PR #30861, opened 2021-02-03
TestData: Fixes issue with forever loading state when all queries are hidden

### LINKED ISSUE #30646: Panel: Cannot disable test data source query when there is only a single query
**What happened**:
Kapture 2021-01-26 at 16 17 32

Single queries should also be capable of being disabled

**Environment**:
- Grafana version: Verified on master and 7.3.7
- Data source type & version: TestData DB
- OS Grafana is installed on: MacOS
- User OS & Browser: MacOS + Google Chrome v87
- Grafana plugins: TestData DB

torkelo:
Fixes #30646


hugohaggmark (approved):
LGTM! 🎉

PR #18580, opened 2019-08-15
SingleStat: use DataFrame results rather than TimeSeries/TableData

ryantxu:
fixes #18536

This tries to move as much logic out of the angular panel and use the same calculations as the new react panels.

TODO:
- [x] migrate field mapping calculations
- [x] get tests passing

Changes:
previously, when in table mode I think it assumed showing the first value.  Now it lets you pick a field and the reducer:
image

See also #18581

PR #30868, opened 2021-02-03
Panels: Remove value mapping of values that have been formatted #26763

### LINKED ISSUE #26763: Value mappings for durations is broken
**What happened**:
Table panel in a dashboard with durations in seconds.
Value mapping for example "range from 1 to 3" applies to:
- 1 second (ok)
- 2 seconds (ok)
- 3 seconds (ok)
- 1 minute (bad)
- 2 minutes (bad)
- 3 minutes (bad)
- 1 hour (bad)
- 2 hours (bad)
- 3 hours (bad)

Value mapping should apply to seconds only:
other values are only converted for presentation,
but the mapping should be intended for the unconverted values only.

Add a new pane to a new dashboard.
Select a SQL Server source and write this query:
`SELECT 1 as a, 2 as b, 60 as c, 120 as d, 3600 as e, 7200 as f, 0 as time`
Format as **table** (not as time serie).
In _Panel_ tab select **Table** visualization.
In _Field_ tab select Standard options > Unit > Time > **duration (s)** .
In _Field_ tab add a value mapping for value 2 with text "banana".
You will see something like this:
image
- column `b` is correct
- column `d` is incorrect and should display `2 minutes` instead of `banana`
- column `f` is incorrect and should display `2 hours` instead of `banana` 

**Environment**:
- Grafana version: 7.0.4
- Data source type & version: built-in SQL Server data source

oscarkilhed:
Value mappings are applied both before and after formatting a value. This leads to issues where a value mapping for the value 1, will also trigger on a value for 1000, if a unit that uses SI prefixes is used ( kilo, mega, giga, etc).

This PR removes the second pass over the value mappings.

However, this removes the ability to use a value mapping for the value 10, and have it match 0.1 with the unit Percent 0.0-1. However this use case is not very user friendly right now anyway as a value mapping with the value 10 will match both the raw value 0.1 and 10.

Other examples are if you use scientific notation you can no longer set the value mapping to 1e-1 and have it match for the raw value 0.1.

There's some hint of why this was added here: https://github.com/grafana/grafana/pull/18580/files/bc264f0c5fa4c2c4555cb76f972f5187052a1e59

Fixes #26763

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.test.ts
+++ b/packages/grafana-data/src/field/displayProcessor.test.ts
@@ -164,7 +164,7 @@ describe('Format value', () => {
     expect(instance(value).numeric).toEqual(1);
   });
 
-  it('should not map 1kw to the value for 1w', () => {
+  it('should not map 1kW to the value for 1W', () => {
     const valueMappings: ValueMapping[] = [{ id: 0, text: 'mapped', type: MappingType.ValueToText, value: '1' }];
     const value = '1000';
     const instance = getDisplayProcessorFromConfig({ decimals: 1, mappings: valueMappings, unit: 'watt' });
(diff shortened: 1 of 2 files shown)


ryantxu (approved):
LGTM

dprokop (approved):

PR #30880, opened 2021-02-03
Exemplars: Change CTA style

zoltanbedi:

Updated the cta in exemplars window to reflect the design changes.
Screenshot 2021-02-03 at 21 51 24


packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+import React from 'react';
+import { useStyles } from '../../themes';
+import { Icon } from '../Icon/Icon';
+import { DataLinkButton } from './DataLinkButton';
+
+type Props = {
+  links: Array<LinkModel<Field>>;
+};
+
+export function FieldLink({ links }: Props) {
```
aocenas, line 12: Should this be a FieldLinkList or something like that?

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return <DataLinkButton link={links[0]} />;
+  }
+
+  const externalLinks = links.filter((link) => link.target === '_blank');
+  const internalLinks = links.filter((link) => link.target === '_self');
+
+  return (
+    <>
+      {internalLinks.map((link, i) => {
+        return <DataLinkButton key={i} link={link} />;
```
aocenas, line 25: So previously `DataLinkButton` was meant to work for both internal and external links. Does this separation now mean we it should not be used for external links? In that case maybe we should have a `InternalDataLinkButton` and `ExternalDataLinkButton` to make that clear?

torkelo (approved):

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+    </>
+  );
+}
+
+const getStyles = (theme: GrafanaTheme) => ({
+  wrapper: css`
+    flex-basis: 150px;
+    width: 100px;
+  `,
+  externalLinksText: css`
```
torkelo, line 45: externalLinksHeading?

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
torkelo: yea, this looks like a list

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+    color: ${theme.colors.textWeak};
+    font-weight: ${theme.typography.weight.regular};
+    font-size: ${theme.typography.size.sm};
+    margin: 0;
+  `,
+  externalLink: css`
```
torkelo, line 51: This should use the external link style (color externalLink & underline on hover)

torkelo (changes requested):
Sorry, did not see this adding a new component to grafana-ui, 

Can you mark them as `@internal`  (With a jsdoc tag)

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
zoltanbedi: fixed

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
zoltanbedi: fixed

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
zoltanbedi: fixed

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
zoltanbedi: Not sure. I think if we only have external link then it makes sense to use the primary button which is now the DataLinkButton but other than use a simple link? WDYT @jessover9000 ?

zoltanbedi:
Updated screenshot
Screenshot 2021-02-05 at 17 07 18

aocenas (approved):

packages/grafana-ui/src/components/DataLinks/FieldLink.tsx
jessover9000: We can also use the primary button for external links, but with the current external link text (like "Go to [long url with lots of characters]"), the button might become very long and therefore the overlay window might become quite chunky too. How do we usually handle long link texts in a button? Do we have a solution for that built into the button component (line breaks or shortening the text)?

jessover9000:
Two alignment nitpicks:
1. The text of the primary button and the traceID are currently slightly misaligned (see red highlight in screenshot).
2. The margins of the "External link" headline are reversed – due to the Gestalt law of proximity, content always seems belong to the surrounding stuff that is closest. So the "External link" text looks like a subheadline to the traceID instead of a headline to the link list (green highlight). The margin in the top should be larger than the bottom one.
Exemplars-nitpicks

A fixed version could look something like this:
Exemplars-fixes

packages/grafana-ui/src/components/DataLinks/FieldLinkList.tsx
```diff
@@ -0,0 +1,71 @@
...
+ */
+export function FieldLinkList({ links }: Props) {
+  const styles = useStyles(getStyles);
+
+  if (links.length === 1) {
+    return <DataLinkButton link={links[0]} />;
+  }
+
+  const externalLinks = links.filter((link) => link.target === '_blank');
+  const internalLinks = links.filter((link) => link.target === '_self');
```
torkelo, line 23: internal links should not use _self, that will cause a full page reload

torkelo (approved):
looks good, but found one issue with using _self for internal links, that will cause angular router to do a full page reload , which we do not want

PR #30886, opened 2021-02-04
AlertingNG:  Test definition 

peterholmberg:
Adds functionality to test a alert definition and get alert instances back.

To test: 
Add a query (A)
Add a reduce expression (B) on A to reduce the query to a single value
Add a math expression on B to compare the value to some other value.

Hit test.


public/app/features/alerting/state/reducers.ts
```diff
@@ -62,11 +62,14 @@ export const initialAlertDefinitionState: AlertDefinitionState = {
...
   },
-  queryOptions: { maxDataPoints: 100, dataSource: { name: '-- Mixed --' }, queries: [] },
   queryRunner: new PanelQueryRunner(dataConfig),
   uiState: { ...store.getObject(ALERT_DEFINITION_UI_STATE_STORAGE_KEY, DEFAULT_ALERT_DEFINITION_UI_STATE) },
   data: [],
   alertDefinitions: [] as AlertDefinition[],
+  /* These are functions as they are mutated later on and redux toolkit will Object.freeze state so
+   * we need to store these using functions instead */
+  getInstances: () => [] as DataFrame[],
+  getQueryOptions: () => ({ maxDataPoints: 100, dataSource: { name: '-- Mixed --' }, queries: [] }),
```
torkelo, line 72: What is mutable in the query options?

peterholmberg: expr in prometheus queries

torkelo: where is that mutated?

peterholmberg: in the query editor.

torkelo: Can't we change that? must be a mistake (or old way of doing things)

peterholmberg: Sure, I can look into it.

torkelo: But old angular editors do modify it, so an option would be to clone it before we give it to angular editors. But let's keep like this for now

torkelo (approved):

PR #30892, opened 2021-02-04
CloudMonitoring: Increase service api page size

### LINKED ISSUE #30417: GCP Monitoring SLO Services query has a limit of 100 services in Grafana Variables
**What happened**:
I have a GCP project with 100+ monitoring services. I created a Grafana dashboard with a variable that queries these SLO Services, so when I go to the dashboard I have a nice dropdown menu where I can select the service I'm interested.
I realized that it's not showing me all the services, just the first 100.

Screen Shot 2021-01-20 at 10 28 06

I would expect this variable to load a complete list of all my SLO Services.

* Create a GCP project with more than 100 monitoring services
* Create a dashboard with a variable with the same configuration than in the image attached to have a dropdown menu with the list of services.
* The dropdown menu will only show the first 100 services.
* 
I think the solution would be setting the optional `pageSize` parameter of the Google Monitoring API. Right now is not set and it uses the default value which is probably 100.

**Environment**:
- Grafana version: 7.1.3
- Data source type & version: Google Cloud Monitoring
- OS Grafana is installed on:
- User OS & Browser: macOS. Any browser
- Grafana plugins:
- Others:

sunker:

By default, the [cloud monitoring services api](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/services/list) has a limit of 100 services. This PR increases that limit to 1000 by setting a query param. For this to work, I also needed to update the cache so that is supports query params. 

Fixes #30417

### DIFF
--- a/public/app/plugins/datasource/cloud-monitoring/api.test.ts
+++ b/public/app/plugins/datasource/cloud-monitoring/api.test.ts
@@ -22,7 +22,7 @@ async function getTestContext({ path = 'some-resource', options = {}, response =
   const fetchMock = jest.spyOn(backendSrv, 'fetch');
 
   fetchMock.mockImplementation((options: any) => {
-    const data = { [options.url.match(/([^\/]*)\/*$/)[1]]: response };
+    const data = { [options.url.match(/([^\/]*)\/*$/)![1].split('?')[0]]: response };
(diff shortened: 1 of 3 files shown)


joshhunt (approved):
nice

PR #30893, opened 2021-02-04
Make value mappings correctly interpret numeric-like strings

### LINKED ISSUE #29949: ValueMappings: Support string value mappings when value can be parsed as a number (For example IP Address)
**What happened**:
Matfching rules for Value=>Text mappings are not working for strings in a table. It appears that it's interpreting the IP addresses as numbers and ignoring everything beginning with the second `.` character. I have tried both setting the table's default Units to String, as well as using field overrides to do it. That made no difference.

I expect to be able to put an IP address in the Value field and a replacement string in the Text field to see my replacement string in the table instead of the IP address.

This table is constructed of Prometheus metrics with the labels shown as columns. All 3 columns in my screenshot are the same underlying data, but I created this to show the difference in behavior. The first column has no overrides. The second one is matching against "169.254.antyhing" to show that it matches every value that begins with `169.254`. The third column shows that `169.253` is not a match.

**Environment**:
- Grafana version: Grafana v7.3.6 (ea06633c34)
- Data source type & version: Prometheus 2.18.1
- OS Grafana is installed on: Ubuntu Bionic
- User OS & Browser: Debian Buster, Firefox 78.6.0esr (64-bit)
- Grafana plugins: Core, Grafana Image Renderer, Pie Chart, Status Map, worldPing
- Others:

image

### LINKED ISSUE #30764: Grafana v7.4.0-beta1: value mappings in override applying to all columns
Hello together,
I wanted to report an Bug.

The value overrides all values in the table, see pictures below

Grafana  Version: Grafana v7.4.0-beta1
In version 7.3.7 everything works

Image1s
Image2

dprokop:
Value mappings were incorrectly applied when mapping was applied to numeric-like strings.
For example a value 20/1 and 20/2 would be mapped to the same text if the mapping was set to 20/1 -> 'Some text'. That's because we were parsing (parseFloat) the incomming value and the result of it would for both values and the mapping be 20.

Fixes https://github.com/grafana/grafana/issues/30764
Fixes https://github.com/grafana/grafana/issues/29949

### DIFF
--- a/packages/grafana-data/src/utils/valueMappings.ts
+++ b/packages/grafana-data/src/utils/valueMappings.ts
@@ -15,20 +15,22 @@ const addValueToTextMappingText = (
     return allValueMappings.concat(valueToTextMapping);
   }
 
-  const valueAsNumber = parseFloat(value as string);
-  const valueToTextMappingAsNumber = parseFloat(valueToTextMapping.value as string);
+  let valueAsNumber, valueToTextMappingAsNumber;
 
-  if (isNaN(valueAsNumber) || isNaN(valueToTextMappingAsNumber)) {
-    if (value === valueToTextMapping.value) {
-      return allValueMappings.concat(valueToTextMapping);
+  if (isNumeric(value as string) && isNumeric(valueToTextMapping.value)) {
+    valueAsNumber = parseFloat(value as string);
+    valueToTextMappingAsNumber = parseFloat(valueToTextMapping.value as string);
+
+    if (valueAsNumber !== valueToTextMappingAsNumber) {
+      return allValueMappings;
     }
   }
 
-  if (valueAsNumber !== valueToTextMappingAsNumber) {
-    return allValueMappings;
+  if (value === valueToTextMapping.value) {
+    return allValueMappings.concat(valueToTextMapping);
   }
 
-  return allValueMappings.concat(valueToTextMapping);
+  return allValueMappings;
 };
 
 const addRangeToTextMappingText = (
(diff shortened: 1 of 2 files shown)


packages/grafana-data/src/utils/valueMappings.ts
```diff
@@ -93,3 +95,9 @@ const isNullValueMap = (mapping: ValueMap): boolean => {
   }
   return mapping.value.toLowerCase() === 'null';
 };
+
+// Ref
```
dprokop, line 99:
```suggestion
// Ref https://stackoverflow.com/a/42356340
```

packages/grafana-data/src/utils/valueMappings.ts
```diff
@@ -93,3 +95,9 @@ const isNullValueMap = (mapping: ValueMap): boolean => {
   }
   return mapping.value.toLowerCase() === 'null';
 };
+
+// Ref https://stackoverflow.com/a/42356340
+export function isNumeric(num: any) {
+  num = '' + num;
+  return !isNaN(num as number) && !isNaN(parseFloat(num));
+}
```
hugohaggmark, line 103: Could almost be written as `return Boolean(Number(num));` but it fails for boolean values :(

hugohaggmark (approved):
Skilled work, love the tests 🎉

packages/grafana-data/src/utils/valueMappings.ts
hugohaggmark, line 103: I guess this would work, at least all the tests are passing 😉 
```suggestion
export function isNumeric(num: any) {
  if (num === true) {
    return false;
  }

  return Boolean(Number(num));
}
```

dprokop:
Ah not, some tests are red!

hugohaggmark:
> Ah not, some tests are red!

My bad I just ran the tests in `packages/grafana-data/src/utils/valueMappings.test.ts`

dprokop:
> My bad, I just checked the tests in packages/grafana-data/src/utils/valueMappings.test.ts
@hugohaggmark not your falt. Singlestat tests surfaced a legitimate issue with the code, fixing it.

PR #30900, opened 2021-02-04
CloudMonitoring: Show service and SLO display name in SLO Query editor

### LINKED ISSUE #29611: Google Cloud Monitor Plugin - Service and SLO display names are not resolving

Resolution of Service and SLO display names when creating a SLO query in Grafana, with Cloud Monitor plugin 

When selecting a gcp CM service through the editor, i’m seeing the service-id, and not the service-name

image

image

And then when I select a SLO, I see the SLO name, but once selected, it displays the SLO-id…  

image

image

**Why is this needed**:

It is very difficult to determine which Service or SLO the panel relates to when using the ID and not the display name.

sunker:
In the SLO Query editor in Cloud Monitoring, the Service field and the SLO field only showed the id in the segment dropdown. This was fine in some scenarios when the id was a human readable value, like `prometheus-service`, but that's not always the case so this PR makes sure SLO/Service display name is being used in the field dropdown in the editor. 

Please note that the display name needs to be loaded from the api before they can be properly displayed, so for existing SLO queries it won't work until SLO/Service field has been clicked and re-selected. 

Fixes #29611

### DIFF
--- a/public/app/plugins/datasource/cloud-monitoring/components/SLOQueryEditor.tsx
+++ b/public/app/plugins/datasource/cloud-monitoring/components/SLOQueryEditor.tsx
@@ -21,7 +21,9 @@ export const defaultQuery: (dataSource: CloudMonitoringDatasource) => SLOQuery =
   aliasBy: '',
   selectorName: 'select_slo_health',
   serviceId: '',
+  serviceName: '',
   sloId: '',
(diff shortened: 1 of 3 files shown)


joshhunt (approved):
looks good to me :)

PR #30908, opened 2021-02-04
Alerts: Update Alert storybook to show more states

torkelo:
Current stories: 

Screen Shot 2021-02-04 at 16 42 25

Updated story: 

Screen Shot 2021-02-04 at 15 03 40


peterholmberg (changes requested):
Very nice, could you perhaps have the overview as a separate story instead of removing the other ones? So we can have a few stories where you can have controls to change the behavior of the component?

torkelo:
There was only one know, for Severity not sure what value that brings when you can see all. But yea a story that had knobs for everything could be a good future addition if users find knobs useful.

PR #30913, opened 2021-02-04
Variables: Extend option pickers to accept custom onChange callback

Clarity-89:
Add optional `onVariableChange` prop to option pickers. This will enable to control the setting of the variable and do some extra post-processing if necessary. 

Also some minor changes to DashboardPicker to make it work better inside forms. 


public/app/features/variables/pickers/OptionsPicker/actions.ts
```diff
@@ -72,7 +72,13 @@ export const filterOrSearchOptions = (searchQuery = ''): ThunkResult<void> => {
   };
 };
 
-export const commitChangesToVariable = (): ThunkResult<void> => {
+const setVariable = async (updated: any) => {
```
hugohaggmark, line 75: This should be VariableWithMultiSupport, right?
```suggestion
const setVariable = async (updated: VariableWithMultiSupport) => {
```

public/app/features/variables/pickers/OptionsPicker/actions.ts
```diff
@@ -72,7 +72,13 @@ export const filterOrSearchOptions = (searchQuery = ''): ThunkResult<void> => {
...
+  const adapter = variableAdapters.get(updated.type);
+  await adapter.setValue(updated, updated.current, true);
+  return;
+};
+
+export const commitChangesToVariable = (callback?: (updated: any) => void): ThunkResult<void> => {
```
hugohaggmark, line 81: Same here:
```suggestion
export const commitChangesToVariable = (callback?: (updated: VariableWithMultiSupport) => void): ThunkResult<void> => {
```

hugohaggmark (commented):
Looks ok so far and I'm going to test this now. I'll get back with more feedback after that.

public/app/features/variables/pickers/OptionsPicker/actions.ts
Clarity-89: Great catch! Forgot to update the types.

public/app/features/variables/pickers/OptionsPicker/actions.ts
```diff
@@ -88,9 +94,11 @@ export const commitChangesToVariable = (): ThunkResult<void> => {
...
     }
 
-    const adapter = variableAdapters.get(updated.type);
-    await adapter.setValue(updated, updated.current, true);
-    return;
+    if (callback) {
+      return callback(updated);
+    } else {
+      return setVariable(updated);
+    }
```
hugohaggmark, line 101: Nit: maybe we could rewrite this as:
```suggestion
    if (callback) {
      return callback(updated);
    } 
    
    return setVariable(updated);
```

public/app/features/variables/textbox/TextBoxVariablePicker.tsx
```diff
@@ -27,7 +27,15 @@ export function TextBoxVariablePicker({ variable }: Props): ReactElement {
...
-    variableAdapters.get(variable.type).updateOptions(variable);
+
+    if (onVariableChange) {
+      onVariableChange({
+        ...variable,
+        current: { ...variable.current, value: updatedValue },
+      });
+    } else {
+      variableAdapters.get(variable.type).updateOptions(variable);
+    }
```
hugohaggmark, line 38: Nit: maybe this is clearer:
```suggestion
    if (onVariableChange) {
      onVariableChange({
        ...variable,
        current: { ...variable.current, value: updatedValue },
      });
      return;
    } 
    
    variableAdapters.get(variable.type).updateOptions(variable);
```

hugohaggmark (approved):
LGTM! 🎉

torkelo (approved):

Issue #30564, opened 2021-01-22
GraphNG (Act 2)


leeoniya:
This is a follow-up to https://github.com/grafana/grafana/issues/27848 (Act 1).

Now that we have most work items finished, let's migrate the few remaining and future (planned and possible) things to this issue. Most of the items here are specific to **Time series**, but I'm leaving the title as GraphNG for continuity. Feel free to change it.

Please help me migrate any remaining features to this issue and let's close the old one.

Cleanup:
- [ ] Renames https://github.com/grafana/grafana/issues/32846

Old features (parity):

- [x] [ms precision in tooltip when user zoomed in enough](https://github.com/grafana/grafana/issues/36017)
- [ ] Stacking an inverted series (* -1 transform?) - Issues: https://github.com/grafana/grafana/issues/30365
- [x] Closest no-null datapoint focus https://github.com/grafana/grafana/issues/31309, https://github.com/grafana/grafana/issues/34542
- [x] stacking (by value, by %)  https://github.com/grafana/grafana/pull/30749
- [x] histograms (move to its own panel)
- [x] thresholds (y)
- [ ] regions (x) https://github.com/grafana/grafana/issues/35486
    - [ ] Bug: #15010 #28017
- [x] annotations
- [x] shared crosshair https://github.com/grafana/grafana/pull/33433
- [ ] sorting legend items #30907
- [x] stat panels no longer use a shared min/max for all graphs, which de-stabilizes ranges across the group. we should look into adding softMin/softMax to Stat, which was designed to stabilize graphs without giving up autoranging (#32765)
- [x] Hiding series via field options: https://github.com/grafana/grafana/issues/33013

New features:

- [ ] focus series on legend hover (torkel request)
- [ ] series symbols - Issues: #28528, PR: #30192
- [ ] padding autosizing - Issues: #33092
- [ ] explicit padding control (for aligning multiple charts) - Issues: #4469
- [ ] grid show/hide (x, y) - Issues: #9305
- [ ] axis tick label rotation - Issues: #3591
- [ ] axis coloring (to visually match series to their y scale) - Issues: #29932
- [ ] bar width control, expose barMaxWidth (for timeseries pseudo-histograms) - Issues: #14130, #18220, Related PR: #30499, Demos: https://leeoniya.github.io/uPlot/demos/line-paths.html (see bottom 2 charts)
- [ ] x/y rectangular zoom, y-only zoom - Issues: #9457
[...]

PR #30919, opened 2021-02-05
Chore: make timeseries the default panel in 8+

ryantxu:
In Grafana 8, we aim to replace the flot panel with the new uPlot based one.  We should make it the default ASAP ;)

image

After we do more migrations work and better testing, we should aim to *deprecate* the graph panel in 7.5.

This PR also makes `barchart` beta and fiddles the presentation order


leeoniya:
until we have better parity with the graph panel (#30564: stacking, thresholds, annotation editing), we should at least warn people about stuff that is still missing so they don't end up setting up a lot of stuff and then realizing there's no reverse-migration path and have to redo their whole config in a blinding rage.

this plays nicely into the whole "recommend the appropriate viz/panel" on our roadmap.

ryantxu:
I'm torn -- I think the sooner people use the new goodness the better.  Then they don't have to risk migrations later.  but yes there are a couple things that are not at parity yet.  

Maybe adding a banner like this:
image

And lets aim to remove it before 7.5 release?

packages/grafana-data/src/panel/PanelPlugin.ts
```diff
@@ -106,6 +106,7 @@ export class PanelPlugin<
   onPanelMigration?: PanelMigrationHandler<TOptions>;
   onPanelTypeChanged?: PanelTypeChangedHandler<TOptions>;
   noPadding?: boolean;
+  editBanner?: ComponentType | null;
```
ryantxu, line 109: or maybe better to hardcode for `timeseries' and 'graph' migration issues

torkelo:
The biggest potential blocker for doing this is alerting

public/app/plugins/panel/timeseries/EditBanner.tsx
```diff
@@ -0,0 +1,28 @@
...
+      `}
+      persistenceId="timeseries-parity-azx"
+      url={'https://github.com/grafana/grafana/issues/30564'}
+      urlTitle="See open github issue"
+    >
+      <p>
+        The new time series panel is now the default panel in grafana. For most things, this is will be the best choice.
+        However a few features exist in the graph panel that are not yet implemented in this panel. Speciffically:
+        <ul>
+          <li>Annotaiton editor</li>
```
hugohaggmark, line 20: nit: small typos
```suggestion
        However, a few features exist in the graph panel that are not yet implemented in this panel. Specifically:
        <ul>
          <li>Annotation editor</li>
```

hugohaggmark (commented):
Found some typos

dprokop:
@ryantxu moving to 8.0. I was thinking about the same thing, but there is still few things we need to address before making this one a default one.

osg-grafana:
Deferring to @achatterjee-grafana on any doc changes for this.

oddlittlebird:
Will the old Graph panel still be an option?

ryantxu:
> Will the old Graph panel still be an option?

For sure existing graph panels will continue to work with flot in 8.  Ideally we can remove the ability to create new flot panels in 8 like we did with the single-stat >> stat migration.  BUT there is a long list of stuff that needs done before we can consider that!  See:  https://github.com/grafana/grafana/issues/30564

ryantxu:
closing in favor of #33639

PR #30935, opened 2021-02-05
Alerting: Dedupe alerts so the same alert message won't repeat

torkelo:
Tired of seeing the below (which we should improve further by auto reloading or something) 

Screenshot from 2021-02-05 15-06-06


hugohaggmark (approved):
Nice 🎉

Clarity-89 (approved):

PR #30939, opened 2021-02-05
Theme: Use higher order theme color variables rather than is light/dark logic

torkelo:
Removes all use of selectThemeVariant and replaces all uses with the more abstract color variables

Also fixes a zIndex issue in explore log context menu where it shows the radio button through it:
Screenshot from 2021-02-05 15-38-29


Clarity-89 (approved):
Looks good!

PR #30957, opened 2021-02-06
Table panel pagination: add some styles according to UX design

### LINKED ISSUE #20739: Table panel pagination highlighting needs better accessibility
**Steps to reproduce**:
If I remove all color, the highlighting of the table panel pagination is hard to see. So might be a problem for people with color blindness. You can use the Spectrum browser extension in Achromatopsia mode to reproduce the problem.
Screenshot 2019-11-28 at 12 40 41

**How to improve**:
I would recommend using not only text color to highlight the active page but also adding a background highlight. The hover state of the active page has a background box, maybe that box could be the regular state and on hover the link text could get an underline.

**Environment**:
- Grafana version: 6.5
- Data source type & version: gdev test data

AmooHashem:
add some styles to table panel pagination, according to description in issue  #20739

Fixes #20739

nothing.

### DIFF
--- a/public/sass/components/_panel_table.scss
+++ b/public/sass/components/_panel_table.scss
@@ -19,8 +19,8 @@
   text-align: center;
   font-size: 90%;
   line-height: 2px;
-  background-color: #f1f1f1;
-  border-radius: 1px;
+  background-color: ghostwhite;
+  border-radius: 3px;
 
   ul {
     position: relative;
--- a/public/sass/components/_panel_table.scss
+++ b/public/sass/components/_panel_table.scss
@@ -19,6 +19,8 @@
   text-align: center;
   font-size: 90%;
   line-height: 2px;
+  background-color: #f1f1f1;
+  border-radius: 1px;
 
   ul {
     position: relative;
@@ -44,6 +46,7 @@
     &.active {
       font-weight: bold;
       color: $blue;
+      text-decoration: underline;
     }
   }
 }


public/sass/components/_panel_table.scss
```diff
@@ -19,6 +19,8 @@
   text-align: center;
   font-size: 90%;
   line-height: 2px;
+  background-color: ghostwhite;
```
torkelo, line 22: this wont work for light theme

AmooHashem: Hmmm,
it's like gray and I thought gray is good for both light and dark mode.

mckn: If I understood @jessover9000 description in the original issue the background color should be the `$tight-form-func-bg;`

public/sass/components/_panel_table.scss
```diff
@@ -44,6 +46,7 @@
     &.active {
       font-weight: bold;
       color: $blue;
+      text-decoration: underline;
```
mckn, line 49: Also, this should only be added on `&:hover` according to the issue.

mckn (changes requested):
Super that you have opened a PR to help out with the project! However I think there are some improvements to be made before we can approve this PR.

If I understood the original issue we should do the following:

1. change background color to `$tight-form-func-bg`.
2. remove the background-color on hover and add a underline.

But, before continue with these changes I think it would be best to ping @jessover9000 to see if this changes still is valid.

jessover9000:
Thanks for tackling this! This issue is quite old indeed. I gave feedback on the following PR as well, which as far as I know is the most recent instance where pagination was tweaked https://github.com/grafana/grafana/pull/30151 

Could we just style the table pagination in this case the same way, if that PR doesn't already fix the styling anyways?

dprokop:
> Could we just style the table pagination in this case the same way, if that PR doesn't already fix the styling anyways?

I agree with @jessover9000 that maybe we could use the Pagination component from grafana-ui. This would require to implement an angular wrapper around this grafana-ui component, similar to what we do with folder picker : https://github.com/grafana/grafana/blob/350dcb999a46e598d4f8aba4c9dfc0799418edc7/public/app/core/angular_wrappers.ts#L175

@torkelo do you think we are going to remove the old table panel in Grafana v8? If so, maybe tweaking CSS is just fine.

torkelo:
Closing this as this panel is deprecated and will be moved to an external plugin in 8.0 or 8.1

PR #30972, opened 2021-02-08
Graph: Fixes so graph is shown for non numeric time values

### LINKED ISSUE #30929: 7.4.0 does not draw Graph from PostgreSQL Data Source
**What happened**:
7.4.0 does not draw Graph from PostgreSQL datasouce

7.4.0 draw Graph from PostgreSQL datasouce
docker run -d -p 3000:3000 --name grafana grafana/grafana:7.4.0
login, add datasource PostgreSQL, create panel
7.3.7 draw: изображение
7.4.0 does not draw:
изображение
In the pop-up window, only start values. 

I upgrade our Grafana and broken many dashbord, had to do downgrade. Please fix.

**Environment**:
- Grafana version: 7.4.0, 7.5.0-11433pre-ubuntu
- Data source type & version: Grafana PostgreSQL Data Source - Native Plugin, version Core
- OS Grafana is installed on: deb and docker
- User OS & Browser: OS: debian, ubuntu; Browser: Firefox 84.0.2, Microsoft Edge 88.0.705.62
- Grafana plugins:
- Others:

hugohaggmark:
When time column includes non numeric time values, i.e. strings like `2021-01-01 00:00:00` then the values where just passed through without conversion to numbers. This results in the graph not being drawn.

Fixes #30929

### DIFF
--- a/packages/grafana-ui/src/components/uPlot/Plot.tsx
+++ b/packages/grafana-ui/src/components/uPlot/Plot.tsx
@@ -4,7 +4,7 @@ import { buildPlotContext, PlotContext } from './context';
 import { pluginLog } from './utils';
 import { usePlotConfig } from './hooks';
 import { PlotProps } from './types';
-import { DataFrame } from '@grafana/data';
+import { DataFrame, dateTime, FieldType } from '@grafana/data';
 import { UPlotConfigBuilder } from './config/UPlotConfigBuilder';
 import usePrevious from 'react-use/lib/usePrevious';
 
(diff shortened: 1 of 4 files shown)


torkelo:
Introduced by this PR https://github.com/grafana/grafana/pull/29402   @ryantxu can you take a look at this? (as your PR introduced this)

Does this impact the time series pane as well? How does it handle time stamps like this?

hugohaggmark:
> Introduced by this PR #29402 @ryantxu can you take a look at this? (as your PR introduced this)
> 
> Does this impact the time series pane as well? How does it handle time stamps like this?

Yes, feel free to change this PR if needed @ryantxu

dprokop:
> Does this impact the time series pane as well? How does it handle time stamps like this?

@torkelo it does affect time series panel too. 


@hugohaggmark I've tested with this data (Direct Input ds):
```
"time","value"
2021-01-01T01:00:00.000Z,1
2021-01-01T02:00:00.000Z,2
2021-01-01T03:00:00.000Z,3
```

and it seems like the graph panel(both angular and time series) does not work still.

dprokop:
@hugohaggmark my bad - wrong time range - the Graph panel works, checking time series now

ryantxu:
I'm a bit concerned about solving this in each panel -- having a string for time *should* be rare, and maybe we should apply this `fix` early in the workflow so that this is not required in every single panel.

I'll take a look at that later today

hugohaggmark:
Thank you @ryantxu, no rush, but we scheduled the patch release for tomorrow Wednesday 😉

dprokop (approved):

torkelo:
I think this is the best patch solution, but created a follow PR to move date parsing to epoch to a more central location:
https://github.com/grafana/grafana/issues/31016

Issue #30940, opened 2021-02-05
Stat Panel: Duration(s) unit not working


twbecker:
![DurationUnitBroken](https://user-images.githubusercontent.com/1576404/107052592-dc630200-679b-11eb-8871-e535c109ee3e.png)

**What happened**:
Immediately upon upgrade from 7.3.7 -> 7.4.0 a stat panel broke, showing the "No Value" value instead of the duration as it's configured to.

**What you expected to happen**:
The duration value should be displayed (e.g. 4 hours).

**How to reproduce it (as minimally and precisely as possible)**:
Choose duration (s) as the unit for a integral field.

**Anything else we need to know?**:

**Environment**:
- Grafana version: 7.4.0
- Data source type & version: Prometheus 
- OS Grafana is installed on: Centos 7
- User OS & Browser: Firefox

dswarbrick:
Confirmed, happening for me too.

peterholmberg:
Possible dupe of https://github.com/grafana/grafana/issues/30938

PR #30979, opened 2021-02-08
StatPanel: Fixes issue formatting date values using unit option

### LINKED ISSUE #30938: Stat panel: Datetime unit not working for time field
**What happened**:
I have the stat panel configured as shown on the screenshots:
image

image

As you can see, the datetime unit is not working. Date shown in my local format + "No date if today" function is not working

Datetime units should also apply if "Fields" is set to "Time"

Described above

-

**Environment**:
- Grafana version: 7.4.0
- Data source type & version: Influxdb 1.8.2
- OS Grafana is installed on: Debian 4.9
- User OS & Browser: Windows 10 Opera Gx
- Grafana plugins: -
- Others: -

torkelo:
Fixes #30938

Not sure what is best here the shouldApply function on field option is constantly causing issues. Makes it so that you have to use overrides in some cases to
apply units that should act on time or string fields.

Wonder if we can remove this for unit? Some things could break?

### DIFF
--- a/packages/grafana-data/src/field/displayProcessor.ts
+++ b/packages/grafana-data/src/field/displayProcessor.ts
@@ -27,9 +27,10 @@ interface DisplayProcessorOptions {
 // Reasonable units for time
 const timeFormats: KeyValue<boolean> = {
   dateTimeAsIso: true,
-  dateTimeAsIsoSmart: true,
+  dateTimeAsIsoNoDateIfToday: true,
   dateTimeAsUS: true,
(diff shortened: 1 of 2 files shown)


peterholmberg:
this covers https://github.com/grafana/grafana/issues/30940 too right?

torkelo:
@peterholmberg nope, that is something else. Fixing that one now (in separate PR)

dprokop (approved):
Maybe we could make the shouldApply function optional and only specify when needed?

PR #30982, opened 2021-02-08
Units: Fixes formatting of duration units

### LINKED ISSUE #30940: Stat Panel: Duration(s) unit not working

torkelo:
Fixes #30940

Also rewrite unit tests to use jest table tests (creates much better test error messages)


peterholmberg (approved):
:+1:

PR #30984, opened 2021-02-08
Remove last synchronisation field from LDAP debug view

undef1nd:
Removes last sync field from LDAP debug view and solves [this Enterprise issue](https://github.com/grafana/grafana-enterprise/issues/314 )


xlson (commented):
Looks great but I think we should remove the prevSync field from the type as well.

undef1nd:
@xlson 
not directly related to this issue, but I think this component is not used anywhere https://github.com/grafana/grafana/blob/master/public/app/features/admin/UserSyncInfo.tsx  Do I get it right?

xlson:
@undef1nd I think so. @alexanderzobnin is this file used?

xlson (approved):
LGTM!

Clarity-89 (approved):
Looks good!

PR #30985, opened 2021-02-08
InfluxDB: Handle columns named "table"

### LINKED ISSUE #29967: "runtime error: invalid memory address or nil pointer dereference" when doing InfluxDB query
**What happened**:
Dashboard fails with "Internal Server Error - Check the Grafana server logs for the detailed error message." when doing an InfluxDB query. Other queries work, but all queries on that particular measurement fail.

Query should work as others do.

Not entirely sure, as I said, it only happens for that same measurement. I am using the MySQL plugin for Telegraf to pull the data. I have everything disabled expect for `gather_table_schema`.

The server logs:
```
t=2020-12-23T10:02:44+0000 lvl=eror msg="Request error" logger=context userId=1 orgId=1 uname=admin error="runtime error: invalid memory address or nil pointer dereference" stack="/usr/local/go/src/runtime/panic.go:212 (0x4569d2)\n/usr/local/go/src/runtime/signal_unix.go:742 (0x456852)\n/drone/src/pkg/tsdb/influxdb/flux/builder.go:243 (0x1a947a6)\n/drone/src/pkg/tsdb/influxdb/flux/executor.go:74 (0x1a9565c)\n/drone/src/pkg/tsdb/influxdb/flux/executor.go:30 (0x1a95304)\n/drone/src/pkg/tsdb/influxdb/flux/flux.go:42 (0x1a95dc4)\n/drone/src/pkg/tsdb/influxdb/influxdb.go:49 (0x1a989eb)\n/drone/src/pkg/tsdb/request.go:17 (0xf5b319)\n/drone/src/pkg/api/metrics.go:74 (0x14bd055)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/drone/src/pkg/api/common.go:36 (0x14e688d)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:112 (0x13f4845)\n/drone/src/pkg/middleware/request_metrics.go:54 (0x13f4830)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:112 (0x13f5586)\n/drone/src/pkg/middleware/request_tracing.go:25 (0x13f5579)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:112 (0x13f46d2)\n/drone/src/pkg/middleware/recovery.go:160 (0x13f46c5)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:79 (0x9c2a32)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:157 (0x9c18d4)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:135 (0x9c1679)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:112 (0x13e99fe)\n/root/go/pkg/mod/github.com/go-macaron/gzip@v0.0.0-20160222043647-cad1c6580a07/gzip.go:96 (0x13e99f1)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/drone/src/pkg/middleware/util.go:37 (0x13f5ad4)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:79 (0x9c2a32)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:157 (0x9c18d4)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:135 (0x9c1679)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:112 (0x13f1bdb)\n/drone/src/pkg/middleware/logger.go:35 (0x13f1bc6)\n/usr/local/go/src/reflect/value.go:476 (0x4a6f46)\n/usr/local/go/src/reflect/value.go:337 (0x4a6438)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:177 (0x9c1bb9)\n/root/go/pkg/mod/github.com/go-macaron/inject@v0.0.0-20160627170012-d8a0b8677191/inject.go:137 (0x9c158a)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/context.go:121 (0x9c2bbc)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/router.go:187 (0x9d42d0)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/router.go:294 (0x9cea2f)\n/root/go/pkg/mod/gopkg.in/macaron.v1@v1.3.9/macaron.go:220 (0x9c7acd)\n/usr/local/go/src/net/http/server.go:2843 (0x7285e2)\n/usr/local/go/src/net/http/server.go:1925 (0x723cec)\n/usr/local/go/src/runtime/asm_amd64.s:1374 (0x478c00)\n"
t=2020-12-23T10:02:44+0000 lvl=eror msg="Request Completed" logger=context userId=1 orgId=1 uname=admin method=POST path=/api/ds/query status=500 remote_addr=xxx.xxx.xx.xxx time_ms=66 size=120 referer="https://xxxxxxxxx/d/xxxxxxx/xxxxxxx?orgId=1"
```

The query:
```
from(bucket: "main-bucket")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "mysql_table_schema")
  |> filter(fn: (r) => r["table"] == "users")
  |> filter(fn: (r) => r["_field"] == "rows")
  |> aggregateWindow(every: v.windowPeriod, fn: last, createEmpty: false)
  |> yield(name: "last")
```
The same query works fine on the InfluxDB dashboard.

I've already had a look at the source code, the error is right here, looks like it has to do with the time column:
https://github.com/grafana/grafana/blob/e375fb0f2b05e614087ac0c5189e73236405985a/pkg/tsdb/influxdb/flux/builder.go#L243

**Environment**:
- Grafana version: 7.3.5 (commit: 11f305f88a, branch: HEAD)
- Data source type & version: InfluxDB 2.0.3 (git: fe04d346df)
- OS Grafana is installed on: Ubuntu 18
- Grafana plugins: N/A
- Others: N/A

gabor:
When data is returned for an influxdb flux-query, there is a column called "table" that contains a number indicating the table-id. Grafana uses it like this ( https://github.com/grafana/grafana/blob/master/pkg/tsdb/influxdb/flux/builder.go#L184 ):
```table, ok := record.ValueByKey("table").(int64)```

The problem is, it can happen that there are two columns in the flux-result named "table". one is the data we want, and the other one may be a user-defined column that happens to be named "table".  When this happens, the above-mentioned code breaks, because it is reading from a `map` structure, where the second table-value overwrote the first table-value.

To reproduce this, simply have an influxdb-query, and rename some column to "table" in the result by inserting a line like this into the query:
```|> rename(columns: {something: "table"})```

The approach to fix this is to simply not use the table-value at all. We create our own table-id. The InfluxDB documentation says https://docs.influxdata.com/influxdb/cloud/reference/glossary/#group-key for data from the same table the fields which are annotated as in-group, do not change. So the idea is to use the values of these non-changing fields as our table-id.

This is only a work-in-progress pull-request, i am not yet sure if this approach is robust enough.

Another, more conservatite approach would be to detect this situation (two table-columns) and return an error saying this is not supported and the user has to rename the column to something else.

this should fix https://github.com/grafana/grafana/issues/29967

### CODE REVIEW

File: pkg/tsdb/influxdb/flux/builder.go [hunk 1]
```diff
@@ -175,14 +183,44 @@ func getTimeSeriesTimeColumn(columns []*query.FluxColumn) *query.FluxColumn {
 	return nil
 }
 
+func getTableID(record *query.FluxRecord, groupColumns []string) []interface{} {
+	result := make([]interface{}, len(groupColumns))
+
+	for i, colName := range groupColumns {
+		result[i] = record.ValueByKey(colName)
```
aocenas, line 190: So I was thinking what will happen here if we have duplicate columns. If there are 2 `table` columns here we still can access only one which would be a bit weird imho. Looking at the docs it seems like table change should be handled by `if result.TableChanged() {` https://github.com/influxdata/influxdb-client-go#querytableresult. And we seem to be doing that here https://github.com/grafana/grafana/blob/e0448513eb29bd9f5f435e4585de3c34a56dee72/pkg/tsdb/influxdb/flux/executor.go#L56. So then in https://github.com/grafana/grafana/blob/cfaca42e973fa86efb62f10ed01586fbd582953a/pkg/tsdb/influxdb/flux/builder.go#L185 as code stands right this is basically an initialisation (because init sets table to -1 and that first append will do this) and probably isn't designed to handle table changes in between appends.

Obviously the `table` column thrown this off but that does not seem like a good way to handle tables anyway.

Personally I think we could try to throw away the code checking table at all inside the builder and just say that this handles single table and you should make sure you use it correctly and create new builder `if result.TableChanged()` (optionally just return error if that went wrong). I think that would simplify this quite a bit. We could then just check if we are doing append for the first time and if so, just run single init code.

gabor: unfortunately flux seems to call two different things "table", both what one might call "series" and a grouping of series that has the same schema, and `TableChanged` is only called for when the second thing changes, and not when the first :-(

regarding the duplicate names, i checked and `flux`, except the table-thing,  does not allow duplicate column-names. will update the code with a comment about this.

aocenas (approved):

PR #30988, opened 2021-02-08
Grafana/UI: Remove DismissableFeatureInfoBox and replace with LocalSt…

oscarkilhed:
This removes the DismissableFeatureInfoBox, and replaces the two instances where it was used with LocalStorageProvider. This moves the concern of how to persist a dismissed box from grafana/ui to the grafana app.

The keys used for local storage has changed. Users that has already dismissed these info boxes will see them again. We could use the old key, but that could confuse things in the future as those keys refer to grafana/ui.


torkelo (approved):
Looks good!

public/app/features/datasources/settings/CloudInfoBox.tsx
```diff
@@ -38,27 +39,38 @@ export const CloudInfoBox: FC<Props> = ({ dataSource }) => {
...
-          target="_blank"
-          rel="noreferrer"
-          title="The free plan includes 10k active metrics and 50gb storage."
-        >
-          free-forever Grafana Cloud plan
-        </a>
-        .
-      </div>
-    </DismissableFeatureInfoBox>
+    <LocalStorageValueProvider<boolean> storageKey="datasources.settings.cloudInfoBox.isDismissed" defaultValue={false}>
```
dprokop, line 42: Nit: maybe we could make it a constant, to keep things consistent with other usages, i.e. https://github.com/grafana/grafana/blob/1d689888b0fc2de2dbed6e606eee19561a3ef006/public/app/core/components/TimePicker/TimePickerWithHistory.tsx#L12

dprokop (approved):
Now it looks great!

PR #30990, opened 2021-02-08
BarGauge: Improvements to value sizing and table inner width calculations 

### LINKED ISSUE #30965: Table: gradient gauge cells are no longer readable
**What happened**:
After upgrading to grafana 7.4.0, values next to gradient gauges in a table are no longer readable:
image

in earlier versions of Grafana the fields with cell display mode set to gradient gauge were readable:
image

Create a table
set cell display mode for 1 column to Gradient gauge
set unit to microseconds

**Environment**:
- Grafana version: 7.4.0
- User OS & Browser: Windows & Firefox

torkelo:
Fixes #30965

* [x]  Adds proper measurement of value sizing instead of always taking a fixed percentage of the width for the value. This will improve the gauge a lot for smaller widths making the value more more readable. (so a narrow gauge can still have readable value)
* [x] Adds some heuristic to try to use the title font size as base when measuring the value width, in order to try two match those to sizes 
* [x] Improves the table inner width calculation and moves it to be shared among all cells (Sparkline cell will need it as well). And fixes the logic so correct innerWidth is calculated for the last chid if there is a scrollbar 
* [x] Moves the last child padding from a inner cell div to the cell div itself (that has the default cell padding) 

Before: 
Screenshot from 2021-02-09 09-31-09

After:
Screenshot from 2021-02-09 09-30-46

Before table: 
Screenshot from 2021-02-09 09-37-46

After table:
Screenshot from 2021-02-09 09-40-00

### DIFF
--- a/packages/grafana-ui/src/components/BarGauge/BarGauge.test.tsx
+++ b/packages/grafana-ui/src/components/BarGauge/BarGauge.test.tsx
@@ -10,6 +10,7 @@ import {
   getTitleStyles,
   getValuePercent,
   BarGaugeDisplayMode,
+  calculateBarAndValueDimensions,
 } from './BarGauge';
 import { getTheme } from '../../themes';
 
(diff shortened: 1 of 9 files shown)


peterholmberg (approved):
Nice, good opportunity to remove a snapshot test too :)

peterholmberg (approved):
Nice

PR #30992, opened 2021-02-08
Elasticsearch: Fix alias field value not being shown in query editor

### LINKED ISSUE #30946: Elasticsearch: Alias field not shown correctly in query editor
**What happened**:
When setting an alias for an ES query, it is correctly applied but it doesn't get shown in the UI after, for instance, reloading the page.

The Alias input field should show correctly what is saved in the query model.

- Create a panel with an ES query
- write anything in the Alias field
- Switch to the transform tab and back to query
- Alias is empty

**Environment**:
- Grafana version: 7.4.0

Elfo404:
The react migration introduced a bug for which the alias field value wasn't being shown in the UI (although it was correctly set and changes modified the query correctly)

Fixes #30946


zoltanbedi (approved):
lgtm🎍

ivanahuckova (approved):
Works nicely!

PR #30993, opened 2021-02-08
Chore: Introduce LibrarySrv for interacting with library panels API

kaydelaney:
Introduces a new frontend service "LibrarySrv", for interacting with the library panels API.
See also, https://github.com/grafana/grafana/pull/30653


torkelo (commented):
not sure this should be a core service, and no need to follow the service class pattern & getLibrarySrv etc here (that is a relic from angularjs services). 

I think it's better to put this as just straight exported functions in features/library-panels/state/api.ts 

similar to https://github.com/grafana/grafana/blob/master/public/app/features/admin/state/apis.ts

hugohaggmark (commented):
I agree with Torkel, I think it's better to export the functions. Maybe even as thunks and if we're thinking of having a separate Redux state for Library Panels?

kaydelaney:
Changed to just export the functions themselves.

torkelo (approved):

hugohaggmark (approved):
LGTM 🎉 !

torkelo:
Good progress! 

Some point (that can be done in follow-up PRs)
* No Apply button for a library panel (should only be an option to Save or Discard, Esc should open modal so user can choose to discard or save). So no way to leave library panel edit without saving or discarding changes

Add panel view:
* The padding is not aligned (same left and right)
Screenshot from 2021-02-12 15-28-50

The library panel view:

* The heading "Popular panels from the panel library"  should be above the filter & search inputs 
* The `Create new reusable panel`  is something we can wait with I feel to follow-up PR as that requires special logic for discard (needs to delete panel) 

Screenshot from 2021-02-12 15-23-40

torkelo:
ah, sorry the ^ above comment was on the wrong PR

dprokop (approved):
@kaydelaney looks great with the functions instead of service class!

Issue #30933, opened 2021-02-05
Time Series Panel: Uncaught Type Error with NaN values


kylebrandt:
**What happened**:
When switching from the Graph Panel to Time series panel a series with a NaN value (dataframe) there is the following error:

```
uPlot.esm.js:3291 Uncaught TypeError: Cannot read property 'map' of null
    at uPlot.esm.js:3291
    at Array.forEach (<anonymous>)
    at drawAxesGrid (uPlot.esm.js:3245)
    at uPlot.esm.js:3464
    at Array.forEach (<anonymous>)
    at _commit (uPlot.esm.js:3464)
```

**How to reproduce it (as minimally and precisely as possible)**:
![image](https://user-images.githubusercontent.com/1692624/107037017-820c7600-6788-11eb-96e5-5aa29edf0609.png)

![image](https://user-images.githubusercontent.com/1692624/107037038-8769c080-6788-11eb-9821-34143c243e29.png)

or using Load Apache arrow data with the time range matching above:

```
[...]

PR #31010, opened 2021-02-09
DataFrameJSON: introduce a json format that supports same semantics as arrow

ryantxu:
We are currently using `json(base64(arrow))` to pass data from backend to the frontend -- needless to say, this is less efficient that it could be :)

@leeoniya and I have explored various options to improve this.  TLDR, all signs point towards json with native browser compression.  One big drawback to JSON is that it does not support: `undefined`, `NaN`, `Infinity`, `-Infinity` -- this PR adds a replacements section to the existing DataFrameDTO so that it can represent the same data as arrow.  (excluding the stuff we don't really support in the frontend anyway -- like binary etc)

For Grafana 8, we will also evaluate removing the `Vector<T>` abstraction and forcing field values to be a simple array.  That is out-of-scope for this PR, but this puts us on a path to greatly simplify the DataFrame.

Current:
`arrow -> gzip(json(base64(arrow)))`

We want:
`arrow -> brotli(json(arrow))`


packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            { name: 'value', type: FieldType.number },
+          ],
+        },
+        data: [
+          [100, 200, 300],
+          ['a', 'b', 'c'],
+          [1, 2, 3],
+        ],
+        replaced: [
+          {}, // nothing to replace, but keeps the index
```
leeoniya, line 21: could even be `null` to skip object allocation/encoding ;)

packages/grafana-data/src/dataframe/DataFrameJSON.ts
```diff
@@ -0,0 +1,153 @@
...
+      if (flen && flen > length) {
+        length = flen;
+      }
+    }
+  }
+
+  const fields = schema.fields.map((f, index) => {
+    let r: FieldValueReplacements = {};
+    let buffer = data ? data[index] : [];
+    buffer.length = length; // will pad with undefined
```
leeoniya, line 135: if this is an array, i think i't best to actually fill it with `undefined`, otherwise you get a sparse often de-optimized array by doing this.

compare the console output of

```js
let x = [];
x.length = 100;
console.log(x);

let y = [];
y.length = 100;
y.fill(undefined, 0, 100);
console.log(y);
```

leeoniya:
something about this version make me like it more ;)

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
leeoniya: this started out without the fill, and the perf difference was quite large:

https://github.com/leeoniya/uPlot/blob/6ad1cb86e85672604515045abca3c891aab94086/src/utils.js#L377

packages/grafana-data/src/dataframe/DataFrameJSON.ts
leeoniya: this started out without the fill, and the perf difference was quite large:

https://github.com/leeoniya/uPlot/blob/6ad1cb86e85672604515045abca3c891aab94086/src/utils.js#L377

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            { name: 'name', type: FieldType.string },
+            { name: 'value', type: FieldType.number },
+          ],
+        },
+        data: [
+          [100, 200, 300],
+          ['a', 'b', 'c'],
+          [1, 2, 3],
+        ],
+        replaced: [
```
leeoniya, line 20: one thing to consider is that `data` and `replacements` will always occur together. not sure if it's worth grouping them:

```js
{
  schema: {
    fields: []
  },
  data: {
    values: [],
    replaced: [],
  }
}
```

would add a bit more overhead on the wire, but semantically a bit better maybe.

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
ryantxu: :+1:

packages/grafana-data/src/dataframe/DataFrameJSON.ts
ryantxu: interesting -- yes, the intention here is that they end up the same length in whatever the most efficient structure is.

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
ryantxu: +1 for `data.values[][]`

each values array ends up on field.values, so there is a bit more consistency

torkelo:
One thing that just struck me, that is not super clear to me: 

> One big drawback to JSON is that it does not support: undefined, NaN, Infinity, -Infinity

Do we need undefined, NaN, Infinity & -Infinity?

Can see a need for NaN, but Infinity & -Infinity? What do we miss by losing NaN & undefined ?

leeoniya:
@torkelo see https://github.com/grafana/grafana/issues/30933

torkelo:
From that issue, I cannot see what we miss but just handling all undefined, NaN, Infinity & -Infinity as null , I am sure we miss something, the question is how important that is

ryantxu:
>  the question is how important that is

I don't feel strongly -- for sure in the backend it is very important, in the front-end we need to filter them out whenever used as numbers anyway.  They are important when shown as strings and understanding what happened in expressions.

@kylebrandt -- any thoughts.

torkelo:
I can see NaN could be valuable for server-side expressions, maybe we can render that as a string as a "NaN" string in the JSON? but that will mean missed type array

kylebrandt:
For me one of the main reason to use arrow was to be able to develop data sources that could query data from sources (e.g. SQL, Azure Data Explorer, Prometheus) and pass it on in a non-destructive way. In this way, any destructive changes to the data are pushed to the end of the data flow (visualization, alerting, etc) etc instead of in the middle of it. 

From a user perspective, as an example, one could query a SQL table and see the data that is in SQL without various stuff being changed.

For a backend plugin developer, with this model the backend developer can use native Go types and then then encode that. Generally before when data sources would return a NaN it would not work (there are various bugs about this in the repo). So in this way, each backend developer doesn't need to know if their datasource returns NaNs and check for it (as it will fail to encode into json). They can effectively pass on any sort of tabular data as it in the source system.

But this also holds true from things like Time (there are different ways to encode time), or (or other things like different types of duration we have not added yet), or large numbers (Although you could still make large numbers string).

leeoniya:
for this PR i evaluated a several options (arrow, base64, protobuf, json in various schemas, brotli & gzip sizes of all options, encode/decode performance in the ui) and the general conclusion was that the overwhelmingly typical case had a very clear advantage by leveraging native browser compression (brotli or gzip, with brotli about 40% smaller) on top of mindfully structured json that's cheap to _optionally_ inflate on the client with primitive loops and simple math or replacements. anything that requires a lib on the client to decompress or decode in js-land is slower by multiple factors, adds a lot to frontend bundle (arrow js lib is 300kb and much slower end-to-end), and adds complexity over native js structs.

if we do need to encode more complex stuff, my recommendation is to use base64-wrapped protocol buffers inside json, coupled with https://github.com/mapbox/pbf on the client for decoding a subset of the json response, instead of doing the same with raw arrow. this will be slower than what's in this PR, faster than what we're doing now, and hopefilly be the exception and not the rule. we're aiming for the 97% use case to be minimal overhead so we can stream at 20hz+ over websockets, where frontend parse/decode easily becomes a bottleneck.

i don't see an issue with supporting optional inflation of NaN's, Infinity, etc. this pr is designed to avoid this both over the wire and during processing if the entities struct is not provided by the backend.

as for cleaning NaNs, etc prior to viz, this has to be done anyways cause they can emerge anyways from doing math on 0s, nulls inside transforms.

(not a rant, just fill disclosure/context :)

kylebrandt:
In terms of Dataframes and the backend it goes something like:
 - Build our Go "SDK" Dataframes (from a user perspective, in a plugin, they just return this)
 - Then the SDK code will turn each frame into an "Arrow Table" (as a serial step), and then base64 encode it (also serial). [Frames.MarshalArrow](https://github.com/grafana/grafana-plugin-sdk-go/blob/v0.86.0/data/arrow.go#L700).

For things on the backend that need to decode it, the reverse is used (https://github.com/grafana/grafana-plugin-sdk-go/blob/v0.86.0/data/arrow.go#L685).

So, if want to change the backend such that we continued to use SDK Go Dataframes, but encode them differently. It would be a breaking change, but the impact of that breaking change I think would be minimal to existing plugin code (although replacing the serialization layer might be a fair amount of work for backend team).

If we want to change the Go Dataframe model itself, the breaking change would likely impact plugin authors a lot more I think.

ryantxu:
To be clear -- this is not about changing anything on the backend SDK and how plugins pass around data.  Arrow gives us a very clear contract that I don't see any value in revisiting.  In particular for supporting new languages (we already have functioning C# and java back ends!)

The goal here is to find an efficient way to pass data from the backend to the browser -- currently we send lots of bytes that are both not used and unusable in the browser -- with a little bit of clever (not lots!) we should be able to get functionally equivalent for the things that are actually supportable.

packages/grafana-data/src/dataframe/DataFrameJSON.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Holds value bases per field so we can encode numbers from fixed points
+   * e.g. [1612900958, 1612900959, 1612900960] -> 1612900958 + [0, 1, 2]
+   */
+  bases?: number[];
+
+  /**
+   * Holds value multipliers per field so we can encode large numbers concisely
+   * e.g. [4900000000, 35000000000] -> 1e9 + [35, 4.9]
```
leeoniya, line 51: i goofed, the 35 and 4.9 should be swapped (or the prior)

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
```diff
@@ -0,0 +1,83 @@
+import { FieldType } from '../types/dataFrame';
+import { DataFrameJSON, dataFrameFromJSON } from './DataFrameJSON';
+
+describe('DataFrame JSON', () => {
+  describe('when called with a DataFrame', () => {
+    it('then it should reverse the order of values in all fields', () => {
```
leeoniya, line 6: "should decode values not supported natively in JSON (e.g. NaN, Infinity)"

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        data: {
+          values: [
+            [100, 200, 300],
+            ['a', 'b', 'c'],
+            [1, 2, 3],
+          ],
+          entities: [
+            null, // nothing to replace, but keeps the index
+            { NaN: [1], Inf: [2] },
+            { NegInf: [2] },
```
leeoniya, line 24: maybe also add an `undefined` test here, since `decodeFieldValueEntities()` supports it.

leeoniya:
only did a read-through and left a few suggestions (have not actually run the code yet).

packages/grafana-data/src/dataframe/DataFrameJSON.test.ts
ryantxu:
```suggestion
    it('should decode values not supported natively in JSON (e.g. NaN, Infinity)', () => {
```

ryantxu:
Thanks @leeoniya  -- for the simple edit suggestions, can you use the github suggestions feature?  It makes commiting changes super super easy
suggestions

packages/grafana-data/src/dataframe/DataFrameJSON.ts
```diff
@@ -0,0 +1,196 @@
...
+  /**
+   * A columnar store that matches fields defined by schema.
+   */
+  values: any[][];
+
+  /**
+   * Since JSON cannot encode NaN, Inf, -Inf, and undefined, these entities
+   * are decoded after JSON.parse() using this struct
+   */
+  entities?: Array<FieldValueEntityLookup | null>;
```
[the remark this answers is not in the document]
ryantxu: This is what you suggested in slack:
```
"entities": {
   "NaN": [[21, [1, 3]]],
   "Inf": [[21, [4]]],
   "NegInf": [[21, [0]]]
}
```

@leeoniya any thoughts?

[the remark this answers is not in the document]
leeoniya: this structure makes it more expensive to ask "does this series have any entities".

generally i expect `entities` to be absent entirely. if this case is atypical, i prefer the positional uniformity in structure between values, entities, enums, bases, factors. i've suggested maybe using `0` instead of `null` to hold the position, and maybe the same for any values that will get entity-replaced, this turns 4 bytes into 1.

i'm thinking to leave the current encoding as-is and see what real use-cases look like. i assume this isn't gonna be a public API commitment and the wire "compression" format can be swapped later.

leeoniya (approved):

PR #31030, opened 2021-02-09
Chore: Replace native select with grafana ui select 

### LINKED ISSUE #28048: Allow free text input for GCP project on dashboard variable query for stackdriver
**What happened**:
Dashboard variable for "Service Level Objectives (SLO)" query type can't be added, if service account doesn't have enough permission to list GCP projects. From browser's console log against https:///api/datasources/proxy/4/cloudresourcemanager/v1/projects access:

```
{
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "status": "PERMISSION_DENIED"
  }
}
```
indicating that the service account doesn't have enough permission.

But as you could see in the screenshot, available SLOs are properly listed under "Preview of values" section:
Screen Shot 2020-10-06 at 17 19 58

The variable can be "Add"-ed, even if the project name is blank.

Or, it should accept free text input in addition to the list item built from `datasource.getProjects()`.

Set up gcp-stackdriver as data source. Set "GCE Default Service Account" as authentication type of the data source.

On the GCP project, don't grant any additional permissions to the GCE default service account. By default, GCE default service account doesn't have organization-wide project listing permission.

Create SLI and SLO on GCP, following https://cloud.google.com/stackdriver/docs/solutions/slo-monitoring/ui/create-slo

I could manually type GCP project name on the query of panel editor,
Screen Shot 2020-10-06 at 17 42 17
 so it'd be natural to expect the same behavior on variables query, **regardless** what the cause of project listing failure.

**Environment**:
- Grafana version: 7.1.3
- Data source type & version: gcp-stackdriver (no specific version, as it's bundled with grafana)
- OS Grafana is installed on: CoreOS / Linux (on GKE)
- User OS & Browser: OSX, Chrome
- Grafana plugins: N/A
- Others: N/A

sunker:
Way back when the Cloud Monitoring VariableQueryEditor was built, styled native select elements were used so that they would look like all the angular dropdowns in the variable edit page. However since then, the variable edit page has been ported to react. In react we use react-select, which has different styling and different capabilities than the native select has. 

This PR refactors the VariableQueryEditor so that is uses react-selet intead of native select. 

Using react-select also made it possible to add custom values to the project field, and that fixes #28048.


hugohaggmark (approved):
LGTM 🎉

PR #31036, opened 2021-02-09
Grafana-data: bump markedjs to v2.x to resolve vulnerability

jackw:
A [vulnerability](https://github.com/markedjs/marked/security/advisories/GHSA-4r62-v4vq-hr96) has been discovered in markedjs. This PR addresses it by bumping marked to v2.0.0.

Having tested the text panel everything seems to work however it's worth noting the following breaking changes were introduced in marked v2.x:

**BREAKING CHANGES**
- em and strong tokenizers have been merged into one emStrong tokenizer.
- code and text tokenizers do not get passed all tokens as a second parameter.
- No longer supporting IE 11. IE 11 may still work but we are not committed to making sure it works with every update. We still provide an es5 version in lib/marked.js but some pollyfills may be needed for IE 11 in the future.

### DIFF
--- a/packages/grafana-data/package.json
+++ b/packages/grafana-data/package.json
@@ -27,7 +27,7 @@
     "apache-arrow": "0.16.0",
     "eventemitter3": "4.0.7",
     "lodash": "4.17.20",
-    "marked": "1.2.2",
+    "marked": "2.0.0",
     "rxjs": "6.6.3",
     "xss": "1.0.6"
   },
--- a/packages/grafana-data/package.json
+++ b/packages/grafana-data/package.json
@@ -27,7 +27,7 @@
     "apache-arrow": "0.16.0",
     "eventemitter3": "4.0.7",
     "lodash": "4.17.20",
-    "marked": "2.0.0",
+    "marked": "2.0.1",
     "rxjs": "6.6.3",
     "xss": "1.0.6"
   },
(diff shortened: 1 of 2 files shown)


mckn (commented):
I thinks this should be fine. The only place I can find where we use this is here:

https://github.com/grafana/grafana/blob/c3ed644f660a1a224a03ab09caee7469585edc99/packages/grafana-data/src/text/markdown.ts#L12

The only question mark that I can think of is if we should support IE11 or not.

jackw:
Thanks for taking the time to look at this @mckn. Unfortunately there's a bug in markedjs that breaks the text panel / variable e2e tests and this has made me hesitant to take it any further. I filed a [bug report](https://github.com/markedjs/marked/issues/1934) with the author and they're already looking into a fix. Do you think it's ok to proceed with the bump or should we wait?

> The only question mark that I can think of is if we should support IE11 or not.

I was under the impression we didn't support IE11? 🤔

hugohaggmark:
> Thanks for taking the time to look at this @mckn. Unfortunately there's a bug in markedjs that breaks the text panel / variable e2e tests and this has made me hesitant to take it any further. I filed a [bug report](https://github.com/markedjs/marked/issues/1934) with the author and they're already looking into a fix. Do you think it's ok to proceed with the bump or should we wait?
> 
> > The only question mark that I can think of is if we should support IE11 or not.
> 
> I was under the impression we didn't support IE11? 🤔

You're correct @jackw but what browser we support should become even clearer in the next few months.

hugohaggmark (commented):
I agree we should wait until the fix has been merged.

mckn:
Super, feels like you have a great plan for moving this forward 🙌 <3

dprokop (approved):

PR #31039, opened 2021-02-09
Usage stats: Adds source/distributor setting

bergquist:
This can be useful for filter out usage stats from other providers if they use this flag. 

### DIFF
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -821,9 +821,9 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
 	GoogleAnalyticsId = analytics.Key("google_analytics_ua_id").String()
 	GoogleTagManagerId = analytics.Key("google_tag_manager_id").String()
 	ReportingEnabled = analytics.Key("reporting_enabled").MustBool(true)
-	reportingDistributor := analytics.Key("reporting_distributor").String()
-	if len(reportingDistributor) >= 100 {
-		ReportingDistributor = reportingDistributor[:100]
+	ReportingDistributor = analytics.Key("reporting_distributor").MustString("grafana_labs")
+	if len(ReportingDistributor) >= 100 {
+		ReportingDistributor = ReportingDistributor[:100]
 	}
 
 	if err := readAlertingSettings(iniFile); err != nil {
--- a/pkg/setting/setting.go
+++ b/pkg/setting/setting.go
@@ -141,10 +141,12 @@ var (
 	appliedCommandLineProperties []string
 	appliedEnvOverrides          []string
 
-	ReportingEnabled   bool
-	CheckForUpdates    bool
-	GoogleAnalyticsId  string
-	GoogleTagManagerId string
+	// analytics
+	ReportingEnabled     bool
+	ReportingDistributor string
+	CheckForUpdates      bool
+	GoogleAnalyticsId    string
+	GoogleTagManagerId   string
 
 	// LDAP
 	LDAPEnabled           bool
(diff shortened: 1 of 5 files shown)


pkg/cmd/grafana-server/main.go
```diff
@@ -143,6 +144,9 @@ func executeServer(configFile, homePath, pidFile, packaging string, traceDiagnos
 	setting.BuildBranch = buildBranch
 	setting.IsEnterprise = extensions.IsEnterprise
 	setting.Packaging = validPackaging(packaging)
+	if len(distributor) >= 100 {
```
RichiH, line 147: Better to truncate?

pkg/cmd/grafana-server/main.go
```diff
@@ -51,10 +51,11 @@ func (e exitWithCode) Error() string {
...
 func main() {
 	var (
-		configFile = flag.String("config", "", "path to config file")
-		homePath   = flag.String("homepath", "", "path to grafana install/home path, defaults to working directory")
-		pidFile    = flag.String("pidfile", "", "path to pid file")
-		packaging  = flag.String("packaging", "unknown", "describes the way Grafana was installed")
+		configFile  = flag.String("config", "", "path to config file")
+		homePath    = flag.String("homepath", "", "path to grafana install/home path, defaults to working directory")
+		pidFile     = flag.String("pidfile", "", "path to pid file")
+		packaging   = flag.String("packaging", "unknown", "describes the way Grafana was installed")
```
SuperQ, line 57: Oh, I didn't know there was a `packaging` flag. Would it be OK to use this field for GitLab embedding like `-packaging="GitLab Omnibus"`?

pkg/cmd/grafana-server/main.go
marefr: It has to have a length greater than 100? Why?

pkg/cmd/grafana-server/main.go
SuperQ: Ahh, nevermind, `validPackaging()` restricts it to specific strings.

bergquist: Gitlab is not a valid package https://github.com/grafana/grafana/pull/31039/files#diff-828368acf3145252e4e01950c3f59e16cc7db5b40df7cb6da9436c6d77e44112R174

SuperQ: Want to add `embedded` to the `validTypes` list?

marefr (commented):
Looks good I think. Think some external parties are using the packaging option already today for similar things. 

How would external parties know that we want them to use this flag when running Grafana? Would it make more sense to make this a build flag instead? Or is the distributor the one who runs Grafana, e.g. Grafana Cloud or similar?

pkg/cmd/grafana-server/main.go
bergquist: fixed

pkg/cmd/grafana-server/main.go
marefr: I've instructed GitLab to use the packaging flag historically. Maybe something did change in regards to this

RichiH:
I would keep it as a configuration. We should add documentation as well, though that's arguably not urgent and can be done at a later point.

pkg/cmd/grafana-server/main.go
```diff
@@ -51,10 +51,11 @@ func (e exitWithCode) Error() string {
...
+		distributor = flag.String("distributor", "unknown", "name of the distributor of the package")
```
torkelo, line 58: why is this a command line option, would have thought it be an ini setting (that way it can be set using env variable as well)

SuperQ: Yes, it makes sense to put this under the `analytics` settings in the ini.

torkelo: yes that was my thinking as well

bergquist: Agree. I'm not sure what I was thinking... I'll fix this once the kids are asleep. Hopefully we can include it in 7.4.1

bergquist: :fist_raised: force pushed!

torkelo (approved):
Looks good!

pkg/setting/setting.go
```diff
@@ -815,10 +817,14 @@ func (cfg *Cfg) Load(args *CommandLineArgs) error {
...
 	analytics := iniFile.Section("analytics")
-	ReportingEnabled = analytics.Key("reporting_enabled").MustBool(true)
 	CheckForUpdates = analytics.Key("check_for_updates").MustBool(true)
 	GoogleAnalyticsId = analytics.Key("google_analytics_ua_id").String()
 	GoogleTagManagerId = analytics.Key("google_tag_manager_id").String()
+	ReportingEnabled = analytics.Key("reporting_enabled").MustBool(true)
+	reportingDistributor := analytics.Key("reporting_distributor").String()
+	if len(reportingDistributor) >= 100 {
+		ReportingDistributor = reportingDistributor[:100]
+	}
```
DanCech, line 827: ReportingDistributor isn't set here unless the length is >= 100

bergquist: :facepalm:

torkelo (approved):
Looks good

DanCech:
Nice

PR #31056, opened 2021-02-09
area/grafana/e2e: ginstall should pull version specified

briangann:

Current CI images only allow "master" and "latest" builds, and ignores specified version.

Bumps version of docker images also, and minor updates to build versions.

Affected images will be pushed with this change included


torkelo (approved):
Looks ok :)

Issue #31035, opened 2021-02-09
Incorporate addVariable flow into core variable e2e tests


hugohaggmark:

vickyyyyyyy:
For more context, our enterprise plugins use flows that are not currently being used in `grafana/e2e` (they are only exported) - these include `addAnnotation` (https://github.com/grafana/grafana/blob/master/packages/grafana-e2e/src/flows/addDashboard.ts#L101) and `addDashboard` (https://github.com/grafana/grafana/blob/master/packages/grafana-e2e/src/flows/addDashboard.ts#L41 additionally.

This means that if we have any breaking changes with latest stable version of Grafana, we won't receive any feedback on these flows being broken until our enterprise plugins' e2e tests fail. Our enterprise plugins builds would also be blocked until the fix is released in the next version. Currently also in discussion with our team on updating our build process to allow flexibility for this.

PR #31059, opened 2021-02-09
test: update addDashboard flow for v7.4.0 changes

vickyyyyyyy:

Currently, our e2e tests for our enterprise plugins are failing because our build runs them against latest stable Grafana (`v7.4.0`) which has introduced some changes to the `Variables` editor, in particular. The flows we use for our e2e tests (`addDashboard`, `addAnnotation` and `addVariable`: https://github.com/grafana/grafana/blob/master/packages/grafana-e2e/src/flows/addDashboard.ts) are currently only being exported by Grafana and not used in existing e2e tests in the package, so relevant updates from the changes were not applied to these flows.

The changes that affected this:
- We are no longer using `select` but `div` here so we are unable to continue using our cypress select

`v7.3.7`:
Screenshot 2021-02-09 at 13 06 04

`v7.4.0`:
Screenshot 2021-02-09 at 13 05 55

- The previously expected CSS class name for `Preview of values` no longer exists

`v7.3.7`:
Screenshot 2021-02-09 at 19 22 30

`v7.4.0`:
Screenshot 2021-02-09 at 19 22 39

I've updated the flows so they coincide with the `v7.4.0` changes so our e2e tests now pass:

**Before**

We get an error in Cypress:
Screenshot 2021-02-09 at 13 06 58

**After**

Testing the flow for `Type = "Constant"` to make sure the following change works:

```
e2e.components.Select.singleValue().should('have.text', 'Query').click().type(`${type}{enter}`);
```

Screenshot 2021-02-09 at 19 09 11

Testing the flow for `Type = "Query"` to make sure the following changes works:

```
e2e.components.Select.input().should('have.text', '').type(`${hide}{enter}`);
e2e.components.Select.input().should('be.visible').type(`${dataSource}{enter}`);
```

Screenshot 2021-02-09 at 19 07 17

- Added `.clear()` because it looks like without this, Cypress no longer removes what is in the textbox and just types so it adds to the default `query0` text.
- Added `.click({ force: true })` to fix flakiness
- Slight refactoring of existing code

Fixes https://github.com/grafana/integrations-team/issues/113, which is a ticket that encompasses all our enterprise plugins.

Ran the e2e tests for this PR locally and they have all passed:

Screenshot 2021-02-09 at 19 20 58

There is an issue open for potentially adding the missing flows in to our core tests to mitigate against something like this happening in the future: https://github.com/grafana/grafana/issues/31035

### DIFF
--- a/packages/grafana-e2e/src/flows/addDashboard.ts
+++ b/packages/grafana-e2e/src/flows/addDashboard.ts
@@ -37,7 +37,6 @@ interface AddVariableRequired {
 export type PartialAddVariableConfig = Partial<AddVariableDefault> & AddVariableOptional & AddVariableRequired;
 export type AddVariableConfig = AddVariableDefault & AddVariableOptional & AddVariableRequired;
 
-// @todo this actually returns type `Cypress.Chainable<AddDashboardConfig>`
 export const addDashboard = (config?: Partial<AddDashboardConfig>) => {
   const fullConfig: AddDashboardConfig = {
     annotations: [],
@@ -64,10 +63,12 @@ export const addDashboard = (config?: Partial<AddDashboardConfig>) => {
 
     fullConfig.variables = addVariables(variables);
 
-    e2e.components.BackButton.backArrow().click();
+    e2e.components.BackButton.backArrow().should('be.visible').click({ force: true });
   }
 
-  setDashboardTimeRange(timeRange);
+  if (timeRange) {
+    setDashboardTimeRange(timeRange);
+  }
 
   e2e.components.PageToolbar.item('Save dashboard').click();
   e2e.pages.SaveDashboardAsModal.newName().clear().type(title);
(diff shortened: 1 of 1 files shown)


packages/grafana-e2e/src/flows/addDashboard.ts
```diff
@@ -64,10 +63,12 @@ export const addDashboard = (config?: Partial<AddDashboardConfig>) => {
 
     fullConfig.variables = addVariables(variables);
 
-    e2e.components.BackButton.backArrow().click();
+    e2e.components.BackButton.backArrow().should('be.visible').click({ force: true });
   }
 
-  setDashboardTimeRange(timeRange);
+  if (timeRange) {
```
hugohaggmark, line 69: Is this if statement necessary? Seems like it's always set?

packages/grafana-e2e/src/flows/addDashboard.ts
```diff
@@ -156,25 +157,38 @@ const addVariable = (config: PartialAddVariableConfig, isFirst: boolean): AddVar
...
-  if (hide !== undefined) {
-    e2e.pages.Dashboard.Settings.Variables.Edit.General.generalHideSelect().select(hide);
+  // Hide only does not show up for VARIABLE_TYPE_CONSTANT
+  // Hide can be '', so we are explicitly checking for undefined
+  if (type !== VARIABLE_TYPE_CONSTANT && hide !== undefined) {
+    e2e.pages.Dashboard.Settings.Variables.Edit.General.generalHideSelect()
+      .should('be.visible')
+      .within(() => {
+        e2e.components.Select.input().should('have.text', '').type(`${hide}{enter}`);
+      });
```
hugohaggmark, line 174: Are there any plugins tests that use this? I think we should remove this option to minimize the length of this flow.

hugohaggmark (approved):
Looks good 🎉  and I left some comments, but I would also have a look at which options we actually use in the plugins e2e tests so we can minimize the length and complexity of this flow.

I would rather split it up in several small flows that we know we use. But that change could be made in a followup PR, i.e. when I add this flow to the Grafana e2e tests.

packages/grafana-e2e/src/flows/addDashboard.ts
torkelo: :+1:

hugohaggmark:
I forgot to mention this, but I really ❤️  your PR descriptions, really nice @vickyyyyyyy!

packages/grafana-e2e/src/flows/addDashboard.ts
vickyyyyyyy: I was thinking of making it more flexible so it's not needed but thinking about it more, I'll remove it this 👍

packages/grafana-e2e/src/flows/addDashboard.ts
vickyyyyyyy: Not at the moment, it looks like our plugins always test the `Constant` type route and prior to v7.4.0, the `hide` option would show up for `Constant` but not anymore, so I'll remove it 😄 

If we need it later, we can split this up into smaller flows (like you mentioned in your comment) 👍

PR #31074, opened 2021-02-10
Docs: Group id should be 0 instead of 1 in Docker upgrade notes

### LINKED ISSUE #30953: Default docker group is root(0) contrary to documents
**Documentation**:
https://grafana.com/docs/grafana/latest/installation/docker/#migrate-to-v51-or-later
"The Grafana Docker image runs with the root group (id 1) instead of the grafana group (id 472), for better compatibility with OpenShift. "

**What happened**:
Running fresh docker container on ubuntu 20  with 
```
version: '3.3'

services:
  grafana:
    container_name: grafana
    image: grafana/grafana
    restart: unless-stopped
    ports:
    - "3000:3000"
```
Checking the user running the grafana process shows the uid to be 472 and gid as 0.

```
bash-5.0$ uname -a
Linux 7a3db0a0ed2e 5.4.0-65-generic #73-Ubuntu SMP Mon Jan 18 17:25:17 UTC 2021 x86_64 Linux
bash-5.0$ id
uid=472(grafana) gid=0(root) groups=0(root)
bash-5.0$ grafana-server -v
Version 7.4.0 (commit: c2203b9859, branch: HEAD)
```

uid=0(root) gid=0(root) groups=0(root)

See the docker compose above

The documentation might need to be updated to clarify the default user is still 472, and the root group is 0 and not one. Since the source code is setting it to 0 [here](https://github.com/grafana/grafana/blob/ee405ef0693def6e8c227bbe5b99e062fc072212/Dockerfile#L40)

marefr:
The upgrade notes for Grafana 7.3 and Docker says that _The Grafana Docker image runs with the root group (id 1)_ which is wrong. It should be _The Grafana Docker image runs with the root group (id 0)_.

Fixes #30953

### DIFF
--- a/docs/sources/installation/docker.md
+++ b/docs/sources/installation/docker.md
@@ -139,7 +139,7 @@ This section contains important information if you want to migrate from previous
 
 ### Migrate to v7.3 or later
 
-The Grafana Docker image runs with the `root` group (id 1) instead of the `grafana` group (id 472), for better compatibility with OpenShift. If you extend the official Docker image you may need to change your scripts to use the `root` group instead of `grafana`.
+The Grafana Docker image runs with the `root` group (id 0) instead of the `grafana` group (id 472), for better compatibility with OpenShift. If you extend the official Docker image you may need to change your scripts to use the `root` group instead of `grafana`.
 
 ### Migrate to v6.5 or later
 


aknuds1:
Isn't the PR title wrong? It says the opposite of what the PR does.

aknuds1:
I set the milestone to 7.5.0 btw, since I figure it should be in the next minor release. Please correct it if I got it wrong.

aknuds1 (approved):

marefr:
Yeah, updated the title

osg-grafana:
> I set the milestone to 7.5.0 btw, since I figure it should be in the next minor release. Please correct it if I got it wrong.

Seems like this change needs to be in all releases that qualify as v7.3 or later. Would it make sense to include these changes in 7.3.# and 7.4.# releases too?

marefr:
@osg-grafana Guess I'm lacking context here - do we have a process for backporting and publishing docs for older versions than the current? I know historically we haven't had this.

osg-grafana:
@briangann, here is an doc infrastructure question that has come up. I have been at Grafana for about four months, so others have more context about versioning works.

marefr:
Discussed and agreed that it's too much work to update docs for older versions so we just don't do that currently

PR #31075, opened 2021-02-10
Drone: Make sure CDN upload is ok before pushing docker images

torkelo:
Want to make sure the CDN upload is ok before pushing docker images

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -186,7 +186,6 @@ steps:
     edition: oss
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: postgres-integration-tests
   image: grafana/build-container:1.3.2
@@ -471,7 +470,6 @@ steps:
       from_secret: docker_user
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: build-docker-images-ubuntu
   image: grafana/drone-grafana-docker:0.3.2
@@ -484,7 +482,6 @@ steps:
       from_secret: docker_user
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: postgres-integration-tests
   image: grafana/build-container:1.3.2
@@ -1821,7 +1818,6 @@ steps:
     edition: oss
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: build-docker-images-ubuntu
   image: grafana/drone-grafana-docker:0.3.2
@@ -1831,7 +1827,6 @@ steps:
     ubuntu: true
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: postgres-integration-tests
   image: grafana/build-container:1.3.2
@@ -2213,7 +2208,6 @@ steps:
     edition: enterprise
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: build-docker-images-ubuntu
   image: grafana/drone-grafana-docker:0.3.2
@@ -2223,7 +2217,6 @@ steps:
     ubuntu: true
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: postgres-integration-tests
   image: grafana/build-container:1.3.2
@@ -2723,7 +2716,6 @@ steps:
     edition: oss
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: build-docker-images-ubuntu
   image: grafana/drone-grafana-docker:0.3.2
@@ -2733,7 +2725,6 @@ steps:
     ubuntu: true
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: postgres-integration-tests
   image: grafana/build-container:1.3.2
@@ -3093,7 +3084,6 @@ steps:
     edition: enterprise
   depends_on:
   - copy-packages-for-docker
-  - upload-cdn-assets
 
 - name: build-docker-images-ubuntu
   image: grafana/drone-grafana-docker:0.3.2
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -482,6 +484,7 @@ steps:
       from_secret: docker_user
   depends_on:
   - copy-packages-for-docker
+  - upload-cdn-assets-ubuntu
```
aknuds1, line 487: There's no step called upload-cdn-assets-ubuntu, a bug?

aknuds1 (changes requested):

.drone.yml
torkelo: hm.. this task maybe has a different sfx variable? the sfx enterprise2 etc stuff was very messy :)

scripts/lib.star
```diff
@@ -769,6 +769,7 @@ def build_docker_images_step(edition, ver_mode, archs=None, ubuntu=False, publis
         'image': grafana_docker_image,
         'depends_on': [
             'copy-packages-for-docker',
+            'upload-cdn-assets' + sfx,
```
aknuds1, line 772: You probably want to make a variable `edition_sfx` or something, since `sfx` here is used for the base image variant:

```suggestion
            'upload-cdn-assets' + edition_sfx,
```

aknuds1 (changes requested):

torkelo:
@aknuds1 Pushed an update, tried to simpify some of the edition logic by extracting the sfx part to a util function, and renaming the ubuntuSfx to avoid my mistake in the future

scripts/lib.star
```diff
@@ -747,14 +743,16 @@ def build_docker_images_step(edition, ver_mode, archs=None, ubuntu=False, publis
     if ver_mode == 'test-release':
         publish = False
 
-    sfx = ''
+    ubuntuSfx = ''
```
aknuds1, line 746: Starlark uses snake case.
```suggestion
    ubuntu_sfx = ''
```

scripts/lib.star
```diff
@@ -747,14 +743,16 @@ def build_docker_images_step(edition, ver_mode, archs=None, ubuntu=False, publis
...
     if ubuntu:
-        sfx = '-ubuntu'
+        ubuntuSfx = '-ubuntu'
```
aknuds1, line 748:
```suggestion
        ubuntu_sfx = '-ubuntu'      
```

scripts/lib.star
```diff
@@ -765,10 +763,11 @@ def build_docker_images_step(edition, ver_mode, archs=None, ubuntu=False, publis
     if archs:
         settings['archs'] = ','.join(archs)
     return {
-        'name': 'build-docker-images' + sfx,
+        'name': 'build-docker-images' + ubuntuSfx,
```
aknuds1, line 766:
```suggestion
        'name': 'build-docker-images' + ubuntu_sfx,
```

scripts/lib.star
```diff
@@ -861,15 +860,16 @@ def deploy_to_kubernetes_step(edition, is_downstream=False):
         ],
     }
 
+def edition_sfx(edition):
```
aknuds1, line 863: Not 100% sure if edition_sfx is a good name, because it's used specifically for the "enterprise2" suffix, while in other contexts we can also use the "enterprise" edition suffix. I find it confusing, because I have to read the function to realize it's only dealing with the enterprise2 edition.

scripts/lib.star
```diff
@@ -861,15 +860,16 @@ def deploy_to_kubernetes_step(edition, is_downstream=False):
...
     if ver_mode == 'master' and edition in ('enterprise', 'enterprise2') and not is_downstream:
         return None
-
-    sfx = ''
-    packages_bucket = ''
-    if edition == 'enterprise2':
-        sfx = '-{}'.format(edition)
-        packages_bucket = ' --packages-bucket grafana-downloads' + sfx
+    
+    packages_bucket = ' --packages-bucket grafana-downloads' + edition_sfx(edition)
```
aknuds1, line 872: I'm not sure this change is necessarily a good idea, because it changes functionality. I.e., currently for other editions than enterprise2, the default bucket is used. With this change, it's assumed that the correct bucket for other editions than enterprise2 is "grafana-downloads". That assumption _could_ break at some point.

Being specific about the bucket _could_ be a good change, but it's not just a code refactoring.

.drone.yml
```diff
@@ -532,7 +532,7 @@ steps:
 - name: upload-packages
   image: grafana/grafana-ci-deploy:1.3.1
   commands:
-  - ./bin/grabpl upload-packages --edition oss
+  - ./bin/grabpl upload-packages --edition oss --packages-bucket grafana-downloads
```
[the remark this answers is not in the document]
torkelo: because I simplified the logic so that this param is always sent (I assume this is the default in the the code of upload-packages) 

But think it's better to have it spelled out here and makes the python logic simpler

scripts/lib.star
torkelo: yea agree that using putting enterprise2 in edition but it's not treated the same was as oss & enterprise edition is super duper confusing, took me a long time to understand that enterprise2 is not really a new edition but only in some new steps in the enterprise pipeline?

scripts/lib.star
torkelo: I think if it did break it we can update it here, better to be explicit than to have hidden defaults that are sometimes used and sometimes overriden?

scripts/lib.star
torkelo: not sure what other name to use, pro_sfx ?)

scripts/lib.star
aknuds1: I'm actually unsure if it's a good idea or not.

aknuds1:
Actually these changes don't work, since Drone requires that a step (upload-cdn-assets) is declared before those depending on it.

aknuds1 (approved):

Issue #30509, opened 2021-01-21
Best effort behaviour for Alertmanager high availability setup


g574:
**What would you like to be added**:
Grafana is now able to connect to two (or more) Alertmanagers: https://github.com/grafana/grafana/issues/24195
However, if one of them is down, the "test" button throws an error message (notification channel settings).
Change the logic in a way that it connects to the one that is available.

**Why is this needed**:
The reason why people set up alertmanager clusters is redundancy. If Grafana does not send alerts because one of them is down that defies the purpose of having this cluster.

aocenas:
@marefr pinging you as this seems like an oversight on our part.

marefr:
Seems to me like this error shouldn't be returned here:
https://github.com/grafana/grafana/blob/0079a275afcc1dfe85f169ad0ddd67f6ef2b53b0/pkg/services/alerting/notifiers/alertmanager.go#L185

However, if all of the URL's fails it should probably return an error.

@g574 interested in opening a PR for fixing this? 

Pinging @alistarle here as well since he implemented support for multiple URL's in #24196.

kurokochin:
Hi @marefr, I'm new to Grafana, but can I try work on this?

PR #31079, opened 2021-02-10
Alerting: Fix case when Alertmanager notifier fails if a URL is not working

kurokochin:
**What this PR does / why we need it**: Fixes behaviour of Notify that returns error when one of the dispatch event return error, to maintain high availability, we should return error when all dispatched events return error instead.

**Which issue(s) this PR fixes**: #30509 

### DIFF
--- a/pkg/services/alerting/notifiers/alertmanager.go
+++ b/pkg/services/alerting/notifiers/alertmanager.go
@@ -2,6 +2,7 @@ package notifiers
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strings"
 	"time"
@@ -170,6 +171,7 @@ func (am *AlertmanagerNotifier) Notify(evalContext *alerting.EvalContext) error 
 
 	bodyJSON := simplejson.NewFromAny(alerts)
 	body, _ := bodyJSON.MarshalJSON()
+	errCnt := 0
 
 	for _, url := range am.URL {
 		cmd := &models.SendWebhookSync{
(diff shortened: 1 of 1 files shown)


marefr:
Please sign the CLA and we'll review this. Thanks

kurokochin:
Hi, sorry, I've signed the CLA, previously my commit not attached to the account that I use to signed the CLA, should be good for now :)

pkg/services/alerting/notifiers/alertmanager.go
```diff
@@ -182,10 +184,16 @@ func (am *AlertmanagerNotifier) Notify(evalContext *alerting.EvalContext) error
...
 			am.log.Error("Failed to send alertmanager", "error", err, "alertmanager", am.Name, "url", url)
-			return err
+			errCnt++
 		}
 	}
 
+	// This happens when every dispatch return error
+	if errCnt == len(am.URL) {
+		am.log.Error("Failed to send alert for all urls")
+		return fmt.Errorf("failed to send alert for all urls")
```
marefr, line 194: Think we can skip logging the error in here and just return it

```suggestion
		return fmt.Errorf("failed to send alert to alertmanager")
```

marefr (changes requested):
In general looks good. Added a suggestion.

pkg/services/alerting/notifiers/alertmanager.go
kurokochin: Thanks for your suggestion! Should be resolved :)

marefr (approved):
LGTM

marefr:

kurokochin:
Glad to help!

PR #31087, opened 2021-02-10
NGAlert: Refactoring that handles cleaning up state

hugohaggmark:
I'll write comments in all the places in the code instead of writing a summary here but basically the connectWithCleanup doesn't work for the QueryRunner in state because you need to call `destroy` on it before deleting it.

### DIFF
--- a/public/app/features/alerting/NextGenAlertingPage.tsx
+++ b/public/app/features/alerting/NextGenAlertingPage.tsx
@@ -1,12 +1,13 @@
 import React, { FormEvent, PureComponent } from 'react';
 import { hot } from 'react-hot-loader';
-import { connect, MapDispatchToProps, MapStateToProps } from 'react-redux';
+import { connect, ConnectedProps } from 'react-redux';
 import { css } from 'emotion';
-import { DataFrame, GrafanaTheme, SelectableValue } from '@grafana/data';
+import { GrafanaTheme, SelectableValue } from '@grafana/data';
 import { PageToolbar, stylesFactory, ToolbarButton } from '@grafana/ui';
+
 import { config } from 'app/core/config';
 import { SplitPaneWrapper } from 'app/core/components/SplitPaneWrapper/SplitPaneWrapper';
-import AlertingQueryEditor from './components/AlertingQueryEditor';
+import { AlertingQueryEditor } from './components/AlertingQueryEditor';
 import { AlertDefinitionOptions } from './components/AlertDefinitionOptions';
 import { AlertingQueryPreview } from './components/AlertingQueryPreview';
 import {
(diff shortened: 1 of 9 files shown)


public/app/features/alerting/NextGenAlertingPage.tsx
```diff
@@ -46,6 +46,7 @@ interface DispatchProps {
   createAlertDefinition: typeof createAlertDefinition;
   evaluateNotSavedAlertDefinition: typeof evaluateNotSavedAlertDefinition;
   onRunQueries: typeof onRunQueries;
+  cleanUpDefinitionState: typeof cleanUpDefinitionState;
```
hugohaggmark, line 49: Created a new Thunk that cleans up state when component is unmounted.

public/app/features/alerting/NextGenAlertingPage.tsx
```diff
@@ -59,6 +60,10 @@ class NextGenAlertingPage extends PureComponent<Props> {
     }
   }
 
+  componentWillUnmount() {
+    this.props.cleanUpDefinitionState();
```
hugohaggmark, line 64: Created a new Thunk that cleans up state when the component is unmounted. Calling it here.

public/app/features/alerting/NextGenAlertingPage.tsx
```diff
@@ -30,7 +30,7 @@ interface OwnProps {
 
 interface ConnectedProps {
   uiState: AlertDefinitionUiState;
-  queryRunner: PanelQueryRunner;
+  queryRunner?: PanelQueryRunner;
```
hugohaggmark, line 33: This is now optional in state so that we can delete it in the reducer.

public/app/features/alerting/NextGenAlertingPage.tsx
```diff
@@ -183,11 +188,10 @@ const mapDispatchToProps: MapDispatchToProps<DispatchProps, OwnProps> = {
   getAlertDefinition,
   evaluateNotSavedAlertDefinition,
   onRunQueries,
+  cleanUpDefinitionState,
 };
 
-export default hot(module)(
-  connectWithCleanUp(mapStateToProps, mapDispatchToProps, (state) => state.alertDefinition)(NextGenAlertingPage)
-);
+export default hot(module)(connect(mapStateToProps, mapDispatchToProps)(NextGenAlertingPage));
```
hugohaggmark, line 194: I removed the connectWithCleanUp and handle the cleanup on unmount instead.

public/app/features/alerting/components/PreviewQueryTab.tsx
```diff
@@ -13,14 +13,21 @@ export const PreviewQueryTab: FC<Props> = ({ data, height, width }) => {
...
-    () => data.series.map((frame, index) => ({ value: index, label: getFrameDisplayName(frame) })),
-    [data.series]
-  );
+  const series = useMemo<Array<SelectableValue<number>>>(() => {
+    if (data?.series) {
+      return data.series.map((frame, index) => ({ value: index, label: getFrameDisplayName(frame) }));
+    }
+
+    return [];
+  }, [data]);
```
hugohaggmark, line 22: So data is either panelData or undefined, just making this clearer in the typings and here.

public/app/features/alerting/components/PreviewQueryTab.tsx
```diff
@@ -13,14 +13,21 @@ export const PreviewQueryTab: FC<Props> = ({ data, height, width }) => {
...
 
   // Select padding
   const padding = 16;
 
+  if (!data?.series?.length) {
+    return <Table data={toDataFrame([])} height={height} width={width} />;
+  }
+
```
hugohaggmark, line 30: I added this for when panelData has no series. Looks awful BTW, maybe we can enhance this experience.

hugohaggmark: One question I have is why we're using the Table component and not the TablePanel? Wouldn't some of this be included in the later?

public/app/features/alerting/state/reducers.ts
```diff
@@ -165,40 +165,47 @@ const alertDefinitionSlice = createSlice({
...
+      state.alertDefinition.id = action.payload.id;
+      state.alertDefinition.uid = action.payload.uid;
+      state.alertDefinition.condition = action.payload.condition;
+      state.alertDefinition.intervalSeconds = action.payload.intervalSeconds;
+      state.alertDefinition.data = action.payload.data;
+      state.alertDefinition.description = action.payload.description;
+      state.getQueryOptions = () => ({
+        ...currentOptions,
+        queries: action.payload.data.map((q: AlertDefinitionQueryModel) => ({ ...q.model })),
+      });
```
hugohaggmark, line 178: I've noticed that you're using the old style of returning state from a reducer. When you use createSlice there is no need for this, you can just update the props directly.

public/app/features/alerting/state/reducers.ts
```diff
@@ -165,40 +165,47 @@ const alertDefinitionSlice = createSlice({
...
-      return { ...state, getInstances: () => action.payload };
+      state.getInstances = () => action.payload;
+    },
+    cleanUpState: (state: AlertDefinitionState, action: PayloadAction<undefined>) => {
+      if (state.queryRunner) {
+        state.queryRunner.destroy();
+        state.queryRunner = undefined;
+        delete state.queryRunner;
+        state.queryRunner = new PanelQueryRunner(dataConfig);
+      }
```
hugohaggmark, line 201: So here is the proper clean up of queryRunner (I think), this should make sure we don't get any mem leaks and hanging observables.

hugohaggmark:
OOoops I've introduced some type errors, I'm going to fix those ASAP.

public/app/features/alerting/components/PreviewQueryTab.tsx
peterholmberg: I believe those have some dependencies on panel a panel?

peterholmberg (approved):
This works !

PR #30009, opened 2020-12-30
Elastic: Support request cancellation properly (Uses new backendSrv.fetch Observable request API)

hugohaggmark:
**What this PR does / why we need it**:
This PR replaces the usage of datasourceRequest in the Elastic data source with fetch and updates tests accordingly.

**Which issue(s) this PR fixes**:
Relates #27222

**Special notes for your reviewer**:
This PR is based on the bug fix for templating in Elastic PR #30003 so please review that PR first so this PR becomes easier to review.
I've used the Elastic7 devenv docker block (`make devenv sources=elastic7`) [together with this dashboard](https://gist.github.com/hugohaggmark/26376048a53c28e3670446e941f9d555) to verify before and after results:

PR #31088, opened 2021-02-10
Elasticsearch: Fix log row context errors

Elfo404:
https://github.com/grafana/grafana/pull/30009 replaced `datasourceRequest` with the new `fetch` API, which returns an observable, this is breaking `getLogRowContext` that was expecting a promise back from `this.post`.


ivanahuckova (approved):
👌

PR #31101, opened 2021-02-10
Transform: Fix Outer Join not assigning unique displayName to all fields

Elfo404:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #31107, opened 2021-02-10
Transformations: ensure duplicate names are not present after transformatoins

### LINKED ISSUE #30943: Multiple Queries & metrics as count 1, count 2 not working as expected after 7.4 upgrade - Elasticsearch Data source
**What happened**:
when we add multiple queries with Count and Unique Count and use transformations(joins & calculations), Earlier to 7.4 upgrade, we used to get them as Count 1, Count 2, etc,. And with those values, transformations were applied. However since the 7.4 upgrade, they aren't working as expected and breaking.

Expected the join and calculation transformations will work even after the upgrade. It was working well until 7.4.beta. 

With 7.3.7 or earlier version - create multiple count queries using elasticsearch datasource. 
you will have count 1, count 2, unique count 1, unique count 2, etc,.
Use transformation to join the count using the common field.
use transformation to make some calculations using count 1, count 2, unique count 1, etc,. 
organize transform and give names to the transformations. 
based on the given names, format using the overrides available. 

Now, Upgrade to 7.4, all these above changes will break. 

when we tested with 7.4-beta, it was all good. after upgrade to 7.4, it is not working.

**Environment**:
- Grafana version: 7.4
- OS Grafana is installed on: Windows/Linux
- User OS & Browser: Mac/Windows & Google Chrome
- Grafana plugins:
- Others:

### LINKED ISSUE #31104: Missing columns in Organize Fields transformation after Outer Join (New in 7.4)
**What happened**:
I used Outer Join transformation to join 3 queries.  this worked correctly and appended a number to common column names to prevent duplicates (i.e. Time 1, Time 2, Time 3).  I then tried to use the Organize Fields transformation to hide the redundant columns, but it only list the original column name (i.e. Time).  You can't hide individual columns.  You can only hide all or none.  This is new behavior in 7.4.  It worked as expected in 7.3.  

May be related to this issue that was fixed in 7.2: https://github.com/grafana/grafana/issues/24744

Screenshots:
table transform
queries

This is the behavior in 7.3
image001 (1)

I expected the column names in Organize Fields to match the post-join columns as shown on the table (i.e. Time 1, Time 2, etc.), and allow you to hide individual columns. 

Add 2 metrics with common columns/labels.  Outer join on a common column.  Try to hide only one of the duplicate columns.

**Environment**:
- Grafana version: 7.4
- Data source type & version:  Prometheus 2.23.0-rc.0
- OS Grafana is installed on:  Ubuntu 20.10
- User OS & Browser: Windows 10, Chrome 88.0.4324.150
- Grafana plugins: none
- Others:

ryantxu:
Alternative approach for https://github.com/grafana/grafana/pull/31101

After transformations some fields in frames might have duplicated names and therefore be ignored by subsequent transformations. This tries to solve this by cleaning duplicated names.

Fixes #31104
Fixes #30943


packages/grafana-data/src/field/fieldState.ts
```diff
@@ -38,14 +38,45 @@ export function getFieldDisplayName(field: Field, frame?: DataFrame, allFrames?:
...
+          }
+        } else {
+          names.set(name, 1);
+        }
+      }
+    }
+  }
+  return frames;
+}
+
```
torkelo, line 79: Looks good, think this is already handled by getFieldDisplayName 
https://github.com/grafana/grafana/blob/master/packages/grafana-data/src/field/fieldState.ts#L136 

So another option could be to clear any cached display name?

torkelo (approved):
Looks good

torkelo:
Would this fix https://github.com/grafana/grafana/issues/31104

Elfo404:
Closing this in favor of https://github.com/grafana/grafana/pull/31101

torkelo:
One big issue is that there is different logic somewhere between duplicate naming. This only adds a 2 for the second item, and not a 1 prefix to the first duplicate. 

So the visualization name has a different naming from the transform 
Screenshot from 2021-02-11 10-37-06

Elfo404:
@torkelo was about to comment the same: 

Screenshot 2021-02-11 at 09 33 57

I wonder if it still makes sense to merge this for 7.4.1 given without this a lot of dashboards would break. (it's also true that this inconsistency might as well break other dashboards)

packages/grafana-data/src/field/fieldState.ts
torkelo: This also needs to run between every transform not just at the end

ivanahuckova:
Just to keep in mind (use-case in my dashboards), any change to column names is going to break/ignore all of the **series overrides by name** that I have. I have to go to each of my panels and change the name of the column manually. 

**7.4-beta1 (working as expected):**

https://user-images.githubusercontent.com/30407135/107623448-9ef9eb00-6c59-11eb-800a-c556b76ea6f1.mov

**7.4.0 (broken):**

https://user-images.githubusercontent.com/30407135/107623630-eed8b200-6c59-11eb-9195-860529a5dfde.mov

**this branch:**
I have to manually re-select each override, it was chaning names (e.g temperature value -> value temperature). Kind of changing everything ad hoc. But at the end, I was able to set is to look as before, but definitely not ideal.
https://user-images.githubusercontent.com/30407135/107624060-99e96b80-6c5a-11eb-8786-f773f619c814.mov

torkelo:
@ryantxu we opted for a simple fix to get this fixed in 7.4.1 without introducing new breaking changes, https://github.com/grafana/grafana/pull/31121 

Still room to improve this to avoid this in other transforms, move clearing displayName to something that runs between every transform for example?

ryantxu:
Closing this since it may introduce other issues

PR #31115, opened 2021-02-11
Logging: sourcemap transform asset urls from CDN in logged stacktraces

domasx2:
When logging stacktraces from frontend, an attempt is made to lookup sourcemaps in build dir and transform frame location to original uncompiled version. Previoulsy the code would check if URL path starts with `/public/build/` to determine if it might belog to grafana before attempting to find the sourcemap. This would not work for CDN based assets that have URLs like `https://storage.googleapis.com/grafana-static-assets/grafana-oss/pre-releases/7.5.0-11925pre/public/build/app.38f83c49518ca194a9f7.js`. 

This PR updates this logic to also handle CDN based URLs

### DIFF
--- a/pkg/api/frontend_logging_test.go
+++ b/pkg/api/frontend_logging_test.go
@@ -46,12 +46,12 @@ func logSentryEventScenario(t *testing.T, desc string, event frontendlogging.Fro
 
 		sc := setupScenarioContext(t, "/log")
 
-		CDNRootURL, e := url.Parse("https://storage.googleapis.com/grafana-static-assets")
+		cdnRootURL, e := url.Parse("https://storage.googleapis.com/grafana-static-assets")
 		require.NoError(t, e)
 
 		cfg := &setting.Cfg{
 			StaticRootPath: "/staticroot",
-			CDNRootURL:     CDNRootURL,
+			CDNRootURL:     cdnRootURL,
 		}
 
 		readSourceMap := func(dir string, path string) ([]byte, error) {
--- a/pkg/api/frontend_logging_test.go
+++ b/pkg/api/frontend_logging_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"testing"
@@ -45,8 +46,12 @@ func logSentryEventScenario(t *testing.T, desc string, event frontendlogging.Fro
 
 		sc := setupScenarioContext(t, "/log")
 
+		CDNRootURL, e := url.Parse("https://storage.googleapis.com/grafana-static-assets")
+		require.NoError(t, e)
+
 		cfg := &setting.Cfg{
 			StaticRootPath: "/staticroot",
+			CDNRootURL:     CDNRootURL,
 		}
 
 		readSourceMap := func(dir string, path string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


pkg/api/frontend_logging_test.go
```diff
@@ -45,8 +46,12 @@ func logSentryEventScenario(t *testing.T, desc string, event frontendlogging.Fro
 
 		sc := setupScenarioContext(t, "/log")
 
+		CDNRootURL, e := url.Parse("https://storage.googleapis.com/grafana-static-assets")
```
aknuds1, line 49: Local variables shouldn't be capitalized.
```suggestion
		cdnRootURL, e := url.Parse("https://storage.googleapis.com/grafana-static-assets")
```

pkg/api/frontendlogging/source_maps.go
```diff
@@ -68,13 +68,16 @@ func (store *SourceMapStore) guessSourceMapLocation(sourceURL string) (*sourceMa
...
-			path:     filepath.Join("build", u.Path[len("/public/build/"):]) + ".map",
-			pluginID: "",
-		}, nil
+	// determine if source comes from grafana core, locally or CDN, look in public build dir on fs
+	if strings.HasPrefix(u.Path, "/public/build/") || (store.cfg.CDNRootURL != nil && strings.HasPrefix(sourceURL, store.cfg.CDNRootURL.String()) && strings.Contains(u.Path, "/public/build/")) {
+		pathParts := strings.SplitN(u.Path, "/public/build/", 2)
+		if len(pathParts) == 2 {
+			return &sourceMapLocation{
+				dir:      store.cfg.StaticRootPath,
+				path:     filepath.Join("build", pathParts[1]+".map"),
```
aknuds1, line 77: Not directly related to the PR, but should this be a filesystem path or a network path?

aknuds1 (approved):
Looks good AFAICT, but see my suggestion.

pkg/api/frontendlogging/source_maps.go
domasx2: filesystem path. This returned struct is a guess at file system location of the source map

pkg/api/frontend_logging_test.go
domasx2: fixed!

pkg/api/frontendlogging/source_maps.go
```diff
@@ -68,13 +68,16 @@ func (store *SourceMapStore) guessSourceMapLocation(sourceURL string) (*sourceMa
...
 
-	// determine if source comes from grafana core, look in public build dir
-	if strings.HasPrefix(u.Path, "/public/build/") {
-		return &sourceMapLocation{
-			dir:      store.cfg.StaticRootPath,
-			path:     filepath.Join("build", u.Path[len("/public/build/"):]) + ".map",
-			pluginID: "",
-		}, nil
+	// determine if source comes from grafana core, locally or CDN, look in public build dir on fs
+	if strings.HasPrefix(u.Path, "/public/build/") || (store.cfg.CDNRootURL != nil && strings.HasPrefix(sourceURL, store.cfg.CDNRootURL.String()) && strings.Contains(u.Path, "/public/build/")) {
```
torkelo, line 72: this can be simplified to simply store.cfg.GetContentDeliveryURL(..) + "/public/build"    (public/build is also part of CDN path)

domasx2: I considered this, but don't have `hs.License` in scope here to call `store.cfg.GetContentDeliveryURL(..)`. Could add it to `SourceMapStore`  struct, but no longer makes it that simpler. WDYT?

PR #31121, opened 2021-02-11
Transforms: Fixes Outer join issue with duplicate field names not getting the same unique field names as before 

### LINKED ISSUE #30943: Multiple Queries & metrics as count 1, count 2 not working as expected after 7.4 upgrade - Elasticsearch Data source

### LINKED ISSUE #31104: Missing columns in Organize Fields transformation after Outer Join (New in 7.4)

torkelo:
Temporary alternative fix to #31107

Fixes #31104
Fixes #30943

### DIFF
--- a/packages/grafana-data/src/field/fieldState.ts
+++ b/packages/grafana-data/src/field/fieldState.ts
@@ -38,10 +38,8 @@ export function getFieldDisplayName(field: Field, frame?: DataFrame, allFrames?:
   }
 
   const displayName = calculateFieldDisplayName(field, frame, allFrames);
-  field.state = {
-    ...field.state,
-    displayName,
-  };
+  field.state = field.state || {};
+  field.state.displayName = displayName;
 
   return displayName;
 }
--- a/packages/grafana-data/src/transformations/transformers/ensureColumns.test.ts
+++ b/packages/grafana-data/src/transformations/transformers/ensureColumns.test.ts
@@ -68,9 +68,6 @@ describe('ensureColumns transformer', () => {
               "config": Object {},
               "labels": Object {},
               "name": "A",
-              "state": Object {
-                "displayName": "A",
-              },
               "type": "number",
               "values": Array [
                 1,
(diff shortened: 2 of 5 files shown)


Elfo404 (approved):
🚀

PR #31131, opened 2021-02-11
Only update usagestats every 30min

bergquist:
These queries can be quite hefty. I dont think we need to run them every minute. 

30 min might be a bit height but I dont see why we would need fresh numbers of these stats. 


marefr (approved):
LGTM Makes sense. 

Just wonder why we had 1 minute in the first place. Something related to being able to monitor these stats with a high update frequency using prometheus/graphite. Was this perhaps earlier used for validating licensing requirements? Would ask enterprise team to have a look as well just in case

bergquist:
> Just wonder why we had 1 minute in the first place

My guess is that few ppl care about :)

> Was this perhaps earlier used for validating licensing requirements?

I'm fairly sure that won't be based on metrics. The admin usage page would still issue this query.

PR #31132, opened 2021-02-11
Permissions: Fix team and role permissions on folders/dashboards not displayed for non Grafana Admin users

### LINKED ISSUE #31134: Team/role permissions on dashboards/folders are not displayed to non Grafana Admin users
**What happened**:
Team permissions are not displayed for non Grafana Admin users. 

Grafana Admin view:
image

Admin user without Grafana Admin rights view:
image

The two screen above should show the same things.

Set one team permission and one role permission on a dashboard with a Grafana Admin user. Then, log in with a non Grafana Admin user and look at the dashboard permissions.

**Environment**:
- Grafana version: 7.4

AgnesToulet:
Fix team and role permissions not displayed for folder and dashboard permissions. 

Fixes https://github.com/grafana/grafana/issues/31134

There are two fixes in this PR:
- The first one is to initialize the `cfg.HiddenUsers` field as empty when the config field is empty.
- The second one is to not even use this config field if the `userLogin` field is empty which is the case for team and role permissions. This isn't necessary but it was my first fix and I think it makes senses anyway.

Also, this issue wasn't catched in the dashboard permissions tests because settings are not initialized in the same way than for a real Grafana instance. E2E tests for dashboard permissions would be great but I don't have the knowledge to include them in this PR.

### DIFF
--- a/pkg/api/dtos/models.go
+++ b/pkg/api/dtos/models.go
@@ -77,7 +77,7 @@ func GetGravatarUrlWithDefault(text string, defaultText string) string {
 }
 
 func IsHiddenUser(userLogin string, signedInUser *models.SignedInUser, cfg *setting.Cfg) bool {
-	if signedInUser.IsGrafanaAdmin || userLogin == signedInUser.Login {
+	if userLogin == "" || signedInUser.IsGrafanaAdmin || userLogin == signedInUser.Login {
 		return false
 	}
 
--- a/pkg/api/dashboard_permission.go
+++ b/pkg/api/dashboard_permission.go
@@ -32,7 +32,7 @@ func (hs *HTTPServer) GetDashboardPermissionList(c *models.ReqContext) response.
 
 	filteredAcls := make([]*models.DashboardAclInfoDTO, 0, len(acl))
 	for _, perm := range acl {
-		if dtos.IsHiddenUser(perm.UserLogin, c.SignedInUser, hs.Cfg) {
+		if perm.UserId > 0 && dtos.IsHiddenUser(perm.UserLogin, c.SignedInUser, hs.Cfg) {
 			continue
 		}
 
--- a/pkg/api/dtos/models_test.go
+++ b/pkg/api/dtos/models_test.go
@@ -1,0 +1,86 @@
+package dtos
+
+import (
+	"testing"
+
+	"github.com/grafana/grafana/pkg/models"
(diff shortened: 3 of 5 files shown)


pkg/api/dtos/models.go
```diff
@@ -77,7 +77,7 @@ func GetGravatarUrlWithDefault(text string, defaultText string) string {
 }
 
 func IsHiddenUser(userLogin string, signedInUser *models.SignedInUser, cfg *setting.Cfg) bool {
-	if signedInUser.IsGrafanaAdmin || userLogin == signedInUser.Login {
+	if userLogin == "" || signedInUser.IsGrafanaAdmin || userLogin == signedInUser.Login {
```
marefr, line 80: Would be good having a test of this method. 

In addition it feels a bit fragile/awkward that this method would be called for any permission, should only be called for user permissions. WDYT?

marefr (changes requested):
Looks good I think. But requested some changes, see comment

pkg/api/dtos/models.go
AgnesToulet: Yes you're right, I added the test in the dashboard permissions file. Now, I'm not sure if we should keep this line or remove it.

xlson: I agree, we should def call this only when we need to.

AgnesToulet: Yes, I added a check for dashboard / folder permissions in [this commit](https://github.com/grafana/grafana/pull/31132/commits/bdb6542ec620803c61e1309e0bb3f77eb941609c) and I made a PR for Enterprise too.

marefr (changes requested):
Great. Almost there

dashboard_permissions.go should be updated as well: https://github.com/grafana/grafana/blob/a18e4d8d6437a0cf57e8ddebe61d36d5ce081e25/pkg/api/dashboard_permission.go#L35

pkg/api/dtos/models.go
marefr: > I'm not sure if we should keep this line or remove it.

If someone uses the API with an API key the `signedInUser.Login` would be empty. But doesn't really matter because permissions shouldn't have an empty `userLogin` is it's a user permissions. This extra check wouldn't hurt I would say :)

marefr (approved):
LGTM great

PR #31144, opened 2021-02-11
Explore: Do not show non queryable data sources in data source picker 

### LINKED ISSUE #30299: Filter data sources in explore to only show ones I can query
image

When I select the ruler, ruler-logs or alertmanager datasources, I get an empty query editor.  We should just not show them.

image

torkelo:
Fixes #30299

*  DataSourceSrv: Filter out non queryable data sources by default


zoltanbedi (approved):
lgtm 🕶

Issue #30182, opened 2021-01-11
Grafana-ui: update stories to use controls


jackw:
**What would you like to be added**:
Now that grafana ui storybook is on v6 it'd be great if we migrate away from knobs to controls as detailed [here](https://github.com/grafana/grafana/blob/master/contribute/style-guides/storybook.md#controls).

**Why is this needed**:
Controls infers props through types and auto generates the form controls that knobs creates. This results in less code and so less maintenance. Below is a list of all remaining stories that make use of `addon-knobs`.

- [x] Alert/Alert.story.tsx #31145 
- [x] Badge/Badge.story.tsx https://github.com/grafana/grafana/pull/30180
- [x] BarGauge/BarGauge.story.tsx #31223
- [x] BigValue/BigValue.story.tsx #31240
- [x] Button/Button.story.tsx https://github.com/grafana/grafana/pull/29433
- [x] ButtonCascader/ButtonCascader.story.tsx #31288
- [x] CallToActionCard/CallToActionCard.story.internal.tsx #31393 
- [x] Card/Card.story.internal.tsx (This doesn't have any knobs)
- [x] Cascader/Cascader.story.tsx #31399
- [x] ClipboardButton/ClipboardButton.story.internal.tsx #31422
- [x] ColorPicker/ColorPicker.story.tsx #31429
- [x] ColorPicker/NamedColorsPalette.story.tsx #31443
- [x] ConfirmButton/ConfirmButton.story.tsx #31476
- [x] ConfirmModal/ConfirmModal.story.tsx #31781
- [x] Drawer/Drawer.story.tsx #31821
- [x]  Dropdown/ButtonSelect.story.internal.tsx 
- [x] FormField/FormField.story.internal.tsx #32018
- [x] Forms/FieldValidationMessage.story.tsx #32020
- [x] Forms/Form.story.tsx #32022
- [x] Forms/Legacy/Input/Input.story.internal.tsx #32059
- [x] Forms/Legacy/Select/Select.story.internal.tsx #32225
- [x] Forms/Legacy/Switch/Switch.story.internal.tsx #32249
- [x] Forms/Legend.story.tsx #32250
- [x] Forms/RadioButtonGroup/RadioButtonGroup.story.tsx #32251
- [x] Graph/Graph.story.tsx #32264
- [x] Graph/GraphLegend.story.tsx (not found in the codebase)
- [x] Graph/GraphWithLegend.story.tsx #32285
[...]

PR #31145, opened 2021-02-11
Alert: update story to use controls

thisisobate:
This PR updates story to use controls instead of knobs
Relates to #30182


torkelo (commented):
This story was not using any knobs from what I can tell, so no need to adds controls here. But does not hurt I guess

torkelo (approved):
But works and looks good!

hugohaggmark:
Going through this today, I've noticed the following:

- I think `onRemove` & `onButtonClick` don't add anything to the story
- `buttonText` doesn't work, or I couldn't figure out how it worked
- `buttonContent`, seems strange you need to enter a string surrounded with `""`, can this be improved somehow?

image

thisisobate:
Alright @hugoh.
Would look into this.

PR #31149, opened 2021-02-11
Plugins: Migrate CloudWatch to backend plugin SDK

### LINKED ISSUE #31095: Plugins: Migrate CloudWatch to Go plugin SDK

wbrowne:
This continues our journey of migrating our core plugins to the Go SDK

Fixes #31095 

I have mostly tried to migrate things as is - however, there are a few areas that could be called into question and I'm happy to address them here as long as the scope doesn't blow up too much

Will definitely need your help @kaydelaney to test out the Live stuff 🙏


package.json
```diff
@@ -216,7 +216,6 @@
     "@types/react-virtualized-auto-sizer": "1.0.0",
     "@types/sockjs-client": "^1.1.1",
     "@types/uuid": "8.3.0",
-    "@types/hoist-non-react-statics": "3.3.1",
```
wbrowne, line 219: dupe - see line 212

public/app/plugins/datasource/cloudwatch/datasource.ts
```diff
@@ -590,9 +592,20 @@ export class CloudWatchDatasource extends DataSourceApi<CloudWatchQuery, CloudWa
     }));
   }
 
+  transformSuggestDataFromDataframes(suggestData: TSDBResponse): Array<{ text: any; label: any; value: any }> {
+    const frames = toDataQueryResponse({ data: suggestData }).data as DataFrame[];
+    const table = toLegacyResponseData(frames[0]) as TableData;
```
wbrowne, line 597: Not sure what's the best strategy here - `frames[0]` might be okay but not 100%

pkg/tsdb/cloudwatch/annotation_query.go
```diff
@@ -120,31 +107,32 @@ func (e *cloudWatchExecutor) executeAnnotationQuery(ctx context.Context, queryCo
...
-func transformAnnotationToTable(data []map[string]string, result *tsdb.QueryResult) {
-	table := &tsdb.Table{
-		Columns: make([]tsdb.TableColumn, 4),
-		Rows:    make([]tsdb.RowValues, 0),
+func transformAnnotationToTable(annotations []map[string]string, query backend.DataQuery) *data.Frame {
+	frame := data.NewFrame(query.RefID,
+		data.NewField("time", nil, []string{}),
+		data.NewField("title", nil, []string{}),
+		data.NewField("tags", nil, []string{}),
+		data.NewField("text", nil, []string{}),
```
marefr, line 122: Feels like these should be able to be typed? Might need update in client side code though. Not sure why this has been handled as all strings earlier since client side will cast it to correct types?

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -62,41 +68,97 @@ func init() {
...
 		logsService: logsService,
+		im:          im,
+	}
+}
+
+func NewInstanceSettings() datasource.InstanceFactoryFunc {
+	return func(settings backend.DataSourceInstanceSettings) (instancemgmt.Instance, error) {
+		var jsonData map[string]string
+
+		err := json.Unmarshal(settings.JSONData, &jsonData)
```
marefr, line 101: This is something I haven't really thought a lot of before. We would marshal the JSON here https://github.com/grafana/grafana/blob/a18e4d8d6437a0cf57e8ddebe61d36d5ce081e25/pkg/plugins/backendplugin/coreplugin/query_endpoint_adapter.go#L29

And then here we Unmarshal. Might need to think about how we can skip this. Maybe the answer is just to migrate away from tsdb.QueryEndpoint/models and we wouldn't need all of these wrappers/adapters from/to SDK/tsdb things. But definitely something to take into consideration what affect this might have in regards to performance and such during the migration path of things.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -62,41 +68,97 @@ func init() {
...
 	rgtaClient resourcegroupstaggingapiiface.ResourceGroupsTaggingAPIAPI
 
 	logsService *LogsService
+	im          instancemgmt.InstanceManager
 }
 
-func (e *cloudWatchExecutor) newSession(region string) (*session.Session, error) {
-	dsInfo := e.getDSInfo(region)
+func (e *cloudWatchExecutor) newSession(region string, pluginCtx backend.PluginContext) (*session.Session, error) {
+	dsInfo, err := e.getDSInfo(pluginCtx)
```
marefr, line 154: It feels a bit odd to have a separate session cache RWMutex. But not easy to migrate away from I definitely understand. Something that we might look into in the future. 

Could potentially build upon the idea of this [instance manager](https://github.com/grafana/grafana-plugin-sdk-go/blob/master/backend/instancemgmt/instance_manager.go) as a way to replace the RWMutex with to remove lock-contention for cloudwatch data source instances?

marefr (commented):
First couple of comments. Not really any blockers so far, more things that I found interesting to highlight.

pkg/tsdb/cloudwatch/query_transformer.go
```diff
@@ -93,10 +91,10 @@ func (e *cloudWatchExecutor) transformQueryResponsesToQueryResult(cloudwatchResp
 		})
 
 		if requestExceededMaxLimit {
-			queryResult.ErrorString = "Cloudwatch GetMetricData error: Maximum number of allowed metrics exceeded. Your search may have been limited."
+			queryResult.Error = fmt.Errorf("cloudwatch GetMetricData error: Maximum number of allowed metrics exceeded. Your search may have been limited")
```
sunker, line 94: The reason ErrorString and not Error was used here is because we want to display an error on the top left of the panel to indicate something went wrong, but at the same the we'd like the time series that we got from the API to be displayed in the panel. By using Error, nothing will be rendered in the panel - we will only get an error message. 

So really this is just a warning. When this was implemented about a year ago, the api didn't have any support to propagate warnings to the panel, but maybe there is a way of doing that today? 

This is pretty hard to reproduce, because in case this limit is reached, the partial data limit is also reached and in that case, the partial data error message will be displayed.

pkg/tsdb/cloudwatch/query_transformer.go
```diff
@@ -93,10 +91,10 @@ func (e *cloudWatchExecutor) transformQueryResponsesToQueryResult(cloudwatchResp
...
 		}
 		if partialData {
-			queryResult.ErrorString = "Cloudwatch GetMetricData error: Too many datapoints requested - your search has been limited. Please try to reduce the time range"
+			queryResult.Error = fmt.Errorf("cloudwatch GetMetricData error: Too many datapoints requested - your search has been limited. Please try to reduce the time range")
```
sunker, line 97: same here. This is easy to reproduce. Just add `SEARCH('', 'Minimum', 3000)` to the expression field.  This will query all metrics in all namespaces. It will probably hit thousands of metrics, but the API has a limit of 100, so only those will be returned.

pkg/tsdb/cloudwatch/time_series_query.go
```diff
@@ -4,126 +4,122 @@ import (
...
-	if len(requestQueriesByRegion) == 0 {
-		return &tsdb.Response{
-			Results: make(map[string]*tsdb.QueryResult),
-		}, nil
-	}
+	for _, q := range req.Queries {
+		startTime := q.TimeRange.From
+		endTime := q.TimeRange.To
+		if !startTime.Before(endTime) {
+			return nil, fmt.Errorf("invalid time range: start time must be before end time")
```
sunker, line 22: Maybe this is something that could be taken care of by the SDK? Is it even possible to achieve this using the timepicker? Maybe only by hacking the dashboard.json

sunker (dismissed):
Looks good! Tested and it works fine. Just need to show warnings correctly when only a part of the data was returned to the panel.

sunker (commented):
Annotations doesn't work for me, and that's because the annotations query response doesn't return table anymore. I think you need to do something similar to what @dsotirakis did for [cloud monitoring](https://github.com/grafana/grafana/blob/master/public/app/plugins/datasource/cloud-monitoring/datasource.ts#L81-L99). Hint: Since the cloudwatch datasource file doesn't implement DatasourceWithBackend, the response obj is lightly different than for CloudMonitoring, so I guess you'll have to pass the whole response to the toDataQueryResponse. Something like this:
```javascript
 map((data) => {
          const dataQueryResponse = toDataQueryResponse({ data });
```

FYI, I just created an alarm in cloudwatch, so if you create an annotation using the following params, it should pop up in the graph as long as the response is handled correctly. Let me know if you need help testing this. 
Screenshot from 2021-03-04 09-15-01

wbrowne:
@ryantxu Would appreciate your eyes since I had some conflicts since your changes was just merged. Also came across this bug https://github.com/grafana/grafana-aws-sdk/issues/10.

sunker (approved):
Tested annotations and it's working fine now. 

Great work!

sunker:
> @ryantxu Would appreciate your eyes since I had some conflicts since your changes was just merged. Also came across this bug [grafana/grafana-aws-sdk#10](https://github.com/grafana/grafana-aws-sdk/issues/10).

I took a look at the conflicts that you resolved and it looks good. Also tested list metrics page limit and cloudwatch auth and it's working fine.

pkg/tsdb/cloudwatch/cloudwatch.go
```diff
@@ -177,158 +267,119 @@ func (e *cloudWatchExecutor) alertQuery(ctx context.Context, logsClient cloudwat
...
-	if err != nil {
-		return plugins.DataResponse{}, err
-	}
+		region := model.Get("region").MustString(defaultRegion)
+		if region == defaultRegion {
+			dsInfo, err := e.getDSInfo(req.PluginContext)
+			if err != nil {
+				return nil, err
+			}
+			model.Set("region", dsInfo.region)
```
guillaumesmo, line 329: shouldn't this have been `model.Set("region", dsInfo.defaultRegion)` ?

PR #31152, opened 2021-02-11
Docs feedback: /sources/http_api/user.md

ademidoff:
I believe it's a mistake in the docs, since http requests against those endpoints are not supposed to be authenticated with API keys:

- `GET /api/user`
- `GET /api/user/orgs`
- `PUT /api/user/password`

If I correctly understand Grafana's auth design, the API key is not issued/tied to a user, therefore there is no mechanism to detect who the user is when the request is authenticated with an API key.

Please note there may be other endpoints like this...

### DIFF
--- a/docs/sources/http_api/user.md
+++ b/docs/sources/http_api/user.md
@@ -298,8 +298,10 @@ Content-Type: application/json
 GET /api/user HTTP/1.1
 Accept: application/json
 Content-Type: application/json
-Authorization: Bearer eyJrIjoiT0tTcG1pUlY2RnVKZTFVaDFsNFZXdE9ZWmNrMkZYbk
+Authorization: Basic YWRtaW46YWRtaW4=
 ```
+
+Requires basic authentication.
 
 **Example Response**:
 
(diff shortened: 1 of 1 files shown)


atymchuk:
@oddlittlebird Could someone review it?

achatterjee-grafana:
@marefr, can you look into this and confirm which docs need to be updated, if any. Thanks a lot!

atymchuk:
@torkelo
Q: it is true that by design we cannot use the API key to find out who the user is? If true, than consider these changes to the docs, since the docs say it's possible.

Maybe you can shed light?

torkelo:
Api keys are not linked to a user right now, only to an org

atymchuk:
> Api keys are not linked to a user right now, only to an org

Great, thanks! Then the docs need to updated I guess, since the quote below sets a wrong expectation:

> GET /api/user HTTP/1.1
Accept: application/json
Content-Type: application/json
Authorization: Bearer eyJrIjoiT0tTcG1pUlY2RnVKZTFVaDFsNFZXdE9ZWmNrMkZYbk

marefr (approved):
LGTM

PR #31156, opened 2021-02-12
Variables: Fix missing empty elements from regex filters

### LINKED ISSUE #31011: Empty elements from variable regex are not possible after upgrade 7.3.7 -> 7.4.0

We are distinguishing between 2 classes of servers via regex depending on the letter G as suffix:

Input: `servername-b2` vs `servername-b2G`
Regex: `servername-[a-c][0-9](G?)`

This returns either G or en empty element.

**What happened**:

After updating to 7.4.0 instead of the empty element all servernames without G were returned.

**Environment**:
- Grafana version: 7.4.0
- Data source type & version: PostgreSQL 12.5
- OS Grafana is installed on: Debian 10
- User OS & Browser: Win10 with Chrome/Firefox
- Grafana plugins: alexanderzobnin-zabbix-app
- Others:

hugohaggmark:
This fix for #31011 was straight down the regex rabbit's hole but in the end I think we now have feature parity with Grafana 7.3.7 again. I'll try to explain what has happened, bare with me.

In 7.4.0 I approved this PR #28625 because I thought it was a substantial addition to the Variable system. I think this was one of my first approvals of the Variable system and to be honest; I didn't do a superb job testing the PR.

I spent the entire day yesterday looking into the difference before and after #28625 and this is what I found:

Before #28625 we implicitly add all matches as options because if you enter the global flag `/g` in the regex field of a query variable we would implicitly reuse the `regex` object without resetting the `lastIndex`

Initialized once and never reset:
https://github.com/grafana/grafana/blob/1e261642f4bff28093de2c30093198fcd231c8a6/public/app/features/variables/query/reducer.ts#L93-L95
https://github.com/grafana/grafana/blob/1e261642f4bff28093de2c30093198fcd231c8a6/public/app/features/variables/query/reducer.ts#L112

This implicit reuse of `regex` works great as long as you don't use `/g` in your regex. And in some edge cases this implicit reuse actually leads to a bug in Grafana < 7.3.7 and rows were excluded because `lastIndex` was not reset and therefore the next match will not match. See this issue in Grafana 7.3.7 in movie below:
variables

So this PR tries to reuse the good parts from 7.3.7, i.e. the implicit finding of all matches with `/g`, but not include the issue where items are excluded as described above. 

This is the reason I introduced the extra loop here:
https://github.com/grafana/grafana/blob/hugoh/issue-variables-regex/public/app/features/variables/query/reducer.ts#L143-L147

I also removed the lodash merging and just pushed results to an array instead, this made the code a ~10x faster for 100000 items:
https://github.com/grafana/grafana/blob/8d339a279bedaa1fad8ab7257d852fdc32942314/public/app/features/variables/query/reducer.ts#L97

If you have any questions, please ping me and I'll try to explain.

Enjoy the review 🐰 

Fixes #31011

I also moved the tests from getAllMatches and added tests for metricNamesToVariableValues instead, because they seemed more valuable.

### DIFF
--- a/public/app/features/variables/query/reducer.test.ts
+++ b/public/app/features/variables/query/reducer.test.ts
@@ -346,7 +346,6 @@ describe('metricNamesToVariableValues', () => {
     { value: 'demo.robustperception.io:9100', text: 'demo.robustperception.io:9100', selected: false },
   ];
 
-  ///.*_(\w+)\{/
   it.each`
     variableRegEx                                          | expected
     ${''}                                                  | ${metricsNames}
--- a/public/app/features/variables/query/reducer.test.ts
+++ b/public/app/features/variables/query/reducer.test.ts
@@ -1,6 +1,7 @@
 import { reducerTester } from '../../../../test/core/redux/reducerTester';
 import {
   getAllMatches,
+  metricNamesToVariableValues,
   queryVariableReducer,
   sortVariableValues,
   updateVariableOptions,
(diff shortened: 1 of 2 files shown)


torkelo (approved):
Looks great!

PR #31161, opened 2021-02-12
Annotations API: also return dasboard uid

### LINKED ISSUE #31111: Annotations API should return dashboard UID
Instead of, or in addition to dashboardId, the annotations http api should return the dashboard UID

**Why is this needed**:
You cannot query dashboard info with dashboardID, only the UID. You cannot go to a dashboard url with the dashboardId. If I want to know what dashboard an annotation belongs to I have to manually go into every dashboard and look at the JSON model under settings.

cunyat:

Make Annotations API include dashboardUid, along with dashboardId.
This is requested in Issue #31111 because dashboardUid is needed for retrieve it from API 

Fixes: #31111

Not so much changes, tested it using postman.
I think thats not necessary to add a test for that..


pkg/services/annotations/annotations.go
```diff
@@ -89,22 +89,23 @@ func (i Item) TableName() string {
...
+	Created      int64            `json:"created"`
+	Updated      int64            `json:"updated"`
+	Time         int64            `json:"time"`
+	TimeEnd      int64            `json:"timeEnd"`
+	Text         string           `json:"text"`
+	Tags         []string         `json:"tags"`
+	Login        string           `json:"login"`
+	Email        string           `json:"email"`
+	AvatarUrl    string           `json:"avatarUrl"`
+	Data         *simplejson.Json `json:"data"`
```
jayaddison, line 110: Note to any other reviewers: this change is best viewed with whitespace differences ignored (link: [example GitHub diff view with whitespace differences ignored](https://github.com/grafana/grafana/pull/31161/files?w=1)).

The only field changed in the list here is the addition of a `DashboardUid` field as described in the intent of the task.

pkg/services/sqlstore/annotation.go
```diff
@@ -127,6 +127,7 @@ func (r *SQLAnnotationRepo) Find(query *annotations.ItemQuery) ([]*annotations.I
 			annotation.epoch as time,
 			annotation.epoch_end as time_end,
 			annotation.dashboard_id,
+			dashboard.uid as dashboard_uid,
```
jayaddison, line 130: This is a small detail, @cunyat but it occurred to me from writing and reading lots of SQL in the past:

It looks like the previous SQL authors here have attempted to group all of the columns in this file by the table name.

If so, it might be a slight consistency improvement to move this `dashboard.uid` line after the `annotation.updated` line.

pkg/services/sqlstore/annotation.go
```diff
@@ -142,6 +143,7 @@ func (r *SQLAnnotationRepo) Find(query *annotations.ItemQuery) ([]*annotations.I
 		FROM annotation
 		LEFT OUTER JOIN ` + dialect.Quote("user") + ` as usr on usr.id = annotation.user_id
 		LEFT OUTER JOIN alert on alert.id = annotation.alert_id
+		LEFT OUTER JOIN dashboard on dashboard.id = annotation.dashboard_id
```
jayaddison, line 146: Since we're introducing another table in the join, there's a small chance of query performance impact here.  I think that's unlikely, since the `dashboard.id` field should (?) be indexed.

jayaddison (commented):
I'm not an approved reviewer for this project, so my input holds little sway - but this looks good to me, with the one note to keep in mind regarding (potential, not expected, and likely tunable even if issues do arise) impact on database query performance.

ryantxu:
I would like to figure out how we can remove `id` from all our APIs and only use `uid`:
https://github.com/grafana/grafana/issues/32853

pkg/services/sqlstore/annotation.go
cunyat: Thanks for your feedback! dashboard id is the primary key, so it should be indexed :)

jayaddison: Good stuff, thanks :)  Reading back I might have added as much noise as useful commentary.. but I hope some of it has been useful.

pkg/services/annotations/annotations.go
```diff
@@ -98,13 +98,13 @@ type ItemDTO struct {
...
 	AlertName    string           `json:"alertName"`
 	DashboardId  int64            `json:"dashboardId"`
-	DashboardUid string           `json:"dashboardUid"`
 	PanelId      int64            `json:"panelId"`
 	UserId       int64            `json:"userId"`
 	NewState     string           `json:"newState"`
 	PrevState    string           `json:"prevState"`
 	Created      int64            `json:"created"`
 	Updated      int64            `json:"updated"`
+	DashboardUid string           `json:"dashboardUid"`
```
jayaddison, line 107: Nitpick: the [relocation of the `dashboard.uid` field in the `sqlstore` submodule](https://github.com/grafana/grafana/pull/31161/commits/b4ea0c7a341d2e7df3d3ba2d554bdf4626c26f9b#diff-5a382284915b31e0cfcc606446bcd3e38ad97c26fce35af6081bbdbc8c8c8d0aR139) looks good, although in this struct perhaps it makes sense to keep `DashboardUid` on the line after `DashboardId` (again, some kind of grouping-by-convention).

cunyat: Yeah, i put that there in the first commit, but i did not know if the ItemDTO must have the same order of the query. I saw that's not necessary, thanks for that point ;)

ying-jeanne:
@zserge hey serge, would this be a apart of API clean up that you are doing?

cunyat:
So, can we do anything here?

Maybe we could add the uuid directly into the annotations table? I don't know the impact that this would have..
If there is anything I can do, please ask for it :)

